PR #1, opened 2021-11-02
Added some plugins

Builditluc:
I've added `colorschemes/tokyonight` and `statuslines/lualine`.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -1,0 +1,15 @@
+{ pkgs, config, lib, ... }:
+with lib;
+let cfg = config.programs.nixvim.plugins.lualine;
+in {
+  options = {
+    programs.nixvim.plugins.lualine = {
+      enable = mkEnableOption "Enable airline";
+    };
+  };
+  if mkIf cfg.enable {
+    programs.nixvim = {
+      extraPlugins = with pkgs.vimPlugins.lualine-nvim;
+    };
+  };
+}
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -1,15 +1,129 @@
 { pkgs, config, lib, ... }:
 with lib;
-let cfg = config.programs.nixvim.plugins.lualine;
+let
+  cfg = config.programs.nixvim.plugins.lualine;
+  helpers = import ../helpers.nix { lib = lib; };
+  separators = mkOption {
+    type = types.submodule {
+      options = {
+        left = mkOption {
+          default = " ";
+          type = types.str;
+          description = "left separator";
+        };
+        right = mkOption {
+          default = " ";
+          type = types.str;
+          description = "right separator";
+        };
+      };
+    };
+    default = { };
+  };
+  component_options = mode:
+    mkOption {
+      type = types.submodule {
+        options = {
+          mode = mkOption {
+            type = types.str;
+            default = "${mode}";
+          };
+          icons_enabled = mkOption {
+            type = types.enum [ "True" "False" ];
+            default = "True";
+            description = "displays icons in alongside component";
+          };
+          icon = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = "displays icon in front of the component";
+          };
+          separator = separators;
+        };
+      };
+      default = { };
+    };
 in {
   options = {
     programs.nixvim.plugins.lualine = {
       enable = mkEnableOption "Enable airline";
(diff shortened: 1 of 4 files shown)


pta2002:
Thanks! Can you just show me an example of how you'd use lualine?

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+          };
+          separator = separators;
+        };
+      };
+      default = { };
+    };
+in {
+  options = {
+    programs.nixvim.plugins.lualine = {
+      enable = mkEnableOption "Enable airline";
```
pta2002, line 49: This should say enable lualine

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+  component_options = mode:
+    mkOption {
+      type = types.submodule {
+        options = {
+          mode = mkOption {
+            type = types.str;
+            default = "${mode}";
+          };
+          icons_enabled = mkOption {
+            type = types.enum [ "True" "False" ];
```
pta2002, line 32: This should be a boolean! Is there any reason in particular you did it this way?

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+        default = [ ];
+        example = ''[ "fzf" ]'';
+        description = "list of enabled extensions";
+      };
+    };
+  };
+  config = mkIf cfg.enable {
+    programs.nixvim = {
+      extraPlugins = [ pkgs.vimPlugins.lualine-nvim ];
+      extraConfigLua = ''require("lualine").setup{''
```
pta2002, line 117: Can probably just use the helper function to convert to a lua object here, instead of doing it line by line :)

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+        };
+      };
+      default = { };
+    };
+in {
+  options = {
+    programs.nixvim.plugins.lualine = {
+      enable = mkEnableOption "Enable airline";
+
+      options = mkOption {
```
pta2002, line 51: I'll remove these into the top-level, don't think there's a need to put this into a submodule

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+        type = types.submodule {
+          options = {
+            theme = mkOption {
+              default = "auto";
+              type = types.str;
+              description = "The theme to use for lualine-nvim.";
+            };
+            section_separators = separators;
```
pta2002, line 59: I'll also rename these so that they're camelCase, more in line with the rest of the project

plugins/statuslines/lualine.nix
```diff
@@ -0,0 +1,129 @@
...
+            };
+          };
+        };
+        default = { };
+        description = "Options for lualine";
+      };
+      sections = mkOption {
+        type = types.nullOr (types.submodule ({ ... }: {
+          options = {
+            lualine_a = component_options "mode";
```
pta2002, line 81: As for these, I think we can just make them a,b,c etc., and drop the lualine_ prefix

pta2002 (commented):
Found some things to improve here, specifically on the lualine plugin. But I'm already on it! I'll take care of those and then I'll get this merged ;) Good job!

plugins/statuslines/lualine.nix
Builditluc: Somehow, when I make it a boolean neovim says it can't resolve this component.

Builditluc:
Thanks!

Builditluc:
I'll write some documentation for these plugins now

Builditluc:
> Thanks! Can you just show me an example of how you'd use lualine?

This is my configuration for lualine (still outdated):
```
plugins.lualine = {
  enable = true;
  options = {
    component_separators = {
      left = "";
      right = "";
    };
    section_separators = {
      left = "";
      right = "";
    };
  };
  sections = {
    lualine_a = {
      mode = "branch";
      icon = "Óú•";
    };
    lualine_b.mode = "filename";
    lualine_c.mode = "";
    lualine_x.mode = "filetype";
    lualine_y.mode = "";
    lualine_z.mode = "mode";
  };
};
```

plugins/statuslines/lualine.nix
pta2002: Interesting, I'll try it out

PR #4, opened 2021-11-23
add bufferline and nvim-lsp/pyright

Builditluc:
I've added the plugin bufferline and nvim-lsp/pyright and the capabilities option for nvim-lsp (will use that later for cmp-nvim)

### DIFF
--- a/plugins/bufferlines/bufferline.nix
+++ b/plugins/bufferlines/bufferline.nix
@@ -1,0 +1,177 @@
+{ config, pkgs, lib, ... }:
+with lib;
+let
+  cfg = config.programs.nixvim.plugins.bufferline;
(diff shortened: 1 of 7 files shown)


Builditluc:
I've also added comment-nvim (another Neovim comment plugin)

pta2002:
Hey, this looks great! Just a thing I'd change though: the default on the options should be, in most cases, null, so that we're not filling up the config with needless variable declarations. I can change that, if you want. I have the `helpers.mkPlugin` interface to help with this, which should work for most plugins that only need to define options as global variables.

Builditluc:
Okay, I'll change that

Builditluc:
All of the options are now null by default

pta2002 (approved):
Looks good!

PR #6, opened 2021-12-12
Add mark-radar and specs.nvim plugins

legendofmiracles:
- specs.nvim: init
- mark-radar: init

### DIFF
--- a/plugins/utils/mark-radar.nix
+++ b/plugins/utils/mark-radar.nix
@@ -1,0 +1,48 @@
+{ pkgs, config, lib, ... }:
+
+with lib;
+let
+  cfg = config.programs.nixvim.plugins.mark-radar;
+  helpers = import ../helpers.nix { inherit lib; };
+  defs = import ../plugin-defs.nix { inherit pkgs; };
+in {
+  options.programs.nixvim.plugins.mark-radar = {
+    enable = mkEnableOption "Enable mark-radar";
+
+    highlight_background = mkOption {
+      type = with types; nullOr bool;
+      default = null;
+    };
+
+    background_highlight_group = mkOption {
+      type = with types; nullOr str;
+      default = null;
+    };
+
+    highlight_group = mkOption {
+      type = with types; nullOr str;
+      default = null;
+    };
+
+    set_default_keybinds = mkOption {
+      type = with types; nullOr str;
+      default = null;
+    };
+  };
+
+  config = let
+    opts = helpers.toLuaObject {
+      inherit (cfg) highlight_group background_highlight_group;
+      set_default_mappings = cfg.set_default_keybinds;
+      background_highlight = cfg.highlight_background;
+    };
+  in mkIf cfg.enable {
+    programs.nixvim = {
+      extraPlugins = [ defs.mark-radar ];
+
+      extraConfigLua = ''
+        require("mark-radar").setup(${opts})
+      '';
+    };
+  };
+}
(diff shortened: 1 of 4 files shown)


plugins/utils/mark-radar.nix
```diff
@@ -0,0 +1,25 @@
...
+with lib;
+let
+  cfg = config.programs.nixvim.plugins.mark-radar;
+  helpers = import ../helpers.nix { inherit lib; };
+  defs = import ../plugin-defs.nix { inherit pkgs; };
+in
+{
+  options.programs.nixvim.plugins.mark-radar = {
+    enable = mkEnableOption "Enable mark-radar";
+  };
```
pta2002, line 12: I'd add the other mark-radar options here! Though I think it's fine to merge this for now, and then I can just add the rest.

plugins/utils/specs.nix
```diff
@@ -0,0 +1,125 @@
...
+      type = with types; listOf string;
+      default = [ "nofile" ];
+    };
+
+  };
+  config = mkIf cfg.enable {
+    programs.nixvim = {
+      extraPlugins = [ pkgs.vimPlugins.specs-nvim ];
+
+      extraConfigLua = ''
```
pta2002, line 107: Maybe move this out to a `let ... in` declaration, so that then it can just be

```lua
require('specs').setup(${setup})
```

pta2002 (changes requested):
Only some minor changed I'd do here :) I can do it if you give me push access. Good job and thanks for the PR!

plugins/utils/mark-radar.nix
legendofmiracles: I'm not sure what other options could be added:
```
local opts = {
    set_default_mappings = true,
    highlight_group = "RadarMark",
    background_highlight = true,                    
    background_highlight_group = "RadarBackground", 
}
```
None of these seem like anything you'd want to override. And if you set `set_default_mappings` to false, there's no way to change it to something else.

pta2002: Sure, but I'd like to expose the options anyway just to be on the safe side of things, even if they are completely sensible defaults that 99% of people wouldn't want to mess with.

pta2002 (commented):
I'd still at the mark-radar config options just to be on the safe side of things. I agree with you they're completely sensible defaults, and they're something 99% of people will have no interest in changing, but just to be on the safe side I like to expose all options upstream provides!

plugins/utils/mark-radar.nix
legendofmiracles: Done.

plugins/utils/mark-radar.nix
```diff
@@ -0,0 +1,48 @@
...
+with lib;
+let
+  cfg = config.programs.nixvim.plugins.mark-radar;
+  helpers = import ../helpers.nix { inherit lib; };
+  defs = import ../plugin-defs.nix { inherit pkgs; };
+in {
+  options.programs.nixvim.plugins.mark-radar = {
+    enable = mkEnableOption "Enable mark-radar";
+
+    highlight_background = mkOption {
```
pta2002, line 12: Sorry for continuing to press on this 😅 but it's best to make these options default to null (type being types.nullOr [original type]), to avoid filling up the init file with unnecessary options. I can take care of this tonight though!

legendofmiracles: No, you're perfectly right! I've pushed a fix. 
Please do keep reviewing if you find something off.

PR #9, opened 2021-12-18
which-key: init plugin

Builditluc:
which-key: init plugin

example of how you can use it: 
https://github.com/Builditluc/dotfiles/blob/8223b2051d89aabf8ac81716dde8d0d6ca93e53b/.config/nixpkgs/hm/nvim.nix#L99

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -31,6 +31,7 @@
     ./utils/intellitab.nix
     ./utils/specs.nix
     ./utils/mark-radar.nix
+    ./utils/which-key.nix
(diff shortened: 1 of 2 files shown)


pta2002:
Sorry for taking so long to review this. I held off on which-key for a long time because I *really* wanted to integrate it with the default nixvim mappings, but I just didn't really know how to do it. I think I will try to do that integration before I merge this though.

Builditluc:
Since it is no longer needed, I'll close this PR

PR #11, opened 2021-12-23
better manage telescope

Builditluc:
I've added a very simple plugin called notify.

Telescope now has it's own folder in `plugins` where it's extensions are defined. You can write extensions for telescope in the same way, the lsp plugins are written.


pta2002:
This looks good! Thank you :)

PR #13, opened 2022-01-17
add nvim-lsp/zls support

imMaturana:
Recently I found Zig and it is a very interesting programming language and one thing I was missing in nixvim was LSP support for it, so I've added.


pta2002:
Oh thanks! I didn't create this at the time I created the zig plugin because IIRC zls wasn't on NixOS yet, glad to see it is now :) Looks good!

PR #23, opened 2022-07-13
added nord colorscheme

Alexnortung:
This PR adds the nord colorscheme as a module with the [`nord-nvim`](https://github.com/shaunsingh/nord.nvim) package

### DIFF
--- a/nixvim.nix
+++ b/nixvim.nix
@@ -204,13 +204,16 @@ in
       });
 
       configure = {
-        customRC = cfg.extraConfigVim + (optionalString (cfg.colorscheme != "" && cfg.colorscheme != null) ''
-          colorscheme ${cfg.colorscheme}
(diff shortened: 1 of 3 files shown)


Alexnortung:
Actually there is an issue with this PR.
The options don't really work since they need to be set before calling `colorscheme nord`. Is there a way we can do this?

pta2002:
Hm, I think you can change nixvim.nix to add the colorscheme call after Lua
- I don't think that'll cause an issue with anything else so I say you can
do it, probably for the best too

A quarta, 13/07/2022, 17:22, Alexander Nortung ***@***.***>
escreveu:

Alexnortung:
I have changed it now and it works with my current config :)

pta2002:
Looks good, thanks!

PR #24, opened 2022-07-13
nvim-tree: git.enable expects a boolean

Alexnortung:
Currently the options `plugins.nvim-tree.git.enable` expects null or a string.
However, this is not correct and results in an error when starting neovim, when nvim-tree is enabled.

This PR changes the option to expect a boolean.


Alexnortung:
@pta2002 if you have time, can you review this one line modification?

pta2002:
Seems good! Thanks

PR #25, opened 2022-07-17
Adding nvim-cmp plugin

Alexnortung:
This pull request is a draft just to share my initial work on this.

As a user of nvim-cmp, I think it would be nice to have cmp as a module which this pull request aims to add.

If you have any inputs to this or any suggestions on how to improve it, please add a comment.

I will probably finish this over the next couple of days.

Config options completed:
- [x] enabled
- [x] performance
- [x] preselect
- [x] mapping
- [x] snippet
- [x] completion
- [x] confirmation
- [x] formatting
- [x] matching
- [x] sorting
- [x] sources
- [x] view
- [x] window
- [x] experimental

### DIFF
--- a/plugins/completion/nvim-cmp/cmp-helpers.nix
+++ b/plugins/completion/nvim-cmp/cmp-helpers.nix
@@ -1,0 +1,54 @@
+{ lib, pkgs, ... }@attrs:
+let
+  helpers = import ../../helpers.nix { lib = lib; };
+in with helpers; with lib;
+{
+  mkCmpSourcePlugin = { name, extraPlugins ? [], useDefaultPackage ? true, ... }: mkPlugin attrs {
+    inherit name;
+    extraPlugins = extraPlugins ++ (lists.optional useDefaultPackage pkgs.vimPlugins.${name});
+    description = "Enable ${name}";
+  };
+
+  pluginAndSourceNames = {
+    "luasnip" = "cmp_luasnip";
+    "snippy" = "cmp-snippy";
+    "ultisnips" = "cmp-nvim-ultisnips";
+    "vsnip" = "cmp-vsnip";
+    "buffer" = "cmp-buffer";
+    "calc" = "cmp-calc";
+    "dictionary" = "cmp-dictionary";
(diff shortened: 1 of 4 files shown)


Alexnortung:
In [a6be8f0](https://github.com/pta2002/nixvim/pull/25/commits/a6be8f0c786322551ef46aa63cda89e9e7c8967a) I have added a lot of cmp source plugins.
The goal was to make nixvim automatically detect if a source was added to the source array and then enable the plugin. However this lead to infinite recursion for some reason and I have no idea why, I hope someone can take a look at it.
Please let me know if you have questions.

pta2002:
Hi! I'd love to have nvim-cmp on NixVim, actually setting up LSP and autocomplete automatically was what made me create it in the first place, just never got around to finishing support... 😅 So this is definitely extremely welcome!

Unfortunately I can't review this right now though, as I'm away for vacation and will be for the next two weeks, but I'll make sure to check it out. If you can get it working and can confirm it works, I'll merge it though!

Alexnortung:
Everything should work now, I haven't tested out all options, since I am not completely sure on some of them.
Source plugins will be automatically installed, but weirdly enough from the plugins that I tested it cannot find the cmp-nvim-lsp plugin.

I think this must be a problem with the nix package of the plugin.

Output from `:CmpStatus`

```
# ready source names
- buffer
- path
- luasnip

# unknown source names
- nvim_lsp
```

But the plugin is installed as I can see from the following output

```
$ ls /nix/store/vyjg8p1cn68yd633ksl63x6q3qhilizf-vim-pack-dir/pack/home-manager/start/ | grep cmp-nvim-lsp
cmp-nvim-lsp
```

Anyway I think the module is fine and is ready for merge, but I might tidy up some of the code.

pta2002:
Could you maybe try using `cmp-nvim-lsp` as the source name?

Alexnortung:
I tried with `cmp-nvim-lsp` and `nvim-lsp`, but I just get the same output from `:CmpStatus`

pta2002:
Alright then, gonna merge this, but please open an issue for the non working plugin.

PR #28, opened 2022-07-24
Refactor language servers and add eslint, jsonls, cssls and html servers.

Alexnortung:
This PR simplifies the basic language servers, by making a new file called `basic-servers.nix`.

If any of the servers will support any new options we can always make new files again.

Other than that I have added the basic language servers for eslint, css, html and json.


pta2002:
Looks good, but why are there no packages specified for some servers?

Alexnortung:
In https://github.com/pta2002/nixvim/pull/28/files#diff-b20ff61378d889cea46c37f0dd1427c31fc7b9a7ebe1765a478527fef74bd280R8 it sets a default package which will just be the name of the server.

But for servers that have `packages = []` it is because they don't need any packages to work. For example, gdscript language server will just connect to Godot through a localhost port.

pta2002:
Got it! Looks good then, thanks!

PR #29, opened 2022-07-24
telescope.media_files: init

Alexnortung:
Just sharing my initial work on telescope media_files extension.
The only problem with this is that the plugin does not exist in nixpkgs yes, so we can either wait for that or depend on https://github.com/m15a/nixpkgs-vim-extra-plugins
But I am marking it as draft until it is added in nixpkgs or we decide to depend on vim-extra-plugins.

A word of appreciation:
A nice thing about nixvim is that we can directly add dependencies such as `überzug` which makes it much easier for the user to configure and not have to manually add it to their list of packages.

Issue in nixpkgs:
NixOS/nixpkgs#182734

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -9,6 +9,7 @@ in
     ./frecency.nix
     ./fzf-native.nix
     ./fzy-native.nix
+    ./media-files.nix
   ];
 
   # TODO:add support for aditional filetypes. This requires autocommands!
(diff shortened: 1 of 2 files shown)


pta2002:
Hey, this plugin is now in nixpkgs, so feel free to get back on this!

pta2002 (approved):
Tried it out and it seems to work!

Issue #16, opened 2022-04-18
use without HM or nixos


edrex:
I would like to use nixvim in a prototype runnable flake "env" project https://github.com/aquiferma/noter (see discussion [^0]). For this, I need a `myNeoVim` package with config/customizations baked in. 

There is a broader design need at the ecosystem level to define an ergonomic relationship between flakes and the module system outside the context of HM/nixos, but nixvim seems an OK place to pick up the conversation.

It seems like a "quick fix" might be to inline or maybe wrap one of the `programs.neovim` modules [^2][^3]. 

Also probably worth a look, @syberant's nix-neovim flake [^4], which IIUC inlines its own module system for config, and only calls `pkgs.wrappedNeoVim`.

[^0]: https://github.com/aquiferma/noter/issues/2
[^1]: https://github.com/pta2002/nixvim/blob/main/nixvim.nix
[^2]: https://github.com/nix-community/home-manager/blob/master/modules/programs/neovim.nix
[^3]: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/neovim.nix
[^4]: https://github.com/syberant/nix-neovim/blob/main/flake.nix

pta2002:
Hey, thanks for the pointers! This is actually something I've wanted to do for a while, but never quite figured out/had the patience to look into. It's definitely doable, so I'll look into what you suggested!

syberant:
Hi! My flake does indeed use a separate module system for config (so no `program.nix-neovim.vim.g.showmode = false` but just `vim.g.showmode = false`).
Doing this is surprisingly easy as you can reuse the library functions implementing NixOS's module system (extra benefit: you can also automatically generate documentation a la `man configuration.nix`).
The relevant piece from my `flake.nix` is this part:
```nix
buildNeovim = { pkgs ? dpkgs, configuration }:
  let
    eval = evalModules {
      modules = getNixFiles ./modules ++ [
        (rec {
          _file = ./flake.nix;
          key = _file;
          config = {
            _module.args.pkgs = mkForce pkgs;
            _module.args.vimLib = import ./lib { lib = pkgs.lib; };
          };
        })

        configuration
      ];
    };
  in eval.config.product.binary;
```

Here `configuration` is your entry point configuration module, `evalModules` is just `pkgs.lib.evalModules` and `product.binary` is the [read-only option that contains the final executable](https://github.com/syberant/nix-neovim/blob/main/modules/product.nix).

[...]

PR #34, opened 2022-08-23
[WIP] Support standalone nixvim

pta2002:
This represents a major rearchitecture for nixvim, so I'm leaving this up to track the progress for now, and to serve as a reference for any breaking changes during transition.

The main change is, of course, being able to use nixvim standalone. To do this, you should use the new `build` function, which takes in two arguments: the system architecture (e.g. `x86_64-linux`) and the configuration. For the new configuration, do not use the `programs.nixvim.` prefix.

For module development, the main change is that you should no longer prefix your modules with `programs.nixvim.`.

Checklist:

- [x] support build function
- [x] support home-manager
- [x] support raw NixOS modules
- [x] update documentation and README to reflect changes
- [ ] get documentation to build
- [x] properly support all systems when using `homeManagerModules/nixosModules.nixvim`

P.S: Addresses #16


pta2002:
For some reason, I can't seem to get the NixOS and home-manager modules working - they build the package, but it does not seem to read any config.

docs/default.nix
```diff
@@ -39,12 +34,15 @@ let
...
+        <d:tocentry xmlns:d="http://docbook.org/ns/docbook" linkend="book-nixvim-manual"><?dbhtml filename="index.html"?>
+          <d:tocentry linkend="ch-options"><?dbhtml filename="options.html"?></d:tocentry>
+          <d:tocentry linkend="ch-release-notes"><?dbhtml filename="release-notes.html"?></d:tocentry>
+        </d:tocentry>
+        </toc>
     '';
   };
-in docs.html
+in
+# TODO: Parse this json or something, since docbook isn't working (and it's kind of terrible anyway)
```
examosa, line 46: Have you considered alternative options for generating docs? Here are a few that I've stumbled across:

- [coricamu](https://github.com/danth/coricamu/)
- [styx](https://github.com/divnix/styx/)

I have also seen some projects using `mdbook` ([example](https://github.com/divnix/std)).

pta2002: From what I can tell, those are generic static site generators and not necessarily made for nix modules, like `nmd` is.

PR #38, opened 2022-09-01
treesitter: add ignoreInstall option to treesitter config

ForsakenHarmony:
Necessary for me because of https://github.com/claytonrcarter/tree-sitter-phpdoc/issues/15

### DIFF
--- a/plugins/languages/treesitter.nix
+++ b/plugins/languages/treesitter.nix
@@ -15,9 +15,15 @@ in
         description = "Install grammars with Nix (beta)";
       };
       ensureInstalled = mkOption {
-        type = with types; oneOf [ (enum [ "all" "maintained" ]) (listOf str) ];
-        default = "maintained";
-        description = "Either \"all\", \"maintained\" or a list of languages";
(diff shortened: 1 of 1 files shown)


plugins/languages/treesitter.nix
```diff
@@ -15,9 +15,15 @@ in
         description = "Install grammars with Nix (beta)";
       };
       ensureInstalled = mkOption {
-        type = with types; oneOf [ (enum [ "all" "maintained" ]) (listOf str) ];
-        default = "maintained";
```
ForsakenHarmony, line 19: `maintained` also seems like it no longer exists?

pta2002: Yep, it no longer exists! Haven't updated this module in a while though, should get rid of "maintained" and default to "all"

ForsakenHarmony: Do you not use `treesitter`? what's your alternative?

pta2002: Oh I do, just had it set to all in my personal config and didn't notice the change

PR #44, opened 2022-10-07
Make sure the runtimepath is set in NixOS 22.05

### LINKED ISSUE #40: Broken config on nixos-stable
Hello, I have just tried to update nixvim in my personal flake. However it seems to be broken, when using the home manager module.

I am getting the following errors

```
Error detected while processing /nix/store/hgclrlyxzazx8b5324qcrkrbm5zfjq25-init.vim:
line    6:
E185: Cannot find color scheme 'nord'
line  396:
E5108: Error executing lua [string ":lua"]:20: module 'cmp' not found:
        no field package.preload['cmp']
        no file '/nix/store/vmln39q308xpq0awd1b5j84i8l58l857-luajit-2.1.0-2022-04-05-env/share/lua/5.1/cmp.lua'
        no file '/nix/store/vmln39q308xpq0awd1b5j84i8l58l857-luajit-2.1.0-2022-04-05-env/share/lua/5.1/cmp/init.lua'
        no file '/nix/store/vmln39q308xpq0awd1b5j84i8l58l857-luajit-2.1.0-2022-04-05-env/lib/lua/5.1/cmp.so'
stack traceback:
        [C]: in function 'require'
        [string ":lua"]:20: in main chunk
```

I also noticed that the config file is no longer saved in the `$HOME/.config/` directory, why is that?

Anyway, I will continue using an older version until we can fix some of these issues.

dwarfmaster:
In NixOS 22.05, the runtimepath is set depending on the `configure.packages` argument to `makeNeovimConfig`. This argument has since been removed and the runtimepath is now set by an argument in the wrapper. To support 22.05 we check for its presence using `functionArgs` and set it appropriately.

Should fix #40

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -91,10 +91,17 @@ in
 
       normalizedPlugins = map (x: defaultPlugin // (if x ? plugin then x else { plugin = x; })) config.extraPlugins;
 
-      neovimConfig = pkgs.neovimUtils.makeNeovimConfig {
+      neovimConfig = pkgs.neovimUtils.makeNeovimConfig ({
         inherit customRC;
         plugins = normalizedPlugins;
-      };
+      }
+      # Necessary to make sure the runtime path is set properly in NixOS 22.05,
+      # or more generally before the commit:
+      # cda1f8ae468 - neovim: pass packpath via the wrapper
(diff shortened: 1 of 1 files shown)


pta2002 (approved):

pta2002:
Just confirmed this fixes issue #40. Thanks!

PR #47, opened 2022-10-13
lualine: fix components config

PhotonQuantum:
The original implementation has several flaws:
- Each section may have only one component. However, lualine supports multiple components in the same section. E.g. `lualine_b = {'branch', 'diff', 'diagnostics'},` in default settings.
- The output format is incorrect. It outputs `{ mode = 'filename', icons = ... }` while lualine expects something like `{ 'filename', file_status = true, ... }`.
- There are many extra settings lualine accepts other than the ones specified in the spec. Adding them to the nixvim managed config results in a type error.

The new implementation changes the accepted section type from `submodule` to `listOf (oneOf str submodule)`, generates the correct output, and adds a new `extraConfig` field for unknown upstream flags.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -20,27 +20,36 @@ let
     });
     default = null;
   };
-  component_options = mode:
+  component_options = defaultName:
     mkOption {
-      type = types.nullOr (types.submodule {
-        options = {
-          mode = mkOption {
-            type = types.str;
-            default = "${mode}";
+      type = types.nullOr (types.listOf (types.oneOf [
+        types.str
+        (types.submodule {
+          options = {
+            name = mkOption {
+              type = types.str;
+              description = "component name";
+              default = defaultName;
+            };
+            icons_enabled = mkOption {
+              type = types.enum [ "True" "False" ];
+              default = "True";
+              description = "displays icons in alongside component";
+            };
+            icon = mkOption {
+              type = types.nullOr types.str;
+              default = null;
+              description = "displays icon in front of the component";
+            };
+            separator = separators;
+            extraConfig = mkOption {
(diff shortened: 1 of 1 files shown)


plugins/statuslines/lualine.nix
```diff
@@ -20,27 +20,35 @@ let
...
-          };
-          icon = mkOption {
-            type = types.nullOr types.str;
-            default = null;
-            description = "displays icon in front of the component";
+      type = types.nullOr (types.listOf (types.oneOf [
+        types.str
+        (types.submodule {
+          options = {
+            name = mkOption {
```
dwarfmaster, line 29: We probably want to keep the default name/mode that were used before so that it keep working for people who do not configure lualine in details.

PhotonQuantum: 'mode' is just the name of a builtin component. I believe this is misleading.

I changed the name because no existing config could ever rely on the old implementation. Anyone setting the 'mode' field will simply find lualine failing. So I changed its misleading name.

But sure it's good to keep backward compatibility! What about adding a 'mode' alias to the 'name' field?

dwarfmaster: Sorry I wasn't clear, my point wasn't about wether it is named name or mode. The problem I was raising is that it had a default value before, so the user that doesn't care much doesn't have to set it. Changing the name of the option is fine from my point of view.

PhotonQuantum: Ok I get your point. I'm adding it back.

dwarfmaster:
I haven't tested the new options, but I can confirm it doen't break my config, so LGTM :+1:

pta2002 (approved):

pta2002:
Looks good, thanks for this!

PR #48, opened 2022-10-13
Allow using global config with hm and nixos

dwarfmaster:
Refactor a bit the nixvim module so that the NixOS and home-manager can set the config by a global file. To do so, the module now includes a `wrapRc` setting defaulting to `false` that control wether the config file is wrapped with the `nvim` script or not. It also includes an internal option `initContent` that contains the final contents of the `init.vim` file. If `wrapRc` is false, the `initContent` is written to `$XDG_CONFIG_HOME/nvim/init.vim` with hm or `/etc/nvim/sysinit.vim` (and $VIM is set appropriately) with NixOS.

`wrapRc` can also be set to true to get back the current mecanism.

Finally, the previous `output` config is renamed to `finalPackage` and made visible, to make it more consistent with other modules.

The motivation is that according to the `makeNeovimConfig` in nixpkgs :
> Indeed, note that wrapping with `-u init.vim` has sideeffects like .nvimrc wont be loaded
> anymore, $MYVIMRC wont be set etc

so it is preferable to avoid when possible.


pta2002 (approved):

pta2002:
I like this! I think this could be expanded too to actually use a lua init file eventually, good job and thanks!

Issue #52, opened 2022-10-18
commit `a38d531b` breaks configs that declare variables in pre and uses it in post.


Alexnortung:
**Plugin affected**: all?
**Nixpkgs channel**: 22.05

## Description

commit a38d531bf4d02974d9e3088146cbaf896f2fcb00 breaks my config. I am using the `extraConfigLuaPre` options, where I define some variables. In `extraConfigLuaPost` I am using those variables. But in the aforementioned commit it seems to break this funcitonality, since the lua part is now split in two.

I have tried to change my config, to use global variables (not sure if I am doing it correctly), but then I am having issues with my cmp mappings.

I can see why it was chosen to be separated into two different lua sections and I am not sure if we should do something about it.

## Config

[nixvim config](https://github.com/Alexnortung/my-nixos-systems/tree/main/profiles/nvim)

Alexnortung:
I just converted my config to use global variables in the pre config, and then renaming my `cmp` variable and that seems to fix it.

pta2002:
I'm not really sure how desirable it can be to fix this, I think the alternative fix to preserve backwards-compatibility would be to just run everything in a lua block and the vim commands would be ran through `vim.cmd` (should probably do that anyway so we can move to init.lua). But at what point would it not be better to just specify that Pre, normal and Post are in different scopes?

PR #53, opened 2022-10-18
New wrapDo helper and wrapping nvim-cmp in do ... end

Alexnortung:
My original goal was to try and fix the cmp issue in #52, but it did not help.

This pull request adds a helper function that will wrap a string (or more preferably lua code) in a `do ... end`.
It also uses this helper in nvim-cmp.


pta2002:
Not sure if it'd help with nvim-cmp, but i'll look into how to fix that! Will merge this in the meantime though

pta2002 (approved):

PR #54, opened 2022-10-24
nvim-tree: added new option

Alexnortung:
Adds new option to nvim-tree.

Also I am not sure what to do with old options. updateCwd should be changed to sync_root_with_cwd, but I don't think we should remove the old option, since it would break some configs.

### DIFF
--- a/plugins/utils/nvim-tree.nix
+++ b/plugins/utils/nvim-tree.nix
@@ -48,7 +48,13 @@ in
       description = "Hijack cursor";
     };
 
+    # TODO: change this to it's new definition sync_root_with_cwd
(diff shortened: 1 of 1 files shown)


pta2002:
Thanks! We could probably just add both and make it an alias, I know there's a function in lib to do that, though home-manager has a way to do deprecation warnings, should probably find out how they do that, could come in handy.

Alexnortung:
I am not sure how to add aliases either. Maybe we could merge this for now, so we have the new option. Then open a new pr once we have figured out how to use aliases. :smile:

pta2002 (approved):

PR #56, opened 2022-10-25
luasnip: init plugin

Alexnortung:
Also added support for paths in `toLuaObject`.

I am not sure how to know what the path of the runtime dir is, but if it is possible I think we should add an option that can include the path of the runtime dir in the paths array.

nonetheless, the plugin module works in my tests.

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -32,6 +32,8 @@
 
     ./pluginmanagers/packer.nix
 
+    ./snippets/luasnip
+
     ./statuslines/airline.nix
     ./statuslines/lightline.nix
     ./statuslines/lualine.nix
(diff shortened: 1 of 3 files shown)


Alexnortung:
I guess the solution to also load the default snippets is to just load multiple times e.g. 

```
require("luasnip.loaders.from_vscode").lazy_load({['paths'] = {'/nix/store/6d2gimfl4rd3j0jw5h56sqw8rgzli1w5-source/profiles/nvim/snippets'}})
require("luasnip.loaders.from_vscode").lazy_load()
```

So it should probably be more flexible and let the user define an array of loads.

I am a bit too tired now, so I might implement this another day.

Alexnortung:
I think this is a good starting point for the plugin and is ready to be merged :)

pta2002:
Yeah, I guess that's a decent compromise for now... Not really sure how to access runtime dir either, probably would have to do it dynamically at runtime.

Alexnortung:
I dont think that it is really needed and I think the module works good enough for a start :)

pta2002 (approved):

PR #60, opened 2022-10-29
treesitter: allow specifying custom parser install dir  @PhotonQuantum

PhotonQuantum:
This option is required if a user opts out nixGrammars because the plugin directory is located in the nix store and is read-only.

However, with this patch, not everything is working due to https://github.com/nvim-treesitter/nvim-treesitter/issues/3605. All tree-sitter plugins will be installed repeatedly.

### DIFF
--- a/plugins/languages/treesitter.nix
+++ b/plugins/languages/treesitter.nix
@@ -21,6 +21,12 @@ in
         description = "Either \"all\" or a list of languages";
       };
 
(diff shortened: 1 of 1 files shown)


pta2002 (approved):

pta2002:
Well this looks like it's an upstream issue, so I see no issue merging this now! Thanks :)

Issue #58, opened 2022-10-26
Add `package` options for plugins


Alexnortung:
It would be nice if I could change which version of a plugin to use (pinning or using unstable/stable) while using nixvim.

I would like to be able to do this in the same manner as in most NixOS modules in nixpkgs handles this, by adding a `package` option.

pta2002:
Good idea, though it'd definitely be a relatively big undertaking. I can get started on it though! Also ensure that all new plugins have that option

Alexnortung:
I can also do some work on this and will open some pull requests ;)

PR #61, opened 2022-10-30
Added `package`option to basic lsp servers

Alexnortung:
This PR adds package options to the basic servers. Related to #58.

You might notice that I also added an extraPackges field in the helper function

```nix
...
    , extraPackages ? { }
...
```

The idea is that we can define which other packages the server should use. So for tsserver, you might define both the main package and the dependency.

```nix
{
  plugins.lsp.servers.tsserver = {
    package = pkgs.nodePackages.typescript-language-server;
    typescriptPackage = pkgs.nodePackages.typescript;
  };
}
```

However, this might not be a good solution, since the typescript package isn't used directly in the output, but instead added to yout global packages.
If another server/plugin also needs the typescript package to be available globally this might cause a conflict and it will be uncertain which package it uses.
I am not sure if we even want to try to solve this problem, since it is already a problem with the following servers if we only used the package option "html", "cssls", "jsonls" and "eslint", since they all depend on `pkgs.nodePackages.vscode-langservers-extracted`.

What do you think @pta2002 ?


pta2002:
I feel like that is a particularly tricky problem to solve (and frankly not even sure how we'd solve it - maybe by just using the `cmds` argument like for elixir?)

On the same note about elixir, should probably have some way of making `cmds` depend on what `package` is... Maybe we can turn it into some sort of module?

e.g. have the option of a server being defined like this:

```nix
{ config, ... }: {
  name = "elixir-ls";
  package = pkgs.elixir_ls;
  cmd = "${config.package}/bin/elixir-ls";
}
```

Alexnortung:
I think using `cmd`s is a very good idea!

I will try to come up with a solution ;)

Although I am still not sure how to solve the problem with extra dependencies. I guess we could make some sort of a derivation which could wrap the specific dependencies. I will probably open a separate PR for this if I figure something out

Alexnortung:
Finally got around to work on this again and thought about some of the things.

I am pretty sure we should use `cmd` as you suggested.
For some language server the user might want to use an older dependency and I think it should be the user's responsibility of wrapping the server binary with that older version.

For other servers that can have direct dependencies in the `cmd` I think we should add an option to change that. For example with `tsserver` I think that it should be possible to define both the typescript version and the language server version and then include the typescript package in the `cmd` with the language server.

pta2002 (approved):
Why do you say something about it feels wrong? Looks fine to me

Alexnortung:
I have thought a bit about it and I think my problem is that if I want to use a nix she'll with a specific version of a language server in it, it would still use the one defined by the cmd. But I don't know how we would deal with that, while also dealing with the other problem 😅

PR #62, opened 2022-10-30
Added extra options

### LINKED ISSUE #30: Ability for users to hook into the setup function's table
Some modules does not provide all options or when they are updated they need to be maintained to provide those options. This can be quite annoying in some instances and can maybe even force the user to rewrite their config in `extraLuaConfig`.
This is because there is no way the user can append anything to the setup object.

To deal with this problem we could make a new option for each module (e.g. `extraSetupAttrs`) which would be an attribute set that could easily be merged with other attribute sets before calling `toLuaObject`. However this could be a quite large task, so if you have any other suggestions, please comment them.

Alexnortung:
This PR aims to resolve part of #30.

Initially I have created the `extraOptionsOptions` helper and tested it with project-nvim and seen that it works.

I will work on adding this option to most other plugins if you think this looks fine @pta2002

### DIFF
--- a/plugins/helpers.nix
+++ b/plugins/helpers.nix
@@ -104,6 +104,17 @@ rec {
     inherit value global;
   };
 
+  extraOptionsOptions = {
+    extraOptions = mkOption {
+      default = { };
+      type = types.attrs;
+      description = ''
+        These attributes will be added to the table parameter for the setup function.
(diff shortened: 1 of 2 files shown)


pta2002:
Looks good to me!

pta2002 (approved):
sorry, completely forgot about this PR!

PR #64, opened 2022-11-05
Use init.lua instead of init.vim

dfangx:
PR to use init.lua instead of init.vim. Resolves https://github.com/pta2002/nixvim/issues/63. Vim setup commands are wrapped inside `vim.cmd()` - this results in the following block (example from my own config):
```
vim.cmd([[
  colorscheme nord

command W w
command Q q

augroup lsp
  autocmd!
  autocmd BufWrite,BufEnter,InsertLeave * :lua vim.diagnostic.setloclist({ open = false })
augroup END

augroup vimStart
  autocmd!
  autocmd VimResized * wincmd =
augroup END                                                                                                                                                                               
]])
```
I've tested this with the home-manager install.

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -89,15 +89,16 @@ in
     let
       customRC =
         (optionalString (config.extraConfigLuaPre != "") ''
-          lua <<EOF
           ${config.extraConfigLuaPre}
-          EOF
         '') +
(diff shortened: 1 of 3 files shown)


pta2002:
This has some issues when being run as standalone, as the generated file is still called init.vim. The extension needs to be changed to .lua for it to work properly

pta2002 (approved):

PR #67, opened 2022-11-14
telescope: added extraOptions and defualtConfig

Alexnortung:
This PR adds extra options to the telescope module.
It also adds defaultConfig which is used to configure telescope.

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -36,7 +36,7 @@ in
       default = { };
     };
 
-    defaultsConfig = mkOption {
+    defaults = mkOption {
       type = types.nullOr types.attrs;
       default = null;
       description = "Telescope default configuration";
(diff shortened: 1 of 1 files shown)


plugins/telescope/default.nix
```diff
@@ -35,6 +35,18 @@ in
       description = "Configuration for the extensions. Don't use this directly";
       default = { };
     };
+
+    defaultsConfig = mkOption {
```
pta2002, line 39: Wouldn't `defaultConfig` be better than `defaultsConfig`?

Alexnortung: What about just `defaults`? It is closer to the native config.

pta2002: sure!

pta2002 (approved):

PR #72, opened 2022-11-23
helpers: fix toLuaObject string escapes

### LINKED ISSUE #71: helpers: `toLuaObject` fails to escape strings containing `'`
**Plugin affected**: many
**Nixpkgs channel**: 22.05
**Home-manager version**: 22.05

## Description
It looks like the issue is `toLuaObject` is using `escapeShellArg`, which doesn't properly handle escaping strings in this situation.

https://github.com/pta2002/nixvim/blob/c320be04c82914933fe94f64282e646d97d342d4/plugins/helpers.nix#L23-L25

When given the string `"<cmd>lua require('gitsigns').blame_line{full=true}<cr>"`
it outputs `'<cmd>lua require('\''gitsigns'\'').blame_line{full=true}<cr>'`

This results in the following error when starting Neovim:

```error
Error detected while processing /nix/store/77xmlnk9bfjw7xd811rf8yipw0jkc2kv-home-manager-files/.config/nvim/init.lua:
E5112: Error while creating lua chunk: ...1rf8yipw0jkc2kv-home-manager-files/.config/nvim/init.lua:92: '}' expected near '\'
```

## Config

```nix
{
  programs.nixvim.maps.normal."<leader>hb" = "<cmd>lua require('gitsigns').blame_line{full=true}<cr>";
}
```

rummik:
This should resolve #71

I was debating rewriting the function, but I'm going to hold off due to work needing to be done on tests first

### DIFF
--- a/plugins/helpers.nix
+++ b/plugins/helpers.nix
@@ -22,9 +22,9 @@ rec {
       "{" + concatMapStringsSep "," toLuaObject args + "}"
     else if builtins.isString args then
       # This should be enough!
-      escapeShellArg args
+      builtins.toJSON args
     else if builtins.isPath args then
-      escapeShellArg args
+      builtins.toJSON (toString args)
     else if builtins.isBool args then
       "${ boolToString args }"
     else if builtins.isFloat args then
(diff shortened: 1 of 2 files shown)


pta2002 (approved):
It seems like this'll work well enough :)

PR #74, opened 2022-11-23
fix: basic flake check should work now

### LINKED ISSUE #73: Fails `nix flake check`
**Nixpkgs channel**: 22.05

## Description

Running `nix flake check --no-write-lock-file` fails with the following error:

```
error: attribute 'helpers' missing

       at /nix/store/46v5hkwaa7aadrdn3gadlcx8ckrv3brj-source/lib/modules.nix:469:28:

          468|         builtins.addErrorContext (context name)
          469|           (args.${name} or config._module.args.${name})
          470|       ) (lib.functionArgs f);
```

The same error is returned when running without `--no-write-lock-file`, however it updates `flake.lock`

rummik:
Should resolve #73

To test, run `nix flake check`

### DIFF
--- a/plugins/completion/lspkind.nix
+++ b/plugins/completion/lspkind.nix
@@ -1,7 +1,8 @@
-{ config, pkgs, lib, helpers, ... }:
+{ config, pkgs, lib, ... }:
 with lib;
 let
   cfg = config.plugins.lspkind;
+  helpers = import ../helpers.nix { inherit lib; };
 in
 {
   options.plugins.lspkind = {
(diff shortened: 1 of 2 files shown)


pta2002:
Ah, I wonder if it's possible to fix it without requiring `helpers` to be manually imported...

rummik:
I imagine there should be a way to do that, but I was more focused on setting up testing 😅

pta2002 (approved):
i'm sure there is a way but honestly haven't found it either :sweat_smile: so i'll merge it for now

PR #75, opened 2022-11-24
options: add viAlias and vimAlias

rummik:
Adds options to alias `vi` and `vim` to `nvim`


pta2002 (approved):
sorry, missed this one when i was clearing out PRs!

PR #82, opened 2022-12-04
lsp: add bufnr to passed argument

LittleQuartZ:
onAttach function have bufnr as its passed argument just like the suggested configurations in [nvim-lspconfig docs](https://github.com/neovim/nvim-lspconfig#suggested-configuration)


pta2002 (approved):

pta2002:
Thanks!

PR #87, opened 2022-12-21
tokyonight: Fix configuration and align more with defaults

traxys:
The global variables don't work for applying configuration, we need to switch to using the 'setup' function. This also aligns the default option values in tokyonight upstream.


pta2002 (approved):
thanks!

Issue #69, opened 2022-11-20
bufferline: `diagnostics_indicator` expects a function


rummik:
**Plugin affected**: bufferline
**Nixpkgs channel**: 22.05 (though this seems to persist into unstable)
**Home-manager version**: 22.05

## Description

> Error detected while processing /nix/store/z19xva0gra1bb071aj4lcbklm5fa8d0k-home-manager-files/.config/nvim/init.lua:
> 
> E5112: Error while creating lua chunk: ...j4lcbklm5fa8d0k-home-manager-files/.config/nvim/init.lua:81: unfinished str
ing near ''function(count, level)'

Relevant lines from `init.lua`:

```lua
require('bufferline').setup{['options'] = {['diagnostics'] = 'nvim_lsp',['diagnostics_indicator'] = 'function(count, level)
  local icon = level:match("error") and " " or ""
  return " " .. icon .. count
end
'}}
```

`plugins.bufferline.diagnosticsIndicator` seems to pass as a string to `diagnostics_indicator`, but the documentation indicates it should receive a function that returns a string

https://github.com/akinsho/bufferline.nvim/blob/d631817de1b52c7b1947e25ec8b7638492914768/doc/bufferline.txt#L259-L262

## Config
```nix
{
  # ...
  plugins.bufferline = {
    enable = true;
    diagnostics = "nvim_lsp";
    diagnosticsIndicator = ''
      function(count, level)
        local icon = level:match("error") and " " or ""
        return " " .. icon .. count
      end
    '';
  };
  # ...
}

**Comment by Alexnortung** [comment]
I think the module is wrong, I can look at it later.
I will probably change the type to use string or rawType.

In your case change 

```
    diagnosticsIndicator = ''
      function(count, level)
        local icon = level:match("error") and " " or ""
        return " " .. icon .. count
      end
    '';
```

tp

```
    diagnosticsIndicator = { __raw = ''
      function(count, level)
        local icon = level:match("error") and " " or ""
        return " " .. icon .. count
      end
    ''; };
```

This doesn't work now, but this will generate the code as raw instead of a string.

From this error it seems that nixvim doesn't support multiline strings @pta2002 do you think this is the case?

Alexnortung:
It should work with the new PR. I should also note that there is a `mkRaw` function that you can use instead of wrapping the string with `__raw`. I am not entirely sure how to use it in your own config, but that should probably be added to the documentation.

Alexnortung:
[...]

PR #89, opened 2022-12-23
add `magma-nvim` to plugins

volkswagenfeature:
[magma-nvim ](https://github.com/dccsillag/magma-nvim) allows you to execute python files as if they were Jupyter notebooks right from inside Neovim. I have added all global options as well, with documentation. I am using the [magama-nvim-goose](https://github.com/WhiteBlackGoose/magma-nvim-goose) fork until their upstream merges their [#69](https://github.com/dccsillag/magma-nvim/pull/69) which allows for external kernels. 

(unrelated, this is my first open source pull request, so am a little excited)

### DIFF
--- a/plugins/plugin-defs.nix
+++ b/plugins/plugin-defs.nix
@@ -84,4 +84,42 @@
       sha256 = "sha256-AtkG2XRVZgvJzH2iLr7UT/U1+LXxenvNckdapnJV+8A=";
     };
   };
+
+  magma-nvim = pkgs.vimUtils.buildVimPlugin rec {
+    pname = "magma-nvim";
+    version = "94370733757d550594fe4a1d65643949d7485989";
+
+    src = pkgs.fetchFromGitHub {
+      # using fork of https://github.com/dccsillag/magma-nvim because there are some unmerged
+      # pull requests that I really want
+      owner = "WhiteBlackGoose";
+      repo = "magma-nvim-goose";
+      rev = version;
+      sha256 = "sha256-IaslJK1F2BxTvZzKGH9OKOl2RICi4d4rSgjliAIAqK4=";
+    };
+
+    passthru.python3Dependencies = ps: with ps;  [
+      pynvim
+      jupyter-client
+      ueberzug
+      pillow
+      cairosvg
+      plotly
+      ipykernel
+      pyperclip
+      (ps.buildPythonPackage rec {
+        pname = "pnglatex";
+        version = "1.1";
+        src = fetchPypi {
+          inherit pname version;
+          hash = "sha256-CZUGDUkmttO0BzFYbGFSNMPkWzFC/BW4NmAeOwz4Y9M=";
+        };
+        doCheck = false;
+        meta = with lib; {
+          homepage = "https://github.com/MaT1g3R/pnglatex";
+          description = "a small program that converts LaTeX snippets to png";
+        };
+      })
+    ];
+  };
 }
--- a/plugins/plugin-defs.nix
+++ b/plugins/plugin-defs.nix
@@ -53,7 +53,7 @@
         pname = "std2";
         version = "48bb39b69ed631ef64eed6123443484133fd20fc";
 
-        doCheck = false;
+        doCheck = true;
 
         src = pkgs.fetchFromGitHub {
           owner = "ms-jpq";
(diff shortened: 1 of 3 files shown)


plugins/utils/magma-nvim.nix
```diff
@@ -0,0 +1,110 @@
...
+    config = mkIf cfg.enable {
+      extraPlugins = [ plugins.magma-nvim ];
+      globals = {
+        magma_image_provider = mkIf (cfg.image_provider != "none") cfg.image_provider;
+        magma_automatically_open_output = mkIf (! cfg.automatically_open_output ) cfg.automatically_open_output;
+        magma_wrap_output = mkIf (! cfg.wrap_output ) cfg.wrap_output;
+        magma_output_window_borders = mkIf (! cfg.output_window_borders ) cfg.output_window_borders;
+        magma_highlight_group = mkIf (cfg.cell_highlight_group != "CursorLine") cfg.cell_highlight_group;
+        magma_show_mimetype_debug = mkIf cfg.show_mimetype_debug cfg.show_mimetype_debug;
+
```
pta2002, line 102: Hey, this looks good! However, I'd just remove this whitespace here, to keep some consistency with the rest of the code.

I'd also remove the whitespsace before the parentheses.

(It's always best to run an autoformatter like `nixfmt` for this kind of thing!)

plugins/plugin-defs.nix
```diff
@@ -84,4 +84,42 @@
...
+    version = "94370733757d550594fe4a1d65643949d7485989";
+
+    src = pkgs.fetchFromGitHub {
+      # using fork of https://github.com/dccsillag/magma-nvim because there are some unmerged
+      # pull requests that I really want
+      owner = "WhiteBlackGoose";
+      repo = "magma-nvim-goose";
+      rev = version;
+      sha256 = "sha256-IaslJK1F2BxTvZzKGH9OKOl2RICi4d4rSgjliAIAqK4=";
+    };
```
matthewmazzanti, line 99: Not entirely certain about using the fork for the core here, might make more sense to use the default and provide a way to override the package.

Typically, for nixpkgs modules, this is done by adding a "pkg" attribute to the options with a default of the package you want. With that, you can change the package used while still providing a "normal" default for most users.

plugins/plugin-defs.nix
```diff
@@ -84,4 +84,42 @@
...
+      ipykernel
+      pyperclip
+      (ps.buildPythonPackage rec {
+        pname = "pnglatex";
+        version = "1.1";
+        src = fetchPypi {
+          inherit pname version;
+          hash = "sha256-CZUGDUkmttO0BzFYbGFSNMPkWzFC/BW4NmAeOwz4Y9M=";
+        };
+        doCheck = false;
```
matthewmazzanti, line 117: Any way to get the checks to run?

plugins/utils/magma-nvim.nix
```diff
@@ -0,0 +1,110 @@
+{pkgs,lib,config, ...}:
+with lib;
+let
+  cfg = config.plugins.magma-nvim;
+  helpers = import ../helpers.nix { lib = lib; };
```
matthewmazzanti, line 5: Doesn't look like this is used anywhere, I would remove.

plugins/utils/magma-nvim.nix
```diff
@@ -0,0 +1,110 @@
...
+  plugins = import ../plugin-defs.nix { inherit pkgs; };
+in
+  {
+    options = {
+      plugins.magma-nvim = {
+        enable = mkEnableOption "Enable magma-nvim?";
+        image_provider = mkOption {
+          type = types.str;
```
matthewmazzanti, line 13: I would use a `types.enum ["none", "ueberzug", "kitty"]` here: https://nixos.org/manual/nixos/stable/index.html#sec-option-types-basic

matthewmazzanti (approved):

matthewmazzanti:
Whoops, ignore that approval I don't know how to github

plugins/plugin-defs.nix
pta2002: That's a good point, and also having the `package` option is something that's (slowly) being implemented (see #58)

pta2002 (approved):
looks good now :)

PR #99, opened 2022-12-30
plugins/git: Add gitsigns plugin

traxys:
This adds the [gitsigns](https://github.com/lewis6991/gitsigns.nvim) plugin. It re-implements the default values for all the setup options.


pta2002 (approved):
this looks great! Thanks a lot :)

PR #100, opened 2022-12-30
null-ls: Add gitsigns code action

traxys:
As this is not an external command but a plugin it adds the gitsigns plugin if enabled.


pta2002 (approved):
Looks good to me

PR #101, opened 2023-01-04
treesitter: Add module for treesitter-refactor

traxys:
As treesitter-refactor is a treesitter module we need a way to pass extra configuration options to the treesitter setup. This is done through a `moduleConfig` attrset. This set should not be used outside nixvim.


pta2002 (approved):

pta2002:
Looks good, thanks!

PR #104, opened 2023-01-05
languages: Add plantuml syntax plugin

traxys:


plugins/default.nix
```diff
@@ -23,6 +23,7 @@
 
     ./languages/ledger.nix
     ./languages/nix.nix
+    ./languages/plantuml_syntax.nix
```
pta2002, line 26: Can you rename the file to `plantuml-syntax.nix` instead? That more closely matches the rest of the project

pta2002 (changes requested):

plugins/default.nix
traxys: Renamed!

pta2002 (approved):

pta2002:
Thanks!

PR #109, opened 2023-01-10
nvim-lsp: add sumneko_lua language server

GaetanLepage:


pta2002:
Doesn't this server require some special configuration to work properly? At least that's what I encountered the last time I looked into it. Will check it out later today.

GaetanLepage:
Indeed...
Here is how I currently have it configured:

```
require 'lspconfig'.sumneko_lua.setup {
    capabilities = capabilities,
    settings = {
        Lua = {
            runtime = {
                -- Tell the language server which version of Lua you're using
                -- (most likely LuaJIT in the case of Neovim)
                version = 'LuaJIT',
            },
            diagnostics = {
                -- Get the language server to recognize the `vim` global
                globals = {'vim'}
            },
            workspace = {
                -- Make the server aware of Neovim runtime files
                library = vim.api.nvim_get_runtime_file("", true),

                -- https://github.com/sumneko/lua-language-server/discussions/1688#discussioncomment-4185003
                checkThirdParty = false,
            },
            -- Do not send telemetry data containing a randomized but unique identifier
            telemetry = {
                enable = false,
            }
        }
    }
}
```

Here is the documentation from the nvim-lspconfig plugin:
https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.txt#sumneko_lua

pta2002 (approved):

PR #111, opened 2023-01-10
Add support for several LSP clients

traxys:
Some of those LSP clients support configuration options.

This commit adds a way to pass configuration options to lsp, using two fields in the mkLsp function:

- extraOptions: allows to define Nix module options
- settings: Translate the config of those options to the settings structure expected by the LSP


pta2002 (approved):
Looks great!

pta2002:
There's some conflicts due to `bashls` being there twice, i'll take care of that tomorrow though :)

PR #113, opened 2023-01-11
nvim-lsp: Add settings for rust-analyzer

traxys:
Because the settings are quite complex they are auto-generated. This commit adds the script creating it & the generated options.


traxys:
By manually what do you mean ? All the options have défaut values that are filled in from the package.JSON

If ra chnages an option that would ne problematic, si I could wrap evrything in nullOr whit null default

pta2002:
Yeah I mean wrapping it in `nullOr` so that we don't have a massive init file generated for options that aren't set.

Just tried it out and just enabling rust-analyzer leads to an almost 6KB init file so definitely something to fix

traxys:
I made the change to add 'nullOr' everywhere. Sometimes there may be some non-optimal behiviour with nullOr (nullOr ...). I can change that, but it would make the autogenerating script more complex

traxys:
@pta2002 I made a significant change, I added a package in the top level flake in order to build the options with the package.json version that corresponds to the rust-analyzer in nixpkgs. I saw that my approach was not very resilient to settings rename

flake.nix
```diff
@@ -35,10 +35,49 @@
...
+                  mkdir -p $out/bin
+                  cp ${./helpers/extract_rust_analyzer.py} $out/bin/extract_rust_analyzer.py
+                '';
+              };
+              extractRustAnalyzerPkg = pkgs.callPackage extractRustAnalyzer { };
             in
             {
-              packages.docs = pkgs.callPackage (import ./docs.nix) {
-                modules = nixvimModules;
+              packages = {
```
pta2002, line 56: not sure if flake.nix is the best place for this - is there any reason this can't just be in the rust-analyzer file? Only real reason I see to update flake.nix is to add inputs, I don't see how you'd need to use `extractRustAnalyzer` externally

traxys: It's just quite easy to do right now, you run `nix run .#` and it generates the file, and it has access to the `pkgs` of the nixvim flake. I don't really see how to generate the .nix file else.

traxys: Alternatively we could only expose the generating script in a nix devolp shell, but I feel it's more complex for not much.

pta2002: Ah I see what you mean, in that case I think this is fine

pta2002: Still, I'll rename the package to runUpdates so that we can extend this to update other stuff in the future

pta2002 (approved):

pta2002 (approved):

PR #117, opened 2023-01-12
plugins/languages/treesitter: change default value for parserInstallDir

GaetanLepage:
Currently, the `nixGrammars = true` method for installing parsers is [broken](https://github.com/pta2002/nixvim/issues/102).

Fortunately, nixvim also supports the _legacy_ way of installing parsers.
The problem is that treesitter tries to put the parsers in its package directory which is not writable when using NixOS.
This can be overcome by setting a custom directory using the `parserInstallDir` option.

This PR proposes to change the default value for this option **when `nixGrammar` is set to `false`** to `$XDG_DATA_HOME/nvim/treesitter`.
Hence, when a NixOS user wants to use the legacy way of installing parsers, he only has to set `nixGrammar` to `false`.

### DIFF
--- a/plugins/languages/treesitter.nix
+++ b/plugins/languages/treesitter.nix
@@ -23,13 +23,15 @@ in
 
       parserInstallDir = mkOption {
         type = types.nullOr types.str;
-        default = if cfg.nixGrammars
-          then
-            null
-          else
-            "$XDG_DATA_HOME/nvim/treesitter"
+        default =
+          if cfg.nixGrammars
+          then null
+          else "$XDG_DATA_HOME/nvim/treesitter"
         ;
-        description = "Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled)";
+        description = ''
+          Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Maybe this is not a good solution.
Here are the other possibilities:

- We leave `null` as the default value, but we add a warning/entry in the documentation: _"If you are using NixOS and if you set `nixGrammars` to `false`, then, you have to set the option `parserInstallDir` to some custom directory such as `$XDG_DATA_HOME/nvim/treesitter`."_
- We set the default of the `parserInstallDir` option to `$XDG_DATA_HOME/nvim/treesitter` independantly of whether or not `nixGrammars` is `true` or `false`.

GaetanLepage:
> Only concern I have is that XDG_DATA_HOME is not defined for macOS (and in some cases might not even be defined in Linux?) so some fallback might be nice

Yes you are right. I thought about this, but I do not think that there is an easy way of implementing a fallback :/

pta2002:
I'll say we leave it like this at the moment then, the option is there to change it anyway and it's just a best-effort default. Will just change the documentation to have that in mind

pta2002 (approved):

PR #120, opened 2023-01-13
modules: add autocmd

GaetanLepage:
This is a basic implementation of an `autocmd` module.

I would be glad to get some feedback on this :)

### DIFF
--- a/modules/autocmd.nix
+++ b/modules/autocmd.nix
@@ -1,0 +1,113 @@
+{ config, lib, ... }:
+with lib;
+let
+  helpers = import ../plugins/helpers.nix { inherit lib; };
+
+  autoCmdOption = types.submodule {
+    options = {
+      event = mkOption {
+        type = types.oneOf [
+          types.str
+          (types.listOf types.str)
+        ];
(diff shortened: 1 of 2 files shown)


modules/autocmd.nix
```diff
@@ -0,0 +1,113 @@
...
+
+      description = mkOption {
+        type = types.nullOr types.str;
+        description = "A textual description of this autocommand.";
+        default = null;
+      };
+
+      callback = mkOption {
+        type = types.nullOr types.str;
+        description = "The name of a Vimscript function to call when this autocommand is triggered. Cannot be used with <command>}.";
```
pta2002, line 48: is this closing bracket meant to be here? 😅

pta2002 (approved):
This looks fine for now. However, later it would be really nice to extend this with features like custom lua code :) Still, thanks for kicking it off! Don't see any reason not to merge this, the rest can be added later

modules/autocmd.nix
GaetanLepage: Probably not haha

PR #123, opened 2023-01-13
modules/highlights: added a wrapper around vim.fn.matchadd

GaetanLepage:
This PR implements a basic `programs.nixvim.match` module.

### DIFF
--- a/modules/highlights.nix
+++ b/modules/highlights.nix
@@ -15,6 +15,17 @@ with lib;
         };
       '';
     };
+
+    match = mkOption {
+      type = types.attrsOf types.str;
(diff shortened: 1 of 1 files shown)


modules/highlights.nix
```diff
@@ -28,6 +39,16 @@ with lib;
         end
       end
       -- }}
+
+      -- Match groups {{
```
pta2002, line 43: this should be on a different mkIf, will fix it myself

pta2002 (changes requested):

pta2002 (approved):

PR #124, opened 2023-01-13
modules/keymaps: add option silentDefault

GaetanLepage:
Ok, this one might be rejected, but I wanted to give it a try.
Personally, I want all my mapping being silent.
In the current state, it would force me to explicitly set the `silent` attribute to `true` for each keymap.

With this PR, I propose to add an extra option: `maps.silentDefault` (`bool`) that sets the default value of the `maps.*.*.silent` attribute.

Unfortunately, it forced me to add `config` to the `helpers.nix` arguments which lead to several file changes.

If you are fine with this approach, we could add the same for the other options: `nowait`, `script`, `expr`, `unique` and `noremap`.

### DIFF
--- a/modules/highlights.nix
+++ b/modules/highlights.nix
@@ -1,6 +1,6 @@
 { config, lib, ... }:
 let
-  helpers = import ../plugins/helpers.nix { inherit lib; };
+  helpers = import ../plugins/helpers.nix { inherit config lib; };
 in
 with lib;
 {
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -1,7 +1,7 @@
 { config, lib, ... }:
 with lib;
 let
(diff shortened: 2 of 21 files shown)


pta2002:
I honestly feel like there might be a nicer way to do this, like providing a function in helpers.nix that would let you do something like this:

```nix
maps = helpers.default { silent = true; } {
  ...
}
```

Should be easier too

GaetanLepage:
> I honestly feel like there might be a nicer way to do this, like providing a function in helpers.nix that would let you do something like this:
> 
> ```nix
> maps = helpers.default { silent = true; } {
>   ...
> }
> ```
> 
> Should be easier too (you can access the `helpers` by using `config.nixvim.helpers`, iirc)

See https://github.com/pta2002/nixvim/pull/236

PR #127, opened 2023-01-16
remaining package options

### LINKED ISSUE #58: Add `package` options for plugins

Alexnortung:
This PR aims to add package option to the remaining plugins that does not have a package option yet.

closes #58

### DIFF
--- a/plugins/bufferlines/barbar.nix
+++ b/plugins/bufferlines/barbar.nix
@@ -7,6 +7,12 @@ in
   options.plugins.barbar = {
     enable = mkEnableOption "Enable barbar.nvim";
 
+    package = mkOption {
+      type = types.package;
+      default = pkgs.vimPlugins.barbar-nvim;
+      description = "Plugin to use for barbar";
+    };
+
     animations = mkOption {
       type = types.nullOr types.bool;
       default = null;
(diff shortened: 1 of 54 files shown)


Alexnortung:
Notable changes:

* the mkPlugin helper now has a `package` parameter.
* null-ls has gotten a `package` and `extraPackages` instead of the old `packages`.

There is a pattern that I am a bit unsure about. In this [code](https://github.com/pta2002/nixvim/pull/127/files#diff-d5e6fe2edcb512c8c2d576adefa425acdeeed2ab50be44ff13c113c1818a94c8R101) I think that the original package value is evaluated, could this be a problem? Does evaluation build/fetch a package and thus taking up more space on the user's disk or more cpu time for building?

pta2002:
> Does evaluation build/fetch a package and thus taking up more space on the user's disk or more cpu time for building?

I don't think it does, though don't quote me on that. AFAIK it only builds the package if you try to access the things on it somehow, not if you just compare it (tested by opening `nix repl` and trying to do `pkgs.google-chrome != null` and noticing it didn't build anything)

pta2002 (approved):

Alexnortung:
Great!

PR #129, opened 2023-01-19
Show warnings for deprecated or changed options

Alexnortung:
This PR is work in progress and needs feedback.

The goal of this PR is to be able to show warnings or throw errors if the user uses options that are deprecated, renamed or removed.
I took inspiration from nixpkgs with the warnings implementation, but it seems like nixpkgs does some extra stuff, thus the need for our own implementation (in `options-warnings.nix`).

I have created the `warnings` module to proxy the warnings from nixvim to the wrapper and then each wrapper can do what they want with the warning.

I have created a small example with `plugins.bufferline.indicatorIcon` which has been deprecated upstream.

Here is how I think we should use it:
For settings in plugins that still work, but are deprecated we should just show a warning (unless there is a certain known alternative).
If an option is deprecated because of a rename, we redirect the value of the option to the new option and show a warning.
For deprecated settings, that no longer work, we throw an error, but still give the user an opportunity to set the option anyway if they use an older version of that plugin.

Other notable changes:
* created a lib folder where helpers have been moved to.


pta2002 (approved):
this is great! Thank you :) When you think it's ready, just mark it as such

Alexnortung:
I am not really sure how to resolve the merge conflict. But I think it works great now!

pta2002:
I'll take care of the conflict, thanks!

PR #131, opened 2023-01-21
tree-sitter: Switch to using nvim-treesitter grammars

traxys:
The grammars in the tree-sitter package and nvim-treesitter package are not the same. The grammars in the nvim-treesitter package are directly taken from the grammars supported by the plugin, whereas tree-sitter grammars are hard-coded in nixpkgs.

This means that there are more tree-sitter grammars available from nvim-treesitter rather than tree-sitter.

### DIFF
--- a/plugins/languages/treesitter.nix
+++ b/plugins/languages/treesitter.nix
@@ -81,7 +81,7 @@ in
 
       grammarPackages = mkOption {
         type = with types; listOf package;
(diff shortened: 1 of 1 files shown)


pta2002 (approved):

pta2002:
Seems like this will fix a lot of the issues with treesitter too, thank you!

PR #132, opened 2023-01-21
plugins/rust-tools: Add a plugins for LSP extensions for rust-analyzer

traxys:
rust-analyzer implements a number of extensions. The rust-tools plugins allows to use them in neovim.

This re-uses the configuration options of the rust-analyzer server.


pta2002 (approved):
looks good, thank you!

PR #134, opened 2023-01-21
plugins: Add nvim-lightbulb

traxys:
nvim-lightbulb is a plugin showing if code actions are available under the cursor.

### DIFF
--- a/plugins/nvim-lsp/nvim-lightbulb.nix
+++ b/plugins/nvim-lsp/nvim-lightbulb.nix
@@ -7,7 +7,7 @@
 }:
 with lib; {
   options.plugins.nvim-lightbulb = {
-    enable = mkEnableOption "Enable nvim-lightbulb, showing available code actions";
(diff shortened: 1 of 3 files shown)


plugins/nvim-lsp/nvim-lightbulb.nix
```diff
@@ -0,0 +1,93 @@
+{
+  pkgs,
+  lib,
+  config,
+  helpers,
+  ...
+}:
+with lib; {
+  options.plugins.nvim-lightbulb = {
+    enable = mkEnableOption "Enable nvim-lightbulb, showing available code actions";
```
pta2002, line 10: just went through a massive change to fix these, mkEnableOption automatically adds "Whether to enable..." to the description so "Enable" shouldn't be part of it. I'll fix it, just leaving it here for future notes

pta2002 (approved):

PR #137, opened 2023-01-22
nvim-lsp: Put LSP settings options in their own namespaces

### LINKED ISSUE #135: error: stack overflow (possible infinite recursion)
After upgrading the nixvim flake I get the following error during `nixos-rebuild build` 
```
"locked": {
        "lastModified": 1674327176,
        "narHash": "sha256-q+HpspHrO5XZnMShBJdLHjHVsCoaWN/Tis7OzRwaF3w=",
        "lastModified": 1674160777,
        "narHash": "sha256-ozE4FVz+UB+5pTiNjhYanS/y52lwXhv8mWzAugbIHak=",
        "owner": "pta2002",
        "repo": "nixvim",
        "rev": "23327700e2a16a3434920d26cb61d90b48478fe7",
        "rev": "d2f714c40ec9bcb20f91a8124f33c81f0576bb68",
        "type": "github"
      },
```
The second one works, the first gives the error.

`error: stack overflow (possible infinite recursion)`

traxys:
Those two options don't control the server settings, and trying to add them in a settings object can result in strange (and broken) behaviour.

Fixes #135

### DIFF
--- a/plugins/nvim-lsp/basic-servers.nix
+++ b/plugins/nvim-lsp/basic-servers.nix
@@ -23,7 +23,7 @@ let
       name = "dartls";
       description = "Enable dart language-server, for dart";
       package = pkgs.dart;
-      extraOptions = {
+      settingsOptions = {
         analysisExcludedFolders = mkOption {
           type = types.nullOr (types.listOf types.str);
           default = null;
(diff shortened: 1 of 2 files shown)


Alexnortung:
Honestly, I do not think this is the correct way to resolve the problem, since we limit the options of each module. I know that `enable` and `package` are uncommon options, but they might still be used. The problem seems to arise from how the settings are used in the `basic-servers.nix` file.

I think a better way would be to make what you did a helper function and that could be used for each server that needs it.

traxys:
`package` and `enable` would never really be possible because the options are flat, you would need to nest them some way (as the enable and package options are already defined).

We could nest everything in a set named `settings` and that would avoid any possible conflict.

pta2002 (approved):

PR #140, opened 2023-01-23
plugins: Add nvim-colorizer

GaetanLepage:
Add the support for plugin [nvim-colorizer](https://github.com/NvChad/nvim-colorizer.lua).

### DIFF
--- a/plugins/utils/nvim-colorizer.nix
+++ b/plugins/utils/nvim-colorizer.nix
@@ -1,0 +1,161 @@
+{ pkgs, config, lib, ... }:
+with lib;
+let
+
+  cfg = config.plugins.nvim-colorizer;
+  helpers = import ../helpers.nix { inherit lib; };
+
+  colorizer-options = {
+    RGB = mkOption {
+      description = "#RGB hex codes";
+      type = types.nullOr types.bool;
+      default = null;
+    };
+    RRGGBB = mkOption {
+      description = "#RRGGBB hex codes";
+      type = types.nullOr types.bool;
+      default = null;
+    };
+    names = mkOption {
+      description = "\"Name\" codes like Blue or blue";
+      type = types.nullOr types.bool;
+      default = null;
+    };
+    RRGGBBAA = mkOption {
+      description = "#RRGGBBAA hex codes";
+      type = types.nullOr types.bool;
+      default = null;
+    };
+    AARRGGBB = mkOption {
+      description = "0xAARRGGBB hex codes";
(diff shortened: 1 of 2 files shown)


plugins/utils/nvim-colorizer.nix
```diff
@@ -0,0 +1,166 @@
...
+            parsers = mkOption {
+              type = types.nullOr types.attrs;
+              default = null;
+            };
+          };
+        }
+      );
+      default = null;
+    };
+    virtualtext = mkOption {
```
pta2002, line 87: mind adding a description here (and for the other ones in the sass colors)? Now with the work on the new doc site I wanted to make sure all of the options are properly documented

plugins/utils/nvim-colorizer.nix
```diff
@@ -0,0 +1,166 @@
...
+        ]
+      );
+      default = null;
+    };
+    sass = mkOption {
+      description = "Enable sass colors";
+      type = types.nullOr (
+        types.submodule {
+          options = {
+            enabled = mkOption {
```
pta2002, line 74: This does not really need to be a submodule, can just do `sass.enabled` and `sass.parsers`, etc. (also, any reason it's named `enabled` instead of `enable` like the rest of the nix options?

plugins/utils/nvim-colorizer.nix
```diff
@@ -0,0 +1,166 @@
...
+                };
+              } // colorizer-options;
+            })
+          ])
+        );
+        default = null;
+      };
+
+      userDefaultOptions = mkOption {
+        description = "Defaults options";
```
pta2002, line 124: Think there's a typo here, shouldn't it be "Default options"?

pta2002 (changes requested):
Looks fine in general, just some minor things to fix up

plugins/utils/nvim-colorizer.nix
GaetanLepage: Done :)

plugins/utils/nvim-colorizer.nix
GaetanLepage: Indeed, fixed.

plugins/utils/nvim-colorizer.nix
GaetanLepage: I changed it. Tell me if it is what you were thinking about :)

pta2002: yep!

pta2002 (approved):

PR #141, opened 2023-01-23
languages: Add clangd_extensions, a plugin to handle clangd LSP extensions

traxys:


plugins/languages/clangd_extensions.nix
```diff
@@ -0,0 +1,189 @@
...
+in
+  with lib; let
+    borderOpt = let
+      bordersTy =
+        types.enum ["double" "rounded" "single" "shadow" "none"];
+    in
+      helpers.defaultNullOpts.mkNullable (types.either bordersTy (types.listOf bordersTy))
+      ''"none"'' "";
+  in {
+    options.plugins.clangd_extensions = {
```
pta2002, line 17: Can you rename this to `clangd-extensions`? Keeps consistency with the rest

plugins/languages/clangd_extensions.nix
```diff
@@ -0,0 +1,189 @@
...
+      enable = mkEnableOption "clangd_extensions, plugin implementing clangd LSP extensions";
+
+      package = mkOption {
```
pta2002, line 20: There's also `mkPackageOption` in the helpers now

pta2002 (changes requested):

plugins/languages/clangd_extensions.nix
traxys: Yeah I was not really sure to respect the case of the plugin or the other plugins, will do!

plugins/languages/clangd_extensions.nix
traxys: Ah right, will edit

traxys:
Should be done, sorry for the force push needed it to rebase on master

pta2002 (approved):

PR #144, opened 2023-01-24
helpers: add mkPackageOption + refactoring

GaetanLepage:
This PR adds a new helper function: `mkPackageOption`.

It replaces:
```nix
package = mkOption {
  type = types.package;
  default = pkgs.vimPlugins.gruvbox-nvim;
  description = "Plugin to use for gruvbox";
};
```

with:
```nix
package = helpers.mkPackageOption "gruvbox" "gruvbox-nvim";
```

It required me to add  `pkgs` as an input of `helpers.nix`... For me it is worth it. What do you think ?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -58,7 +58,7 @@
                   modules = nixvimModules;
                 };
                 runUpdates = pkgs.callPackage
-                  ({ pkgs, stdenv }: stdenv.mkDerivation { 
+                  ({ pkgs, stdenv }: stdenv.mkDerivation {
                     pname = "run-updates";
                     version = pkgs.rust-analyzer.version;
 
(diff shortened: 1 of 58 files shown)


pta2002:
Hey, I'd say it's better to just pass in the package directly to the second option instead of a string. That way we can package stuff that isn't on nixpkgs, or is nested inside some module (like `nodePackages` for example), also makes it so that `helpers` does not need to be added

GaetanLepage:
> Hey, I'd say it's better to just pass in the package directly to the second option instead of a string. That way we can package stuff that isn't on nixpkgs, or is nested inside some module (like `nodePackages` for example), also makes it so that `helpers` does not need to be added

Ok I understand. Now it should be alright :)

pta2002 (approved):

pta2002:
Hey @GaetanLepage, please ensure that `helpers` is declared in all the files that you've changed to use this, there's tons of errors right now because of it. Building the docs evaluates all modules, so just try building them to see if you got everything (`nix build .#docs`). When you're done open a new pr, sorry for the mess, my bad for not checking thoroughly

PR #148, opened 2023-01-24
Refactor gitsigns to use nullable options

traxys:
This adds a few helpers for enums to refactor gitsigns.

It changes all options to be nullable, significantly (by 80 lines) reducing the length of an init.lua that only enables gitsigns.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -88,6 +88,8 @@ rec {
     mkInt = default: mkNullable lib.types.int (toString default);
     mkBool = default: mkNullable lib.types.bool (if default then "true" else "false");
     mkStr = default: mkNullable lib.types.str ''"${default}"'';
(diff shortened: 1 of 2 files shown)


lib/helpers.nix
```diff
@@ -86,8 +86,10 @@ rec {
     );
 
     mkInt = default: mkNullable lib.types.int (toString default);
-    mkBool = default: mkNullable lib.types.bool (toString default);
+    mkBool = default: mkNullable lib.types.bool (if default then "true" else "false");
```
pta2002, line 89: why this change?

pta2002: nvm, got it, `toString true/false` returns `1/0`

pta2002 (approved):

pta2002:
thanks!

lib/helpers.nix
traxys: Worse than that, it returns "1"/""

PR #150, opened 2023-01-25
helpers: add mkPackageOption + refactoring

GaetanLepage:
Indeed, https://github.com/pta2002/nixvim/pull/144 was full of errors.
This is my second attempt.
The docs build successfully.


pta2002 (approved):
alright, seems like this fixes the issues with the previous version :) let's try again

PR #154, opened 2023-01-25
plugins/git: Add gitmessenger plugin

traxys:


plugins/git/gitmessenger.nix
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  config =
+    let
+      cfg = config.plugins.gitmessenger;
+    in
+    mkIf cfg.enable {
+      extraPlugins = [ cfg.package ];
+      globals = {
+        git_messenger_close_on_cursor_moved = helpers.mkIfNonNull cfg.closeOnCursorMoved;
```
pta2002, line 101: There's no real need to use the `mkIfNonNull` here. When creating a global, it uses `toLuaObject` under the hood, so if it's null it's just going to not be set.

pta2002 (changes requested):

plugins/git/gitmessenger.nix
traxys: I never really know when to know mkIf vs if (as is the case here, because toLuaObject would use an 'if')...

pta2002: AFAIK `mkIf` makes it not set, where as `if` just sets it to null. However, in this case, it's irrelevant, since `toLuaObject` ignores anything that's set to null (since in lua, setting something to nil is the same as not setting it)

traxys:
Should have fixed the gitmessenger options

pta2002 (approved):

PR #158, opened 2023-01-30
nvim-lsp: added nvim-cmp completion capabilities

GaetanLepage:
As the [`nvim-cmp` documentation states](https://github.com/hrsh7th/nvim-cmp#recommended-configuration), the LSP servers need to be made aware of the lsp capabilities:

```lua
-- Set up lspconfig.
local capabilities = require('cmp_nvim_lsp').default_capabilities()
-- Replace <YOUR_LSP_SERVER> with each lsp server you've enabled.
require('lspconfig')['<YOUR_LSP_SERVER>'].setup {
  capabilities = capabilities
}
```

This PR automatically injects this lua code when the `cmp-nvim-lsp` plugin is enabled.

### DIFF
--- a/plugins/nvim-lsp/default.nix
+++ b/plugins/nvim-lsp/default.nix
@@ -44,6 +44,12 @@ in {
         default = "";
       };
 
+      capabilities = mkOption {
+        type = types.lines;
+        description = "Lua code that modifies inplace the `capabilities` table.";
+        default = "";
+      };
+
       setupWrappers = mkOption {
         type = with types; listOf (functionTo str);
         description = "Code to be run to wrap the setup args. Takes in an argument containing the previous results, and returns a new string of code.";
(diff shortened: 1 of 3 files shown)


plugins/nvim-lsp/default.nix
```diff
@@ -69,6 +69,13 @@ in {
       if wrappers == []
       then s
       else (head wrappers) (runWrappers (tail wrappers) s);
+
+    getCompletionCapabilities = optionalString config.plugins.cmp-nvim-lsp.enable ''
```
pta2002, line 73: This should probably not live here, but instead on the nvim-cmp plugin itself

pta2002: More flexible this way

pta2002 (changes requested):

pta2002 (approved):
Looks good now!

PR #160, opened 2023-01-31
plugins/utils/nvim-tree: marked setup options as deprecated

GaetanLepage:
Some nvim-tree options have been deprecated.
See: https://github.com/nvim-tree/nvim-tree.lua/wiki/Open-At-Startup#legacy-migration
This PR marks them as deprecated.

In the wiki, they suggest implementations that a user could put in his config to replicate these behaviors.
Maybe we could implement them and restore the corresponding options later.
This is not the goal of this PR.


pta2002:
Agree with the manual implementation!

pta2002 (approved):

PR #163, opened 2023-02-02
plugins/utils/indent-blankline: fix enable option

GaetanLepage:
The `enable` option of the indent-blankline plugin was wrongly implemented which was causing a bug when building any nixvim configuration:
```
error: ‘mkIf’ called with a non-Boolean condition
```

The bug was introduced in https://github.com/pta2002/nixvim/commit/89bf2d660b6696b7548214e22f74d343397c9d4e.


traxys:
Oops sorry

pta2002 (approved):

PR #167, opened 2023-02-14
modules/options: use vim.opt instead of vim.o

s1341:
By setting the options on `vim.opt` instead of `vim.o` we are able to use table values (dictionaries/lists) when specifying options.


GaetanLepage:
LGTM !

pta2002 (approved):
looks good to me

PR #173, opened 2023-02-19
plugins/utils/nvim-tree: huge option refactoring + implementation of auto-open deprecated functionality

GaetanLepage:
The options of the `nvim-tree` plugin were not up-to-date.
This PR ensures they are 1:1 compatible with the upstream plugin.
The documentation has also been ported.
My only concern is that this module is now huge.
Maybe, we may prefer to do something more [RFCS-0042](https://github.com/Infinisil/rfcs/blob/config-option/rfcs/0042-config-option.md)-style.
I would like to get your opinion on this.

Furthermore, a manual implementation of the [`open_on_setup` deprecated options](https://github.com/nvim-tree/nvim-tree.lua/wiki/Open-At-Startup).

~~NOTE: This PR requires https://github.com/pta2002/nixvim/pull/172 to be merged.~~ now merged.

### DIFF
--- a/plugins/utils/nvim-tree.nix
+++ b/plugins/utils/nvim-tree.nix
@@ -9,6 +9,27 @@ with lib; let
   helpers = import ../helpers.nix {inherit lib;};
   optionWarnings = import ../../lib/option-warnings.nix args;
   basePluginPath = ["plugins" "nvim-tree"];
+
+  openWinConfigOption =
+    helpers.defaultNullOpts.mkNullable
+    # Type
+    types.attrs
+    # Default
+    ''
+      {
+        col = 1;
+        row = 1;
+        relative = "cursor";
+        border = "shadow";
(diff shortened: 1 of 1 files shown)


pta2002 (changes requested):
Yeah future options should definitely be in an `extraOptions` thing to avoid the exponential growth of the number of options. Since you've already gone through the effort to add all this I think it's fine to keep it in, but for everyone's sanity an `extraOptions` setting here would be for the best

GaetanLepage:
Yeah future options should definitely be in an extraOptions thing to avoid the exponential growth of the number of options. Since you've already gone through the effort to add all this I think it's fine to keep it in, but for everyone's sanity an extraOptions setting here would be for the best

@pta2002 done :)

pta2002 (approved):

PR #178, opened 2023-02-21
modules/keymaps: bind lua function calls

GaetanLepage:
Right now, it is not possible to bind a raw lua function to a key.
Well, it is possible by doing this:
```nix
maps.normal.x = ":lua my_lua_func()<cr>";
```
which is not very pleasant.

This PR adds a new field to the action submodule for mappings. The example above would become:
```nix
maps.normal.x.luaFunc = "my_lua_func";
```

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -55,7 +55,13 @@ with lib; rec {
           nowait = false;
           action = action;
         }
-        else action)
+        else {
+          inherit (action) silent expr unique noremap script nowait;
+          action =
+            if action.lua
+            then mkRaw action.action
+            else action;
+        })
       maps;
   in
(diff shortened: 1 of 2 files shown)


GaetanLepage:
> I would me more in favour of having a `maps.normal.x.lua = true` and use `action` for `my_lua_func` I think (though I'm not opposed to this approach)

Yes, I would prefer this as well.
What do you think @pta2002 @Alexnortung ?

pta2002:
Agree with @traxys, I think that way it's cleaner

GaetanLepage:
Here we are !

traxys:
LGTM

traxys (approved):

Alexnortung (approved):
I think it's fine :)

PR #179, opened 2023-02-21
plugins/telescope: add keymaps and keymapsSilent options

GaetanLepage:
Before:
```nix
maps.normal = {
  "<leader>fg" = {
    action = ":lua require('telescope.builtin').live_grep()<cr>";
    silent = true;
  };
  "<C-p>" = {
    action = ":lua require('telescope.builtin').git_files()<cr>";
    silent = true;
  };
};
```

After:
```nix
plugins.telescope  = {
  keymapsSilent = true;
  keymaps = {
    "<leader>fg" = "live_grep";
    "<C-p>" = "git_files";
  };
};
```

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -23,6 +23,22 @@ in {
 
     package = helpers.mkPackageOption "telescope.nvim" pkgs.vimPlugins.telescope-nvim;
 
+    keymaps = mkOption {
+      type = types.attrsOf types.str;
+      description = "Keymaps for telescope.";
+      default = {};
+      example = {
+        "<leader>fg" = "live_grep";
+        "<C-p>" = "git_files";
+      };
+    };
+
+    keymapsSilent = mkOption {
+      type = types.bool;
+      description = "Whether telescope keymaps should be silent";
+      default = false;
+    };
+
     highlightTheme = mkOption {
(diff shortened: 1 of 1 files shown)


plugins/telescope/default.nix
```diff
@@ -74,16 +89,28 @@ in {
...
-        for i, extension in ipairs(__telescopeExtensions) do
-          require('telescope').load_extension(extension)
+          for i, extension in ipairs(__telescopeExtensions) do
+            require('telescope').load_extension(extension)
+          end
         end
-      end
-    '';
+      ''
+      + (optionalString (cfg.keymaps != {}) ''
```
traxys, line 104: Instead of adding keymaps here would you not be able to do something like `config.maps.normal = mapAttrs (<some function to build the correct keymaps>) cfg.keymaps` ?

I think this would avoid repetition with the keymaps, and it will benefit from any change we do to the keymaps.

GaetanLepage: Yes, I thought about it, but I wanted to `require("telescope.builtin")` only once.

GaetanLepage: Yes, I will wait for the merge of https://github.com/pta2002/nixvim/pull/178 and will use the nixvim way then.

traxys (approved):

PR #181, opened 2023-02-21
plugins/nvim-lsp: added options for lua_ls

GaetanLepage:
Added the most useful options for `lua_ls` and initialized the defaults as recommended by the [`nvim-lsp` documentation](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.txt#lua_ls).

### DIFF
--- a/plugins/nvim-lsp/language-servers/default.nix
+++ b/plugins/nvim-lsp/language-servers/default.nix
@@ -5,7 +5,8 @@
   ...
 } @ args:
 with lib; let
-  helpers = import ../helpers.nix args;
+  lspHelpers = import ../helpers.nix args;
+  helpers = import ../../helpers.nix {inherit lib;};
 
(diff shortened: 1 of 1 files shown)


plugins/nvim-lsp/basic-servers.nix
```diff
@@ -225,6 +226,58 @@ with lib; let
...
+      # https://github.com/LuaLS/lua-language-server/wiki/Settings
+      settingsOptions = {
+        runtime = {
+          version = mkOption {
+            type = types.str;
+            description = ''
+              Tell the language server which version of Lua you're using
+              (most likely LuaJIT in the case of Neovim)
+            '';
+            default = "LuaJIT";
```
traxys, line 240: I feel like still having types.nullOr is nice because then a user has a way to set the  value to the 'default' according to the LSP without searching for the default.

traxys (approved):

PR #184, opened 2023-02-23
modules/keymaps: fix bug in keymaps generation

GaetanLepage:
A bug was introduced in https://github.com/pta2002/nixvim/pull/178.
This PR fixes it.


traxys (approved):
Did not see that when reviewing :x

PR #191, opened 2023-02-24
plugins/nvim-lsp/trouble: refactoring + new options

GaetanLepage:
Fixes https://github.com/pta2002/nixvim/issues/190.

### DIFF
--- a/plugins/nvim-lsp/trouble.nix
+++ b/plugins/nvim-lsp/trouble.nix
@@ -7,24 +7,222 @@
   cfg = config.plugins.trouble;
   helpers = import ../helpers.nix {inherit lib;};
 in
-  with lib;
-  # with helpers;
-    {
-      options.plugins.trouble = {
+  with lib; {
+    options.plugins.trouble =
+      helpers.extraOptionsOptions
+      // {
         enable = mkEnableOption "trouble.nvim";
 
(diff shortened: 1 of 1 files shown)


plugins/nvim-lsp/trouble.nix
```diff
@@ -7,24 +7,222 @@
...
         enable = mkEnableOption "trouble.nvim";
 
         package = helpers.mkPackageOption "trouble-nvim" pkgs.vimPlugins.trouble-nvim;
 
-        position = helpers.mkNullOrOption (types.enum ["top" "left" "right" "bottom"]) "Position of the list";
-        height = helpers.mkNullOrOption types.int "Height of the trouble list when position is top or bottom";
-        width = helpers.mkNullOrOption types.int "Width of the trouble list when position is left or right";
-        icons = helpers.mkNullOrOption types.bool "Use devicons for filenames";
-      };
+        position = helpers.defaultNullOpts.mkEnum ["top" "left" "right" "bottom"] "bottom" ''
```
traxys, line 18: If you prefer there is a way to have the first element of the enum as default (it allows to type less characters) :P

traxys (approved):
LGTM

plugins/nvim-lsp/trouble.nix
GaetanLepage: Yes, I know.
I am not really sure about which one is best but I tend to prefer setting the default explicitly.

PR #192, opened 2023-02-24
plugins/nvim-lsp/trouble: typo

GaetanLepage:
Just a minor typo which prevents the `signs` options to work.


traxys (approved):
Yeah, we should really look into testing...

PR #193, opened 2023-02-24
plugins/languages/neorg: add package neorg

GaetanLepage:
Add package [neorg](https://github.com/nvim-neorg/neorg).

### DIFF
--- a/plugins/utils/neorg.nix
+++ b/plugins/utils/neorg.nix
@@ -1,0 +1,149 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}: let
+  cfg = config.plugins.neorg;
+  helpers = import ../helpers.nix {inherit lib;};
+in
+  with lib; {
+    options.plugins.neorg =
+      helpers.extraOptionsOptions
+      // {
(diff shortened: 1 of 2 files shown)


plugins/utils/neorg.nix
```diff
@@ -0,0 +1,149 @@
...
+              level = "warn";
+            };
+            error = {
+              hl = "ErrorMsg";
+              level = "error";
+            };
+            fatal = {
+              hl = "ErrorMsg";
+              level = "5";
+            };
```
traxys, line 47: It  seemed  like level == attrName but the last one says nope....

traxys (approved):
LGTM (if it was tested with at least the mode options as they are complicated)

plugins/utils/neorg.nix
traxys: Maybe you could check if the defaults set has a `level` attr and act differently, but I feel it's too complicated

GaetanLepage: Yes, I didn't want to overdo it. Sometimes, a bit more redundant but simpler code is better !

PR #195, opened 2023-02-24
plugins/nvim-lsp/pylsp: fix bug in settings generation

GaetanLepage:
I forgot to set the settings generation function option which led the pylsp settings declared in `pylsp.nix` ignored.


traxys (approved):
I am going to require a screenshot of a generated module with the correct options :P

PR #197, opened 2023-02-24
plugins/lualine: add nvim-web-devicons to extraPlugins

GaetanLepage:
`nvim-web-devicons` is an optional dependency of `lualine` so I guess we can install it by default.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -86,6 +86,12 @@ in {
 
       package = helpers.mkPackageOption "lualine" pkgs.vimPlugins.lualine-nvim;
 
+      iconsEnabled = mkOption {
+        type = types.bool;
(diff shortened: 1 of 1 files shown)


plugins/statuslines/lualine.nix
```diff
@@ -160,7 +160,10 @@ in {
     };
   in
     mkIf cfg.enable {
-      extraPlugins = [cfg.package];
+      extraPlugins = with pkgs.vimPlugins; [
+        cfg.package
+        nvim-web-devicons
```
traxys, line 165: We should  maybe to this optionally if it's optional in lualine ? For people who want a small config

We could add an option "enableIcons" and conditionally add it.

pta2002: I agree!

GaetanLepage: Done :)

PR #199, opened 2023-02-24
tests: Implement a way to test nixvim

traxys:
The tests can be executed using `nix flake check`, they check that modules can be built, and they execute in neovim without any errors.

This commit only implements tests for tokyonight-nvim upstream defaults

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -62,6 +62,10 @@
             };
           extractRustAnalyzerPkg = pkgs.callPackage extractRustAnalyzer {};
         in {
+          checks = import ./tests/checks.nix {
+            inherit pkgs;
+            makeNixvim = self.legacyPackages."${system}".makeNixvim;
+          };
           packages = {
             docs = pkgs.callPackage (import ./docs.nix) {
               modules = nixvimModules;
--- a/tests/checks.nix
+++ b/tests/checks.nix
@@ -1,0 +1,36 @@
+{
+  makeNixvim,
+  pkgs,
+}: let
+  tests = import ./plugins {inherit (pkgs) lib;};
+in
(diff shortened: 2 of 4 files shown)


pta2002:
I'd like to check on this one myself before merging, so give me a moment before approving @GaetanLepage :)

Seems like a great idea just skimming though

pta2002 (changes requested):
This is great and I like the way you approach it. However, could you move each check to a separate file? And then just load all files automatically, like it's done for the module loading in flake.nix

traxys:
Sure!
By automatically you mean creating a file with lots of `imports = [./<...> ./<....> ./<....>]`

traxys:
@pta2002 is this more what you had in mind  ?

pta2002:
I meant more in the sense of listing the directory automatically, like it's done in flake.nix to load the modules, that way there's no need to list everything manually

traxys:
I feel like this would be enough automation, you just have list a new file in attrset, I fear that too much automation would make things brittle

traxys:
@pta2002 Could you point me to the exact lines you mean ? I don't really understand

pta2002:
Yes, here you go:

```nix
      nixvimModules = map (f: ./modules + "/${f}") (attrNames (builtins.readDir ./modules));
```

traxys:
This should be more automatic!

pta2002 (approved):

pta2002:
Awesome!

pta2002:
@GaetanLepage requesting your review since it's a big change

GaetanLepage (approved):
Seems good now !

PR #200, opened 2023-02-25
plugins/statuslines/lualine: fix type of color option

GaetanLepage:
While refactoring I mis-defined the type of option `color` for the components.
`str or (attrsOf str)` --> `attrsOf str`


traxys (approved):
You can start to look into my test PR to implement tests :P

PR #203, opened 2023-02-25
plugins/neorg: fix logger.modes option

GaetanLepage:


plugins/utils/neorg.nix
```diff
@@ -120,23 +125,26 @@ in
...
+                    level = modeConfig.level;
+                  in
+                    if isNull level
+                    then null
+                    else
+                      (
+                        if (isInt level)
+                        then level
+                        else helpers.mkRaw "vim.log.levels.${strings.toUpper level}"
+                      );
```
traxys, line 145: This should not be needed as the level can't be null

traxys (changes requested):

plugins/utils/neorg.nix
GaetanLepage: Indeed. Fixed !

traxys (approved):

PR #204, opened 2023-02-26
plugins: floaterm: fix options

s1341:
Looks like the refactor broke things.

We should really have CI that at least tries to evaluate all plugins.


traxys:
Yeah work on that started in #199, we'll still need to create the tests for all the plugins

GaetanLepage:
Indeed !!!
Thank you for the fix @s1341 .

LGTM

GaetanLepage (approved):

GaetanLepage:
Well, I already fixed that bug in yet unmerged https://github.com/pta2002/nixvim/pull/189.
Anyway, thank you for reporting it :)

PR #205, opened 2023-02-26
flake.nix: update inputs

GaetanLepage:
Update flake inputs.
plenary-nvim was added as a dependency to harpoon in nixpkgs (https://github.com/NixOS/nixpkgs/pull/218291).


traxys:
Should be mentioned that it is for harpoon

traxys (approved):

PR #206, opened 2023-02-27
plugins/nvim-lsp: Add a user option to set the cmd of each language server

GaetanLepage:
This is a proposal to implement the `plugins.lsp.servers.<name>.cmd` option.
This lets the user provide a custom command to run the language servers.
For some language servers, nixvim sets a default value for it. For others, it is left to `null`.

### DIFF
--- a/plugins/nvim-lsp/helpers.nix
+++ b/plugins/nvim-lsp/helpers.nix
@@ -39,6 +39,10 @@
         plugins.lsp.servers.${name} =
           {
             enable = mkEnableOption description;
+            cmd = mkOption {
+              type = with types; nullOr (listOf str);
+              default = cmd cfg;
+            };
             settings = settingsOptions;
             extraSettings = mkOption {
               default = {};
@@ -62,7 +66,7 @@
             {
               name = serverName;
               extraOptions = {
-                cmd = cmd cfg;
+                cmd = cfg.cmd;
                 settings = settings (cfg.settings // cfg.extraSettings);
               };
             }


plugins/nvim-lsp/helpers.nix
```diff
@@ -62,7 +66,7 @@
             {
               name = serverName;
               extraOptions = {
-                cmd = cmd cfg;
+                cmd = cfg.cmd;
```
pta2002, line 69: Don't you have to pass in the argument here? so `cfg.cmg package`

pta2002: nvm i got it

plugins/nvim-lsp/helpers.nix
```diff
@@ -10,7 +10,7 @@
     serverName ? name,
     package ? pkgs.${name},
     extraPackages ? {},
-    cmd ? (cfg: null),
+    cmd ? (package: null),
```
pta2002, line 13: Is there any reason you're changing it to `package` instead of `config`? The config seems more versatile

GaetanLepage: I did that because I had an infinite recursion. But it might have been caused by something else...
Let me try again

pta2002: Oh yeah i can see why it happens, since `cmd` is itself part of cfg. Maybe you can pass in `cfg` without `cmd`?

GaetanLepage: Ok it seems to work !

pta2002 (approved):
This looks alright!

PR #210, opened 2023-02-28
plugins/nvim-lsp: add a fallback for the lua LS package

GaetanLepage:
The lua language server package has been renamed [in nixpkgs](https://github.com/NixOS/nixpkgs/pull/215057) and thus also [in nixvim](https://github.com/pta2002/nixvim/pull/169).
However, the users following a stable version of nixpkgs should be using the old name.
Now, nixvim should be robust to both stable/unstable state of nixpkgs.


pta2002 (approved):
looks good to me

PR #211, opened 2023-02-28
plugins/nvim-autopairs: refactor + update options

GaetanLepage:
I went through the [nvim-autopairs](https://github.com/windwp/nvim-autopairs) plugin and cleaned the module:
- new options
- increased use of helpers for more concise code
- removed deprecated options

### DIFF
--- a/plugins/utils/nvim-autopairs.nix
+++ b/plugins/utils/nvim-autopairs.nix
@@ -3,58 +3,131 @@
   config,
   lib,
   ...
-}:
+} @ args:
 with lib; let
   cfg = config.plugins.nvim-autopairs;
   helpers = import ../helpers.nix {inherit lib;};
+  optionWarnings = import ../../lib/option-warnings.nix args;
+  basePluginPath = ["plugins" "nvim-autopairs"];
 in {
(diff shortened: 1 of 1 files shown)


plugins/utils/nvim-autopairs.nix
```diff
@@ -3,59 +3,130 @@
...
+      checkTs = helpers.defaultNullOpts.mkBool false "Use treesitter to check for a pair.";
+
+      tsConfig = helpers.mkNullOrOption (types.nullOr types.attrs) ''
+        Configuration for TreeSitter.
+
+        Default:
+        {
+          lua = [ "string" "source" ];
+          javascript = [ "string" "template_string" ];
+        };
```
traxys, line 86: Wouldn't this need to be in \`\`\` to render correctly in the documentation

traxys (changes requested):

plugins/utils/nvim-autopairs.nix
GaetanLepage: Done :)

traxys (approved):

PR #213, opened 2023-02-28
helpers: fix toLuaObject ignoring empty objects

pta2002:
At the moment, toLuaObject ignores empty objects, leading to some settings like `neorg`'s modules not working

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -20,7 +20,7 @@ with lib; rec {
               if head (stringToCharacters n) == "@"
               then toLuaObject v
               else "[${toLuaObject n}] = " + (toLuaObject v))
-            (filterAttrs (n: v: !isNull v && toLuaObject v != "{}") args)))
+            (filterAttrs (n: v: !isNull v) args)))
         + "}"
     else if builtins.isList args
     then "{" + concatMapStringsSep "," toLuaObject args + "}"


GaetanLepage:
Some regressions have been caused by this PR.

Ok I get what you mean: if the user does not provide any config, the setup option is called with a bunch of useless empty lua tables.

We have 2 choices:

**-> SOLUTION 1**
Rewrite the relevant options by substituting

```
containerOption = {
  optionA = ...;
  optionB = ...;
};
```

with

```
containerOption = helpers.mkNullOrOption
  (types.submodule {
    options = {
      optionA = ...;
      optionB = ...;
    };
  })
  "This is a container option"
;
```

without forgetting the proper handling in the implementation part (the option value can now be `null`).

**-> SOLUTION 2**
As @pta2002 suggested, we add an optional `ignoreEmpty? false` option to the `toLuaObject` function.


I started to write PRs implementing solution 1 ([#214](https://github.com/pta2002/nixvim/pull/214) and [#215](https://github.com/pta2002/nixvim/pull/214)), but, while writting the one for nvim-tree, I wondered if it was not too much work.
Thus, _solution 2_ might be a nicer choice.

pta2002:
I don't mind solution 2 at all, but I'm not sure how you'd go about doing it without breaking everything, since toLuaObject isn't designed for taking parameters. If you have any idea do tell

pta2002:
Could just create a separate function obviously

GaetanLepage:
I am not sure which is the cleaner.
In the case of having two functions, what would be the "default" version: the one ignoring or not ignoring empty dictionaries ?

GaetanLepage:
@traxys any thoughts ?

traxys:
Not really

pta2002:
IMO makes more sense to ignore empty, makes it so we need to change less stuff @GaetanLepage

GaetanLepage:
Ok, I am actually still wondering.
What I like about the new behavior (empty attrs translates to empty lua table) is that it makes more sense.
If I (the user) set an option to the value `{}`, I want it to appear as such in my lua.
Furthermore, it would leave us with a single implementation of `toLuaObject`, without the need of `mkEmpty`.
The only problem is on our side (the developers) to make sure that the default values for options that have suboption is `null`.

Hence, we should look for all the options in the codebase that have such composite options and translate them to the clean submodule style.

I am all right with tackling this myself.

PR #219, opened 2023-03-01
Added flake template

Alexnortung:
I have added a flake template, which I think can be beneficial for new Nixvim users and users who do not yet use standalone.

I have also exposed some of the lib in the flake. This allows the template to use the check function defined in lib.
Another benefit of exposing lib is that it might be more intuitive to use the helpers this way, compared to the current one where you need to access `config.nixvim.helpers`, instead you can use `inputs.nixvim.lib.helpers`.

### DIFF
--- a/templates/simple/flake.nix
+++ b/templates/simple/flake.nix
@@ -2,8 +2,7 @@
   description = "A nixvim configuration";
 
   inputs = {
-    # nixvim.url = "github:pta2002/nixvim";
-    nixvim.url = "path:../..";
+    nixvim.url = "github:pta2002/nixvim";
     flake-utils.url = "github:numtide/flake-utils";
   };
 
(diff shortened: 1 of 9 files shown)


GaetanLepage:
Thanks for this contribution.
I have not tested it, but looking at the code, the changes look good to me !

GaetanLepage (approved):

templates/simple/flake.nix
```diff
@@ -0,0 +1,40 @@
...
+    nixpkgs,
+    nixvim,
+    flake-utils,
+    ...
+  } @ inputs: let
+    nixvimLib = nixvim.lib;
+    config = import ./config; # import the module directly
+  in
+    flake-utils.lib.eachDefaultSystem (system: let
+      pkgs = nixpkgs.legacyPackages.${system};
```
pta2002, line 20: Can this be changed to use `import nixpkgs { inherit system; }` instead?

pta2002 (changes requested):

pta2002 (approved):
that's it, great! thanks :)

PR #220, opened 2023-03-02
Added nix check and pre commit hook for formatting

Alexnortung:
After making my first commits to this project in a while I was met with an angry CI that didn't like I did not run `nix fmt` before committing/pushing. So I thought a good idea would be to provide a pre-commit hook, so I and others don't forget to format before committing.

Thanks to [`cachix/pre-commit-hooks.nix`](https://github.com/cachix/pre-commit-hooks.nix) this was quite easy to implement.

To install the pre commit hooks simply open the devShell in the repository (`nix develop`). Or if you have direnv it will load automatically.

This was also referenced a while back in another issue https://github.com/pta2002/nixvim/issues/114#issuecomment-1384582491

PR #225, opened 2023-03-07
gitignore: add .direnv

GaetanLepage:
The direnv integration was introduced in #220.
Using it requires to create a `.direnv` directory which is currently not ignored by git.


Alexnortung (approved):
Thanks, I completely forgot about that

traxys (approved):

PR #234, opened 2023-03-09
bugfix: Set default for `programs.nixvim`.

### LINKED ISSUE #224: Home manager won't build if no config is set in nixvim
**Plugin affected**: What plugin does this bug relate to (if applicable)
**Nixpkgs channel**: 22.11
**Home-manager version**: 22.11

## Description
If I have the nixvim module added in my home-manager config and I do not use any options from nixvim, home-manager will fail to build with an error:

```
error: The option `programs.nixvim' is used but not defined.
```

This happened as I removed my config from my files, but didn't remove the module import yet.

Alexnortung:
fixes #224

setting a default for the nixvim submodule, makes it possible to build with no options set.

### DIFF
--- a/modules/colorscheme.nix
+++ b/modules/colorscheme.nix
@@ -12,8 +12,8 @@ with lib; {
     };
   };
 
-  config = {
-    extraConfigVim = optionalString (config.colorscheme != "" && config.colorscheme != null) ''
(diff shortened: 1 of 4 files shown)


GaetanLepage (approved):
LGTM

pta2002:
damn can't believe it was this simple <.<

PR #236, opened 2023-03-10
helpers: add mkMaps and mkModeMaps

GaetanLepage:
Add a way to declare several keymaps without having to duplicate the defaults for each one of them:
```nix
maps = config.nixvim.helpers.mkMaps { silent = true; } {
  normal = {
    ...
  };
  visual = {
    ...
  };
};
```

Or for a single mode:
```nix
maps.normal = config.nixvim.helpers.mkModeMaps { silent = true; } {
  ...
};
```

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -75,6 +75,54 @@ with lib; rec {
       })
       normalized);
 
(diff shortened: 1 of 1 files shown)


traxys (approved):
Really nice approach!

PR #237, opened 2023-03-10
plugins/nvim-cmp: refactor + tests

GaetanLepage:
Refactoring of the `nvim-cmp` plugin.
It now doesn't suffer from the [new `toLuaObject` behavior](https://github.com/pta2002/nixvim/pull/213).
I also made sure that all of the options were correct regarding upstream.

### DIFF
--- a/plugins/completion/nvim-cmp/default.nix
+++ b/plugins/completion/nvim-cmp/default.nix
@@ -7,89 +7,53 @@
 with lib; let
   cfg = config.plugins.nvim-cmp;
   helpers = import ../../helpers.nix {inherit lib;};
-  mkNullOrOption = helpers.mkNullOrOption;
   cmpLib = import ./cmp-helpers.nix args;
-  # functionName should be a string
-  # parameters should be a list of strings
-  wrapWithFunction = functionName: parameters: let
-    parameterString = strings.concatStringsSep "," parameters;
-  in ''${functionName}(${parameterString})'';
 in {
   options.plugins.nvim-cmp = {
     enable = mkEnableOption "nvim-cmp";
 
     package = helpers.mkPackageOption "nvim-cmp" pkgs.vimPlugins.nvim-cmp;
 
-    performance = mkOption {
-      default = null;
-      type = types.nullOr (types.submodule ({...}: {
-        options = {
-          debounce = mkOption {
-            type = types.nullOr types.int;
-            default = null;
-          };
-          throttle = mkOption {
-            type = types.nullOr types.int;
-            default = null;
-          };
-        };
-      }));
-    };
+    performance = helpers.mkCompositeOption "Performance options" {
+      debounce = helpers.defaultNullOpts.mkInt 60 ''
+        Sets debounce time
+        This is the interval used to group up completions from different sources
+        for filtering and displaying.
+      '';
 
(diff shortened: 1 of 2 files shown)


traxys (commented):
Could we have some tests so that we don't break things going forwards ?

GaetanLepage:
> Could we have some tests so that we don't break things going forwards ?

Yes, I plan to write several tests in the following days.

GaetanLepage:
Wow, I found so many errors when writing the test !

plugins/completion/nvim-cmp/default.nix
```diff
@@ -115,103 +79,158 @@ in {
...
-          keyword_length = mkOption {
-            default = null;
-            type = types.nullOr types.int;
-          };
+    mappingPresets = mkOption {
+      default = [];
+      type = types.listOf (types.enum [
+        "insert"
+        "cmdline"
+        # Not sure if there are more or if this should just be str
```
traxys, line 87: There are no other presets (lua/cmp/config/mapping.lua)

plugins/completion/nvim-cmp/default.nix
```diff
@@ -115,103 +79,158 @@ in {
...
+          ```
+          function(args)
+            vim.fn["vsnip#anonymous"](args.body) -- For `vsnip` users.
+            -- require('luasnip').lsp_expand(args.body) -- For `luasnip` users.
+            -- require('snippy').expand_snippet(args.body) -- For `snippy` users.
+            -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
+          end
+          ```
+        '';
+    };
```
traxys, line 118: I feel the problem with this one is  that we don't really know which engine is selected so we can't install it. Maybe in a future commit we could be a bit more intelligent with an enum for the diffrent possible engines, and still leave an escape hatch to plug your own ?

plugins/completion/nvim-cmp/default.nix
GaetanLepage: Ok thanks ! fixed :)

plugins/completion/nvim-cmp/default.nix
GaetanLepage: Ok, I'll do it in another PR.

plugins/completion/nvim-cmp/default.nix
```diff
@@ -115,103 +79,157 @@ in {
...
-          completeopt = mkOption {
-            default = null;
-            type = types.nullOr types.str;
-          };
-        };
-      }));
+      keywordPattern =
+        helpers.defaultNullOpts.mkStr
+        "[[\%(-\?\d\+\%(\.\d\+\)\?\|\h\w*\%(-\w*\)*\)]]"
+        "The default keyword pattern.";
```
traxys, line 127: Same tretament as todo-comments, mkRaw with [[ ]] appendended, and  a note in the documentation

traxys (changes requested):

plugins/completion/nvim-cmp/default.nix
GaetanLepage: Done :)

traxys (approved):
LGTM

PR #242, opened 2023-03-12
plugins/todo-comments: fix missing mkRaw for pattern options

GaetanLepage:
The `highlight.pattern` and `search.pattern` options were not exported as pure lua code.
Added `mkRaw` to fix this.

### DIFF
--- a/plugins/utils/todo-comments.nix
+++ b/plugins/utils/todo-comments.nix
@@ -114,9 +114,15 @@ in {
             "fg" or "bg" or empty.
           '';
 
-          pattern = helpers.mkNullOrOption (with types; either str (listOf str)) ''
-            Pattern or list of patterns, used for highlighting (vim regex)
-          '';
+          pattern =
+            helpers.defaultNullOpts.mkNullable
+            (with types; either str (listOf str))
+            ".*<(KEYWORDS)\\s*:"
+            ''
+              Pattern or list of patterns, used for highlighting (vim regex)
+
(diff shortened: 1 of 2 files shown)


traxys (commented):
This was not seen by the tests right ?could you write a mall test to avoid regressions, or does it require too much detection in neovim ?

GaetanLepage:
> This was not seen by the tests right ?could you write a mall test to avoid regressions, or does it require too much detection in neovim ?

No, it does not prevent vim from starting succesfully :/

plugins/utils/todo-comments.nix
```diff
@@ -226,16 +230,27 @@ in {
             before
             keyword
             after
-            pattern
             exclude
             ;
+          pattern =
+            helpers.ifNonNull' cfg.highlight.pattern
+            (helpers.mkRaw cfg.highlight.pattern);
```
traxys, line 237: I don't really want us to use a mkRaw here, because if people hardcode a pattern as a nix string ex: " TODO:", then it would break because we dumped this as a pure lua code, whereas an user would expect the string to be quoted.

I think we should add `[[` and `]]` to the strings, and mention it in the documentation.

traxys (changes requested):

plugins/utils/todo-comments.nix
GaetanLepage: Now it should be better.

traxys (changes requested):
There are `[[]]` in tests to change

traxys (approved):
LGTM

PR #245, opened 2023-03-13
plugins/neo-tree: init + tests

GaetanLepage:
Add [neo-tree](https://github.com/nvim-neo-tree/neo-tree.nvim) plugins and the relevant tests.


traxys (approved):
Did not see any major issues, we'll see with time

Issue #49, opened 2022-10-13
Support setting ftconfig


dwarfmaster:
There should be a way to set config specific to certain filetypes using `ftplugin` files, for both performance and convenience. For examples for setting mappings that are only relevant on certain filetypes.

I am willing to implement it, but I'd like to discuss the design first, hence this issue.

The first challenge is how to organise the options, since mostly we want to pretty much duplicate all the options for filetype specific commands, so we would need to find a way to keep it both ergonomic to use but hopefully without too much nix wizardry.

The second one is that this feature is not supported by the neovim infrastructure in `nixpkgs` which this plugin is based on. One possible workaround would be to put it in functions in `init.vim` that are then autocalled. It would allow working on this on nixvim side while in parallel trying to extend things on the nixpkgs side.

pta2002:
I think the best-case scenario for this would be something like

```nix
{
  ft."nix" = {
    plugins.nix.enable = true;
    options.numbers = true;
  };
}
```

as in, allowing the full range of nixvim options inside of it. This would definitely require some reworking though. I'd say the same options should work for autocommands, too. Not fully sure how to implement this though, but it's definitely doable with some effort

GaetanLepage:
Fixed by https://github.com/pta2002/nixvim/pull/246

PR #246, opened 2023-03-13
Enable creation of other config files than init.lua

dwarfmaster:
This is an attempt at solving #49 . It separates options into file-level (ie options that add to a file), and global options (that controls how nvim is instantiated). All plugins are file-level options. Using file-level options at the top-level adds to `init.lua`, but those options can also be used in `files.*` to create other files in the config. For example, one could do:
```nix
{
  programs.nixvim = {
    files."ftdetect/myft.lua".autoCmd = [
      {
        event = [ "BufRead" "BufNewFile" ];
        pattern = [ "*.myft" ];
        command = "set ft=myft";
      }
    ];
  };
}
```

By using options in `ftplugin/ft.lua` files, it allows setting options only in for some filetypes. Furthermore, it supports generating both lua and vim files (and guess which one to use depending on the extension of the file).

Notes:
- It should be backward compatible.
- I haven't included the top-level options in the docs. It shouldn't be too hard, but if I do that almost all options will appear twice, once at the top level and once in `files.*`, which may be problematic.
- I haven't tested the nixos module yet, but the home-manager one seems to work, both in wrapped and unwrapped mode.

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -27,28 +27,6 @@ with lib; let
   };
 in {
   options = {
-    viAlias = mkOption {
-      type = types.bool;
-      default = false;
-      description = ''
-        Symlink <command>vi</command> to <command>nvim</command> binary.
-      '';
-    };
-
-    vimAlias = mkOption {
-      type = types.bool;
-      default = false;
-      description = ''
-        Symlink <command>vim</command> to <command>nvim</command> binary.
-      '';
-    };
-
-    package = mkOption {
-      type = types.package;
-      default = pkgs.neovim-unwrapped;
-      description = "Neovim to use for nixvim";
-    };
-
     extraPlugins = mkOption {
       type = with types; listOf (either package pluginWithConfigType);
       default = [];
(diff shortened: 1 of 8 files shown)


pta2002 (approved):
I like this solution, and it feels like a nice first step :) I'll approve this so we can get CI going, since I am assuming that some things might break

GaetanLepage:
Wow, impressive work @dwarfmaster !
The diff looks good to me. I would still need to test it a bit.
Could you please fix the conflict in `modules/output.nix` ?

dwarfmaster:
I rebased it, the checks pass but for some reason my config fails to build with it, and I'm not sure I'll have the time to debug it before a few weeks. If you could check if it works on your configs it's be great.

GaetanLepage:
My config works fine !
@traxys could you please check with yours so that we are extra-sure before merging ?

dwarfmaster:
Turns out I found the issue, and indeed it isn't linked to the PR.

PR #247, opened 2023-03-13
plugins: Fetch the plugin sources from flake inputs

traxys:
Using flake inputs is easier to manage than hardcoding the rev/hash in the plugins def file


GaetanLepage (approved):
LGTM !

traxys:
@pta2002 if you could look at this too, I'm not sure if you tried to do this in the past and hit an hurdle somewhere that I missed

GaetanLepage:
Works fine for me (using home-manager).

pta2002:
I have two conflicting opinions on this:

1. This makes managing the external plugins a lot easier, since it's all in the main flake, but...
2. It fills up our flake.nix with things that are better handled elsewhere, and ends up with a ton of inputs for people who might not want to use this

Two possible solutions:
1. Create a new file, and dynamically populate the flake's inputs (something like `inputs = import plugins/external.nix // {...}`
2. Do something like nixpkgs does, with a simple list and then a script that generates the inputs using something like `nix-prefetch-git`

pta2002:
I recon something like `import` should still work, no? It's a matter of trying. Still, I'd be more in favor of the nvfetcher approach

traxys:
When doing something like this you get the following error: `
error: expected a set but got a thunk at /nix/store/8r4pmhh4szpm4i7pabjmld4g1gpkjg8a-source/flake.nix:3:3`

traxys:
@pta2002 Do you like this approach better ?

flake.nix
```diff
@@ -124,7 +127,34 @@
...
+                dontUnpack = true;
+                dontBuild = true;
+                installPhase = let
+                  script = writeShellScript "alejandra-excludes.sh" ''
+                    ${alejandra}/bin/alejandra --exclude ./plugins/_sources "$@"
+                  '';
+                in ''
+                  mkdir -p $out/bin
+                  cp ${script} $out/bin/alejandra-excludes
+                  cat $out/bin/alejandra-excludes
```
GaetanLepage, line 153: Why do you `cat` the script content here ?

traxys: Oops forgot to remove after debugging

pta2002:
I like it, is there any way to add `nvfetch` to one of the `nix run` options?

traxys:
Yeah I can look into that

PR #248, opened 2023-03-13
plugins/vim-bbye: init + tests

GaetanLepage:
Add plugin [vim-bbye](https://github.com/moll/vim-bbye) and the relevant tests.


traxys (approved):
ok

PR #249, opened 2023-03-13
neo-tree: Allow other plugins to insert their sources

traxys:
Some plugins define neo-tree sources. For theese plugins having an option 'neoTreeIntegration = true' would be useful. The problem is that if we set the plugins.neo-tree.sources to ["<source>"] then we ditch the default values of ["filesystem" ...]. The idea here is to define an internal extraSources option that takes care of the following:

- If sources is null set it to the default sources
- Concatenate sources and extraSources


GaetanLepage (approved):
Very good idea.
LGTM :)

PR #251, opened 2023-03-14
plugins/null-ls: refactoring + tests

GaetanLepage:
Refactoring of null-ls: added more options.
I also wrote tests.
Fixes https://github.com/pta2002/nixvim/issues/32.

### DIFF
--- a/plugins/null-ls/default.nix
+++ b/plugins/null-ls/default.nix
@@ -12,38 +12,237 @@ in {
     ./servers.nix
   ];
 
-  options.plugins.null-ls = {
-    enable = mkEnableOption "null-ls";
+  options.plugins.null-ls =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "null-ls";
 
-    package = mkOption {
-      type = types.package;
-      default = pkgs.vimPlugins.null-ls-nvim;
-      description = "Plugin to use for null-ls";
+      package = mkOption {
+        type = types.package;
+        default = pkgs.vimPlugins.null-ls-nvim;
+        description = "Plugin to use for null-ls";
+      };
+
+      border = helpers.defaultNullOpts.mkStr "null" ''
+        Defines the border to use for the `:NullLsInfo` UI window.
+        Uses `NullLsInfoBorder` highlight group (see [Highlight Groups](#highlight-groups)).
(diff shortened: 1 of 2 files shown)


plugins/null-ls/default.nix
```diff
@@ -12,38 +12,244 @@ in {
...
+
+        Note that by default, Neovim will not display updated diagnostics in insert mode.
+        Together with the option above, you need to pass `update_in_insert = true` to
+        `vim.diagnostic.config` for diagnostics to work as expected.
+        See `:help vim.diagnostic.config` for more info.
+      '';
+
+      sourcesItems = mkOption {
+        default = null;
+        # type = with types; nullOr (either (listOf str) (listOf attrsOf str));
```
traxys, line 202: I think this is a mistake

plugins/null-ls/default.nix
```diff
@@ -12,38 +12,244 @@ in {
...
+        type = with types; nullOr (listOf (attrsOf str));
+        description = "The list of sources to enable, should be strings of lua code. Don't use this directly";
+      };
+
+      # sources = mkOption {
+      #   default = null;
+      #   type = with types; nullOr attrs;
+      # };
```
traxys, line 210: Also this one

plugins/null-ls/default.nix
```diff
@@ -12,38 +12,244 @@ in {
...
+          (helpers.mkRaw cfg.onAttach);
+        on_init =
+          helpers.ifNonNull' cfg.onInit
+          (helpers.mkRaw cfg.onInit);
+        on_exit =
+          helpers.ifNonNull' cfg.onExit
+          (helpers.mkRaw cfg.onExit);
+        root_dir =
+          helpers.ifNonNull' cfg.rootDir
+          (helpers.mkRaw cfg.rootDir);
```
traxys, line 243: We should  maybe look into a helper for this in the future

traxys (changes requested):

plugins/null-ls/default.nix
GaetanLepage: Fixed !

plugins/null-ls/default.nix
GaetanLepage: yes it could be useful...

traxys (approved):

PR #252, opened 2023-03-14
plugins/nvim-lsp: add filetypes and autostart options for LS

GaetanLepage:
Add the `filetypes` and `autostart` options for each language server.
Fixes https://github.com/pta2002/nixvim/issues/96.


traxys (approved):
ok

PR #260, opened 2023-03-15
tests: Allow to skip running neovim

traxys:
Some plugins don't really play nicely with being launched in a headless sandboxed environment. This adds the pseudo option 'tests.dontRun' to avoid running those tests


GaetanLepage (approved):
LGTM

PR #261, opened 2023-03-15
plugins/nvim-osc52: init + tests

traxys:
nvim-osc52 is a plugin that allows to copy from neovim using OSC52 control sequences (works through SSH for example)


GaetanLepage (approved):
LGTM

PR #265, opened 2023-03-16
editorconfig: Add options to configure builtin editorconfig

traxys:


traxys:
This feature is not yet able in relesed neovim, only in nightly.

houstdav000:
There are also plugins until then, like <https://github.com/gpanders/editorconfig.nvim>.

I'm interested to see this supported asap, since it's how I primarily manage indent size for different languages across editors and across repos

GaetanLepage (approved):
LGTM

PR #266, opened 2023-03-17
plugins/inc-rename: init + tests

traxys:
inc-rename is a plugin that allows to show incrementally an LSP renaming (like %s)


GaetanLepage (approved):
LGTM

PR #270, opened 2023-03-18
plugins/packer: improve lua generation

GaetanLepage:
When not providing any lua_rocks plugin, avoid to write `use_rocks {}` in the config.


traxys (approved):
LGTM

PR #274, opened 2023-03-18
plugin/luasnip: Add jsregexp lua dependency

traxys:
The lua dependency is required to perform transforms (luasnip can fallback to a simpler algorithm if it's absent). This requires us to expose the `extraLuaPackages` option of makeNeovimConfig.


GaetanLepage (approved):
LGTM

PR #280, opened 2023-03-20
plugins/nvim-autopairs: cleanup + fix mistyped option + tests

GaetanLepage:
- Finally removed deprecated options
- Fix type of option `disabledFiletypes`  (fixes https://github.com/pta2002/nixvim/issues/279)
- Add tests


traxys:
The options were entirely removed by upstream right ?

PR #282, opened 2023-03-21
plugins/coq-thirdparty: init + tests

GaetanLepage:
Add the [coq-thirdparty](https://github.com/ms-jpq/coq.thirdparty) plugin.
Answer to the request: https://github.com/pta2002/nixvim/issues/278


plugins/completion/coq-thirdparty.nix
```diff
@@ -0,0 +1,46 @@
...
+  cfg = config.plugins.coq-thirdparty;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.coq-thirdparty = {
+    enable = mkEnableOption "coq-thirdparty";
+
+    package = helpers.mkPackageOption "coq-thirdparty" pkgs.vimPlugins.coq-thirdparty;
+
+    sources = mkOption {
+      type = types.listOf types.attrs;
```
traxys, line 17: This is not really a random attribute set. Per the documentation it must have   'src' and optionally shortName. Then the options are source specific.

I'd see maybe something like a listOf {
   src = ...;
   shortName = ....;
   extraOptions  = ....;
}

In addition some sources may depend on external plugins, so we could analyze the sources and enable the right plugins when we find them ? Or is it going too far ?

traxys (changes requested):

plugins/completion/coq-thirdparty.nix
```diff
@@ -0,0 +1,65 @@
...
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.coq-thirdparty = {
+    enable = mkEnableOption "coq-thirdparty";
+
+    package = helpers.mkPackageOption "coq-thirdparty" pkgs.vimPlugins.coq-thirdparty;
+
+    sources = mkOption {
+      type = types.listOf (types.submodule {
+        freeformType = types.attrsOf types.str;
```
traxys, line 18: Wouldn't that be types.attrsOf types.anything ?

tests/test-sources/plugins/completion/coq-thirdparty.nix
```diff
@@ -0,0 +1,29 @@
...
+          src = "nvimlua";
+          short_name = "nLUA";
+        }
+        {
+          src = "vimtex";
+          short_name = "vTEX";
+        }
+        {
+          src = "copilot";
+          short_name = "COP";
```
traxys, line 22: maybe shortName (depending on my previous statements) ?

plugins/completion/coq-thirdparty.nix
```diff
@@ -0,0 +1,65 @@
...
+        {src = "demo";}
+      ];
+    };
+  };
+
+  config = mkIf cfg.enable {
+    extraPlugins = [cfg.package];
+
+    extraConfigLua = ''
+      require('coq_3p')(${helpers.toLuaObject cfg.sources})
```
traxys, line 62: shortName does not have the correct case, either leave short_name or you need to filter it out

plugins/completion/coq-thirdparty.nix
```diff
@@ -0,0 +1,65 @@
...
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.coq-thirdparty;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.coq-thirdparty = {
```
traxys, line 11: We should add the extraOptions option right ?

traxys (changes requested):

plugins/completion/coq-thirdparty.nix
GaetanLepage: The `require('coq_3p')` function only takes a list of sources as input. There is no such thing as a table of options.

plugins/completion/coq-thirdparty.nix
GaetanLepage: I left `short_name` because the user can provide other custom attributes.
Hence, it makes no sense to mix camelCase and snake_case in the same option.
Furthermore, It removes the need to filter anything.

plugins/completion/coq-thirdparty.nix
GaetanLepage: Yes ! Fixed :)

PR #283, opened 2023-03-23
plugins/bufferline: refactor + tests

GaetanLepage:


plugins/bufferlines/bufferline.nix
```diff
@@ -6,359 +6,377 @@
...
+    guibg = helpers.mkNullOrOption types.str "background color";
   };
 in {
-  imports = [
-    (optionWarnings.mkDeprecatedOption {
-      option = ["plugins" "bufferline" "indicatorIcon"];
-      alternative = ["plugins" "bufferline" "indicator" "icon"];
-    })
-  ];
-
```
traxys, line 36: A reason for having deleted the warning ? Or was it deleted upstream ?

GaetanLepage: Because `mkDeprecatedOption` does not work with options that are nested in `submodules` (i.e. options inside `mkCompositOption` options).

traxys: Then we need to find a way to make it work with submodules, we can't break the options of the users like this, we need to be forgiving to deprecated names. We also should add the date to the introduction of the option, and delete it maybe something like after 2 months/ 1.5 month I'm not really sure

GaetanLepage: Yes, you are completely right.

GaetanLepage: Fixed :)

traxys (approved):

PR #285, opened 2023-03-24
presence-nvim: init plugin

SaiProton:
Adds plugin for Discord Rich Presence for Neovim: 
https://github.com/andweeb/presence.nvim

Also adds tests for this plugin.

### DIFF
--- a/plugins/utils/presence-nvim.nix
+++ b/plugins/utils/presence-nvim.nix
@@ -7,18 +7,6 @@
 with lib; let
   cfg = config.plugins.presence-nvim;
   helpers = import ../helpers.nix {inherit lib;};
-
-  # Given an attribute with the child attributes of:
-  # - Some static attribute
-  # - Some attribute which represents a lua function as a string
-  # This function will prioritize choosing the static attribute, then will
-  # fall back to the raw lua function.
-  staticOrFunc = attribute: static_name:
-    helpers.ifNonNull' attribute (
-      if attribute.${static_name} != null
-      then attribute.${static_name}
-      else helpers.mkRaw attribute.function
-    );
 in {
   options = {
     plugins.presence-nvim =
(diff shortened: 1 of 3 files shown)


plugins/utils/presence-nvim.nix
```diff
@@ -0,0 +1,291 @@
...
+          Update activity based on autocmd events.
+          If `false`, map or manually execute
+          `:lua package.loaded.presence:update()`
+        '';
+
+        neovimImageText = helpers.defaultNullOpts.mkStr ''"The One True Text Editor"'' ''
+          Text displayed when hovered over the Neovim image.
+        '';
+
+        mainImage = helpers.defaultNullOpts.mkStr ''"neovim"'' ''
```
GaetanLepage, line 41:
```suggestion
        mainImage = helpers.defaultNullOpts.mkStr "neovim" ''
```
same here

GaetanLepage (changes requested):
Hey !
Thank you for contributing :)

First, here are a minor modifications.

SaiProton:
Thanks for the tips, applied the changes :)

GaetanLepage:
Ok, thanks for addressing my comments.
I would suggest to remove your `staticOrFunc` implementation and do something simpler:

```nix
editingText = helpers.defaultNullOpts.mkNullable (types.either types.str helpers.rawType) ''"Editing %s"'' ''
  String rendered when `enableLineNumber` is set to `true` to display the current line number.
  Can also be a  lua function:
  `function(line_number: number, line_count: number): string`.
''
```

GaetanLepage:
In several other plugins, we rely on the user to use the `foo.__raw = "lua code"` syntax.

GaetanLepage (approved):
Great ! Now it is good for me :)

SaiProton:
Yeah, I didn't actually know you could actually specify raw types like that which is cool.

Also, when I updated the `buttons` option to be able to accept either a raw type or list of submodules, I noticed the documentation for the submodule attributes no longer appeared in the generated docs, so I specified the docs for those attributes under `buttons` if that's okay.

PR #286, opened 2023-03-24
plugins/magma-nvim: refactor + tests

GaetanLepage:


plugins/utils/magma-nvim.nix
```diff
@@ -1,105 +1,151 @@
...
-        default = "none";
-        example = "ueberzug";
-        description = " This configures how to display images. The following options are available:
-          none -- don't show imagesmagma_image_provider.
-          ueberzug -- use Ueberzug to display images.
-          kitty -- use the Kitty protocol to display images.";
+  optionWarnings = import ../../lib/option-warnings.nix args;
+  basePluginPath = ["plugins" "magma-nvim"];
+in
+  {
```
traxys, line 12: With the date and it would be perfect!

traxys (changes requested):

plugins/utils/magma-nvim.nix
GaetanLepage: Done :)

PR #287, opened 2023-03-24
plugins/mark-radar: refactor + tests

GaetanLepage:


traxys (changes requested):
We can't just rename options like this, we need to pass by a deprecation step with both co-existing and creating a warning

GaetanLepage:
> We can't just rename options like this, we need to pass by a deprecation step with both co-existing and creating a warning

Fixed !

plugins/utils/mark-radar.nix
```diff
@@ -3,50 +3,72 @@
...
+    })
+    (optionWarnings.mkRenamedOption {
+      option = basePluginPath ++ ["highlight_group"];
+      newOption = basePluginPath ++ ["highlightGroup"];
+    })
+    (optionWarnings.mkRenamedOption {
+      option = basePluginPath ++ ["set_default_keybinds"];
+      newOption = basePluginPath ++ ["setDefaultMappings"];
+    })
+  ];
```
traxys, line 31: Just add a date in a comment at the top to know when we can delete them

traxys (changes requested):

plugins/utils/mark-radar.nix
GaetanLepage: Done

traxys (approved):

PR #288, opened 2023-03-24
plugins/inc-rename: formatting

GaetanLepage:


plugins/nvim-lsp/inc-rename.nix
```diff
@@ -14,19 +14,24 @@ in {
...
       whether an empty new name should be previewed; if false the command preview will be cancelled
       instead
     '';
+
     showMessage = helpers.defaultNullOpts.mkBool true ''
       whether to display a `Renamed m instances in n files` message after a rename operation
     '';
-    inputBufferType = helpers.defaultNullOpts.mkNullable (types.enum ["dressing"]) "null" ''
+
+    inputBufferType = helpers.defaultNullOpts.mkEnum ["dressing"] "null" ''
```
traxys, line 31: Won't this render as `"null"` in the documentation ?

GaetanLepage: Oh yes you are right.

traxys (approved):

PR #290, opened 2023-03-24
plugins/notify: Fix typo for background_colour option

rummik:
`nvim-notify` uses `background_colour` rather than `background_color`

See: https://github.com/rcarriga/nvim-notify/blob/50d037041ada0895aeba4c0215cde6d11b7729c4/doc/nvim-notify.txt#L23

This PR additionally renames the configuration option to the correct camelCase name, and uses `optionWarnings.mkRenamedOption` to both warn about this change and map the old option to the new one.

### DIFF
--- a/plugins/utils/notify.nix
+++ b/plugins/utils/notify.nix
@@ -16,6 +16,7 @@ with lib; let
 in {
   imports = [
     (optionWarnings.mkRenamedOption {
+      # 2023-03-24
       option = basePluginPath ++ ["backgroundColor"];
       newOption = basePluginPath ++ ["backgroundColour"];
     })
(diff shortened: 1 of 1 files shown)


plugins/utils/notify.nix
```diff
@@ -7,11 +7,20 @@
...
   helpers = import ../helpers.nix {inherit lib;};
+  optionWarnings = import ../../lib/option-warnings.nix {inherit lib;};
+  basePluginPath = ["plugins" "notify"];
   icon = mkOption {
     type = types.nullOr types.str;
     default = null;
   };
 in {
+  imports = [
+    (optionWarnings.mkRenamedOption {
```
traxys, line 18: Could you add a comment of the date, in case we want to cleanup

traxys (changes requested):

plugins/utils/notify.nix
rummik: Comment added

traxys (approved):

Issue #299, opened 2023-03-28
mappings `description` is not included by genMaps


andrevmatos:
**Plugin affected**: 
**Nixpkgs channel**: unstable
**Home-manager version**: master

## Description
`description` field in `maps.*` doesn't get included in `config` param; it seems `helpers.genMaps` doesn't take them into account, even if they're described in mappings options. It'd be nice to rename it to `desc` as well, to be more in line with what people are used to in lua and disambiguate with nix options's `description`.

https://github.com/pta2002/nixvim/blob/dbd25dffd035974b37b0406a1d74c68040751344/lib/helpers.nix#L44-L77

GaetanLepage:
Fixed by #300

PR #300, opened 2023-03-28
modules/keymaps: refactoring + fix bug in keymaps generation

GaetanLepage:
This is an opinionated refactoring of the keymaps module:

1. Moved the `genMaps` function from `lib/helpers.nix` to `modules/keymaps.nix` as the letter is the only place where this function is used.
2. Factorized a bit the `genMaps` code
3. Renamed option `description` -> `desc`. No one will notice I hope as it was broken until know anyway.

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -6,70 +6,119 @@
 with lib; let
   helpers = import ../lib/helpers.nix {inherit lib;};
 
+  # These are the configuration options that change the behavior of each mapping.
+  mapConfigOptions = {
+    silent =
+      helpers.defaultNullOpts.mkBool false
+      "Whether this mapping should be silent. Equivalent to adding <silent> to a map.";
+
+    nowait =
+      helpers.defaultNullOpts.mkBool false
+      "Whether to wait for extra input on ambiguous mappings. Equivalent to adding <nowait> to a map.";
+
+    script =
+      helpers.defaultNullOpts.mkBool false
+      "Equivalent to adding <script> to a map.";
+
+    expr =
+      helpers.defaultNullOpts.mkBool false
+      "Means that the action is actually an expression. Equivalent to adding <expr> to a map.";
+
+    unique =
+      helpers.defaultNullOpts.mkBool false
+      "Whether to fail if the map is already defined. Equivalent to adding <unique> to a map.";
+
+    noremap =
+      helpers.defaultNullOpts.mkBool true
+      "Whether to use the 'noremap' variant of the command, ignoring any custom mappings on the defined action. It is highly advised to keep this on, which is the default.";
+
+    remap =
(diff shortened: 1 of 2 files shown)


GaetanLepage:
Fixes #299

modules/keymaps.nix
```diff
@@ -6,6 +6,65 @@
 with lib; let
   helpers = import ../lib/helpers.nix {inherit lib;};
 
+  # Generates maps for a lua config
+  genMaps = mode: maps: let
+    action2Config = action: {
```
traxys, line 11: maybe we can do something like
```nix
action2Config = {
    action,
    silent ? false,
    expr ? false,
    desc ? "",
    <fill in other options>
} @ args: args
```

Then we can  call `action2config {inherit action;}` in case of string and `action2config action` in case of dict ? It avoids repeating the names twice

modules/keymaps.nix
```diff
@@ -6,6 +6,65 @@
...
+          lib.filterAttrs
+          (
+            _: v:
+              if isBool v
+              then v
+              else if isString v
+              then v != ""
+              else !isNull v
+          )
+          (action2Config action);
```
traxys, line 62: Didn't we already call action2config when building normalized ? Why do we need it here too ?

modules/keymaps.nix
```diff
@@ -6,6 +6,65 @@
...
+        action = action.action;
+        config =
+          lib.filterAttrs
+          (
+            _: v:
+              if isBool v
+              then v
+              else if isString v
+              then v != ""
+              else !isNull v
```
traxys, line 60: This should not be possible once we went through the normalization step, we only have attr sets right ?

traxys (changes requested):

modules/keymaps.nix
```diff
@@ -6,70 +6,119 @@
...
+        inherit (action) <values of the config options that have been explicitly set by the user>
+      };
+    }
+    */
+    normalizeAction = action:
+      if isString action
+      # Case 1: action is a string
+      then {
+        inherit action;
+        config = helpers.emptyTable;
```
traxys, line 72: Why do we want an emptyTable ? won't a simple {} not cause any problem as it will pickup the upstream defaults ?

modules/keymaps.nix
```diff
@@ -6,70 +6,119 @@
...
+        # Case 2: action is an attrs
+        let
+          # Extract the values of the config options that have been explicitly set by the user
+          config =
+            filterAttrs (n: v: v != null)
+            (getAttrs (attrNames mapConfigOptions) action);
+        in {
+          config =
+            if config == {}
+            then helpers.emptyTable
```
traxys, line 84: Same question

traxys (approved):

PR #305, opened 2023-03-31
wrappers/standalone: handle assertions

GaetanLepage:
Assertions (which are used for deprecated and renamed options) where properly handled when using nixvim with home-manager and NixOS but not in standalone mode.
This is now fixed by applying the same logic as in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/62d5fcc12cca592022133807ff01e642c9e15a91/nixos/modules/system/activation/top-level.nix#L122-L126).

### DIFF
--- a/wrappers/standalone.nix
+++ b/wrappers/standalone.nix
@@ -9,5 +9,14 @@ default_pkgs: modules: {
   eval = lib.evalModules {
     modules = (modules pkgs) ++ [module wrap];
   };
+
+  handleAssertions = config: let
+    failedAssertions = map (x: x.message) (lib.filter (x: !x.assertion) config.assertions);
+  in
+    if failedAssertions != []
+    then throw "\nFailed assertions:\n${builtins.concatStringsSep "\n" (map (x: "- ${x}") failedAssertions)}"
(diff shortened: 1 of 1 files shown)


tests/test-sources/example-configurations/issues.nix
```diff
@@ -81,7 +81,7 @@
       nvim-tree = {
         enable = true;
         openOnSetup = true;
-        openOnTab = true;
+        tab.sync.open = true;
```
traxys, line 84: This should be in a nvim-tree PR no ? or is the test currently broken ?

tests/test-sources/plugins/filetrees/nvim-tree.nix
```diff
@@ -74,7 +74,6 @@
         ignoreDirs = [];
       };
       onAttach = "default";
-      removeKeymaps = false;
```
traxys, line 77: Same ?

tests/test-sources/example-configurations/issues.nix
GaetanLepage: Yes, I made a separate PR

traxys (approved):

PR #307, opened 2023-03-31
jdtls: add default data dir

fabianpage:


fabianpage:
https://github.com/pta2002/nixvim/pull/308/files is the better solution

PR #311, opened 2023-03-31
treesitter-playground: init plugin

fhilgers:


traxys (changes requested):
Could you please add a test for the plugin ? You can look into tests/test-sources for examples of other plugins. 
We recently added a CONTRIBUTING.md if you want more explanation of the helpers and why we use them.

plugins/languages/treesitter-playground.nix
```diff
@@ -0,0 +1,67 @@
...
+      ];
+
+      extraPlugins = [cfg.package];
+
+      plugins.treesitter.moduleConfig.playground = {
+        enable = true;
+        disable = cfg.disabledLanguages;
+        updatetime = cfg.updateTime;
+        persist_queries = cfg.persistQueries;
+        keybindings = mkIf (cfg.keybindings != null) {
```
GaetanLepage, line 53:
```suggestion
        keybindings = helpers.ifNonNull' cfg.keybindings {
```

traxys (approved):

PR #314, opened 2023-04-03
added tmux navigator

fabianpage:


plugins/utils/tmux-navigator.nix
```diff
@@ -0,0 +1,42 @@
...
+  options.plugins.tmux-navigator = {
+    enable = mkEnableOption "Enable Tmux-Navigator (see https://github.com/christoomey/vim-tmux-navigator for tmux installation instruction)";
+
+    package = helpers.mkPackageOption "tmux-navigator" pkgs.vimPlugins.tmux-navigator;
+
+    tmuxNavigatorSaveOnSwitch = helpers.mkNullOrOption lib.types.int ''
+      null: don't save on switch
+      1: update (write the current buffer, but only if changed)
+      2: wall (write all buffers)
+    '';
```
traxys, line 20: If this is an enum couldn't you use a `helpers.mkNullOrOption (lib.types.enum [1 2])` to be more type aware ? Morever can you add to the description the upstream default value

plugins/utils/tmux-navigator.nix
```diff
@@ -0,0 +1,42 @@
...
+
+    tmuxNavigatorDisableWhenZoomed = helpers.mkNullOrOption lib.types.int ''
```
traxys, line 22: Same as the other option

traxys (changes requested):
Could you also add a test with the empty  configuration and the default configuration please ?

plugins/utils/tmux-navigator.nix
fabianpage: ah nice, didn't knew this type.

plugins/utils/tmux-navigator.nix
fabianpage: here only null or 1 is allowed, should i put helpers.mkNullOrOption (lib.types.enum [1]) ?

traxys: It's a bit verbose but yes I think it's usefull

Issue #321, opened 2023-04-06
Plugin Request: nvim-ts-rainbow


cyntheticfox:
**Plugin**: nvim-ts-rainbow
**Plugin repository**: https://github.com/p00f/nvim-ts-rainbow
**Is the plugin on nixpkgs**: Yes

## Extra information
Needs support for color options, and hard dependency on tree-sitter plugin

GaetanLepage:
Done !
I chose to set the [nvim-ts-rainbow2](https://gitlab.com/HiPhish/nvim-ts-rainbow2) version by default because it seems to be the most maintained one.

PR #323, opened 2023-04-06
plugins/treesitter-rainbow: add plugin + test

GaetanLepage:
Add plugin [nvim-ts-rainbow2](https://gitlab.com/HiPhish/nvim-ts-rainbow2).
Fulfills #321.


plugins/languages/treesitter/treesitter-rainbow.nix
```diff
@@ -0,0 +1,66 @@
...
+        helpers.defaultNullOpts.mkNullable (types.listOf types.str) "[]"
+        "List of languages you want to disable the plugin for.";
+
+      query =
+        helpers.defaultNullOpts.mkStr "rainbow-parens"
+        "The query to use for finding delimiters.";
+
+      strategy =
+        helpers.defaultNullOpts.mkStr "require('ts-rainbow').strategy.global"
+        "The query to use for finding delimiters.";
```
traxys, line 27: Mention in the documentaiton that this should be lua code

traxys (changes requested):
Could you check in the other treesitter  plugins if we have the warning for when tree-sitter is not  enabled while you are at it ?

plugins/languages/treesitter/treesitter-rainbow.nix
GaetanLepage: Done

traxys (approved):

PR #325, opened 2023-04-07
plugins/nvim-lsp: add terraform-ls language server

devusb:
Add [language server for terraform](https://github.com/hashicorp/terraform-ls).


GaetanLepage (approved):
LGTM.
Thanks for the contribution !

PR #330, opened 2023-04-13
plugins/spider: add plugin + test

GaetanLepage:
Add plugin [nvim-spider](https://github.com/chrisgrieser/nvim-spider)

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -87,6 +87,7 @@
     ./utils/project-nvim.nix
     ./utils/presence-nvim.nix
     ./utils/specs.nix
+    ./utils/spider.nix
     ./utils/startify.nix
     ./utils/surround.nix
     ./utils/tmux-navigator.nix
--- a/plugins/utils/spider.nix
+++ b/plugins/utils/spider.nix
@@ -1,0 +1,79 @@
+{
+  pkgs,
+  config,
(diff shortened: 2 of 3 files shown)


fabianpage:
Looks good for me. The only thing is that this plugin doesn't set any keymappings. So to use it you always need to define the mapping, would you also add something for that? The mapping which is recommended by the plugin author:
```
vim.keymap.set({"n", "o", "x"}, "w", "<cmd>lua require('spider').motion('w')<CR>", { desc = "Spider-w" })
vim.keymap.set({"n", "o", "x"}, "e", "<cmd>lua require('spider').motion('e')<CR>", { desc = "Spider-e" })
vim.keymap.set({"n", "o", "x"}, "b", "<cmd>lua require('spider').motion('b')<CR>", { desc = "Spider-b" })
vim.keymap.set({"n", "o", "x"}, "ge", "<cmd>lua require('spider').motion('ge')<CR>", { desc = "Spider-ge" })
```

GaetanLepage:
> Looks good for me. The only thing is that this plugin doesn't set any keymappings. So to use it you always need to define the mapping, would you also add something for that? The mapping which is recommended by the plugin author:

Done !

traxys (approved):

PR #335, opened 2023-04-15
editorconfig: Migrate to builtin plugin

traxys:
As we updated to neovim 0.9.0 we can use the included editorconfg


GaetanLepage (approved):
LGTM !

PR #336, opened 2023-04-15
modules/lua-loader: add support for the new lua-loader

GaetanLepage:
Add support for the experimental lua-loader introduced in neovim 0.9.
Documentation: https://neovim.io/doc/user/lua.html#lua-loader

### DIFF
--- a/modules/lua-loader.nix
+++ b/modules/lua-loader.nix
@@ -1,0 +1,35 @@
+{
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.luaLoader;
+in {
+  options.luaLoader = {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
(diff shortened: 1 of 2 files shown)


modules/lua-loader.nix
```diff
@@ -0,0 +1,35 @@
...
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.luaLoader;
+in {
+  options.luaLoader = {
+    enable = mkOption {
+      type = types.bool;
+      default = true;
```
traxys, line 12: Shouldn't we set it to false by default ? As it is false right now, we want to forward neovim defaults right ?

GaetanLepage: Indeed. Fixed !

traxys (approved):

PR #339, opened 2023-04-17
plugins/conjure: init plugin

pta2002:
Creates an option for [Conjure](https://github.com/Olical/conjure). No options,
because there really isn't anything to configure.


traxys (approved):
LGTM

PR #341, opened 2023-04-19
plugins/nvim-lsp: add per-server onAttach option

GaetanLepage:
Add a new option `plugins.lsp.servers.<name>.onAttach` that lets the user override `plugins.lsp.onAttach` by providing a server-specific implementation of the `on_attach` callback.


traxys:
I'd suggest to either name the function overrideOnAttach or to make it still run the defaut on attach, if we have plugins that set the global onAttach for them to work it would be a bit counter intuitive

traxys (approved):

PR #343, opened 2023-04-21
modules/output: Allow to specify text to add as extra files

traxys:
To enable some features (like adding tree-sitter queries) we need to add files to specific directories in the runtime path (queries/lang/file.scm for tree-sitter queries for example).

This commit adds support for specifying such files. You must be careful to not have any collisions between `files` and `extraFiles`.


GaetanLepage (approved):
LGTM

GaetanLepage (approved):
Good !

PR #344, opened 2023-04-21
plugins/treesitter: Add injections to highlight lua in nixvim options

traxys:
This can look something like this: 
image


GaetanLepage (approved):
Good job !

PR #346, opened 2023-04-24
modules/clipboard: add new module to manage clipboard option and provider

GaetanLepage:
Add a new module that lets the user easily manage the clipboard.
It allows to add a provider package.

### DIFF
--- a/modules/clipboard.nix
+++ b/modules/clipboard.nix
@@ -1,0 +1,56 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.clipboard;
+in {
+  options = {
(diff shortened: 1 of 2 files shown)


modules/clipboard.nix
```diff
@@ -0,0 +1,63 @@
...
+            (enum ["xclip" "xsel"])
+          );
+        default = null;
+        example = "xclip";
+      };
+    };
+  };
+
+  config = {
+    extraConfigLuaPre =
```
traxys, line 42: Can't we use the nixvim options.clipboard ?

GaetanLepage: Definitely. Done

traxys (approved):

PR #351, opened 2023-04-27
plugins/nvim-lsp: add extraOptions for nvim-lsp language servers

GaetanLepage:
Some `nvim-lsp` language servers need to be configured with options that are not explicitly wrapped by nixvim.
This commit adds an `extraOptions` to each language server.

Example:
```nix
plugins.nvim-lsp.servers = {
  ruff-lsp = {
    enable = true;
    extraOptions = {
      init_options.settings.args = [
        "--config=/path/to/config.toml"
      ];
    };
  };
};
```

### DIFF
--- a/plugins/nvim-lsp/helpers.nix
+++ b/plugins/nvim-lsp/helpers.nix
@@ -20,6 +20,7 @@
     pkgs,
     config,
     lib,
+    options,
     ...
   }:
     with lib; let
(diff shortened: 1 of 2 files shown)


traxys (commented):
Isn't the extraSettings and extraOptions redunduant ? maybe we should deprate extraSettings in favour of extraOptions.settings ?

GaetanLepage:
> Isn't the extraSettings and extraOptions redunduant ? maybe we should deprate extraSettings in favour of extraOptions.settings ?

Indeed we could... I can make a deprecation warning

traxys:
Yes please

traxys (approved):

PR #356, opened 2023-05-08
envrc will now create .direnv folder

Alexnortung:
For direnv users, just using `direnv allow` will not always work

We could also consider removing the use flake function from `.envrc` since it should be built into direnv.


Alexnortung:
Good idea, it seems to be working great for me!

traxys (approved):

PR #357, opened 2023-05-08
Added futhark lsp

Alexnortung:


GaetanLepage:
Could you please add this server to the nvim-lsp `all-servers` test case ?

GaetanLepage:
I added the test but it seems that `futhark` is currently broken on nixpkgs.
Hence, I suggest that we wait for it to be fixed before merging.

Alexnortung:
> I added the test

Thanks!

> `futhark` is currently broken on nixpkgs

Where do you find that? I can run it just fine locally and checking `mkDerivation` function on nixpkgs, i can't find anywhere it says broken.
Also the tests seem to pass after updating the nixpkgs input, so it may be that futhark was broken in an earlier version?

GaetanLepage:
> Where do you find that? I can run it just fine locally and checking `mkDerivation` function on nixpkgs, i can't find anywhere it says broken.
> Also the tests seem to pass after updating the nixpkgs input, so it may be that futhark was broken in an earlier version?

Yes it was failing in the previous version of nixos-unstable but now it seems fine !
Maybe, just re-organize the commits to have
1. One that updates the flake inputs
2. A second one that adds the LSP server

Alexnortung:
That should do it :D

GaetanLepage (approved):

GaetanLepage:
Merged ! Thank you @Alexnortung :)

PR #360, opened 2023-05-13
plugins/catppuccin: add colorscheme + test

Tanish2002:
Added the catppuccin colorscheme (https://github.com/catppuccin/nvim)

### DIFF
--- a/plugins/colorschemes/catppuccin.nix
+++ b/plugins/colorschemes/catppuccin.nix
@@ -1,0 +1,352 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+} @ args:
+with lib; let
+  cfg = config.colorschemes.catppuccin;
+  helpers = import ../helpers.nix args;
+in {
+  options = {
+    colorschemes.catppuccin = {
+      enable = mkEnableOption "catppuccin";
+
+      package =
+        helpers.mkPackageOption "catppuccin" pkgs.vimPlugins.catppuccin-nvim;
+
+      flavour = helpers.defaultNullOpts.mkEnumFirstDefault [
+        "latte"
+        "mocha"
+        "frappe"
+        "macchiato"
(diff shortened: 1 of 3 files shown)


GaetanLepage (changes requested):
Very good !
I personally prefer when the option declarations are spaced (1 blank line between each).
Besides that, I have nothing to say, good job :)

GaetanLepage:
Seems like the pipeline is failing.
Be sure to
```sh
# Format the code
nix fmt

# Run the tests
nix flake check
```

GaetanLepage:
I pushed to your branch to do a little bit of formatting.
I changed the type of the `customHighlights` option to be either an `attrsOf attrs` or a `str` (the code of a lua function).
The implementation part for this is lacking.

Tanish2002:
I believe we could just simply do a `str` for a lua function and add a example in the description.
Otherwise do you have a idea for implementing this :thinking:

GaetanLepage:
> @GaetanLepage I tried to implement the `customHighlights` option such that it can be a `str` or a `attrsOf attrs`. I believe it works.

This implementation looks good to me.
The only issue is the `@` prefix of the keys of the attrs.
Indeed, they will be ignored by nixvim when being translated to lua:
https://github.com/pta2002/nixvim/blob/3600698abad00756947eb3a3d837beedf1c2f551/lib/helpers.nix#L22-L23

I doubt that it would work... But the user can always fall back on a custom function declaration.

Tanish2002:
yeah, for complex configs the user could fallback to the string.
Do I need to work on anything else?

GaetanLepage:
> yeah, for complex configs the user could fallback to the string. Do I need to work on anything else?

No, I think that it is alright like this

GaetanLepage (approved):
Approved

PR #361, opened 2023-05-13
Added test for template and fixed check export for lib

Alexnortung:
lib/default.nix was missing the check file, which is now moved to test-derivation.nix.
Here it was important to export both functions as the config for mkTestDerivation does not support imports from a configuration.

the check workflow also got a test for the template so we can keep it up to date, when making some breaking changes.

### DIFF
--- a/templates/_wrapper/simple/flake.nix
+++ b/templates/_wrapper/simple/flake.nix
@@ -1,0 +1,15 @@
+{
+  inputs = {
+    nixvim.url = "path:../../..";
+    # flake-utils = {
+    #   url = "github:numtide/flake-utils";
+    # };
+    simple = {
+      url = "path:../../simple";
+      inputs.nixvim.follows = "nixvim";
+      inputs.flake-utils.follows = "nixvim/flake-utils";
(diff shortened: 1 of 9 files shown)


traxys (changes requested):
Could you add a README in the templates/_wrapper directory to explain that this directory stores tests for the templates ?

Alexnortung:
Sure! I am not sure what you mean by the same command in the pre-commit hook? Do you mean I should add the check command to the pre-commit hooks?

traxys:
Sorry I meant the command in the action workflow

templates/_wrapper/simple/flake.nix
```diff
@@ -0,0 +1,15 @@
+{
+  inputs = {
+    nixvim.url = "path:../../..";
+    # flake-utils = {
+    #   url = "github:numtide/flake-utils";
+    # };
```
GaetanLepage, line 6: Shouldn't this be removed ?
```suggestion
```

traxys (approved):

PR #366, opened 2023-05-14
modules/commands: allow creation of user commands

pta2002:
Allow the creation of custom user commands, using the `userCommands` option.

### DIFF
--- a/modules/commands.nix
+++ b/modules/commands.nix
@@ -1,0 +1,79 @@
+{
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  helpers = import ../lib/helpers.nix {inherit lib;};
+  commandAttributes = types.submodule {
+    options = {
+      command = mkOption {
+        type = types.str;
+        description = "The command to run";
+      };
+
+      nargs = mkOption {
+        type = types.nullOr (types.enum ["0" "1" "*" "?" "+"]);
+        default = null;
+        description = "The number of arguments to expect, see :h command-nargs";
+      };
+      complete = mkOption {
+        # TODO: This can take a lua function
+        type = types.nullOr types.str;
+        default = null;
+        description = "Tab-completion behaviour, see :h command-complete";
+      };
+      range = mkOption {
+        type = types.nullOr (types.oneOf [
+          types.bool
+          (types.enum ["%"])
+          types.int
+        ]);
+        default = null;
+        description = "Whether the command accepts a range, see :h command-range";
+      };
+      count = mkOption {
+        type = types.nullOr (types.oneOf [
+          types.bool
+          types.int
+        ]);
+        default = null;
+        description = "Whether the command accepts a count, see :h command-range";
+      };
+      addr = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Whether special characters relate to other things, see :h command-addr";
+      };
+      bang = helpers.defaultNullOpts.mkBool false "Whether this command can take a bang (!)";
+      bar = helpers.defaultNullOpts.mkBool false "Whether this command can be followed by a \"|\" and another command";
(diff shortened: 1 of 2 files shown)


modules/commands.nix
```diff
@@ -0,0 +1,79 @@
...
+
+      # TODO: command-preview, need to grab a function here.
+    };
+  };
+in {
+  options.userCommands = mkOption {
+    type = types.attrsOf commandAttributes;
+    default = {};
+    description = "A list of user commands to add to the configuration";
+  };
```
GaetanLepage, line 63: Shouldn't this be a list ?
```suggestion
  options.userCommands = mkOption {
    type = types.listOf commandAttributes;
    default = [];
    description = "A list of user commands to add to the configuration";
  };
```

pta2002: Nope! The attribute specifies the name of the new command, so for example, I want to bind `:Write` to `:w` I'd do `Write = { command = ":w<CR>"; }`

tests/test-sources/modules/commands.nix
```diff
@@ -0,0 +1,10 @@
+{
+  example = {
+    userCommands = {
+      "W" = {
+        command = ":W<CR>";
```
Alexnortung, line 5: Should this be `:w<CR>` instead of `:W<CR>`?

pta2002: Oh yeah, doesn't really matter since it's the test but I'll change it.

modules/commands.nix
```diff
@@ -0,0 +1,79 @@
...
+
+  config = let
+    cleanupCommand = _: cmd: {
+      command = cmd.command;
+      options = filterAttrs (name: _: name != "command") cmd;
+    };
+  in
+    mkIf (config.userCommands != []) {
```
GaetanLepage, line 71:
```suggestion
    mkIf (config.userCommands != {}) {
```

modules/commands.nix
GaetanLepage: Indeed !

GaetanLepage:
Apart from the 2 remaining TODOs, this seems very good to me :)

modules/commands.nix
```diff
@@ -0,0 +1,79 @@
...
+        description = "Whether the command accepts a range, see :h command-range";
+      };
+      count = mkOption {
+        type = types.nullOr (types.oneOf [
+          types.bool
+          types.int
+        ]);
+        default = null;
+        description = "Whether the command accepts a count, see :h command-range";
+      };
```
GaetanLepage, line 42:
```suggestion
      count = helpers.mkNullOrOption (with types; either bool int) ''
        Whether the command accepts a count, see :h command-range.
      '';
```

GaetanLepage (changes requested):

modules/commands.nix
```diff
@@ -0,0 +1,79 @@
...
+        description = "Whether special characters relate to other things, see :h command-addr";
+      };
+      bang = helpers.defaultNullOpts.mkBool false "Whether this command can take a bang (!)";
+      bar = helpers.defaultNullOpts.mkBool false "Whether this command can be followed by a \"|\" and another command";
+      register = helpers.defaultNullOpts.mkBool false "The first argument to the command can be an optional register";
+      keepscript = helpers.defaultNullOpts.mkBool false "Do not use the location of where the user command was defined for verbose messages, use the location of where the command was invoked";
+      force = helpers.defaultNullOpts.mkBool false "Overwrite an existing user command";
+      desc = helpers.defaultNullOpts.mkStr "" "A description of the command";
+
+      # TODO: command-preview, need to grab a function here.
```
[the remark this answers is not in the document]
pta2002: I think we can leave this for later, it's quite a niche feature basically only useful for plugin authors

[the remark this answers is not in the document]
GaetanLepage: Ok great !

PR #373, opened 2023-05-21
feat: register mapping with no actions in which-key when enabled

Tanish2002:
With these changes I can create bindings without a action attribute when which-key is enabled.
This is helpful for nested bindings.
Example: 
\<leader>ff = something
\<leader>fw = something_else

In which key when you press \<leader> it'll show "\<leader>f" as `+prefix`. We can now do
```nix
"<leader>f" = {
  desc = " Files";
};
```
so it shows the desc in which-key instead of just `+prefix`

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -105,10 +105,14 @@ with lib; let
       options =
         mapConfigOptions
         // {
-          action = mkOption {
-            type = types.str;
-            description = "The action to execute.";
-          };
+          action =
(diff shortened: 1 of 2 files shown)


GaetanLepage:
Could you please add some test case for this in the `which-key.nix` test file ?

Tanish2002:
done!

GaetanLepage (approved):
LGTM !

PR #377, opened 2023-05-22
nvim-lsp: add installLanguageServer option for servers

GaetanLepage:
By default, for each enabled `plugins.lsp.servers`, `nixvim` installs the corresponding language server thanks to `nixpkgs`.
For some reason, a user might want to disable this behavior and limit the role of `nixvim` to wrap the `nvim-lspconfig` options without interfering more.

Hence, this PR adds a `installLanguageServer` option (defaulting to `true`) that, if disabled sets the `cmd` option default value to `null` and ignores the `package` option value.

Please note that the `package` option does not accept `null` anymore.

### DIFF
--- a/plugins/lsp/helpers.nix
+++ b/plugins/lsp/helpers.nix
@@ -32,7 +32,7 @@
         then {
           package = mkOption {
             default = package;
-            type = types.nullOr types.package;
+            type = types.package;
           };
         }
         else {};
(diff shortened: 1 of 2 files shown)


traxys:
Maybe an option wrapPackages set to true by default ? Negative booleans are best avoided I find

traxys (approved):

PR #378, opened 2023-05-22
plugins/alpha-nvim: add alpha-nvim plugin module + tests

CaptainEureka:
Added the alpha greeter for Neovim https://github.com/goolord/alpha-nvim).

Feel free to comment on the design choice(s). There's basically two options:
1. override the default themes: startify, dashboard, theta (I'm not sure how to implement this)
2. Allow access to the lower level API through the opts (I think this fits best the Nixvim philosophy)

A few things I'm not sure how to implement but would be nice to have:
- allowing the user to provide a `button` function to be used in the sections.
- Providing guardrails so a user can't mix configuration options i.e. only padding sections can have val is int etc.

### DIFF
--- a/plugins/utils/alpha.nix
+++ b/plugins/utils/alpha.nix
@@ -1,0 +1,232 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.alpha;
+
+  helpers = import ../helpers.nix {inherit lib;};
+
+  sectionType = types.enum ["group" "padding" "text"];
+  themeType = types.enum ["dashboard" "startify" "theta"];
+
+  mkAlphaSectionOption = type:
+    types.submodule {
+      options = {
+        type = mkOption {
+          type = type;
+          description = "Type of section";
+        };
+
+        val = mkOption {
+          type = with types;
+            oneOf [
+              (either str int)
+              (listOf (
+                either
+                str
+                (submodule {
+                  options = {
+                    shortcut = helpers.mkNullOrOption str "Shortcut for keymap";
+                    desc = helpers.mkNullOrOption str "Description to display for keymap";
+                    command = helpers.mkNullOrOption str "Command to run for keymap";
+                  };
+                })
+              ))
+            ];
+          default = null;
+          description = "Value for section";
+        };
+
+        opts = mkOption {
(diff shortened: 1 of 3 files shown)


plugins/utils/alpha.nix
```diff
@@ -0,0 +1,232 @@
...
+                }
+                {
+                  shortcut = "SPC sr";
+                  desc = "  Recently used files";
+                  command = ":Telescope oldfiles <CR>";
+                }
+                {
+                  shortcut = "SPC sg";
+                  desc = "  Find text";
+                  command = ":Telescope live_grep <CR>";
```
Tanish2002, line 137: These default shortcuts are dependent on Telescope and will fail if Telescope is disabled.

GaetanLepage: Yes but I don't feel like we should do anything here.
I would find it a bit over-killed to automatically install telescope for instance.

CaptainEureka: I agree, we don't need to go as far as installing Telescope with Alpha. Though more minimal shortcuts are appropriate.

plugins/utils/alpha.nix
```diff
@@ -0,0 +1,222 @@
...
+  helpers = import ../helpers.nix {inherit lib;};
+
+  sectionType = types.enum ["group" "padding" "text"];
+  themeType = types.enum ["dashboard" "startify" "theta"];
+
+  mkAlphaSectionOption = type:
+    types.submodule {
+      options = {
+        type = mkOption {
+          type = type;
```
GaetanLepage, line 19:
```suggestion
          inherit type;
```

GaetanLepage (changes requested):

GaetanLepage:
Could you please add a more exhaustive testing configuration ?

CaptainEureka:
> Could you please add a more exhaustive testing configuration ?

Yes, I'll add a testcase with exhaustive configuration.

Tanish2002:
@CaptainEureka Just saw a bit of code in alpha. I believe it has a implementation of startify's fortune inside it. Couldn't find legitimate documentation on how to use it but found a issue that tells about it
https://github.com/goolord/alpha-nvim/issues/154

GaetanLepage (approved):

PR #379, opened 2023-05-23
flake: update inputs

GaetanLepage:
Regular `nix flake update`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -94,11 +94,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


traxys:
We should introduce a workflow to automatically trigger this every week I think ?

traxys (approved):

GaetanLepage:
> We should introduce a workflow to automatically trigger this every week I think ?

Yes, that would be great ! Would you know how to do it ?

PR #381, opened 2023-05-23
barbar.nvim: fix 'requires' (fixes #380)

### LINKED ISSUE #380: barbar.nvim: adds require('bufferline') in rendered init.lua
**Plugin affected**: barbar.nvim
**Nixpkgs channel**: unstable
**Home-manager version**: ref: 58eb96

## Description
Enabling barbar adds `require("bufferline")` instead of `require("barbar")` as in docs at: 

## Config
```nix
# barbar.nix
{
  programs.nixvim.plugins = {
    barbar.enable = true;
    barbar.autoHide = true;
  };
}
```

pta2002:
The config was requiring 'bufferline' instead of 'barbar', as pointed out in #380. This fixes that.

### DIFF
--- a/plugins/bufferlines/barbar.nix
+++ b/plugins/bufferlines/barbar.nix
@@ -129,7 +129,8 @@ in {
       icons =
         stateOptions
         // (
-          mapAttrs (name: description:
+          mapAttrs
+          (name: description:
             mkOption {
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Nice catch !

PR #382, opened 2023-05-23
nvim-lsp/pylsp: fix pylsp plugins

GaetanLepage:
Currently, my fancy implementation of auto-installing pylsp plugins does not have any effect.
Indeed, it is needed to change the pylsp derivation itself by adding the different plugins to its `propagatedBuildInputs`.

See https://github.com/NixOS/nixpkgs/issues/229337 for more details

### DIFF
--- a/plugins/lsp/language-servers/pylsp.nix
+++ b/plugins/lsp/language-servers/pylsp.nix
@@ -499,11 +499,15 @@ in {
   config =
     mkIf cfg.enable
     {
-      extraPackages = let
+      # WARNING: tricky stuff below:
+      # We need to fix the `python-lsp-server` derivation by adding all of the (user enabled)
+      # plugins to its `propagatedBuildInputs`.
+      # See https://github.com/NixOS/nixpkgs/issues/229337
+      plugins.lsp.servers.pylsp.package = let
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/pylsp.nix
```diff
@@ -528,21 +532,98 @@ in {
...
+        );
+
+        # Final list of pylsp plugins to install
+        pylspPlugins = nativePlugins ++ thirdPartyPlugins;
+      in
+        mkDefault (
+          # This is the final default package for pylsp
+          pkgs.python3Packages.python-lsp-server.overridePythonAttrs (
+            old: {
+              propagatedBuildInputs = pylspPlugins ++ old.propagatedBuildInputs;
```
GaetanLepage, line 625: I am not sure rather we should add `doCheck = false;` here.
In practice, the testing is annoying for the user and lasts for ~30s.
On the other hand, keeping it like this (enabled) adds extra robustness as it would cause our CI to fail in case of a problem.

What do you think ?

pta2002: Honestly, I'd say `doCheck = true`, especially since we can get CI up and running now. But we can probably add an option that allows disabling that.

pta2002 (approved):

PR #384, opened 2023-05-23
Initialize cachix

pta2002:
So, this _should_ get cachix working, at least for commits from maintainers. Since it requires a secret, I think that PRs won't commit tests to the cache unless they're made from maintainers, but any push should still keep the test cache up-to-date, so only the thing that changed will get rebuilt anyway.

Let's see how this goes :)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,8 @@
 name: Run checks on code
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
 
 jobs:
   check:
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Great ! Is it linked to the `nixvim` cachix then ?
How could we (maintainers) configure this substituter on our local machines ?

pta2002:
Yeah, it's linked to that cachix. Just need to do `cachix use nixvim` and it should work :) but let's see how the CI goes, I have terrible luck when it comes to getting it right on the first try

pta2002:
Alright, I think that works! If you check the last step, it's fetching the results straight from cachix :D

GaetanLepage:
> Alright, I think that works! If you check the last step, it's fetching the results straight from cachix :D

Good job !
LGTM :)

GaetanLepage (approved):

GaetanLepage:
Maybe you can add some instructions in the README to configure it !

PR #385, opened 2023-05-24
plugins/nvim-cmp: fix window.completion bug

GaetanLepage:
Fix a very tricky bug in the `nvim-cmp` implementation.

### DIFF
--- a/plugins/completion/nvim-cmp/default.nix
+++ b/plugins/completion/nvim-cmp/default.nix
@@ -607,38 +607,34 @@ in {
 
       inherit (cfg) view;
       window = helpers.ifNonNull' cfg.window {
-        completion = helpers.ifNonNull' cfg.window.completion {
-          inherit
-            (cfg.window.completion)
-            border
-            winhighlight
-            zindex
-            scrolloff
-            scrollbar
-            ;
-          col_offset = cfg.window.completion.colOffset;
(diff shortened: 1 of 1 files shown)


pta2002:
What was the issue? Can't really understand what changed here

plugins/completion/nvim-cmp/default.nix
```diff
@@ -607,38 +607,34 @@ in {
...
-            if isInt maxWidth
-            then maxWidth
-            else helpers.ifNonNull' maxWidth (helpers.mkRaw maxWidth);
-          max_height = let
-            inherit (cfg.window.documentation) maxHeight;
-          in
-            if isInt maxHeight
-            then maxHeight
-            else helpers.ifNonNull' maxHeight (helpers.mkRaw maxHeight);
-        };
```
GaetanLepage, line 641: > What was the issue? Can't really understand what changed here

Indeed, the indenting changed due to formatting.
Actually the problem was in this section and each `cfg.window.completion` should have been `cfg.window.documentation`. So basically it was a copy-paste error.

pta2002: Ah, got it, will mark it as resolved then :)

pta2002 (approved):

PR #388, opened 2023-05-26
plugins/nvim-cmp: update options

GaetanLepage:
- The `source[n].max_item_count` option has been removed upstream. Hence, remove the corresponding `maxItemCount` option
- Add missing `performance` options: `asyncBudget` and `maxViewEntries`


GaetanLepage:
Unfortunately, as the `nvim-cmp.sources[].maxItemCound` is nested in a `submodule` type, I cannot add a deprecation warning.

traxys (approved):

PR #395, opened 2023-05-28
plugins/typst-vim: init plugin + test

GaetanLepage:
Add support for the [`typst.vim`](https://github.com/kaarmu/typst.vim) plugin.

### DIFF
--- a/plugins/languages/typst/typst-vim.nix
+++ b/plugins/languages/typst/typst-vim.nix
@@ -1,0 +1,42 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.typst-vim;
+  helpers = import ../../helpers.nix {inherit lib;};
+in {
+  options.plugins.typst-vim =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "typst.vim";
+
+      package = helpers.mkPackageOption "typst-vim" pkgs.vimPlugins.typst-vim;
+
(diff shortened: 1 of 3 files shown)


traxys (changes requested):
We should either set the `typst_cmd` to a nix store path, or add `typst` to the extraPackages I think. The first one may be more flexible as you can easily customize the typst version you have (or set it to null to manage it externally)

plugins/languages/typst/typst-vim.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  config = mkIf cfg.enable {
+    extraPlugins = [cfg.package];
+
+    # Add the typst compiler to nixvim packages
+    extraPackages = with pkgs; [typst];
+
+    maps.normal = with cfg.keymaps;
+      helpers.mkModeMaps {inherit silent;} {
+        ${watch} = ":TypstWatch<CR>";
```
traxys, line 39: Does this work when watch `watch` is null ? I'm not sure

GaetanLepage: Yes it works. Cf the `empty` test case.
I also tested locally without any mapping.
However, maybe it is overkill to wrap keybindings for this plugin. It does not really add any value...

traxys: Ah sorry I though the keymaps were optional. It's not really a problem to provide keymaps, it helps discovery of options.

traxys (approved):

PR #415, opened 2023-06-03
plugins/treesitter: add parsersToFiletypes option

GaetanLepage:
Add a wrapper option to the `vim.treesitter.language.register` option.
This allows to map specific parsers to one or more filetypes.

I don't know if the name I chose (`parsersToFiletypes`) is a good idea.
I thought about `register`, but it is more adpated to a function name (an action) rather than this option which allows to do several _registrations_ (i.e. an attrs).

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -80,6 +80,20 @@ in {
 
       folding = mkEnableOption "tree-sitter based folding";
 
+      languageRegister = mkOption {
+        type = with types; attrsOf (either str (listOf str));
+        description = ''
+          This is a wrapping of the `vim.treesitter.language.register` function.
+          Register specific parsers to one or several filetypes.
+          The keys are the parser names and the values are either one or several filetypes.
+        '';
+        default = {};
+        example = {
+          cpp = "onelab";
+          python = ["myFiletype" "anotherFiletype"];
+        };
+      };
+
       grammarPackages = mkOption {
         type = with types; listOf package;
         default = cfg.package.passthru.allGrammars;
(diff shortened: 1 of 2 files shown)


traxys (commented):
It would make more sense to have the mapping the other way around I feel.

Having an attr set from `filetypes` to the parser used for the filetype

And in this maybe naming it something like `filteypeParser` ?

GaetanLepage:
> It would make more sense to have the mapping the other way around I feel.
> 
> Having an attr set from `filetypes` to the parser used for the filetype

Hmm actually the original function `vim.treesitter.language.register` is the following:
```
register({lang}, {filetype})              *vim.treesitter.language.register()*
    Register a parser named {lang} to be used for {filetype}(s).

    Parameters:  
      • {lang}      string Name of parser
      • {filetype}  string|string[] Filetype(s) to associate with lang
```
Thus, it seems to me that the most logical wrapper should map parsers to one or more filetypes.

traxys:
We can call the function like this by inverting the attrset (through something like `zipAttrsWith`). I feel that mapping filetypes as keys makes more sense from an user point of view.

I think more along the lines of "I want <ft> files to be like <python>" and not "I want <python> to parse <ft> files`

GaetanLepage:
> We can call the function like this by inverting the attrset (through something like `zipAttrsWith`). I feel that mapping filetypes as keys makes more sense from an user point of view.
> 
> I think more along the lines of `"I want <ft> files to be like <python>"` and not `"I want <python> to parse <ft> files"`

I agree with the fact that we can do it but I feel like we should be sticking to the upstream "signature".

traxys:
Ok for staying along side upstream, but mabye name the option something like `treesitter.languageRegister` to be easier to discover knowing the upstream method

GaetanLepage:
> Ok for staying along side upstream, but mabye name the option something like `treesitter.languageRegister` to be easier to discover knowing the upstream method

Indeed ! I just did the rename

traxys (approved):

PR #424, opened 2023-06-13
plugins/toggleterm: init + test

Tanish2002:
Added  [toggleterm.nvim](https://github.com/akinsho/toggleterm.nvim) with tests

### DIFF
--- a/plugins/utils/toggleterm.nix
+++ b/plugins/utils/toggleterm.nix
@@ -1,0 +1,168 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.toggleterm;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.toggleterm = {
+    enable = mkEnableOption "toggleterm";
+
+    package = helpers.mkPackageOption "toggleterm" pkgs.vimPlugins.toggleterm-nvim;
+
+    size =
+      helpers.defaultNullOpts.mkNullable
+      (with types; either number str) "" ''
+        Size of the terminal.
+        `size` can be a number or function
+        Example:
+        ```
+        size = 20
+        ```
+        OR
+        ```
+        size = function(term)
+          if term.direction == "horizontal" then
+            return 15
+          elseif term.direction == "vertical" then
+            return vim.o.columns * 0.4
+          end
+        end
+        ```
+      '';
+
+    openMapping = helpers.defaultNullOpts.mkStr "[[<c-\\>]]" "Setting the open_mapping key to use for toggling the terminal(s) will set up mappings for normal mode.";
+
+    onCreate = helpers.mkNullOrOption types.str "Function to run when the terminal is first created.";
+
+    onOpen = helpers.mkNullOrOption types.str "Function to run when the terminal opens.";
+
+    onClose = helpers.mkNullOrOption types.str "Function to run when the terminal closes.";
+
+    onStdout = helpers.mkNullOrOption types.str "Callback for processing output on stdout.";
+
+    onStderr = helpers.mkNullOrOption types.str "Callback for processing output on stderr.";
+
+    onExit = helpers.mkNullOrOption types.str "Function to run when terminal process exits.";
+
+    hideNumbers = helpers.defaultNullOpts.mkBool true "Hide the number column in toggleterm buffers.";
+
+    shadeFiletypes = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "[]" "";
+
+    autochdir = helpers.defaultNullOpts.mkBool false "When neovim changes it current directory the terminal will change it's own when next it's opened.";
+
+    highlights = helpers.defaultNullOpts.mkNullable (with types; (attrsOf (attrsOf str))) "{}" ''
+        Highlights which map to a highlight group name and a table of it's values.
+        Example:
+        ```
+        highlights = {
+          Normal = {
+            guibg = "<VALUE-HERE>";
+          };
+          NormalFloat = {
+            link = "Normal";
+          },
+          FloatBorder = {
+            guifg = "<VALUE-HERE>";
+            guibg = "<VALUE-HERE>";
+          };
+        };
+      ```
+    '';
+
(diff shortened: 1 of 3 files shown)


Tanish2002:
Would really like to add a `customMappings` option to setup custom key mappings. Like this: [here](https://github.com/akinsho/toggleterm.nvim#custom-terminals) and [here](https://github.com/akinsho/toggleterm.nvim#custom-terminal-usage).
Though I'm sure on how to implement this.

plugins/utils/toggleterm.nix
```diff
@@ -0,0 +1,160 @@
...
+        size = 20
+        OR
+        size = function(term)
+          if term.direction == "horizontal" then
+            return 15
+          elseif term.direction == "vertical" then
+            return vim.o.columns * 0.4
+          end
+        end
+      '';
```
GaetanLepage, line 30:
```suggestion
    size =
      helpers.defaultNullOpts.mkNullable
      (with types; either number str) "" ''
        `size` can be a number or function which is passed the current terminal.
        Example:
        ```
          size = 20
        ```
        OR
        ```
          size = function(term)
            if term.direction == "horizontal" then
              return 15
            elseif term.direction == "vertical" then
              return vim.o.columns * 0.4
            end
          end
        ```
      '';
```

Try to wrap code with "``" in documentation.

plugins/utils/toggleterm.nix
```diff
@@ -0,0 +1,160 @@
...
+
+    openMapping = helpers.defaultNullOpts.mkStr "[[<c-\\>]]" "Setting the open_mapping key to use for toggling the terminal(s) will set up mappings for normal mode";
```
GaetanLepage, line 32: Sorry for being picky but I will ask you to start sentences with a capital letter and end them with a period ('.').
This is so that we keep consistency across documentation :)

plugins/utils/toggleterm.nix
```diff
@@ -0,0 +1,160 @@
...
+
+    direction = helpers.defaultNullOpts.mkEnumFirstDefault ["vertical" "horizontal" "tab" "float"] "";
+
+    closeOnExit = helpers.defaultNullOpts.mkBool true "close the terminal window when the process exits";
+
+    shell = helpers.defaultNullOpts.mkStr "" "change the default shell";
+
+    autoScroll = helpers.defaultNullOpts.mkBool true "automatically scroll to the bottom on terminal output";
+
+    float_opts = {
```
GaetanLepage, line 91:
```suggestion
    floatOpts = {
```

plugins/utils/toggleterm.nix
```diff
@@ -0,0 +1,160 @@
...
+        ''
+          function(term)
+            return term.name
+          end
+        '' "";
+    };
+  };
+  config = let
+    setupOptions = with cfg; {
+      inherit autochdir highlights direction shell float_opts;
```
GaetanLepage, line 124:
```suggestion
      inherit autochdir highlights direction shell;
      ...
      float_opts = floatOpts;
```

GaetanLepage (changes requested):

GaetanLepage:
> Would really like to add a `customTerminals` option to setup custom key mappings. Like this: [here](https://github.com/akinsho/toggleterm.nvim#custom-terminals) and [here](https://github.com/akinsho/toggleterm.nvim#custom-terminal-usage).
> Though I'm sure on how to implement this.

You can create an option (or a set of options) so that the user can specify in some way its table of preferences.
Then, on the implementation side, you can do:
```nix
maps.normal.${cfg.terminal.key} = {
  lua = true;
  silent = cfg.terminal.keymapSilent;
  action = ''
    require('toggleterm.terminal').Terminal(${helpers.toLuaObject cfg.terminal.options}):toggle()
  '';
};
```

You can also create the function before hand using some `extraConfigLua` and then call it in the keymap.

Tanish2002:
> You can create an option (or a set of options) so that the user can specify in some way its table of preferences. Then, on the implementation side, you can do:
> 
> ```nix
> maps.normal.${cfg.terminal.key} = {
>   lua = true;
>   silent = cfg.terminal.keymapSilent;
>   action = ''
>     require('toggleterm.terminal').Terminal:new(${helpers.toLuaObject cfg.terminal.options}):toggle()
>   '';
> };
> ```
> 
> You can also create the function before hand using some `extraConfigLua` and then call it in the keymap.

How would I implement this for all the available editing modes? For e.g: If someone wants to setup a custom terminal for visual mode.

plugins/utils/toggleterm.nix
```diff
@@ -0,0 +1,168 @@
...
+
+    onOpen = helpers.defaultNullOpts.mkStr "null" "Function to run when the terminal opens.";
+
+    onClose = helpers.defaultNullOpts.mkStr "null" "Function to run when the terminal closes.";
+
+    onStdout = helpers.defaultNullOpts.mkStr "null" "Callback for processing output on stdout.";
+
+    onStderr = helpers.defaultNullOpts.mkStr "null" "Callback for processing output on stderr.";
+
+    onExit = helpers.defaultNullOpts.mkStr "null" "Function to run when terminal process exits.";
```
GaetanLepage, line 49: For all of those, as the default is `null`, use
```nix
onFoo = helpers.mkNullOrOption types.str "Description of the option.";
```

GaetanLepage (approved):
LGTM !

MarkusLohmayer:
Hi there!  think this may have been too quick as the documentation is broken after https://pta2002.github.io/nixvim/#_plugins_toggleterm_highlights 
Also,  `openMapping = "[[<c-\>]]";`, which supposedly is default, has no effect. Perhaps the other options in the broken part of the docs don't work either, but did not try.

GaetanLepage:
> Hi there! @GaetanLepage, I think this may have been too quick as the documentation is broken after https://pta2002.github.io/nixvim/#_plugins_toggleterm_highlights Also, `openMapping = "[[<c-\>]]";`, which supposedly is default, has no effect. Perhaps the other options in the broken part of the docs don't work either, but did not try.

Indeed ! Nice catch :)
Fixed by #440

PR #428, opened 2023-06-14
plugins/auto-session: init + tests

### LINKED ISSUE #426: Plugin request: auto-session
**Plugin**:  auto-session
**Plugin repository**: https://github.com/rmagatti/auto-session
**Is the plugin on nixpkgs**: https://search.nixos.org/packages?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=auto-session

## Extra information
Session Lens was once its own plugin, so it has its own enable/settings/etc. Thank you!

GaetanLepage:
Add support for the [`auto-session`](https://github.com/rmagatti/auto-session) plugin.

Fixes #426.

### DIFF
--- a/plugins/utils/auto-session.nix
+++ b/plugins/utils/auto-session.nix
@@ -1,0 +1,187 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.auto-session;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.auto-session =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "auto-session";
+
+      package = helpers.mkPackageOption "auto-session" pkgs.vimPlugins.auto-session;
+
+      logLevel =
+        helpers.defaultNullOpts.mkEnum
+        ["debug" "info" "warn" "error"]
+        "error"
+        "Sets the log level of the plugin.";
+
+      autoSession = {
+        enabled = helpers.defaultNullOpts.mkBool true ''
+          Enables/disables auto creating, saving and restoring.
+        '';
+
+        enableLastSession = helpers.defaultNullOpts.mkBool false ''
+          Whether to enable the "last session" feature.
+        '';
+
+        rootDir =
+          helpers.defaultNullOpts.mkNullable
+          (with types; either str helpers.rawType)
+          "{__raw = \"vim.fn.stdpath 'data' .. '/sessions/'\";}"
+          ''
+            Root directory for session files.
+            Can be either a string or lua code (using `{__raw = 'foo';}`).
+          '';
+
+        createEnabled = helpers.mkNullOrOption types.bool ''
+          Whether to enable auto creating new sessions
+        '';
+
+        suppressDirs = helpers.mkNullOrOption (with types; listOf str) ''
+          Suppress session create/restore if in one of the list of dirs.
+        '';
+
+        allowedDirs = helpers.mkNullOrOption (with types; listOf str) ''
+          Allow session create/restore if in one of the list of dirs.
+        '';
+
+        useGitBranch = helpers.mkNullOrOption types.bool ''
+          Include git branch name in session name to differentiate between sessions for different
+          git branches.
+        '';
+      };
+
+      autoSave = {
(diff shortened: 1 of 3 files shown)


plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "auto-session";
+
+      package = helpers.mkPackageOption "auto-session" pkgs.vimPlugins.auto-session;
+
+      logLevel =
+        helpers.defaultNullOpts.mkEnum
+        ["debug" "info" "warn" "error"]
+        "error"
```
traxys, line 21: Upstream default seems to be info ?

plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+        "Sets the log level of the plugin.";
+
+      autoSession = {
+        enable = helpers.defaultNullOpts.mkBool true ''
```
traxys, line 25: Default is nil. Do we leave `enable` when the upstream option is `enabled` ?

plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+        rootDir =
+          helpers.defaultNullOpts.mkNullable
+          (with types; either str helpers.rawType)
+          "{__raw = \"vim.fn.stdpath 'data' .. '/sessions/'\";}"
+          ''
+            Root directory for session files.
+            Can be either a string or lua code (using `{__raw = 'foo';}`).
+          '';
+
+        createEnabled = helpers.mkNullOrOption types.bool ''
```
traxys, line 42: default is true

plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+        '';
+
+        useGitBranch = helpers.mkNullOrOption types.bool ''
+          Include git branch name in session name to differentiate between sessions for different
+          git branches.
+        '';
+      };
+
+      autoSave = {
+        enable = helpers.defaultNullOpts.mkNullable types.bool "null" ''
```
traxys, line 61: do we leave `enable` when the upstream option is `enabled` ?

plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+          Whether to enable auto saving session.
+        '';
+      };
+
+      autoRestore = {
+        enable = helpers.defaultNullOpts.mkNullable types.bool "null" ''
```
traxys, line 67: `enable` or `enabled` ?

plugins/utils/auto-session.nix
```diff
@@ -0,0 +1,187 @@
...
+          Whether to enable auto restoring session.
+        '';
+      };
+
+      cwdChangeHandling =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            either
+            str
```
traxys, line 77: Why can this be a str ?

traxys (changes requested):

plugins/utils/auto-session.nix
GaetanLepage: This line suggests `"error"` no ?
https://github.com/rmagatti/auto-session/blob/7afbb149f87be279778689596c781882014f7eef/lua/auto-session/init.lua#L52

traxys: It seems like, the readme was telling `info` though....

plugins/utils/auto-session.nix
GaetanLepage: I have changed to `enabled` to stick to the upstream naming.

plugins/utils/auto-session.nix
GaetanLepage: Default is actually `true`.
Renamed to `enabled`.

plugins/utils/auto-session.nix
GaetanLepage: Default is actuall `nil`.

plugins/utils/auto-session.nix
GaetanLepage: My mistake. It can either be a table **or `false`**. I fixed it.

traxys (approved):

PR #434, opened 2023-06-18
plugins/noice: turn routes into a list

t184256:
Judging by [the routes example in noice readme](https://github.com/folke/noice.nvim#-routes) and by [`route.lua` using `list_extend` on it](https://github.com/folke/noice.nvim/blob/a3318600bc1eba2cca84e879048c1ab8d4a0262d/lua/noice/config/routes.lua#L13), it seems to me that `plugins.noice.routes` is supposed to be a list.

### DIFF
--- a/plugins/ui/noice.nix
+++ b/plugins/ui/noice.nix
@@ -264,7 +264,7 @@ in {
       '';
 
       views = helpers.defaultNullOpts.mkNullable (types.attrsOf types.anything) "{}" "";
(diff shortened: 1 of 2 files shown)


traxys (approved):
Indeed... I rember when I packaged noice.nvim it was a hassle to track down all the types....

If you are interested you can try to formalize a bit more the types (removing some `anything`) ... but that is quite a dauting work.

PR #439, opened 2023-06-20
plugins/copilot: switch to from copilot.vim to copilot.lua

### LINKED ISSUE #437: Add copilot.lua
**Plugin**: copilot.lua
**Plugin repository**: https://github.com/zbirenbaum/copilot.lua/
**Is the plugin on nixpkgs**: https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/applications/editors/vim/plugins/generated.nix#L2242

## Extra information
The current plugin uses `copilot-vim` which is great but can be a resource hog and source of crashes. It would be awesome if we could use the newer pure lua plugin which also has a nice integration as a completion source with https://github.com/zbirenbaum/copilot-cmp.

GaetanLepage:
This PR adds the support of the [`copilot.lua`](https://github.com/zbirenbaum/copilot.lua) while keeping the support for the previously used [`copilot.vim`](https://github.com/github/copilot.vim).

Fixes #437.

**EDIT:**
This is the final approach:
- The `copilot` plugin has been renamed `copilot-vim` as it wraps `copilot.vim`.
- The `copilot.lua` plugin was added under `copilot-lua`.
- The  `copilot` source for `nvim-cmp` is now using [`copilot-cmp`](https://github.com/zbirenbaum/copilot-cmp/tree/master) which depends on `copilot.lua`.
  The support for [`cmp-copilot`](https://github.com/hrsh7th/cmp-copilot) was thus dropped.

Note: A user cannot enable both plugins at the same time.


traxys:
I'm not really in favour of desiging the copilot.vim plugin as "legacy", as is it is maintained under github, and as such would be a better "source of truth".

I'd rather either rather:
- Package copilot.lua as `copilot-lua` and leave `copilot` as `copilot.vim`
- Rename the current copilot as `copilot-vim` and make `copilot` the copilot.lua implementation
- Rename the current copilot as `copilot-vim` and implement this as `copilot-lua` (I think this is the clearest option).

I don't think `nixvim` should be opinionated on what is the "best" plugin, it should be a simple mapping from Nix to lua/vim configurations.

If we want something more opinionated I think we can have another module system upon nixvim that would act like a distrubution, you could imagine something ala `LunarNixVim`.

GaetanLepage:
> I'm not really in favour of desiging the copilot.vim plugin as "legacy", as is it is maintained under github, and as such would be a better "source of truth".
> 
> I'd rather either rather:
> 
>     * Package copilot.lua as `copilot-lua` and leave `copilot` as `copilot.vim`
> 
>     * Rename the current copilot as `copilot-vim` and make `copilot` the copilot.lua implementation
> 
>     * Rename the current copilot as `copilot-vim` and implement this as `copilot-lua` (I think this is the clearest option).
> 
> 
> I don't think `nixvim` should be opinionated on what is the "best" plugin, it should be a simple mapping from Nix to lua/vim configurations.
> 
> If we want something more opinionated I think we can have another module system upon nixvim that would act like a distrubution, you could imagine something ala `LunarNixVim`.

After a bit of thinking I came to the same conclusion and now agree with your comment.
I went with your recommended solution (the third one) as it is indeed the most logical one.

traxys (approved):

PR #440, opened 2023-06-22
plugins/toggleterm: fix openMapping and documentation bug

GaetanLepage:
- Fix a bug of documentation formatting in the `highlights` option
- Fix the implementation of the `openMapping` key binding.

Thank you @MarkusLohmayer for the catch !

### DIFF
--- a/plugins/utils/toggleterm.nix
+++ b/plugins/utils/toggleterm.nix
@@ -34,7 +34,7 @@ in {
         ```
       '';
 
-    openMapping = helpers.defaultNullOpts.mkStr "[[<c-\\>]]" "Setting the open_mapping key to use for toggling the terminal(s) will set up mappings for normal mode.";
+    openMapping = helpers.defaultNullOpts.mkStr "<c-\\>" "Setting the open_mapping key to use for toggling the terminal(s) will set up mappings for normal mode.";
 
     onCreate = helpers.mkNullOrOption types.str "Function to run when the terminal is first created.";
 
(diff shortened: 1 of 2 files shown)


traxys (approved):

MarkusLohmayer:
Hi thanks for fixing it! I can now configure the terminal I want with
```
    plugins.toggleterm = {
      enable = true;
      openMapping = "<c-\\>";
      direction = "float";
      floatOpts.border = "curved";
    };
```
however, I still see two issues with the docs, namely:
the docs say that `<c-\>` is the default for `openMapping`, however
1. it should probably say that the default is `<-\\>` and
2. the default is not set somehow, i.e. without explicitly specifying `openMapping` as above it does not work.

GaetanLepage:
> Hi thanks for fixing it! I can now configure the terminal I want with
Great !

>     1. it should probably say that the default is `<-\\>` and
Hmm yes I see. I have indeed set it to `<c-\\>` in the doc but it renders as `<c-\>` once translated to html. Not sure if we can do better here.

>     2. the default is not set somehow, i.e. without explicitly specifying `openMapping` as above it does not work.
I checked and the default is indeed not this. There is no keymapping by default.
If a user wants a mapping, he needs to set it explicitly.

MarkusLohmayer:
> I have indeed set it to `<c-\\>` in the doc but it renders as `<c-\>` once translated to html. Not sure if we can do better here.

Maybe it would then just be the best option to mention in the docs that the backslash must be escaped.

PR #446, opened 2023-06-25
plugins/nvim-ufo: init + test

Wolbyte:
Added [nvim-ufo](https://github.com/kevinhwang91/nvim-ufo) + test

### DIFF
--- a/plugins/utils/nvim-ufo.nix
+++ b/plugins/utils/nvim-ufo.nix
@@ -15,7 +15,7 @@ in {
 
       package = helpers.mkPackageOption "nvim-ufo" pkgs.vimPlugins.nvim-ufo;
 
-      openFoldHighlightTimeout = helpers.defaultNullOpts.mkInt 400 ''
+      openFoldHlTimeout = helpers.defaultNullOpts.mkInt 400 ''
         Time in millisecond between the range to be highlgihted and to be cleared
         while opening the folded line, `0` value will disable the highlight
       '';
@@ -31,9 +31,9 @@ in {
         run `UfoInspect` for details if your provider has extended the kinds.
       '';
 
-      foldVirtualTextHandler = helpers.defaultNullOpts.mkStr "null" "A lua function to customize fold virtual text";
+      foldVirtTextHandler = helpers.defaultNullOpts.mkStr "null" "A lua function to customize fold virtual text";
 
-      enableGetFoldVirtualText = helpers.defaultNullOpts.mkBool false ''
+      enableGetFoldVirtText = helpers.defaultNullOpts.mkBool false ''
         Enable a function with `lnum` as a parameter to capture the virtual text
         for the folded lines and export the function to `get_fold_virt_text` field of
         ctx table as 6th parameter in `fold_virt_text_handler`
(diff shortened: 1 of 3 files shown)


plugins/utils/nvim-ufo.nix
```diff
@@ -0,0 +1,80 @@
...
+      '';
+
+      closeFoldKinds = helpers.mkNullOrOption types.attrs ''
+        After the buffer is displayed (opened for the first time), close the
+        folds whose range with `kind` field is included in this option. For now,
+        'lsp' provider's standardized kinds are 'comment', 'imports' and 'region',
+        run `UfoInspect` for details if your provider has extended the kinds.
+      '';
+
+      foldVirtualTextHandler = helpers.defaultNullOpts.mkStr "null" "A lua function to customize fold virtual text";
```
traxys, line 34: same `foldVirtTextHandler`

plugins/utils/nvim-ufo.nix
```diff
@@ -0,0 +1,80 @@
...
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.nvim-ufo =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "nvim-ufo";
+
+      package = helpers.mkPackageOption "nvim-ufo" pkgs.vimPlugins.nvim-ufo;
+
+      openFoldHighlightTimeout = helpers.defaultNullOpts.mkInt 400 ''
```
traxys, line 18: For the option names we try to map very closely to the upstream names, so I think `openFoldHlTimeout` would be more appropriate

plugins/utils/nvim-ufo.nix
```diff
@@ -0,0 +1,80 @@
...
+      closeFoldKinds = helpers.mkNullOrOption types.attrs ''
+        After the buffer is displayed (opened for the first time), close the
+        folds whose range with `kind` field is included in this option. For now,
+        'lsp' provider's standardized kinds are 'comment', 'imports' and 'region',
+        run `UfoInspect` for details if your provider has extended the kinds.
+      '';
+
+      foldVirtualTextHandler = helpers.defaultNullOpts.mkStr "null" "A lua function to customize fold virtual text";
+
+      enableGetFoldVirtualText = helpers.defaultNullOpts.mkBool false ''
```
traxys, line 36: Same `enableGetFoldVirtText`

plugins/utils/nvim-ufo.nix
```diff
@@ -0,0 +1,80 @@
...
+        Enable a function with `lnum` as a parameter to capture the virtual text
+        for the folded lines and export the function to `get_fold_virt_text` field of
+        ctx table as 6th parameter in `fold_virt_text_handler`
+      '';
+
+      preview = {
+        winConfig = {
+          border = helpers.defaultNullOpts.mkStr "rounded" "The border for preview window, `:h nvim_open_win() | call search('border:')`";
```
traxys, line 44: Border can be either a string or a list of 8 strings (maybe not checking the `8` part, but it's usefull to have the list part)

traxys (changes requested):

GaetanLepage:
Although nixvim doesn't have a clear and deterministic set of rules to name plugins, I would call this one `ufo` instead of `nvim-ufo`. What do you think ?

Wolbyte:
Given that the current code base tries to keep the name/case of the plugin (mostly) untouched, I decided to name it as is to keep things a little bit more convenient.

traxys (approved):
I agree, I would find just `ufo` a bit jarring

PR #456, opened 2023-06-29
plugins/neogen: init + test

anriha:
Add a support for [neogen](https://github.com/danymat/neogen) plugin.

### DIFF
--- a/plugins/utils/neogen.nix
+++ b/plugins/utils/neogen.nix
@@ -1,0 +1,179 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.neogen;
+  helpers = import ../helpers.nix {inherit lib;};
+
+  keymapDef = {
+    generate = {
+      command = "";
+      description = ''
+        The only function required to use Neogen.
+
+        It'll try to find the first parent that matches a certain type.
+        For example, if you are inside a function, and called `generate({ type = "func" })`,
+        Neogen will go until the start of the function and start annotating for you.
+      '';
+    };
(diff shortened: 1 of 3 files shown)


plugins/utils/neogen.nix
```diff
@@ -0,0 +1,172 @@
...
+        {
+            csharp = {
+                template = {
+                    annotation_convention = "...";
+                };
+            };
+        }
+        ```
+      '';
+      snippetEngine = helpers.defaultNullOpts.mkStr null ''
```
GaetanLepage, line 90:
```suggestion
      snippetEngine = helpers.mkNullOrOption types.str ''
```

GaetanLepage (changes requested):
Very good !

Could you simply add a line break between each option declaration (to keep formatting coherent in the project) ?
Other than that, it looks good to me :)

plugins/utils/neogen.nix
anriha: Missed that function, thanks. Btw, is there some documentation of these helper functions, or is the only option just going through the code?

GaetanLepage: Unfortunately, as the number of contributors has been fairly limited, we have not an amazing documentation for internal things.
However, you can have a look at the [CONTRIBUTING.md](https://github.com/pta2002/nixvim/blob/main/CONTRIBUTING.md) file. It gives a few information about the tools and the rules we use to package plugins.

anriha:
Added the line breaks, hopefully it should be ok now.

GaetanLepage (approved):
I changed a few minor details.
Good to go !

PR #458, opened 2023-07-01
plugins/null-ls: add eslint, eslint_d, and prettier_d sources, extend test for sources, sort attr sets

garaiza-93:
Add eslint, eslint_d, and prettier_d sources. Also sorted attribute sets alphabetically :^)

Because Nix refuses to evaluate `pkgs.nodePackages.vscode-langservers-extracted` because vscode has an unfree license, this can be an alternative.

I am aware that `allowUnfree = true` in Nix's config or using `--impure` flag on Flake commands allows that derivation to be evaluated, but others may want a out-of-the-box solution like using null-ls.

### DIFF
--- a/plugins/null-ls/servers.nix
+++ b/plugins/null-ls/servers.nix
@@ -7,6 +7,12 @@
   helpers = import ./helpers.nix args;
   serverData = {
     code_actions = {
+      eslint = {
+        package = pkgs.nodePackages.eslint;
+      };
+      eslint_d = {
+        package = pkgs.nodePackages.eslint_d;
+      };
       gitsigns = {};
       shellcheck = {
         package = pkgs.shellcheck;
@@ -17,31 +23,85 @@
     };
     completion = {};
     diagnostics = {
-      flake8 = {
-        package = pkgs.python3Packages.flake8;
-      };
-      shellcheck = {
-        package = pkgs.shellcheck;
+      alex = {
+        package = pkgs.nodePackages.alex;
       };
       cppcheck = {
         package = pkgs.cppcheck;
       };
+      deadnix = {
+        package = pkgs.deadnix;
+      };
+      eslint = {
+        package = pkgs.nodePackages.eslint;
+      };
+      eslint_d = {
+        package = pkgs.nodePackages.eslint_d;
+      };
+      flake8 = {
+        package = pkgs.python3Packages.flake8;
+      };
       gitlint = {
         package = pkgs.gitlint;
       };
-      deadnix = {
-        package = pkgs.deadnix;
+      hadolint = {
+        package = pkgs.hadolint;
+      };
+      luacheck = {
+        package = pkgs.luaPackages.luacheck;
+      };
+      mypy = {
+        package = pkgs.mypy;
+      };
+      protolint = {
+        package = pkgs.protolint;
+      };
+      pylint = {
+        package = pkgs.pylint;
+      };
+      shellcheck = {
+        package = pkgs.shellcheck;
       };
       statix = {
         package = pkgs.statix;
       };
+      vale = {
+        package = pkgs.vale;
+      };
+      vulture = {
+        package = pkgs.python3Packages.vulture;
+      };
     };
     formatting = {
-      phpcbf = {
-        package = pkgs.phpPackages.phpcbf;
-      };
       alejandra = {
         package = pkgs.alejandra;
+      };
+      black = {
+        package = pkgs.python3Packages.black;
+      };
+      cbfmt = {
+        package = pkgs.cbfmt;
+      };
+      eslint = {
+        package = pkgs.nodePackages.eslint;
+      };
+      eslint_d = {
+        package = pkgs.nodePackages.eslint_d;
+      };
+      fnlfmt = {
+        package = pkgs.fnlfmt;
+      };
+      fourmolu = {
+        package = pkgs.haskellPackages.fourmolu;
+      };
+      isort = {
+        package = pkgs.isort;
+      };
+      jq = {
+        package = pkgs.jq;
+      };
+      markdownlint = {
+        package = pkgs.nodePackages.markdownlint-cli;
       };
       nixfmt = {
         package = pkgs.nixfmt;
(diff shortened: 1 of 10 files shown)


plugins/null-ls/servers.nix
```diff
@@ -7,72 +7,39 @@
...
+      cbfmt = {package = pkgs.cbfmt;};
+      eslint = {package = pkgs.nodePackages.eslint;};
+      eslint_d = {package = pkgs.nodePackages.eslint_d;};
+      fnlfmt = {package = pkgs.fnlfmt;};
+      fourmolu = {package = pkgs.haskellPackages.fourmolu;};
+      nixfmt = {package = pkgs.nixfmt;};
+      nixpkgs_fmt = {package = pkgs.nixpkgs-fmt;};
+      phpcbf = {package = pkgs.phpPackages.phpcbf;};
+      prettier = {package = pkgs.nodePackages.prettier;};
+      prettier_d = {package = pkgs.nodePackages.prettier_d_slim;};
```
Wolbyte, line 39: I think that renaming all instances of `prettier_d` to `prettier_d_slim` would be a better option since a `prettierd` source already exists; doing so would avoid confusions.

Wolbyte (changes requested):

tests/test-sources/plugins/null-ls.nix
```diff
@@ -36,31 +34,37 @@
...
+          deadnix.enable = true;
+          eslint.enable = true;
+          eslint_d.enable = true;
           flake8.enable = true;
           gitlint.enable = true;
           shellcheck.enable = true;
           statix.enable = true;
-          deadnix.enable = true;
         };
         formatting = {
```
Wolbyte, line 53: It seems that you forgot to enable `prettier_d` in the tests

GaetanLepage:
What we have done in nixvim about the `pkgs.nodePackages.vscode-langservers-extracted` unfree license is to [set `allowUnfree` to `true` in the tests](https://github.com/pta2002/nixvim/commit/0bd5ea698495dcc165f057f0016b69d16dc42e40) so that everything can be checked by the CI.
Then, we let the user decides for himself when it comes to individual configuration.

plugins/null-ls/servers.nix
garaiza-93: This makes sense. I also noticed that null-ls has both `prettierd` and `prettier_d_slim` sources, so I may as well make this change.

tests/test-sources/plugins/null-ls.nix
garaiza-93: Good catch! Thanks

garaiza-93:
> Then, we let the user decides for himself when it comes to individual configuration.
That's a good idea. This PR is meant to provide more choice, in case the user doesn't want to bother enable unfree stuff.

Wolbyte (approved):

GaetanLepage:
> That's a good idea. This PR is meant to provide more choice, in case the user doesn't want to bother to enable unfree stuff.

Have you compared the version of those individual sources with the ones provided by `vscode-langservers-extracted` ?
A few weeks ago, I noticed that for language servers, the ones packaged individually as node packages were very out of date.

garaiza-93:
I'm not sure what you mean. They're fairly up-to-date, only behind by some minor versions.
`nodePackages.eslint` and `nodePackages_latest.eslint` are at 8.41.0 on nixpkgs unstable. Most recent version on npm is 8.44.0.
`nodePackages.prettier` and `nodePackages_latest.prettier` are at 2.8.8, which is the latest release.

traxys:
If you look at the language servers for json or html ypu can see that the nixpkgs version is really out of date

GaetanLepage:
> If you look at the language servers for json or html ypu can see that the nixpkgs version is really out of date

This is what I was referring to in my previous answer. And those language servers are the reason why we decided to use `pkgs.vscode-langservers-extracted` instead.

GaetanLepage:
I built the `pkgs.vscode-langservers-extracted` package and it only has the following outputs:
```
nixpkgs on  master [$] 
✦ ✗ l result/bin                                              
Permissions Size User Group Date Modified Name
.r-xr-xr-x   310 root root   1 Jan  1970  vscode-css-language-server*
.r-xr-xr-x   311 root root   1 Jan  1970  vscode-html-language-server*
.r-xr-xr-x   311 root root   1 Jan  1970  vscode-json-language-server*
.r-xr-xr-x   315 root root   1 Jan  1970  vscode-markdown-language-server*
```

Hence, I guess that there is no debate to have because there is a single available choice: using `pkgs.nodePackages.{eslint, eslint_d, prettier, prettier_d_slim}`.

What do you think @traxys ?

traxys:
I don't really see why we are talking about `vscode-langservers-extracted` as it does not seem to appear in the PR ? I am ok for the change, but you have conflicts due to #464

GaetanLepage:
> I don't really see why we are talking about `vscode-langservers-extracted`

The reason is that @garaiza-93 mentioned it in the text of the PR itself.

> but you have conflicts due to https://github.com/pta2002/nixvim/pull/464

Yes, it needs to be rebased.
+ I don't see why we bother changing formatting here.

garaiza-93:
The formatting is just what I got from running `nix fmt`. I have no other reason.
I'll resolve merge conflicts.

tests/test-sources/plugins/null-ls.nix
```diff
@@ -1,8 +1,6 @@
 {
   # Empty configuration
-  empty = {
-    plugins.null-ls.enable = true;
-  };
+  empty = {plugins.null-ls.enable = true;};
```
GaetanLepage, line 3:
```suggestion
  empty = {
    plugins.null-ls.enable = true;
  };
```

plugins/null-ls/servers.nix
```diff
@@ -123,12 +140,9 @@
...
   # }]
-  serverDataFormatted =
-    lib.mapAttrsToList
-    (
-      sourceType: lib.mapAttrsToList (name: attrs: attrs // {inherit sourceType name;})
-    )
-    serverData;
+  serverDataFormatted = lib.mapAttrsToList (sourceType:
+    lib.mapAttrsToList (name: attrs: attrs // {inherit sourceType name;}))
+  serverData;
```
GaetanLepage, line 145:
```suggestion
  serverDataFormatted =
    lib.mapAttrsToList
    (
      sourceType: lib.mapAttrsToList (name: attrs: attrs // {inherit sourceType name;})
    )
    serverData;
```

plugins/null-ls/servers.nix
```diff
@@ -26,6 +32,8 @@
       deadnix = {
         package = pkgs.deadnix;
       };
+      eslint = {package = pkgs.nodePackages.eslint;};
+      eslint_d = {package = pkgs.nodePackages.eslint_d;};
```
GaetanLepage, line 36:
```suggestion
      eslint = {
        package = pkgs.nodePackages.eslint;
      };
      eslint_d = {
        package = pkgs.nodePackages.eslint_d;
      };
```

GaetanLepage (changes requested):
Minor formatting requests.

GaetanLepage (approved):
All good ! Thanks :)

PR #466, opened 2023-07-02
plugins/wilder-nvim: init + test

anriha:
Add a support for [wilder.nvim](https://github.com/gelguy/wilder.nvim).

### DIFF
--- a/plugins/utils/wilder-nvim.nix
+++ b/plugins/utils/wilder-nvim.nix
@@ -1,0 +1,80 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.wilder-nvim;
+  helpers = import ../helpers.nix {inherit lib;};
+
+  boolToInt = value:
+    if value == null
+    then null
+    else if value
+    then "1"
+    else "0";
+in {
+  options.plugins.wilder-nvim =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "wilder-nvim";
+
+      package = helpers.mkPackageOption "wilder-nvim" pkgs.vimPlugins.wilder-nvim;
+
+      modes = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "['/', '?']" ''
+        List of modes which wilderw will be active in.
+        Possible elements: '/', '?' and ':'
+      '';
+
+      enableCmdlineEnter = helpers.defaultNullOpts.mkBool true ''
+        If true calls wilder#enable_cmdline_enter().
+        Creates a new |CmdlineEnter| autocmd to which will start wilder
+        when the cmdline is entered.
+      '';
+
+      wildcharm = helpers.defaultNullOpts.mkStr "&wildchar" ''
+        Key to set the 'wildcharm' option to. can be set to v:false to skip the setting.
+      '';
+
+      nextKey = helpers.defaultNullOpts.mkStr "<Tab>" ''
+        A key to map to wilder#next() providing next suggestion.
+      '';
+
+      prevKey = helpers.defaultNullOpts.mkStr "<S-Tab>" ''
+        A key to map to wilder#prev() providing previous suggestion.
+      '';
+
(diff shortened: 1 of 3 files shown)


plugins/utils/wilder-nvim.nix
```diff
@@ -0,0 +1,80 @@
...
+    else "0";
+in {
+  options.plugins.wilder-nvim =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "wilder-nvim";
+
+      package = helpers.mkPackageOption "wilder-nvim" pkgs.vimPlugins.wilder-nvim;
+
+      modes = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "['/', '?']" ''
```
traxys, line 25: The default value should be the nix version (as it is shown in the documentation and is easier to copy paste), this means `["/" "?"]`.

plugins/utils/wilder-nvim.nix
```diff
@@ -0,0 +1,80 @@
...
+        Possible elements: '/', '?' and ':'
+      '';
+
+      enableCmdlineEnter = helpers.defaultNullOpts.mkBool true ''
+        If true calls wilder#enable_cmdline_enter().
+        Creates a new |CmdlineEnter| autocmd to which will start wilder
+        when the cmdline is entered.
+      '';
+
+      wildcharm = helpers.defaultNullOpts.mkStr "&wildchar" ''
```
traxys, line 36: Well this default would be complex to express in Nix, it can be left as-is.

traxys (changes requested):

plugins/utils/wilder-nvim.nix
anriha: As far as I understand this should be the same as default right? This would just translate into lua string, which should be the same I think.

traxys: This is a vimscript expression right ? In any case the `mkStr` will add `""` around so it can't be an arbitrary expression, you would need to do an `types.either (str helpers.rawType)` for that, but I'm not sure it's worth it. You can leave it as this.

plugins/utils/wilder-nvim.nix
```diff
@@ -22,7 +22,7 @@ in {
 
       package = helpers.mkPackageOption "wilder-nvim" pkgs.vimPlugins.wilder-nvim;
 
-      modes = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "['/', '?']" ''
+      modes = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "[\"/\", \"?\"]" ''
```
traxys, line 25: There are no `,` in nix lists

traxys (changes requested):

plugins/utils/wilder-nvim.nix
traxys: Also you can use `''["/" "?"]''` to avoid escaping

plugins/utils/wilder-nvim.nix
anriha: Right always get mixed up how strings work in Nix, it should be fixed now.

plugins/utils/wilder-nvim.nix
anriha: Right, didn't notice that it actually is vimscript, at this point I just assume everything is lua. To be honest, I couldn't really find what it does anyway.

plugins/utils/wilder-nvim.nix
anriha: Thanks, Python habits die hard :smile:

traxys (approved):

s1341:
It would be nice to expose the renderer option via nixvim...

anriha:
> It would be nice to expose the renderer option via nixvim...

I could try to do that, but not really sure how useful it will be. As far as I can tell, the only option would be to just add wilder-nvim.renderer.wildmenuRenderer which would accept function as an argument. Not sure if it is worth it as opposed to just adding some extraConfigLua. But if there is a better way, I would be happy to expose it.

PR #472, opened 2023-07-05
plugins/vim-illuminate: init + test

anriha:
Add support for [vim-illuminate](https://github.com/RRethy/vim-illuminate).

### DIFF
--- a/plugins/utils/illuminate.nix
+++ b/plugins/utils/illuminate.nix
@@ -69,9 +69,9 @@ in {
   with defaultNullOpts;
     helpers.extraOptionsOptions
     // {
-      enable = mkEnableOption "illuminate";
+      enable = mkEnableOption "vim-illuminate";
 
-      package = mkPackageOption "illuminate" pkgs.vimPlugins.vim-illuminate;
+      package = mkPackageOption "vim-illuminate" pkgs.vimPlugins.vim-illuminate;
 
       filetypeOverrides = mkOption {
         type = types.listOf (types.submodule {
--- a/plugins/utils/illuminate.nix
+++ b/plugins/utils/illuminate.nix
@@ -5,7 +5,7 @@
   ...
 }:
 with lib; let
-  cfg = config.plugins.vim-illuminate;
+  cfg = config.plugins.illuminate;
   helpers = import ../helpers.nix {inherit lib;};
 
   mkListStr = helpers.defaultNullOpts.mkNullable (types.listOf types.str);
@@ -65,13 +65,13 @@ with lib; let
     '';
   };
 in {
-  options.plugins.vim-illuminate = with helpers;
+  options.plugins.illuminate = with helpers;
   with defaultNullOpts;
     helpers.extraOptionsOptions
     // {
-      enable = mkEnableOption "vim-illuminate";
+      enable = mkEnableOption "illuminate";
 
-      package = mkPackageOption "vim-illuminate" pkgs.vimPlugins.vim-illuminate;
+      package = mkPackageOption "illuminate" pkgs.vimPlugins.vim-illuminate;
 
       filetypeOverrides = mkOption {
         type = types.listOf (types.submodule {
@@ -96,7 +96,7 @@ in {
         description = ''
           Config to use for large files (based on large_file_cutoff).
           Supports the same keys passed to .configure
-          If null, vim-illuminate will be disabled for large files.
+          If null, illuminate will be disabled for large files.
         '';
         default = {};
       };
--- a/plugins/utils/vim-illuminate.nix
+++ b/plugins/utils/vim-illuminate.nix
@@ -1,0 +1,139 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.vim-illuminate;
+  helpers = import ../helpers.nix {inherit lib;};
+
+  mkListStr = helpers.defaultNullOpts.mkNullable (types.listOf types.str);
+
+  commonOptions = with helpers.defaultNullOpts; {
+    providers = mkListStr ''[ "lsp" "treesitter" "regex" ]'' ''
+      Provider used to get references in the buffer, ordered by priority
+    '';
+
+    delay = mkInt 100 ''
+      Delay in milliseconds
+    '';
+
+    modesDenylist = mkListStr "[]" ''
+      Modes to not illuminate, this overrides modes_allowlist
+      See `:help mode()` for possible values
+    '';
+
+    modesAllowlist = mkListStr "[]" ''
+      Modes to illuminate, this is overriden by modes_denylist
+      See `:help mode()` for possible values
+    '';
+
+    providersRegexSyntaxDenylist = mkListStr "[]" ''
+      Syntax to not illuminate, this overrides providers_regex_syntax_allowlist
+      Only applies to the 'regex' provider
+      Use :echom synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')
+    '';
+
+    providersRegexSyntaxAllowlist = mkListStr "[]" ''
+      Syntax to illuminate, this is overriden by providers_regex_syntax_denylist
+      Only applies to the 'regex' provider
+      Use :echom synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')
+    '';
+
+    underCursor = mkBool true ''
+      Whether or not to illuminate under the cursor
+    '';
+
+    largeFileCutoff = helpers.mkNullOrOption types.int ''
+      Number of lines at which to use large_file_config
+      The `under_cursor` option is disabled when this cutoff is hit
+    '';
+
+    minCountToHighlight = mkInt 1 ''
+      Minimum number of matches required to perform highlighting
+    '';
+  };
+
+  filetypeOptions = {
+    filetypesDenylist = mkListStr ''[ "dirvish" "fugitive" ]'' ''
+      Filetypes to not illuminate, this overrides filetypes_allowlist
+    '';
+
+    filetypesAllowlist = mkListStr "[]" ''
+      Filetypes to illuminate, this is overriden by filetypes_denylist
+    '';
+  };
+in {
+  options.plugins.vim-illuminate = with helpers;
+  with defaultNullOpts;
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "vim-illuminate";
+
+      package = mkPackageOption "vim-illuminate" pkgs.vimPlugins.vim-illuminate;
+
+      filetypeOverrides = mkOption {
+        type = types.listOf (types.submodule {
+          options = {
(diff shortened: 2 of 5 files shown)


plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+  ...
+}:
+with lib; let
+  cfg = config.plugins.vim-illuminate;
+  helpers = import ../helpers.nix {inherit lib;};
+
+  mkListStr = helpers.defaultNullOpts.mkNullable (types.listOf types.str);
+
+  commonOptions = with helpers.defaultNullOpts; {
+    providers = mkListStr ''[ "lsp" "treesitter" "regex" ]'' ''
```
GaetanLepage, line 14:
```suggestion
    providers = mkListStr ''["lsp" "treesitter" "regex"]'' ''
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+      Provider used to get references in the buffer, ordered by priority
```
GaetanLepage, line 15:
```suggestion
      Provider used to get references in the buffer, ordered by priority.
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+    '';
+
+    delay = mkInt 100 ''
+      Delay in milliseconds
```
GaetanLepage, line 19:
```suggestion
      Delay in milliseconds.
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+    '';
+
+    modesDenylist = mkListStr "[]" ''
+      Modes to not illuminate, this overrides modes_allowlist
+      See `:help mode()` for possible values
+    '';
```
GaetanLepage, line 25:
```suggestion
    modesDenylist = mkListStr "[]" ''
      Modes to not illuminate, this overrides `modes_allowlist`.
      See `:help mode()` for possible values.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    modesAllowlist = mkListStr "[]" ''
+      Modes to illuminate, this is overriden by modes_denylist
+      See `:help mode()` for possible values
+    '';
```
GaetanLepage, line 30:
```suggestion
    modesAllowlist = mkListStr "[]" ''
      Modes to illuminate, this is overriden by `modes_denylist`.
      See `:help mode()` for possible values.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    providersRegexSyntaxDenylist = mkListStr "[]" ''
+      Syntax to not illuminate, this overrides providers_regex_syntax_allowlist
+      Only applies to the 'regex' provider
+      Use :echom synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')
+    '';
```
GaetanLepage, line 36:
```suggestion
    providersRegexSyntaxDenylist = mkListStr "[]" ''
      Syntax to not illuminate, this overrides `providers_regex_syntax_allowlist`.
      Only applies to the 'regex' provider.
      Use `:echo synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')`.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    providersRegexSyntaxAllowlist = mkListStr "[]" ''
+      Syntax to illuminate, this is overriden by providers_regex_syntax_denylist
+      Only applies to the 'regex' provider
+      Use :echom synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')
+    '';
```
GaetanLepage, line 42:
```suggestion
    providersRegexSyntaxAllowlist = mkListStr "[]" ''
      Syntax to illuminate, this is overriden by `providers_regex_syntax_denylist`.
      Only applies to the 'regex' provider.
      Use `:echo synIDattr(synIDtrans(synID(line('.'), col('.'), 1)), 'name')`.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    underCursor = mkBool true ''
+      Whether or not to illuminate under the cursor
```
GaetanLepage, line 45:
```suggestion
      Whether or not to illuminate under the cursor.
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+    '';
+
+    largeFileCutoff = helpers.mkNullOrOption types.int ''
+      Number of lines at which to use large_file_config
+      The `under_cursor` option is disabled when this cutoff is hit
+    '';
```
GaetanLepage, line 51:
```suggestion
    largeFileCutoff = helpers.mkNullOrOption types.int ''
      Number of lines at which to use `large_file_config`.
      The `under_cursor` option is disabled when this cutoff is hit.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    minCountToHighlight = mkInt 1 ''
+      Minimum number of matches required to perform highlighting
```
GaetanLepage, line 54:
```suggestion
      Minimum number of matches required to perform highlighting.
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+    '';
+  };
+
+  filetypeOptions = {
+    filetypesDenylist = mkListStr ''[ "dirvish" "fugitive" ]'' ''
+      Filetypes to not illuminate, this overrides filetypes_allowlist
+    '';
```
GaetanLepage, line 61:
```suggestion
    filetypesDenylist = mkListStr ''["dirvish" "fugitive"]'' ''
      Filetypes to not illuminate, this overrides `filetypes_allowlist`.
    '';
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+
+    filetypesAllowlist = mkListStr "[]" ''
+      Filetypes to illuminate, this is overriden by filetypes_denylist
```
GaetanLepage, line 64:
```suggestion
      Filetypes to illuminate, this is overriden by `filetypes_denylist`.
```

plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+      {
+        large_file_overrides = commonSetupOptions largeFileOverrides // filetypeSetupOptions largeFileOverrides;
+
+        filetype_overrides = let
+          override = attr: {${attr.filetype} = commonSetupOptions attr.overrides;};
+        in
+          map override filetypeOverrides;
+      }
+      // filetypeSetupOptions cfg
+      // commonSetupOptions cfg;
```
GaetanLepage, line 135:
```suggestion
        large_file_overrides = (commonSetupOptions largeFileOverrides) // (filetypeSetupOptions largeFileOverrides);

        filetype_overrides = let
          override = attr: {${attr.filetype} = commonSetupOptions attr.overrides;};
        in
          map override filetypeOverrides;
      }
      // (filetypeSetupOptions cfg)
      // (commonSetupOptions cfg);
```

This one is only a suggestion. Those parenthesizes are not mandatory but help with readability (I think).

tests/test-sources/plugins/utils/vim-illuminate.nix
```diff
@@ -0,0 +1,35 @@
...
+          };
+        }
+      ];
+
+      largeFileOverrides = {
+        delay = 20;
+        underCursor = true;
+      };
+    };
+  };
```
GaetanLepage, line 34:
```suggestion
      enable = true;

      delay = 50;
      providers = ["lsp"];
      modesDenylist = ["n"];
      modesAllowlist = ["v"];
      underCursor = false;
      largeFileCutoff = 10;
      minCountToHighlight = 2;
      filetypesDenylist = ["csharp"];
      filetypesAllowlist = ["python"];
      filetypeOverrides = [
        {
          filetype = "c";
          overrides = {
            delay = 10;
            providers = ["treesitter"];
          };
        }
      ];
      largeFileOverrides = {
        delay = 20;
        underCursor = true;
      };
    };
  };
```

GaetanLepage (changes requested):
Thanks for this nice contribution !
I just have a few formatting change requests. Apart from this, it looks good.

GaetanLepage:
Also, I would rather see this plugin named `illuminate` instead of `vim-illuminate`.

anriha:
> Also, I would rather see this plugin named `illuminate` instead of `vim-illuminate`.

What is the guideline for this? Just use name that you use in lua require?

GaetanLepage:
> What is the guideline for this? Just use name that you use in lua require?

Unfortunately we do not have a strong and deterministic guideline on the naming.
The current codebase already has some discrepancies in the naming of plugins.

If there is no clash or other exception, I tend to prefer sticking to the _namespace name_ (i.e. `telescope`, `harpoon`, `treesitter`...).
So in this case, I would go for `illuminate`.

anriha:
> > What is the guideline for this? Just use name that you use in lua require?
> 
> Unfortunately we do not have a strong and deterministic guideline on the naming. The current codebase already has some discrepancies in the naming of plugins.
> 
> If there is no clash or other exception, I tend to prefer sticking to the _namespace name_ (i.e. `telescope`, `harpoon`, `treesitter`...). So in this case, I would go for `illuminate`.

Ok, makes sense, renamed :+1:

plugins/utils/illuminate.nix
```diff
@@ -0,0 +1,144 @@
...
+    '';
+  };
+in {
+  options.plugins.illuminate = with helpers;
+  with defaultNullOpts;
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "illuminate";
+
+      package = mkPackageOption "illuminate" pkgs.vimPlugins.vim-illuminate;
```
GaetanLepage, line 74:
```suggestion
      enable = mkEnableOption "vim-illuminate";

      package = mkPackageOption "vim-illuminate" pkgs.vimPlugins.vim-illuminate;
```

Maybe just here, we can leave the real plugin's name.

GaetanLepage: Well even the github page of the plugin uses both `vim-illuminate` and `illuminate.vim` X)

anriha: Never really understood why someone uses .vim -nvim or .nvim makes at least some sense, but this is weird, especially when it is mainly neovim plugin. I think it still makes sense to change those two, so they match nixpkgs.

GaetanLepage (approved):

PR #473, opened 2023-07-05
plugins/colorschemes/oxocarbon: init

BattleCh1cken:
Added support for the oxocarbon-nvim colorscheme.

Homepage: https://github.com/nyoom-engineering/oxocarbon.nvim

I have tested and formatted my code.

### DIFF
--- a/plugins/colorschemes/oxocarbon.nix
+++ b/plugins/colorschemes/oxocarbon.nix
@@ -21,7 +21,7 @@ in {
     extraPlugins = [cfg.package];
 
     options = {
-      termguicolors = true;
+      termguicolors = mkDefault true;
     };
   };
 }
--- a/plugins/colorschemes/oxocarbon.nix
+++ b/plugins/colorschemes/oxocarbon.nix
@@ -12,7 +12,7 @@ in {
     colorschemes.oxocarbon = {
       enable = mkEnableOption "oxocarbon";
 
-      package = helpers.mkPackageOption "one" pkgs.vimPlugins.oxocarbon-nvim;
+      package = helpers.mkPackageOption "oxocarbon" pkgs.vimPlugins.oxocarbon-nvim;
     };
   };
 
--- a/plugins/colorschemes/oxocarbon.nix
+++ b/plugins/colorschemes/oxocarbon.nix
@@ -1,0 +1,27 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.colorschemes.oxocarbon;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options = {
+    colorschemes.oxocarbon = {
+      enable = mkEnableOption "oxocarbon";
+
+      package = helpers.mkPackageOption "one" pkgs.vimPlugins.oxocarbon-nvim;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    colorscheme = "oxocarbon";
+    extraPlugins = [cfg.package];
+
+    options = {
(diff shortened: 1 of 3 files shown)


plugins/colorschemes/oxocarbon.nix
```diff
@@ -0,0 +1,27 @@
...
+}:
+with lib; let
+  cfg = config.colorschemes.oxocarbon;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options = {
+    colorschemes.oxocarbon = {
+      enable = mkEnableOption "oxocarbon";
+
+      package = helpers.mkPackageOption "one" pkgs.vimPlugins.oxocarbon-nvim;
```
GaetanLepage, line 15:
```suggestion
      package = helpers.mkPackageOption "oxocarbon" pkgs.vimPlugins.oxocarbon-nvim;
```

plugins/colorschemes/oxocarbon.nix
```diff
@@ -0,0 +1,27 @@
...
+      package = helpers.mkPackageOption "oxocarbon" pkgs.vimPlugins.oxocarbon-nvim;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    colorscheme = "oxocarbon";
+    extraPlugins = [cfg.package];
+
+    options = {
+      termguicolors = true;
```
GaetanLepage, line 24:
```suggestion
      termguicolors = true;
```

What is the idea behind this ?
If we happen to need it, then it would be better to set it to `mkDefault true`.

BattleCh1cken: I'm ngl, I just copied the "one" color scheme and changed some things, as you've probably already guessed. I definetly agree that it would be better to use mkDefault, thanks for catching that.

GaetanLepage: Ok, can you please make the change then ?

GaetanLepage (changes requested):

plugins/colorschemes/oxocarbon.nix
BattleCh1cken: Sorry, I went to bed. I'll do that right now.

GaetanLepage (approved):

BattleCh1cken:
Looking over CONTRIBUTING.MD I'm realizing that there's also a requirement for writing unit tests, should I write some for this, or is that overkill for color schemes?

GaetanLepage:
Indeed, you may add a test for this. Have a look at the examples to get inspiration :)

tests/test-sources/plugins/colorschemes/oxocarbon.nix
```diff
@@ -0,0 +1,10 @@
+{
+  empty = {colorschemes.oxocarbon.enable = true;};
```
GaetanLepage, line 2:
```suggestion
  empty = {
    colorschemes.oxocarbon.enable = true;
  };
```

tests/test-sources/plugins/colorschemes/oxocarbon.nix
```diff
@@ -0,0 +1,10 @@
...
+
+  # All the upstream default options of oxocarbon
+  defaults = {
+    colorschemes.oxocarbon = {
+      enable = true;
+    };
+  };
+}
```
GaetanLepage, line 10:
```suggestion
```

As there are no options, no need to duplicate the `empty` test.

GaetanLepage (changes requested):

GaetanLepage (approved):

PR #489, opened 2023-07-20
fidget-nvim: init plugin + tests

ShyProton:
Adds plugin that shows LSP loading status:
https://github.com/j-hui/fidget.nvim

Also, since the plugin is in the process of being rewritten, the author's suggested to pin the config to the `legacy` branch for the time being, so I've got it set to pull from the latest commit from there.
I'm not entirely sure if that's the best way to go about it, so please let me know if there's a better way :)


traxys:
I'd say you should either package the plugin in `nixpkgs` (it's not very hard, you can ask @GaetanLepage for tips / to do it if youd want) or use `nvfetcher` to fetch it (there is a `nvfetcher.toml` file in `plugins` and a `plugins/plugin-defs.nix` for that). Using nvfecher you can point to a branch

ShyProton:
Ok, made it so it uses `nvfetcher` to pull from the `legacy` branch.
Also the plugin's already in nixpkgs [here](https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/applications/editors/vim/plugins/generated.nix#L3158), but points to the `main` branch instead of `legacy`.
Maybe I could try changing that at some point and if that goes through, I can change the reference back here to just use the package directly from nixpkgs whenever that happens.

GaetanLepage:
Hi !
Honestly I would just wait for the rewrite to be completed. Hence, we can add the relevant options in nixvim for good.
I don't really think that it is worth it to add a temporary solution.

Will the options change significantly ? If not, I am ok to merge the options right now and leave the individual users set `plugins.fidget.package` to whatever they prefer.

plugins/lsp/fidget-nvim.nix
```diff
@@ -0,0 +1,240 @@
...
+  lib,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.fidget-nvim;
+  helpers = import ../helpers.nix {inherit lib;};
+  plugin-defs = import ../plugin-defs.nix {inherit pkgs;};
+in {
+  options = {
+    plugins.fidget-nvim =
```
GaetanLepage, line 13:
```suggestion
    plugins.fidget =
```
I would have this plugin renamed `fidget`.

tests/test-sources/plugins/lsp/fidget-nvim.nix
```diff
@@ -0,0 +1,62 @@
...
+              task_name
+            )
+          end
+        '';
+      };
+
+      debug = {
+        logging = false;
+        strict = false;
+      };
```
GaetanLepage, line 59:
```suggestion
      text = {
        spinner = "pipe";
        done = "✔";
        commenced = "Started";
        completed = "Completed";
      };
      align = {
        bottom = true;
        right = true;
      };
      timer = {
        spinnerRate = 125;
        fidgetDecay = 2000;
        taskDecay = 1000;
      };

      window = {
        relative = "win";
        blend = 100;
        zindex = null;
        border = "none";
      };
      fmt = {
        leftpad = true;
        stackUpwards = true;
        maxWidth = 0;
        fidget.__raw = ''
          function(fidget_name, spinner)
            return string.format("%s %s", spinner, fidget_name)
          end
        '';
        task.__raw = ''
          function(task_name, message, percentage)
            return string.format(
              "%s%s [%s]",
              message,
              percentage and string.format(" (%s%%)", percentage) or "",
              task_name
            )
          end
        '';
      };
      debug = {
        logging = false;
        strict = false;
      };
```

plugins/lsp/fidget-nvim.nix
```diff
@@ -0,0 +1,240 @@
...
+      helpers.extraOptionsOptions
+      // {
+        enable = mkEnableOption "Enable fidget-nvim.";
+        package = helpers.mkPackageOption "fidget-nvim" plugin-defs.fidget-nvim;
```
GaetanLepage, line 17:
```suggestion
        enable = mkEnableOption "Enable fidget-nvim.";

        package = helpers.mkPackageOption "fidget-nvim" plugin-defs.fidget-nvim;
```

ShyProton:
Looks like there's going to be at least *some* amount of breaking changes, since their github says:
> NOTE: fidget.nvim will soon be completely rewritten. In the meantime, please pin your plugin config to the legacy tag to avoid breaking changes.

I'm not sure how much it's going to change though since it looks like the rewrite process has only begun somewhat recently.

GaetanLepage:
> I'm not sure how much it's going to change though since it looks like the rewrite process has only begun somewhat recently.

Ok I see. I made a [PR](https://github.com/NixOS/nixpkgs/pull/244652) to switch the upstream `pkgs.vimPlugins.fidget-nvim` to use the `legacy` branch.
If it gets merged, we can then add the options corresponding to the legacy fidget.
Once the rewrite will be completed, we will be able to target back the `master` branch on `nixpkgs` and to change the options in `nixvim` accordingly.

GaetanLepage:
@ShyProton I messed up this PR sorry. Can you re-open it please ?

ShyProton:
Yeah no problem, here's the new PR: #491

PR #491, opened 2023-07-26
fidget-nvim: init plugin + tests

ShyProton:
Adds plugin that shows LSP loading status:
https://github.com/j-hui/fidget.nvim

Currently pinned to the `legacy` branch.


GaetanLepage:
Actually, `nixpkgs` now tracks the `legacy` branch of `fidget-nvim`, so you do not need to add it to nvfetcher.
Can you remove it please ?

ShyProton:
Updated `fidget` so it uses the package from `nixpkgs` now.
Also the test fails here since the reference to `nixpkgs` in the lockfile points to one before the change to `fidget` using `legacy` upstream; I tested it using a newer reference to `nixpkgs` on my end and it works as intended :+1:

PR #505, opened 2023-08-08
Fix  incorrect command for barbar plugin goToN keymaps

meatcoder:
Currently, the `plugins.barbar.keymaps` option for `gotoN` (e.g. goto1, goto2, etc.)  generates a mapping to an incorrect editor command of the form `BufferGoTo N`. The correct command is `BufferGoto N`, with a lower case "T". This PR puts in the fix to generate the correct command name.


GaetanLepage (approved):
Nice catch thanks !

PR #510, opened 2023-08-10
helpers: add highlightType and defaultNullOpts.mkHighlight

GaetanLepage:
Many plugins expose options to define highlight groups.
In this case, we can now use the new `highlightType` which represents an `attrs` that will be later provided to `nvim_set_hl`.


traxys (changes requested):
Maybe add a test in `tests/lib-tests.nix`

traxys (approved):

PR #511, opened 2023-08-10
colorschemes/ayu: add plugin + test

GaetanLepage:


plugins/colorschemes/ayu.nix
```diff
@@ -0,0 +1,48 @@
...
+      // {
+        enable = mkEnableOption "ayu";
+
+        package = helpers.mkPackageOption "ayu" pkgs.vimPlugins.neovim-ayu;
+
+        mirage = helpers.defaultNullOpts.mkBool false ''
+          Set to `true` to use `mirage` variant instead of `dark` for dark background.
+        '';
+
+        overrides =
```
traxys, line 23: From the documentation:

Alternatively, overrides can be a function that returns a dictionary of the same format. You can use the function to override based on a dynamic condition, such as the value of 'background'.

traxys (changes requested):

plugins/colorschemes/ayu.nix
GaetanLepage: Done !

traxys (approved):

PR #526, opened 2023-08-17
plugins/dap: ensure extensions load after the main plugin

Wolbyte:
This pull request fixes issues with `nvim-dap` and extensions that add custom configurations/adapters to `nvim-dap`. All extensions now load after `nvim-dap` is initialized to prevent user-defined settings from overriding the configurations added by them.

### DIFF
--- a/plugins/dap/default.nix
+++ b/plugins/dap/default.nix
@@ -45,7 +45,14 @@ in
           dapBreakpointRejected = mkSignOption "R" "Sign to indicate breakpoints rejected by the debug adapter.";
         };
 
-        extensionConfigLua = helpers.mkNullOrOption types.lines "Extension configuration for dap. Don't use this directly!";
+        extensionConfigLua = mkOption {
+          type = types.lines;
+          description = ''
+            Extension configuration for dap. Don't use this directly !
+          '';
+          default = "";
+          internal = true;
+        };
       };
 
     config = let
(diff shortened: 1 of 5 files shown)


plugins/dap/default.nix
```diff
@@ -44,6 +44,8 @@ in
 
           dapBreakpointRejected = mkSignOption "R" "Sign to indicate breakpoints rejected by the debug adapter.";
         };
+
+        extensionConfigLua = helpers.mkNullOrOption types.lines "Extension configuration for dap. Don't use this directly!";
```
GaetanLepage, line 48:
```suggestion
        extensionConfigLua = mkOption {
          type = types.lines;
          description = ''
            Extension configuration for dap. Don't use this directly !
          '';
          default = "";
          internal = true;
        };
```

I think that this would fit better for this option.

plugins/dap/default.nix
```diff
@@ -86,6 +88,7 @@ in
             for sign_name, sign in pairs(__dap_signs) do
               vim.fn.sign_define(sign_name, sign)
             end
-          '');
+          '')
+          + (optionalString (cfg.extensionConfigLua != null) cfg.extensionConfigLua);
```
GaetanLepage, line 92:
```suggestion
          + cfg.extensionConfigLua);
```

GaetanLepage (changes requested):
Very nice solution !
Thanks for the fix :)

GaetanLepage (approved):
LGTM !
I prefer to wait for @traxys 's review before merging.

traxys (approved):

PR #533, opened 2023-08-21
fix(docs): typo + line wrap

anntnzrb:
1. `congig/default.nix` -> `config/default.nix`
2. Wrap at 80 chars


GaetanLepage (approved):
Nice catch, thanks !

PR #534, opened 2023-08-22
plugins/clangd-extensions: fix options following upstream changes + test

GaetanLepage:


plugins/languages/clangd-extensions.nix
```diff
@@ -20,157 +68,171 @@ in
...
+      plugins.lsp.postConfig = let
+        extraCaps =
+          if cfg.enableOffsetEncodingWorkaround
+          then ''
+            local __clangdCaps = vim.lsp.protocol.make_client_capabilities()
+            __clangdCaps.offsetEncoding = { "utf-16" }
+          ''
+          else "";
+      in ''
+        ${extraCaps}
```
traxys, line 234: This define the __clangdCaps in the postConfig, but it is used in the LSP config itself. I think you need to do that in the preConfig

traxys (changes requested):

plugins/languages/clangd-extensions.nix
GaetanLepage: Fixed

traxys (approved):

PR #537, opened 2023-08-23
plugins/nvim-cmp: Correctly handle multi mode mappings

### LINKED ISSUE #536: nvim-cmp: can no longer define mappings with a set/submodule
**Plugin affected**: nvim-cmp
**Nixpkgs channel**: unstable
**Home-manager version**: 23.11-pre

## Description
This bug seems to be added by #525.

_The `...` are placeholders here_

What still works: `nvim-cmp.mapping."..." = "...";`
What no longer works: `nvim-cmp.mapping."..." = {modes = [ ... ]; action = ...; };`

If this is simply a breaking change and it should be defined differently now it would be nice if you changed the example to reflect the change.
If this is a bug it seems to have passed the tests because it is simply not tested for.

```
error:

       … while evaluating derivation 'home-manager-generation'
         whose name attribute is located at /nix/store/1rp9mgpy83k4w11gi58hqfcqhk3nnahp-source/pkgs/stdenv/generic/make-derivation.nix:300:7

       … while evaluating attribute 'buildCommand' of derivation 'home-manager-generation'

         at /nix/store/1rp9mgpy83k4w11gi58hqfcqhk3nnahp-source/pkgs/build-support/trivial-builders/default.nix:87:14:

           86|       enableParallelBuilding = true;
           87|       inherit buildCommand name;
             |              ^
           88|       passAsFile = [ "buildCommand" ]

       error: attribute 'action' missing

       at /nix/store/v2dx1b6i3hvg324hskkxy78pblvrswc4-source/plugins/completion/nvim-cmp/default.nix:508:39:

          507|                         ("," + (helpers.toLuaObject modes));
          508|                     in "cmp.mapping(${mapping.action}${modesString})"
             |                                       ^
          509|                   )
```

## Config
This is a minimal config using the example for the mappings
```nix
{
  programs.nixvim = {
    enable = true;
    plugins = {
      nvim-cmp = {
        enable = true;
        mapping = {
          # This still works
          "<CR>" = "cmp.mapping.confirm({ select = false })";
          # This no longer works after the PR even though it is the example in the module
	  "<Tab>" = {
            modes = ["i" "s"];
            action = ''
              function(fallback)
                if cmp.visible() then
                  cmp.select_next_item()
                elseif luasnip.expandable() then
                  luasnip.expand()
                elseif luasnip.expand_or_jumpable() then
                  luasnip.expand_or_jump()
                elseif check_backspace() then
                  fallback()
                else
                  fallback()
                end
              end
            '';
          };
        };
      };
    };
  };
}
```

traxys:
Fixes #536

### DIFF
--- a/plugins/completion/nvim-cmp/default.nix
+++ b/plugins/completion/nvim-cmp/default.nix
@@ -505,7 +505,7 @@ in {
                           && ((length modes) >= 1)
                         )
                         ("," + (helpers.toLuaObject modes));
-                    in "cmp.mapping(${mapping.action}${modesString})"
+                    in "cmp.mapping(${action.action}${modesString})"
                   )
               )
               mapping);
--- a/tests/test-sources/plugins/completion/nvim-cmp.nix
+++ b/tests/test-sources/plugins/completion/nvim-cmp.nix
@@ -11,6 +11,34 @@
     };
   };
 
+  # Issue #536
+  mappings = {
+    plugins.nvim-cmp = {
+      enable = true;
+      mapping = {
+        "<CR>" = "cmp.mapping.confirm({ select = false })";
+        "<Tab>" = {
+          modes = ["i" "s"];
+          action = ''
+            function(fallback)
+              if cmp.visible() then
+                cmp.select_next_item()
+              elseif luasnip.expandable() then
+                luasnip.expand()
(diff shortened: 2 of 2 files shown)


GaetanLepage (approved):
Good catch ! Sorry

PR #545, opened 2023-08-24
docs: ignore option aliases

Wolbyte:
This pull request prevents option aliases from being generated and shown in the book.
It also removes any empty page from the book.


GaetanLepage (approved):
LGTM !

PR #551, opened 2023-08-27
plugins/neo-tree: fix incorrect option - filtered_items

adtya:
The `plugins.neo-tree.filesystem.filteredItems` is creating the respective option in the neovim config incorrectly. it should be `filtered_items` instead of `filteredItems`. This PR fixes that.


GaetanLepage (approved):
Nice catch ! Thanks for the fix :)

PR #556, opened 2023-08-28
lib: Use mkOptionType for `rawType`

traxys:
This allows the type to be rendered as `raw lua code` instead of `submodule` in the documentation.


GaetanLepage:
It does not impact how people would use it ?

traxys:
It should not, at least the checks still pass (this is not mkRaw)

GaetanLepage (approved):

PR #562, opened 2023-08-28
plugins/lsp: Extract rust-analyzer options from the source directly

traxys:
This directly reads the package.json file in pkgs.rust-analyzer.src in order to generate the list of options for rust-analyzer.

This avoids the need for a generator script, and makes sure that it is always in sync.


GaetanLepage (approved):
LGTM although I trust you on the very details

PR #571, opened 2023-09-04
plugins/nvim-lsp: add digestif language server

ptonner:


plugins/lsp/language-servers/default.nix
```diff
@@ -147,6 +147,11 @@ with lib; let
       description = "Enable denols, for Deno";
       package = pkgs.deno;
     }
+    {
+      name = "digestif";
+      description = "Enable digestif, for LaTeX";
+      package = pkgs.lua54Packages.digestif;
```
traxys, line 153: Could you add a coment saying that `luaPackages.digestif` is broken ?

traxys (changes requested):

plugins/lsp/language-servers/default.nix
ptonner: it's possible that I just failed to use the right package, I used `lua54` because that was the most recent one I found here: https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=digestif

is there somewhere else I should have looked?

traxys: You can use `luaPackages.digestif` but when I tried that I got:

```
       error: Package ‘lua5.2-digestif-dev-1’ in /nix/store/69snz0369qhm5k2wh18snsj7m2a3af6q-source/pkgs/development/lua-modules/generated-packages.nix:429 is marked as broken, refusing to evaluate.

       a) To temporarily allow broken packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_BROKEN=1

        Note: For `nix shell`, `nix build`, `nix develop` or any other Nix 2.4+
        (Flake) command, `--impure` must be passed in order to read this
        environment variable.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowBroken = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowBroken = true; }
       to ~/.config/nixpkgs/config.nix.
```

(`nix shell nixpkgs#luaPackages.digestif`)
I think this does not appear in the nixos search, but when possible it's better to ommit the version in luaPackages, phpPackages, pythonPackages, .... as it is better to let nixpkgs choose what is the "official" version

ptonner: okay cool, thanks! maybe it doesn't show up in search if it's broken

I added a quick comment about it below

traxys (approved):

Issue #573, opened 2023-09-05
plugins.lsp.nixd: implicit dependency on nixpkgs-fmt for formatting


jhvst:
**Plugin affected**: `plugins.lsp.nixd`
**Nixpkgs channel**: unstable
**Home-manager version**: unstable

## Description
When LSP formatting is enabled, and so is `nixd`, formatting does not work unless `nixpkgs-fmt` is added as extraPackage. Looking at the [nixd source code](https://github.com/search?q=repo%3Anix-community%2Fnixd%20nixpkgs-fmt&type=code) it has an implicit dependency on `nixpkgs-fmt`.

`nixpkgs-fmt` should be added as an additional package to install if `nixd` is enabled, and formatting is set as default or as `nixpkgs-fmt`.

## Config
N/A

GaetanLepage:
After having taken a look, it seems that `nixpkgs-fmt` is not really a _dependency_ of `nixd`.
`nixd` can run `nixpkgs-fmt`.
As such, it does not really make sense to systematically add it as a dependency.
If the user needs it, he should indeed add it to `extraPackages`.

jhvst:
Fair enough. Could you comment on LSPs that do the same? Like bash-ls integrates with `shellcheck`, it is not a required dependency, but one could assume that if the user wants to enable bash-ls, they most likely also want `shellcheck` to be enabled as well. I wonder what is your stance in these cases?

GaetanLepage:
After a second thought, I have proposed #577 which will automatically add the dependency if you enable formatting in the `nixd` settings (see PR for more details).
I think that this is a nice middle-ground.

PR #577, opened 2023-09-08
plugins/nixd: automatically add `nixpkgs-fmt` if needed

GaetanLepage:
If the user explicitly sets `plugins.lsp.servers.nixd.settings.formatting.command = "nixpkgs-fmt"`, then `pkgs.nixpkgs-fmt` will be automatically added to `extraPackages`.

Context: #573

### DIFF
--- a/plugins/lsp/language-servers/nixd.nix
+++ b/plugins/lsp/language-servers/nixd.nix
@@ -1,0 +1,64 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  helpers = import ../../helpers.nix {inherit lib;};
+  cfg = config.plugins.lsp.servers.nixd;
+in {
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/nixd.nix
```diff
@@ -0,0 +1,64 @@
...
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  helpers = import ../../helpers.nix {inherit lib;};
+  cfg = config.plugins.lsp.servers.nixd;
+in {
+  # Options: https://github.com/sveltejs/language-tools/tree/master/packages/language-server#settings
```
traxys, line 11: why sveltejs ?

traxys (changes requested):

traxys (approved):

PR #589, opened 2023-09-15
feat(options): add `defaultEditor` option

mrtnvgr:
This PR adds `defaultEditor` option from home-manager's `neovim` module

### DIFF
--- a/wrappers/modules/output.nix
+++ b/wrappers/modules/output.nix
@@ -94,7 +94,7 @@ in {
     config.extraPlugins;
 
     neovimConfig = pkgs.neovimUtils.makeNeovimConfig ({
-        inherit (config) viAlias vimAlias defaultEditor;
+        inherit (config) viAlias vimAlias;
         # inherit customRC;
         plugins = normalizedPlugins;
       }
--- a/wrappers/modules/output.nix
+++ b/wrappers/modules/output.nix
@@ -31,7 +31,7 @@ in {
       type = types.bool;
       default = false;
       description = ''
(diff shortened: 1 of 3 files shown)


traxys:
I don't think this should be added in this place, as it is a HM specific option. I think this has more it's place in the nixvim/wrappers/hm.nix file

wrappers/modules/output.nix
```diff
@@ -43,6 +43,14 @@ in {
       '';
     };
 
+    defaultEditor = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Configure <command>nvim</command> as the default editor using the EDITOR environment variable
```
GaetanLepage, line 50: Shouldn't we simply use `nvim` here (because the doc is generated from markdown) ?

traxys:
I feel like the option definition itself should be in the hm file (and maybe nixos) because it makes no sense for the standalone module. Something like programs.nixvim.defaultEditor should work ?

mrtnvgr:
> I feel like the option definition itself should be in the hm file (and maybe nixos) because it makes no sense for the standalone module. Something like programs.nixvim.defaultEditor should work ?

I don't quite understand how can I do that

traxys:
```nix
    programs.nixvim = mkOption {
      default = {};
      type = types.submodule ([
          {
            options = {
              enable = mkEnableOption "nixvim";
              defaultEditor = mkEnableOption "Set nixvim as the default option";
            };
          }
        ]
        ++ shared.topLevelModules);
    };
```

Something like that should work (not tested)

traxys (approved):

PR #590, opened 2023-09-16
plugins/gitblame: init + test

adtya:
PR to add support for the [git-blame.nvim](https://github.com/f-person/git-blame.nvim) plugin

### DIFF
--- a/plugins/git/gitblame.nix
+++ b/plugins/git/gitblame.nix
@@ -16,7 +16,7 @@ in {
 
       messageTemplate = mkOption {
         type = types.str;
-        default = "<author> • <date> • <summary>";
+        default = "  <author> • <date> • <summary>";
         description = "The template for the blame message that will be shown.";
       };
 
@@ -28,7 +28,7 @@ in {
 
       messageWhenNotCommitted = mkOption {
         type = types.str;
-        default = "Not Committed Yet";
+        default = "  Not Committed Yet";
         description = "The blame message that will be shown when the current modification hasn't been committed yet.";
       };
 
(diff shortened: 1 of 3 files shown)


plugins/git/gitblame.nix
```diff
@@ -0,0 +1,99 @@
...
+  cfg = config.plugins.gitblame;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options = {
+    plugins.gitblame = {
+      enable = mkEnableOption "gitblame";
+
+      package = helpers.mkPackageOption "gitblame" pkgs.vimPlugins.git-blame-nvim;
+
+      messageTemplate = mkOption {
```
pta2002, line 17: Thanks for the pr! Just a note though, don't have these as the default! Use the helper functions in helpers.nix in order to define options.

Defining them like this leads to quickly filling up the generated config, since we are redefining every default option.

pta2002 (changes requested):

plugins/git/gitblame.nix
adtya: Sure, I'll update those. I used mkOption because I was looking at how other plugins are written 😅

adtya: done.

though for one option, `display_virtual_text`, I had to keep `mkOption`.
The plugin expects 1, 0 or null with default being 1 (if null).

when using `helpers.defaultNullOpts.mkNullable (types.nullOr types.bool) true`, I am not able to convert it to 1 or 0 in the config

traxys: You need to do it manually with something like 
```
display_virtual_text = let disp = cfg.displayVirtualText in 
   helpers.ifNonNull' cfg.displayVirtualText (if disp then 1 else 0);
```

adtya: That seems to have worked :+1:

adtya: resolving this as all the options are now using helpers

traxys (approved):

pta2002 (approved):
that's more like it!

PR #595, opened 2023-09-19
plugins/notify: add missing options + test

### LINKED ISSUE #584: [Notify] max size, levels, and top_down options missing
**Plugin affected**: notify
**Nixpkgs channel**: unstable
**Home-manager version**: n/a

## Description

Options for configuring notify-nvim's `max_width`, `max_height`, `levels`, and `top_down` settings are currently missing. We should add these to allow for customization by users.

## Config

n/a

GaetanLepage:
Fixes #584

### DIFF
--- a/plugins/utils/notify.nix
+++ b/plugins/utils/notify.nix
@@ -7,67 +7,150 @@
 with lib; let
   cfg = config.plugins.notify;
   helpers = import ../helpers.nix {inherit lib;};
-  icon = mkOption {
-    type = types.nullOr types.str;
-    default = null;
-  };
 in {
-  options.plugins.notify = {
-    enable = mkEnableOption "notify";
+  options.plugins.notify =
+    helpers.extraOptionsOptions
+    // {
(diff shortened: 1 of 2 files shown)


plugins/utils/notify.nix
```diff
@@ -7,67 +7,141 @@
...
+
+      maxWidth = helpers.mkNullOrOption (with types; either ints.unsigned helpers.rawType) ''
+        Max number of columns for messages.
+      '';
+
+      maxHeight = helpers.mkNullOrOption (with types; either ints.unsigned helpers.rawType) ''
+        Max number of lines for a message.
+      '';
+
+      stages =
```
traxys, line 41: It seems that the stages should either be a string or an array of functions, do we leave the array of functions to do in the raw type or do we use listOf rawType ?

GaetanLepage: Good point, I chose a `listOf str` that are `mkRaw`ed afterwards.

traxys (approved):

PR #599, opened 2023-09-20
plugins/todo-comments: Allow to pass keymap options

### LINKED ISSUE #598: desc attribute in action.key for todo-comments
Would it be possible to supply an attribute set like `{ key = "key; desc = "Show all TODOs in telescope overlayer"` in the  [todo-comments keymap](https://github.com/nix-community/nixvim/blob/ada99e57635b9ec60fab30d361927820fa0f890d/plugins/utils/todo-comments.nix#L196)?

Or matybe add a new parameter `desc` to make the keymapping compatible with **which-key**?

I'm not sure if such a cross-plugin features are something that should be introduced or considered in **nixvim** or if this is something that should go into the custom keymap of every user.

traxys:
This re-uses the keymap options defined in the keymaps module to allow to define the options of todo-comments key mappings.

Fixes #598

### DIFF
--- a/plugins/utils/todo-comments.nix
+++ b/plugins/utils/todo-comments.nix
@@ -15,6 +15,13 @@ with lib; let
     todoTelescope = "TodoTelescope";
   };
 in {
+  imports = [
+    (
+      mkRemovedOptionModule
+      ["plugins" "todo-comments" "keymapsSilent"]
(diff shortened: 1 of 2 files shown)


GaetanLepage:
This can now be rebased on `main`.

GaetanLepage (approved):

PR #612, opened 2023-09-29
plugins/lsp/hls: add missing --lsp cmd flag

NyCodeGHG:
hls just crashed everytime i tried to use it, because it's missing the `--lsp` flag to work correctly.


GaetanLepage:
Good catch !
Thank you for your contribution.

PR #614, opened 2023-09-30
github-worflow: delete useless files on GitHub runner to free up space

GaetanLepage:
`github` runners fails by running out of space.
Those few lines remove unnecessary things from the container.


traxys:
Maybe use something like this:
https://github.com/marketplace/actions/maximize-build-disk-space
It does what you did, but should be easier to maintain going forwards

traxys (approved):

PR #617, opened 2023-10-02
modules/keymaps: factor out helper functions for use in plugins

GaetanLepage:
Move some useful declarations/functions from `modules/keymaps.nix` to `lib/keymap-helpers.nix`.
Like so, a plugin could use them to define some keymap-related options.

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -5,142 +5,6 @@
 }:
 with lib; let
   helpers = import ../lib/helpers.nix {inherit lib;};
-
-  # These are the configuration options that change the behavior of each mapping.
-  mapConfigOptions = {
-    silent =
-      helpers.defaultNullOpts.mkBool false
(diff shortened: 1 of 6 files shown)


modules/keymaps.nix
```diff
@@ -157,21 +22,26 @@ in {
                 either
                 str
                 (
-                  mkMapOptionSubmodule
+                  keymap-helpers.mkMapOptionSubmodule
```
traxys, line 25: Why not put in helpers.keymaps ? like this it avoids to add another kind of special import, and you don't have to add the necessary stuff to make it import automatically in home-assistant/nixos modules & standalone configurations.

traxys (changes requested):

modules/keymaps.nix
GaetanLepage: Very good idea ! Done :)

PR #621, opened 2023-10-03
plugins/harpoon: fix keymaps + test

GaetanLepage:
Harpoon has never been tested. Hence, a few bugs were found and fixed.

### DIFF
--- a/plugins/utils/harpoon.nix
+++ b/plugins/utils/harpoon.nix
@@ -10,11 +10,11 @@ with lib; let
 
   projectConfigModule = types.submodule {
     options = {
-      termCommands = helpers.mkNullOrOption (types.listOf types.str) ''
+      termCommands = helpers.mkNullOrOption (with types; listOf str) ''
         List of predefined terminal commands for this project.
       '';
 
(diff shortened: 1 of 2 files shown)


plugins/utils/harpoon.nix
```diff
@@ -155,27 +158,40 @@ in {
...
 
-        menu = {
-          inherit (cfg.menu) width height;
-          borderchars = cfg.menu.borderChars;
+        menu = with menu; {
+          inherit
+            width
+            height
+            ;
+          borderchars = borderChars;
```
traxys, line 194: I'd prefer to avoid having nested with foo; with bar; It makes it harder to check where a values comes from, and could easily introduce bugs if we add an option with the same name in `cfg` and in `cfg.menu`

traxys (changes requested):

plugins/utils/harpoon.nix
GaetanLepage: You are right (although the innermost scope should prevail... I guess).
Anyway, I removed it.

traxys (approved):

PR #622, opened 2023-10-03
plugins/barbecue: init + tests

hmajid2301:
PR to add support for the [barbecue-nvim](https://github.com/utilyre/barbecue.nvim plugin)

### DIFF
--- a/plugins/bufferlines/barbecue.nix
+++ b/plugins/bufferlines/barbecue.nix
@@ -164,7 +164,8 @@ in {
         include_buftypes = includeBuftypes;
         exclude_filetypes = excludeFiletypes;
         modifiers = {
-          inherit (modifiers)
+          inherit
+            (modifiers)
             dirname
             basename
             ;
@@ -172,20 +173,21 @@ in {
         show_dirname = showDirname;
         show_basename = showBasename;
         show_modified = showModified;
-        modified = helpers.ifNonNull' modified; (helpers.mkRaw modified);
+        modified = helpers.ifNonNull' modified (helpers.mkRaw modified);
         show_navic = showNavic;
         lead_custom_section = helpers.ifNonNull' leadCustomSection (helpers.mkRaw leadCustomSection);
         custom_section = helpers.ifNonNull' customSection (helpers.mkRaw customSection);
         inherit theme;
         context_follow_icon_color = contextFollowIconColor;
         symbols = {
-          inherit (symbols);
+          inherit
+            (symbols)
             modified
             ellipsis
             separator
             ;
         };
-        inherit kinds;       
+        inherit kinds;
       }
       // cfg.extraOptions;
   in
--- a/plugins/bufferlines/barbecue.nix
+++ b/plugins/bufferlines/barbecue.nix
@@ -28,7 +28,7 @@ in {
         Buftypes to enable winbar in.
       '';
 
-      excludeFiletypes = mkListStr ''[ "netrw" "toggleterm" ]'' ''
+      excludeFiletypes = mkListStr ''["netrw" "toggleterm"]'' ''
         Filetypes not to enable winbar in.
       '';
 
--- a/plugins/bufferlines/barbecue.nix
+++ b/plugins/bufferlines/barbecue.nix
@@ -5,7 +5,7 @@
   ...
 }:
 with lib; let
-  cfg = config.plugins.navic;
+  cfg = config.plugins.barbecue;
   helpers = import ../helpers.nix {inherit lib;};
   mkListStr = helpers.defaultNullOpts.mkNullable (types.listOf types.str);
 in {
--- a/plugins/bufferlines/barbecue.nix
+++ b/plugins/bufferlines/barbecue.nix
@@ -24,7 +24,7 @@ in {
         Whether to create winbar updater autocmd.
       '';
 
-      includeBuftypes = mkListStr ''[ "" ]'' ''
+      includeBuftypes = mkListStr ''[""]'' ''
         Buftypes to enable winbar in.
       '';
 
--- a/plugins/bufferlines/barbecue.nix
+++ b/plugins/bufferlines/barbecue.nix
@@ -68,11 +68,17 @@ in {
         Whether to show/use navic in the winbar.
       '';
 
-      leadCustomSection = helpers.defaultNullOpts.mkStr ''function() return " " end,'' ''
-            Get leading custom section contents.
-
-        NOTE: This function shouldn't do any expensive actions as it is run on each render.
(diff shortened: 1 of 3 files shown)


plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,172 @@
...
+
+      showBasename = helpers.defaultNullOpts.mkBool true ''
+        Whether to display file name.
+      '';
+
+      showModified = helpers.defaultNullOpts.mkBool false ''
+        Whether to replace file icon with the modified symbol when buffer is modified.
+      '';
+
+      modified = helpers.defaultNullOpts.mkStr ''function(bufnr) return vim.bo[bufnr].modified end,'' ''
```
hmajid2301, line 61: Not sure the best type for this lua function, I saw some other packages doing this!

hmajid2301: I guess this is also whats throwing an error in CI (with the `lead_custom_section`. Thanks.

GaetanLepage: ->Question: When to choose between an `types.str` and a `helpers.rawType` option for lua code ?

- If the upstream option expects only lua code, we use `types.str`. There is no ambiguity and the user doesn't have to bother with providing a `rawType` value.
In this case, we do the conversion ourselves in the `config` section of the plugin: `foo = helpers.ifNonNull' foo (helpers.mkRaw foo);`
- If the upstream option accept both strings and lua code (or when there is some kind of ambiguity), we rely on `helpers.rawType`. In this case, there is no need to convert the provided value.

plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,172 @@
...
+
+      attachNavic = helpers.defaultNullOpts.mkBool true ''
+        Whether to attach navic to language servers automatically.
+      '';
+
+      createAutocmd = helpers.defaultNullOpts.mkBool true ''
+        Whether to create winbar updater autocmd.
+      '';
+
+      includeBuftypes = mkListStr ''[ "" ]'' ''
```
GaetanLepage, line 27:
```suggestion
      includeBuftypes = mkListStr ''[""]'' ''
```

plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,172 @@
...
+        Buftypes to enable winbar in.
+      '';
+
+      excludeFiletypes = mkListStr ''[ "netrw" "toggleterm" ]'' ''
```
GaetanLepage, line 31:
```suggestion
      excludeFiletypes = mkListStr ''["netrw" "toggleterm"]'' ''
```

plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,172 @@
...
+
+      showNavic = helpers.defaultNullOpts.mkBool true ''
+        Whether to show/use navic in the winbar.
+      '';
+
+      leadCustomSection = helpers.defaultNullOpts.mkStr ''function() return " " end,'' ''
+            Get leading custom section contents.
+
+        NOTE: This function shouldn't do any expensive actions as it is run on each render.
+      '';
```
GaetanLepage, line 75:
```suggestion
      leadCustomSection =
        helpers.defaultNullOpts.mkStr
        ''
          function()
            return " "
          end
        ''
        ''
          Get leading custom section contents.
          NOTE: This function shouldn't do any expensive actions as it is run on each render.
        '';
```
Prefer this formatting for those options.

tests/test-sources/plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,61 @@
+{
+  empty = {
+    plugins.barbecue.enable = true;
+  };
+
+  defaults = {
+    plugins.barbecue = {
+      enable = true;
+      attachNavic = true;
```
GaetanLepage, line 9:
```suggestion
      enable = true;

      attachNavic = true;
```

GaetanLepage (commented):
Very good looking !
Here are some first minor remarks.

plugins/bufferlines/barbecue.nix
hmajid2301: Okay just to replay it back to you; we carry on using the str type. I just need to update the config section with this helpers function (?) you defined.

plugins/bufferlines/barbecue.nix
hmajid2301: sure!

plugins/bufferlines/barbecue.nix
hmajid2301: yeh looks much better, will update the others.

plugins/bufferlines/barbecue.nix
```diff
@@ -0,0 +1,190 @@
...
+        exclude_filetypes = excludeFiletypes;
+        show_dirname = showDirname;
+        show_basename = showBasename;
+        show_modified = showModified;
+        show_navic = showNavic;
+        lead_custom_section = leadCustomSection;
+        custom_section = customSection;
+        context_follow_icon_color = contextFollowIconColor;
+      }
+      // cfg.extraOptions;
```
GaetanLepage, line 181:
```suggestion
        attach_navic = attachNavic;
        create_autocmd = createAutocmd;
        include_buftypes = includeBuftypes;
        exclude_filetypes = excludeFiletypes;
        modifiers = {
          inherit (modifiers)
            dirname
            basename
            ;
        };
        show_dirname = showDirname;
        show_basename = showBasename;
        show_modified = showModified;
        modified = helpers.ifNonNull' modified; (helpers.mkRaw modified);
        show_navic = showNavic;
        lead_custom_section = helpers.ifNonNull' leadCustomSection (helpers.mkRaw leadCustomSection);
        custom_section = helpers.ifNonNull' customSection (helpers.mkRaw customSection);
        inherit theme;
        context_follow_icon_color = contextFollowIconColor;
        symbols = {
          inherit (symbols);
            modified
            ellipsis
            separator
            ;
        };
        inherit kinds;       
      }
      // cfg.extraOptions;
```

GaetanLepage: 1. You had forgotten to convert (`helpers.mkRaw`) the values of the options that expect lua code
2. Keep the same order in `setupOptions` as the one used when declaring options.

hmajid2301: 1. yeh I was waiting to confirm above if 1 was the right approach!
2. ahh yes the ordering good shout

GaetanLepage (approved):
LGTM

PR #623, opened 2023-10-05
nvim-tree: removes deprecated 'hideRootFolder' option

ShyProton:
nvim-tree has [deprecated the `hide_root_folder` option](https://github.com/nvim-tree/nvim-tree.lua/pull/2431); it cannot be set anymore.

Also adds missing option to specify a boolean for `rootFolderLabel`. Setting it to `false` will do the same thing as what `hideRootFolder` used to do.

### DIFF
--- a/plugins/filetrees/nvim-tree.nix
+++ b/plugins/filetrees/nvim-tree.nix
@@ -365,10 +365,6 @@ in {
           Increase if you experience performance issues around screen refresh.
         '';
 
-        hideRootFolder = helpers.defaultNullOpts.mkBool false ''
-          Hide the path of the current working directory on top of the tree.
-        '';
-
         width =
           helpers.defaultNullOpts.mkNullable
           (types.oneOf [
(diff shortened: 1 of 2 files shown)


plugins/filetrees/nvim-tree.nix
```diff
@@ -365,10 +365,6 @@ in {
           Increase if you experience performance issues around screen refresh.
         '';
 
-        hideRootFolder = helpers.defaultNullOpts.mkBool false ''
-          Hide the path of the current working directory on top of the tree.
-        '';
-
```
GaetanLepage, line 371: Please, mark the option as deprecated using `mkRemovedModule`.
Example here: https://github.com/nix-community/nixvim/blob/73f21c063dac16340c60d99a143dd3c7c07a1e70/plugins/filetrees/neo-tree.nix#L13-L24

GaetanLepage (changes requested):

GaetanLepage (approved):
LGTM !
Thank you for your contribution :)

PR #624, opened 2023-10-05
telescope.harpoon: init 

hmajid2301:
Added telescope extension: [harpoon](https://github.com/ThePrimeagen/harpoon#telescope-support)


GaetanLepage (commented):
Shouldn't we check/require that harpoon (the plugin) is enabled in this case ?

hmajid2301:
Yes, whats the best way to do that ? `config.plugins.utils.harpoon.enable` ?

traxys:
I feel we should add this in the harpoon plugin directly, with something like `enableTelescope = mkEnableOption "telescope intergration for harpoon"`.

Though we should raise a warning if telescope is not enabled 
(example for treesitter:
```nix
      warnings = mkIf (!config.plugins.treesitter.enable) [
        "Nixvim: ts-context-commentstring needs treesitter to function as intended"
      ];
```)

**Comment by hmajid2301** [comment]
So I guess we want something like:

```nix
  mkIf cfg.enableTelescope {
  warnings = mkIf (!config.plugins.telescope.enable) [
    "Nixvim: The harpoon telescope integration needs telescope to function as intended"
  ];

  extraConfigLua = ''
   require("telescope").load_extension('harpoon')
  '';
};
```

Where do I put this ? I dont think I can put in the harpoon block i.e. nested `mkIf`, shall I do something like this with mkMerge ? 

```
config = mkIf cfg.enable (mkMerge [
   (mkIf condition1 {
      ...
   })
   (mkIf condition2 {
      ...
   })
   ....
]);
```

traxys:
Hm not sure, you can try to do nested `mkIf`. For warnings you can do a `lib.optional` I think?

hmajid2301:
Trying to do:

```
 mkIf cfg.enable {
      mkIf cfg.enableTelescope {
      warnings = mkIf (!config.plugins.telescope.enable) [
        "Nixvim: The harpoon telescope integration needs telescope to function as intended"
      ];

      extraConfigLua = ''
        require("telescope").load_extension('harpoon')
      '';
    };
```

gives me a syntax error. `syntax error, unexpected ID, expecting '.' or '=' `

traxys:
Yeah you want be able to merge them like this.

You can do the following: 
```nix
mkIf cfg.enable  {
      warnings = mkIf (!config.plugins.telescope.enable && cfg.enableTelescope) [
        "Nixvim: The harpoon telescope integration needs telescope to function as intended"
      ];
      
      extraConfigLua = let
        telescopeCfg = ''require(....).load....'';
      in ''
          //curent harpoon content of extraConfigLua
          ${if cfg.enableTelescope then telescopeCfg else ""}
      '';
}
```

In the harpoon file.

Maybe more of an assertion than a warning, as the config will error.
```nix
assertions = [
{
   assertion = /* boolean */;;
  message = '' .... '';
}
];
```

hmajid2301:
I believe I have made the changes!

plugins/utils/harpoon.nix
```diff
@@ -198,10 +200,21 @@ in {
       // cfg.extraOptions;
   in
     mkIf cfg.enable {
+      warnings = mkIf (!config.plugins.telescope.enable && cfg.enableTelescope) [
+        "Nixvim: The harpoon telescope integration needs telescope to function as intended"
+      ];
```
traxys, line 205: Please use an assertion instead of a warning, see my previous comment about those

traxys (changes requested):

plugins/utils/harpoon.nix
hmajid2301: apologies i forgot to make that change!

tests/test-sources/plugins/utils/harpoon.nix
```diff
@@ -15,6 +15,7 @@
     plugins.harpoon = {
       enable = true;
 
+      enableTelescope = true;
```
traxys, line 18: You are going to need to enable telescope in that test, else the assertion will trigger :P

traxys (changes requested):

traxys:
I think you need that operator:
https://nixos.org/manual/nix/stable/language/operators#logical-implication

tests/test-sources/plugins/utils/harpoon.nix
hmajid2301: yeh i disabled it but maybe it makes more sense to test with it enabled

traxys:
> > I think you need that operator:
> > https://nixos.org/manual/nix/stable/language/operators#logical-implication
> 
> Do you mean in the assertion bit ? `!config.plugins.telescope.enable -> cfg.enableTelescope`

Yes, but I think you need to revert it. cfg.enableTelescope implies that config.plugins.telescope.enable

hmajid2301:
> > > I think you need that operator:
> > > https://nixos.org/manual/nix/stable/language/operators#logical-implication
> > 
> > 
> > Do you mean in the assertion bit ? `!config.plugins.telescope.enable -> cfg.enableTelescope`
> 
> Yes, but I think you need to revert it. cfg.enableTelescope implies that config.plugins.telescope.enable

Ahh ofc yes, basic logic :facepalm: !

hmajid2301:
> Can you also write a test where enableTelescope is false ? That code could be error prone in the future.

Yes ofc!

traxys (approved):

PR #626, opened 2023-10-06
plugins/better-escape: init + tests

hmajid2301:
PR to add support for the [better-escape-nvim](https://github.com/max397574/better-escape.nvim) plugin

I have a quick questions about the PR to fix before its ready.

### DIFF
--- a/plugins/utils/better-escape.nix
+++ b/plugins/utils/better-escape.nix
@@ -11,9 +11,9 @@ in {
   options.plugins.better-escape =
     helpers.extraOptionsOptions
     // {
-      enable = mkEnableOption "better-escape";
+      enable = mkEnableOption "better-escape.nvim";
 
-      package = helpers.mkPackageOption "better-escape" pkgs.vimPlugins.better-escape-nvim;
+      package = helpers.mkPackageOption "better-escape.nvim" pkgs.vimPlugins.better-escape-nvim;
 
       mapping = helpers.mkNullOrOption (with types; listOf str) ''
         List of mappings to use to enter escape mode.
--- a/plugins/utils/better-escape.nix
+++ b/plugins/utils/better-escape.nix
@@ -19,9 +19,13 @@ in {
         List of mappings to use to enter escape mode.
       '';
 
-      timeout = helpers.defaultNullOpts.mkInt 100 ''
-        The time in which the keys must be hit in ms. Use option timeoutlen by default.
-      '';
+      timeout =
+        helpers.defaultNullOpts.mkNullable
+        (with types;
+            either int helpers.rawType)
+        "vim.o.timeoutlen" ''
+          The time in which the keys must be hit in ms. Use option timeoutlen by default.
+        '';
 
       clearEmptyLines = helpers.defaultNullOpts.mkBool false ''
         Clear line after escaping if there is only whitespace.
@@ -50,7 +54,11 @@ in {
   config = let
     setupOptions = with cfg;
       {
-        inherit mapping timeout keys;
+        inherit mapping keys;
+        timeout =
+          if isInt timeout
+          then helpers.mkRaw "timeout"
+          else timeout;
         clear_empty_lines = clearEmptyLines;
       }
       // cfg.extraOptions;
--- a/plugins/utils/better-escape.nix
+++ b/plugins/utils/better-escape.nix
@@ -56,7 +56,7 @@ in {
       {
         inherit mapping keys;
         timeout =
-          if isInt timeout
+          if isString timeout
           then helpers.mkRaw "timeout"
           else timeout;
         clear_empty_lines = clearEmptyLines;
--- a/plugins/utils/better-escape.nix
+++ b/plugins/utils/better-escape.nix
@@ -21,8 +21,9 @@ in {
 
       timeout =
         helpers.defaultNullOpts.mkNullable
-        (with types;
-          either ints.unsigned str
+        (
+          with types;
+            either ints.unsigned str
         )
         "vim.o.timeoutlen"
         ''
@@ -36,8 +37,9 @@ in {
 
       keys =
         helpers.defaultNullOpts.mkNullable
-        (with types;
-          either str helpers.rawType
+        (
+          with types;
+            either str helpers.rawType
         )
         "<ESC>"
         ''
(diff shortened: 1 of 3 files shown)


plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,65 @@
...
+
+      timeout = helpers.defaultNullOpts.mkInt 100 ''
+        The time in which the keys must be hit in ms. Use option timeoutlen by default.
+      '';
+
+      clearEmptyLines = helpers.defaultNullOpts.mkBool false ''
+        Clear line after escaping if there is only whitespace.
+      '';
+
+      keys =
```
hmajid2301, line 30: I believe this could also be a list of keys as well, how should I type this ? I tried `listof str`

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,65 @@
...
+    // {
+      enable = mkEnableOption "better-escape";
+
+      package = helpers.mkPackageOption "better-escape" pkgs.vimPlugins.better-escape-nvim;
+
+      mapping = helpers.mkNullOrOption (with types; listOf str) ''
+        List of mappings to use to enter escape mode.
+      '';
+
+      timeout = helpers.defaultNullOpts.mkInt 100 ''
```
hmajid2301, line 22: by default this uses a value from `vim.o.timeoutlen`, I set this to a 100 just to create the PR. but is there a way to fetch this ? I saw something getting `vim.g.` values

hmajid2301: I think I fixed this

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,73 @@
...
+          	};
+        '';
+    };
+
+  config = let
+    setupOptions = with cfg;
+      {
+        inherit mapping keys;
+        timeout =
+          if isInt timeout
```
GaetanLepage, line 59:
```suggestion
          if isString timeout
```

tests/test-sources/plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,15 @@
+{
+  empty = {
+    plugins.better-escape.enable = true;
+  };
+
+  example = {
+    plugins.better-escape = {
+      enable = true;
+      mapping = ["jj" "jk"];
```
GaetanLepage, line 9:
```suggestion
      enable = true;

      mapping = ["jj" "jk"];
```

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,73 @@
...
+with lib; let
+  cfg = config.plugins.better-escape;
+  helpers = import ../helpers.nix {inherit lib;};
+in {
+  options.plugins.better-escape =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "better-escape";
+
+      package = helpers.mkPackageOption "better-escape" pkgs.vimPlugins.better-escape-nvim;
```
GaetanLepage, line 16:
```suggestion
      enable = mkEnableOption "better-escape.nvim";

      package = helpers.mkPackageOption "better-escape.nvim" pkgs.vimPlugins.better-escape-nvim;
```

Here, we put the name of the real plugin.

plugins/utils/better-escape.nix
hmajid2301: im confused is this is a string shouldnt it be treated as "lua" ?

GaetanLepage: If it is a string, then it should be converted to Lua.
Otherwise (`null` or int), it should be passed as is.

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,73 @@
...
+
+          keys = {
+          	_raw = \'\'
+          		function()
+          			return vim.api.nvim_win_get_cursor(0)[2] > 1 and '<esc>l' or '<esc>'
+          		 end,.
+          		 \'\'
+          	};
+        '';
+    };
```
GaetanLepage, line 52: No, it cannot be a list.
The option declaration is fine.

Minor formatting changes:
```suggestion
      keys =
        helpers.defaultNullOpts.mkNullable
        (with types;
          either str helpers.rawType
        )
        "<ESC>"
        ''
          Keys used for escaping, if it is a function will use the result everytime.

          Example (recommended):

          keys.__raw = \'\'
            function()
              return vim.api.nvim_win_get_cursor(0)[2] > 1 and '<esc>l' or '<esc>'
            end
          \'\';
        '';
    };
```

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,72 @@
...
+
+  config = let
+    setupOptions = with cfg;
+      {
+        inherit mapping keys;
+        timeout =
+          if isString timeout
+          then helpers.mkRaw "timeout"
```
GaetanLepage, line 59:
```suggestion
          then helpers.mkRaw timeout
```

plugins/utils/better-escape.nix
GaetanLepage: No, it is indeed either a string or a lua function.

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,72 @@
...
+        List of mappings to use to enter escape mode.
+      '';
+
+      timeout =
+        helpers.defaultNullOpts.mkNullable
+        (with types;
+            either ints.unsigned str)
+        "vim.o.timeoutlen" ''
+          The time in which the keys must be hit in ms. Use option timeoutlen by default.
+        '';
```
GaetanLepage, line 28:
```suggestion
        helpers.defaultNullOpts.mkNullable
        (with types;
            either ints.unsigned str)
        "vim.o.timeoutlen"
        ''
          The time in which the keys must be hit in ms.
          Uses the value of `vim.o.timeoutlen` (`options.timeoutlen` in nixvim) by default.
        '';
```

plugins/utils/better-escape.nix
```diff
@@ -0,0 +1,73 @@
...
+    setupOptions = with cfg;
+      {
+        inherit mapping keys;
+        timeout =
+          if isInt timeout
+          then helpers.mkRaw "timeout"
+          else timeout;
+        clear_empty_lines = clearEmptyLines;
+      }
+      // cfg.extraOptions;
```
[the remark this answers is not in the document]
hmajid2301: ahh okay, i thought we wanted to keep the inherits together? I thought we did that for another PR. :+1: Ahh well this makes more sense. Will keep it in mind for other PRs

GaetanLepage (approved):
LGTM

PR #632, opened 2023-10-09
fix(helpers): use "__unkeyed" prefix instead of "@" in `toLuaObject`

### LINKED ISSUE #627: can't define treesitter capture highlight groups starting with `@`
**Plugin affected**: n/a
**Nixpkgs channel**: unstable
**Home-manager version**: 23.11-pre

## Description

due to the definition of `helpers.toLuaObject`, key names in `nixvim.highlight` cannot begin with `@`. but treesitter highlight groups start with `@`: <https://github.com/nvim-treesitter/nvim-treesitter#highlight>

is there a workaround for this? or do we want to discuss an alternative to `@` in `toLuaObject`?

## Config

<https://codeberg.org/municorn/dotfiles/src/branch/nixvim/homes/muni/nvim/highlight.nix>

muni-corn:
so that neovim (treesitter) highlight groups can start with "@".

this may break existing configs which use "@" for key-less table entries.

i'm open to other solutions or a different prefix name :)

closes #627

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -21,7 +21,7 @@ with lib; rec {
         + (concatStringsSep ","
           (mapAttrsToList
             (n: v:
-              if head (stringToCharacters n) == "@"
+              if (builtins.match "__unkeyed.*" n) != null
               then toLuaObject v
               else if n == "__emptyString"
               then "[''] = " + (toLuaObject v)
--- a/plugins/filetrees/neo-tree.nix
+++ b/plugins/filetrees/neo-tree.nix
@@ -990,7 +990,7 @@ in {
             name: value: {
               name =
                 if name == "name"
-                then "@"
+                then "__unkeyed"
                 else name;
               value =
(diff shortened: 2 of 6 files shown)


traxys (changes requested):
There are some places where the syntax was used in nixvim, like in `tests/lib-tests.nix`, `plugins/statusline/lualine.nix`, `plugins/filetrees/neo-tree.nix`, `plugins/utils/oil.nix`

GaetanLepage:
Thank you for your proposition !
I think that this change is possible.
Could you look at the occurences of `@`-prefixed keys in the nixvim repo please ?
They would need to be fixed in the same PR.

muni-corn:
on it! o7

muni-corn:
done 😎

pta2002 (approved):

GaetanLepage (approved):

traxys (approved):

PR #641, opened 2023-10-13
plugins/null-ls: migrate `null-ls` to `none-ls`

siph:
`null-ls` was archived [some time ago](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1621) and [`none-ls`](https://github.com/nvimtools/none-ls.nvim) is a fork and a drop-in replacement.

`none-ls` is packaged but not merged onto unstable yet (https://nixpk.gs/pr-tracker.html?pr=258774), so I am marking this as a draft.

This pr just replaces the default package with `none-ls`. I don't think this requires any deprecations or warnings because it doesn't make any breaking changes, plus the package is still configurable. From the `none-ls` readme:
```
Only the repo name is changed for compatibility concerns.
All the API and future changes will keep in place as-is.
```

FWIW, `LazyVim` has also migrated to `none-ls`: https://github.com/LazyVim/LazyVim/pull/1517

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -72,7 +72,7 @@
     ./lsp/nvim-lightbulb.nix
     ./lsp/trouble.nix
 
-    ./null-ls
+    ./none-ls
 
     ./dap
 
--- a/plugins/none-ls/default.nix
+++ b/plugins/none-ls/default.nix
@@ -5,22 +5,27 @@
   ...
 }:
 with lib; let
-  cfg = config.plugins.null-ls;
+  cfg = config.plugins.none-ls;
   helpers = import ../helpers.nix {inherit lib;};
 in {
   imports = [
     ./servers.nix
+    (
+      mkRenamedOptionModule
+      ["plugins" "null-ls"]
+      ["plugins" "none-ls"]
+    )
   ];
 
-  options.plugins.null-ls =
+  options.plugins.none-ls =
     helpers.extraOptionsOptions
     // {
-      enable = mkEnableOption "null-ls";
+      enable = mkEnableOption "none-ls";
 
       package = mkOption {
         type = types.package;
-        default = pkgs.vimPlugins.null-ls-nvim;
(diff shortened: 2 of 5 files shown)


GaetanLepage:
Good idea !
I would also rename the plugin in nixvim to `none-ls`.
- Rename the file (and change `plugins/default.nix` accordingly), and the test file as well
- Add a rename warning in `none-ls.nix`:
```
imports = [
  (
    mkRenamedOptionModule
    ["plugins" "null-ls"]
    ["plugins" "none-ls"]
];
```

GaetanLepage (approved):
LGTM !
Thank you for doing that :)

GaetanLepage:
@siph This PR was marked as a draft. Do you agree to merge it now ?

siph:
I think this is ready. I left the config documentation as-is because `none-ls` still refers to itself as `null-ls` in its docs.

GaetanLepage:
Shouldn't this be `none-ls` @siph ?
https://github.com/nix-community/nixvim/blob/77a3f3c29817e8ec14448022c934ae708968fd22/plugins/none-ls/helpers.nix#L65

There are also still some occurences of `null-ls` in the option descriptions and tests.

traxys:
It think not according to
> Only the repo name is changed for compatibility concerns.
> All the API and future changes will keep in place as-is.
From the PR request. Maybe they changed that stance, but that's another problem.

siph:
It's kind of weird. `none-ls` is basically just the repo name. The plugin is still named `null-ls` internally. For example, running `:h null-ls` still works and all the documentation (and the APIs) still refer to `null-ls`, with only a small section about migrating to `none-ls`.

That's why this PR originally only changed the default package and nothing else. When you asked me to rename to `none-ls` I left the documentation unchanged because it's basically unchanged in `none-ls`.

GaetanLepage:
Ok, makes sense. Thank you for those explanations !

PR #644, opened 2023-10-15
plugins/nix-develop: init + tests

hmajid2301:
Add [nix-develop](https://github.com/figsoda/nix-develop.nvim/tree/main) plugin.

It seems there are no config options! So this one is pretty simple ?

### DIFF
--- a/plugins/utils/nix-develop.nix
+++ b/plugins/utils/nix-develop.nix
@@ -24,21 +24,18 @@ in {
       };
     };
 
-  config = let
-    inherit (cfg) ignoredVariables separatedVariables;
-  in
-    mkIf cfg.enable {
-      extraPlugins = [cfg.package];
-      extraConfigLua = ''
-        local ignored_variables = ${helpers.toLuaObject ignoredVariables}
-        for ignoredVariable, shouldIgnore in ipairs(ignored_variables) do
-        	require("nix-develop").ignored_variables[ignoredVariable] = shouldIgnore
-        end
+  config = mkIf cfg.enable {
+    extraPlugins = [cfg.package];
+    extraConfigLua = ''
+      local __ignored_variables = ${helpers.toLuaObject cfg.ignoredVariables}
+      for ignoredVariable, shouldIgnore in ipairs(__ignored_variables) do
+      	require("nix-develop").ignored_variables[ignoredVariable] = shouldIgnore
+      end
 
-        local separated_variables = ${helpers.toLuaObject separatedVariables}
-        for variable, separator in ipairs(separated_variables) do
-        	require("nix-develop").separated_variables[variable] = separator
-        end
-      '';
-    };
+      local __separated_variables = ${helpers.toLuaObject cfg.separatedVariables}
+      for variable, separator in ipairs(__separated_variables) do
+      	require("nix-develop").separated_variables[variable] = separator
+      end
+    '';
+  };
 }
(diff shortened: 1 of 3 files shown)


traxys:
I'd say there is no `setup` but there still is some [configuration options](https://github.com/figsoda/nix-develop.nvim/blob/afea026f5c478c000a8af8de87f7b711676387ab/doc/nix-develop.txt#L18-L27).

I think `mkOption { type = types.attrsOf types.bool;}` and `mkOption { type = types.attrsOf types.str; }`  for both options should be the good option.

Though we can't just assign them to the require, we need to assign them key by key, like it's done for autocommands for example

```lua
            local __nixvim_autocommands = ${helpers.toLuaObject autoCmd}

            for _, autocmd in ipairs(__nixvim_autocommands) do
                  -- Stuff
            end
```

plugins/utils/nix-develop.nix
```diff
@@ -0,0 +1,44 @@
...
+        type = types.attrsOf types.bool;
+      };
+
+      separatedVariables = mkOption {
+        type = types.attrsOf types.str;
+      };
+    };
+
+  config = let
+    inherit (cfg) ignoredVariables separatedVariables;
```
hmajid2301, line 28: im struggling a bit with the correct syntax for the inherit ?

traxys: I think it would be clearer if you skip the let (cfg) inherit .... in, and directly to cfg.ignoredVariables & cfg.separatedVariables in extraConfigLua

plugins/utils/nix-develop.nix
```diff
@@ -14,9 +14,31 @@ in {
...
+  in
+    mkIf cfg.enable {
+      extraPlugins = [cfg.package];
+      extraConfigLua = ''
+        local ignored_variables = ${helpers.toLuaObject ignoredVariables}
+        for ignoredVariable, shouldIgnore in ipairs(ignored_variables) do
+        	require("nix-develop").ignored_variables[ignoredVariable] = shouldIgnore
+        end
+
+        local separated_variables = ${helpers.toLuaObject separatedVariables}
```
traxys, line 38: please prefix the {ignored,separated}_variables with `__` to avoid risks of mixing that with an users configuration

traxys (changes requested):

plugins/utils/nix-develop.nix
hmajid2301: ahh good shout!

plugins/utils/nix-develop.nix
```diff
@@ -0,0 +1,41 @@
...
+in {
+  options.plugins.nix-develop =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "nix-develop.nvim";
+
+      package = helpers.mkPackageOption "nix-develop.nvim" pkgs.vimPlugins.nix-develop-nvim;
+
+      ignoredVariables = mkOption {
+        type = types.attrsOf types.bool;
```
traxys, line 19: you need to provide a `default = {}` for both fields. Else nix will throw an error if enable = true  and the variables are unset by the user.

traxys (changes requested):

plugins/utils/nix-develop.nix
hmajid2301: ahh yeh!

traxys (approved):

PR #646, opened 2023-10-15
plugin/hmts: init + tests

hmajid2301:
Add the [hmts plugin](https://github.com/calops/hmts.nvim).

### DIFF
--- a/plugins/utils/hmts.nix
+++ b/plugins/utils/hmts.nix
@@ -18,9 +18,5 @@ in {
 
   config = mkIf cfg.enable {
     extraPlugins = [cfg.package];
-
-    extraConfigLua = ''
-      require('hmts').setup()
-    '';
   };
 }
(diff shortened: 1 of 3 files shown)


plugins/utils/hmts.nix
```diff
@@ -0,0 +1,26 @@
...
+
+      package = helpers.mkPackageOption "hmts.nvim" pkgs.vimPlugins.hmts-nvim;
+    };
+
+  config = mkIf cfg.enable {
+    extraPlugins = [cfg.package];
+
+    extraConfigLua = ''
+      require('hmts').setup()
+    '';
```
traxys, line 24: I think we don't need that:
https://github.com/calops/hmts.nvim#faq

traxys (changes requested):

plugins/utils/hmts.nix
hmajid2301: yeh i also thought about leaving it out! wasnt sure if it better to be more consistent with other plugins!

hmajid2301: should be fixed now.

traxys (approved):

PR #654, opened 2023-10-19
 plugins/telescope.undo: init

hmajid2301:
Added telescope extension: [telescope-undo](https://github.com/debugloop/telescope-undo.nvim)

### DIFF
--- a/plugins/telescope/undo.nix
+++ b/plugins/telescope/undo.nix
@@ -62,7 +62,7 @@ in {
       default = "";
     };
 
-    mappings = with mappings; {
+    mappings = {
       i =
         helpers.ifNonNull' i
         (
--- a/plugins/telescope/undo.nix
+++ b/plugins/telescope/undo.nix
@@ -42,8 +42,13 @@ in {
     };
 
     diffContextLines = mkOption {
-      type = types.nullOr types.int;
-      description = '''';
+      type =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            either ints.unsigned str
+        );
+      description = ''Defaults to the scrolloff'';
       default = "vim.o.scrolloff";
     };
 
@@ -60,18 +65,72 @@ in {
       '';
       default = "";
     };
-    mappings = {};
+
+    mappings = {
+      i =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            attrsOf str
+        )
+        ''
+               {
+          "<cr>" = "yank_additions";
+          "<s-cr>" = "yank_deletions";
+          "<c-cr>" = "restore";
+               }
+        '';
+      n =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            attrsOf str
+        )
+        ''
+               {
+          "y" = "yank_additions";
+          "Y" = "yank_deletions";
+          "u" = "restore";
+               }
+        '';
+    };
   };
 
   config = let
-    configuration = {
-      inherit (cfg) fuzzy;
-      override_generic_sorter = cfg.overrideGenericSorter;
-      override_file_sorter = cfg.overrideFileSorter;
-      case_mode = cfg.caseMode;
+    configuration = with cfg; {
+      use_delta = useDelta;
+      use_custom_command = useCustomCommand;
+      side_by_side = sideBySide;
+      diff_context_lines =
+        if isInt diffContextLines
+        then helpers.mkRaw "diffContextLines"
+        else diffContextLines;
+      entry_format = entryFormat;
+      time_format = timeFormat;
+      mappings = with mappings; {
+        i =
+          helpers.ifNonNull' i
+          (mapAttrs
+            (
+              key: action:
+                helpers.mkRaw "require('telescope-undo.actions').${action}"
+            )
+            mappings);
+
+        n =
+          helpers.ifNonNull' n
+          (mapAttrs
+            (
+              key: action:
+                helpers.mkRaw "require('telescope-undo.actions').${action}"
+            )
+            mappings);
+      };
     };
   in
-    mkIf cfg.enable {
+    mkIf
+    cfg.enable
+    {
       extraPlugins = [cfg.package];
 
       plugins.telescope.enabledExtensions = ["undo"];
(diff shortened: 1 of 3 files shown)


plugins/telescope/undo.nix
```diff
@@ -0,0 +1,139 @@
...
+          helpers.ifNonNull' i
+          (mapAttrs
+            (
+              key: action:
+                helpers.mkRaw "require('telescope-undo.actions').${action}"
+            )
+            mappings);
+
+        n =
+          helpers.ifNonNull' n
```
hmajid2301, line 121: also not sure if referencing `n` here will cause issues with `shadowing` let me know :) (and same for the i)

plugins/telescope/undo.nix
```diff
@@ -0,0 +1,139 @@
...
+
+    timeFormat = mkOption {
+      type = types.nullOr types.str;
+      description = ''
+        Can be set to a [Lua date format string](https://www.lua.org/pil/22.1.html).
+      '';
+      default = "";
+    };
+
+    mappings = {
```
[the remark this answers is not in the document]
GaetanLepage: That looks a perfectly valid solution to me !

plugins/telescope/undo.nix
GaetanLepage: `shadowing` ?

GaetanLepage (changes requested):

plugins/telescope/undo.nix
hmajid2301: "reusing" the name var `n` twice ?

GaetanLepage: No, this should work just fine.
Could you please add some test in `tests/test-sources/plugins/telescope/undo.nix` please ?

plugins/telescope/undo.nix
```diff
@@ -0,0 +1,139 @@
...
+        )
+        ''
+           {
+                 "y" = "yank_additions";
+                 "Y" = "yank_deletions";
+                 "u" = "restore";
+          }
+        '';
+    };
+  };
```
GaetanLepage, line 97: Warning, you missed the description.
```suggestion
      i =
        helpers.defaultNullOpts.mkNullable
        (
          with types;
            attrsOf str
        )
        ''
          {
               "<cr>" = "yank_additions";
               "<s-cr>" = "yank_deletions";
               "<c-cr>" = "restore";
          }
        ''
        "<INSERT_DESCRIPTION>";
      n =
        helpers.defaultNullOpts.mkNullable
        (
          with types;
            attrsOf str
        )
        ''
           {
                 "y" = "yank_additions";
                 "Y" = "yank_deletions";
                 "u" = "restore";
          }
        ''
        "<INSERT_DESCRIPTION>";
    };
  };
```

plugins/telescope/undo.nix
```diff
@@ -0,0 +1,139 @@
...
+    diffContextLines = mkOption {
+      type =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            either ints.unsigned str
+        );
+      description = ''Defaults to the scrolloff'';
+      default = "vim.o.scrolloff";
+    };
```
GaetanLepage, line 53:
```suggestion
    diffContextLines = mkOption {
      type = with types;
        either ints.unsigned str;
      description = ''Defaults to the scrolloff'';
      default = "vim.o.scrolloff";
    };
```

GaetanLepage (changes requested):
Why did you use `mkOption` instead of the `null`-defaulting options that we usually use ?

hmajid2301:
> Why did you use `mkOption` instead of the `null`-defaulting options that we usually use ?

I used `fzf-native` plugin as an example and they used mkOption there. There also didnt seem to be any tests for these, hence I didn't create any for these :sweat:! 

But will add some tests now!!! :)

tests/test-sources/plugins/telescope/undo.nix
```diff
@@ -0,0 +1,34 @@
...
+      extensions.undo = {
+        enable = true;
+
+        useDelta = true;
+        useCustomCommand = ["bash" "-c" "echo '$DIFF' | delta"];
+        sideBySide = true;
+        diffContextLines = 8;
+        entryFormat = "state #$ID";
+        timeFormat = "!%Y-%m-%dT%TZ";
+        mappings = {
```
hmajid2301, line 19: I think the test is failing here:

```bash
error:
       … while checking flake output 'checks'

         at /nix/store/p58d2j0ac7zvja5jl14xzbc19fakjxh2-source/lib.nix:39:17:

           38|               {
           39|                 ${key} = (attrs.${key} or { })
             |                 ^
           40|                   // { ${system} = ret.${key}; };

       … while checking the derivation 'checks.x86_64-linux.plugins-telescope-undo-example'

         at «none»:0: (source not available)

       error: cannot coerce a set to a string

```

GaetanLepage:
> I used `fzf-native` plugin as an example and they used mkOption there. There also didnt seem to be any tests for these, hence I didn't create any for these 😓!

Yes, this is totally fair. However, this is an example of an older part of the code base (untested + not much using the `lib/helpers.nix` functions).
It is better to switch to `helpers.defaultNullOpts.MkXxxx` where possible as well as testing.

plugins/telescope/undo.nix
```diff
@@ -0,0 +1,137 @@
...
+        n =
+          helpers.ifNonNull' n
+          (mapAttrs
+            (
+              key: action:
+                helpers.mkRaw "require('telescope-undo.actions').${action}"
+            )
+            mappings);
+      };
+    };
```
GaetanLepage, line 127:
```suggestion
      mappings = with mappings; {
        i =
          helpers.ifNonNull' i
          (
            mapAttrs
            (key: action:
              helpers.mkRaw "require('telescope-undo.actions').${action}")
            i
          );

        n =
          helpers.ifNonNull' n
          (
            mapAttrs
            (key: action:
              helpers.mkRaw "require('telescope-undo.actions').${action}")
            n
          );
      };
    };
```

tests/test-sources/plugins/telescope/undo.nix
GaetanLepage: Tests now pass

plugins/telescope/undo.nix
GaetanLepage: This still needs to be fixed.

GaetanLepage (approved):
Good job !

PR #656, opened 2023-10-20
fix: harpoon telescope integration

hmajid2301:
At the moment if your config doesn't specify `enableTelescope` in the harpoon options you get the following:

Harpoon config:
```
      harpoon = {
        enable = true;
        tmuxAutocloseWindows = true;
        keymaps = {
          addFile = "<leader>ha";
          toggleQuickMenu = "<leader>ht";
          navNext = "<leader>hn";
          navPrev = "<leader>hp";
        };
      };
```

Error:
```bash
error:

         at /nix/store/4smdp3v5a9bcg1hpdhl2hs70rq385ykq-source/lib/lists.nix:57:9:

           56|       fold' = n:
           57|         if n == len
             |         ^
           58|         then nul

       … while calling the 'length' builtin

         at /nix/store/4smdp3v5a9bcg1hpdhl2hs70rq385ykq-source/lib/lists.nix:55:13:

           54|     let
           55|       len = length list;
             |             ^
           56|       fold' = n:

       error:
       Failed assertions:
       - Nixvim: The harpoon telescope integration needs telescope to function as intended
```

I think we should default it to `false` and make users turn it on.

### DIFF
--- a/plugins/utils/harpoon.nix
+++ b/plugins/utils/harpoon.nix
@@ -29,6 +29,12 @@ in {
 
       enableTelescope = mkEnableOption "enable telescope integration";
 
+      keymapsSilent = mkOption {
+        type = types.bool;
+        description = "Whether harpoon keymaps should be silent.";
+        default = false;
+      };
+
       keymaps = {
         addFile = helpers.mkNullOrOption types.str ''
           Keymap for marking the current file.";
--- a/plugins/utils/harpoon.nix
+++ b/plugins/utils/harpoon.nix
@@ -27,7 +27,7 @@ in {
 
       package = helpers.mkPackageOption "harpoon" pkgs.vimPlugins.harpoon;
 
-      enableTelescope = mkEnableOption "enable telescope integration";
+      enableTelescope = mkEnableOption "telescope integration";
 
       keymapsSilent = mkOption {
         type = types.bool;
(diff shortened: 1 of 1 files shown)


traxys:
This seems strange, because mkEnableOption should already default to false

hmajid2301:
it could be the assertion logic is incorrect ?

hmajid2301:
yeh i think the assertion logic shouldve been reversed my bad :sweat: 

if harpoon extension enabled -> telescope is enabled

GaetanLepage:
Why have you removed the `keymapsSilent` option ?

hmajid2301:
my bad fixed now!

plugins/utils/harpoon.nix
```diff
@@ -27,7 +27,7 @@ in {
 
       package = helpers.mkPackageOption "harpoon" pkgs.vimPlugins.harpoon;
 
-      enableTelescope = mkEnableOption "Enable telescope integration";
+      enableTelescope = mkEnableOption "enable telescope integration";
```
GaetanLepage, line 30:
```suggestion
      enableTelescope = mkEnableOption "telescope integration";
```

GaetanLepage: `mkEnableOption` already prefixes your provided description with `Whether to enable `.

hmajid2301:
good to know! For next time.

GaetanLepage (approved):
Well done !

PR #668, opened 2023-10-24
flake: add overlay as an alternative to legacyPackages

PatrickDaG:
This adds an overlay containing both makeNixvim functions. This simplifies usage of the functions, as well as enabling you to use your own preinstantiated package set, instead of the one from this flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,6 +127,18 @@
         nixDarwinModules.nixvim = import ./wrappers/darwin.nix modules;
         rawModules.nixvim = nixvimModules;
 
+        overlays.default = final: prev: {
+          nixvim = rec {
+            makeNixvimWithModule = import ./wrappers/standalone.nix prev modules;
+            makeNixvim = configuration:
+              makeNixvimWithModule {
+                module = {
+                  config = configuration;
+                };
+              };
+          };
+        };
+
         templates = let
           simple = {
             path = ./templates/simple;


traxys:
I'm not too sold out on the overlay, you can easily access those functions through module._args if you use NixOS modules, or through extraSpecialArgs for Home-Manager.

For the case of using a predefined `nixpkgs` there is the option with `makeNixvimWithModule` to pass a `pkgs` argument.

I find it a bit weird to add non package (even non derivation) attributes to an overlay.

PatrickDaG:
I would like to use nixvim standalone, without needing to enable the nixos or home-manager modules. This enables me to have multiple vim configurations completely independent of each other.

PatrickDaG:
I would like to refrain form using either the nixos or home manager modules just to get the functions, as I don't intend to use the options.
With the current legacyPackages approach, this requires me to be able to refer to my flake's inputs at the point where I want to call makeNixvimWithModule, and also requires me to know the current system. So this would be come this rather unwieldy expression:

  ```nix
  # home-manager.extraSpecialArgs.flakeInputs = flakeInputs;
  { flakeInputs, pkgs, ... }: {
    home.shellAliases.secondaryVim = lib.getExe (flakeInputs.nixvim.legacyPackages.${pkgs.stdenv.hostPlatform.system}.makeNixvimWithModule {
      inherit pkgs;
      module.config = {
        # <nixvim config>
        # globals.mapleader = ",";
      };
    });
  }
  ```

  By exposing the function as an overlay you would implicitly add the correct system context without any extra effort, and this becomes much more readable:

  ```nix
  { pkgs, ... }: {
    home.shellAliases.secondaryVim = lib.getExe (pkgs.nixvim.makeNixvim {
      # <nixvim config>
      # globals.mapleader = ",";
    });
  }
  ```

GaetanLepage (approved):
As your proposal is simply an addition and would not affect any current user, I have nothing against it.
I don't see any reason why not to add that as yet an extra way to declare a nixvim configuration.
@traxys don't you think that we could just go and experiment with that ? Removing will obviously remain an option if this causes any negative effect.

PR #670, opened 2023-10-25
plugins/catppuccin!: update integrations

mrtnvgr:
This PR updates catppuccin integrations and sorts them to alphabetical order

### DIFF
--- a/plugins/colorschemes/catppuccin.nix
+++ b/plugins/colorschemes/catppuccin.nix
@@ -233,7 +233,7 @@ in {
         mason = helpers.defaultNullOpts.mkBool false "";
 
         mini = {
-          enabled = helpers.defaultNullOpts.mkBool config.plugins.mini.enable "";
+          enabled = helpers.defaultNullOpts.mkBool false "";
 
           indentscope_color = helpers.defaultNullOpts.mkStr "" "";
         };
--- a/plugins/colorschemes/catppuccin.nix
+++ b/plugins/colorschemes/catppuccin.nix
@@ -144,9 +144,19 @@ in {
       integrations = {
         aerial = helpers.defaultNullOpts.mkBool false "";
 
-        alpha = helpers.defaultNullOpts.mkBool false "";
+        alpha = helpers.defaultNullOpts.mkBool true "";
 
         barbar = helpers.defaultNullOpts.mkBool config.plugins.barbar.enable "";
+
+        barbecue = {
+          dim_dirname = helpers.defaultNullOpts.mkBool true "";
+
+          bold_basename = helpers.defaultNullOpts.mkBool true "";
+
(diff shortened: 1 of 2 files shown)


plugins/colorschemes/catppuccin.nix
```diff
@@ -199,6 +273,10 @@ in {
 
         noice = helpers.defaultNullOpts.mkBool config.plugins.noice.enable "";
 
+        # TODO: NormalNvim = ;
```
mrtnvgr, line 276: How can I rename this option to snake_case?

mrtnvgr (commented):
Should all integrations relay on their plugin enabled state?

mrtnvgr:
I think we need a type which would allow for `integrations.mini = enable` and `integrations.mini.enabled = true`. Catppuccin [does this at compile time](https://github.com/catppuccin/nvim/blob/ea52fe8a0b1e4a820df0d0cf9a6a5a0e18c3eaa0/lua/catppuccin/lib/mapper.lua#L29-L39).

plugins/colorschemes/catppuccin.nix
GaetanLepage: `normal_nvim` ?

mrtnvgr: > `normal_nvim` ?

This integration's name is in camel case, will it work? https://github.com/catppuccin/nvim/commit/0e3c128eea8a7de692778d52b8429817df5c9040

GaetanLepage: Oh, I didn't understand well. I would say that we leave it as `NormalNvim` (same as upstream).

GaetanLepage:
@mrtnvgr is your PR now ready for review ? If so, don't hesitate to undraft it accordingly so that I take a closer look.

plugins/colorschemes/catppuccin.nix
```diff
@@ -144,21 +144,52 @@ in {
...
+          bold_basename = helpers.defaultNullOpts.mkBool true "";
+
+          dim_context = helpers.defaultNullOpts.mkBool false "";
+
+          alt_background = helpers.defaultNullOpts.mkBool false "";
+        };
+
         beacon = helpers.defaultNullOpts.mkBool false "";
 
+        # TODO: bufferline = ;
```
GaetanLepage, line 163: You still have a few TODOs left

mrtnvgr: These TODOs are special integrations which require additional setup in other plugins. I need help to implement them

GaetanLepage: Ok, let's leave them as is for now.

GaetanLepage (approved):

PR #672, opened 2023-10-27
fantomas: init

MikaelFangel:
### Changes
 * Adds the fantomas formatter for F#, which is used by fsautocomplete and vim-Ionide.


traxys:
If fsautocomplete requires this program to be installed to function correctly I'd suggest to add it automatically. This can be done by adding the fantomas package with `extraPackages = [fantomas]`. 

You can look at `plugins/lsp/language-servers/nixd.nix` for an example

traxys (approved):

PR #676, opened 2023-10-30
plugins/refactoring: init + test

hmajid2301:
Add new plugin [refactoring.nvim])https://github.com/ThePrimeagen/refactoring.nvim)

### DIFF
--- a/plugins/utils/refactoring.nix
+++ b/plugins/utils/refactoring.nix
@@ -32,7 +32,9 @@ in {
           }
         ''
         ''
-          For certain languages like Golang, types are required for functions that return an object(s) and parameters of functions. Unfortunately, for some parameters and functions there is no way to automatically find their type. In those instances, we want to provide a way to input a type instead of inserting a placeholder value.
+          For certain languages like Golang, types are required for functions that return an object(s) and parameters of functions.
+          Unfortunately, for some parameters and functions there is no way to automatically find their type. In those instances,
+          we want to provide a way to input a type instead of inserting a placeholder value.
         '';
 
       promptFuncParamType =
@@ -53,14 +55,16 @@ in {
           }
         ''
         ''
-          For certain languages like Golang, types are required for functions that return an object(s) and parameters of functions. Unfortunately, for some parameters and functions there is no way to automatically find their type. In those instances, we want to provide a way to input a type instead of inserting a placeholder value.
+          For certain languages like Golang, types are required for functions that return an object(s) and parameters of functions.
+          Unfortunately, for some parameters and functions there is no way to automatically find their type. In those instances,
+          we want to provide a way to input a type instead of inserting a placeholder value.
         '';
 
       printVarStatements =
         helpers.defaultNullOpts.mkNullable
         (
           with types;
-            attrsOf listof str
+            attrOf (listOf str)
         )
         ''
           {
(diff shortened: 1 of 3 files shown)


plugins/utils/refactoring.nix
```diff
@@ -0,0 +1,165 @@
...
+      package = helpers.mkPackageOption "refactoring.nvim" pkgs.vimPlugins.refactoring-nvim;
+
+      promptFuncReturnType =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            attrsOf bool
+        )
+        ''
+          {
```
hmajid2301, line 24: Im not sure if this it the best approach ?

plugins/utils/refactoring.nix
```diff
@@ -0,0 +1,165 @@
...
+  config = let
+    cfg = config.plugins.refactoring;
+  in
+    mkIf cfg.enable {
+      extraPlugins = [cfg.package];
+
+      extraConfigLua = let
+        opts = {
+          inherit (cfg);
+          prompt_func_return_type = promptFuncReturnType;
```
hmajid2301, line 155: do we need to do any transformation here ?

plugins/utils/refactoring.nix
traxys: I think it is

plugins/utils/refactoring.nix
```diff
@@ -0,0 +1,165 @@
...
+        )
+        ''
+          {
+            "go" = "false";
+            "java" = "false";
+            "cpp" = "false";
+            "c" = "false";
+            "h" = "false";
+            "hpp" = "false";
+            "cxx" = "false";
```
traxys, line 31: There are no quotes around "false"

plugins/utils/refactoring.nix
```diff
@@ -0,0 +1,165 @@
...
```
traxys, line 54: No quotes around false

plugins/utils/refactoring.nix
traxys: I don't think so

traxys (changes requested):

plugins/utils/refactoring.nix
traxys: You need to do `XXXX = cfg.XXXX;` though (see the CI fail)

hmajid2301: I had it committed, just not pushed!

plugins/utils/refactoring.nix
hmajid2301: done

plugins/utils/refactoring.nix
```diff
@@ -0,0 +1,164 @@
...
+          For certain languages like Golang, types are required for functions that return an object(s) and parameters of functions.
+          Unfortunately, for some parameters and functions there is no way to automatically find their type. In those instances,
+          we want to provide a way to input a type instead of inserting a placeholder value.
+        '';
+
+      printVarStatements =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            attrOf (listOf str)
```
traxys, line 67: CI fails due to typo: attrOf --> attrsOf
You should run `nix flake check`, it would find those sorts of things

traxys (changes requested):

plugins/utils/refactoring.nix
hmajid2301: Yeh sorry, I just usually like to push my changes. I'll try to check before pushing :sweat:

traxys (approved):

PR #681, opened 2023-11-02
fix(plugins/utils/spider): `ge` mapping not working

trueNAHO:
Reference: https://github.com/chrisgrieser/nvim-spider/blob/86236bfc573ae45193bf0eb663be8a7184458281/README.md?plain=1#L119-L124


GaetanLepage:
Nice catch, thanks !

GaetanLepage (approved):

PR #683, opened 2023-11-04
plugins/harpoon: fix busted keybind

zachcoyle:
fix keybinding where the action was being called at the time of creation of the keybind instead of when the keybind was triggered


GaetanLepage (approved):
Nice catch, thank you !

PR #688, opened 2023-11-05
fix: Enable `helpers` to be passed as an argument in nixvim

traxys:
With this change `helpers` can be imported using `{pkgs, lib, helpers, ...}` instead of requiring to `import` it.


GaetanLepage (approved):
Good idea !
We might want to change all the files in consequence, no ?

traxys:
Maybe over time, there are over 150 files to change ....

PR #695, opened 2023-11-11
Add support for the Lean language

GaetanLepage:
- add `plugins.lsp.servers.lean` ([leanls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#leanls))
- add `plugins.lean` ([lean.nvim](https://github.com/Julian/lean.nvim))

### DIFF
--- a/plugins/languages/lean.nix
+++ b/plugins/languages/lean.nix
@@ -1,0 +1,316 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.lean;
+in {
+  options.plugins.lean =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "lean-nvim";
+
+      package = helpers.mkPackageOption "lean-nvim" pkgs.vimPlugins.lean-nvim;
+
+      leanPackage = mkOption {
+        type = with types; nullOr package;
+        default = pkgs.lean4;
+        description = "Which package to use for lean.";
+        example = null;
+      };
+
+      lsp =
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            submodule {
+              freeformType = types.attrs;
+              options = {
+                enable = helpers.defaultNullOpts.mkBool true ''
+                  Whether to enable the Lean language server(s) ?
+
+                  Set to `false` to disable, otherwise should be a table of options to pass to
+                  `leanls`.
+                  See https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#leanls
+                  for details.
+                '';
+
+                on_attach = helpers.mkNullOrOption types.str ''
+                  The LSP handler.
+                '';
+
+                init_options = helpers.mkNullOrOption (with types; attrsOf anything) ''
+                  The options to configure the lean language server.
+                  See `Lean.Lsp.InitializationOptions` for details.
+                '';
+              };
+            }
+        )
+        "{}"
+        "LSP settings.";
+
+      ft = {
+        default = helpers.defaultNullOpts.mkStr "lean" ''
+          The default filetype for Lean files.
+        '';
+
+        nomodifiable = helpers.mkNullOrOption (with types; listOf str) ''
+          A list of patterns which will be used to protect any matching Lean file paths from being
+          accidentally modified (by marking the buffer as `nomodifiable`).
+
+          By default, this list includes the Lean standard libraries, as well as files within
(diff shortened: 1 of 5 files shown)


plugins/languages/lean.nix
```diff
@@ -0,0 +1,316 @@
...
+                  `leanls`.
+                  See https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#leanls
+                  for details.
+                '';
+
+                on_attach = helpers.mkNullOrOption types.str ''
+                  The LSP handler.
+                '';
+
+                init_options = helpers.mkNullOrOption types.attrs ''
```
traxys, line 45: Could you use types.attrsOf types.anything? 

They are better handled in merging:
```
    attrs = mkOptionType {
      merge = loc: foldl' (res: def: res // def.value) {};
    };
```

vs 

```
    attrsOf = elemType: mkOptionType rec {
      merge = loc: defs:
        mapAttrs (n: v: v.value) (filterAttrs (n: v: v ? value) (zipAttrsWith (name: defs:
            (mergeDefinitions (loc ++ [name]) elemType defs).optionalValue
          )
          # Push down position info.
          (map (def: mapAttrs (n: v: { inherit (def) file; value = v; }) def.value) defs)));
    };

    anything = mkOptionType {
      merge = loc: defs:
        let
          getType = value:
            if isAttrs value && isStringLike value
            then "stringCoercibleSet"
            else builtins.typeOf value;

          # Returns the common type of all definitions, throws an error if they
          # don't have the same type
          commonType = foldl' (type: def:
            if getType def.value == type
            then type
            else throw "The option `${showOption loc}' has conflicting option t
ypes in ${showFiles (getFiles defs)}"
          ) (getType (head defs).value) defs;

          mergeFunction = {
            # Recursively merge attribute sets
            set = (attrsOf anything).merge;
            # Safe and deterministic behavior for lists is to only accept one d
efinition
            # listOf only used to apply mkIf and co.
            list =
              if length defs > 1
              then throw "The option `${showOption loc}' has conflicting defini
tions, in ${showFiles (getFiles defs)}."
              else (listOf anything).merge;
            # This is the type of packages, only accept a single definition
            stringCoercibleSet = mergeOneOption;
            lambda = loc: defs: arg: anything.merge
              (loc ++ [ "<function body>" ])
              (map (def: {
                file = def.file;
                value = def.value arg;
              }) defs);
            # Otherwise fall back to only allowing all equal definitions
          }.${commonType} or mergeEqualOption;
        in mergeFunction loc defs;
  };
```

plugins/languages/lean.nix
```diff
@@ -0,0 +1,316 @@
...
+        '';
+
+        lean3 = {
+          showFilter = helpers.defaultNullOpts.mkBool true "";
+
+          mouseEvents = helpers.defaultNullOpts.mkBool false ''
+            Setting this to `true` will simulate mouse events in the Lean 3 infoview, this is buggy
+            at the moment so you can use the I/i keybindings to manually trigger these.
+          '';
+        };
```
traxys, line 145: shouldn this be in the global scope  and not in `infoview` ? (as from the README.rst):

```lua
  -- Redirect Lean's stderr messages somehwere (to a buffer by default)
  stderr = {
    enable = true,
    -- height of the window
    height = 5,
    -- a callback which will be called with (multi-line) stderr output
    -- e.g., use:
    --   on_lines = function(lines) vim.notify(lines) end
    -- if you want to redirect stderr to `vim.notify`.
    -- The default implementation will redirect to a dedicated stderr
    -- window.
    on_lines = nil,
  },

  -- Legacy Lean 3 support (on_attach is as above, your LSP handler)
  lsp3 = { on_attach = on_attach },

  -- mouse_events = true will simulate mouse events in the Lean 3 infoview, this is buggy at the moment
  -- so you can use the I/i keybindings to manually trigger these
  lean3 = { mouse_events = false },
}
```

traxys (changes requested):

plugins/languages/lean.nix
GaetanLepage: Oh I didn't know that !
Done

plugins/languages/lean.nix
GaetanLepage: This is indeed what the documentation suggests, but after looking at the code, this is clearly under `infoview`.
This isn't the only error I spotted in this documentation section.

traxys: Indeed.... I tried to search for a compatibility layer but there does not seem to be any

traxys (approved):

PR #696, opened 2023-11-13
plugins/efm: Install mypy & ruff

GaetanLepage:


traxys (approved):
The commit name is misleading, it's more along the lines of "plugin/efm: Install mypy & ruff".

I think we need a warning if the user selects programs that are not in this list.

PR #698, opened 2023-11-13
plugins/lsp: add biome language server

zachcoyle:


plugins/lsp/language-servers/default.nix
```diff
@@ -20,6 +20,12 @@ with lib; let
       description = "Enable bashls, for bash.";
       package = pkgs.nodePackages.bash-language-server;
     }
+    {
+      name = "biome";
+      description = "Enable Biome, Toolchain of the Web";
+      package = pkgs.biome;
+      cmd = cfg: ["${cfg.package}/bin/biome" "lsp-proxy"];
```
GaetanLepage, line 27: I don't think that this is needed as `nvim-lspconfig` has set `biome lsp-proxy` as the default command.
As `cfg.package` will be added to `extraPackages` (and thus `"${cfg.package}/bin"` will be added to the `$PATH`), the command should resolve nicely.

plugins/lsp/language-servers/default.nix
```diff
@@ -20,6 +20,11 @@ with lib; let
       description = "Enable bashls, for bash.";
       package = pkgs.nodePackages.bash-language-server;
     }
+    {
+      name = "biome";
+      description = "Enable Biome, Toolchain of the Web";
+      package = pkgs.biome;
+    }
```
GaetanLepage, line 27:
```suggestion
    {
      name = "biome";
      description = "Enable Biome, Toolchain of the Web";
    }
```
Actually we don't need this line either because by default, we assume that the corresponding package has the same name as the language server.

zachcoyle: ah! makes sense. I was wondering why some didn't have a `package`

GaetanLepage (approved):
All good ! Thanks :)

PR #703, opened 2023-11-15
plugins/dap-go: add buildFlags option

hmajid2301:
We are missing the `build_flags` config option, https://github.com/leoluz/nvim-dap-go#configuring


GaetanLepage (approved):
Thank you for this addition :

hmajid2301:
I need it for my debugger :sweat_smile: as well :).

PR #705, opened 2023-11-16
plugins/efmls-configs: Improve tool installation

traxys:
This adds a few missing tools, and adds the following behaviours: 
- Warning when selecting a tool nixvim is not able to install (it is either too new and was not added, or not packaged in nixpkgs)
- Allow to override tool packages (this is usefull for example for `mdformat` that can have plugins added to it)

### DIFF
--- a/plugins/lsp/language-servers/efmls-configs.nix
+++ b/plugins/lsp/language-servers/efmls-configs.nix
@@ -14,65 +14,91 @@ with lib; let
   toolPkgs = {
     inherit
       (pkgs)
+      actionlint
+      alejandra
       ameba
       astyle
       bashate
+      beautysh
+      biome
       black
       cbfmt
+      checkmake
       clazy
+      codespell
       cppcheck
       cpplint
+      dfmt
+      djlint
       dmd
       dprint
       fish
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -9,12 +9,119 @@ with lib; let
...
+    mcs = pkgs.mono;
+    php_cs_fixer = pkgs.phpPackages.php-cs-fixer;
+    prettier_d = pkgs.prettierd;
+    slither = pkgs.slither-analyzer;
+    staticcheck = pkgs.go-tools;
+    terraform_fmt = pkgs.terraform;
+    vint = pkgs.vim-vint;
+    write_good = pkgs.write-good;
+    yq = pkgs.yq-go;
+  };
```
GaetanLepage, line 102: I know it was like this before, but a `with pkgs;` could make this more readable.

GaetanLepage (changes requested):
Very cool ! Just one minor suggestion.

GaetanLepage (approved):

PR #709, opened 2023-11-17
tests: Install all efmls-configs tools

traxys:
This requires to pass the efmls-configs file to the derivation helpers, if we try to import it from the test file directly it fails.

We then look into the types to find the available languages & corresponding tools, to enable them all.

We then filter out the broken tools or the unpackaged tools, in order to detect more easily the introduction of future tools.

### DIFF
--- a/tests/test-sources/plugins/lsp/efmls-configs.nix
+++ b/tests/test-sources/plugins/lsp/efmls-configs.nix
@@ -1,6 +1,65 @@
-{
+{efmls-options, ...}: {
   empty = {
     plugins.efmls-configs.enable = true;
+  };
+
+  all = {
+    plugins.efmls-configs = let
+      options = efmls-options.options.plugins.efmls-configs;
+      # toolOptions is an attrsets of the form:
+      # { <lang> = { linter = tools; formatter = tools; }; }
+      # Where tools is the option type representing the valid tools for this language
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/lsp/efmls-configs.nix
```diff
@@ -1,8 +1,50 @@
...
+        "xo"
+      ];
+
+      toolsFromOptions = opt:
+        builtins.filter (t: !builtins.elem t (brokenTools ++ unpackaged))
+        opt.type.nestedTypes.right.nestedTypes.elemType.nestedTypes.left.functor.payload;
+    in {
+      enable = true;
+
+      setup = builtins.mapAttrs (_: builtins.mapAttrs (_: toolsFromOptions)) toolOptions;
```
GaetanLepage, line 44: I wouldn't mind having either (or both) a little clearer implementation (yet more verbose) or some comments to explain what is going on.
(Especially for the `setup` line).

traxys: Done! I agree that it's better, finding the correct invocation was a lot of `pkgs.lib.debug.traceSeqN`

GaetanLepage (approved):
Good !

PR #716, opened 2023-11-18
neo-tree: add window.mappings to document_symbol

andrevmatos:
In current code, `filesystem.window.mappings`, when set, became `filesystem.window.mappings.mappings`.
`document_symbol` can also receive its own custom mappings.


traxys:
I prefered the solution in #725 (sorry for taking too long to review this one). Though that PR does not handle the `document_symbol` mappings, so if you could rebase only including that?

Though you need to add the `window` processing in the `document_symbol = { ... }` part of the file (around line 1206), see other windows

andrevmatos:
Done

traxys:
You still to add the processing to the setup options (see around line 1200 of the file, and other window options)

traxys (approved):

Issue #707, opened 2023-11-17
nixvim standalone broken after staging-next 2023-10-26


andrevmatos:
**Nixpkgs channel**: master

## Description
nixpkgs just got a staging-next merge to master (NixOS/nixpkgs#263535), which seems to have broken nixvim. It maybe due to NixOS/nixpkgs#237425 making `pkgs.neovim` not a `symlinkJoin` anymore, while nixvim standalone seems to rely on `paths` attribute here:
https://github.com/nix-community/nixvim/blob/a304b68a882357a55c4d76eb6db299e4d11780c5/wrappers/standalone.nix#L34-L35

```
error:
       … while calling the 'getAttr' builtin

         at /builtin/derivation.nix:19:19: (source not available)

       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'neovim-20231115_d92dd2a'
         whose name attribute is located at /nix/store/kjg0gm2y4f5zgw6cqidh4w6c5n3q6fww-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'paths' of derivation 'neovim-20231115_d92dd2a'

         at /nix/store/1ann4n8aba6qd5sr93gx1yjj7pinfk2d-source/wrappers/standalone.nix:34:5:

           33|   config.finalPackage.overrideAttrs (oa: {
           34|     paths =
             |     ^
           35|       oa.paths

       error: attribute 'paths' missing

       at /nix/store/1ann4n8aba6qd5sr93gx1yjj7pinfk2d-source/wrappers/standalone.nix:35:7:

           34|     paths =
           35|       oa.paths
             |       ^
           36|       ++ (pkgs.lib.optional config.enableMan self.packages.${pkgs.system}.man-docs);
```

cc @teto

andrevmatos:
My workaround to make the `oa.paths` stop erroring, is to overlay `wrapNeovimUnstable`, used to create `finalPackage` before standalone:
```nix
final: prev: {
  wrapNeovimUnstable = pkg: args: (prev.wrapNeovimUnstable pkg args).overrideAttrs (_: {paths = [];});
}
```

GaetanLepage:
Thank you for having reported this.
It is very annoying indeed. Do you think a similar solution would be made directly into the `nixvim` code ?
Do you have any suggestion ?

teto:
[...]

PR #717, opened 2023-11-19
bug when paths does not exist build fails

ivandimitrov8080:
When I try to build standalone I get this error:

`error: attribute 'paths' missing`

That change fixes the error, but I'm not totally sure how it works.


ivandimitrov8080:
Why am I getting this error in the first place then? It wasn't like that yesterday.

GaetanLepage:
Yes we are aware of this problem.
You can learn more in #707.
We will try to deploy a fix quickly.

PR #718, opened 2023-11-20
plugins/lsp: add pest-language-server

NickHu:


NickHu:
Done

plugins/lsp/language-servers/default.nix
```diff
@@ -421,6 +421,11 @@ with lib; let
         '';
       };
     }
+    {
+      name = "pest_ls";
+      description = "Enable pest-language-server, for pest";
+      package = pkgs.pest-ide-tools;
+    }
```
GaetanLepage, line 428:
```suggestion
    {
      name = "pest-ls";
      description = "Enable pest-language-server, for pest";
      package = pkgs.pest-ide-tools;
      serverName = "pest_ls";
    }
```

We would adopt "nix-style" in this case.

PR #720, opened 2023-11-21
ci/check: Run checks for all systems (not only x86_64-linux)

GaetanLepage:
Right now, the CI only performs checks for x86_64-linux.
However, bugs can occur on other platforms without impacting x86 linux.
Hence, we have nothing to lose by running checks for all systems.


traxys (approved):

traxys:
Seems like CI is not happy with darwin

PR #729, opened 2023-11-23
ktlint: enable formatter and diagnostics

MikaelFangel:
### Changes
 * Enable ktlint as a none-ls formatter
 * Enable ktlint as a none-ls diagnostics tool
 * Add tests for both cases

### Checks
 - [x] Alejandra style
 - [x] Flake check


GaetanLepage (approved):
LGTM
Thanks !

PR #736, opened 2023-11-23
wrapper: Use symlink join to install man pages for standalone

traxys:
Patching the derivation forces a rebuild of neovim, which is undesirable


GaetanLepage (approved):
Amazing ! Thanks :)

PR #737, opened 2023-11-23
plugin/rust-tools: Correctly take into account on_attach

traxys:
Nixvim uses a common function for the on attach function of lsp, __lspOnAttach. With the current implementation the function was not taken into account if no server options were present, this fixes this issue.


GaetanLepage (approved):
LGTM

PR #744, opened 2023-11-24
plugins/lsp: add emmet_ls language server

GaetanLepage:
Add support for [emmet-ls](https://github.com/aca/emmet-ls).

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -169,6 +169,11 @@ with lib; let
       package = pkgs.elmPackages.elm-language-server;
     }
     {
+      name = "emmet_ls";
+      description = "Enable emmet_ls, emmet support based on LSP";
+      package = pkgs.emmet-ls;
+    }
+    {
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -168,6 +168,11 @@ with lib; let
       description = "Enable elmls, for Elm.";
       package = pkgs.elmPackages.elm-language-server;
     }
+    {
+      name = "emmet-ls";
```
traxys, line 172: Isn't it prefereable to have `emmet_ls` like upstream nvim-lspconfig? It's easier to adapt a normal neovim config

traxys (changes requested):

plugins/lsp/language-servers/default.nix
GaetanLepage: I am ok with that, but it is not the convention we have sued until now.

traxys (approved):

PR #751, opened 2023-11-27
Fix manpage generation

Iggle:
NixOS/nixpkgs#269942 changed the signature of the NixOS ManpageConverter to require a header and a footer.

This PR fixes the incompatibility and refactors the NixvimManpageConverter to use those options


Iggle:
Apparently, `self.render()` can't be used before calling `super().__init__()`.

That said, looking at the previous implementation, it seems the `render` lines are just copy and pasted from the nixos-render-docs implementation and probably aren't necessary - would it be fine to replace them with raw strings?

traxys:
I'm pretty sure the only differences between the code here and the upstream implementation is to edit the footer/header, so if there is an option to do that we should use this instead of the ugly workaround we have right now

Iggle:
In that case, I removed the nixvim-render-docs along with the flake output to use the new header and footer parameters of nixos-render-docs.

That said, the new nixos-render docs has not reached nixos-unstable yet and is only in nixpkgs-unstable and nixos-23.11 so far. Until it reaches nixos-unstable I've marked this PR as a draft.

GaetanLepage:
@Iggle the new nixos-render docs has now reached `nixos-unstable`. Would your fix be working if I bump nixpkgs now ?

Iggle:
It's a bit hard for me to check, as cbfmt is currently [broken on darwin](https://hydra.nixos.org/job/nixpkgs/nixpkgs-23.11-darwin/cbfmt.aarch64-darwin) so the checks fail, but I cherry-picked the commit from your PR. The man-docs package builds fine.
By the way, I noticed that the nixvim man-pages are the only ones not gzipped, meaning the compress-man-pages.sh setup hook is not being executed. Any idea how to incorporate this in the runCommand builder?

GaetanLepage:
I cherry picked your commit and added it to #760.
The CI seems to be happy, so I think that we will merge this.

Iggle:
Thank you!
Adding `compressManPages $out` at the end of the build script will apparently work to compress the Man Page as well.

PR #756, opened 2023-11-28
plugins/none-ls: add enableLspFormat option

### LINKED ISSUE #755: Lsp-format doesn't work easily with none-ls
To make none-ls work with lsp-format you need to add a custom onAttach:

```
none-ls.onAttach = "require('lsp-format').on_attach";
```

This could be automatic depending on if lsp-format is enabled, or an extra option on none-ls.

GaetanLepage:
This PR adds an option in `none-ls` to configure the use of `lsp-format` with it.

I also changed the behavior of `mkRaw` to ensure that `mkRaw ""` gives `null` instead of `""`.

Fixes #755

### DIFF
--- a/plugins/none-ls/default.nix
+++ b/plugins/none-ls/default.nix
@@ -28,6 +28,15 @@ in {
         description = "Plugin to use for none-ls";
       };
 
+      enableLspFormat = mkOption {
+        type = types.bool;
+        default = config.plugins.lsp-format.enable;
+        description = ''
+          Automatically enable the `lsp-format` plugin and configure `none-ls` accordingly.
+        '';
+        example = true;
+      };
+
       border = helpers.defaultNullOpts.mkBorder "null" "`:NullLsInfo` UI window." ''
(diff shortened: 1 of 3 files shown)


GaetanLepage:
I don't know if this integration should be enabled automatically if the user has `lsp-format` enabled...

plugins/none-ls/default.nix
```diff
@@ -28,6 +28,15 @@ in {
         description = "Plugin to use for none-ls";
       };
 
+      enableLspFormat = mkOption {
+        type = types.bool;
+        default = false;
```
traxys, line 33: we could maybe set the default to `plugins.lsp-format.enable` indeed?

traxys (changes requested):

plugins/none-ls/default.nix
GaetanLepage: Yes this makes sense.

GaetanLepage: Done.

traxys (approved):

PR #762, opened 2023-11-29
plugins/formatter-nvim: init

VoreckLukas:
Add support for [formatter-nvim](https://github.com/mhartington/formatter.nvim/tree/master)

TODO:
- Look for missing formatter packages (help apreciated)
- Allow user to choose custom formatter packages
- allow disabling auto install of formatters


VoreckLukas:
The only thing still missing are some packages. I couldn't find packages for all supported formatters, in part because some seem to be old and unmaintained

PR #769, opened 2023-12-02
Support more options for icon in lualine

adalessa:
From lualine documentation the icons can be configure more than just an string 

https://github.com/nvim-lualine/lualine.nvim#general-component-options

example 
```lua
{'branch', icon = {'', align='right', color={fg='green'}}}
```
Currently is defined as string which cover for most of the cases but if you want to set a color or more is not possible. I think that adding support for raw type makes more extensible.

I think also setting an object is possible but I lack the knowledge to do so, with this I was able to extend my configuration as I needed.

Let me know what do you think.

Kind regards, thanks for maintaining the project.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -36,7 +36,23 @@ with lib; let
               Enables the display of icons alongside the component.
             '';
 
-            icon = helpers.mkNullOrOption (with types; either str (attrsOf anything)) ''
+            icon = helpers.mkNullOrOption
+              (
+                with types;
+                  either
+                  str
+                  (submodule {
+                    freeformType = attrsOf anything;
+                    
+                    options = {
+                      icon = mkOption {
+                        type = str;
+                        description = "Icon character.";
+                      };
+                    };
+                  })
+              )
+              "Defines the icon to be displayed in front of the component.";
               Defines the icon to be displayed in front of the component.
             '';
 
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -36,7 +36,7 @@ with lib; let
               Enables the display of icons alongside the component.
             '';
 
-            icon = helpers.mkNullOrOption types.str ''
+            icon = helpers.mkNullOrOption (types.either types.str helpers.rawType) ''
               Defines the icon to be displayed in front of the component.
             '';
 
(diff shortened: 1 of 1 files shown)


plugins/statuslines/lualine.nix
```diff
@@ -36,7 +36,7 @@ with lib; let
               Enables the display of icons alongside the component.
             '';
 
-            icon = helpers.mkNullOrOption types.str ''
+            icon = helpers.mkNullOrOption (types.either types.str helpers.rawType) ''
```
GaetanLepage, line 39:
```suggestion
            icon = helpers.mkNullOrOption (with types; either str (attrsOf anything)) ''
```

GaetanLepage (changes requested):
Good remark, we can definitely do this.
See my proposition for what type to set :)

plugins/statuslines/lualine.nix
adalessa: Thanks for showing me the change.
I am trying to test locally before pushing the change.
I am having a problem on the generation. 

I think something else needs to be define maybe here https://github.com/nix-community/nixvim/blob/main/plugins/statuslines/lualine.nix#L214. in order to handle the unkeyed value for the icon. 
I've tried several things with no success.  maybe you can point me to some example to use as base. 
Thanks.

adalessa:
@GaetanLepage After several tries and many errors I was able to find the solution to properly map the icon property, and from my tests works as expected. Please let me know what you think.

plugins/statuslines/lualine.nix
```diff
@@ -36,7 +36,7 @@ with lib; let
               Enables the display of icons alongside the component.
             '';
 
-            icon = helpers.mkNullOrOption types.str ''
+            icon = helpers.mkNullOrOption (with types; either str (attrsOf anything)) ''
```
GaetanLepage, line 39:
```suggestion
            icon = helpers.mkNullOrOption
              (
                with types;
                  either
                  str
                  (submodule {
                    freeformType = attrsOf anything;
                    
                    options = {
                      icon = mkOption {
                        type = str;
                        description = "Icon character.";
                      };
                    };
                  })
              )
              "Defines the icon to be displayed in front of the component.";
```

GaetanLepage: Good job ! I suggest this for the option definition.
It ensures that the `icon.icon` option is set with a string.

adalessa: That's for sure better, I have applied it, need to remove the extra line, but after my testing works as expected.

GaetanLepage (approved):
Perfect ! Thanks :)

PR #785, opened 2023-12-06
Replace flake-utils with flake-parts

GaetanLepage:
flake-parts seems to be considered as the "clean" and less "hacky" solution to flake bootstrapping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,27 +141,38 @@ can use the following:
   description = "A very basic flake";
 
   inputs.nixvim.url = "github:nix-community/nixvim";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = {
     self,
-    nixpkgs,
     nixvim,
-    flake-utils,
-  }: let
+    flake-parts,
+  } @ inputs: let
     config = {
       colorschemes.gruvbox.enable = true;
     };
   in
-    flake-utils.lib.eachDefaultSystem (system: let
-	  nixvim' = nixvim.legacyPackages."${system}";
-      nvim = nixvim'.makeNixvim config;
-    in {
-      packages = {
-        inherit nvim;
-        default = nvim;
+    flake-parts.lib.mkFlake {inherit inputs;} {
+      systems = [
+        "aarch64-darwin"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "x86_64-linux"
+      ];
+
+      perSystem = {
+        pkgs,
+        system,
+        ...
+      }: let
+        nixvim' = nixvim.legacyPackages."${system}";
+        nvim = nixvim'.makeNixvim config;
+      in {
+        packages = {
+          inherit nvim;
+          default = nvim;
+        };
       };
-    });
+    };
 }
 ```
 </details>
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -43,7 +43,7 @@ If you have any question, please use the [discussions page](https://github.com/n
 > If you want to use NixVim with nixpkgs 23.05 you should use the `nixos-23.05` branch.
 
 ### Without flakes
-NixVim now ships with `flake-compat`, which makes it usable from any system.
```
traxys, line 46: I don't think that changed? I think flake-compat is the `default.nix` at the root of the projet

README.md
```diff
@@ -141,27 +141,38 @@ can use the following:
...
   inputs.nixvim.url = "github:nix-community/nixvim";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = {
     self,
-    nixpkgs,
     nixvim,
-    flake-utils,
-  }: let
+    flake-parts,
```
traxys, line 148: This is missing the flake-parts input I think?

flake.nix
```diff
@@ -2,7 +2,6 @@
   description = "A neovim configuration system for NixOS";
 
   inputs = {
-    flake-utils.url = "github:numtide/flake-utils";
```
traxys, line 5: This is missing the flake-parts input I think?

templates/simple/flake.nix
```diff
@@ -3,42 +3,52 @@
...
     nixvim.url = "github:nix-community/nixvim";
-    flake-utils.url = "github:numtide/flake-utils";
   };
 
   outputs = {
     self,
-    nixpkgs,
     nixvim,
-    flake-utils,
+    flake-parts,
```
traxys, line 11: Missing flake-parts input?

traxys (changes requested):

README.md
GaetanLepage: Actually, you don't need it thanks to the magic flake registry: https://github.com/NixOS/flake-registry
The examples provided in the flake-parts repo itself don't put it in the inputs.

traxys: Well I don't think it's good to rely on the flake registry, because we don't control the versions. 
image
I don't agree that the example does not provide it.

traxys: I think it's different for a "library" where we need to tightly control the versions of our inputs to avoid random breakage on a machine & a "end-user" flake, where taking what's on the machine is ok.

README.md
GaetanLepage: Indeed. Roll back done.

flake-modules/modules.nix
```diff
@@ -0,0 +1,47 @@
...
+
+      nixvimExtraArgsModule = rec {
+        _file = ./flake.nix;
+        key = _file;
+        config = {
+          _module.args = {
+            pkgs = pkgs.lib.mkForce pkgs;
+            inherit (pkgs) lib;
+            helpers = import ../lib/helpers.nix {inherit (pkgs) lib;};
+            # TODO: Not sure why the modules need to access the whole flake inputs...
```
traxys, line 25: Could be if we have certain packages as flake inputs. I don't think this is really mandatory.

traxys (approved):

flake-modules/modules.nix
GaetanLepage: I will make some tests...

PR #793, opened 2023-12-08
plugins/fzf-lua: init + test

budimanjojo:


budimanjojo:
I don't know why the test fails, I tested locally without my addition and it fails too.

plugins/utils/fzf-lua.nix
```diff
@@ -0,0 +1,39 @@
...
+in {
+  options.plugins.fzf-lua =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "fzf-lua";
+
+      package = helpers.mkPackageOption "fzf-lua" pkgs.vimPlugins.fzf-lua;
+
+      profile = mkOption {
+        type = with types; oneOf [(enum ["default" "fzf-native" "fzf-tmux" "fzf-vim" "max-perf" "telescope" "skim"])];
```
GaetanLepage, line 19:
```suggestion
        type = types.enum ["default" "fzf-native" "fzf-tmux" "fzf-vim" "max-perf" "telescope" "skim"];
```
You don't need `oneOf` here.

budimanjojo:
@GaetanLepage Done!

budimanjojo:
@GaetanLepage is there any other issue that I need to fix?

GaetanLepage:
Hey, sorry for taking so long.
Everything looks nice for now. However, I would encourage you to wrap the plugin options like we do for the other plugins.

The simpler way would be to limit the scope to the ["global options"](https://github.com/ibhagwan/fzf-lua#default-options) that would end up in the `setupOptions` attrs.

If you want to go a bit further, you could create a `keymaps` option that allows the user to create mappings easily (like we do for telescope or harpoon for instance).
You could even add options for the keymaps to be configurable.

I suggest to limit the scope of this PR to the default options though.

budimanjojo:
Thanks! I'll try to add `keymaps` and `keymapsSilent` option for now. I don't think I understand the "wrap the plugin options like we do for the other plugins" part. Do u mean I need to add more options instead of making users to use `extraOptions` for everything else?

GaetanLepage:
> Do u mean I need to add more options instead of making users to use `extraOptions` for everything else?

Yes exactly. It is a bit of a pain to do, but it is one of the added value of nixvim ;)

budimanjojo:
@GaetanLepage I added `keymaps` options. I don't know if my nix knowledge is good enough to create all those setup options without making the code looks messy. Maybe it's just me but I think people should just be good with changing the provided `profile` (or theme) and set those options per keymap like I provided in the example.

Maybe you can accept this PR and maybe if somebody else want to improve this plugin then he/she/they can make another PR?

budimanjojo:
I also agree that having the setup options added in can be some kind of type checking at build time in case the user provided a wrong options it will fail to build resulting in the user can fix that up before having a broken neovim config. But I'm just not good enough to make it robust. I think the better way to go is to have all the available options in a variable that can also be used for the `actionOpts` in the `keymaps` in my last commit.

But, there are a lot of options that are action name specific, i.e `cwd` is only available in `files` action and this makes my brain hurts just by thinking about it :sweat_smile:

plugins/utils/fzf-lua.nix
```diff
@@ -0,0 +1,89 @@
...
+        type = types.bool;
+        description = "Whether Fzf-Lua keymaps should be silent";
+        default = false;
+      };
+    };
+
+  config = let
+    options =
+      {
+        __unkeyedProfile = cfg.profile;
```
GaetanLepage, line 51:
```suggestion
        __unkeyed_profile = cfg.profile;
```

plugins/utils/fzf-lua.nix
```diff
@@ -0,0 +1,89 @@
...
+            lua = true;
+
+            options =
+              {
+                silent = cfg.keymapsSilent;
+              }
+              // actionProps;
+          }
+        )
+        cfg.keymaps;
```
GaetanLepage, line 83:
```suggestion
          key: action: let
            actionStr =
              if isString action
              then "${action}()"
              else "${action.action}(${helpers.toLuaObject action.actionOpts})";
            actionProps =
              if isString action
              then {}
              else filterAttrs (n: v: n != "action" && n != "actionOpts") action;
          in {
            mode = "n";
            inherit key;
            action = "function() require('fzf-lua').${actionStr} end";
            lua = true;

            options =
              {
                silent = cfg.keymapsSilent;
              }
              // actionProps;
          }
        )
        cfg.keymaps;
```

Maybe, we can remove the `keymapsSilent` option as the user can set `actionProps.silent = true`.

plugins/utils/fzf-lua.nix
```diff
@@ -0,0 +1,89 @@
...
+    // {
+      enable = mkEnableOption "fzf-lua";
+
+      package = helpers.mkPackageOption "fzf-lua" pkgs.vimPlugins.fzf-lua;
+
+      profile = mkOption {
+        type = with types; (enum ["default" "fzf-native" "fzf-tmux" "fzf-vim" "max-perf" "telescope" "skim"]);
+        default = "default";
+        description = "Preconfigured profile to use";
+      };
```
GaetanLepage, line 22: This should be made optional (i.e. `helpers.defaultNullOpts.mkEnumFirstDefault`).
We will then only call the `setup` if the user has explicitly provided a profile and/or options.

GaetanLepage (changes requested):
I think that what you have done until now is good.

I would still want to package the options.
If you allow me to do so, I can complete this work by force-pushing to your branch.
Does that sound good to you ?

budimanjojo:
@GaetanLepage I have commited the requested changes. Yes feel free to push to the branch! Thank you!

GaetanLepage:
This is finally ready ! Sorry @budimanjojo for taking that long...

traxys (approved):

GaetanLepage (approved):

budimanjojo:
Thank you!

PR #796, opened 2023-12-08
fix: typo in mini module example

hmajid2301:
If you use the example, nvim will throw an error that surrounds does not exist. `surround` is also used in the test.


GaetanLepage (approved):
Cool, thanks !

PR #805, opened 2023-12-12
*WIP* Add Octo

alisonjenkins:
Hey @GaetanLepage ,

This is still very much WIP but I thought I would check with you if I am heading in the right direction with this.

Does this look sane with what I have done so far?

Thanks,

Alison

### DIFF
--- a/plugins/git/octo.nix
+++ b/plugins/git/octo.nix
@@ -1,0 +1,226 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.octo;
+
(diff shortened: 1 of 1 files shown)


GaetanLepage:
It looks pretty good !
One thing though, you want to use the `helpers.defaultNullOpts.mkBool` / `mkStr` / `mkUnsignedInt` / `mkInt` or, `mkNullable <TYPE>`.
Have a look at other packages to see how we use those.

traxys (commented):
You should define the nix options as `camelCase` (and then in the `setupOptions` convert the camel case name to snake case).

One of these days we should look into [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)

PR #816, opened 2023-12-15
plugins/wtf: init

### LINKED ISSUE #771: Add support for `wtf.nvim`
**Plugin**: `wtf.nvim`
**Plugin repository**: https://github.com/piersolenski/wtf.nvim
**Is the plugin on nixpkgs**: `false`

GaetanLepage:
Add support for the [wtf.nvim](https://github.com/piersolenski/wtf.nvim) plugin.

Fixes #771

### DIFF
--- a/plugins/lsp/wtf.nix
+++ b/plugins/lsp/wtf.nix
@@ -1,0 +1,124 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.wtf;
+
+  defaultKeymaps = {
+    ai = {
+      key = "gw";
+      mode = ["n" "x"];
+      action = "require('wtf').ai";
+      lua = true;
+    };
+
+    search = {
+      key = "gW";
+      mode = "n";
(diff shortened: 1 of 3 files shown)


plugins/lsp/wtf.nix
```diff
@@ -0,0 +1,122 @@
...
+              )
+              "Keymap for the ${action} action."
+          )
+          defaultKeymaps;
+
+        popupType = helpers.defaultNullOpts.mkEnum ["popup" "horizontal" "vertical"] "popup" ''
+          Default AI popup type.
+        '';
+
+        openaiApiKey = helpers.mkNullOrOption types.str "An alternative way to set your API key";
```
traxys, line 54: It would be nice to be either str or raw, to allow using an env var if you don't want a private key in the nix store

traxys (changes requested):

plugins/lsp/wtf.nix
GaetanLepage: Good idea ! Done.

traxys (approved):

PR #837, opened 2023-12-22
helpers: Allow to set raw lua code in more places

traxys:
With this change a user should be able to set raw lua code in more places than before.

The `mkStr` helper was _not_ changed as a large number of places in nixvim are using this helper to directly implement raw lua code.

Also note that types of the form `oneOf [submodule <something else>]` are not well rendered in the documentation, and should be avoided (this includes `listOf [submodule <something else>]`, so the helpers should be avoided in this case)


GaetanLepage (approved):
Good change.

PR #838, opened 2023-12-22
plugins/startup: init

### LINKED ISSUE #791: startup.nvim
**Plugin**: startup.nvim
**Plugin repository**: [URL to the repository of the plugin](https://github.com/startup-nvim/startup.nvim)
**Is the plugin on nixpkgs**: No

## Extra information
Customizable startup screen for neovim, similar to startify. 
Depends on plenary.nvim, and should depend on telescope.nvim for default setup.
I don't really understand fully how nixvim works, so I'm not sure how to add this myself, but if it would be simple for someone to implement this, I'd be grateful!

GaetanLepage:
Add support for [startup.nvim](https://github.com/startup-nvim/startup.nvim).

Fixes #791

### DIFF
--- a/plugins/utils/startup.nix
+++ b/plugins/utils/startup.nix
@@ -1,0 +1,350 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.startup;
+in {
+  meta.maintainers = [maintainers.GaetanLepage];
+
+  options.plugins.startup =
+    helpers.extraOptionsOptions
+    // {
+      enable = mkEnableOption "startup.nvim";
+
+      package = helpers.mkPackageOption "startup.nvim" pkgs.vimPlugins.startup-nvim;
+
+      theme = helpers.defaultNullOpts.mkStr "dashboard" ''
+        Use a pre-defined theme.
+      '';
+
+      sections = let
+        sectionType = with types;
+          submodule {
+            options = {
+              type = helpers.defaultNullOpts.mkEnumFirstDefault ["text" "mapping" "oldfiles"] ''
+                - "text" -> text that will be displayed
+                - "mapping" -> create mappings for commands that can be used.
+                  use `mappings.executeCommand` on the commands to execute.
+                - "oldfiles" -> display oldfiles (can be opened with `mappings.openFile`/`openFileSplit`)
+              '';
+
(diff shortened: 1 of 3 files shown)


plugins/utils/startup.nix
```diff
@@ -0,0 +1,350 @@
...
+      {
+        assertion = (cfg.options.paddings == null) || (length cfg.options.paddings) == numSections;
+        message = ''
+          Nixvim (plugins.startup): Make sure that `plugins.startup.options.paddings` has the same
+          number of elements as there are sections.
+        '';
+      }
+      {
+        assertion =
+          ((length cfg.parts) == numSections)
```
traxys, line 276: I think this should be a `<` you could define more sections than you use in parts

traxys (changes requested):

plugins/utils/startup.nix
GaetanLepage: I don't think so. In the README, it is mentioned "all sections in order" for the `parts` options.

traxys: I think this means "in order you want them to be shown"

traxys: Lookig at the code it seems that way:
```
    for _, part in ipairs(parts) do
        utils.empty(settings.options.paddings[padding_nr])
        padding_nr = padding_nr + 1
        current_section = part
        local options = settings[part]
        options = utils.validate_settings(options)
```

GaetanLepage: Ok, I switched to `<` and changed the assertion message.

plugins/utils/startup.nix
```diff
@@ -0,0 +1,350 @@
...
+          number of elements as there are sections.
+        '';
+      }
+      {
+        assertion =
+          ((length cfg.parts) <= numSections)
+          && (
+            all
+            (sectionName: elem sectionName cfg.parts)
+            sectionNames
```
traxys, line 280: Sorry this one seems off too, I think you need to check `all (part: builtins.hasAttr part cfg.sections) cfg.parts` in order to check that all parts are defined.

And the message should be `You should not have parts that are not defined in 'plugins.startup.sections'`, I think that condition also avoids the need to check the length

traxys (changes requested):

plugins/utils/startup.nix
GaetanLepage: Done !

traxys (approved):

PR #845, opened 2023-12-27
plugins/rust-tools: Allow null serverPackage

ErrorNoInternet:
Some users might want to use rust-analyzer beta/nightly from their own nix develop shell.


traxys (changes requested):
Please mention in the option description that setting it to null will not install rust analyzer

traxys (approved):

PR #849, opened 2023-12-28
plugins/comment-nvim: add {pre,post}Hook options

hoppla20:
Adding support for [comment-nvim's hooks](https://github.com/numToStr/Comment.nvim?tab=readme-ov-file#-hooks). These are especially important for integrating comment-nvim with ts-context-commentstring (see linked comment-nvim docs).

### DIFF
--- a/tests/test-sources/plugins/statuslines/lualine.nix
+++ b/tests/test-sources/plugins/statuslines/lualine.nix
@@ -59,6 +59,7 @@
           # you can specify only the sections you want to change
           {
             name = "filename";
+            icon = "-";
             extraConfig = {
               newfile_status = true;
               path = 1;
@@ -75,6 +76,10 @@
         lualine_a = [
           {
             name = "buffers";
+            icon = {
+              icon = "X";
+              align = "right";
+            };
             extraConfig.mode = 4;
             extraConfig.filetype_names = {
               TelescopePrompt = "Telescope";


hoppla20:
Hmm okay this is not working. I tried the following:

```nix
{
  plugins.ts-context-commentstring = {
    enable = true;
    extraOptions.enable_autocmd = false;
  };
  plugins.comment-nvim.preHook = "require('ts_context_commentstring.integrations.comment_nvim').create_pre_hook()";
}
```

This should match [ts-context-commentstring documentation](https://github.com/JoosepAlviste/nvim-ts-context-commentstring/wiki/Integrations#commentnvim).

The resulting init.lua looks like this:

```lua
do
  local nixvim_globals = {
   ...
    ["skip_ts_context_commentstring_module"] = true,
    ...
  }

  for k, v in pairs(nixvim_globals) do
    vim.g[k] = v
  end
end


...

require("Comment").setup({
   ["pre_hook"] = "require('ts_context_commentstring.integrations.comment_nvim').create_pre_hook()",
})

...

require("ts_context_commentstring").setup({ ["enable_autocmd"] = false })

...
```

I'm not entirely sure, why this is not working, I just started switching to neovim, so I'm not so familiar how to debug these problems. The following error is printed:

```
[Comment.nvim] Invalid commentstring for nix! Read `:h commentstring` for help.
```

However, when I look at the commentstring it has the following value: `commentstring=# %s`. This should be a valid commentstring, so I am a bit confused. When I use ts-context-commentstring without the hook integration, it works. However, this is suboptimal, because comment-nvim has [some problems without ts-context-commentstring](https://github.com/numToStr/Comment.nvim?tab=readme-ov-file#-treesitter).

Do you maybe have an idea why this is not working?

plugins/utils/comment-nvim.nix
```diff
@@ -29,6 +29,8 @@ in {
         description = "Lines to be ignored while comment/uncomment";
         default = null;
       };
+      preHook = helpers.defaultNullOpts.mkStr "" "Lua function called before (un)comment.";
+      postHook = helpers.defaultNullOpts.mkStr "" "Lua function called after (un)comment.";
```
traxys, line 33: I'd suggest using `helpers.defaultNullOpts.mkNullable types.str "null" "...."` as a type instead, to denote the null value.

plugins/utils/comment-nvim.nix
```diff
@@ -94,6 +96,8 @@ in {
   config = let
     setupOptions = {
       inherit (cfg) padding sticky ignore toggler opleader mappings;
+      pre_hook = cfg.preHook;
+      post_hook = cfg.postHook;
```
traxys, line 100: you need `pre_hook = helpers.mkRaw cfg.preHook` to make it as lua code (same for the other line)

traxys (changes requested):

plugins/utils/comment-nvim.nix
hoppla20: Thank you, this worked!

hoppla20:
Tested with the above config snippet, should be working now. Maybe in the future one could add a `plugins.ts-context-commentstring.integrations` options, which configures ts-context-commentstring and the corresponding comment plugin (comment-nvim, commentary, ...).

traxys (changes requested):

PR #855, opened 2023-12-29
plugins: Introduce helpers.defaultNullOpts.mkLuaFn

traxys:
This allows to avoid calling `mkRaw` on lua functions, as they will get applied automatically.

This could also help in the future to refactor the use of Lua code to make it more user-friendly.


GaetanLepage (approved):
Very good idea !

PR #857, opened 2023-12-30
helpers/mkLuaFn: refactor description implementation

GaetanLepage:
This allows to pass an empty default.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -144,27 +144,26 @@ with lib; rec {
       lib.mkOption {
         type = lib.types.nullOr lib.types.str;
         default = null;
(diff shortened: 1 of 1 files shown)


traxys:
Why would we want empty defaults instead of null?

GaetanLepage:
> Why would we want empty defaults instead of a lua nil

I meant an empty default string to disable the `defaultDesc` block.
For instance, if the actual default value of the plugin needs a weird description.

traxys:
I think #866 is the better approach

GaetanLepage:
> I think #866 is the better approach

Indeed, closing.

PR #860, opened 2023-12-30
plugins/lsp: add dhall language server

gwg313:
add [dhall](https://hackage.haskell.org/package/dhall-lsp-server) language server.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -159,6 +159,11 @@ with lib; let
       package = pkgs.deno;
     }
     {
+      name = "dhall_lsp_server";
+      description = "Enable dhall-lsp-server, for Dhall.";
+      package = pkgs.dhall-lsp-server;
+    }
+    {
       name = "digestif";
       description = "Enable digestif, for LaTeX";
       # luaPackages.digestif is currently broken, using lua54Packages instead
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -159,9 +159,9 @@ with lib; let
       package = pkgs.deno;
     }
     {
-      name = "dhall_lsp_server";
+      name = "dhall-lsp-server";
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -158,6 +158,11 @@ with lib; let
       description = "Enable denols, for Deno";
       package = pkgs.deno;
     }
+    {
+      name = "dhall_lsp_server";
+      description = "Enable dhall-lsp-server, for Dhall.";
+      package = pkgs.dhall-lsp-server;
+    }
```
GaetanLepage, line 165:
```suggestion
    {
      name = "dhall-lsp-server";
      description = "Enable dhall-lsp-server, for Dhall.";
      serverName = "dhall_lsp_server";
    }
```
We use the `foo-bar` naming convention for language servers.
Here, we then need to set `serverName` to the real server name (`dhall_lsp_server`).
The package name though is set by default to the server name.

tests/test-sources/plugins/lsp/_lsp.nix
```diff
@@ -88,6 +88,7 @@
         cssls.enable = true;
         dartls.enable = true;
         denols.enable = true;
+        dhall_lsp_server.enable = true;
```
GaetanLepage, line 91:
```suggestion
        dhall-lsp-server.enable = true;
```

plugins/lsp/language-servers/default.nix
```diff
@@ -158,6 +158,11 @@ with lib; let
       description = "Enable denols, for Deno";
       package = pkgs.deno;
     }
+    {
+      name = "dhall-lsp-server";
+      description = "Enable dhall-lsp-server, for Dhall.";
```
GaetanLepage, line 163:
```suggestion
      description = "Enable dhall-lsp-server, for Dhall";
```

GaetanLepage (approved):
LGTM !
Thanks for your contribution :)

PR #866, opened 2023-12-31
helpers: Add mkLua,mkLuaFn (defaultNullOpts or not)

traxys:
This makes it such that defaultNullOpts.mkStr never represents a lua string, easing documenting where a string is a lua string vs lua code.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -119,6 +119,27 @@ with lib; rec {
   mkCompositeOption = desc: options:
     mkNullOrOption (types.submodule {inherit options;}) desc;
 
+  mkNullOrLua = desc:
+    lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = desc;
+      apply = mkRaw;
+    };
+
+  mkNullOrLuaFn = desc: let
+    fnDesc = "(lua function)";
+  in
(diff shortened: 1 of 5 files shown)


lib/helpers.nix
```diff
@@ -140,33 +161,34 @@ with lib; rec {
...
-            (lua function)
+    mkLua = default: desc: let
+      defaultDesc = "default: `${default}`";
+    in
+      mkNullOrLua
+      (
+        if desc == ""
+        then defaultDesc
+        else ''
+          ${desc}
```
GaetanLepage, line 172:
```suggestion
    mkLua = default: desc: let
      defaultDesc = "default: `${default}`";
    in
      mkNullOrLua
      (
        (optionalString (desc != "") ''
          ${desc}
          
        '')
        + ''
          default: `${default}`
        '';
```

This looks more readable to me. What do you think ?

traxys: No real preference

GaetanLepage (approved):
This looks good to me !
Are you sure that there are no more occurrences of `mkStr`  (or `mkNullOrOption types.str`) options across the codebase ?

traxys:
I checked all the mkRaw places, and it should be good. There are a few other places, but it's a bit different (in listOf or attrsOf for example)

PR #871, opened 2024-01-01
helpers: move rawType to helpers.nixvimTypes

traxys:
This is done in preparation of adding new (lua) types to help the documentation.


GaetanLepage (approved):
I like that a lot !

PR #872, opened 2024-01-01
helpers: Add types for lua function & lua code

traxys:
This allows to display `lua function string` and `lua code string` in the documentation for related options.


GaetanLepage (approved):

GaetanLepage:
Let's merge #874 first, and then rebase.

PR #876, opened 2024-01-01
helpers: Add option helpers for types that are either a lua str or a value

traxys:
Those types needed the following code in apply:

if builtins.isString value
then mkRaw value
else value

This commit avoids this boilerplate, and clarifies that the `str` is lua code in the documentation.


GaetanLepage (approved):
Cool !

PR #887, opened 2024-01-03
 misc/wrapRc: ignore user "after" directory when wrapRc is true 

GaetanLepage:
This seems to be necessary as well...


GaetanLepage:
I got confirmation that this patch indeed prevents neovim to load user's files in `~/.config/nvim/after`

traxys (approved):

PR #901, opened 2024-01-06
tests: Add a test for the home-manager module

traxys:


tests/modules/check.sh
```diff
@@ -0,0 +1,30 @@
...
+	if ! home-manager --no-write-lock-file build --flake ${self}/home-manager#nixvim-${system}; then
+		failures+=("home-manager-${system}")
+		ok=false
+	fi
+done
+
+if [[ $ok = false ]]; then
+	echo -e "\nFAILURES: ${failures[@]}"
+	exit 1
+fi
```
GaetanLepage, line 30: Can't we make a derivation out of this ?

traxys: I tried, but was not able because home-manager was not able to find the nix profile directory for the user, which seems coherent as I don't think that in derivations a full nix environement exists

GaetanLepage:
I think that you have to rebase on `main`

GaetanLepage (approved):
Amazing !

PR #902, opened 2024-01-06
flake/templates: propagate the template's checks to the main flake's

GaetanLepage:


templates/simple/flake.nix
```diff
@@ -35,7 +34,7 @@
           module = config;
           # You can use `extraSpecialArgs` to pass additional arguments to your module files
           extraSpecialArgs = {
-            inherit self;
+            # inherit (inputs) foo;
```
traxys, line 37: Why leave this as a comment? It's to show how to pass something?

GaetanLepage: Yes, it is just to show that it can be done.

traxys (approved):

PR #905, opened 2024-01-06
plugins/lsp: add templ language server

rsHalford:
Adds lsp support for [templ](https://templ.guide), the HTML templating language for Go.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -546,6 +546,11 @@ with lib; let
       package = pkgs.taplo;
     }
     {
+      name = "templ";
+      description = "Enable the templ language server for the templ HTML templating language";
+      package = pkgs.templ;
+    }
+    {
       name = "terraformls";
       description = "Enable terraform-ls, for terraform";
       package = pkgs.terraform-ls;
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -548,7 +548,6 @@ with lib; let
     {
       name = "templ";
       description = "Enable the templ language server for the templ HTML templating language";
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -545,6 +545,11 @@ with lib; let
       description = "Enable taplo, for TOML";
       package = pkgs.taplo;
     }
+    {
+      name = "templ";
+      description = "Enable the templ language server for the templ HTML templating language";
+      package = pkgs.templ;
+    }
```
GaetanLepage, line 552:
```suggestion
      description = "Enable the templ language server for the templ HTML templating language";
    }
```
As the package has the same name as the LS, you don't need to specify it.

GaetanLepage (approved):
LGTM ! Thanks for your contribution :)

psyos10:
Hey, new to nixos, may be my problem. I have been getting nixos-rebuild issues that point to this addition in nixvim, **`error: attribute 'templ' missing` at plugins/lsp/helpers.nix:6:15** , my config worked fine before this.

traxys:
There could be multiple reasons for this:
- Are you using the correct nixvim branch for your environement? This means that if you are using nixpkgs/nixos/home-manager 23.11 then you need to use the 23.11 branch of nixvim
- If you are using nixpkgs/nixos/home-manager unstable they need to be up-to-date, templ was added on the 21st of december, so you need to have at least this version. Generally we only guarantee that the master of nixvim works with a very recent (not more than a week old) nixpkgs

psyos10:
Apologies yep, the 23.11 works for me, I was initially confused by why the templ package wasn't there. My nixpkgs isn't that up to date. Thanks.

PR #907, opened 2024-01-06
tests: Add a test for the nixos module

traxys:


flake-modules/tests.nix
```diff
@@ -39,6 +40,17 @@
           nixvim = self;
         })
         .activationPackage;
+
+      nixos =
```
GaetanLepage, line 44:
```suggestion
      nixos-module =
```

And similarly, `home-manager-module` above.
I feel like it's more explicit.

traxys: Done. Also disabled the nixos checks for darwin

GaetanLepage (approved):
LGTM !

PR #910, opened 2024-01-06
flake/wrappers: move wrappers checks to wrappers.nix

GaetanLepage:


traxys (commented):
You should be able to remove `self` and `inputs` from tests.nix if you do this

traxys (approved):

PR #914, opened 2024-01-06
standalone: pass `extraSpecialArgs` to lib.evalModules directly

mirkolenz:
Currently, `extraSpecialArgs` are assigned to `_module.args` in the standalone wrapper. I would like to pass some arguments that shall be used for imports, so `_module.args` sadly does not work. This PR proposes to change the behavior and align it with what Home Manager for example does: Pass the property `extraSpecialArgs` as `specialArgs` to `lib.evalModules`. This would enable my use case.

Strictly speaking this is a breaking change, but since Nix/NixOS openly advocates the use of `_modules.args` for regular parameters anyway, I assume users are familiar with it and people wanting to retain the current behavior (i.e., overriding values in `extraSpecialArgs`) should hopefully be using `_modules.args` already.

In any case, this change should **not** be backported to the `23.11` branch.

### DIFF
--- a/tests/extra-args.nix
+++ b/tests/extra-args.nix
@@ -2,14 +2,26 @@
   makeNixvimWithModule,
   pkgs,
 }: let
+  defaultModule = {regularArg, ...}: {
+    extraConfigLua = ''
+      -- regularArg=${regularArg}
+    '';
+  };
   generated = makeNixvimWithModule {
-    module = {importedArgument, ...}: {
-      extraConfigLua = ''
-        -- importedArgument=${importedArgument}
-      '';
+    module = {
+      regularArg,
+      extraModule,
+      ...
+    }: {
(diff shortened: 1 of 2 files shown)


traxys:
would it be possible to add a test that passes with the new behaviour and not the previous one? We have a test for extra-args in tests/extra-args.nix

mirkolenz:
Totally makes sense to add a test for this behavior. I just pushed a commit where in addition to checking `_modules.args`, `specialArgs` is tested as well. I did it by adding a new parameter `extraModule` that is then used in the module's `imports`. If passing such a module using the regular `_module.args` there will be an infinite recursion (I tested that as well by temporarily reverting the initial commit of this PR). I hope the adjusted test makes sense to you.

traxys (approved):

PR #915, opened 2024-01-06
misc: correct molten image provider options

ptonner:
fixes the option for `imageProvider` to match [configuration](https://github.com/benlubas/molten-nvim/tree/main#configuration) of molten (use "image.nvim" instead of "image_nvim")

### DIFF
--- a/plugins/utils/molten.nix
+++ b/plugins/utils/molten.nix
@@ -1,209 +1,205 @@
-{
-  lib,
-  pkgs,
-  ...
-} @ args:
+{ lib, pkgs, ... }@args:
 with lib;
-with import ../helpers.nix {inherit lib;};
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Thank you for this fix !
However, you have changed the formatting (no need to add `@args` at the beginning).
Please, make sure to run `nix fmt` before pushing :)

ptonner:
ah crap, I forgot my bad

ptonner:
(also, I think the `args` was there already. I can drop if needed)

GaetanLepage:
> (also, I think the `args` was there already. I can drop if needed)

Indeed ! No, leave them, they are used.

GaetanLepage (approved):

PR #925, opened 2024-01-10
plugins/none-ls: add bandit source

garaiza-93:
Adds none-ls' source for [bandit](https://github.com/PyCQA/bandit)


GaetanLepage (approved):
LGTM !

PR #926, opened 2024-01-11
plugins/lsp: add nushell language server

siph:
Support for nushells built-in [language server](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#nushell).


GaetanLepage (approved):
LGTM, thanks !

PR #927, opened 2024-01-11
Various refactors around `mkPlugin` (for legacy 'globals'-based plugins)

GaetanLepage:


lib/helpers.nix
```diff
@@ -386,23 +386,13 @@ with lib; rec {
...
-    else val;
-
   mkDefaultOpt = {
     type,
     global,
     description ? null,
     example ? null,
     default ? null,
-    value ? v: (globalVal v),
+    value ? v: v,
```
traxys: there is  `lib.id` that may be clearer

traxys (approved):

lib/helpers.nix
GaetanLepage: This option gets removed in one of the following commits actually.

traxys: true...... I even read it

PR #947, opened 2024-01-17
plugins/alpha: fix example (for buttons)

GaetanLepage:
The alpha doc is misleading: all of the `opts` options should be optional.
However, for buttons, not specifying `opts.shortcut` is causing an error.
This patch fixes the affected examples.


GaetanLepage:
I opened an issue upstream: https://github.com/goolord/alpha-nvim/issues/255

PR #949, opened 2024-01-18
plugins/rustaceanvim: init

### LINKED ISSUE #893: Request: rustaceanvim
**Plugin**: rustaceanvim
**Plugin repository**: https://github.com/mrcjkb/rustaceanvim
**Is the plugin on nixpkgs**: Yes

## Extra information
`rust-tools` has been archived and effectively replaced by this fork. A flake is available as well.

GaetanLepage:
Add support for [rustaceanvim](https://github.com/mrcjkb/rustaceanvim).

Fixes #893


plugins/languages/rustaceanvim.nix
```diff
@@ -0,0 +1,271 @@
...
+          Set to `null` to disable its automatic installation.
+        '';
+        example = null;
+      };
+
+      tools = {
+        executor =
+          helpers.defaultNullOpts.mkEnumFirstDefault
+          ["termopen" "quickfix" "toggleterm" "vimux"]
+          "How to execute terminal commands.";
```
traxys, line 34: I think this can be a function: https://github.com/mrcjkb/rustaceanvim/blob/a095a1f9397310d108ead239edc068b2624695b6/lua/rustaceanvim/executors/init.lua#L13-L14

plugins/languages/rustaceanvim.nix
```diff
@@ -0,0 +1,271 @@
...
+
+          Default:
+          ```lua
+            function()
+              return { 'rust-analyzer', '--log-file', RustaceanConfig.server.logfile }
+            end
+          ```
+        '';
+
+        settings = helpers.mkNullOrStrLuaFnOr (with types; attrsOf anything) ''
```
traxys, line 145: We have a list of all rust-analyzer options somewhere, we should re-use that

traxys (changes requested):

plugins/languages/rustaceanvim.nix
GaetanLepage: Indeed, changed.

plugins/languages/rustaceanvim.nix
traxys: Well we already do it for rust-tools and the LSP options, so let's do it here too

GaetanLepage: Ok, done

plugins/languages/rustaceanvim.nix
```diff
@@ -0,0 +1,281 @@
...
+        '';
+        example = null;
+      };
+
+      tools = {
+        executor =
+          helpers.defaultNullOpts.mkNullable
+          (
+            with helpers.nixvimTypes;
+              maybeRaw
```
traxys, line 35: I'd suggest either strLuaFn (enum ....) to help in the documentation

traxys (changes requested):

plugins/languages/rustaceanvim.nix
GaetanLepage: Done !

PR #953, opened 2024-01-19
plugins/none-ls: add formatting/code_action source: ltrs

GeoffreyFrogeye:


traxys (changes requested):
I think we have a test file where we enable all none-ls sources, could you add an entry in it?

traxys (approved):

PR #958, opened 2024-01-21
plugins/rest: init

GaetanLepage:
Add support for [rest.nvim](https://github.com/rest-nvim/rest.nvim).

### DIFF
--- a/plugins/utils/rest.nix
+++ b/plugins/utils/rest.nix
@@ -1,0 +1,183 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.rest;
+in {
+  meta.maintainers = [maintainers.GaetanLepage];
+
+  options.plugins.rest =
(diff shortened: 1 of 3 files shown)


plugins/utils/rest.nix
```diff
@@ -0,0 +1,179 @@
...
+              )
+          )
+          "false"
+          ''
+            Table of curl `--write-out` variables or false if disabled.
+            For more granular control see [Statistics Spec](https://github.com/rest-nvim/rest.nvim?tab=readme-ov-file#statistics-spec).
+          '';
+
+        formatters =
+          helpers.defaultNullOpts.mkAttrsOf types.str
```
traxys, line 90: (either str (enum [false]))

traxys (changes requested):

plugins/utils/rest.nix
GaetanLepage: Done

traxys (approved):

Issue #964, opened 2024-01-22
[PLUGIN REQUEST] typescript-tools


garaiza-93:
| Field    | Description |
|----------|-------------|
| Plugin   | `typescript-tools`  | 
| Homepage | `https://github.com/pmizio/typescript-tools.nvim/`     | 
| Nixpkgs  | `https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/applications/editors/vim/plugins/generated.nix#L10345` | 

## Extra Information

It's a suite of tools similar to rust-tools/rustacean, but for TS! This plugin aims to solve typescript-language-server woes.

PR #966, opened 2024-01-23
plugins/typescript-tools: init + test

garaiza-93:
Adds  typescript-tools: https://github.com/pmizio/typescript-tools.nvim/

Fixes https://github.com/nix-community/nixvim/issues/964

### DIFF
--- a/plugins/languages/typescript-tools.nix
+++ b/plugins/languages/typescript-tools.nix
@@ -1,0 +1,178 @@
+{
+  lib,
+  pkgs,
+  helpers,
+  config,
+  ...
+}:
+with lib; let
+  cfg = config.plugins.typescript-tools;
+in {
+  options.plugins.typescript-tools = {
+    enable = mkEnableOption "typescript-tools";
+    package =
+      helpers.mkPackageOption "typescript-tools"
+      pkgs.vimPlugins.typescript-tools-nvim;
+
+    onAttach =
+      helpers.defaultNullOpts.mkLuaFn "__lspOnAttach"
+      "Lua code to run when tsserver attaches to a buffer.";
+    handlers = mkOption {
+      type = with helpers.nixvimTypes; nullOr (attrsOf strLuaFn);
+      apply = v: helpers.ifNonNull' v (mapAttrs (_: helpers.mkRaw) v);
+      default = null;
+      description = "How tsserver should respond to LSP requests";
+      example = {
+        "textDocument/publishDiagnostics" = ''
+          require("typescript-tools.api").filter_diagnostics(
+            -- Ignore 'This may be converted to an async function' diagnostics.
+            { 80006 }
+          )
+        '';
+      };
+    };
+
+    settings = {
+      separateDiagnosticServer =
+        helpers.defaultNullOpts.mkBool true
+        "Spawns an additional tsserver instance to calculate diagnostics";
+
+      publishDiagnosticOn =
+        helpers.defaultNullOpts.mkEnum ["change" "insert_leave"]
+        "insert_leave" ''
+          Either "change" or "insert_leave". Determines when the client asks the server about diagnostics
+        '';
+
+      exposeAsCodeAction = mkOption {
+        type = with types;
+          either (enum ["all"]) (listOf (enum [
+            "fix_all"
+            "add_missing_imports"
+            "remove_unused"
+            "remove_unused_imports"
+            "organize_imports"
+            "insert_leave"
+          ]));
+        default = [];
+        description = "Specify what to expose as code actions.";
+      };
+
+      tsserverPath = helpers.mkNullOrStr ''
+        Specify a custom path to `tsserver.js` file, if this is nil or file under path
+        doesn't exist then standard path resolution strategy is applied
+      '';
+
+      tsserverPlugins = with helpers.nixvimTypes;
+        helpers.mkNullOrOption (listOf (maybeRaw str)) ''
+          List of plugins for tsserver to load. See this plugins's README
+          at https://github.com/pmizio/typescript-tools.nvim/#-styled-components-support
+        '';
+
+      tsserverMaxMemory = helpers.mkNullOrOption (with helpers.nixvimTypes;
+          maybeRaw (either ints.unsigned (enum ["auto"]))) ''
+        This value is passed to: https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes
+        Memory limit in megabytes or "auto"(basically no limit)
+      '';
+
+      tsserverFormatOptions = mkOption {
+        type = with types; nullOr (attrsOf anything);
+        default = null;
+        description = "Configuration options that well be passed to the tsserver instance. Find available options [here](https://github.com/microsoft/TypeScript/blob/v5.0.4/src/server/protocol.ts#L3418)";
+        example = {
+          "tsserver_file_preferences" = ''
+            {
+              includeInlayParameterNameHints = "all",
+              includeCompletionsForModuleExports = true,
+              quotePreference = "auto",
+              ...
+            }
+          '';
+        };
+      };
+
+      tsserverFilePreferences = mkOption {
+        type = with types; nullOr (attrsOf anything);
+        default = null;
+        description = "Configuration options that well be passed to the tsserver instance. Find available options [here](https://github.com/microsoft/TypeScript/blob/v5.0.4/src/server/protocol.ts#L3439)";
+        example = {
+          "tsserver_format_options" = ''
+            {
+              allowIncompleteCompletions = false,
+              allowRenameOfImportPath = false,
+              ...
+            }'';
+        };
+      };
+
+      tsserverLocale = helpers.defaultNullOpts.mkStr "en" ''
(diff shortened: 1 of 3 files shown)


plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,42 @@
+{ lib, pkgs, helpers, config, ... }:
+with lib;
+let cfg = config.plugins.typescript-tools;
+in {
+  extraPlugins = [ cfg.package ];
+  options.plugins.typescript-tools = {
```
GaetanLepage, line 6:
```suggestion
in {
  options.plugins.typescript-tools = {
```

GaetanLepage:
Thank you for taking a shot at this one :)

garaiza-93:
Felt right since you're already working on rustacean for me :p

plugins/languages/typescript-tools.nix
garaiza-93: I don't know what's up with nil_ls, but it looooves to butcher this repo's formatting. As of now I'm just rolling with it and running `nix fmt` afterwards. :sob:
If you know of something I should incorporate into my nvim setup, please let me know because this is ANNOYING

garaiza-93:
I've added most of the settings. There's a hurdle though.
How could I add handlers without wanting to tear my hair out? Looking at typescript-tools [readme](https://github.com/pmizio/typescript-tools.nvim/#%EF%B8%8F-configuration) it is a list of variables (or attributes in Nix's case). There's a number of [supported methods](https://github.com/pmizio/typescript-tools.nvim/#supported-lsp-methods) too, and I hope I don't have to add these one-by-one!!

GaetanLepage:
> How could I add handlers without wanting to tear my hair out?

```nix
handlers = mkOption {
  type = with helpers.nixvimTypes; nullOr (attrsOf strLuaFn);
  apply = v: helpers.ifNonNull' v (mapAttrs (_: helpers.mkRaw) v);
  default = null;
  description = "...";
  example = {
    "textDocument/publishDiagnostics" = ''
      api.filter_diagnostics(
        -- Ignore 'This may be converted to an async function' diagnostics.
        { 80006 }
      )
    '';
  };
};
```

plugins/languages/typescript-tools.nix
GaetanLepage: Don't forget to remove this wrong line (`extraPlugins = [ cfg.package ];`)

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,95 @@
...
+in {
+  extraPlugins = [cfg.package];
+  options.plugins.typescript-tools = {
+    enable = mkEnableOption "typescript-tools";
+    package =
+      helpers.mkPackageOption "typescript-tools"
+      pkgs.vimPlugins.typescript-tools-nvim;
+
+    onAttach =
+      helpers.defaultNullOpts.mkLuaFn "null"
```
traxys, line 19: I _think_ we should set the default value of this to `__lspOnAttach` (so avoiding the use of a defaultNullOpts).
We do some heavy lifting in __lspOnAttach to handle integrations with other plugins, so it'd be nice to benefit from it here too.

traxys (changes requested):

plugins/languages/typescript-tools.nix
garaiza-93: I'll change it, but I'm not understanding what this does. I know you answered me before, but where can I find where it does?
I know `mkRaw` takes a thing and makes it into raw Lua code. However when I search for `__lspOnAttach` [this line of code](https://github.com/nix-community/nixvim/blob/ddce82d0d1fb88c957ecb7debed2ae6c63b60750/plugins/lsp/default.nix#L152) is as close as I can get. I highly, highly doubt this is it, because it doesn't tell me much.

traxys: This is exactly that!

If you  search for `onAttach` in nixvim there plugins that add things to it

garaiza-93: I think I get it now. We're inserting `__lspOnAttach` as raw Lua code, thus it references the value assigned to this in the generated init.lua. The value assigned to `__lspOnAttach` is determined by `plugins.lsp.onAttach`!

traxys: Yes, we only do that sort of trickery for `__lspOnAttach` I think

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+        '';
+      };
+    };
+
+    settings = {
+      separateDiagnosticServer =
+        helpers.defaultNullOpts.mkBool true
+        "Spawns an additional tsserver instance to calculate diagnostics";
+
+      publishDiagnosticOn = mkOption {
```
traxys, line 37: please use helpers.defaultNullOpts.mkEnum

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+        type = with types; enum ["change" "insert_leave"];
+        default = "insert_leave";
+        description = ''
+          Either "change" or "insert_leave". Determines when the client asks the server about diagnostics'';
+      };
+
+      exposeAsCodeAction = mkOption {
+        type = with types;
+          either (enum ["all"]) (enum [
```
traxys, line 46: you should be able to merge the two enum, though it should be a `helpers.defaultNullOpts.mkListOf (enum ....)`

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+            "remove_unused"
+            "remove_unused_imports"
+            "organize_imports"
+            "insert_leave"
+          ]);
+        default = [];
+        description = "Specify what to expose as code actions.";
+      };
+
+      tsserverPath = helpers.mkNullOrOption types.str ''
```
traxys, line 58: helpers.mkNullOrStr, this allows the user to provide raw lua values instead of a string if he requires it

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+        Specify a custom path to `tsserver.js` file, if this is nil or file under path
+        doesn't exist then standard path resolution strategy is applied
+      '';
+
+      tsserverPlugins = helpers.mkNullOrOption (types.listOf types.str) ''
```
traxys, line 63: the type should be (with helpers.nixvimTypes; listOf (maybeRaw str)), to allow the inclusion of raw lua

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+        List of plugins for tsserver to load. See this plugins's README
+        at https://github.com/pmizio/typescript-tools.nvim/#-styled-components-support
+      '';
+
+      tsserverMaxMemory =
+        helpers.mkNullOrOption
+        (with types; either ints.unsigned (enum ["auto"])) ''
```
traxys, line 70: with nixvimTypes; maybeRaw (either  ....)

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+          This value is passed to: https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes
+          Memory limit in megabytes or "auto"(basically no limit)
+        '';
+
+      tsserverFormatOptions = mkOption {
+        type = with types; nullOr (attrsOf str);
```
traxys, line 76: attrsOf anything  I think, the option does not only include raw lua code, but strings, bools, ....

(This also means that you can remove the `apply`

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+              includeCompletionsForModuleExports = true,
+              quotePreference = "auto",
+              ...
+            }
+          '';
+        };
+      };
+
+      tsserverFilePreferences = mkOption {
+        type = with types; nullOr (attrsOf str);
```
traxys, line 94: same, attrsOf anything

plugins/languages/typescript-tools.nix
```diff
@@ -0,0 +1,177 @@
...
+      completeFunctionCalls = helpers.defaultNullOpts.mkBool false ''
+        Mirror of VSCode's `typescript.suggest.completeFunctionCalls`
+      '';
+
+      includeCompletionsWithInsertText = helpers.defaultNullOpts.mkBool true ''
+        Mirror of VSCode's `typescript.suggest.completeFunctionCalls`
+      '';
+
+      codeLens = mkOption {
+        type = with types; (enum ["off" "all" "implementations_only" "references_only"]);
```
traxys, line 122: helpers.defaultNullOpts.mkEnum

traxys (changes requested):

plugins/languages/typescript-tools.nix
garaiza-93: This plugin's docs state that it has to be either "all" or any number of the other options. If I merge these, wouldn't that allow something incorrect like `["all", "fix_all"]`?

traxys: Yes it would. You need something of the form (either (enum ["all"]) (listOf (enum [....])))

traxys (approved):

PR #977, opened 2024-01-25
helpers: init `neovim-plugin.mkSetupOptionsOption`

GaetanLepage:


lib/neovim-plugin.nix
```diff
@@ -0,0 +1,37 @@
...
+  extraOptionsOptions = {
+    extraOptions = mkOption {
+      default = {};
+      type = types.attrs;
+      description = ''
+        These attributes will be added to the table parameter for the setup function.
+        (Can override other attributes set by nixvim)
+      '';
+    };
+  };
```
traxys, line 36: We won't need this in the future right? As the setupOptions are exactly the same?

While we are at it can you switch from attrs to attrsOf anything

plugins/utils/hmts.nix
traxys, line 1: Why is this diff here?

traxys (changes requested):

lib/neovim-plugin.nix
GaetanLepage: > We won't need this in the future right? As the setupOptions are exactly the same?

In the future no, we won't need that.
But for backwards compatibility, we have to keep it for now.



> While we are at it can you switch from attrs to attrsOf anything

done !

plugins/utils/hmts.nix
traxys, line 1: I think this file is not in the right place?

traxys (changes requested):

plugins/utils/hmts.nix
GaetanLepage: Done.

traxys (approved):

PR #979, opened 2024-01-25
plugins/rustaceanvim: Allow to add on_attach to rustaceanvim

traxys:
This is useful to display inlay hints in neovim >=0.10 for example

### DIFF
--- a/plugins/languages/rustaceanvim.nix
+++ b/plugins/languages/rustaceanvim.nix
@@ -135,6 +135,8 @@ in {
           ```
         '';
 
+        onAttach = helpers.defaultNullOpts.mkLuaFn "__lspOnAttach" "Function to call on attach";
+
         cmd = helpers.mkNullOrStrLuaFnOr (with types; listOf str) ''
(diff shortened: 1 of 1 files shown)


plugins/languages/rustaceanvim.nix
```diff
@@ -278,5 +274,18 @@ in {
...
+      extraPlugins = [cfg.package];
+
+      extraPackages =
+        optional
+        (cfg.rustAnalyzerPackage != null)
+        cfg.rustAnalyzerPackage;
+
+      plugins.lsp.postConfig = ''
+        vim.g.rustaceanvim = ${helpers.toLuaObject globalOptions}
+      '';
```
GaetanLepage, line 289:
```suggestion
      plugins.lsp.postConfig = let
        globalOptions = with cfg; {
          ...
        };
      in ''
        vim.g.rustaceanvim = ${helpers.toLuaObject globalOptions}
      '';
```

I would rather preserve the "scope locality".

GaetanLepage (approved):
LGTM !

PR #980, opened 2024-01-26
plugins: clean up module descriptions

foo-dogsquared:
Only for the `enable` option though, I haven't fully explored the whole manual yet. Also did some quick corrections on capitalization and whatnot. I just mainly did this to clean up the manual page.


GaetanLepage (approved):
Thank you for this cleaning :)

PR #983, opened 2024-01-27
plugins/base16: minor improvements

### LINKED ISSUE #986: [BUG] lualine doesnt recognize `base16.customColorScheme`
| Plugin       | `lualine`, `nvim-base16`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

If `plugins.lualine.theme` is set to `"base16"`, it follows `base16.colorscheme` and ignores `base16.customColorScheme`.
The reason appears to be that `plugins.lualine.theme` and `colorscheme.base16.customColorScheme` both use `extraConfigLua`.

```nix
programs.nixvim = {
  enable = true;
  colorscheme.base16 = {
    enable = true;
    customColorScheme = {
      base00 = "#191724";
      base01 = "#1f1d2e";
      base02 = "#26233a";
      base03 = "#6e6a86";
      base04 = "#908caa";
      base05 = "#e0def4";
      base06 = "#e0def4";
      base07 = "#524f67";
      base08 = "#eb6f92";
      base09 = "#f6c177";
      base0A = "#ebbcba";
      base0B = "#31748f";
      base0C = "#9ccfd8";
      base0D = "#c4a7e7";
      base0E = "#f6c177";
      base0F = "#524f67";
    };
  };
  plugins.lualine = {
    enable = true;
    theme = "base16";
  };
}
```

player131007:
Fixes #986

### DIFF
--- a/plugins/colorschemes/base16.nix
+++ b/plugins/colorschemes/base16.nix
@@ -87,6 +87,7 @@ in {
     extraPlugins = [cfg.package];
 
     plugins.airline.theme = mkIf cfg.setUpBar "base16";
+    plugins.lualine.theme = mkIf cfg.setUpBar "base16";
     plugins.lightline.colorscheme = null;
 
     options.termguicolors = mkIf cfg.useTruecolor true;
--- a/plugins/colorschemes/base16.nix
+++ b/plugins/colorschemes/base16.nix
@@ -22,9 +22,9 @@ in {
       };
 
       colorscheme = mkOption {
-        type = types.enum themes;
+        type = types.nullOr (types.enum themes);
         description = "The base16 colorscheme to use";
-        default = head themes;
+        default = null;
       };
 
       setUpBar = mkOption {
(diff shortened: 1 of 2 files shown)


GaetanLepage:
Thank you for the contribution.
Most of the changes make sense.
However, what would be the point of having the `colorscheme` option set to `null` ?

player131007:
We initialized  `customColorScheme` in `extraConfigLuaPre`. If `colorscheme` was not `null`, it would come after `customColorScheme` in the config and reset the colorscheme

GaetanLepage:
> We initialized `customColorScheme` in `extraConfigLuaPre`. If `colorscheme` was not `null`, it would come after `customColorScheme` in the config and reset the colorscheme

I am not sure to get what you mean ?

Why do a user would want to leave `colorscheme.base16.colorscheme = null` ?
I would say that this is an "invalid" configuration, don't you think ?

player131007:
> I am not sure to get what you mean ?

If both `colorscheme` and `customColorScheme` is not `null`, the resulting init.lua will look something like this:
```lua
-- extraConfigLuaPre goes here
require('base16-colorscheme').setup(customColorScheme)


-- extraConfigVim goes here (which includes colorscheme)
vim.cmd [[
  colorscheme base16-colorscheme
]]
```
The resulting colorscheme is `colorscheme.base16.colorscheme`.

> Why do a user would want to leave `colorscheme.base16.colorscheme = null` ?
If they want to set a custom theme, they could leave `colorscheme = null` and set `customColorScheme` instead.

traxys:
I think you need to update some tests

traxys (approved):

PR #996, opened 2024-02-02
plugins/lsp: set correct name for the elixirls package

573:
plugins/lsp: set correct name for the elixirls package.


GaetanLepage:
Thanks ! Can you set the commit message (and PR title) to:
> plugins/lsp: set correct name for the elixirls package

573:
See #998

PR #1001, opened 2024-02-02
plugins/lsp: set correct package level for the ruff-lsp package

573:


GaetanLepage (approved):

GaetanLepage:
The CI is annoying to run.
Also, there is no problem with having just `pkgs.ruff-lsp`. I will then leave it as is.

PR #1011, opened 2024-02-03
docs: Use a nixpkgs patch instead of copy/pasting nixpkgs functions

traxys:
This makes the code more maintainable, as it only depends on our small patch, instead of a large number of internals of nixpkgs.


GaetanLepage (approved):
LGTM !

PR #1014, opened 2024-02-04
plugins/lsp: add vala-language-server

foo-dogsquared:
Add LSP support for [vala-language-server](https://github.com/vala-lang/vala-language-server).

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -598,6 +598,12 @@ with lib; let
       package = pkgs.typst-lsp;
     }
     {
+      name = "vala-ls";
+      description = "vala_ls for Vala";
+      serverName = "vala_ls";
+      package = pkgs.vala-language-server;
+    }
+    {
       name = "vls";
       description = "vls for V";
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -597,6 +597,11 @@ with lib; let
       description = "typst-lsp for typst";
       package = pkgs.typst-lsp;
     }
+    {
+      name = "vala-language-server";
+      serverName = "vala_ls";
+      description = "vala-language-server for Vala";
+    }
```
GaetanLepage, line 604:
```suggestion
    {
      name = "vala-ls";
      description = "vala_ls for Vala";
      serverName = "vala_ls";
      package = pkgs.vala-language-server;
    }
```

foo-dogsquared: Done. I assume we're following the naming convention from [lspconfig's filetree](https://github.com/neovim/nvim-lspconfig/tree/master/lua/lspconfig/server_configurations) (just in kebab-case), right?

GaetanLepage (approved):
LGTM, thanks !

PR #1016, opened 2024-02-04
modules/highlights: post -> pre loading

khaneliman:
Highlights currently get configured in extraConfigLuaPost. Any plugin configs that reference them before they get injected will throw errors. This moves the global highlights to be injected at the beginning of the config with extraConfigLuaPre.

### DIFF
--- a/modules/highlights.nix
+++ b/modules/highlights.nix
@@ -30,7 +30,7 @@ with lib; {
   };
 
   config = mkIf (config.highlight != {} || config.match != {}) {
-    extraConfigLuaPost =
+    extraConfigLuaPre =
       (optionalString (config.highlight != {}) ''
         -- Highlight groups {{
         do


traxys (approved):

ErrorNoInternet:
Some of my own highlights aren't working anymore, e.g. CursorLine and FoldColumn with the nord colorscheme

traxys:
They depend on values defined by nord? We may need to add a section for highlight groups post, I had not thought of that

khaneliman:
> They depend on values defined by nord? We may need to add a section for highlight groups post, I had not thought of that
Yeah I think just expand the highlight option to support pre and post.

budimanjojo:
I have the same problem as @ErrorNoInternet.
Before this commit, `highlight` are set before `colorscheme`.
After this commit, `highlight` are set after `colorscheme`.

I want to override some colors from the colorscheme, so it should be after `colorscheme` command. Is there something to be done here?

PR #1017, opened 2024-02-04
feat: add neoscroll plugin

jakehamilton:
This commit initializes a new `neoscroll` plugin option.

Customization of this plugin is currently managed only via `extraOptions`, but this can be upgraded to full options for the module.


traxys:
We have a new way to manage such plugins, with a `mkSettingsOption` that allows to make a freeform type, see #991 for an inspiration (though it does _not_ use the mkSettingsOption)

jakehamilton:
I've gone ahead and refactored the module to use `mkSettingsOption`. I think I got it right, but let me know if there are any other tweaks that need to happen :+1:

GaetanLepage:
Hi !
Actually, you will want to use the `mkNeovimPlugin` helper.
You can have a look at our [TEMPLATE.nix](https://github.com/nix-community/nixvim/blob/5293373732f51f83017484831bc26365caecd965/plugins/TEMPLATE.nix) and at this [real example](https://github.com/nix-community/nixvim/blob/main/plugins/utils/bacon.nix).

jakehamilton:
Gotcha, I'll take a look at refactoring this 👍

GaetanLepage:
@jakehamilton Do you need some help with finalizing this plugin ? I can take over if you wish :)

jakehamilton:
> @jakehamilton Do you need some help with finalizing this plugin ? I can take over if you wish :)

I haven't gotten back around to reworking this one, but if you want to take over feel free! Otherwise it may take me a bit to allocate time for it.

traxys (approved):

PR #1018, opened 2024-02-04
feat: allow custom formatting for status components

jakehamilton:
Lualine allows you to specify a custom formatter function which can process the final status component string. This can be useful for augmenting existing status components and tweaking the output dynamically.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -80,6 +80,17 @@ with lib; let
               "1"
               "Adds padding to the left and right of components.";
 
+            fmt = helpers.mkNullOrLuaFn ''
+              A lua function to format the component string.
+
+              Example:
+              ```lua
+                function(text)
+                  return text .. "!!!"
+                end
+              ```
+            '';
+
             extraConfig = mkOption {
               type = types.attrs;
               default = {};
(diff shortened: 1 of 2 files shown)


plugins/statuslines/lualine.nix
```diff
@@ -80,6 +80,10 @@ with lib; let
               "1"
               "Adds padding to the left and right of components.";
 
+            fmt = helpers.mkNullOrOption helpers.nixvimTypes.rawLua ''
```
traxys, line 83: I'd suggest making an option of type `nullOr strLuaFn` and using an `apply` to map it to a raw value, so the user can simply input a string (see other options that use `apply`, many of them do that)

traxys (changes requested):

tests/test-sources/plugins/statuslines/lualine.nix
```diff
@@ -85,6 +85,11 @@
               TelescopePrompt = "Telescope";
               NvimTree = "NvimTree";
             };
+            fmt.__raw = ''
```
traxys, line 88: You need to remove the `__raw`

traxys (changes requested):

tests/test-sources/plugins/statuslines/lualine.nix
jakehamilton: Oh whoops! Thank you for catching this!

traxys (approved):

PR #1020, opened 2024-02-04
fix: add ripgrep for todo-comments functionality

jakehamilton:
Attempting to use todo-comments without `ripgrep` isn't very useful since the plugin won't be able to find comments in your files based on keywords. This commit adds a new option, defaulting enabled, to add `ripgrep` to the resulting neovim build.

### DIFF
--- a/plugins/utils/todo-comments.nix
+++ b/plugins/utils/todo-comments.nix
@@ -33,6 +33,15 @@ in {
           "todo-comments"
           pkgs.vimPlugins.todo-comments-nvim;
 
+        ripgrep = {
+          enable = mkOption {
+            type = types.bool;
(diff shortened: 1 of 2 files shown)


traxys (changes requested):
There a are a few other places where we have those kind of optional but set by default packages, like gcc for treesitter, and we use a single option, of type null or package, and don't add the package if it's null. Could you do the same here please?

jakehamilton:
I've updated this PR to match how things are done in the treesitter module. Let me know if there's anything else that should be tweaked :+1:

traxys:
Could you add a test disabling ripgrep in tests/test-sources/plugins/utils/todo-comments.nix

traxys (approved):

PR #1021, opened 2024-02-04
feat: add option for customizing lsp capabilities

jakehamilton:
Some servers like `tsserver` conflict with additional LSPs when they are active on the same buffer. Other servers may misreport their capabilities or need tweaking to match user preferences. Previously this needed to be done in a much more ad-hoc way in the user's configuration or bespoke lua config would need to be written to handle these cases. This commit provides a central, nix-based solution for toggling lsp capabilities declaratively.

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -58,6 +58,12 @@ in {
                   description = "The server's name";
                 };
 
+                capabilities = mkOption {
+                  type = types.nullOr (types.attrsOf types.bool);
+                  description = "Control resolved capabilities for the language server.";
+                  default = null;
+                };
+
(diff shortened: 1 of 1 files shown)


plugins/lsp/default.nix
```diff
@@ -58,6 +58,12 @@ in {
                   description = "The server's name";
                 };
 
+                capabilities = mkOption {
+                  type = types.nullOr (types.attrsOf types.bool);
+                  description = "Control resolved capabilities for the language server.";
+                  default = null;
```
traxys, line 64: Maybe add an example

traxys (changes requested):

traxys (approved):

PR #1028, opened 2024-02-07
plugins/gitsigns: add gitPackage option

GaetanLepage:
gitsigns needs git to operate. Add `pkgs.git` as a dependency.

### DIFF
--- a/plugins/git/gitsigns.nix
+++ b/plugins/git/gitsigns.nix
@@ -37,7 +37,15 @@ with lib; let
 in {
   options.plugins.gitsigns = {
     enable = mkEnableOption "gitsigns plugin";
+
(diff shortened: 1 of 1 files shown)


plugins/git/gitsigns.nix
```diff
@@ -321,6 +329,9 @@ in {
   in
     mkIf cfg.enable {
       extraPlugins = [cfg.package];
+
+      extraPackages = optional (cfg.gitPackage != null) [cfg.gitPackage];
```
traxys, line 333: Optionals ? Or remove the list

traxys (changes requested):

plugins/git/gitsigns.nix
GaetanLepage: Indeed.

PR #1033, opened 2024-02-08
misc: refactor helpers propagation

GaetanLepage:
This PR moves the `helpers` from `_module.args` to `specialArgs`.
This is well more adapted and will allow us to not have to import `helpers.nix` from anywhere in the code base.

### DIFF
--- a/wrappers/nixos.nix
+++ b/wrappers/nixos.nix
@@ -1,6 +1,7 @@
 {
   modules,
   self,
+  getHelpers,
 }: {
   pkgs,
   config,
@@ -8,7 +9,8 @@
   ...
 } @ args: let
   inherit (lib) mkEnableOption mkOption mkOptionType mkForce mkMerge mkIf types;
-  shared = import ./_shared.nix modules args;
+  helpers = getHelpers pkgs;
+  shared = import ./_shared.nix {inherit modules helpers;} args;
   cfg = config.programs.nixvim;
   files =
     shared.configFiles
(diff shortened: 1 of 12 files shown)


GaetanLepage:
Thank you so much @roberth for your help with this !

traxys (approved):

wrappers/nixos.nix
```diff
@@ -19,16 +21,21 @@ in {
...
-            options = {
-              enable = mkEnableOption "nixvim";
-              defaultEditor = mkEnableOption "nixvim as the default editor";
-            };
-            config.wrapRc = mkForce true;
-          }
-        ]
-        ++ shared.topLevelModules);
+      type = types.submoduleWith {
+        shorthandOnlyDefinesConfig = true;
```
GaetanLepage, line 25: @roberth I have set this because I saw that [it was set by default when you use the `submodule` type](https://github.com/NixOS/nixpkgs/blob/master/lib/types.nix#L677-L680).
Do you have an opinion on that ?

roberth: You could probably remove it. By doing so, you enable a bit more syntax to be valid, which might be nice for users. IIRC they can then write `programs.nixvim.imports = [ ... ];`.

> Enabling this only has a benefit when the submodule defines an option named `config` or `options`.

https://nixos.org/manual/nixos/stable/#sec-option-types-submodule

GaetanLepage: Ok, thanks for the answer :)

PR #1035, opened 2024-02-09
plugins/toggleterm: Add custom terminals declaration

DeviousStoat:
One common use case with this plugin is to create custom terminals that you bind to keys as explained [here](https://github.com/akinsho/toggleterm.nvim?tab=readme-ov-file#custom-terminal-usage).

This tries to add a nice way to declare these custom terminals through the plugin configuration in `nixvim` directly.


DeviousStoat:
Oh I just realized that we could probably have `customTerms` as an attribute set of the keymap to the config instead of a list with the keymap. Since the keymap is required in my implementation it would avoid checking `keymap != null` and this awkward thing where custom terminals without a keymap are ignored. I will probably change to that. It sounds like it makes more sense.

traxys (changes requested):
Hey sorry for taking so long to respond!

I'm not sure this is exactly the right abstraction. I think we should add some (nixvim) options to the customTerms submodule:
- `name`: a string
- `toggle.enable`: a boolean controlling whether or not to add a  toggle keymap
- `toggle.key`: the key for toggling the terminal
- `toggle.options`: Keymap options for the toggle mapping

We would then generate `name = Terminal:new { ..... }` in the extraConfigLua(Pre?) and then add the keymapping to our keymaps option

Alternatively we could add a `helpers.mkToggletermTerminal {options}` function to the helpers, and this would generate the `Terminal:new{....}` snippet, that the user could use as he wants in the init.lua

DeviousStoat:
wow sorry I kinda stopped using this plugin and completely forgot about this PR.

I updated the PR, let me know what you think @traxys

PR #1048, opened 2024-02-10
Google java format

redyf:
Add an option to enable google-java-format for none-ls.


redyf:
I forgot to run the test before pushing this one, will redo.

PR #1049, opened 2024-02-10
plugins/none-ls: add google-java-format

redyf:
Add an option to enable google-java-format none-ls source.


traxys:
I think you encountered a bug in the way we make the none ls config, seeing the arithmetic error :/

redyf:
> I think you encountered a bug in the way we make the none ls config, seeing the arithmetic error :/

What should I do? That's pretty much the only thing holding me back from switching to none-ls. ;-;

traxys:
I think you need `_` instead of `-`. See https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTINS.md#google_java_format

redyf:
That makes sense, I'll try again in a bit

PR #1058, opened 2024-02-10
 plugins/none-ls: add prettierd

redyf:
add prettierd formatter source for none-ls.


redyf:
@traxys I did what you suggested and learned something new, thanks! Sorry for all confusion, I'm still on my way to learn git.

traxys (approved):
No problem! Git is not very easy!

PR #1060, opened 2024-02-11
plugins/none-ls add stylelint

tarantoj:
Adds stylelint to none-ls

### DIFF
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -96,6 +96,9 @@ with lib; let
       statix = {
         package = pkgs.statix;
       };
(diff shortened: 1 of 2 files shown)


tarantoj:
Looks like the build errors might be transient?

Not sure if I can trigger them again without a commit

GaetanLepage:
> Looks like the build errors might be transient?
> 
> Not sure if I can trigger them again without a commit

I re-triggered it.
Also, to artificially change your commit you can do:
```
git commit --amend --no-edit
```

GaetanLepage (approved):
LGTM

PR #1063, opened 2024-02-11
plugins/emmet: remove to strict typing on option 'mode'

GaetanLepage:
Values such as `"inv"` should be allowed.


traxys:
CI fails seem unrelated

GaetanLepage:
Indeed, now passes.

Issue #1074, opened 2024-02-13
[FEATURE] Environment-specific config inside NixVim modules


foo-dogsquared:
In home-manager (HM) where it can be deployed as part of NixOS or nix-darwin system, it has a neat feature of peeking from their configuration through certain attributes passed as a module argument. For example, you can peek your NixOS config with the `nixosConfig` attribute or your nix-darwin config with `darwinConfig`. This allows you to have environment-specific configuration for home-manager like in the following snippet.

```nix
{
    options.suites.desktop.pipewire.enable = lib.mkOption {
      type = lib.types.bool;
      default = attrs ? nixosConfig && lib.attrByPath [ "services" "pipewire" "enable" ] false attrs.nixosConfig;
      description = ''
        Enable whether to install Pipewire-related applications.

        This module is implicitly enabled if used as part of the NixOS
        configuration and has Pipewire service enabled.
      '';
    };
}
```

I'm aware there are various ways to have NixOS-/nix-darwin-specific home-manager settings but I find this approach nice as it allows some flexibility if you deploy HM configs both as part of NixOS systems and standalone.

This could be applied for NixVim where it can be deployed as part of home-manager, NixOS, nix-darwin, AND standalone for similar reasons stated above. This is useful for Neovim plugins that integrates with other programs outside of Neovim itself. For example, I could have parts of my NixVim config where I can implicitly enable my firenvim config if either home-manager or NixOS `programs.firefox.enable` is set where I would want Firenvim extension to be installed alongside it.

I have a patch ready to be made as a PR but I want some discussion first if this is suitable for NixVim. What does the maintainers think?

PR #608, opened 2023-09-28
docs: add user guide pages

Wolbyte:
This pull request aims to add a new `User Guide` section with pages covering the basics of nixvim installation and usage, as well as some FAQs and tips.

@GaetanLepage is the one helping to document those topics.

PR #1075, opened 2024-02-13
wrappers: add env-specific config as part of NixVim module

foo-dogsquared:
See #1074 for more details. I just went ahead to make this PR since discussion can also take place here anyways. ;p

### DIFF
--- a/docs/user-guide/install.md
+++ b/docs/user-guide/install.md
@@ -57,6 +57,9 @@ The imports can be added as a `imports = [ <nixvim_import> ]` in a configuration
 You will then be able to enable nixvim through `programs.nixvim.enable = true`, and configure the
 options as `programs.nixvim.<path>.<to>.<option> = <value>`.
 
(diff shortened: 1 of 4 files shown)


traxys:
I think this is a very good idea! We have a number of options that would benefit from this like choosing to install GCC or not for treesitter depending on standalone vs hmManager for example.

Though we need to add documentation for this feature, I'm not too sure where would be the best place

foo-dogsquared:
> I think this is a very good idea! We have a number of options that would benefit from this like choosing to install GCC or not for treesitter depending on standalone vs hmManager for example.

Having install extra packages when deployed as standalone is also (very very) nice use case too!

> Though we need to add documentation for this feature, I'm not too sure where would be the best place

Documentation, what documentation? ;p

I suppose it would be best put in a user guide (if it has one like in #608).

traxys:
For reference I added a documentation section in #1093, you can do something similar

traxys:
An user guide was added in #1121!

foo-dogsquared:
Added a section on the user guide.

traxys (approved):

PR #1079, opened 2024-02-13
helpers/vim-plugin/mkVimPlugin: add extraOptions and extraConfig options

GaetanLepage:
Add two options to the `vim-plugin.mkVimPlugin` helpers:
- `extraOptions` allows to add arbitrary options to the plugin module
- `extraConfig` allows to do extra fancy configuration

### DIFF
--- a/lib/vim-plugin.nix
+++ b/lib/vim-plugin.nix
@@ -18,6 +18,7 @@ with lib; {
     addExtraConfigRenameWarning ? false,
     extraOptions ? {},
     # config
+    extraConfig ? cfg: {},
     extraPlugins ? [],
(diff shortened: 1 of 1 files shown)


traxys (commented):
What is the use case for extra options instead of settings?

GaetanLepage:
> What is the use case for extra options instead of settings?

Although the plan is to have the plugin settings put in the `settings` option, the more nixvim-specific options should still be declared at the top-level of the plugin module.

Examples:
- `keymaps`
- `keymapsSilent`
- `gccPackage`
- ...

traxys (approved):

PR #1084, opened 2024-02-14
plugins/julia-cell: switch to mkVimPlugin

GaetanLepage:


plugins/languages/julia/julia-cell.nix
```diff
@@ -37,78 +35,66 @@
...
+      defaultPackage = pkgs.vimPlugins.vim-julia-cell;
+      globalPrefix = "julia_cell_";
 
-      package = helpers.mkPackageOption "julia-cell" pkgs.vimPlugins.vim-julia-cell;
+      maintainers = [maintainers.GaetanLepage];
 
-      delimitCellsBy = helpers.defaultNullOpts.mkEnumFirstDefault ["marks" "tags"] ''
-        Specifies if cells are delimited by 'marks' or 'tags'.
-      '';
+      # TODO introduced 2024-02-19: remove 2024-03-19
```
traxys, line 47: I'd suggest removal in 2024-04-19

traxys (changes requested):

plugins/languages/julia/julia-cell.nix
GaetanLepage: Done !

traxys (approved):

PR #1087, opened 2024-02-14
plugins/undotree: switch to settings-style

GaetanLepage:


plugins/utils/undotree.nix
```diff
@@ -13,120 +13,55 @@ with helpers.vim-plugin;
...
+    # TODO introduced 2024-02-19: remove 2024-03-19
+    imports = let
+      basePluginPath = ["plugins" "undotree"];
+    in
+      map
+      (
+        optionName:
+          mkRenamedOptionModule
+          (basePluginPath ++ [optionName])
+          (basePluginPath ++ ["settings" optionName])
```
traxys, line 25: The renamed options won't have the right name I think, they need  to start by an uppercase letter

traxys (commented):
I think before removing the options we need a way to link to the plugin homepage, or at the minimum add a `meta.url` option in order to fetch it from the documentation.

plugins/utils/undotree.nix
```diff
@@ -13,120 +13,55 @@ with helpers.vim-plugin;
...
-
-      showCursorLine = mkDefaultOpt {
-        type = types.bool;
-        description = ''
-          Show cursor line
-
-          Default: `true`
-        '';
-      };
+    # TODO introduced 2024-02-19: remove 2024-03-19
```
traxys, line 16: 04-19

traxys (changes requested):

plugins/utils/undotree.nix
GaetanLepage: Done

traxys (approved):

PR #1091, opened 2024-02-14
helpers/neovim-plugin/mkNeovimPlugin: add processSettings option

GaetanLepage:


traxys (commented):
Why would we want to process settings in a RFC-42 style plugin? If the plugin is _not_ RFC-42 style I'd suggest we don't use this function

GaetanLepage:
> Why would we want to process settings in a RFC-42 style plugin? If the plugin is _not_ RFC-42 style I'd suggest we don't use this function

Yes indeed.

PR #1093, opened 2024-02-14
docs: Add a short sections on helpers

traxys:


docs/helpers.md
```diff
@@ -0,0 +1,37 @@
...
+}: {
+    # Your config
+}
+```
+
+A certain number of helpers are defined that can be useful:
+
+- `helpers.emptyTable`: An empty lua table `{}` that will be included in the final lua configuration.
+
+- `helpers.mkRaw str`: Write the string `str` as raw lua in the final lua configuration.
```
GaetanLepage, line 21: Maybe we could give the equivalence here: `.__raw = "value"` and `__empty = <whatever>`.

traxys: Yes!

GaetanLepage (approved):
Great idea !!!

PR #1095, opened 2024-02-14
tests: Introduce a way to disable options when running in tests

### LINKED ISSUE #1085: [BUG] mkTestDerivationFromNvim fails for image.nvim
| Plugin       | `image`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `<VERSION>` | 

## Description

The `image.nvim` plugin breaks the `mkTestDerivationFromNvim` function in the build phase. This is due to a failure to get the current terminal size. I'm not too familiar with the `mkTestDerivationFromNvim` function, but I expect it tries to load the `image.nvim` plugin in a headless mode, which would obviously fail to get the terminal size. 

The function which fails is on [this line](https://github.com/3rd/image.nvim/blob/4c6cb5ad93ee93d8d7b7c84e1eb291cee99f0a0e/lua/image/utils/term.lua#L34), where the [`ioctl`](https://www.man7.org/linux/man-pages/man2/ioctl.2.html) function is called. In the same file, the function which calls `ioctl` is called as part of the `term.lua` file [here](https://github.com/3rd/image.nvim/blob/4c6cb5ad93ee93d8d7b7c84e1eb291cee99f0a0e/lua/image/utils/term.lua#L46C2-L46C3). 

This means that when this file is included in while this plugin is loaded, this function is called and we get an assert error. 

I think this would be best to solve in the `image.nvim` plugin itself, but I figured I'd create the issue here first as this was how I found it. 

For simplicity, start with the standalone example: 
```bash
mkdir /tmp/test-config
cd /tmp/test-config
nix flake init --template github:nix-community/nixvim
```

From here, `nix flake check .` works as expected.

Then add to `config/default.nix`:
```nix
plugins.image.enable = true;
```

Now `nix flake check .` will break with a log similar to this:

```
@nix { "action": "setPhase", "phase": "patchPhase" }Running phase: patchPhase@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running phase: configurePhase
no configure script, doing nothing
@nix { "action": "setPhase", "phase": "buildPhase" }
Running phase: buildPhase
ERROR: Error detected while processing /nix/store/4ccnws6j00gwma81nzhaygg76w85c782-init.lua:
E5113: Error while calling lua chunk: ...NeovimPackages/start/image.nvim/lua/image/utils/term.lua:34: Failed to get terminal size
stack traceback:
        [C]: in function 'assert'
        ...NeovimPackages/start/image.nvim/lua/image/utils/term.lua:34: in function 'update_size'
        ...NeovimPackages/start/image.nvim/lua/image/utils/term.lua:46: in main chunk
        [C]: in function 'require'
        ...NeovimPackages/start/image.nvim/lua/image/utils/init.lua:7: in main chunk
        [C]: in function 'require'
        ...myNeovimPackages/start/image.nvim/lua/image/renderer.lua:2: in main chunk
        [C]: in function 'require'
        ...ck/myNeovimPackages/start/image.nvim/lua/image/image.lua:2: in main
```

traxys:
In our basic template we used to provide a check based on `mkTestDerivationFromNvim`. The issue with this check (that is handled correctly internally) is that some plugins _can't_ be used in the test environment, for example image.nvim like in #1085.

This commit introduces a new function to generate such checks, `mkTestDerivationFromNixvimModule`, that wraps a nixvim configuration instead of a built nvim instance.

Then a configuration can rely on the newly added
`helpers.enableExceptInTests` attribute to disable parts of the configuration depending if it is evaluated in tests or in a real final configuration.

Resolves #1085

### DIFF
--- a/docs/helpers.md
+++ b/docs/helpers.md
@@ -37,3 +37,6 @@ A certain number of helpers are defined that can be useful:
   ```lua
     {"a", "b", [foo] = "bar"}
   ```
+
+- `helpers.enableExceptInTests`: Evaluates to `true`, except in `mkTestDerivationFromNixvimModule`
+  where it evaluates to `false`. This allows to skip instantiating plugins that can't be run in tests.
--- a/flake-modules/helpers.nix
+++ b/flake-modules/helpers.nix
@@ -1,7 +1,7 @@
 {getHelpers, ...}: {
-  _module.args.getHelpers = pkgs:
+  _module.args.getHelpers = pkgs: _nixvimTests:
     import ../lib/helpers.nix {
-      inherit pkgs;
+      inherit pkgs _nixvimTests;
       inherit (pkgs) lib;
     };
 
@@ -10,6 +10,6 @@
     config,
     ...
   }: {
-    _module.args.helpers = getHelpers pkgs;
+    _module.args.helpers = getHelpers pkgs false;
   };
 }
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -13,7 +13,7 @@
       }:
         import ../lib {
           inherit pkgs lib;
-          inherit (config.legacyPackages) makeNixvim;
+          inherit (config.legacyPackages) makeNixvim makeNixvimWithModule;
         }
     )
   );
--- a/flake-modules/tests.nix
+++ b/flake-modules/tests.nix
@@ -1,4 +1,4 @@
-{
+{self, ...}: {
   perSystem = {
     pkgs,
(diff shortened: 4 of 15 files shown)


GaetanLepage (approved):
A bit annoying that we have to do so much black magic but hey...
At least good job of figuring out this solution !

PR #1096, opened 2024-02-15
docs: Fix deprecation message for installLanguageServer

iynaix:
`plugins.lsp.servers.${serverName}.package` should be `null` instead of `false`


traxys:
I saw that and then forgot to ask for a change... Sorry and thank you!

traxys (approved):

PR #1101, opened 2024-02-15
plugins/neocord: init

redyf:
Add support for https://github.com/IogaMaster/neocord

### DIFF
--- a/tests/test-sources/plugins/utils/neocord.nix
+++ b/tests/test-sources/plugins/utils/neocord.nix
@@ -10,7 +10,7 @@
       # General options.
       autoUpdate = true;
       logo = "auto";
-      logo_tooltip = null;
+      logoTooltip = null;
 
       mainImage = "language";
       clientId = "1157438221865717891";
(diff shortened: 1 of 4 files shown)


redyf:
@traxys Hi traxys, I made PR for this plugin on nixpkgs and it was merged already. It shows up in nixpkgs search but only for the unstable branch. Is that why the checks failed?
```nix
 error: attribute 'neocord' missing
at /nix/store/sv7l2h04qc6k0yg3mikx58id5wql195y-source/plugins/utils/neocord.nix:14:51:
13| enable = mkEnableOption "neocord";
14| package = helpers.mkPackageOption "neocord" pkgs.vimPlugins.neocord;
```

traxys:
If the PR recently tricked to nixos-unstable we first need to update the flake.lock

traxys:
You should rebase onto main, we recently updated the flake.lock

tests/test-sources/plugins/utils/neocord.nix
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+  defaults = {
+    plugins.neocord = {
+      enable = true;
+
+      # General options.
+      autoUpdate = true;
+      logo = "auto";
+      logo_tooltip = null;
```
traxys, line 13: That would be logoTooltip

redyf: Thanks! I was looking for why this error was being caused for a while now hahaha

PR #1108, opened 2024-02-16
fix: neocord logo and logo_tooltip

redyf:
User couldn't set the logo url, also logo_tooltip didn't allow custom string because of the enum.


plugins/utils/neocord.nix
```diff
@@ -20,11 +20,11 @@ with lib;
         `:lua package.loaded.neocord:update()`
       '';
 
-      logo = helpers.defaultNullOpts.mkEnum ["auto" "url"] "auto" ''
+      logo = helpers.defaultNullOpts.mkStr "auto" ''
         Update the Logo to the specified option ("auto" or url).
       '';
 
-      logo_tooltip = helpers.defaultNullOpts.mkEnum ["null" "string"] "null" ''
+      logo_tooltip = helpers.defaultNullOpts.mkStr "Nixvim" ''
```
GaetanLepage, line 27:
```suggestion
      logo_tooltip = helpers.mkNullOrStr ''
```
The actual upstream default value is `null`.

tests/test-sources/plugins/utils/neocord.nix
```diff
@@ -14,7 +14,7 @@
       settings = {
         # General options.
         logo = "auto";
-        logo_tooltip = null;
+        logo_tooltip = "Nixvim";
```
GaetanLepage, line 17:
```suggestion
        logo_tooltip = null;
```

GaetanLepage: You could have left the one in the `example` config.

redyf: > You could have left the one in the `example` config.

Yeah, I should've done that. Thanks again for guiding me haha

PR #1105, opened 2024-02-15
plugins/lsp: Add statix

tarantoj:
Adds statix to lspconfig

PR #1111, opened 2024-02-17
Revert "plugins/lsp: Add statix"

tarantoj:
Reverts nix-community/nixvim#1105

After a bit of digging, it looks like [statix](https://github.com/nerdypepper/statix) [does not](https://github.com/neovim/nvim-lspconfig/issues/2887) actually implement the lsp.

I suggest we pull this out, or possibly mark it as "broken" to avoid any further confusion.


GaetanLepage (approved):
Ok makes sense.

PR #1121, opened 2024-02-18
docs: Introduce an user guide

traxys:
This user guide currently only documents how to install & some frequently asked questions. It also includes the previously introduced documentation on helpers.


docs/user-guide/faq.md
```diff
@@ -0,0 +1,26 @@
...
+extraPlugins = [(pkgs.vimUtils.buildVimPlugin {
+    name = "my-plugin";
+    src = pkgs.fetchFromGitHub {
+        owner = "<owner>";
+        repo = "<repo>";
+        rev = "<commit hash>";
+        hash = "<nix NAR hash>";
+    };
+})];
+```
```
GaetanLepage, line 24:
```suggestion
extraPlugins = [
  (pkgs.vimUtils.buildVimPlugin {
    name = "my-plugin";
    src = pkgs.fetchFromGitHub {
        owner = "<owner>";
        repo = "<repo>";
        rev = "<commit hash>";
        hash = "<nix NAR hash>";
    };
  })
];
```
```

GaetanLepage:
That is a very good idea !

docs/user-guide/faq.md
traxys: I specifically don't want to do that, in order to reduce the length of the example section.

GaetanLepage: Ok, as you want.
Personally, I find it more readable to break a line but you get to choose :)

traxys: I agree on readability of the code, but reducing the length of the doc will increase overall readability I think

GaetanLepage (approved):
LGTM ! Great work :)

PR #1125, opened 2024-02-19
Add _nixvimTests = false to flake.lib output

robhanlon22:
In my personal configuration, I rely on `nixvim.lib.${system}.helpers` to get access to things like `mkRaw` and `toLuaObject` for use in my extended `lib`.

With the changes made in 6d7e429537ae8143d8405c44a61479248776758c, `nixvim.lib.${system}.helpers` can no longer be accessed directly from the `nixvim` input as `_nixvimTests` is not provided as an input attribute, so loading `helpers.nix` fails.

This small change restores what I *think* is the intended behavior of this module.

### DIFF
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -14,7 +14,6 @@
         import ../lib {
           inherit pkgs lib;
           inherit (config.legacyPackages) makeNixvim makeNixvimWithModule;
-          _nixvimTests = false;
         }
     )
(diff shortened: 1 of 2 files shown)


traxys:
I think this should be added in lib/dafault.nix, only for the helpers import

traxys:
Yes, but utils.nix is only available through `helpers`

Issue #1072, opened 2024-02-13
[BUG]  Some errors after switching from Null-ls to none-ls


dracdavamp:
| Field        | Description |
|--------------|-------------|
| Plugin       | `none-ls`  | 
| Nixpkgs      | `23.11` | 
| Home Manager | `23.11` | 

## Description

After switching to newer nixvim (23.11) and after switching from null-ls to none-ls I get the following error
```
Error executing vim.schedule lua callback: ...imPackages/start/none-ls.nvim/lua/null-ls/generators.lua:10: module 'plenary.async' not found: 
        no field package.preload['plenary.async']
        no file '/nix/store/7fg2440ssbcwdmwm0ls7klhibibcnwp4-luajit-2.1.1693350652-env/share/lua/5.1/plenary/async.lua'
        no file '/nix/store/7fg2440ssbcwdmwm0ls7klhibibcnwp4-luajit-2.1.1693350652-env/share/lua/5.1/plenary/async/init.lua'
        no file '/nix/store/7fg2440ssbcwdmwm0ls7klhibibcnwp4-luajit-2.1.1693350652-env/lib/lua/5.1/plenary/async.so'
        no file '/nix/store/7fg2440ssbcwdmwm0ls7klhibibcnwp4-luajit-2.1.1693350652-env/lib/lua/5.1/plenary.so'
stack traceback:
        [C]: in function 'require'
        ...imPackages/start/none-ls.nvim/lua/null-ls/generators.lua:10: in function 'run'
        ...imPackages/start/none-ls.nvim/lua/null-ls/generators.lua:156: in function 'run_registered'
        ...mPackages/start/none-ls.nvim/lua/null-ls/diagnostics.lua:171: in function 'handler'
        .../myNeovimPackages/start/none-ls.nvim/lua/null-ls/rpc.lua:91: in function 'notify'
        ...eovim-unwrapped-0.9.4/share/nvim/runtime/lua/vim/lsp.lua:722: in function 'text_document_did_open_handler'
        ...eovim-unwrapped-0.9.4/share/nvim/runtime/lua/vim/lsp.lua:1526: in function '_on_attach'
        ...eovim-unwrapped-0.9.4/share/nvim/runtime/lua/vim/lsp.lua:1357: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```
I tries adding `extraPlugins = with pkgs; [luajitPackages.plenary-nvim];` but still no luck...

[...]

PR #1128, opened 2024-02-19
plugins/none-ls: add missing plenary-nvim dependency (#1072)

nawordar:


GaetanLepage:
Thank you for your PR !
However, plugin dependencies should be upstreamed if possible.
I opened a [PR](https://github.com/NixOS/nixpkgs/pull/289919) to do just that :)

PR #1137, opened 2024-02-20
modules/colorscheme: Set colorscheme after everything else

mxkrsv:
Without that, e.g. require('gruvbox').setup() was run *after* setting the colorsheme. Thus all settings were ignored.

### DIFF
--- a/modules/colorscheme.nix
+++ b/modules/colorscheme.nix
@@ -13,8 +13,8 @@ with lib; {
   };
 
   config = mkIf (config.colorscheme != "" && config.colorscheme != null) {
-    extraConfigVim = ''
-      colorscheme ${config.colorscheme}
+    extraConfigLuaPost = ''
+      vim.cmd.colorscheme("${config.colorscheme}")
     '';
   };
 }


traxys:
I'm not sure I want to unilaterally change the colorscheme setup, as I believe it's better to set them up early in the init.lua.

I think what we should do is create a new option `extraColorschemeLua` that would contain all the code for the `setup` of the selected colorscheme(s) and _after_ this section we could add the `vim.cmd.colorscheme("${config.colorscheme}")` statement.

GaetanLepage:
> Without that, e.g. require('gruvbox').setup() was run _after_ setting the colorsheme. Thus all settings were ignored.

[I am rewriting all of the `colorschemes.*` plugin modules](https://github.com/nix-community/nixvim/pulls?q=is%3Apr+colorschemes+author%3AGaetanLepage).
With the new method, this ordering problem should not exist anymore.

Could you test if `gruvbox` now works fine please ?

mxkrsv:
> [I am rewriting all of the `colorschemes.*` plugin modules](https://github.com/nix-community/nixvim/pulls?q=is%3Apr+colorschemes+author%3AGaetanLepage). With the new method, this ordering problem should not exist anymore.
> 
> Could you test if `gruvbox` now works fine please ?

Works now, thanks!

PR #1138, opened 2024-02-20
plugins/lsp: add typos-lsp

tarantoj:
Adds the typos-lsp


plugins/lsp/default.nix
```diff
@@ -90,6 +90,7 @@ in {
         "terraformls"
         "texlab"
         "tsserver"
+        "typos-lsp"
         "typst-lsp"
```
GaetanLepage, line 94:
```suggestion
        "tsserver"
        "typst-lsp"
```

This is for deprecation warnings, as this is a new server, no need to mention it here.

tarantoj:
Ah shoot, sorry. I had those changes locally and thought I pushed them up. Probably need to make a rule for myself not to contribute late at night haha.

Will get this fixed up ASAP!

plugins/lsp/language-servers/default.nix
```diff
@@ -596,6 +596,12 @@ with lib; let
       description = "tsserver for TypeScript";
       package = pkgs.nodePackages.typescript-language-server;
     }
+    {
+      name = "typos-lsp";
+      serverName = "typos_lsp";
+      description = "Source code spell checker for Visual Studio Code and LSP clients";
+      package = pkgs.typos-lsp;
```
GaetanLepage, line 603:
```suggestion
```
Actually, you don't even need this as we use `pkgs.${name}` for `package` by default.

tarantoj: Ah I thought that was the case!

GaetanLepage (approved):
LGTM ! Thanks

Issue #1143, opened 2024-02-21
[PLUGIN REQUEST] indent-o-matic 


alisonjenkins:
| Field    | Description |
|----------|-------------|
| Plugin   | `indent-o-matic `  | 
| Homepage | `https://github.com/Darazaki/indent-o-matic`     | 
| Nixpkgs  | `vimPlugins.indent-o-matic` | 

## Extra Information

Dumb automatic fast indentation detection for Neovim written in Lua.

PR #1144, opened 2024-02-21
plugins/indent-o-matic: init

alisonjenkins:
Adds indent-o-matic will resolve: https://github.com/nix-community/nixvim/issues/1143


traxys:
You are not forced to create nix options for every upstream plugin option. You can simply add it to the example to show that it can be done. `settings` is a freeformType, this means that users can add arbitrary keys even if they are not defined in nix.

traxys (approved):

PR #1150, opened 2024-02-22
Add ruff source to none-ls

nfelber:
Thank you for the amazing project :)

This adds the ruff linter and formatter as a source for none-ls.
I basically copied similar previous PRs and tested locally, so please let me know if you see any issue.

### DIFF
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -93,6 +93,9 @@ with lib; let
       staticcheck = {
         pacakge = pkgs.go-tools;
       };
+      ruff = {
+        package = pkgs.ruff;
+      };
       statix = {
         package = pkgs.statix;
       };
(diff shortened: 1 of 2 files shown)


plugins/none-ls/servers.nix
```diff
@@ -93,6 +93,9 @@ with lib; let
       staticcheck = {
         pacakge = pkgs.go-tools;
       };
+      ruff = {
+        package = pkgs.ruff;
+      };
```
GaetanLepage, line 98:
```suggestion
      ruff = {
        package = pkgs.ruff;
      };
```
Please, preserve ordering (`staticcheck` should go below `ruff`)

tests/test-sources/plugins/none-ls.nix
```diff
@@ -71,6 +71,7 @@
           ltrs.enable = true;
           markdownlint.enable = true;
           shellcheck.enable = true;
+          ruff.enable = true;
```
GaetanLepage, line 74:
```suggestion
          ruff.enable = true;
          shellcheck.enable = true;
```

Same here

plugins/none-ls/servers.nix
nfelber: Sorry I missed that! Fixed

GaetanLepage (approved):
All good !
Thanks for your contribution :)

PR #1159, opened 2024-02-26
fix-kanagawa-colorscheme

adalessa:
The theme is not applied to the colorscheme. for example setting the theme `dragon` has no effect since always the colorscheme was set to `kanagawa`.

I went to the approach of just checking to not be null and use the selected theme.

### DIFF
--- a/plugins/colorschemes/kanagawa.nix
+++ b/plugins/colorschemes/kanagawa.nix
@@ -188,7 +188,10 @@ in {
       // cfg.extraOptions;
   in
     mkIf cfg.enable {
-      colorscheme = "kanagawa";
+      colorscheme =
+        if cfg.theme == null
+        then "kanagawa"
+        else "kanagawa-${cfg.theme}";
 
       extraPlugins = [cfg.package];
 


traxys:
I find this perplexing because the kanagawa readme mentions that we should be able to do this by passing `theme` to the setup function, which we seem to do (https://github.com/nix-community/nixvim/blob/21512692ebf174e038351e991ebe77beb9e25a98/plugins/colorschemes/kanagawa.nix#L183)

adalessa:
Thanks for pointing that out.
I think the problem is other now.
Yes the theme is being inherit. but from the plugin readme
```
theme = "wave",              -- Load "wave" theme when 'background' option is not set
    background = {               -- map the value of 'background' option to a theme
        dark = "wave",           -- try "dragon" !
        light = "lotus"
    },
```
If the background is set will not be use.

in the existing code there is a default set 
```
background = {
          light = helpers.defaultNullOpts.mkStr "lotus" ''
            The theme to use when `vim.o.background = "light"`.
          '';

          dark = helpers.defaultNullOpts.mkStr "wave" ''
            The theme to use when `vim.o.background = "dark"`.
          '';
        };
```
So the theme is not properly applied.

since the default is wave the theme is not taking effect.

Possible solutions that I can think is remove default for background and add a validation letting the user know that they can not be set both properties at the same time. 

@traxys what do you think ?

traxys:
Those should default to `null` (even if they mention a default, it's not the nix default!). I agree that we need some validation to prevent both theme & background from being set

adalessa:
I have put some time investigating this.
In order to the theme from the configuration to be taken into effect background needs to be not set, not just not be present, if not present the default values takes effect which makes the `wave` theme. This is not in nixvim it's more in the plugin side.
https://github.com/rebelot/kanagawa.nvim/blob/master/lua/kanagawa/init.lua#L26

For this is required to send the `background = {}` but I haven't been able to achieve that. Any suggestion is welcome.

traxys:
We should have an helper for that, helpers.emptyTable (in https://nix-community.github.io/nixvim/user-guide/helpers.html)

GaetanLepage:
Hasn't this been fixed by #1269 ?

MattSturgeon:
> Hasn't this been fixed by #1269 ?

With that PR there's still the caveat mentioned in the docs:

> You can select the theme in two ways:
      - Set `colorschemes.kanagawa.settings.theme` AND explicitly unset `vim.o.background` (i.e. `options.background = ""`).
      - Set `colorschemes.kanagawa.settings.background` (the active theme will depend on the value of `vim.o.background`).

Although this PR is kinda stale and would need updating to use `settings`, etc, anyway.

Closing for now, someone can always open a new PR if they want to take another stab at improving this.

Issue #1149, opened 2024-02-22
[PLUGIN REQUEST] crates.nvim


alisonjenkins:
| Field    | Description |
|----------|-------------|
| Plugin   | `crates.nvim`  | 
| Homepage | `https://github.com/Saecki/crates.nvim`     | 
| Nixpkgs  | `vimPlugins.crates-nvim` | 

## Extra Information

The plugin has been partially added already but only as a completion source for cmp. It is currently lacking the overlays that show the current versions of the crates when editing a Cargo.toml as you can see in the video on the repo page: https://github.com/Saecki/crates.nvim

siph:
The overlays are working for me:
![Screenshot_20240222_103957](https://github.com/nix-community/nixvim/assets/6619112/12ebd7f4-afa8-4aba-80e1-14021467004c)

alisonjenkins:
[...]

PR #1165, opened 2024-02-27
plugins/crates-nvim: init

alisonjenkins:
Adds crates-nvim as a plugin instead of just a completion source for cmp.

Resolves: https://github.com/nix-community/nixvim/issues/1149

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -118,6 +118,7 @@
     ./utils/commentary.nix
     ./utils/conjure.nix
     ./utils/coverage.nix
+    ./utils/crates-nvim.nix
     ./utils/cursorline.nix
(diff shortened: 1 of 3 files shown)


traxys (changes requested):
Because it can be a cmp source in addition of a plugin could you add a test enabling the plugin in cmp too?

Eveeifyeve:
This PR Is a copy of #1205

Eveeifyeve:
> Because it can be a cmp source in addition of a plugin could you add a test enabling the plugin in cmp too?

A PR has been made for this [here](1205)

GaetanLepage:
FWIW I had opened a PR a while back: https://github.com/nix-community/nixvim/pull/1205

PR #1170, opened 2024-02-29
plugins/lightline: Add option for inactive components

DoctorDalek1963:
The lightline plugin has an option for which components to display in active panes, which is currently supported by nixvim, but it also has an equivalent option for inactive panes. This PR just adds that option.

### DIFF
--- a/plugins/statuslines/lightline.nix
+++ b/plugins/statuslines/lightline.nix
@@ -76,15 +76,15 @@ in {
         default = null;
         type = types.nullOr (types.submodule {
           options = let
-            listType = with types; nullOr (listOf (listOf str));
+            listType = with helpers.nixvimTypes; maybeRaw (listOf (listOf str));
           in {
-            left = mkOption {
+            left = helpers.mkNullOrOption {
               type = listType;
               description = "List of components that will show up on the left side of the bar";
               default = null;
             };
 
-            right = mkOption {
+            right = helpers.mkNullOrOption {
               type = listType;
               description = "List of components that will show up on the right side of the bar";
               default = null;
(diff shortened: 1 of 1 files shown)


plugins/statuslines/lightline.nix
```diff
@@ -72,6 +72,27 @@ in {
         });
       };
 
+      inactive = mkOption {
+        default = null;
+        type = types.nullOr (types.submodule {
+          options = let
+            listType = with types; nullOr (listOf (listOf str));
```
traxys, line 79: listType = with helpers.nixvimTypes; maybeRaw (listOf (listOf str)).

This allows users to add lua code everywhere.

plugins/statuslines/lightline.nix
```diff
@@ -72,6 +72,27 @@ in {
...
+          in {
+            left = mkOption {
```
traxys, line 81: Prefer using helpers.mkNullOrOption

plugins/statuslines/lightline.nix
```diff
@@ -72,6 +72,27 @@ in {
...
+              type = listType;
+              description = "List of components that will show up on the left side of the bar";
+              default = null;
+            };
+
+            right = mkOption {
```
traxys, line 87: same

traxys (changes requested):

DoctorDalek1963:
I'm very new to Nix, so I just copied this from the `active` option that was already there. Should I apply these changes to that option as well or should that be a separate PR?

traxys:
No you can leave that to another PR

traxys:
Be carefull `helpers.mkNullOrOption` does not take the same arguments as `mkOption`. You can search in the code base for uses (or look at the definition in lib/options.nix)

traxys (approved):

Issue #797, opened 2023-12-09
[New module] lazy.nvim


pta2002:
**Plugin**: lazy.nvim
**Plugin repository**: https://github.com/folke/lazy.nvim
**Is the plugin on nixpkgs**: [Yes](https://github.com/NixOS/nixpkgs/blob/b4372c4924d9182034066c823df76d6eaf1f4ec4/pkgs/applications/editors/vim/plugins/generated.nix#L4789)

## Extra information
Following the discussions around #421, one thing has become clear: lazy.nvim already implements pretty much everything we need.

Therefore, instead of reinventing the wheel, the best course of action starts with adding a module to allow for plugin management via `lazy.nvim`. This can later be implemented to load plugins from `extraPlugins`, but at first it can just be a separate option from the module, given that supporting that would require a flag to tell nixvim to not install plugins the other way.

Either way, this should follow the same general idea as the [packer module](https://github.com/nix-community/nixvim/blob/main/plugins/pluginmanagers/packer.nix). Since `lazy.nvim` supports loading plugins from a directory (via the `dir` option), we can sidestep its plugin downloading completely, too, keeping the config reproducible :) (Though we should still allow a user to have `lazy.nvim` download plugins, in order to let them easily install out-of-tree plugins. However, IMO everything officially supported by us should come from Nix, since it's not that much extra work to package).

Eventually™ we can fork out lazy.nvim's lazy-loading code to a separate library, and have a module for that, to make builds smaller. But as I said in the beginning, no need to reinvent the wheel for now! Let's just use what works and see how it goes.

ehsan2003:
I kinda guess that reinventing the wheel in this case is pretty much appropriate

pta2002:
@ehsan2003 there was a fairly long discussion about this in #421, and the general consensus was basically "lazy.nvim already does this and we can use it to try out the idea at first and then eventually roll our own", but obviously there's merits to doing this ourselves. If you have any insights feel free to leave them in the original feature issue

ck3mp3r:
I've started experimenting with creating a plugin manager for lazy.nvim. I've taken the current packer.nix implementation as inspiration as I am not that familiar with the internals just yet.
Still a fair amount of work left (need to implement some more of the lazy.nvim basic properties), but this is a working version.
https://github.com/ck3mp3r/nixvim/pull/1/files
[...]

Issue #421, opened 2023-06-11
[Feature] Lazy loading


GaetanLepage:
-->Summary: https://notes.glepage.com/rA_fyfKESvC9OxGGNHOYXw#Lazy-Loading <--

Currently, nixvim relies on the neovim built-in way of loading plugins.
This is implemented by nixpkgs neovim's module. It has the advantage of existing and leveraging the collection of plugins packaged in nixpkgs.

With that said, the very popular [`lazy.nvim`](https://github.com/folke/lazy.nvim) plugin manager proposes a more performant way of loading plugins.
Imitating their lazy loading process could be a very valuable addition to nixvim.
I would guess that this feature should be optional so that conventional loading would stay the default.

As of today, no-one has started to work on this issue.
Feel free to go and make a PR, even if the work is in draft-stage. I will personally be glad to review and discuss a proposed implementation.

pta2002:
A (relatively) easy way to do this, compared to `lazy.nvim`'s approach of basically redoing the whole lazy-loading machinery in neovim, is to make use of optional plugin loading - this is supported, I don't really remember how, but it's a matter of checking how `wrapNeovim` works.

This way, we can `:packadd` the packages only when they're needed - AFAIK this is basically how plugin lazy-loading has been done in the past.

Second, probably easier, option - the vast majority of the plugins we support are lua-based plugins. The majority of the loading time there is spent on the `setup` function. What if we just avoid calling that function until it's needed?

Either way, I will be looking into how lazy-loading in neovim has been traditionally done, since I don't have much experience with it, and will try to come up with some way it could work.

pta2002:
Also, another feature `lazy.nvim` supports is bytecode pre-compilation. This is actually totally something we could do! Basically just need to wrap the plugin's derivations to add that. Something to look into, but I'd say this is better off as a separate feature.

pta2002:
Some references:

- [Examples of lazyloading (Neovim discourse)](https://neovim.discourse.group/t/examples-of-lazyloading/2897) - Basically, create some convenience functions for delaying lazy-loading, and then bind them that way. We have all the machinery required right now - we can create autocommands, rebind commands and rebind keys. With this, we can just shim a plugin's commands and `:packadd` the plugin when they're called, or wait for an autocommand to load them.

GaetanLepage:
> Either way, I will be looking into how lazy-loading in neovim has been traditionally done, since I don't have much experience with it, and will try to come up with some way it could work.

[...]

PR #1175, opened 2024-03-01
plugins/lazy.nvim: new plugin manager

ck3mp3r:
This PR addresses #797

### DIFF
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -49,31 +49,53 @@ in {
 
               name = helpers.mkNullOrOption str "Name of the plugin to install";
 
-              dev = helpers.defaultNullOpts.mkBool false "When true, a local plugin directory will be used instead. See config.dev";
+              dev = helpers.defaultNullOpts.mkBool false ''
+                When true, a local plugin directory will be used instead.
+                See config.dev
+              '';
 
-              lazy = helpers.defaultNullOpts.mkBool true "When true, the plugin will only be loaded when needed. Lazy-loaded plugins are automatically loaded when their Lua modules are required, or when one of the lazy-loading handlers triggers";
+              lazy = helpers.defaultNullOpts.mkBool true ''
+                When true, the plugin will only be loaded when needed.
+                Lazy-loaded plugins are automatically loaded when their Lua modules are required,
+                or when one of the lazy-loading handlers triggers
+              '';
 
-              enabled = helpers.defaultNullOpts.mkBool true "When false then this plugin will not be included in the spec";
+              enabled = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`" ''
+                When false then this plugin will not be included in the spec. (accepts fun():boolean)
+              '';
 
               cond =
-                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
-                "When false, or if the function returns false, then this plugin will not be loaded. Useful to disable some plugins in vscode, or firenvim for example.";
+                helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`"
+                ''
+                  When false, or if the function returns false,
+                  then this plugin will not be loaded. Useful to disable some plugins in vscode,
+                  or firenvim for example. (accepts fun(LazyPlugin):boolean)
+                '';
 
               dependencies =
                 helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str listOfPlugins)
                 "Plugin dependencies";
 
               init =
-                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
+                helpers.mkNullOrLuaFn
                 "init functions are always executed during startup";
 
-              config =
-                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
-                "config is executed when the plugin loads. The default implementation will automatically run require(MAIN).setup(opts). Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name. See also opts. To use the default implementation without opts set config to true.";
+              config = helpers.mkNullOrStrLuaFnOr (types.enum [true]) ''
+                config is executed when the plugin loads.
+                The default implementation will automatically run require(MAIN).setup(opts).
+                Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name.
+                See also opts. To use the default implementation without opts set config to true.
+              '';
 
-              main = helpers.mkNullOrOption str "You can specify the main module to use for config() and opts(), in case it can not be determined automatically. See config()";
+              main = helpers.mkNullOrOption str ''
+                You can specify the main module to use for config() and opts(),
+                in case it can not be determined automatically. See config()
+              '';
 
-              submodules = helpers.defaultNullOpts.mkBool true "When false, git submodules will not be fetched. Defaults to true";
+              submodules = helpers.defaultNullOpts.mkBool true ''
+                When false, git submodules will not be fetched.
+                Defaults to true
+              '';
 
               event = with helpers.nixvimTypes;
                 helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
@@ -91,14 +113,29 @@ in {
                 helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
                 "Lazy-load on key mapping";
 
-              module = helpers.defaultNullOpts.mkBool false "Do not automatically load this Lua module when it's required somewhere";
+              module = helpers.mkNullOrOption (enum [false]) ''
+                Do not automatically load this Lua module when it's required somewhere
+              '';
 
-              priority = helpers.mkNullOrOption number "Only useful for start plugins (lazy=false) to force loading certain plugins first. Default priority is 50. It's recommended to set this to a high number for colorschemes.";
+              priority = helpers.mkNullOrOption number ''
+                Only useful for start plugins (lazy=false) to force loading certain plugins first.
+                Default priority is 50. It's recommended to set this to a high number for colorschemes.
+              '';
 
-              optional = helpers.defaultNullOpts.mkBool false "When a spec is tagged optional, it will only be included in the final spec, when the same plugin has been specified at least once somewhere else without optional. This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part of the user's plugins";
+              optional = helpers.defaultNullOpts.mkBool false ''
+                When a spec is tagged optional, it will only be included in the final spec,
+                when the same plugin has been specified at least once somewhere else without optional.
+                This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part
+                of the user's plugins
+              '';
 
               opts = with helpers.nixvimTypes;
-                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) "opts should be a table (will be merged with parent specs), return a table (replaces parent specs) or should change a table. The table will be passed to the Plugin.config() function. Setting this value will imply Plugin.config()";
+                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) ''
+                  opts should be a table (will be merged with parent specs),
+                  return a table (replaces parent specs) or should change a table.
+                  The table will be passed to the Plugin.config() function.
+                  Setting this value will imply Plugin.config()
+                '';
             };
           });
 
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -56,12 +56,7 @@ in {
               enabled = helpers.defaultNullOpts.mkBool true "When false then this plugin will not be included in the spec";
 
               cond =
-                helpers.mkNullOrOption
-                (oneOf [
-                  str
-                  helpers.nixvimTypes.rawLua
-                  (listOf (either str helpers.nixvimTypes.rawLua))
-                ])
+                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
                 "When false, or if the function returns false, then this plugin will not be loaded. Useful to disable some plugins in vscode, or firenvim for example.";
 
               dependencies =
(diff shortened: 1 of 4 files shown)


plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                ])
+                "Lazy-load on key mapping";
+
+              module = helpers.defaultNullOpts.mkBool false "Do not automatically load this Lua module when it's required somewhere";
+
+              priority = helpers.defaultNullOpts.mkNum null "Only useful for start plugins (lazy=false) to force loading certain plugins first. Default priority is 50. It's recommended to set this to a high number for colorschemes.";
+
+              optional = helpers.defaultNullOpts.mkBool false "When a spec is tagged optional, it will only be included in the final spec, when the same plugin has been specified at least once somewhere else without optional. This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part of the user's plugins";
+
+              opts = helpers.mkNullOrOption (either attrs helpers.nixvimTypes.rawLua) "opts should be a table (will be merged with parent specs), return a table (replaces parent specs) or should change a table. The table will be passed to the Plugin.config() function. Setting this value will imply Plugin.config()";
```
traxys, line 129: helpers.mabyeRaw (attrsOf str)

Please avoir using `attrs`, and instead use `attrsOf str`, merging is handling better.

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                helpers.mkNullOrOption
+                (oneOf [
+                  (either str (listOf str))
+                  helpers.nixvimTypes.rawLua
+                ])
+                "Lazy-load on key mapping";
+
+              module = helpers.defaultNullOpts.mkBool false "Do not automatically load this Lua module when it's required somewhere";
+
+              priority = helpers.defaultNullOpts.mkNum null "Only useful for start plugins (lazy=false) to force loading certain plugins first. Default priority is 50. It's recommended to set this to a high number for colorschemes.";
```
traxys, line 125: Don't use a `null` as a default with `defaultNullOpts`, pleas use the `mkNullOrOption` directly

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                helpers.mkNullOrOption
+                (oneOf [
+                  (either str (listOf str))
+                  helpers.nixvimTypes.rawLua
+                ])
+                "Lazy-load on filetype";
+
+              keys =
+                helpers.mkNullOrOption
+                (oneOf [
```
traxys, line 117: maybeRaw

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                  (either str (listOf str))
+                  helpers.nixvimTypes.rawLua
+                ])
+                "Lazy-load on command";
+
+              ft =
+                helpers.mkNullOrOption
+                (oneOf [
+                  (either str (listOf str))
```
traxys, line 110: maybeRaw

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                  helpers.nixvimTypes.rawLua
+                ])
+                "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
+
+              cmd =
+                helpers.mkNullOrOption
+                (oneOf [
+                  (either str (listOf str))
```
traxys, line 102: mabyeRaw

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+              dependencies =
+                helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str listOfPlugins)
+                "Plugin dependencies";
+
+              init =
+                helpers.mkNullOrOption (either bool helpers.nixvimTypes.rawLua)
+                "init functions are always executed during startup";
+
+              config =
+                helpers.mkNullOrOption (either bool helpers.nixvimTypes.rawLua)
```
traxys, line 76: maybeRaw bool

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+                  (listOf (either str helpers.nixvimTypes.rawLua))
+                ])
+                "When false, or if the function returns false, then this plugin will not be loaded. Useful to disable some plugins in vscode, or firenvim for example.";
+
+              dependencies =
+                helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str listOfPlugins)
+                "Plugin dependencies";
+
+              init =
+                helpers.mkNullOrOption (either bool helpers.nixvimTypes.rawLua)
```
traxys, line 72: maybeRaw bool

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+
+              dev = helpers.defaultNullOpts.mkBool false "When true, a local plugin directory will be used instead. See config.dev";
+
+              lazy = helpers.defaultNullOpts.mkBool true "When true, the plugin will only be loaded when needed. Lazy-loaded plugins are automatically loaded when their Lua modules are required, or when one of the lazy-loading handlers triggers";
+
+              enabled = helpers.defaultNullOpts.mkBool true "When false then this plugin will not be included in the spec";
+
+              cond =
+                helpers.mkNullOrOption
+                (oneOf [
```
traxys, line 60: maybeRaw (either ....)

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,216 @@
...
+        mkOption {
+          type = listOfPlugins;
+          default = [];
+          description = "List of plugins";
+        };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    extraPlugins = [(pkgs.vimPlugins.lazy-nvim.overrideAttrs (_: {pname = "lazy.nvim";}))];
```
traxys, line 144: Why do we override the name?
If it's important there should be a comment

traxys (changes requested):
Really nice!

I think we could improve a bit in the future the `opts` (or `config`, not sure) field by providing a function of  the form `helpers.initLuaFromModule ({...}: {#nixvim module}`,  but this is larger than the scope of simply lazy.nvim.

jla2000:
Would it somehow be possible to configure a plugin that already comes with nixvim (like eg telescope) using the options provided in nixvim, but then use lazy to load this plugin?

GaetanLepage:
> Would it somehow be possible to configure a plugin that already comes with nixvim (like eg telescope) using the options provided in nixvim, but then use lazy to load this plugin?

This is the goal in the long term but we are not there yet.
Discussion is happening in https://github.com/nix-community/nixvim/issues/421.

plugins/pluginmanagers/lazy.nix
ck3mp3r: The ops is a table, so (attrsOf str) will fail.
`error: A definition for option `plugins.lazy.plugins."[definition 1-entry 7]".opts.ensure_installed' is not of type `string'. Definition values:
       - In `<unknown-file>': { }`

ck3mp3r:
> I have thoughts on the 'build' part of Lazy though. Ideally the build would be unnecessary in Nixvim as it should have already been built and put into Nixpkgs as part of the packaging of dependencies. Is this not the case for plugins you have tested?

Totally, I should have come to that conclusion myself 😆

plugins/pluginmanagers/lazy.nix
traxys: Ah sorry, attrsOf anything

plugins/pluginmanagers/lazy.nix
ck3mp3r: Good question, will see what happens when I pull this.

plugins/pluginmanagers/lazy.nix
ck3mp3r: either I am doing something wrong or this doesn't quite work out:
```
error: attempt to call something which is not a function but a set

       at /nix/store/i8d20zf065xdkjydxhclcr517ya0d09g-source/plugins/pluginmanagers/lazy.nix:59:17:

           58|               cond =
           59|                 helpers.mkNullOrOption
             |                 ^
           60|                 helpers.nixvimTypes.maybeRaw (either (listOf str) str)

```

ck3mp3r: On second thought, reading the description again, I think this needs to be `maybeRaw bool` in any case...

plugins/pluginmanagers/lazy.nix
ck3mp3r: Removed the override, has no detrimental effect I can determine...

ck3mp3r:
@traxys is there a slightly more elegant way than doing this?
```
helpers.mkNullOrOption (either (maybeRaw str) (maybeRaw (listOf str)))
```
It does the trick but it has a certain "smell" :)

plugins/pluginmanagers/lazy.nix
traxys: You are missing some parentheses

traxys:
> @traxys is there a slightly more elegant way than doing this?
> 
> ```
> helpers.mkNullOrOption (either (maybeRaw str) (maybeRaw (listOf str)))
> ```
> 
> It does the trick but it has a certain "smell" :)

helpers.mkNullOrOption (maybeRaw (either str (listOf str))) ?

ck3mp3r:
I tried that before...

https://github.com/nix-community/nixvim/pull/1175#discussion_r1509987207

ck3mp3r:
Just slightly meta: the revisions to this PR were done in a nixvim instance that uses this plugin manager.
I think, unless there are any other remarks/suggestions, I'll mark this PR as ready to merge...

ck3mp3r:
@traxys can we land this soonish?

jla2000:
@ck3mp3r Was the "import" lazy.nvim option forgotten or intentionally left out?
I tried doing something like this, but there seems to be no nix option for that:
```nix
plugins.lazy = {
  enable = true;
  plugins = with pkgs.vimPlugins; [
  {
    pkg = LazyVim;
      import = "lazyvim.plugins";
    }
  ];
};
```

traxys (changes requested):
There are quite a lot of really long lines, mainly due to descriptions. Could you use `''` quotes, and split them on multiple lines?

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+                description = "Vim plugin to install";
+              };
+
+              name = helpers.mkNullOrOption str "Name of the plugin to install";
+
+              dev = helpers.defaultNullOpts.mkBool false "When true, a local plugin directory will be used instead. See config.dev";
+
+              lazy = helpers.defaultNullOpts.mkBool true "When true, the plugin will only be loaded when needed. Lazy-loaded plugins are automatically loaded when their Lua modules are required, or when one of the lazy-loading handlers triggers";
+
+              enabled = helpers.defaultNullOpts.mkBool true "When false then this plugin will not be included in the spec";
```
traxys, line 56: Could you use helpers.defaultNullOpts.mkStrLuaFnOr types.bool?

And mention in the description that it can take a `fun():boolean`

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+
+              cond =
```
traxys, line 58: Could you use helpers.defaultNullOpts.mkStrLuaFnOr types.bool and metion it can take a `fun(LazyPlugin):boolean`

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
+                "When false, or if the function returns false, then this plugin will not be loaded. Useful to disable some plugins in vscode, or firenvim for example.";
+
+              dependencies =
+                helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str listOfPlugins)
+                "Plugin dependencies";
+
+              init =
+                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
```
traxys, line 67: Shouldn't this be a helpers.mkNullOrLuaFn ?

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+                "init functions are always executed during startup";
+
+              config =
+                helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw bool)
```
traxys, line 71: This should be helpers.mkNullOrStrLuaFnOr (types.enum [true])

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+                "config is executed when the plugin loads. The default implementation will automatically run require(MAIN).setup(opts). Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name. See also opts. To use the default implementation without opts set config to true.";
+
+              main = helpers.mkNullOrOption str "You can specify the main module to use for config() and opts(), in case it can not be determined automatically. See config()";
+
+              submodules = helpers.defaultNullOpts.mkBool true "When false, git submodules will not be fetched. Defaults to true";
+
+              event = with helpers.nixvimTypes;
+                helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
```
traxys, line 79: The type seems pretty complex:
```
string? or string[] or fun(self:LazyPlugin, event:string[]):string[] or {event:string[]|string, pattern?:string[]|string}
```

That seems to translate to `let item = oneOf [str (listOf str) rawLua]; in either item (listOf item)`

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+              cmd = with helpers.nixvimTypes;
+                helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
+                "Lazy-load on command";
+
+              ft = with helpers.nixvimTypes;
+                helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
+                "Lazy-load on filetype";
+
+              keys = with helpers.nixvimTypes;
+                helpers.mkNullOrOption (maybeRaw (either str (listOf str)))
```
traxys, line 91: We could map the `LazyKeysSpec` submodule, but it's not really necessary because the user could do it with a `maybeRaw`

plugins/pluginmanagers/lazy.nix
```diff
@@ -0,0 +1,187 @@
...
+                "Lazy-load on key mapping";
+
+              module = helpers.defaultNullOpts.mkBool false "Do not automatically load this Lua module when it's required somewhere";
```
traxys, line 94: I think this is a helpers.mkNullOrOption (enum [false])

traxys (changes requested):

ck3mp3r:
I'll see when I have time to address these issues. For the time being I need to get back to my day job.

jla2000:
@ck3mp3r Just curious, how did you install treesitter grammars with the lazy.nvim setup? Does it even work at the moment?

ck3mp3r:
> > Just slightly meta: the revisions to this PR were done in a nixvim instance that uses this plugin manager.
> 
> @ck3mp3r Just curious, how did you install treesitter grammars with the lazy.nvim setup? Does it even work at the moment?

@jla2000 have a look here, this is my current testing branch. It works fine actually :) 
https://github.com/ck3mp3r/xvim/blob/lazy/config/treesitter.nix

jla2000:
Ah I see, you have to add it to the runtime path again since lazy clears it :) Thanks!

ck3mp3r:
> There are quite a lot of really long lines, mainly due to descriptions. Could you use `''` quotes, and split them on multiple lines?

@traxys  This is what alejandra does with the formatting when I use '' ''....
```
              dev =
                helpers.defaultNullOpts.mkBool false
                ''                  When true, a local plugin directory will be used instead.
                                                                    See config.dev'';

```

plugins/pluginmanagers/lazy.nix
ck3mp3r:
```
cond = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "When false, or if the function returns false, then this plugin will not be loaded. Useful to disable some plugins in vscode, or firenvim for example.";

```
results in:
```
 error: An option declaration for `plugins.lazy.plugins.*.cond' has type
       `lambda' rather than an attribute set.
       Did you mean to define this outside of `options'?

```

Any idea how I need to define this?

plugins/pluginmanagers/lazy.nix
ck3mp3r: somehow the nix version reads easier than the description... yet doesn't quite "compile"

plugins/pluginmanagers/lazy.nix
ck3mp3r: How do you want to address this?

traxys:
> I'll see when I have time to address these issues. For the time being I need to get back to my day job. Alternatively we can merge this, it does the trick, and we can then refactor and improve, as I don't see any of the interfaces change the way this plugin is going to be used.
> 
> On a different note: I am not sure the abstraction being added by nixvim isn't doing more harm than good. The more I use it the more I want to keep things simple in the nix realm and allow configuration that isn't related to packaging and distribution to be just handled by the neovim ecosystem...

There are two main ways to write nixvim modules, the way this PR does it (the "old" way) where we type every option strictly, but there is another way we are starting to look into, inspired by the NixOS RFC 42, that uses freeform options and a lot less magic

plugins/pluginmanagers/lazy.nix
traxys: What  do you mean it doesn't "compile" ?

plugins/pluginmanagers/lazy.nix
traxys: Leave it like that, we'll see if we get complaints

plugins/pluginmanagers/lazy.nix
traxys: Though I think I mistyped the option, it looks like more of a `oneOf [str (listOf str) rawLua (submodule  {event = ...; pattern= ...;})];`, but we could skip the submodule, and simply use `oneOf [str (listOf str) rawLua]`.

You can choose what you prefer

PR #1176, opened 2024-03-01
plugins/goyo: switch to settings option

GaetanLepage:


traxys (changes requested):
mkNullOr -> mkNullOrOption

PR #1178, opened 2024-03-01
flake/dev: add a convenient devshell

GaetanLepage:
Create a convenient devshell that provides the following commands:
- checks: Run all nixvim checks
- docs: Build nixvim documentation
- format: Format the entire codebase
- menu: prints this menu
- serve-docs: Build and serve documentation locally
- tests: Run nixvim tests

### DIFF
--- a/flake-modules/dev/devshell.nix
+++ b/flake-modules/dev/devshell.nix
@@ -1,0 +1,66 @@
+{inputs, ...}: {
+  imports = [
+    inputs.devshell.flakeModule
+  ];
+
+  perSystem = {
+    pkgs,
+    config,
+    system,
+    ...
+  }: {
+    devshells.default = {
+      devshell.startup.pre-commit.text = config.pre-commit.installationScript;
+
+      commands = let
+        # Thanks to this, the user can choose to use `nix-output-monitor` (`nom`) instead of plain `nix`
+        nix = "$([ '$\{NIXVIM_NOM:-0}' = '1' ] && echo ${pkgs.lib.getExe pkgs.nom} || echo nix)";
+      in [
+        {
+          name = "checks";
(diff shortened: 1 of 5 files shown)


GaetanLepage:
`nom` is great to visualize like builds like those but it has the drawback of not showing warnings...
Maybe we should stick to plain `nix`...

flake-modules/dev/devshell.nix
```diff
@@ -0,0 +1,67 @@
...
+    pkgs,
+    config,
+    system,
+    ...
+  }: {
+    devshells.default = {
+      devshell.startup.pre-commit.text = config.pre-commit.installationScript;
+
+      packages = with pkgs; [
+        python3
```
traxys, line 16: I don't think  we should add either `python3` or `nix-output-monitor` to the packages. We should simply reference their paths in the commands (`"${lib.getExe pkgs.python3}"`).

This would risk poluting what people are using

traxys (changes requested):

flake-modules/dev/devshell.nix
GaetanLepage: Done

flake-modules/dev/devshell.nix
```diff
@@ -0,0 +1,71 @@
...
+  perSystem = {
+    pkgs,
+    config,
+    system,
+    ...
+  }: {
+    devshells.default = {
+      devshell.startup.pre-commit.text = config.pre-commit.installationScript;
+
+      packages = with pkgs; [
```
traxys, line 15: Please remove  entirely the `packages`!

traxys (changes requested):

traxys (approved):

PR #1179, opened 2024-03-01
plugins/emmet: switch to settings option

GaetanLepage:


plugins/utils/emmet.nix
```diff
@@ -12,23 +12,42 @@ with helpers.vim-plugin;
...
+    # TODO introduced 2024-03-01: remove 2024-05-01
     deprecateExtraConfig = true;
+    optionsRenamedToSettings = [
+      "mode"
+    ];
+    imports = [
+      (
+        mkRenamedOptionModule
+        ["plugins" "emmet" "leader"]
+        ["plugins" "emmet" "settings" "leader"]
```
traxys, line 25: shouldn't this be `leader_key` in the second list?

traxys (changes requested):

plugins/utils/emmet.nix
GaetanLepage: Yes indeed... good catch

traxys (approved):

PR #1169, opened 2024-02-28
plugin/none-ls: Implement all builtins

traxys:
This is done by parsing the upstream provided JSON document, and creating a big attribute set mapping packages.
Many builtins don't have a nixpkgs package, we force the user to acknowledge it and install the package externally.

Closes #768 as we always have nix options for all the none-ls sources

PR #1182, opened 2024-03-01
plugins/none-ls: fix withArgs type, syntax

robhanlon22:
This was previously null or string, but was changed to null or Lua in nix-community/nixvim#1169. Internally, `withArgs` is expected to be a string, but `mkNullOrLua` applies `mkRaw`, leading to a type error when generating the output, as it is attempting to interpolate a set rather than a string.

Additionally, there was an errant extra } when interpolating the args, which led to a broken init.lua.

Tested locally, and verified that it fixes my setup.

### DIFF
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -383,7 +383,7 @@ in {
             withArgs =
               if source.withArgs == null
               then sourceItem
-              else "${sourceItem}.with(${source.withArgs}})";
+              else "${sourceItem}.with(${source.withArgs})";
           in
             helpers.mkRaw ''
               require("null-ls").builtins.${withArgs}
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -331,9 +331,7 @@ in {
         (source: _:
           {
             enable = mkEnableOption "the ${source} ${sourceType} source for none-ls";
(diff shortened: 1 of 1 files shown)


plugins/none-ls/servers.nix
```diff
@@ -331,7 +331,7 @@ in {
         (source: _:
           {
             enable = mkEnableOption "the ${source} ${sourceType} source for none-ls";
-            withArgs = helpers.mkNullOrLua ''
```
traxys, line 334: the right type  should be `helpers.mkNullOrOption helpers.nivimTypes.strLua`! Sorry for the issue.

traxys (changes requested):

plugins/none-ls/servers.nix
GaetanLepage: So the type is a "lua string" but we are not converting it to raw right ?

traxys: exactly, we pass it directly. We should improve it to take arbitrary lua values with a toLuaObject, but we can do that later.

traxys (approved):
Sorry for the regression

traxys (changes requested):
You are lacking a description, this is why tests are failing

traxys (approved):

PR #1184, opened 2024-03-02
plugins/ltex-extra: init

loicreynier:
Add an option to `plugins.lsp.servers.ltex` to enable the [LTeX_extra](https://github.com/barreiroleo/ltex_extra.nvim) plugins which enable LTeX LSP code action, external file handling and more features. 

I let this PR as a draft as I have not properly tested it yet and I am not sure if adding this plugin as an option to `plugins.lsp.servers.ltex` is the preferred way.

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -53,6 +53,7 @@
     ./languages/lean.nix
     ./languages/ledger.nix
     ./languages/lint.nix
+    ./languages/ltex-extra.nix
     ./languages/markdown-preview.nix
     ./languages/nix.nix
     ./languages/nvim-jdtls.nix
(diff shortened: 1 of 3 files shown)


traxys (commented):
We have some other plugins that do stuff like this, like clangd_extensions & rustacean.nvim, you could check how they do things.

traxys:
Sorry I have forgotten to mention that new plugins should try to use the `mkNeovimPlugin` helper (see for occurences). This will require a bit of a refactor :/

loicreynier:
> It looks like the upstream options are in `snake_case`: [barreiroleo/ltex_extra.nvim#configuration](https://github.com/barreiroleo/ltex_extra.nvim?tab=readme-ov-file#configuration)
> 
> We need the names to match for the module to work.

Thanks for the heads up. I just realized it testing it. I'll fix it right away.

loicreynier:
Tested locally, seems to work fine.

traxys (approved):

PR #1195, opened 2024-03-02
plugins/copilot-vim: switch to settings option

GaetanLepage:


plugins/completion/copilot-vim.nix
```diff
@@ -12,33 +12,72 @@ with helpers.vim-plugin;
...
+      "filetypes"
+      "proxy"
+    ];
 
-    options = {
-      nodeCommand = mkDefaultOpt {
-        global = "node_command";
+    settingsOptions = {
+      node_command = mkOption {
         type = types.str;
```
traxys, line 26: I know it was not the case before, but this should be a `nullOr; types.str`, maybe the user does not want to set it to use the environement provided nodejs

plugins/completion/copilot-vim.nix
```diff
@@ -12,33 +12,72 @@ with helpers.vim-plugin;
...
+    optionsRenamedToSettings = [
+      "nodeCommand"
+      "filetypes"
+      "proxy"
+    ];
 
-    options = {
-      nodeCommand = mkDefaultOpt {
-        global = "node_command";
+    settingsOptions = {
```
traxys, line 24: this is missing the `enabled` settings. This settings could be useful in some cases, for example when using extra files in the nixvim configuration for certain filetypes, or to dynamically enable/disable the plugin (plus it's a documented upstream option)

traxys (changes requested):

plugins/completion/copilot-vim.nix
GaetanLepage: The `enabled` setting is a `buffer` option, not `global` ([`b:copilot_enabled`](https://github.com/github/copilot.vim/blob/57a0115908895f465eb3476f03a0aaa7096e8fe1/doc/copilot.txt#L64-L67)). So, according to me, it doesn't make sense to declare this option here.

plugins/completion/copilot-vim.nix
GaetanLepage: Done !

plugins/completion/copilot-vim.nix
GaetanLepage: It is especially made to locally override the `filetypes` option.

traxys: Yes indeed!

plugins/completion/copilot-vim.nix
```diff
@@ -12,33 +12,70 @@ with helpers.vim-plugin;
...
+      "filetypes"
+      "proxy"
+    ];
 
-      filetypes = mkDefaultOpt {
-        type = with types; attrsOf bool;
-        description = ''
-          A dictionary mapping file types to their enabled status
+    settingsOptions = {
+      node_command = helpers.defaultNullOpts.mkStr "${pkgs.nodejs-18_x}/bin/node" ''
```
traxys, line 25: I don't think this  is  the right type, because this won't set the "${pkgs.nodejs-18_x}/bin/node" as the nix default, I think we still  want that

traxys (changes requested):

plugins/completion/copilot-vim.nix
GaetanLepage: Oh yes you are right !

GaetanLepage: Done

PR #1217, opened 2024-03-06
treewide: fix typos

loicreynier:
Found a few typos so I ran `typos` to fix them. Most of the typos were in option descriptions. Checks passed locally.

Also added a pre-commit hook for it in the devshell. Let me know if we should keep this hook or only merge the typo fixes.


plugins/utils/neogen.nix
```diff
@@ -112,53 +112,53 @@ in {
           - `"vsnip"` (https://github.com/hrsh7th/vim-vsnip)
       '';
 
-      placeholderHighligt = helpers.defaultNullOpts.mkStr "DiagnosticHint" ''
+      placeholderHighlight = helpers.defaultNullOpts.mkStr "DiagnosticHint" ''
```
traxys, line 115: We  need to introduce a warning that this option was renamed, even if it was a  typo :/ See how other renames are done

plugins/utils/vim-matchup.nix
```diff
@@ -78,7 +78,7 @@ with lib; {
         helpers.defaultNullOpts.mkInt 60
         "Adjust timeouts in milliseconds for matchparen highlighting";
 
-      deffered = {
+      deferred = {
```
traxys, line 81: You need to rename both sub options  in  this case

traxys (changes requested):

loicreynier:
Just added renames using `mkRenamedOptionModule`. It is what you intended? Haven't go the time to rebase into `main` yet.

traxys:
Yes exactly!

loicreynier:
Rebased into `main`.

traxys (approved):

PR #804, opened 2023-12-11
Add conform function support

alisonjenkins:

PR #1221, opened 2024-03-06
plugins/conform: add str type for formatAfterSave

jonboh:
this is the same change that was performed previously on formatOnSave in https://github.com/nix-community/nixvim/pull/804

With this change it is possible to define a function to be run in formatAfterSave. It makes possible implementing the recipe to run slow formatters asynchronously: https://github.com/stevearc/conform.nvim/blob/master/doc/recipes.md#automatically-run-slow-formatters-async

pd: I found the devshell to be very nice, easing formatting and testing, thanks! I think I'll integrate that as well in some of my projects :D

### DIFF
--- a/plugins/lsp/conform-nvim.nix
+++ b/plugins/lsp/conform-nvim.nix
@@ -40,9 +40,9 @@ in {
       formatOnSave =
         helpers.defaultNullOpts.mkNullable
         (
-          with types;
+          with helpers.nixvimTypes;
             either
-            str
+            strLuaFn
             (
               submodule {
                 options = {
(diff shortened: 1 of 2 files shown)


plugins/lsp/conform-nvim.nix
```diff
@@ -69,15 +69,23 @@ in {
...
-              default = true;
-              description = "See :help conform.format for details.";
-            };
-          };
-        })
+        helpers.defaultNullOpts.mkNullable
+        (
+          with types;
+            either
+            str
```
traxys, line 76: please use `with nixvimTypes; either [strLuaFn ....]` so that it is documented correctly.

traxys (changes requested):
Could you add a test in 'tests/test-sources/plugins/lsp/conform-nvim.nix' using the raw version?

plugins/lsp/conform-nvim.nix
jonboh: I've applied this change in saveOnSave as well

GaetanLepage (approved):

PR #1222, opened 2024-03-06
plugins/cmp: add rename warning for plugins.nvim-cmp.enable

GaetanLepage:


traxys (approved):
I did not think of that....

PR #1225, opened 2024-03-07
plugins/none-ls: automatic check for tools packages declaration

GaetanLepage:
Add two assertions to ensure that we declare exactly the tools that are supported upstream (no more, no less).

The big cleaning comes from https://github.com/nvimtools/none-ls.nvim/pull/77.

### DIFF
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -21,9 +21,6 @@ with lib; let
       alejandra
       asmfmt
       astyle
-      autoflake
-      beancount
-      beautysh
       bibclean
       biome
       buf
@@ -34,13 +31,11 @@ with lib; let
       codespell
       commitlint
       cppcheck
-      cpplint
       csharpier
       deadnix
       dfmt
       djhtml
       djlint
-      dprint
       erlfmt
       fantomas
       fish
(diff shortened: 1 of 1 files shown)


traxys (approved):

plugins/none-ls/servers.nix
```diff
@@ -376,6 +297,47 @@ in {
...
+          # Keep tool names which are not in upstream
+          (toolName: !(elem toolName upstreamToolNames))
+          localToolNames;
+      in [
+        {
+          assertion = (length undeclaredToolNames) == 0;
+          message = ''
+            Nixvim (plugins.none-ls): Some tools from upstream are not declared locally in `builtinPackages`.
+            -> [${concatStringsSep ", " undeclaredToolNames}]
+          '';
```
andrevmatos, line 330: Could this be a warning, instead of an error assertion?
getting the following on current unstable-small:
```
error:
Failed assertions:
- Nixvim (plugins.none-ls): Some tools from upstream are not declared locally in `builtinPackages`.
-> [puppet_lint, findent, gleam_format]
```

purepani: I'm currently getting a error even on normal unstable.

arjan-s: Same here

PR #1226, opened 2024-03-07
plugins/cmp: allow negative values in `col_offset`

mrtnvgr:
An update broke my configuration. I used a negative int in this option before, everything worked fine 🙂


GaetanLepage (approved):
Thank you for the fix, LGTM !

PR #1235, opened 2024-03-08
plugins/lsp: use `attrsOf anything` instead of `attrs` for extraOptions

traxys:
`attrs` is not merged correctly, this can introduce conflicts with efmls-configs for example when trying to register `efm` sources not defined by efmls-configs.

### DIFF
--- a/plugins/lsp/helpers.nix
+++ b/plugins/lsp/helpers.nix
@@ -77,7 +77,7 @@
 
           extraOptions = mkOption {
             default = {};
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):

GaetanLepage:
Eventually `with types; attrsOf anything` but I don't care that much.

traxys:
For short types like this I generaly avoid doing it, because it does not save much typing

PR #1236, opened 2024-03-09
plugins/project-nvim: take ownership of telescope integration

khaneliman:
Telescope extension option being toggled independently will throw an error in config trying to load a non existent plugin.


traxys:
I'd rather add a warning than toggling it explictely

khaneliman:
I'm not sure why, though. If I enable the extension manually I expected it to enable the plugin that it's dependent on for me. 

It's a bug currently to try and load an extension that doesn't exist and breaks your init.lua just enabling this alone. The other instances of extensions fetch the package for you, as well.

khaneliman:
I wrote the warning around the concept of you possibly supplying the project-nvim outside of the nixvim options, if that was  what you were thinking.  

```nix
{ lib
, config
, ...
}:
with lib; let
  cfg = config.plugins.telescope.extensions.project-nvim;
  project-nvim-cfg = config.plugins.project-nvim;
in
{
  options.plugins.telescope.extensions.project-nvim = {
    enable = mkEnableOption "project-nvim telescope extension";
  };

  config = mkIf cfg.enable {
    plugins.telescope.enabledExtensions = [ "projects" ];

    warnings =
      optional
        (
          cfg.enable && (!project-nvim-cfg.enable)
        )
        ''
          You have enabled the `project-nvim` extension in telescope.
          You need to provide the project-nvim plugin itself otherwise this option will break your config.

          - To enable this extension, set `plugins.project-nvim.enable` to `true`.
          - If you've supplied the plugin outside of Nixvim, to silence this warning, explicitly set the option to `false`.
        '';

    plugins.project-nvim.enable = true;
  };
}
```

traxys:
Well if it breaks I'd suggest adding an assertion (so an error) instead of a warning, but I'd prefer we avoid having too much magic.

Ultimately I think I'd prefer having an option in the project.nvim plugin options called `enableTelescope` that would do this, and avoid using the `config.plugins.telescope.extensions.project-nvim` option, maybe even deprecating this way to enable telescope extensions.

I feel like it is more appropriate to have project.nvim manage this (and maybe warn the user if enableTelescope is set to `true` but telescope is not enabled)

khaneliman:
> I feel like it is more appropriate to have project.nvim manage this (and maybe warn the user if enableTelescope is set to `true` but telescope is not enabled)

I agree with this sentiment. I'll go this direction.

khaneliman:
In testing, I noticed telescope is automatically enabled in yanky. So I had to mkForce disable to test the warning. Should that be removed to get rid of the magic you were talking about? 

I also saw that harpoon is doing an assertion instead of a warning for the telescope integration. Should we change that to a warning like this or make this one an assertion?

traxys (approved):

PR #1238, opened 2024-03-09
plugins/indent-blankline: swith to mkNeovimPlugin

GaetanLepage:


plugins/utils/indent-blankline.nix
```diff
@@ -5,87 +5,48 @@
...
+    luaName = "ibl";
+    defaultPackage = pkgs.vimPlugins.indent-blankline-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    # TODO introduced 2024-03-10: remove 2024-05-10
+    deprecateExtraOptions = true;
+    optionsRenamedToSettings = [
+      "debounce"
+      "viewportBuffer"
```
traxys, line 21: does this work with  the suboptions (viewportBuffer.min, viewportBuffer.max) ?

traxys (changes requested):

plugins/utils/indent-blankline.nix
GaetanLepage: The warning is triggered for the root option.
Maybe it is better to specify for each of the suboptions still.

traxys: I'd find this more re-assuring  (and more helpful too)

GaetanLepage: I agree, done.

traxys (approved):

PR #1249, opened 2024-03-12
plugins/lsp: add fortls

loicreynier:
Add `fortls` LSP plugin. 

This plugin features numerous options, some of which are experimental. At the moment, I haven't implemented any `settingsOptions` because I'm not sure it's necessary given that the LSP is entirely configured by its command line arguments (it could then be configured through `fortls.cmd = [ ... ]`. Plus, it can be configured via a configuration file. Adding the options in nixvim would add significant maintenance work. However, I'm not closed to the idea of adding the options if you think it's necessary (as mentioned in the contribution guideline).

As for the default options, I've chosen those selected by [`nvim-lspconfig`](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/fortls.lua) (except for the initialization notification one

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -212,6 +212,7 @@ with lib; let
     }
     {
       name = "fortls";
+      description = "fortls for Fortran";
       cmd = cfg: [
         "${cfg.package}/bin/fortls"
         "--hover_signature"
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -211,6 +211,17 @@ with lib; let
       cmd = cfg: ["${cfg.package}/bin/elixir-ls"];
     }
     {
+      name = "fortls";
+      package = pkgs.fortls;
+      cmd = cfg: [
+        "${cfg.package}/bin/fortls"
+        "--notify_init"
+        "--hover_signature"
+        "--hover_language=fortran"
+        "--use_signature_help"
+      ];
+    }
+    {
       name = "fsautocomplete";
       description = "fsautocomplete for F#";
       package = pkgs.fsautocomplete;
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -210,6 +210,16 @@ with lib; let
       package = pkgs.elixir-ls;
       cmd = cfg: ["${cfg.package}/bin/elixir-ls"];
     }
+    {
+      name = "fortls";
+      package = pkgs.fortls;
```
GaetanLepage, line 215:
```suggestion
```
As the package has the same name as the LSP, you can omit this line.

plugins/lsp/language-servers/default.nix
```diff
@@ -210,6 +210,15 @@ with lib; let
...
       cmd = cfg: ["${cfg.package}/bin/elixir-ls"];
     }
+    {
+      name = "fortls";
+      cmd = cfg: [
+        "${cfg.package}/bin/fortls"
+        "--hover_signature"
+        "--hover_language=fortran"
+        "--use_signature_help"
+      ];
```
GaetanLepage, line 220:
```suggestion
```

As this is the default `nvim-lspconfig` value for `cmd`, I am not sure that you have to add this...

GaetanLepage:
Could you please add a `description`:
```nix
description = "fortls for Fortran";
```

plugins/lsp/language-servers/default.nix
loicreynier: It's not exactly the default command, I remove the startup notification.

GaetanLepage: Ok fine !

GaetanLepage (approved):

traxys (approved):

PR #1250, opened 2024-03-12
colorschemes/base16: use mkVimPlugin

GaetanLepage:
This is, in my honest opinion, a way nicer module design to handle this plugin.


plugins/colorschemes/base16/default.nix
```diff
@@ -0,0 +1,118 @@
...
+    name = "base16";
+    isColorscheme = true;
+    originalName = "base16.nvim";
+    defaultPackage = pkgs.vimPlugins.base16-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    imports = [
+      (
+        mkRemovedOptionModule
```
traxys, line 21: Can't we use a mkRenamedOption?

plugins/colorschemes/base16/default.nix
```diff
@@ -0,0 +1,118 @@
...
+        then {
+          colorscheme = "base16-${cfg.colorscheme}";
+        }
+        # ...otherwise, call the `setup function`.
+        else {
+          # Explicitly set the `colorscheme` option to `null` as it has been set to
+          # `mkDefault "base16"` by `mkVimPlugin`.
+          colorscheme = null;
+          extraConfigLuaPre = ''
+            require('base16-colorscheme').setup(${helpers.toLuaObject cfg.customColorScheme})
```
traxys, line 114: Shouldn't this be cfg.colorscheme?

plugins/colorschemes/base16/default.nix
```diff
@@ -0,0 +1,118 @@
...
+    imports = [
+      (
+        mkRemovedOptionModule
+        ["colorschemes" "base16" "customColorScheme"]
+        "You can directly provide an attrs to the `colorschemes.base16.colorscheme` option."
+      )
+    ];
+
+    extraOptions = {
+      useTruecolor = mkOption {
```
traxys, line 28: This option seems pretty useless as it simply maps to a builtin neovim option. Shouldn't we deprecate it with a mkRenameOption to the correct neovim builtin option?

traxys (changes requested):

plugins/colorschemes/base16/default.nix
GaetanLepage: Sure.

plugins/colorschemes/base16/default.nix
GaetanLepage: Indeed, good idea !

plugins/colorschemes/base16/default.nix
```diff
@@ -0,0 +1,118 @@
...
+    in [
+      (
+        mkRenamedOptionModule
+        (basePluginPath ++ ["customColorScheme"])
+        (basePluginPath ++ ["colorscheme"])
+      )
+      (
+        mkRenamedOptionModule
+        (basePluginPath ++ ["useTruecolor"])
+        ["termguicolors"]
```
traxys, line 34: isn't this `["options" "termguicolors"]` ? Our very unfortunate mapping to neovim options...

traxys (changes requested):

traxys (approved):

PR #1251, opened 2024-03-12
helpers: refactor mk[Neo]vimPlugin for colorschemes handling

GaetanLepage:
This design is more flexible.

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -32,13 +32,15 @@ with lib; rec {
 
   mkNeovimPlugin = config: {
     name,
-    colorscheme ? false,
     maintainers,
     url ? defaultPackage.meta.homepage,
     imports ? [],
     # deprecations
     deprecateExtraOptions ? false,
     optionsRenamedToSettings ? [],
+    # colorscheme
+    isColorscheme ? false,
+    colorscheme ? name,
     # options
     originalName ? name,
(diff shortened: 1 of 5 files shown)


lib/neovim-plugin.nix
```diff
@@ -128,10 +130,8 @@ with lib; rec {
               require('${luaName}').setup(${toLuaObject cfg.settings})
             '';
           }
-          (optionalAttrs colorscheme {
-            # We use `mkDefault` here to let individual plugins override this option.
-            # For instance, setting it to `null` a specific way of setting the coloscheme.
-            colorscheme = lib.mkDefault name;
+          (optionalAttrs (isColorscheme && (colorscheme != null)) {
+            inherit colorscheme;
```
GaetanLepage, line 134: Otherwise, we can do:
```nix
(optionalAttrs isColorscheme {
  colorscheme = mkDefault colorscheme;
})
```
Then, the plugins that want to manually unset it would add 
```nix
colorscheme = null;
```
to their `extraConfig`.

@traxys which option do you prefer ?

traxys: I don't really care about either way

traxys (approved):

lib/neovim-plugin.nix
GaetanLepage: Ok fine !

PR #1252, opened 2024-03-12
plugins/gitsigns: Use the newer way to define lua functions for attach

### LINKED ISSUE #1116: [BUG] Example for plugins.gitsigns.onAttach is wrong / confusing
| Plugin       | `gitsigns`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

The example for the `onAttach` option of the `gitsigns` plugin in the documentation looks like this:

```
''
  {
    function = \'\'
      function(bufnr)
        if vim.api.nvim_buf_get_name(bufnr):match(<PATTERN>) then
          -- Don't attach to specific buffers whose name matches a pattern
          return false
        end
        -- Setup keymaps
        vim.api.nvim_buf_set_keymap(bufnr, 'n', 'hs', '<cmd>lua require"gitsigns".stage_hunk()<CR>', {})
        ... -- More keymaps
      end
    \'\'
  }
''
```

That makes it look like the option should be assigned a multiline string (which of course will fail when building). The documentation says that it should be `null` or a submodule, but as a very inexperienced Nix user I was confused, because I did not know how the submodule is supposed to look.

A better way to display the example could look like this:

```
{
  onAttach.function = ''
      function(bufnr)
        if vim.api.nvim_buf_get_name(bufnr):match(<PATTERN>) then
          -- Don't attach to specific buffers whose name matches a pattern
          return false
        end
        -- Setup keymaps
        vim.api.nvim_buf_set_keymap(bufnr, 'n', 'hs', '<cmd>lua require"gitsigns".stage_hunk()<CR>', {})
        ... -- More keymaps
      end
  ''
}
```

traxys:
Fixes #1116

### DIFF
--- a/plugins/git/gitsigns.nix
+++ b/plugins/git/gitsigns.nix
@@ -106,7 +106,9 @@ in {
         '';
       };
     onAttach = mkOption {
-      type = types.nullOr luaFunction;
+      # TODO remove luaFunction in type
+      # Introduced 12-03-2024, remove 12-05-2024
+      type = with helpers.nixvimTypes; nullOr (either strLuaFn luaFunction);
       default = null;
       description = ''
         Callback called when attaching to a buffer. Mainly used to setup keymaps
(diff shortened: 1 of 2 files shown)


GaetanLepage:
r

PR #1254, opened 2024-03-12
plugins/coq-thirdparty: Fix example & improve documentation

### LINKED ISSUE #1113: [BUG] `coq-thirdparty` is missing `acceptKey` option and string cases don't match with example
| Plugin       | `coq-thirdparty`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

In the example listed under [plugins.coq-thirdparty.sources](https://nix-community.github.io/nixvim/plugins/coq-thirdparty/sources.html#pluginscoq-thirdpartysources), there are two errors:
- `acceptKey` is not handled by nixvim
- `shortName` is actually `short_name`

```nix
programs.nixvim = {
  plugins.coq-nvim.enable = true;
  plugins.coq-thirdparty = {
    sources = [{
      acceptKey = "<c-f>";
      shortName = "COP";
      src = "copilot";
    }];
  };
}
```

traxys:
Fixes #1113

### DIFF
--- a/plugins/completion/coq-thirdparty.nix
+++ b/plugins/completion/coq-thirdparty.nix
@@ -35,8 +35,8 @@ in {
         };
       });
       description = ''
-        List of sources. Each source is a free form type, so additional settings like `accept_key`
-        may be specified even if they are not declared by nixvim
+        List of sources.
+        Each source is a free-form type, so additional settings like `accept_key` may be specified even if they are not declared by nixvim.
       '';
       default = [];
       example = [
--- a/plugins/completion/coq-thirdparty.nix
+++ b/plugins/completion/coq-thirdparty.nix
@@ -34,7 +34,10 @@ in {
           };
         };
       });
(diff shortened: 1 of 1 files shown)


plugins/completion/coq-thirdparty.nix
```diff
@@ -34,21 +34,24 @@ in {
           };
         };
       });
-      description = "List of sources";
+      description = ''
+        List of sources. Each source is a free form type, so additional settings like `accept_key`
+        may be specified even if they are not declared by nixvim
```
GaetanLepage, line 39:
```suggestion
        List of sources.
        Each source is a free-form type, so additional settings like `accept_key` may be specified even if they are not declared by nixvim.
```

Hopefully I said that I would stop nitpicking :')

GaetanLepage (approved):
LGTM !

Issue #1219, opened 2024-03-06
feature: register LSP binds on `LspAttach`


MattSturgeon:
Currently, LSP's keymap registration is done together with most other keymaps, as LSP keymaps actually get added to `programs.nixvim.keymaps` after a bit of processing.

It's common practice in neovim configs to only register LSP-related keymaps when the buffer is attached to a language server, so that vanilla keys can be overridden with LSP functions (etc).

This is demonstrated by [kickstart.nvim here](https://github.com/nvim-lua/kickstart.nvim/blob/66e2a5a425a4f3b2e4f9456d82b29718c24a8993/init.lua#L440-L446).

- This could be configured using a new option such as `plugins.lsp.keymaps.onAttach`

If this is supported, users may wish to add arbitrary keymaps when the LSP is attached, not just the `vim.lsp.*` functions. E.g. to make use of [telescope's lsp functions](https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#neovim-lsp-pickers).

- This could be configured by adding an "extra" list; `plugins.lsp.keymaps.extra`, which is a list of normal keymap submodules.
- Alternatively, users could add keymaps themselves, and manually set the keymap event to `"LspAttach"`.
- Ideally, both configuration methods would be available.

### Potential strategy

One way to approach this might be to add a "registered on event" attribute to the keymaps submodule...

When rendering keymaps to lua, first _grouped them_ using the new `event` attribute.

The "`null`" group would be registered the same way keymaps are now, since it represents keymaps with no `event`.

All other groups would be mapped to an `autoCmd` submodule, with `event` being the group's key and the `callback` function registering the group's keymaps like normal.

The LSP plugin could then take advantage of the new system by having it's keymaps use `event = "LspAttach"` by default. Similar to how `options.silent` is configured currently.

Draft implementation of grouping keymaps by event in [this comment](https://github.com/nix-community/nixvim/issues/1219#issuecomment-1984855570) below.

Bodleum:
I think I can tackle this next week.
The only hurdle I can think of right now is we would want `event` to be `null` or `string` or `list of string` just like `autoCmd.*.event`, and in the `list of string` case we would want the keymap to be registered in both groupings of `event` and I'm not sure how to implement that.

The basic `string` case I'm pretty sure I can get going though!

MattSturgeon:
> I think I can tackle this next week.

👀 

[...]

PR #1260, opened 2024-03-13
Keymaps: Add onEvent key to keymaps to load keymap on specified event

Bodleum:
What the title says.
Groups keymaps by new `onEvent` key and loads with an autocmd.
Allows for easy fix of #1219. Not done here because this PR adds a new feature.

This is definitely not the cleanest code - I've only used nix for configuration so far, so I'd appreciate help on that front! I'm also not sure about how good the tests are either.

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -19,27 +19,33 @@ with lib; {
         }
       ];
     };
+
+    keymapsOnEvents = mkOption {
+      type = types.attrsOf (types.listOf helpers.keymaps.mapOptionSubmodule);
+      default = {};
+      example = {
+        "InsertEnter" = [
+          {
+            key = "<C-y>";
+            action = ''require("cmp").mapping.confirm()'';
+            lua = true;
+          }
+          {
+            key = "<C-n>";
+            action = ''require("cmp").mapping.select_next_item()'';
+            lua = true;
+          }
+        ];
+      };
+      description = ''
+        Register keymaps on an event instead of when nvim opens.
+        Keys are the events to register on, and values are lists of keymaps to register on each event.
+      '';
+    };
   };
 
   config = {
     extraConfigLua = let
-      # Divide mappings if they have "onEvent" set or not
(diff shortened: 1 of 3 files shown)


modules/keymaps.nix
```diff
@@ -48,7 +71,30 @@ with lib; {
             vim.keymap.set(map.mode, map.key, map.action, map.options)
           end
         end
-        -- }}}
-      '';
+      ''
+      + foldl (a: b: a + "\n" + b)
+      ""
+      (
+        mapAttrsToList
```
traxys, line 78: We have a wrapper for auto commands. I think you should be able to use it instead of calling directly vim.api.nvim_create_autocmd

traxys (changes requested):

modules/keymaps.nix
Bodleum: Sorry if I'm being an idiot here but I don't know how to use the helpers in `lib/autocmd-helpers.nix`. Looks to me just options and types and things, not any machinery for actually defining autocmds.
Additionally, `modules/keymaps.nix` evaluates to just a string, so I don't know how to also set nixvim configs there as well.

traxys: You would be able to add in the `config` section the following:
```nix
   autoCmd = mapAttrsToList (
       event: mappings: {
            inherit event;
            group = helpers.mkRaw ''vim.api.nvim_create_augroup("nixvim_binds_${event}", { clear = true })'';
            callback = helpers.mkRaw ''
                 function()
                    do
                      local __nixvim_binds = ${helpers.toLuaObject (map normalizeMapping mappings)}
                      for i, map in ipairs(__nixvim_binds) do
                        vim.keymap.set(map.mode, map.key, map.action, map.options)
                      end
                    end
                 end
            '';
       }
   ) config.keymapsOnEvents;
```

Though maybe we need to allow `autoCmd.*.group` to be a raw lua string

traxys (approved):

PR #1261, opened 2024-03-13
plugins/alpha: Allow `terminal` to not set `val`

### LINKED ISSUE #1140: [BUG] Alpha terminal
| Plugin       | `alpha`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `24.05-pre` | 

## Description

Nixvim configuration for Alpha is incomplete for "alpha" layout. Each layout require a `val` attribute although Alpha `terminal` layout do not.

[Alpha documentation for the `terminal` layout](https://github.com/goolord/alpha-nvim/blob/1356b9ef31b985d541d94314f2cf73c61124bf1d/doc/alpha.txt#L166)

```nix
{
  plugins.alpha =
    {
      enable = true;
      layout = [
        {
          type = "terminal";
          command = "thisisfine";
          width = 46;
          height = 25;
          opts = {
            redraw = true;
          };
        }
      ];
    };
}

```

Result

```
error: A definition for option `plugins.alpha.layout."[definition 1-entry 3]".val' is not of type `string or signed integer or list of (string or attribute set of anything)'. Definition values:
```

traxys:
Fixes #1140

### DIFF
--- a/plugins/utils/alpha.nix
+++ b/plugins/utils/alpha.nix
@@ -22,9 +22,10 @@ with lib; let
         description = "Type of section";
       };
 
-      val = mkOption {
-        type = with helpers.nixvimTypes;
-          oneOf [
+      val =
+        helpers.mkNullOrOption
+        (with helpers.nixvimTypes;
+          nullOr (oneOf [
             # "button", "text"
             str
             # "padding"
@@ -36,10 +37,8 @@ with lib; let
               # "group"
               (attrsOf anything)
             ))
-          ];
-        default = null;
-        description = "Value for section";
-      };
+          ]))
+        "Value for section";
 
       opts = mkOption {
         type = with types; attrsOf anything;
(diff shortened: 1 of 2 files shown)


plugins/utils/alpha.nix
```diff
@@ -24,7 +24,7 @@ with lib; let
 
       val = mkOption {
```
GaetanLepage, line 25:
```suggestion
      val = helpers.mkNullOrOption
        (with helpers.nixvimTypes;
          oneOf [
            str
            int
            (listOf (either str (attrsOf anything)))
          ]
        )
        "Value for section."
```

GaetanLepage: This is a suggestion, I am not agains the `mkOption`.

traxys: I hesitated, but I feel like the mkNullOprOption would be more consistent

GaetanLepage (approved):

PR #1265, opened 2024-03-14
colorschemes/base16: set termguicolors to true

### LINKED ISSUE #1264: [BUG] base16 colorscheme not applied
| Plugin       | `base16`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description
When the base16 colorscheme is enabled as seen in the MRE, it does not apply itself as the Neovim colorscheme. Strangely, it _does_ apply to lualine. I've tried multiple base16 colorschemes, all of them fail to apply. Possibly related to #1189 or #1228 (though the onedark plugin works fine for me).

```nix
programs.nixvim = {
  enable = true;
  colorschemes.base16 = {
    enable = true;
    colorscheme = "tomorrow-night";
  };
}
```

r3n3gad3p3arl:
With the rewrite at c2cd3cb7a13c0677c49f46d55a8c65b50b7d9431, the option `useTruecolor` was removed in favor of `options.termguicolors`. `useTruecolor` was set to `true` by default, but this is not done after the removal of this option. This causes the colors to not be set properly. I have set `options.termguicolors` to `true` again to fix this.

Fixes #1264.

### DIFF
--- a/plugins/colorschemes/base16/default.nix
+++ b/plugins/colorschemes/base16/default.nix
@@ -103,6 +103,8 @@ with lib;
           plugins.airline.settings.theme = mkIf cfg.setUpBar "base16";
           plugins.lualine.theme = mkIf cfg.setUpBar "base16";
           plugins.lightline.colorscheme = null;
+
+          options.termguicolors = mkDefault true;
         }
         (mkIf (isString cfg.colorscheme) {
           colorscheme = "base16-${cfg.colorscheme}";


traxys:
Maybe we should instead of adding the option ourselves add a warning (or an asertion) if it is not set?

plugins/colorschemes/base16/default.nix
```diff
@@ -103,6 +103,7 @@ with lib;
           plugins.airline.settings.theme = mkIf cfg.setUpBar "base16";
           plugins.lualine.theme = mkIf cfg.setUpBar "base16";
           plugins.lightline.colorscheme = null;
+          options.termguicolors = true;
```
GaetanLepage, line 106:
```suggestion
          options.termguicolors = mkDefault true;
```

traxys (approved):

PR #1266, opened 2024-03-14
plugins/gitsigns: switch to mkNeovimPlugin

### LINKED ISSUE #1253: plugins/gitsigns: Replace `luaFunction` with `strLuaFn`
`luaFunction` was an old attempt at having a nice documentation. It was a bit confusing & not very explicit. Nowdays we have strLuaFn that fills the same role, but better

GaetanLepage:
Fixes #1253


plugins/git/gitsigns/options.nix
```diff
@@ -0,0 +1,428 @@
...
+      hl = "GitSignsAdd";
+      text = "┆";
+      numhl = "GitSignsAddNr";
+      linehl = "GitSignsAddLn";
+    };
+  };
+
+  worktrees = let
+    worktreeType = types.submodule {
+      options = {
```
traxys, line 69: shouldn't we set this as  a freeform  type?

plugins/git/gitsigns/options.nix
```diff
@@ -0,0 +1,428 @@
...
+  '';
+
+  show_deleted = helpers.defaultNullOpts.mkBool false ''
+    Show the old version of hunks inline in the buffer (via virtual lines).
+
+    Note: Virtual lines currently use the highlight `GitSignsDeleteVirtLn`.
+  '';
+
+  diff_opts = let
+    diffOptType = types.submodule {
```
traxys, line 152: shouldn't we set  this as a freeform type

traxys (changes requested):

traxys (approved):

PR #1269, opened 2024-03-15
colorschemes/kanagawa: switch to mkNeovimPlugin

GaetanLepage:
"Fixes" #1159 -> By properly unsetting `vim.o.background` (`option.background = "";`) as mentioned in the plugin README, everything works as expected.

### DIFF
--- a/plugins/colorschemes/kanagawa.nix
+++ b/plugins/colorschemes/kanagawa.nix
@@ -5,195 +5,225 @@
   config,
   ...
 }:
-with lib; let
-  cfg = config.colorschemes.kanagawa;
-in {
-  options = {
-    colorschemes.kanagawa =
-      helpers.neovim-plugin.extraOptionsOptions
-      // {
-        enable = mkEnableOption "kanagawa";
+with lib;
+  helpers.neovim-plugin.mkNeovimPlugin config {
+    name = "kanagawa";
+    isColorscheme = true;
+    originalName = "kanagawa.nvim";
+    defaultPackage = pkgs.vimPlugins.kanagawa-nvim;
 
-        package = helpers.mkPackageOption "kanagawa" pkgs.vimPlugins.kanagawa-nvim;
+    description = ''
+      You can select the theme in two ways:
+      - Set `colorschemes.kanagawa.settings.theme` AND explicitly unset `vim.o.background` (i.e. `options.background = ""`).
+      - Set `colorschemes.kanagawa.settings.background` (the active theme will depend on the value of `vim.o.background`).
+    '';
 
-        compile = helpers.defaultNullOpts.mkBool false "Enable compiling the colorscheme.";
+    maintainers = [maintainers.GaetanLepage];
 
-        undercurl = helpers.defaultNullOpts.mkBool true "Enable undercurls.";
+    # TODO introduced 2024-03-15: remove 2024-05-15
+    deprecateExtraOptions = true;
+    imports = let
+      basePluginPath = ["colorschemes" "kanagawa"];
+    in
+      (
+        map
+        (
+          optionPath:
+            mkRenamedOptionModule
+            (basePluginPath ++ optionPath)
+            (basePluginPath ++ ["settings"] ++ optionPath)
+        )
+        [
+          ["compile"]
+          ["undercurl"]
+          ["commentStyle"]
+          ["functionStyle"]
+          ["keywordStyle"]
+          ["statementStyle"]
+          ["typeStyle"]
+          ["transparent"]
+          ["dimInactive"]
+          ["terminalColors"]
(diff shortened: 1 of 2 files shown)


plugins/colorschemes/kanagawa.nix
```diff
@@ -5,195 +5,217 @@
...
+          "functionStyle"
+          "keywordStyle"
+          "statementStyle"
+          "typeStyle"
+          "transparent"
+          "dimInactive"
+          "terminalColors"
+          "colors"
+          "theme"
+          "background"
```
traxys, line 43: Shouldn't this be shouldn't we rename background.dark & background.light instead of simply background?

plugins/colorschemes/kanagawa.nix
```diff
@@ -5,195 +5,217 @@
...
+          "undercurl"
+          "commentStyle"
+          "functionStyle"
+          "keywordStyle"
+          "statementStyle"
+          "typeStyle"
+          "transparent"
+          "dimInactive"
+          "terminalColors"
+          "colors"
```
traxys, line 41: shouldn't this be colors.palette & colors.theme ?

traxys (commented):
We should add some documentation (or warnings) to properly handle #1159

plugins/colorschemes/kanagawa.nix
GaetanLepage: If I'm not mistaking, as those are already "snake case", we can limit the warning to the root option.

traxys: Yeah, but `colors` is not an option though? It simply is an attrset?

GaetanLepage: I've switched to writing the full paths.

plugins/colorschemes/kanagawa.nix
GaetanLepage: I've switched to writing the full paths.

plugins/colorschemes/kanagawa.nix
traxys: This does not work because you need to have a `isString` or `isList`

GaetanLepage:
> We should add some documentation (or warnings) to properly handle #1159

I added a warning.

plugins/colorschemes/kanagawa.nix
```diff
@@ -5,195 +5,232 @@
...
-      colorscheme = "kanagawa";
+      };
+      overrides = "function(colors) return {} end";
+      theme = "wave";
+    };
 
-      extraPlugins = [cfg.package];
+    extraConfig = cfg: {
+      warnings =
+        optional ((config.options.background or null) == null)
```
traxys, line 226: I _think_ if colorschemes.kanagawa.settings.background is set to `emptyTable` _then_ background is not set. It is also possible to add 
```
vim.cmd("colorscheme kanagawa-wave")
vim.cmd("colorscheme kanagawa-dragon")
vim.cmd("colorscheme kanagawa-lotus")
```

manually to the extraConfigLua, but I guess this is outside of what we want to help.

The readme says:

Themes can be changed in three ways:

    Setting config.theme to the desired theme. Note that vim.o.background must be unset.
    Using the background option: Any change to the value of vim.o.background will select the theme mapped by config.background. Use vim.o.background = "" to unset this option.
    Loading the colorscheme directly with: 

This is what is telling me that the warning is not sufficient, as it only handles case 2.

traxys (changes requested):

plugins/colorschemes/kanagawa.nix
GaetanLepage: I planned to support cases 1 and 2, but not 3.
In both cases 1 and 2, background must be explicitly **set** (in the nixvim sense), to either `""`, `"dark"` or `"light"`.

GaetanLepage: Hmmm.

GaetanLepage: So, indeed, this module target use cases 1 and 2.
However the warning is useless. If the user doesn't set `options.background`, its default value is `"dark"` in neovim, and this is fine.

traxys:
I think we should add some documentation (on the global plugin desc) on this behaviour? Maybe by copy-pasting the case 1 & 2 from the README to try and avoid user confusion on this plugin

GaetanLepage:
> I think we should add some documentation (on the global plugin desc) on this behaviour? Maybe by copy-pasting the case 1 & 2 from the README to try and avoid user confusion on this plugin

I am not against it, but at the same time we are 100% in plugin-usage-information land.
According to me, this is not our responsibility.

Nonetheless, I added instructions in the description.

traxys (approved):

PR #1280, opened 2024-03-17
plugins/tmux-navigator: Add keymap config options

MattSturgeon:
First time opening a PR here, so let me know if I've missed some etiquette.

#### Summary

- New option `tmuxNavigatorNoMappings` maps to global [`tmux_navigator_no_mappings`](https://github.com/christoomey/vim-tmux-navigator?tab=readme-ov-file#custom-key-bindings)
- New attrset `keymaps` maps keymaps to the various `:TmuxNavigate*` commands
  - Set of either `enum` or `mapConfigOptions` submodule (with some tweaks)

#### Future refactoring

I don't see why the various options should include the `tmuxNavigator` prefix; that is surely implied by being in the tmux-navigator plugin...

I also think that the `null` or `1` options could be refactored as `bool` options, to make more sense in nix...

I can open this as a separate issue/PR if it warrants discussion, or I could lump it into this PR if that's easier to review?

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -60,11 +60,7 @@ rec {
       # ["" "n" "v" ...]
       (map ({ short, ... }: short) (attrValues modes));
 
-  mapOptionSubmodule = mkMapOptionSubmodule {
-    hasKey = true;
-    hasAction = true;
-    rawAction = true;
-  };
+  mapOptionSubmodule = mkMapOptionSubmodule { };
 
   mkModeOption =
     default:
@@ -85,9 +81,10 @@ rec {
   mkMapOptionSubmodule =
     {
       defaults ? { },
-      hasKey ? false,
-      hasAction ? false,
-      rawAction ? true,
+      # key and action can be true/false to enable/disable adding the option,
+      # or an attrset to enable the option and add/override mkOption args.
+      key ? true,
+      action ? true,
     }:
     # TODO remove assert once `lua` option is gone
     # This is here to ensure no uses of `mkMapOptionSubmodule` set a `lua` default
@@ -96,22 +93,24 @@ rec {
       with types;
       submodule {
         options =
-          (optionalAttrs hasKey {
+          (optionalAttrs (isAttrs key || key) {
             key = mkOption (
               {
                 type = str;
                 description = "The key to map.";
                 example = "<C-m>";
               }
+              // (optionalAttrs (isAttrs key) key)
               // (optionalAttrs (defaults ? key) { default = defaults.key; })
             );
           })
-          // (optionalAttrs hasAction {
+          // (optionalAttrs (isAttrs action || action) {
             action = mkOption (
               {
-                type = if rawAction then nixvimTypes.maybeRaw str else str;
+                type = nixvimTypes.maybeRaw str;
                 description = "The action to execute.";
               }
+              // (optionalAttrs (isAttrs action) action)
               // (optionalAttrs (defaults ? action) { default = defaults.action; })
             );
           })
--- a/plugins/utils/tmux-navigator.nix
+++ b/plugins/utils/tmux-navigator.nix
@@ -6,30 +6,6 @@
   ...
 }:
 with lib;
-let
-  # TODO: Introduced 2024-03-19, remove on 2024-05-19
-  deprecations =
-    let
-      pluginPath = [
-        "plugins"
-        "tmux-navigator"
-      ];
-      option = s: pluginPath ++ [ s ];
-      setting =
-        s:
-        pluginPath
-        ++ [
-          "settings"
-          s
-        ];
-      settingStr = s: concatStringsSep "." (setting s);
-    in
-    [
-      (mkRenamedOptionModule (option "tmuxNavigatorSaveOnSwitch") (setting "save_on_switch"))
-      (mkRemovedOptionModule (option "tmuxNavigatorDisableWhenZoomed") "Use `${settingStr "disable_when_zoomed"}` option.")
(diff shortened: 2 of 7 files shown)


plugins/utils/tmux-navigator.nix
```diff
@@ -6,6 +6,43 @@
...
+    "right"
+    "previous"
+  ];
+
+  # Define a submodule so we get docs and enum type-safety for `action`
+  keymapOptions = {
+    # FIXME: Do we really want _all_ of `mapConfigOptions` here?
+    #    Should we consider filtering it? Etc?
+    options =
+      (recursiveUpdate helpers.keymaps.mapConfigOptions (mapAttrs (n: v: {${n}.default = v;}) defaultOptions))
```
MattSturgeon, line 23: Thoughts on this? I can remove the `FIXME` if this approach is ok. Otherwise, how would you set-up the submodule options?

MattSturgeon:
> I don't see why the various options should include the `tmuxNavigator` prefix; that is surely implied by being in the tmux-navigator plugin...
> 
> I also think that the `null` or `1` options could be refactored as `bool` options, to make more sense in nix...
> 
> I can open this as a separate issue/PR if it warrants discussion, or I could lump it into this PR if that's easier to review?

IMO this PR should be reworked after the plugin is ported to `mkVimPlugin`. I've opened #1293 to do that.

GaetanLepage:
Is this PR still relevant ? If so, it is necessary to rebase it on `main` as it has been a while ^^

GaetanLepage:
cc @MattSturgeon

MattSturgeon:
Unless I have some time to kill earlier, I'll take a look when I get back to the UK.

It'll probably need a rewrite due to #1293.

GaetanLepage:
Could you rebase this on main please ?

plugins/utils/tmux-navigator.nix
```diff
@@ -147,38 +121,72 @@ helpers.vim-plugin.mkVimPlugin config {
...
+          action = "right";
+        }
+        {
+          key = "<C-w>\\";
+          action = "previous";
+        }
+      ];
+      default = [ ];
+      type = types.listOf (
+        helpers.keymaps.mkMapOptionSubmodule {
```
MattSturgeon, line 170: This probably needs a `defaults.mode = "n"` or something. I can't recall what `mode = ""` does in nvim ottomh.

plugins/utils/tmux-navigator.nix
```diff
@@ -147,38 +121,72 @@ helpers.vim-plugin.mkVimPlugin config {
...
+            example = "left";
+          };
+        }
+      );
+    };
+  };
+
+  extraConfig = cfg: {
+    keymaps = map (
+      v: v // { action = "<cmd>TmuxNavigate${helpers.upperFirstChar v.action}<cr>"; }
```
MattSturgeon, line 189: Maybe `k` or `m` instead of `v`? Or be verbose and call it `keymap`, `mapping`, or something?

plugins/utils/tmux-navigator.nix
GaetanLepage: `mode = ""` is the same as normal-visual-op.

lib/keymap-helpers.nix
```diff
@@ -85,9 +81,10 @@ rec {
   mkMapOptionSubmodule =
     {
       defaults ? { },
-      hasKey ? false,
-      hasAction ? false,
-      rawAction ? true,
+      # key and action can be true/false to enable/disable adding the option,
+      # or an attrset to enable the action and add/override mkOption args.
+      key ? true,
+      action ? true,
```
MattSturgeon, line 87: Still not a perfect function signature, but at least we have unobstructed access to some of the underlying `mkOption` calls now :thinking: 

A `mode` argument makes sense too, though `mkModeOption` is currently sitting in the way of `mkOption`. Same for `options.*`.

tests/test-sources/plugins/utils/tmux-navigator.nix
```diff
@@ -18,4 +20,35 @@
       };
     };
   };
+
+  withKeymap = {
```
GaetanLepage, line 24: `with-keymap`

plugins/utils/tmux-navigator.nix
GaetanLepage: verbose would be better.

GaetanLepage:
So you chose the "list" approach which is definitely more straightforward to implement.
However, wouldn't be an attrs approach clearer for the user ?
`keymaps.<action>.key = "<C-m>";` ?

MattSturgeon:
> So you chose the "list" approach which is definitely more straightforward to implement. However, wouldn't be an attrs approach clearer for the user ? `keymaps.<action>.key = "<C-m>";` ?

I felt like a list was a better analogue for the underlying keymaps.

In particular, if a user wanted multiple "left" or "previous" binds, that wouldn't be possible with an `attrsOf (either str submodule)`.

I'm still torn as to whether most/all of our plugin keybinds should be transitioning towards lists at some point, but that's way beyond the scope of this PR 😂

GaetanLepage:
> I felt like a list was a better analogue for the underlying keymaps.
> 
> In particular, if a user wanted multiple "left" or "previous" binds, that wouldn't be possible with an `attrsOf (either str submodule)`.
> 
> I'm still torn as to whether most/all of our plugin keybinds should be transitioning towards lists at some point, but that's way beyond the scope of this PR 😂

Yes indeed, hard to choose.
I'm OK with this.

GaetanLepage (approved):

PR #1298, opened 2024-03-21
plugins/codeium-vim: fix keymaps

### LINKED ISSUE #1156: [BUG] copilot-vim.keymaps do not work properly
| Plugin       | `codeium-vim`  | 
| Nixpkgs      | `nixos-unstable` | 

## Description

When i try to use the bind, it throws "E15: Invalid expression: <80><fd>hcodeium#Accept()^M", the default one works as expected.
@GaetanLepage 

```nix
programs.nixvim = {
  plugins = {
    codeium-vim = {
      enable = true;
      keymaps.accept = "<M-Tab>";
    };
  };
}
```

GaetanLepage:
Fixes #1156

### DIFF
--- a/plugins/completion/codeium-vim.nix
+++ b/plugins/completion/codeium-vim.nix
@@ -10,27 +10,27 @@ with lib; let
     clear = {
       default = "<C-]>";
       description = "Keymap for clearing current suggestion.";
-      command = "codeium#Clear()";
+      command = "vim.fn['codeium#Clear']()";
     };
     next = {
       default = "<M-]>";
       description = "Keymap for cycling to the next suggestion.";
-      command = "codeium#CycleCompletions(1)";
+      command = "vim.fn['codeium#CycleCompletions'](1)";
     };
     prev = {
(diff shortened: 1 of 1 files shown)


GaetanLepage:
This patch fixes the error described in the linked issue.
However, I can't make the `accept` keybind works... Seems related to https://github.com/Exafunction/codeium.vim/issues/215.

GaetanLepage:
@hyperpuncher @spl3g could you test this fix please ?
It is working better than the current solution, but I am curious to see if `accept` will work for you.

plugins/completion/codeium-vim.nix
```diff
@@ -127,9 +127,9 @@ in
           (v != null)
           {
             key = v;
-            action = let
+            action.__raw = let
```
traxys, line 130: I'd prefer we use `helpers.mkRaw`, it is a bit more descriptive

traxys (changes requested):

traxys (approved):

Issue #1215, opened 2024-03-06
feature: keymap aliasing


MattSturgeon:
Currently if you wish to have multiple keymaps that do the same thing, you must either duplicate your definition for each alias or `map` a list of alias strings into a list of keymap attribute sets.

It may be handy to provide an `aliases` option accepting `null`, `str` or `listOf str`. If used, the keymap will be created for `key` as well as each `alias`.

If this is considered unnecessary feature bloat, perhaps document an example somewhere (README, Wiki, or docs) demonstrating how `map` can be used to produce duplicate keymaps?

MattSturgeon:
@traxys this is probably wontfix for the same reason as #1214?

I guess, at most, the scope of this issue could be reduced to documenting an example of mapping a list of string-keys into multiple keybinds.

If you think such an example adds more value than confusion...

traxys:
Yeah I think like #1214 this is a bit out of the scope of nixvim.

traxys:
But we could add a section Tips & Tricks to the user guide were we could add Nix snippets to do stuff like this (and maybe to simulate something like #1214)

PR #1301, opened 2024-03-22
docs: Add keymap alias FAQ

MattSturgeon:
Document one approach to solving #1215

Added as an FAQ for now, but may be better suited to a new "Tips and Tricks" section, [as suggested](https://github.com/nix-community/nixvim/issues/1215#issuecomment-1992737399) by @traxys.


traxys (approved):

traxys:
If we start having more tips like this we'll create a new section, for now let's leave it in the FAQ

PR #1304, opened 2024-03-22
plugins/cmp: fix examples for snippet.expand

GaetanLepage:
This was not adapted to the new implementation.

### DIFF
--- a/tests/test-sources/plugins/completion/cmp.nix
+++ b/tests/test-sources/plugins/completion/cmp.nix
@@ -6,7 +6,7 @@
   snippet-engine = {
     plugins.cmp = {
       enable = true;
-      settings.snippet.expand = "luasnip";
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/completion/cmp.nix
```diff
@@ -6,7 +6,7 @@
   snippet-engine = {
     plugins.cmp = {
       enable = true;
-      settings.snippet.expand = "luasnip";
+      snippet.expand = "function(args) require('luasnip').lsp_expand(args.body) end";
```
traxys, line 9: why remove the `settings` ?

traxys (changes requested):

tests/test-sources/plugins/completion/cmp.nix
GaetanLepage: Because copy-paste...
Fixed.

traxys (approved):

PR #1315, opened 2024-03-24
plugins/sleuth: init

GaetanLepage:
Add support for tpope's [vim-sleuth](https://github.com/tpope/vim-sleuth).


plugins/utils/sleuth.nix
```diff
@@ -0,0 +1,16 @@
...
+  ...
+}:
+helpers.vim-plugin.mkVimPlugin config {
+  name = "sleuth";
+  originalName = "vim-sleuth";
+  defaultPackage = pkgs.vimPlugins.vim-sleuth;
+
+  maintainers = [lib.maintainers.GaetanLepage];
+
+  # In typical tpope fashion, this plugin has no config options
```
traxys, line 15: That is not true! It at least has the following options:

- `g:sleuth_{filetype}_heuristics`
- `g:sleuth_heuristics`
- `g:sleuth_no_filetype_indent_on`
(I'm not suggesting we package them though, we can do as we want)

plugins/utils/sleuth.nix
```diff
@@ -0,0 +1,41 @@
...
+}:
+helpers.vim-plugin.mkVimPlugin config {
+  name = "sleuth";
+  originalName = "vim-sleuth";
+  defaultPackage = pkgs.vimPlugins.vim-sleuth;
+  globalPrefix = "sleuth_";
+
+  maintainers = [lib.maintainers.GaetanLepage];
+
+  # In typical tpope fashion, this plugin has no config options
```
traxys, line 16: You can remove this comment as we expose options now!

traxys (changes requested):

traxys (approved):

Issue #1218, opened 2024-03-06
feature: option to enable refactor.nvim's telescope extension


MattSturgeon:
[refactoring.nvim](https://github.com/ThePrimeagen/refactoring.nvim) includes a Telescope extension that can be enabled using

```nix
extraConfigLuaPre = /* lua */ ''
  -- load refactoring Telescope extension
  require("telescope").load_extension("refactoring")
'';
```

And used in keymaps as follows:

```nix
keymps = [
  {
    mode = "n";
    key = "<leader>rr";
    action = /* lua */ ''
      function() require("telescope").extensions.refactoring.refactors() end
    '';
    lua = true;
    options.desc = "Select refactoring option";
  }
];
```

It'd be great if `plugins.refactoring` had options to configure this

#### Related

Perhaps `plugins.telescope.enabledExtensions` could be made public or an `plugins.telescope.extraExtensions` could be introduced to handle other unsupported telescope extensions?

Bodleum:
Would it make sense to do this? You could have something like:

```nix
plugins.refactoring = {
    enable = true;
    telescopeExtension.enable = true;
};
```

but that wouldn't guarantee that `plugins.telescope` was enabled.
Of course, by enabling the extension from the refactoring side, it could also enable telescope, but I get the feeling that silently enabling other plugins is a bad idea!

Just my thoughts, push back if you disagree!

GaetanLepage:
Maybe, we should merge the telescope refactor first...

MattSturgeon:
> Maybe, we should merge the telescope refactor first...

Agreed.

It'd be great if that pr added public access to `plugins.telescope.enabledExtensions` too, for any other telescope extensions we don't support.

> Would it make sense to do this? You could have something like: `plugins.refactoring.telescopeExtension.enable`

Looks fine to me. Is this how other telescope plugins (not included in telescope itself) are handled?

PR #1320, opened 2024-03-25
plugins/telescope: refactor

GaetanLepage:
Refactoring of the `plugins.telescope` module:
- Switch to `mkNeovimPlugin`
- Better handling of extensions:
  Adopt a similar approach as in `neotest`: Add a helper function to create the submodule for each extension.
  This reduces the amount of boilerplate code.
- Fix wrong options
- Add tests


traxys:
Please describe in more details in a message what is a "better handling of extensions" and why would it mandate a 1000+ lines refactor

GaetanLepage:
> Please describe in more details in a message what is a "better handling of extensions" and why would it mandate a 1000+ lines refactor

Yes, indeed !
Actually the enormous diff is caused by `git` not handling the file renames properly and counted them as deletion-creation...

plugins/telescope/default.nix
```diff
@@ -59,78 +69,86 @@ in {
...
+
+      extensions = mkOption {
+        type = with types; attrsOf anything;
+        default = {};
+        # NOTE: We hide this option from the documentation as users should use the top-level
+        # `extensions.*` options.
+        # They can still directly change this `settings.adapters` option.
+        # In this case, they are responsible for explicitly installing the manually added extensions.
+        visible = false;
+      };
```
MattSturgeon, line 129: Could this be visible to mitigate issues like #1218?

> Perhaps `plugins.telescope.enabledExtensions` could be made public or an `plugins.telescope.extraExtensions` could be introduced to handle other unsupported telescope extensions?

Or is there a better solution?

traxys: wouldn't `plugins.telescope.settings.extensions` work for that? It's the main adavantage of RFC 42 style plugins, you should _not_ be limited to what nixvim provides

traxys (approved):

plugins/telescope/extensions/media-files.nix
```diff
@@ -0,0 +1,120 @@
...
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+  (
+    import ./_helpers.nix {inherit lib helpers config pkgs;}
+  )
+  .mkExtension {
+    name = "media-files";
```
MattSturgeon, line 13: I get `error: The option `plugins.telescope.extensions.media_files' does not exist.`

I think this should've had a rename deprecation added.

plugins/telescope/default.nix
```diff
@@ -5,37 +5,47 @@
...
+          (
+            either
+            str
+            (submodule {
+              options = {
+                action = mkOption {
+                  type = types.str;
+                };
+                mode = helpers.keymaps.mkModeOption "n";
+                options = helpers.keymaps.mapConfigOptions;
```
MattSturgeon, line 38: Not sure if it's possible for attrsOf submodule, but ideally this should've had a rename deprecation too:

`error: The option `plugins.telescope.keymaps."<leader>bb".desc' does not exist`

plugins/telescope/extensions/media-files.nix
GaetanLepage: Was it necessary to rename it to `media-files`... Not sure now.

plugins/telescope/default.nix
GaetanLepage: Yes, this is not really possible unfortunately.
The only alternative is to make a `mkRemovedOptionModule` for the whole `keymaps` option to force the user to re-define it manually.

PR #1322, opened 2024-03-25
LSP: Register keymaps on `LspAttach`

### LINKED ISSUE #1219: feature: register LSP binds on `LspAttach`

Bodleum:
Register keymaps defined in `plugins.lsp.keymaps` on `LspAttach`.
Added `plugins.lsp.keymaps.extra` to register arbitrary keymaps.

Fixes #1219

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -260,37 +260,35 @@ in {
     mkIf cfg.enable {
       extraPlugins = [pkgs.vimPlugins.nvim-lspconfig];
 
-      keymapsOnEvents = {
-        "LspAttach" = let
-          mkMaps = prefix:
-            mapAttrsToList
-            (
-              key: action: let
-                actionStr =
-                  if isString action
-                  then action
-                  else action.action;
-                actionProps =
-                  if isString action
-                  then {}
-                  else filterAttrs (n: v: n != "action") action;
-              in {
-                mode = "n";
-                inherit key;
-                action.__raw = prefix + actionStr;
+      keymapsOnEvents.LspAttach = let
+        mkMaps = prefix:
+          mapAttrsToList
+          (
+            key: action: let
+              actionStr =
+                if isString action
+                then action
+                else action.action;
+              actionProps =
+                if isString action
+                then {}
+                else filterAttrs (n: v: n != "action") action;
+            in {
+              mode = "n";
+              inherit key;
+              action = helpers.mkRaw (prefix + actionStr);
 
-                options =
-                  {
-                    inherit (cfg.keymaps) silent;
-                  }
-                  // actionProps;
-              }
-            );
-        in
-          (mkMaps "vim.diagnostic." cfg.keymaps.diagnostic)
-          ++ (mkMaps "vim.lsp.buf." cfg.keymaps.lspBuf)
-          ++ cfg.keymaps.extra;
-      };
+              options =
+                {
+                  inherit (cfg.keymaps) silent;
+                }
+                // actionProps;
+            }
+          );
+      in
+        (mkMaps "vim.diagnostic." cfg.keymaps.diagnostic)
+        ++ (mkMaps "vim.lsp.buf." cfg.keymaps.lspBuf)
+        ++ cfg.keymaps.extra;
 
       # Enable all LSP servers
       extraConfigLua = ''
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -114,7 +114,7 @@ in {
 
         diagnostic = mkOption {
           type = with types; attrsOf (either str attrs);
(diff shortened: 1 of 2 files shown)


plugins/lsp/default.nix
```diff
@@ -134,6 +134,35 @@ in {
...
+              action = "<CMD>LspRestart<Enter>";
+            }
+            {
+              key = "gd";
+              action = "require('telescope.builtin').lsp_definitions()";
+              lua = true;
+            }
+            {
+              key = "K";
+              action = "<CMD>Lspsaga hover_doc<Enter>";
```
MattSturgeon, line 161: I wonder if this would add more confusion, since most users will use the "normal" hover?

plugins/lsp/default.nix
```diff
@@ -231,34 +260,37 @@ in {
...
+                  then action
+                  else action.action;
+                actionProps =
+                  if isString action
+                  then {}
+                  else filterAttrs (n: v: n != "action") action;
+              in {
+                mode = "n";
+                inherit key;
+                action.__raw = prefix + actionStr;
```
MattSturgeon, line 280: If this line is being indented anyway, is it worth changing it to use `helpers.mkRaw`?

plugins/lsp/default.nix
```diff
@@ -231,34 +260,37 @@ in {
...
-              actionProps =
-                if isString action
-                then {}
-                else filterAttrs (n: v: n != "action") action;
-            in {
-              mode = "n";
-              inherit key;
-              action.__raw = prefix + actionStr;
+      keymapsOnEvents = {
+        "LspAttach" = let
```
MattSturgeon, line 264: Could avoid indenting the lines below by doing something like this:
```suggestion
      keymapsOnEvents."LspAttach" = let
```

MattSturgeon (approved):
Looks good. Left a few comments to probe discussion.

plugins/lsp/default.nix
Bodleum: Yes, I just added that, and the telescope mapping as examples of why you would use this as well as how.

Easily removed if you think that'd be clearer.

MattSturgeon: No strong opinions either way. More examples is usually good, this one just felt like it could potentially confuse new users.

I'll let the maintainers decide if it should be changed.

MattSturgeon:
Probably a good idea to update the description for `diagnostics` and `lspBuf` mentioning that the keymaps will be registered on `LspAttach`.
Similar to your description for `extra`.

plugins/lsp/default.nix
MattSturgeon: I think this is fine now, with the updated description.

MattSturgeon (approved):
LGTM

GaetanLepage (approved):
LGTM !
Thanks @Bodleum for this work and @MattSturgeon for kindly reviewing it.

As this PR affects the LSP code, I will wait for an extra review from @traxys before merging.

traxys (approved):

PR #1326, opened 2024-03-26
plugins/sandwich: init

### LINKED ISSUE #1161: [PLUGIN REQUEST] vim-sandwich
| Field    | Description |
|----------|-------------|
| Plugin   | `machakann/vim-sandwich`  | 
| Homepage | `[<URL>](https://github.com/machakann/vim-sandwich)`     | 
| Nixpkgs  | `vimPlugins.vim-sandwich` | 

## Extra Information
Pretty basic plugin, I made it work using extraPlugins by just putting it in the list. So I guess no specific settings or anything in that regard

GaetanLepage:
Add support for [vim-sandwich](https://github.com/machakann/vim-sandwich).

Fixes #1161

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -174,6 +174,7 @@
     ./utils/quickmath.nix
     ./utils/refactoring.nix
     ./utils/rest.nix
+    ./utils/sandwich.nix
     ./utils/sleuth.nix
     ./utils/smart-splits.nix
     ./utils/specs.nix
--- a/plugins/utils/sandwich.nix
+++ b/plugins/utils/sandwich.nix
@@ -1,0 +1,43 @@
+{
+  lib,
+  config,
+  helpers,
+  pkgs,
+  ...
+}:
+with lib;
+  helpers.vim-plugin.mkVimPlugin config {
+    name = "sandwich";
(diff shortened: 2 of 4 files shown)


traxys (changes requested):
I don't think this is a nice way to package the plugin, because most of the actual configuration options use the `#` syntax.

I think this plugin does not fit very nicely in the raw "settings" model, we may need to do some manual work for it

GaetanLepage:
> I don't think this is a nice way to package the plugin, because most of the actual configuration options use the `#` syntax.
> 
> I think this plugin does not fit very nicely in the raw "settings" model, we may need to do some manual work for it

I see your point. However, there is no perfect solution for this plugin. I would like to avoid our previous model of crafting custom options to wrap the plugin ones.
I would argue that the user of this plugin will be very unlikely to customize the behavior to the point of needing those `#` settings. This is backed up by a research across publicly available neovim configurations on GitHub.

Also, I have added a notice in the plugin description.
My opinion would be to ship this module as is.

traxys:
> This is backed up by a search across publicly available neovim configurations on GitHub.

Good enough for me

traxys (approved):

PR #1344, opened 2024-03-28
colorschemes/catpuccin: switch to mkNeovimPlugin

GaetanLepage:


plugins/colorschemes/catppuccin.nix
```diff
@@ -5,408 +5,254 @@
...
+
+    maintainers = [maintainers.GaetanLepage];
+
+    # TODO introduced 2024-03-27: remove 2024-05-27
+    optionsRenamedToSettings = [
+      "flavour"
+      ["background" "light"]
+      ["background" "dark"]
+      "transparentBackground"
+      ["dimInactive" "enabled"]
```
MattSturgeon, line 22: How does `optionsRenamedToSettings` handle renames like this, where the new name or path is different?

Is camelCase automatically converted to snake\_case?

Is the path different for any of these renames?

Should any of these rename/removals be handled manually instead?

GaetanLepage: [Here](https://github.com/nix-community/nixvim/blob/c86a0aca63ce91c482b1b86242e0959ceb882667/lib/neovim-plugin.nix#L89-L105) is the implementation of `optionsRenamedToSettings`.

The `mkNeovimPlugin` helper declares, for every option in this list (`["myOldOption" "subOption"]`) a rename warning from the old path (`plugins.foo.myOldOption.subOption`) to the new path (`plugins.foo.settings.my_old_option.sub_option`).

Notes:
- Elements that are simple strings (`"foo"`), this will be converted to a singleton (`["foo"]`).
- Yes, camelCase is automatically converted to snake_case.

GaetanLepage: In some cases, it is necessary to create a warning manually.

GaetanLepage: Can I mark this issue as resolved ?

MattSturgeon (approved):

PR #1350, opened 2024-03-30
tests: fix occurences of deprecated "options" option

GaetanLepage:


tests/test-sources/plugins/bufferlines/barbar.nix
```diff
@@ -17,7 +17,6 @@
     };
   };
 
-  # All the upstream default options of barbar
```
traxys, line 20: This is completly unrelated (there are lot's of other changes like this in the PR)

traxys (changes requested):

tests/test-sources/plugins/bufferlines/barbar.nix
GaetanLepage: Sure, I removed the commit.

traxys (approved):

PR #1352, opened 2024-03-30
colorschemes/dracula: switch backend to dracula.nvim

GaetanLepage:
Switch to using the [dracula.nvim](https://github.com/Mofiqul/dracula.nvim) plugin for `colorschemes.dracula`.

### DIFF
--- a/plugins/colorschemes/dracula.nix
+++ b/plugins/colorschemes/dracula.nix
@@ -1,87 +1,75 @@
 {
   lib,
-  config,
-  pkgs,
   ...
 }:
 let
-  inherit (lib) types;
-  cfg = config.colorschemes.dracula;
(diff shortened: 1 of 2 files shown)


traxys (commented):
I'm not sure this is the best approach, as there is no "single truth" about the dracula colorscheme. I'd rather rename the old plugin to "dracula-vim" and use this name for the nvim one.
Or create "dracula-vim" and "dracula-nvim"

MattSturgeon:
> there is no "single truth" about the dracula colorscheme.

This probably applies to a bunch of stuff.  
I think it's only valuable to maintain both if there's a legitimate reason the user would want to use the other **and** having "native" support provides some benefit over just using `extraPackages`

PR #1353, opened 2024-03-30
wrappers/hm: add vimdiffAlias option

### LINKED ISSUE #1351: [FEATURE] Add vimdiff alias support as well
| Plugin       | `core NixVim`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

Compaired to the vanilla NeoVim configureation, NixVim supports setting up aliases for `vi` and `vim` to NixVim, but the option of doing the same for vimdiff is missing from NixVim. I would love to see this be added as well.

GaetanLepage:
Implement the `programs.neovim.vimdiffAlias` option from HM in nixvim.
Note: this only applies to the HM module.

Fixes #1351

### DIFF
--- a/wrappers/modules/hm.nix
+++ b/wrappers/modules/hm.nix
@@ -4,5 +4,13 @@ with lib;
   options = {
     enable = mkEnableOption "nixvim";
     defaultEditor = mkEnableOption "nixvim as the default editor";
+
+    vimdiffAlias = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Alias `vimdiff` to `nvim -d`.
+      '';
+    };
   };
 }
--- a/wrappers/hm.nix
+++ b/wrappers/hm.nix
@@ -53,5 +53,10 @@ in
       inherit (cfg) warnings assertions;
       home.sessionVariables = mkIf cfg.defaultEditor { EDITOR = "nvim"; };
     }
+    {
+      programs.bash.shellAliases = mkIf cfg.vimdiffAlias { vimdiff = "nvim -d"; };
+      programs.fish.shellAliases = mkIf cfg.vimdiffAlias { vimdiff = "nvim -d"; };
+      programs.zsh.shellAliases = mkIf cfg.vimdiffAlias { vimdiff = "nvim -d"; };
+    }
   ]);
 }


traxys:
This needs to be documented somewhere, because as is I don't think the option will appear in the documentation.

traxys:
I though about this a bit more, I think we should add this feature in the same place as the viAlias/vimAlias and do the work for all the wrappers (nix-darwin, home-manager, nixos, standalone) or we should not do it at all

MattSturgeon:
Could we use `pkgs.wrapProgram`, `pkgs.makeWrapper`, or similar on platforms where we can't make shell aliases?

traxys:
Yeah I think we would need to do this for the standalone wrapper. It may be the best to do it for everyone though (mabye even do this in nixpkgs proper? The viAlias & vimAlias is implemented there IIRC)

wrappers/modules/hm.nix
```diff
@@ -3,5 +3,13 @@ with lib; {
   options = {
     enable = mkEnableOption "nixvim";
     defaultEditor = mkEnableOption "nixvim as the default editor";
+
+    vimdiffAlias = mkOption {
```
MattSturgeon, line 7: This could use `mkEnableOption`

MattSturgeon (changes requested):

MattSturgeon:
> This needs to be documented somewhere, because as is I don't think the option will appear in the documentation.

This should have been resolved by #1355.



> > I think we should add this feature in the same place as the viAlias/vimAlias and do the work for all the wrappers (nix-darwin, home-manager, nixos, standalone)
> 
> Could we use `pkgs.wrapProgram`, `pkgs.makeWrapper`, or similar on platforms where we can't make shell aliases?

We already have `defaultEditor` exclusive to home-manager. Perhaps porting these could be handled in a separate PR?

I don't think not having this everywhere should necessarily be a blocker.

wrappers/modules/hm.nix
GaetanLepage: Yes we could, but it would force the description to start with "Whether to enable" which is not ideal.

MattSturgeon: > which is not ideal

Not sure I agree "Whether to enable `vimdiff` alias" is not ideal, but I'll agree to disagree :grinning:

MattSturgeon (approved):

MattSturgeon:
> maybe even do this in nixpkgs proper? The viAlias & vimAlias is implemented there IIRC

That's correct. I'm happy for this to be merged as a stop-gap, but I agree long-term it'd be better to have this everywhere or even in nixpkgs' nvim wrapper.

IMO a lot of our wrapper stuff could be unified, so maybe there's some future (unrelated) refactoring to be done there...

Issue #1347, opened 2024-03-28
Feat - Export portable neovim configuration


PowerUser64:
I think it would be really cool if nixvim could export a portable version of the configuration files it generates. This would allow configurations to be more easily used in limited environments such as servers where a nix might not exist, or there could be other restrictions related to installing additional software.

Related: https://github.com/ralismark/nix-appimage, https://github.com/DavHau/nix-portable

traxys:
I think exporting to windows is going to be very hard, because nixvim does not only bundle the user configuration & plugins, but also adds _a lot_ of programs like LSPs, linters, formmaters,...

I think If you build your configuration as [a standalone nixvim module](https://nix-community.github.io/nixvim/#standalone-usage) you could use  https://github.com/ralismark/nix-appimage without any support on our side

MattSturgeon:
While I agree this would be a great feature to have, I can't think of a practical way to do it.

Some basics like options, globals, & keymaps could probably be done, but anything that uses plugins or packages from nixpkgs would be very difficult to make portable.

I believe there are other nix/vim frameworks that may be more suitable to portable configs, though I'm on my phone currently so I don't have any links.

I've been thinking about using a nixvim config to generate other non-nixvim files, such as `vimrc`, `ideavimrc`, etc...

Maybe the standalone nixvim build could also contain a `config` attribute which would be the final config evaluated by the module system. That could then be used by end-users to share some basics parts of the config in other files they write... Subject to all the same limitations mentioned above, of course.

PowerUser64:
> Some basics like options, globals, & keymaps could probably be done, but anything that uses plugins or packages from nixpkgs would be very difficult to make portable.

I figured, yeah. The output would probably have to at least exclude compiled things like LSP's and prebuilt treesitter modules. The nvim treesitter plugin can download+compile parsers on its own, so at least treesitter-powered highlighting is portable. As for language servers, we'd probably want to disable or ommit them in a way that they could still be enabled if the target system has them. (maybe just comment each of them out?)

Loading plugins is the thing I'm most unsure of. It would be _super_ cool if plugins could be bundled up too, because plugins can be very important to how some configurations work. Is there anything about bundling plugins that you were specifically thinking would be difficult?

> I've been thinking about using a nixvim config to generate other non-nixvim files, such as `vimrc`, `ideavimrc`, etc...

Ooh, that would be really cool too!

MattSturgeon:
[...]

PR #1356, opened 2024-03-31
wrappers: expose `config` in standalone output

MattSturgeon:
Expose the resolved `config` as an attribute on the standalone derivation, allowing users to reference options configured for nixvim in their nixos/home-manager/darwin modules.

May be useful for sharing some basic values, such as key bindings, or syncing up settings in partially related configurations.

Related to #1347


traxys:
This woukd require a mention in the documentation

MattSturgeon:
Should this be documented as an FAQ entry? Or maybe hijack the [Extending a standalone configuration](https://nix-community.github.io/nixvim/user-guide/extending-config.html) page and add a section there?

I guess it could be renamed something like "Using a standalone configuration in another module" (open to suggestions)

traxys:
Maybe we could add it in the newly introduced 'Module Specific Options' as 'standalone usage'? (And migrate extend to that too)

Maybe renaming 'Module Specific Options' to 'Module Specific Usage'?

GaetanLepage (approved):

MattSturgeon:
Added some initial docs. Would benefit from an example in the future...

traxys (approved):

PR #1371, opened 2024-04-03
plugins/surround: init

### LINKED ISSUE #1346: [PLUGIN REQUEST] nvim-surround
| Field    | Description |
|----------|-------------|
| Plugin   | `nvim-surround`  | 
| Homepage | <https://github.com/kylechui/nvim-surround>     | 
| Nixpkgs  | YES, `vimPlugins.nvim-surround` | 

Lua-based alternative for surround.vim with support for nvim-treesitter and nvim-treesitter-textobjects

## Extra Information

Default config:
https://github.com/kylechui/nvim-surround/blob/main/lua/nvim-surround/config.lua

theabm:
This PR adds the nvim-surround plugin. The previous plugin vim-surround is removed but can be enabled by adding it to extraPlugins since it does not require particular configuration. 

closes #1346.

### DIFF
--- a/plugins/utils/surround.nix
+++ b/plugins/utils/surround.nix
@@ -5,10 +5,429 @@
   pkgs,
   ...
 }:
-helpers.vim-plugin.mkVimPlugin config {
-  name = "surround";
-  originalName = "surround.vim";
-  defaultPackage = pkgs.vimPlugins.surround;
+with lib;
+  helpers.neovim-plugin.mkNeovimPlugin config {
+    name = "surround";
+    luaName = "nvim-surround";
+    originalName = "nvim-surround";
+    defaultPackage = pkgs.vimPlugins.nvim-surround;
 
-  maintainers = [lib.maintainers.GaetanLepage];
-}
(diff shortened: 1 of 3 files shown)


GaetanLepage:
Make a separate commit for adding you in the maintainers list.

plugins/utils/surround.nix
```diff
@@ -5,10 +5,429 @@
   pkgs,
   ...
 }:
-helpers.vim-plugin.mkVimPlugin config {
-  name = "surround";
-  originalName = "surround.vim";
-  defaultPackage = pkgs.vimPlugins.surround;
```
MattSturgeon, line 11: Is this functionally identical to the old surround.vim when options are left as defaults?

MattSturgeon (commented):
This looks good, thanks.

"plugins/surround: init" is misleading because "plugins/surround" already existed; this is actually a migration/switch from surround.vim to nvim-surround.

I'm also wondering if this would potentially change users' configs without them (initially) noticing? That's not a blocker, but it's worth considering.

PR #1372, opened 2024-04-04
plugins/jdtls: fix binary name

khaneliman:
Fixes incorrectly hard coded binary name for jdtls.

### DIFF
--- a/plugins/languages/nvim-jdtls.nix
+++ b/plugins/languages/nvim-jdtls.nix
@@ -25,7 +25,7 @@ in {
         plugins.nvim-jdtls = {
           enable = true;
           cmd = [
-            "$\{pkgs.jdt-language-server}/bin/jdt-language-server"
+            (lib.getExe pkgs.jdt-language-server)
             "-data" "/path/to/your/workspace"
             "-configuration" "/path/to/your/configuration"
             "-foo" "bar"
@@ -105,7 +105,7 @@ in {
           else cfg.data;
       in
         [
-          "${pkgs.jdt-language-server}/bin/jdt-language-server"
+          (lib.getExe pkgs.jdt-language-server)
         ]
         ++ ["-data" data]
         ++ (


plugins/languages/nvim-jdtls.nix
```diff
@@ -25,7 +25,7 @@ in {
         plugins.nvim-jdtls = {
           enable = true;
           cmd = [
-            "$\{pkgs.jdt-language-server}/bin/jdt-language-server"
+            "$\{pkgs.lib.getExe pkgs.jdt-language-server}"
```
GaetanLepage, line 28:
```suggestion
            (lib.getExe pkgs.jdt-language-server)
```

plugins/languages/nvim-jdtls.nix
```diff
@@ -105,7 +105,7 @@ in {
           else cfg.data;
       in
         [
-          "${pkgs.jdt-language-server}/bin/jdt-language-server"
+          "${pkgs.lib.getExe pkgs.jdt-language-server}"
```
GaetanLepage, line 108:
```suggestion
          (lib.getExe pkgs.jdt-language-server)
```

GaetanLepage (changes requested):

GaetanLepage (approved):

MattSturgeon (approved):

plugins/languages/nvim-jdtls.nix
```diff
@@ -105,7 +105,7 @@ in {
           else cfg.data;
       in
         [
-          "${pkgs.jdt-language-server}/bin/jdt-language-server"
+          "${lib.getExe pkgs.jdt-language-server}"
```
MattSturgeon, line 108: Why use string templating here instead of an expression?

```suggestion
          (lib.getExe pkgs.jdt-language-server)
```

GaetanLepage: I agree, this is what I had suggested at first.
But I'm trying to stop nitpicking !

khaneliman: Sorry, I missed that in your original review. I just glanced at the suggested change from my phone and thought it was just the lib reference.

GaetanLepage: No worry, I fixed it :)

Issue #1374, opened 2024-04-04
[BUG] nixfmt was renamed to nixfmt-classic


sebastianwebber:
| Field        | Description |
|--------------|-------------|
| Plugin       | `none-ls.sources.formatting`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `24.05 ` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

After my latest flake update I started to get some warnings related both the `nixfmt` and `htmlbeutifier`.

```shell
❯ alias home-switch
home-switch='home-manager switch --flake ~/.config/home-manager'
❯ home-switch
trace: warning: nixfmt was renamed to nixfmt-classic. The nixfmt attribute may be used for the new RFC 166-style formatter in the future, which is currently available as nixfmt-rfc-style
trace: warning: [DEV] Nixvim (plugins.none-ls): Some tools from upstream are not declared locally in `builtinPackages`.
-> [htmlbeautifier]

trace: warning: nixfmt was renamed to nixfmt-classic. The nixfmt attribute may be used for the new RFC 166-style formatter in the future, which is currently available as nixfmt-rfc-style
Starting Home Manager activation
Activating checkFilesChanged
Activating checkLaunchAgents
Activating checkLinkTargets
Activating writeBoundary
Activating installPackages
replacing old 'home-manager-path'
installing 'home-manager-path'
Activating linkGeneration
Cleaning up orphan links from /Users/seba
Creating profile generation 490
Creating home file links in /Users/seba
Activating onFilesChange
Activating setupLaunchAgents
trace: warning: nixfmt was renamed to nixfmt-classic. The nixfmt attribute may be used for the new RFC 166-style formatter in the future, which is currently available as nixfmt-rfc-style
trace: warning: [DEV] Nixvim (plugins.none-ls): Some tools from upstream are not declared locally in `builtinPackages`.
-> [htmlbeautifier]
```

I'm assuming that the possible fix is to change the [package in server.nix](https://github.com/nix-community/nixvim/blob/052012d2d98d0e97f488111380c28cc53812ea48/plugins/none-ls/servers.nix#L57) file to `nixfmt-rfc-style`. Please let me know if I can send a PR.

## Minimal, Reproducible Example (MRE)

This is my nixvim config:
```nix
{ pkgs, lib, ... }:
let
  toLua = str: ''
    lua << EOF
    ${str}
    EOF
[...]

PR #1375, opened 2024-04-04
fix: replace nixfmt

sebastianwebber:
since `nixfmt` package was renamed, this PR replaces the package by the new `nixfmt-rfc-style`.

Related to https://github.com/nix-community/nixvim/issues/1374

### DIFF
--- a/plugins/none-ls/servers.nix
+++ b/plugins/none-ls/servers.nix
@@ -54,7 +54,7 @@ with lib; let
       mdformat
       mdl
       mypy
-      nixfmt
+      nixfmt-rfc-style
       pmd
       prettierd
       proselint


MattSturgeon:
> I still don't know if we should use `nixfmt-classic` (what `nixfmt` has been aliased to, which has the same behavior) or the new `nixfmt-rfc-style`...

Presumably we should offer both, but if users are using a specific formatter (classic nixfmt) then we probably shouldn't change them to a new formatter (rfc-style) without some kinda deprecation notice, warning, or opt-in...

If there's an alias then we may not need to do anything immediately?

Asside: I'm assuming `nixfmt-rfc-style` is functionality identical to the beta `nixpkgs-fmt` that was built during the RFC?

jalil-salame:
> Asside: I'm assuming `nixfmt-rfc-style` is functionality identical to the beta `nixpkgs-fmt` that was built during the RFC?

`nixpkgs-fmt` was built for [RFC 101][1]. [RFC 101][1] was rejected and a rework was proposed ([RFC 166][2]). [RFC 166][2] was accepted and is being implemented into [`nixfmt-rfc-style`][3].

`nixpkgs-fmt` and `nixfmt-rfc-style` have meaningful differences.

[1]: https://github.com/NixOS/rfcs/pull/101
[2]: https://github.com/NixOS/rfcs/pull/166
[3]: https://github.com/NixOS/nixfmt/issues/153

GaetanLepage:
I will go on and replace it with `nixfmt-classic` for now. Like so, we avoid breaking anything on the user side.

PR #1379, opened 2024-04-05
plugins/lsp/pylsp: replace "--replace" to remove warnings

RTUnreal:
removes `substituteStream(): WARNING: '--replace' is deprecated, use --replace-{fail,warn,quiet}.` warnings by reaplacing `--replace`

See https://github.com/NixOS/nixpkgs/pull/260776


MattSturgeon (approved):
Looks like the only other use of stdenv substitute\* is already using `--replace-fail`.

LGTM, thanks

PR #1404, opened 2024-04-11
plugins/cmake-tools: init

nfelber:
This PR adds the cmake-tools plugin to nixvim.

A substantial part of this plugin's default config sets options for external plugins such as dap or overseer (optional), therefore I didn't create nixvim options for these and used generic attribute sets instead as they seem to be beyond the scope of cmake-tools itself and aren't exhaustive. Please let me know if this is not the right approach - one issue I noticed is that large attribute set default values don't render well in the docs.

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -70,4 +70,10 @@
     githubId = 4646110;
     name = "Nikita Shirokov";
   };
+  NathanFelber = {
+    email = "nathan@pascaline.org";
+    github = "nfelber";
+    githubId = 46248772;
+    name = "Nathan Felber";
+  };
 }
--- a/plugins/languages/cmake-tools.nix
+++ b/plugins/languages/cmake-tools.nix
@@ -1,0 +1,165 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "cmake-tools";
+  originalName = "cmake-tools.nvim";
+  defaultPackage = pkgs.vimPlugins.cmake-tools-nvim;
+
+  maintainers = [ helpers.maintainers.NathanFelber ];
+
+  settingsOptions = {
+    cmake_command = helpers.defaultNullOpts.mkStr "cmake" ''
+      This is used to specify cmake command path.
+    '';
+
+    ctest_command = helpers.defaultNullOpts.mkStr "ctest" ''
+      This is used to specify ctest command path.
+    '';
+
+    cmake_regenerate_on_save = helpers.defaultNullOpts.mkBool true ''
+      Auto generate when save CMakeLists.txt.
+    '';
+
+    cmake_generate_options =
+      helpers.defaultNullOpts.mkAttrsOf lib.types.anything { "-DCMAKE_EXPORT_COMPILE_COMMANDS" = 1; }
+        ''
+          This will be passed when invoke `CMakeGenerate`.
+        '';
+
+    cmake_build_options = helpers.defaultNullOpts.mkAttrsOf lib.types.anything { } ''
+      This will be passed when invoke `CMakeBuild`.
+    '';
+
+    cmake_build_directory = helpers.defaultNullOpts.mkStr "out/\${variant:buildType}" ''
+      This is used to specify generate directory for cmake, allows macro expansion, relative to vim.loop.cwd().
+    '';
+
+    cmake_soft_link_compile_commands = helpers.defaultNullOpts.mkBool true ''
+      This will automatically make a soft link from compile commands file to project root dir.
+    '';
+
+    cmake_compile_commands_from_lsp = helpers.defaultNullOpts.mkBool false ''
+      This will automatically set compile commands file location using lsp, to use it, please set `cmake_soft_link_compile_commands` to false.
+    '';
+
+    cmake_kits_path = helpers.defaultNullOpts.mkStr null ''
+      This is used to specify global cmake kits path, see CMakeKits for detailed usage.
+    '';
+
+    cmake_variants_message = {
+      short = helpers.defaultNullOpts.mkAttrsOf lib.types.anything { show = true; } ''
+        Whether to show short message.
+      '';
+
+      long =
+        helpers.defaultNullOpts.mkAttrsOf lib.types.anything
+          {
+            show = true;
+            max_length = 40;
+          }
+          ''
+            Whether to show long message.
+          '';
+    };
+
+    cmake_dap_configuration = {
+      name = helpers.defaultNullOpts.mkStr "cpp" ''
+        Name of the launch configuration.
+      '';
+
+      type = helpers.defaultNullOpts.mkStr "codelldb" ''
+        Debug adapter to use.
+      '';
+
+      request = helpers.defaultNullOpts.mkStr "launch" ''
+        Session initiation method.
+      '';
+    };
+
+    cmake_executor = {
+      name = helpers.defaultNullOpts.mkStr "quickfix" ''
+        Name of the executor.
+      '';
+
+      opts = helpers.defaultNullOpts.mkAttrsOf lib.types.anything { } ''
+        The options the executor will get, possible values depend on the executor type.
+      '';
+    };
+
+    cmake_runner = {
+      name = helpers.defaultNullOpts.mkStr "terminal" ''
+        Name of the runner.
+      '';
+
+      opts = helpers.defaultNullOpts.mkAttrsOf lib.types.anything { } ''
+        The options the runner will get, possible values depend on the runner type.
+      '';
+    };
+
+    cmake_notifications = {
+      runner.enabled = helpers.defaultNullOpts.mkBool true "";
+
+      executor.enabled = helpers.defaultNullOpts.mkBool true "";
+
+      spinner =
+        helpers.defaultNullOpts.mkListOf lib.types.str
+          [
+            "⠋"
+            "⠙"
+            "⠹"
+            "⠸"
+            "⠼"
+            "⠴"
+            "⠦"
+            "⠧"
+            "⠇"
+            "⠏"
+          ]
+          ''
+            Icons used for progress display.
+          '';
+
+      refresh_rate_ms = helpers.defaultNullOpts.mkPositiveInt 100 ''
+        How often to iterate icons.
+      '';
+    };
+  };
+
+  settingsExample = {
+    cmake_regenerate_on_save = false;
+    cmake_build_directory = "build/\${variant:buildtype}";
+    cmake_soft_link_compile_commands = false;
+    cmake_dap_configuration = {
+      name = "Launch file";
+      type = "codelldb";
+      request = "launch";
+      program.__raw = ''
+        function()
+          return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
+        end
+      '';
+      cwd = "\${workspaceFolder}";
+      stopOnEntry = false;
+    };
+    cmake_executor.name = "toggleterm";
+    cmake_runner.name = "toggleterm";
+    cmake_notifications = {
+      spinner = [
+        "▱▱▱▱▱▱▱"
+        "▰▱▱▱▱▱▱"
+        "▰▰▱▱▱▱▱"
+        "▰▰▰▱▱▱▱"
+        "▰▰▰▰▱▱▱"
+        "▰▰▰▰▰▱▱"
+        "▰▰▰▰▰▰▱"
+        "▰▰▰▰▰▰▰"
+      ];
+      refresh_rate_ms = 80;
+    };
+  };
+}
--- a/tests/test-sources/plugins/languages/cmake-tools.nix
+++ b/tests/test-sources/plugins/languages/cmake-tools.nix
@@ -1,0 +1,190 @@
+{
+  empty = {
+    plugins.cmake-tools.enable = true;
+  };
+
+  default = {
+    plugins.cmake-tools = {
+      enable = true;
+
+      settings = {
+        cmake_command = "cmake";
+        ctest_command = "ctest";
+        cmake_regenerate_on_save = true;
+        cmake_generate_options = {
+          "-DCMAKE_EXPORT_COMPILE_COMMANDS" = 1;
+        };
+        cmake_build_options = { };
+        cmake_build_directory = "out/\${variant:buildType}";
+        cmake_soft_link_compile_commands = true;
+        cmake_compile_commands_from_lsp = false;
+        cmake_kits_path = null;
+
+        cmake_variants_message = {
+          short = {
+            show = true;
+          };
+          long = {
+            show = true;
+            max_length = 40;
+          };
+        };
+
+        cmake_dap_configuration = {
+          name = "cpp";
+          type = "codelldb";
+          request = "launch";
+          stopOnEntry = false;
+          runInTerminal = true;
+          console = "integratedTerminal";
+        };
+
+        cmake_executor = {
+          name = "quickfix";
+          opts = { };
+          default_opts = {
+            quickfix = {
+              show = "always";
+              position = "belowright";
+              size = 10;
+              encoding = "utf-8";
+              auto_close_when_success = true;
+            };
+
+            toggleterm = {
+              direction = "float";
+              close_on_exit = false;
+              auto_scroll = true;
+            };
+
+            overseer = {
+              new_task_opts = {
+                strategy.__unkeyed-1 = "terminal";
+              };
+              on_new_task.__raw = ''
+                function(task) end
+              '';
+            };
+
+            terminal = {
+              name = "Main Terminal";
+              prefix_name = "[CMakeTools]: ";
+              split_direction = "horizontal";
+              split_size = 11;
+
+              single_terminal_per_instance = true;
+              single_terminal_per_tab = true;
+              keep_terminal_static_location = true;
+
+              start_insert = false;
+              focus = false;
+            };
+          };
+        };
+
+        cmake_runner = {
+          name = "terminal";
+          opts = { };
+          default_opts = {
+            quickfix = {
+              show = "always";
+              position = "belowright";
+              size = 10;
+              encoding = "utf-8";
+              auto_close_when_success = true;
+            };
+
+            toggleterm = {
+              direction = "float";
+              close_on_exit = false;
+              auto_scroll = true;
+            };
+
+            overseer = {
+              new_task_opts = {
+                strategy.__unkeyed-1 = "terminal";
+              };
+              on_new_task.__raw = ''
+                function(task) end
+              '';
+            };
+
+            terminal = {
+              name = "Main Terminal";
+              prefix_name = "[CMakeTools]: ";
+              split_direction = "horizontal";
+              split_size = 11;
+              single_terminal_per_instance = true;
+              single_terminal_per_tab = true;
+              keep_terminal_static_location = true;
+              start_insert = false;
+              focus = false;
+            };
+          };
+        };
+
+        cmake_notifications = {
+          runner.enabled = true;
+          executor.enabled = true;
+          spinner = [
+            "⠋"
+            "⠙"
+            "⠹"
+            "⠸"
+            "⠼"
+            "⠴"
+            "⠦"
+            "⠧"
+            "⠇"
+            "⠏"
+          ];
+          refresh_rate_ms = 100;
+        };
+      };
+    };
+  };
+
+  example = {
+    plugins.toggleterm.enable = true;
+
+    plugins.cmake-tools = {
+      enable = true;
+
+      settings = {
+        cmake_regenerate_on_save = false;
+        cmake_build_directory = "build/\${variant:buildtype}";
+        cmake_soft_link_compile_commands = false;
+
+        cmake_dap_configuration = {
+          name = "Launch file";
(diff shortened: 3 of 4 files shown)


plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+      console = "integratedTerminal";
+    };
+
+    cmake_executor = {
+      name = "quickfix";
+      opts = {};
+      default_opts = {
+        quickfix = {
+          show = "always"; # "always", "only_on_error"
+          position = "belowright"; # "vertical", "horizontal", "leftabove", "aboveleft", "rightbelow", "belowright", "topleft", "botright", use `:h vertical` for example to see help on them
```
MattSturgeon, line 288: comments in `settingsExample` won't show up in the docs. Most of these should probably be converted to enum types of moved to description text as appropriate, unless they're meant as internal comments for nixvim maintainers.

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+      {
+        runner = helpers.mkCompositeOption "" {
+          enabled = helpers.defaultNullOpts.mkBool true "";
+        };
+
+        executor = helpers.mkCompositeOption "" {
+          enabled = helpers.defaultNullOpts.mkBool true "";
+        };
+
+        spinner = helpers.defaultNullOpts.mkListOf lib.types.str ''[ "⠋" "⠙" "⠹" "⠸" "⠼" "⠴" "⠦" "⠧" "⠇" "⠏" ]'' ''
```
MattSturgeon, line 244: @traxys @GaetanLepage if `[ "⠋" "⠙" "⠹" "⠸" "⠼" "⠴" "⠦" "⠧" "⠇" "⠏" ]` is the [plugin's default](https://github.com/Civitasv/cmake-tools.nvim?tab=readme-ov-file#balloon-configuration), should _our_ default value just be `null`? How do we normally handle defaults like this?

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+        name = helpers.defaultNullOpts.mkStr "quickfix" ''
+          Name of the executor.
+        '';
+
+        opts = helpers.defaultNullOpts.mkAttributeSet "{}" ''
+          The options the executor will get, possible values depend on the executor type. See `default_opts` for possible values.
+        '';
+
+        default_opts =
+          helpers.defaultNullOpts.mkAttributeSet ''
```
MattSturgeon, line 111: Would some of these `mkAttributeSet` options be better off as submodues? How do we normally handle this while sticking to [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)?

plugins/languages/cmake-tools.nix
nfelber: Thanks for the review; these comments are copied from the default cmake-tools configuration upstream. I wasn´t aware that they wouldn't show up in the docs - then there is no point leaving them for nixvim maintainers. I'll just wait for further input to your other comments before proceeding, and accordingly remove them or move them to option descriptions.

MattSturgeon: Personally, I think we should use enum options when upstream document a list of acceptable values.

Unless that list is likely to change frequently; that'd become a maintainance burden.

GaetanLepage: > Personally, I think we should use enum options when upstream document a list of acceptable values.
> 
> Unless that list is likely to change frequently; that'd become a maintainance burden.

100% agree.

GaetanLepage:
I haven't read everything, but keep in mind that it is absolutely not mandatory to package every single option.
This one has a lot of options, so we might be better of selecting the most valuable ones.

As you have already went through the work, we can also keep them all, but keep in mind that it is not a requirement.

nfelber:
I'm not sure I understand what would be best to do. On one hand I totally agree that not every option should be packaged in nixvim. This is partly why I left several options as attribute sets instead of composite options. On the other hand, I also agree that options listed upstream would be better packaged as enum values whenever possible, but I don't know how to properly satisfy both constraints.

For example, this is the default upstream `cmake_dap_configuration` value:

```lua
cmake_dap_configuration = { -- debug settings for cmake
  name = "cpp",
  type = "codelldb",
  request = "launch",
  stopOnEntry = false,
  runInTerminal = true,
  console = "integratedTerminal",
}
```

I packaged it as a simple attribute set for the following reasons:

- To limit the number of available options as discussed above.
- This config is directly passed to nvim-dap (if used), and is not immediately related to cmake-tools.
- While some attributes like `name`, `type` and `request` are required by nvim-dap, the others are either optional or specific to the debug adapter ("codelldb" in this case), and listing them all would be very difficult to maintain.
- If I were to add proper options for the `name`, `type` and `request` attributes (strings and enum), I don't think arbitrary additional attributes would be supported anymore  (?).

A similar reasoning applies to the `cmake_executor.default_opts` and `cmake_runner.default_opts` options.

Hopefully this clarifies the confusion, please let me know what you think and how to go about it.

plugins/languages/cmake-tools.nix
GaetanLepage: This is the right approach. The default provided to the `helpers.defaultNullOpts.*` helpers is the one of the plugin.
The actual default value of the nixvim option will always be `null`.

MattSturgeon:
@nfelber do you need any help with this one and are you still working on it?

If you prefer, one of us could continue what you started when we get chance. Otherwise we can answer any specific questions (here or on matrix) and do another round of reviews.

nfelber:
@MattSturgeon yes, I'd like to continue working on it. I haven't made progress in a while because I'm unsure what direction to take to improve the packaging, as described in my comment above.

If you could have a look, that would be very helpful. And feel free to let me know if anything is unclear.

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+          ''
+            A list of default and possible values for executors.
+          '';
+      };
+
+    cmake_runner =
+      helpers.mkCompositeOption ''
+        Runner to use.
+      ''
+      {
```
MattSturgeon, line 168: Again, prefer not using composite options, but you can if you think the description is necessary:

```nix
cmake_runner = {
```

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+                # Running Tasks
+                start_insert = false; # If you want to enter terminal with :startinsert upon using :CMakeRun
+                focus = false; # Focus on terminal when cmake task is launched.
+              }; # terminal executor uses the values in cmake_terminal
+            }
+          ''
+          ''
+            A list of default and possible values for executors.
+          '';
+      };
```
MattSturgeon, line 162: Same as `default_opts`, make this into a set of options if you think it's worth keeping.

If it's the same options that `default_opts` uses you can declare them in a `let in` block and reuse them, or perhaps make a `submodule` type that can be reused.

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+          type = "codelldb";
+          request = "launch";
+          stopOnEntry = false;
+          runInTerminal = true;
+          console = "integratedTerminal";
+        }
+      ''
+      ''
+        Dap configuration for cmake.
+      '';
```
MattSturgeon, line 95:
```suggestion
    cmake_dap_configuration =
      helpers.defaultNullOpts.mkAttrsOf lib.types.anything {
          name = "cpp";
          type = "codelldb";
          request = "launch";
          stopOnEntry = false;
          runInTerminal = true;
          console = "integratedTerminal";
        } ''
        Dap configuration for cmake.
      '';
```

- We no longer quote plugin-defaults
- `attrsOf anything` is preferred over `attrs`
- If `name`, `type`, `request`, etc are well known options, you could make them actual options instead. Or define a submodule type to use here.

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+
+    cmake_variants_message =
+      helpers.mkCompositeOption ''
+        Settings for cmake variants messages.
+      ''
+      {
+        short =
+          helpers.defaultNullOpts.mkAttributeSet ''
+            { show = true; }
+          ''
```
MattSturgeon, line 68:
```suggestion
        short =
          helpers.defaultNullOpts.mkAttrsOf lib.types.anything { show = true; }
```

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+
+    cmake_kits_path = helpers.defaultNullOpts.mkStr null ''
+      This is used to specify global cmake kits path, see CMakeKits for detailed usage.
+    '';
+
+    cmake_variants_message =
+      helpers.mkCompositeOption ''
+        Settings for cmake variants messages.
+      ''
+      {
```
MattSturgeon, line 64: Again:

```suggestion
    cmake_variants_message = {
```

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+    '';
+
+    cmake_regenerate_on_save = helpers.defaultNullOpts.mkBool true ''
+      Auto generate when save CMakeLists.txt.
+    '';
+
+    cmake_generate_options =
+      helpers.defaultNullOpts.mkAttributeSet ''
+        { "-DCMAKE_EXPORT_COMPILE_COMMANDS" = 1; }
+      ''
```
MattSturgeon, line 35:
```suggestion
      helpers.defaultNullOpts.mkAttrsOf lib.types.anything {
        "-DCMAKE_EXPORT_COMPILE_COMMANDS" = 1;
      }
```

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+      ''
+        This will be passed when invoke `CMakeGenerate`.
+      '';
+
+    cmake_build_options = helpers.defaultNullOpts.mkAttributeSet "{}" ''
```
MattSturgeon, line 40:
```suggestion
    cmake_build_options = helpers.defaultNullOpts.mkAttrsOf lib.types.anything { } ''
```

lib/maintainers.nix
MattSturgeon, line 1: The maintainers commit should be titled:

> `maintainers: add NathanFelber`

MattSturgeon (changes requested):
I hope this is helpful. I've gone over everything quickly, so probably missed some stuff that'll be picked up in later reviews.

Note, you'll also need to run `nix fmt` in order for CI to pass.

Let me know here or on matrix if you have any questions of if you'd like me to push up any changes to your branch.

tests/test-sources/plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,143 @@
+{
+  empty = {
+    plugins.cmake-tools.enable = true;
+  };
+
+  example = {
```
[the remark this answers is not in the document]
nfelber: Yes, this is correct. Thanks for pointing it out.

tests/test-sources/plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,143 @@
...
+              auto_scroll = true;
+            };
+
+            overseer = {
+              new_task_opts = {
+                strategy.__raw = ''
+                  {
+                    "terminal",
+                  },
+                '';
```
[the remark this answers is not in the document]
nfelber: Yes, it is indeed an attrs type, I had no idea how to implement it - this is exactly what I need. Isn't it `strategy.__unkeyed-1` however (from looking at `to-lua.nix`)?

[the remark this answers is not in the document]
MattSturgeon: Yes, that was a typo on my end

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+          ''
+            A list of default and possible values for runner.
+          '';
+      };
+
+    cmake_notifications =
+      helpers.mkCompositeOption ''
+        Notification settings.
+      ''
+      {
```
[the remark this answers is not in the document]
nfelber: That is cleaner indeed. I probably thought every attribute had to be an option for some reason.

nfelber:
Thank you very much for your detailed feedback. It is incredibly helpful. I'll continue going through the remaining suggestions when I get time later this week.

plugins/languages/cmake-tools.nix
nfelber: In this case, `name`, `type` and `request` are well known required options while the others are optional and depend on the debug adapter being used (e.g. `codelldb`). It would be nice to package the first three but also let the user add arbitrary attributes, and I'm not sure how to properly do that. mkCompositeOption doesn't seem to take a plugin default, and mkAttrsOf anything doesn't seem to take child options. Any idea?

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,391 @@
...
+                # Running Tasks
+                start_insert = false; # If you want to enter terminal with :startinsert upon using :CMakeRun
+                focus = false; # Focus on terminal when cmake task is launched.
+              };
+            }
+          ''
+          ''
+            A list of default and possible values for runner.
+          '';
+      };
```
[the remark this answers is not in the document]
nfelber: Yes, thinking twice about it, `default_opts` are probably not worth keeping. I'll have a closer look, but I don't think they are supposed to be modified anyways as they are part of the `cmake-tools.nvim` default config and are overridden by the user-defined `opts`. And they are also specific to optional external plugins, therefore out of the scope of `cmake-tools.nvim` as far as option packaging is concerned.

plugins/languages/cmake-tools.nix
MattSturgeon: > It would be nice to package the first three but also let the user add arbitrary attributes, and I'm not sure how to properly do that

The `settings` option has a "freeform" type (`attrsOf anything`), so this is the default behaviour, unless we add additional restrictions via option types.

If that's what you want here then all you need to do is:

```nix
cmake_dap_configuration = {
  name = /*mkOption*/;
  type = /*mkOption*/;
  request = /*mkOption*/;
};
```

If this needs to be done in a re-usable way, then there's a few ways to do this. The simplest would be to define the option attrs in a let block and re-use it that way.

With this approach you could also define "extra" options using the `commonOpts // {/* extraOpts */}` pattern (if you wanted to).

---

It's also possible to do this kinda thing with submodules, however `mkCompositeOption` is probably not the best way to define a submodule option, because it doesn't set a freeform type and AFAIK submodules don't inherit the parent's freeform type.

A submodule option _would_ allow you to define things like `example`, `description` and `pluginDefault` though.

I'd suggest ignoring submodules for now, focus on getting something simple working and then revisit the idea later if you're not satisfied with the end result.

tests/test-sources/plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,143 @@
...
+              };
+              on_new_task = {
+                __raw = ''
+                  function(task) end
+                '';
+              };
```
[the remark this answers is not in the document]
nfelber: As I've decided to not package this option, I'll stick with `__raw` in the test file.

plugins/languages/cmake-tools.nix
nfelber: I see, thank you for the clarification. Then I'll go the simplest route and only declare these three options as you suggest.

MattSturgeon:
@nfelber looks like you've rebased `main` onto your branch instead of your branch onto `main`

nfelber:
Oh, so that's what happened. My bad, I'll try to fix it later today.

nfelber:
@MattSturgeon Looks like it's now fixed. I think I integrated all your suggestions into the PR and it should be ready for review. However, some of the checks are failing and I can't test locally as my system (16 GB RAM + 16 GB swap) runs out of memory while running `nix flake check --all-systems`  and kills the process. Is that normal?

MattSturgeon:
> @MattSturgeon Looks like it's now fixed. I think I integrated all your suggestions into the PR and it should be ready for review. However, some of the checks are failing and I can't test locally as my system (16 GB RAM + 16 GB swap) runs out of memory while running `nix flake check --all-systems` and kills the process. Is that normal?

Try entering `nix develop` (to get a devshell) and using the `tests` command, instead of running `nix flake check`.

You can also try omitting `--all-systems` as that'll make the tests _much_ heavier.

For additional context, we're having issues with our CI currently, so ignore any ci failure complaining about "broke pipe".

Testing locally shouldn't have that issue though.

Also: @traxys recently managed to dramatically improve the memory usage of the tests, you should benefit from that now you've rebase.

Before the tests needed ~32gb ram, now they use around 3gb or less.

MattSturgeon:
The fact the CI is passing on most systems, and failing with "broken pipe" on two tells me the tests are passing for this PR.

I'll review again when I have chance, hopefully over the weekend!

Thanks for your efforts!

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,134 @@
...
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "cmake-tools";
+  originalName = "cmake-tools.nvim";
+  defaultPackage = pkgs.vimPlugins.cmake-tools-nvim;
+
+  maintainers = [ helpers.maintainers.NathanFelber ];
+
+  extraPlugins = with pkgs.vimPlugins; [ plenary-nvim ];
```
GaetanLepage, line 15: This should be dealt with at the nixpkgs level.

plugins/languages/cmake-tools.nix
```diff
@@ -0,0 +1,134 @@
...
+
```
GaetanLepage, line 16: Could you please add a `settingsExample` section as well ?

nfelber:
> Try entering `nix develop` (to get a devshell) and using the `tests` command, instead of running `nix flake check`.
> 
> You can also try omitting `--all-systems` as that'll make the tests _much_ heavier.

I see, thanks for the tips and context. I'll give it another try tomorrow then!

plugins/languages/cmake-tools.nix
MattSturgeon: Note for context: I'd previously suggested removing the example as it just contained the plugin defaults.

If we can come up with a useful non-default example that'd be ideal though.

It could be extensive, or it could just set one or two "important" or frequently configured options.

Maybe try searching `lang:lua "cmake-tools"` accross GitHub to see real world examples, or check if upstream has any examples in its docs?

nfelber:
> The fact the CI is passing on most systems, and failing with "broken pipe" on two tells me the tests are passing for this PR.
> 
> I'll review again when I have chance, hopefully over the weekend!

Awesome, thank you then!

plugins/languages/cmake-tools.nix
nfelber: The dependency on `plenary.nvim` you mean? I can have a look when I get some time.

plugins/languages/cmake-tools.nix
nfelber: Sure, I'll add one with non-default values.

plugins/languages/cmake-tools.nix
GaetanLepage: https://github.com/NixOS/nixpkgs/pull/328534

plugins/languages/cmake-tools.nix
nfelber: I added a basic example. It only slightly deviates from the default configuration, but in what I consider could be a useful (or at least usable) way. I wasn't able to find examples online or upstream unfortunately, but the `cmake_dap_configuration` part comes from [the nvim-dap wiki](https://github.com/mfussenegger/nvim-dap/wiki/C-C---Rust-(via--codelldb)).

GaetanLepage (changes requested):
Good !
Could you please squash your commits so that you have only the two following:
- `helpers/maintainers: add NathanFelber`
- `plugins/cmake-tools: init`

nfelber:
Done. Let me know if there's anything else I can do!

GaetanLepage (approved):
LGTM !
Thanks @nfelber and the reviewers for this nice work !

GaetanLepage (changes requested):
We need to wait for the next flake update to get the `plenary` dependency in.
In the meantime, you can remove the relevant line from your patch.

plugins/languages/cmake-tools.nix
MattSturgeon: We should have https://github.com/NixOS/nixpkgs/pull/328534 in nixos-unstable within a day or two and our CI should also be back online soon.

Once we've updated our lockfile you'll be able to rebase and remove this line. I think we can wait for that before merging.

MattSturgeon (approved):
No issues from me, but lets hold off until we can run our CI checks and drop the plenary line.

Thanks for your hard work and persistence on this!

nfelber:
Yes, no problem. I just amended my last commit with the relevant line removed.
And thank you for taking the time to guide me through the process!

GaetanLepage (approved):
All good for me !
Thanks @nfelber

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio refresh

MattSturgeon:
@mergifyio dequeue

PR #1408, opened 2024-04-12
plugins/coq-nvim: fix settings only being set when lsp is enabled

Kareem-Medhat:


plugins/completion/coq.nix
```diff
@@ -67,9 +67,14 @@ with lib;
...
+      globals = {
+        coq_settings = cfg.settings;
+      };
+
+      extraConfigLua = "require('coq')";
+
       plugins.lsp = {
         preConfig = ''
-          vim.g.coq_settings = ${helpers.toLuaObject cfg.settings}
           local coq = require 'coq'
```
traxys, line 78: This is superflous now, isn't it?

Kareem-Medhat: it is needed for use in "setupWrappers" as it references that variable

traxys (approved):

PR #1409, opened 2024-04-13
Format init.lua using StyLua

MattSturgeon:
@GaetanLepage mentioned it'd be nice to format `init.lua` using stylua. This PR does that by introducing a new derivation builder `helpers.writeLua`.

The options parameter is mostly there to allow for unforeseen future use-cases. It's currently unused.

This is currently only used for `init.lua`, the other files (e.g. ftplugins) will also need to migrate to `writeLua` to benefit.

In the future, we could expose an option to have either raw, minified, or prettified lua files, although this probably isn't really necessary.

### DIFF
--- a/lib/builders.nix
+++ b/lib/builders.nix
@@ -1,0 +1,30 @@
+{
+  lib,
+  pkgs,
+}: {
+  /*
+  Write a lua file to the nix store, formatted using stylua.
+
+  # Type
+
+  ```
+  writeLua :: String -> String -> Derivation
+  ```
+
+  # Arguments
+
+  - [name] The name of the derivation
+  - [text] The content of the lua file
+  */
+  writeLua = name: text:
+    pkgs.runCommand name {inherit text;} ''
+      echo -n "$text" > "$out"
+
+      ${lib.getExe pkgs.stylua} \
+        --no-editorconfig \
+        --line-endings Unix \
+        --indent-type Spaces \
+        --indent-width 4 \
+        "$out"
+    '';
+}
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -4,6 +4,7 @@
   _nixvimTests,
   ...
 }: let
+  nixvimBuilders = import ./builders.nix {inherit lib pkgs;};
   nixvimTypes = import ./types.nix {inherit lib nixvimOptions;};
   nixvimUtils = import ./utils.nix {inherit lib _nixvimTests;};
   nixvimOptions = import ./options.nix {inherit lib nixvimTypes nixvimUtils;};
@@ -20,3 +21,4 @@ in
   }
   // nixvimUtils
   // nixvimOptions
+  // nixvimBuilders
--- a/wrappers/modules/output.nix
+++ b/wrappers/modules/output.nix
@@ -2,6 +2,7 @@
   pkgs,
   config,
   lib,
+  helpers,
   ...
 }:
 with lib; {
@@ -115,7 +116,7 @@ with lib; {
       ''
       + config.content;
 
-    init = pkgs.writeText "init.lua" customRC;
+    init = helpers.writeLua "init.lua" customRC;
     initPath = toString init;
 
     extraWrapperArgs = builtins.concatStringsSep " " (


GaetanLepage:
This looks great to me.
The similar `pkgs.writeShellApplication` has the `text` as an argument in the "options" attrs.
Should we do that as well ?

MattSturgeon:
> The similar `pkgs.writeShellApplication` has the `text` as an argument in the "options" attrs.
> Should we do that as well ?

Do you prefer using `pkgs.writeText "name" "text"` or [its sister](https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing) `pkgs.writeTextFile { name = "name"; text = "text"; }`?

GaetanLepage:
> > The similar `pkgs.writeShellApplication` has the `text` as an argument in the "options" attrs.
> > Should we do that as well ?
> 
> Do you prefer using `pkgs.writeText "name" "text"` or [its sister](https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing) `pkgs.writeTextFile { name = "name"; text = "text"; }`?

Let's go for the first one now as it is the most compact.
There is no issue with changing it later as this is basically an internal helper.

MattSturgeon:
> There is no issue with changing it later as this is basically an internal helper.

Following that philosophy, is there any reason to include the options param at this point, or should we assume the defaults will always be used and simplify further by dropping the argument?

GaetanLepage:
> Following that philosophy, is there any reason to include the options param at this point, or should we assume the defaults will always be used and simplify further by dropping the argument?

Exactly.

GaetanLepage:
Diff LGTM.
Have you tested this patch locally (on your config for instance) ?

MattSturgeon:
> Have you tested this patch locally (on your config for instance) ?

Tested by printing `packages.x86_64-linux.nvim.config.initPath` in `nix repl` and then opening the printed path in vim.

Vim still runs ok, and the init file appears to be formatted correctly.

I haven't checked if this affects the init file written to `~/.config/nvim` (etc) by the home-manager/nixos/darwin modules, but I'm assuming that's just a copy of `packages.x86_64-linux.nvim.config.initPath`.

GaetanLepage:
> > Have you tested this patch locally (on your config for instance) ?
> 
> Tested by printing `packages.x86_64-linux.nvim.config.initPath` in `nix repl` and then opening the printed path in vim.
> 
> Vim still runs ok, and the init file appears to be formatted correctly.
> 
> I haven't checked if this affects the init file written to `~/.config/nvim` (etc) by the home-manager/nixos/darwin modules, but I'm assuming that's just a copy of `packages.x86_64-linux.nvim.config.initPath`.

Cool !

GaetanLepage (approved):

lib/builders.nix
```diff
@@ -0,0 +1,31 @@
...
+    raw = pkgs.writeText name text;
+    stylua = lib.getExe pkgs.stylua;
+  in
+    pkgs.runCommand name {} ''
+      ${stylua} \
+        --no-editorconfig \
+        --line-endings Unix \
+        --indent-type Spaces \
+        --indent-width 4 \
+        -- - < ${raw} 1> $out
```
MattSturgeon, line 29: This might be more readable if we do something like

```shell
cp -TL ${raw} $out
${stylua} $out
```

`-- - < ${raw} 1> $out` is a little cryptic...

lib/builders.nix
```diff
@@ -0,0 +1,31 @@
...
+  # Arguments
+
+  - [name] The name of the derivation
+  - [text] The content of the lua file
+  */
+  writeLua = name: text: let
+    raw = pkgs.writeText name text;
+    stylua = lib.getExe pkgs.stylua;
+  in
+    pkgs.runCommand name {} ''
```
traxys, line 23: Couldn't we use the env vars to pass the raw text, and use ` <<< "$luaFile` instead of creating an unformatted text file in the store?

lib/builders.nix
GaetanLepage: Not a huge deal for me ?
@traxys any opinion ?

traxys (changes requested):

lib/builders.nix
MattSturgeon: Was just about to ask if `lib.escapeShellArg` was preferable to writing an extra file to the store. I take it you think so?

I figured it added extra overhead escaping a large/complex file, but writing redundant stuff to the store isn't ideal either.

traxys: Env vars shold avoide the neeed to esacape

MattSturgeon: Oh you mean the env vars attrset passed to `runCommand`, good idea

traxys (approved):

MattSturgeon:
Looks like `extraFiles` will need a variant of `writeLua` that writes to a subdir within `$out`. `pkgs.writeTextFile` [handles this](https://github.com/NixOS/nixpkgs/blob/92aece79b3c74d0af58475b477868864aa072fc1/pkgs/build-support/trivial-builders/default.nix#L112-L113) with an optional `destination` attribute.

traxys:
> Looks like `extraFiles` will need a variant of `writeLua` that writes to a subdir within `$out`. `pkgs.writeTextFile` [handles this](https://github.com/NixOS/nixpkgs/blob/92aece79b3c74d0af58475b477868864aa072fc1/pkgs/build-support/trivial-builders/default.nix#L112-L113) with an optional `destination` attribute.
> 
> This can be refactored in a followup PR

Yeah I guess we need the same hierarchy of functions that writeTextFile & writeText.
I think we should do that in a subsequent PR

PR #1411, opened 2024-04-13
plugins/improved-search: init

### LINKED ISSUE #1338: [PLUGIN REQUEST] improved-search.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | improved-search.nvim  | 
| Homepage | [github:backdround/improved-search.nvim](https://github.com/backdround/improved-search.nvim)     | 
| Nixpkgs  | false | 

## Extra Information

Would be nice to have options to enable some (or all) of the suggested configuration options presented in the readme, as I think most people will use it in this way.

GaetanLepage:
Add support for [improved-search.nvim](https://github.com/backdround/improved-search.nvim).

Fixes #1338

### DIFF
--- a/plugins/utils/improved-search.nix
+++ b/plugins/utils/improved-search.nix
@@ -1,0 +1,123 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+# This plugin is only configured through keymaps, so we use `mkVimPlugin` without the
+# `globalPrefix` argument to avoid the creation of the `settings` option.
+  helpers.vim-plugin.mkVimPlugin config {
+    name = "improved-search";
+    originalName = "improved-search.nvim";
+    defaultPackage = pkgs.vimPlugins.improved-search-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    extraOptions = {
+      keymaps = mkOption {
+        description = ''
+          Keymap definitions for search functions
+
+          See [here](https://github.com/backdround/improved-search.nvim?tab=readme-ov-file#functions-and-operators) for the list of available callbacks.
+        '';
+        type = with helpers.nixvimTypes;
+          listOf (
+            submodule {
+              options = {
+                key = mkOption {
+                  type = str;
+                  description = "The key to map.";
+                  example = "!";
+                };
+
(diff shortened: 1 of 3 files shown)


plugins/utils/improved-search.nix
```diff
@@ -0,0 +1,122 @@
...
+                        "forward"
+                        "forward_strict"
+                        "backward"
+                        "backward_strict"
+                      ]
+                    );
+                  description = ''
+                    The action to execute.
+
+                    See https://github.com/backdround/improved-search.nvim?tab=readme-ov-file#functions-and-operators
```
MattSturgeon, line 57: IMO this'd be better as a markdown link, matching the keymaps description
```suggestion
                    See [here](https://github.com/backdround/improved-search.nvim?tab=readme-ov-file#functions-and-operators) for the list of available callbacks.
```

May also be nice to define the enum list in a `let` block and include it directly in the description text too?

Is it worth mentioning that the action can be raw lua, or does the type information make that obvious?

GaetanLepage: > May also be nice to define the enum list in a `let` block and include it directly in the description text too?

Not sure this would be necessary as the list of allowed values is already shown on the documentation.

> Is it worth mentioning that the action can be raw lua, or does the type information make that obvious?

Same, it will be visible on the documentation + Here it would be a fairly niche usage. So I would say no, we don't need to add it explicitly to the description.

plugins/utils/improved-search.nix
```diff
@@ -0,0 +1,122 @@
...
+          }
+          {
+            mode = "x";
+            key = "#";
+            action = "backward";
+          }
+          {
+            mode = "n";
+            key = "|";
+            action = "in_place";
```
MattSturgeon, line 101: It'd be nice if one or two examples & tests showed using the keymap `options` submodule, e.g. to set `desc`.

MattSturgeon (approved):
Left a comment regarding keymaps, but don't think it should block merging.

Otherwise diff LGTM (not tested)

plugins/utils/improved-search.nix
GaetanLepage: I added such an example.

MattSturgeon (approved):

PR #1422, opened 2024-04-15
colorschemes/rose-pine: switch to mkNeovimPlugin

GaetanLepage:


MattSturgeon (approved):
`settings.enable.*` is pretty cursed, maybe this is an example of somewhere we should stray from rfc42 and provide some better naming abstractions?

Regardless, the changes look good. LGTM

GaetanLepage:
> is pretty cursed, maybe this is an example of somewhere we should stray from rfc42 and provide some better naming abstractions?

By "cursed" you mean that it could get interpreted as a boolean option ?
Here we follow the upstream structure, so this does not bother me too much.
Also, providing `colorschemes.rose-pine.settings.enable = true` would lead to an error as it expects an `attrs`?

GaetanLepage:
> > is pretty cursed, maybe this is an example of somewhere we should stray from rfc42 and provide some better naming abstractions?
> 
> By "cursed" you mean that it could get interpreted as a boolean option ? Here we follow the upstream structure, so this does not bother me too much. Also, providing `colorschemes.rose-pine.settings.enable = true` would lead to an error as it expects an `attrs`.

PR #1430, opened 2024-04-16
plugins/cmp-async-path: init

FelixBoehme:
Add support for [cmp-async-path](https://codeberg.org/FelipeLema/cmp-async-path).

### DIFF
--- a/plugins/completion/cmp/sources.nix
+++ b/plugins/completion/cmp/sources.nix
@@ -1,5 +1,6 @@
 # Maps each source name to the plugin that provides it
 {
+  async_path = "cmp-async-path";
   buffer = "cmp-buffer";
   calc = "cmp-calc";
   dap = "cmp-dap";
--- a/tests/test-sources/plugins/completion/cmp-async-path.nix
+++ b/tests/test-sources/plugins/completion/cmp-async-path.nix
@@ -1,0 +1,16 @@
+{
+  example = {
+    plugins.cmp = {
+      enable = true;
+      settings.sources = [
+        {
+          name = "async_path";
+          option = {
+            trailing_slash = false;
+            label_trailing_slash = true;
+          };
+        }
+      ];
+    };
+  };
+}


GaetanLepage:
Thanks for the contribution ! Have you been able to test it locally ?

FelixBoehme:
It works locally, but I forgot to check whether the auto enabling of the sources was working, which was indeed not working, because the plugin uses an underscore instead of a hyphen.
Now it is fixed and the plugin works, as well as the auto enabling.

And quick question, because I wanted to double check and haven't been able to figure it out, how do I obtain the store path of the currently used init.lua?

Btw. great project, I have really enjoyed using it so far, keep up the great work.

MattSturgeon:
> how do I obtain the store path of the currently used init.lua

You can inspect the NixVim build in `nix repl` and print out `config.initPath`.

E.g.

```
$ nix repl
>> :lf .
>> packages.x86_64-linux.nvim.config.initPath
```

FelixBoehme:
> > how do I obtain the store path of the currently used init.lua
> 
> One method is to inspect the build using `nix repl`, in particular you want `config.initPath`.
> 
> E.g.
> 
> ```
> $ nix repl
> >> :lf .
> >> nvim = packages.x86_64-linux.nvim
> >> nvim.config.initPath
> ```

Thanks for the answer, sadly this doesn't work for me as `packages.x86_64-linux` is an empty attribute set for me. 
I would suspect that flake-parts is causing the issue, and the path to nvim probably has to be different but I'm not sure.

MattSturgeon:
> `packages.x86_64-linux` is an empty attribute set for me. 
I would suspect that flake-parts is causing the issue.

Nothing to do with flake-parts, I'm also using that. It'll either be

- You forgot to load your flake into the repl (`:lf <path>`)
- You're not using standalone nixvim (hence no nvim package)
- You're using `legacyPackages` instead of `packages`
  - you can run `nix flake show` to see what you should be printing, or use tab-completion in the repl

MattSturgeon (approved):
> Now it is fixed and the plugin works, as well as the auto enabling.

If this works locally then I'm happy

FelixBoehme:
> > `packages.x86_64-linux` is an empty attribute set for me.
> > I would suspect that flake-parts is causing the issue.
> 
> Nothing to do with flake-parts, I'm also using that. It'll either be
> 
>     * You forgot to load your flake into the repl (`:lf <path>`)
> 
>     * You're not using standalone nixvim (hence no nvim package)
> 
>     * You're using `legacyPackages` or some other output attr instead of `packages`
>       
>       * you can run `nix flake show` to see your outputs, or use tab-completion in the repl
> 
> 
> _Sorry, I accidentally edited your comment instead of posting this reply. Mobile GitHub can be fun sometimes... Should be fixed now._

Finally found it, as you said it was because I'm not using the standalone version. Don't know how I didn't find it yesterday when searching for it but now I got it.
Thanks again for your help and patience, really appreciate it.

GaetanLepage:
Could you please add some test case somewhere ?

FelixBoehme:
Hope this simple test is what you were looking for, let me know if anything else comes up.

GaetanLepage:
@FelixBoehme we have made the necessary changes to the `cmp` plugin.
Please, rebase on `main` and address the conflicts.

GaetanLepage (approved):
All good ! Thanks for your patience @FelixBoehme !

PR #1436, opened 2024-04-19
plugins/lsp: add ruff language server

GaetanLepage:


traxys:
What is the difference with ruff-lsp

GaetanLepage:
> What is the difference with ruff-lsp?

This is the rust implementation that directly uses the `ruff` binary. The python implementation `ruff-lsp` will still be maintained while this one reaches a more stable state.

traxys (approved):

PR #1444, opened 2024-04-22
plugins/gitignore: init

### LINKED ISSUE #1418: [PLUGIN REQUEST] `gitignore.nvim`
| Field    | Description |
|----------|-------------|
| Plugin   | `gitignore.nvim`  | 
| Homepage | `https://github.com/wintermute-cell/gitignore.nvim`     | 
| Nixpkgs  | `false` | 

## Extra Information

From the README:

> A neovim plugin for generating .gitignore files in seconds, by allowing you to select from a huge number of different technologies.

GaetanLepage:
Add support for [gitignore.nvim](https://github.com/wintermute-cell/gitignore.nvim).

Fixes #1418

### DIFF
--- a/plugins/git/gitignore.nix
+++ b/plugins/git/gitignore.nix
@@ -1,0 +1,68 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+# We use `mkVimPlugin` to avoid having a `settings` option.
+# Indeed, this plugin is not configurable in the common sense (no `setup` function).
+  helpers.vim-plugin.mkVimPlugin config {
+    name = "gitignore";
+    originalName = "gitignore.nvim";
+    defaultPackage = pkgs.vimPlugins.gitignore-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    extraOptions = {
+      keymap = mkOption {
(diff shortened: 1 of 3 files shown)


plugins/git/gitignore.nix
```diff
@@ -0,0 +1,68 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+# We use `mkVimPlugin` to avoid having a `settings` option.
```
MattSturgeon, line 9: Wouldn't it be better to provide a way for `mkNeovimPlugin` to do this? E.g. a boolean parameter

GaetanLepage: Maybe. I recently thought about merging those two helpers as they are very redundant.

MattSturgeon: Not sure that's necessary for this PR.

As this plugin doesn't need us to call `setup()` it probably isn't an issue to use `mkVimPlugin` anyway.

Maybe just also mention that in the comment, to avoid confusion?

GaetanLepage: I clarified the comment. Will think about unifying the helpers later :)

MattSturgeon (approved):

PR #1446, opened 2024-04-22
plugins/godot: init

### LINKED ISSUE #1359: [PLUGIN REQUEST] vim-godot
| Field    | Description |
|----------|-------------|
| Plugin   | `vim-godot`  | 
| Homepage | `https://github.com/habamax/vim-godot`     | 
| Nixpkgs  | `false` | 

## Extra Information
The plugin has basically no config options, but does provide instructions for enabling its completions in popular completion plugins like `cmp`. Maybe additional config options could be added there to auto-enable this plugin?

GaetanLepage:
Add support for [vim-godot](https://github.com/habamax/vim-godot).

Fixes #1359

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -49,6 +49,7 @@
 
     ./languages/clangd-extensions.nix
     ./languages/debugprint.nix
+    ./languages/godot.nix
     ./languages/haskell-scope-highlighting.nix
     ./languages/helm.nix
     ./languages/julia/julia-cell.nix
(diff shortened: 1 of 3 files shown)


traxys:
Shouldn't we add godot to the path? Or is that implemented in nixpkgs?

GaetanLepage:
> Shouldn't we add godot to the path? Or is that implemented in nixpkgs?

No you are right. I forgot to add it. Done now.

MattSturgeon (approved):

PR #1447, opened 2024-04-22
plugins/spectre: init

### LINKED ISSUE #1403: [PLUGIN REQUEST] nvim-spectre
| Field    | Description |
|----------|-------------|
| Plugin   | `nvim-spectre`  | 
| Homepage | `https://github.com/nvim-pack/nvim-spectre`     | 
| Nixpkgs  | `https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/editors/vim/plugins/generated.nix#L8105` | 

## Extra Information

I started this super rough MR: https://github.com/nix-community/nixvim/pull/974,
which is very out of date, doesn't even use the new way of doing things but mostly I was stuck on how to do the mappings properly.

GaetanLepage:
Add support for [nvim-spectre](https://github.com/nvim-pack/nvim-spectre).

Note: I didn't implement any auto-installation method for the find/replace tools (`rg`, `ag`, ...).

Fixes #1403

### DIFF
--- a/plugins/utils/spectre.nix
+++ b/plugins/utils/spectre.nix
@@ -1,0 +1,283 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+  helpers.neovim-plugin.mkNeovimPlugin config {
+    name = "spectre";
+    originalName = "nvim-spectre";
+    defaultPackage = pkgs.vimPlugins.nvim-spectre;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    description = ''
+      You may want to set the package for your find/replace tool(s) like shown below:
+
+      ```nix
+        plugins.spectre.findPackage = pkgs.rg;
+        plugins.spectre.replacePackage = pkgs.gnused;
+      ```
+    '';
+
+    settingsOptions = let
+      mkEngineOption = type:
+        helpers.mkNullOrOption
+        (
+          with types;
+            attrsOf (submodule {
+              options = {
+                cmd = mkOption {
+                  type = types.str;
+                  description = "Executable to run.";
+                };
+
+                args = helpers.defaultNullOpts.mkListOf types.str "[]" ''
+                  List of arguments to provide to the engine.
+                '';
+
+                options =
+                  helpers.defaultNullOpts.mkAttrsOf
+                  (types.submodule {
+                    options = {
+                      value = mkOption {
+                        type = types.str;
+                        example = "-i";
+                        description = "The option flag.";
+                      };
+
+                      icon = mkOption {
+                        type = types.str;
+                        example = "[I]";
+                        description = "The option icon.";
+                      };
+
+                      desc = helpers.mkNullOrStr ''
+                        The description for this option.
+                      '';
+                    };
+                  })
+                  "{}"
+                  "The options for this engine.";
+              };
+            })
+        )
+        ''
+          Definition of the ${type} engines.
+
+          default: see [here](https://github.com/nvim-pack/nvim-spectre/blob/master/lua/spectre/config.lua)
+        '';
+    in {
+      color_devicons = helpers.defaultNullOpts.mkBool true ''
+        Whether to enable color devicons.
+      '';
+
+      open_cmd = helpers.defaultNullOpts.mkStr "vnew" ''
+        The open command.
+      '';
+
+      live_update = helpers.defaultNullOpts.mkBool false ''
+        Auto execute search again when you write to any file in vim.
+      '';
+
+      lnum_for_results = helpers.defaultNullOpts.mkBool false ''
+        Show line number for search/replace results.
+      '';
+
+      line_sep_start =
+        helpers.defaultNullOpts.mkStr
+        "┌──────────────────────────────────────────────────────"
+        "Start of the line separator";
+
+      result_padding = helpers.defaultNullOpts.mkStr "│  " ''
+        Result padding string.
+      '';
+
+      line_sep =
+        helpers.defaultNullOpts.mkStr
+        "└──────────────────────────────────────────────────────"
+        "Line separator.";
+
+      highlight =
+        helpers.defaultNullOpts.mkAttrsOf types.str
+        ''
+          {
+            headers = "SpectreHeader";
+            ui = "SpectreBody";
+            filename = "SpectreFile";
+            filedirectory = "SpectreDir";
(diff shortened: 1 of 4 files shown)


plugins/utils/spectre.nix
```diff
@@ -0,0 +1,236 @@
...
+          default: see [here](https://github.com/nvim-pack/nvim-spectre/blob/master/lua/spectre/config.lua)
+        '';
+
+      find_engine = mkEngineOption "find";
+
+      replace_engine = mkEngineOption "replace";
+
+      default = {
+        find = {
+          cmd = helpers.defaultNullOpts.mkStr "rg" ''
```
GaetanLepage, line 152: @traxys do you think that we should automatically install the `find` and `replace` engine ?
```nix
extraPackages = (optional (elem cfg.settings.default.find.cmd ["rg" "ag" ...]) pkgs.${cfg.settings.default.find.cmd}) ++ (optional (elem cfg.settings.default.replace.cmd ["rg" "ag" ...]) pkgs.${cfg.settings.default.find.cmd})
```

traxys: It would be nicer to do it

MattSturgeon: Is is possible to have the `cmd` option type accept either string or package, and use some `apply` magic to set `cmd` to `lib.getExe pkg`?

If we do an auto-install there should probably be a documented opt-in/out somewhere.

GaetanLepage: > Is is possible to have the `cmd` option type accept either string or package, and use some `apply` magic to set `cmd` to `lib.getExe pkg`?

According to me, this would be against the concept of RFC-42 `settings` and pretty unusual (at least from Nixvim). But why not ?


> If we do an auto-install there should probably be a documented opt-in/out somewhere.
Yes, actually I wanted to go without any installation at all here. @traxys what did you have in mind exactly ?

khaneliman:
Kept meaning to come back to setting up this module but forgot... thanks for doing this, I love this plugin.

plugins/utils/spectre.nix
GaetanLepage: I feel like there is no good option here, that satisfies:
- automatically detect the values of `settings.default.find.cmd`
- Allow the user to prevent the auto-install or change the package
- Conform with RFC 42

Hence, I would advocate to ship this plugin without any installation.
If the user wants to have `rg` installed, he can add it to `extraPackages`.
Also, we could simply add a warning in the description.

plugins/utils/spectre.nix
```diff
@@ -0,0 +1,282 @@
...
+      userCommandSettings = config.plugins.spectre.settings.default;
+    in {
+      findPackage = mkOption {
+        type = with types; nullOr package;
+        default =
+          {
+            rg = pkgs.ripgrep;
+          }
+          # `toString` will turn `null` into `"null"` to allow for the attrs indexation.
+          .${toString userCommandSettings.find.cmd}
```
MattSturgeon, line 247: I wonder if it'd be more readable to have `{rg = pkgs.ripgrep;}` and `{sed = pkgs.gnused; inherit (pkgs) sd;}` defined in a let-block elsewhere. Perhaps alongside `userCommandSettings`?

MattSturgeon (commented):
Looks good. One minor comment

plugins/utils/spectre.nix
GaetanLepage: Is it fine for you now ? I agree that the readability is indeed better.

plugins/utils/spectre.nix
```diff
@@ -0,0 +1,282 @@
...
+    defaultPackage = pkgs.vimPlugins.nvim-spectre;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    description = ''
+      You may want to add your find/replace tool(s) to the `extraPackages` list:
+
+      ```nix
+        extraPackages = [pkgs.rg];
+      ```
```
MattSturgeon, line 21:
```suggestion
      You may want to add your find/replace tool(s) to the `extraPackages` list:

      ```nix
        extraPackages = [pkgs.rg];
      ```

      or:

      ```nix
        plugins.spectre.findPackage = pkgs.rg;
        plugins.spectre.replacePackage = pkgs.gnused;
      ```
```

This description could be updated now that `{find,replace}Package` exist. Maybe even go further than my above suggestion and remove the old `extraPackages` stuff entirely?

plugins/utils/spectre.nix
```diff
@@ -0,0 +1,282 @@
...
+      # `toString` will turn `null` into `"null"` to allow for the attrs indexation.
+      findDefaultPackage = findPackages.${toString userCommandSettings.find.cmd} or null;
+      replaceDefaultPackage = replacePackages.${toString userCommandSettings.replace.cmd} or null;
+    in {
+      findPackage = mkOption {
+        type = with types; nullOr package;
+        default = findDefaultPackage;
+        description = ''
+          Which package to install for the find command.
+          Defaults to `pkgs.$\{settings.default.find.cmd}`.
```
MattSturgeon, line 258: This is kinda a white-lie, but I don't have any better suggestion so I'm happy to merge as-is.

tests/test-sources/plugins/utils/spectre.nix
```diff
@@ -0,0 +1,272 @@
...
+            options = ["word" "hidden"];
+          };
+          replace = {
+            cmd = "sed";
+          };
+        };
+      };
+    };
+  };
+
```
MattSturgeon, line 62:
```suggestion

  no-packages = {
    plugins.spectre = {
      inherit enable;
      
      findPackage = null;
      replacePackage = null;

      settings = {
        default = {
          find.cmd = "rg";
          replace.cmd = "sed";
        };
      };
    };
  };
```

Does it make sense to have a test with non-default values for `{find,replace}Package`?

plugins/utils/spectre.nix
GaetanLepage: You are definitely right. done

MattSturgeon (approved):

tests/test-sources/plugins/utils/spectre.nix
GaetanLepage: In this case, yes why not ! done

plugins/utils/spectre.nix
GaetanLepage: Yes, I agree. I haven't had a better idea either...

PR #1452, opened 2024-04-23
plugins/languages/texpresso: init

NickHu:
There are no configuration options for this package.
This just makes the `texpresso` binary (necessary to use this plugin) available to neovim.

### DIFF
--- a/plugins/languages/texpresso.nix
+++ b/plugins/languages/texpresso.nix
@@ -1,0 +1,33 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+# This plugin has no configuration, so we use `mkVimPlugin` without the `globalPrefix` argument to
+# avoid the creation of the `settings` option.
+  helpers.vim-plugin.mkVimPlugin config {
+    name = "texpresso";
+    originalName = "texpresso.vim";
+    defaultPackage = pkgs.vimPlugins.texpresso-vim;
+
+    maintainers = [maintainers.nickhu];
+
+    extraOptions = {
+      texpressoPackage = mkOption {
+        type = with types; nullOr package;
+        default = pkgs.texpresso;
+        example = null;
+        description = ''
+          The `texpresso` package to use.
+          Set to `null` to not install any package.
+        '';
+      };
+    };
+
+    extraConfig = cfg: {
+      extraPackages = [cfg.texpressoPackage];
+    };
+  }
--- a/tests/test-sources/plugins/languages/texpresso.nix
+++ b/tests/test-sources/plugins/languages/texpresso.nix
@@ -1,0 +1,6 @@
+{pkgs, ...}: {
+  empty = {
+    # texpresso is broken on darwin
+    plugins.texpresso.enable = !pkgs.stdenv.isDarwin;
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -70,6 +70,7 @@
     ./languages/rustaceanvim.nix
     ./languages/sniprun.nix
     ./languages/tagbar.nix
+    ./languages/texpresso.nix
     ./languages/treesitter/hmts.nix
     ./languages/treesitter/rainbow-delimiters.nix
     ./languages/treesitter/treesitter-context.nix


plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,16 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.vim-plugin.mkVimPlugin config {
```
MattSturgeon, line 8: Would be nice to have a comment stating that the plugin has no options to configure

MattSturgeon (approved):
lgtm

plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,16 @@
...
+  name = "texpresso";
+  originalName = "texpresso.vim";
+  defaultPackage = pkgs.vimPlugins.texpresso-vim;
+
+  maintainers = with lib.maintainers; [nickhu];
+
+  extraPackages = with pkgs; [texpresso];
```
GaetanLepage, line 15:
```suggestion
  extraOptions = {
    texpressoPackage = mkOption {
        type = with types; nullOr package;
        default = pkgs.texpresso;
        description = ''
          The fzf package to use.
          Set to `null` to not install any package.
        '';
      };
  };
  
  extraPackages = [cfg.texpressoPackage];
```

@MattSturgeon @traxys should we do this systematically or here it is fine to just auto-install `texpresso` ? I have no strong opinion.
It would be nice to agree on a general rule: either
A) Systematically add a `fooPackage` option that can be set to `null`
OR
B) If it makes some sense for a user to select a custom package for it, add the option, otherwise, silent auto-install is OK (more subjective)

On the one hand (A) seems more straightforward to enforce and requires less discussion. It also maximizes the user's control over its setup.
On the other hand, adding `pkgs.git` to `extraPackages` for `plugins.gitsigns` doesn't shock me either.

MattSturgeon: I prefer option "A"

GaetanLepage: Sure, but a user could want nixvim not to install texpresso automatically by setting this option to `null`.
For instance, you could have another version in your `$PATH`.
Also, you could override the `texpresso` derivation right here.
Anyway, my point is that it does no harm having the option.

GaetanLepage: > I would like in the future to have _all_ packages installed by nixvim be nullable to support externally managed programs

So be it !
@NickHu please implement the proposed snippet to add `texpressoPackage`.

MattSturgeon: > and there is no alternative package that would work

A user may wish to use a custom package (e.g. patched), a package from another channel (e.g. stable vs unstable) or some third party flake.

In addition to the null option discussed above.

plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,27 @@
...
+  defaultPackage = pkgs.vimPlugins.texpresso-vim;
+
+  maintainers = with lib.maintainers; [nickhu];
+
+  extraOptions = {
+    texpressoPackage = mkOption {
+        type = with types; nullOr package;
+        default = pkgs.texpresso;
+        description = ''
+          The fzf package to use.
```
GaetanLepage, line 20:
```suggestion
          The texpresso package to use.
```

plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,31 @@
...
+        example = null;
+        description = ''
+          The `texpresso` package to use.
+          Set to `null` to not install any package.
+        '';
+      };
+    };
+
+    extraConfig = cfg: {
+      extraPackages = optional (cfg.texpressoPackage != null) cfg.texpressoPackage;
```
GaetanLepage, line 29:
```suggestion
      extraPackages = [cfg.texpressoPackage];
```

Since [this commit](https://github.com/nix-community/nixvim/commit/f2f97d844bb39559f3356e209b49c92900d860b8), you can just dump `null` values to `extraPackages` and they will get properly ignored.

plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,31 @@
...
+  pkgs,
+  ...
+}:
+with lib;
+  helpers.vim-plugin.mkVimPlugin config {
+    name = "texpresso";
+    originalName = "texpresso.vim";
+    defaultPackage = pkgs.vimPlugins.texpresso-vim;
+
+    maintainers = with maintainers; [nickhu];
```
GaetanLepage, line 14:
```suggestion
    maintainers = [maintainers.nickhu];
```

tests/test-sources/plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.texpresso.enable = true;
+  };
```
MattSturgeon, line 4: To be exhaustive, there ought to be a non-default `texpressoPackage` test too. I don't think that's necessary though.

plugins/languages/texpresso.nix
```diff
@@ -0,0 +1,31 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+  helpers.vim-plugin.mkVimPlugin config {
```
MattSturgeon, line 9:
```suggestion
  # This plugin has no configuration, so we use `mkVimPlugin` without the
  # `globalPrefix` argument to avoid the creation of the `settings` option.
  helpers.vim-plugin.mkVimPlugin config {
```

MattSturgeon (commented):
Minor nits, otherwise looks good.

tests/test-sources/plugins/languages/texpresso.nix
GaetanLepage: Yes, this is not critical.

plugins/languages/texpresso.nix
GaetanLepage: Done

GaetanLepage (approved):
LGTM now, thanks @NickHu !

PR #1453, opened 2024-04-23
misc: allow null in extraPackages

GaetanLepage:
=> This is a proposal and expects discussion.

## Current situation
Currently, the `extraPackages` option has type `listOf package`.
Furthermore, we often rely on _optional_ package options (i.e. typed as `nullOr package`) where `null` means: "do not install any package".

Then, the implementation looks like this:
```nix
extraPackages =
  optional
  (cfg.fooPackage == null)
  cfg.fooPackage;
```

## Proposition:
Have `extraOptions` to be typed `listOf (nullOr package)`. I.e., make it accept `null`.

### Pros:
- Further reduce boilerplate. Simplify the snippet above to
```nix
extraPackages = [cfg.fooPackage];
```

### Cons:
- Reduction in clarity. -> More prone to errors/omission.

What is your opinion about this ?

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -1,7 +1,6 @@
 {
   lib,
   config,
-  pkgs,
   ...
 }:
 with lib; let
@@ -34,9 +33,10 @@ in {
     };
 
     extraPackages = mkOption {
-      type = types.listOf types.package;
+      type = with types; listOf (nullOr package);
       default = [];
       description = "Extra packages to be made available to neovim";
+      apply = builtins.filter (p: p != null);
     };
 
     extraPython3Packages = mkOption {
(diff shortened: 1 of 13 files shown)


traxys (approved):
I think it simplifies a lot the implementation, without any user facing footguns!

GaetanLepage:
> I'm of two minds; this definitely simplifies a common use case, however it does make the option type unnecessarily loose.
> 
> We could achieve _almost_ the same thing with a helper function:
> 
> ```nix
> nullable = a: lib.optional (a != null) a;
> ```
> 
> This would change your example to:
> 
> ```nix
> extraPackages = helpers.nullable cfg.fooPackage;
> ```
> 
> That said, I've not got a strong opinion either way.

Indeed, this is on okay idea. I would go on with this as adding more helpers is not too satisfying either.
Let's go with this change. As it is internal, we can easily revert it later if we discover new drawbacks.

Thanks for your feedback.

PR #1459, opened 2024-04-24
plugins/cloak: init

GaetanLepage:
Add support for [cloak.nvim](https://github.com/laytan/cloak.nvim). Cloak allows you to overlay *'s over defined patterns in defined files.


MattSturgeon (approved):
Can't see any issues

PR #1464, opened 2024-04-26
plugins/lsp: add ast-grep language server

### LINKED ISSUE #1463: [PLUGIN REQUEST] `ast-grep lsp`
| Field    | Description |
|----------|-------------|
| Plugin   | `ast-grep`  | 
| Homepage | [`nvim-lspconfig/ast_grep`](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ast_grep)     | 
| Nixpkgs  | [`nixpkgs/ast-grep`](https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/by-name/as/ast-grep/package.nix#L46) | 

## Extra Information
It is already available in nvim-lspconfig and nixpkgs so it should be easy to add.

GaetanLepage:
Add support for the [ast-grep](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ast_grep) language server.

Fixes #1463

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -16,6 +16,14 @@ with lib; let
       cmd = cfg: ["${cfg.package}/bin/ansible-language-server" "--stdio"];
     }
     {
+      name = "ast-grep";
+      description = ''
+        ast-grep(sg) is a fast and polyglot tool for code structural search, lint, rewriting at large scale.
+        ast-grep LSP only works in projects that have `sgconfig.y[a]ml` in their root directories.
+      '';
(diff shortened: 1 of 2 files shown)


uttarayan21:
I get the following error message when trying to use it
```
Failed to setup: ...mPackages/start/nvim-lspconfig/lua/lspconfig/configs.lua:71: attempt to index local 'user_config' (a nil value)
```
But the plugin does work

MattSturgeon (approved):
Diff looks good. I've tested the LSP is listed in `:LspInfo`, but I'm not familiar enough with this LSP to try actually loading/using it.

PR #1467, opened 2024-04-26
plugins/lsp/ccls: fix initOptions

khaneliman:
ccls wasn't passing the init options to it's own configuration so they were being ignored..


GaetanLepage (approved):
Nice catch, thanks !

khaneliman:
> Nice catch, thanks !

I was trying to look into an issue in a project and was confused why my compile_commands.json wasn't being loaded lol

PR #1469, opened 2024-04-27
ci: introduce `nixf-tidy-action`

inclyc:
nixf-tidy-action is a nix linter based on libnixf. It detects deadcode, unused `rec`, and syntax errors. 

nixf-tidy-action will only report warnings on changed files in diff (thus it will not be very pedantic).

nixf-sample

### DIFF
--- a/.github/workflows/nixf-tidy-action.yaml
+++ b/.github/workflows/nixf-tidy-action.yaml
@@ -1,0 +1,25 @@
+name: nixf-tidy code linter
+
+on:
+  [ pull_request ]
+
+jobs:
+  nixf-tidy:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+
+    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+      with:
+        github_access_token: ${{ secrets.GITHUB_TOKEN }}
+        extra_nix_config: |
+          trusted-public-keys = nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+          substituters = https://nix-community.cachix.org https://cache.nixos.org/
+    - run: nix profile install github:nix-community/nixd#nixd
+    - uses: inclyc/nixf-tidy-action@v1


.github/workflows/nixf-tidy-action.yaml
```diff
@@ -0,0 +1,25 @@
+name: nixf-tidy code linter
+
+on:
+  [ pull_request ]
```
MattSturgeon, line 4: Prefer YAML syntax to JSON-style

Also, I'm not sure [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) is the right event here. That runs whenever a PR is opened, edited, reviewed, etc (see "activity types").

Instead, we'd want this to run whenever a branch is updated ([push](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push)).

We _could_ use `branches-ignore` to prevent running on `main` and `nixos-*`, but IMO I don't see an issue with having the CI on these branches too.

Note if the action expects `github.event.pull_request.head.sha`, this'll be missing when using `on.push`. Instead `github.sha` or `GITHUB_SHA` should be used for non-pr events.

```suggestion
  push
```

.github/workflows/nixf-tidy-action.yaml
```diff
@@ -0,0 +1,25 @@
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+      with:
+        github_access_token: ${{ secrets.GITHUB_TOKEN }}
+        extra_nix_config: |
+          trusted-public-keys = nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+          substituters = https://nix-community.cachix.org https://cache.nixos.org/
+    - run: nix profile install github:nix-community/nixd#nixd
```
MattSturgeon, line 24: IMO this should be the responsibility of the `nixf-tidy-action` action.

Is there some technical reason why _we_ need to install the action's dependency in _our_ workflow?

.github/workflows/nixf-tidy-action.yaml
```diff
@@ -0,0 +1,25 @@
...
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+
+    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
```
MattSturgeon, line 18:
```suggestion
    - uses: actions/checkout@v4
    - uses: cachix/install-nix-action@v26
```

MattSturgeon (changes requested):
Thanks, this looks interesting!

.github/workflows/nixf-tidy-action.yaml
inclyc: > IMO this should be the responsibility of the `nixf-tidy-action` action.
> 
> Is there some technical reason why _we_ need to install the action's dependency in _our_ workflow?

Thanks, I just wrote `nixf-tidy-action` action in purely typescript. Thus I think it is more elegant to handle dependencies in nix?

(Otherwise these installation should be performed in "js" exec, looks bad!)

.github/workflows/nixf-tidy-action.yaml
inclyc: > Also, I'm not sure [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) is the right event here. That runs whenever a PR is opened, edited, reviewed, etc (see "activity types"). 

I haven't test it exactly for events other than "pull_request", maybe I should update the action itself then.

GaetanLepage:
I am personally not against trying this. However, I would much prefer having a pre-commit hook based solution (using flake-parts).
Indeed, this could be dealt with within the development environment and also, naturally checked by the CI.

.github/workflows/nixf-tidy-action.yaml
MattSturgeon: > Also, I'm not sure [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) is the right event here. That runs whenever a PR is opened, edited, reviewed, etc (see "activity types").

I think some of my assumptions were off here. The `pull_request` event does get triggered by a bunch of stuff, but pushes to a PR branch don't _always_ seem to trigger the `push` event...

saygo-png:
Superseded by https://github.com/nix-community/nixvim/pull/3735

PR #1472, opened 2024-04-29
plugins/lsp: add nginx-language-server

### LINKED ISSUE #1435: Nginx LSP does not exist
I trying to setup nginx_language_server but i get the following error:

       `error: The option 'programs.nixvim.plugins.lsp.servers.nginx_language_server' does not exist. Definition values:
       - In '/nix/store/c48l8qfaiigv7f7f1inwdyqdbpkk3cky-source/modules/nvim/plugins/lsp.nix':
           {
             enable = true;
           }`

If the LSP is not supported by default, how can I workaround to enable it?

GaetanLepage:
Add support for the [nginx-language-server](https://github.com/pappasam/nginx-language-server).

Fixes #1435

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -417,6 +417,11 @@ with lib; let
       description = "metals for Scala";
     }
     {
+      name = "nginx-language-server";
+      description = "nginx-language-server for `nginx.conf`";
+      serverName = "nginx_language_server";
+    }
+    {
       name = "nil_ls";
       description = "nil for Nix";
       package = pkgs.nil;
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -416,6 +416,11 @@ with lib; let
       name = "metals";
       description = "metals for Scala";
     }
+    {
+      name = "nginx-language-server";
```
MattSturgeon, line 420: `nginx-ls` (or simply `nginx`) would feel more consistent with (most) other LSP attr names, although there's already a handful that use the "full" `***-language-server` style...

Maybe we should consider providing less verbose and/or more consistent aliases? Not in scope here, of course.

MattSturgeon (approved):

plugins/lsp/language-servers/default.nix
GaetanLepage: Actually, for the name, we use the ones that [nvim-lspconfigs](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.txt) gives. We only substitute `_` for `-` (`foo_bar` -> `foo-bar`).

PR #1473, opened 2024-04-29
plugins/hydra: init

### LINKED ISSUE #1305: [PLUGIN REQUEST] hydra.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `hydra.nvim`  | 
| Homepage | `https://github.com/nvimtools/hydra.nvim`     | 
| Nixpkgs  | `vimPlugins.hydra-nvim` |

GaetanLepage:
Add support for [hydra.nvim](https://github.com/nvimtools/hydra.nvim).

Fixes #1305

### DIFF
--- a/plugins/utils/hydra/default.nix
+++ b/plugins/utils/hydra/default.nix
@@ -1,0 +1,52 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+  helpers.neovim-plugin.mkNeovimPlugin config {
+    name = "hydra";
+    originalName = "hydra.nvim";
+    defaultPackage = pkgs.vimPlugins.hydra-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    extraOptions = {
+      # A list of `Hydra` definitions
+      hydras = import ./hydras-option.nix {inherit lib helpers;};
+    };
+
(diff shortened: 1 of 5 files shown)


plugins/utils/hydra/default.nix
```diff
@@ -0,0 +1,52 @@
...
+  helpers.neovim-plugin.mkNeovimPlugin config {
+    name = "hydra";
+    originalName = "hydra.nvim";
+    defaultPackage = pkgs.vimPlugins.hydra-nvim;
+
+    maintainers = [maintainers.GaetanLepage];
+
+    extraOptions = {
+      # A list of `Hydra` definitions
+      hydras = import ./hydras-option.nix {inherit lib helpers;};
```
MattSturgeon, line 18: Is there a reason you tend to favor inheriting specific attrs over doing something like this:

```nix
{
  lib,
  helpers,
  config,
  pkgs,
  ...
} @attrs: {
  hydras = import ./hydras-option.nix attrs;
}
```

?

This is more a general question, not feedback on this PR.

MattSturgeon (approved):
Diff looks good

plugins/utils/hydra/default.nix
GaetanLepage: I find this way to be more explicit and less obfuscating, but it does not matter to much to me.

PR #1476, opened 2024-04-29
tests/cmp: add a test that enables all sources

GaetanLepage:
Add an exhaustive test case for the cmp sources.


plugins/completion/cmp/cmp-helpers.nix
```diff
@@ -20,6 +20,8 @@ with lib; rec {
       maintainers = [maintainers.GaetanLepage];
     };
 
+  # WARNING: When adding a source, make sure to append the corresponding test in
+  # tests/test-sources/plugins/completion/cmp-all-sources.nix
   pluginAndSourceNames = {
```
traxys, line 25: We could maybe put this in a separate file, and `import` it here & in the tests, so that we can automatically enable all sources

traxys (changes requested):

tests/test-sources/plugins/completion/cmp-all-sources.nix
```diff
@@ -0,0 +1,66 @@
...
+        enable = true;
+
+        panel.enabled = false;
+        suggestion.enabled = false;
+      };
+
+      cmp = {
+        enable = true;
+        settings.sources =
+          [
```
traxys, line 14: You should be able to do something like:
```nix
settings.sources = let
    knownSources = import ../../../cmp/sources.nix;
    filteredSources =
            pkgs.lib.optional
            (pkgs.stdenv.hostPlatform.system != "aarch64-linux")
            {name = "cmp_tabnine";}
in 
    builtins.filter (src: !(builtins.elem src filteredSources)) (mapAttrsToList (k: _: {name = k;}) knownSources)
```

To avoid needing to list each source

traxys (changes requested):

tests/test-sources/plugins/completion/cmp-all-sources.nix
GaetanLepage: Done.

traxys (approved):

PR #1368, opened 2024-04-03
plugins/ccc: Add more settings

janurskremer:
I tried to add more settings to the plugin with suggestions by @GaetanLepage. 

`checks` failed with this error: 

```
error: builder for '/nix/store/jm8y8klh0zvjbn4fkkspj840mkda5kv7-plugins-utils-ccc-example.drv' failed with exit code 1;
       last 8 log lines:
       > Running phase: patchPhase
       > Running phase: configurePhase
       > no configure script, doing nothing
       > Running phase: buildPhase
       > ERROR: Error detected while processing /nix/store/07i48i1sqr60dir3prki4p2cax0aai0s-init.lua:
       > E5113: Error while calling lua chunk: /nix/store/07i48i1sqr60dir3prki4p2cax0aai0s-init.lua:30: attempt to index global 'ccc' (a nil value)
       > stack traceback:
       >  /nix/store/07i48i1sqr60dir3prki4p2cax0aai0s-init.lua:30: in main chunk
       For full logs, run 'nix log /nix/store/jm8y8klh0zvjbn4fkkspj840mkda5kv7-plugins-utils-ccc-example.drv'.
error: 1 dependencies of derivation '/nix/store/9iayb3kxg7lm0vs05agq9pn4b6zzb4pi-nixvim-tests.drv' failed to build
error: build of '/nix/store/9aa86a39l8vz3h974f2dcsnyd6jfwfjv-darwin-system-24.05.20240329.d8fe5e6+darwin4.36524ad.drv', '/nix/store/9iayb3kxg7lm0vs05agq9pn4b6zzb4pi-nixvim-tests.drv', '/nix/store/c2vg4dkn92yzpc8y5kfyszwri1fy7lf4-special-arg-test.drv', '/nix/store/gl8kcfjz2yjjvbfrfngk0817fkh7rzv6-enable-except-in-tests.drv', '/nix/store/gnm94gsiswc13gvi0bjxfkai88zzkdk8-no-flakes.drv', '/nix/store/jsmf0hf1a7zmbd53g7h89c1lh6bsbq6w-extend-test.drv', '/nix/store/ms8j2g1dnc7x0f1vr22by7fd3s2qql0w-home-manager-generation.drv', '/nix/store/sgbs5pn0c1vr5m9z18rygh72ld468rq9-nixvim-check.drv' failed
```

Please help me figure out what I'm missing.

PR #1482, opened 2024-05-02
plugins/ccc: Add more settings

GaetanLepage:
Follow up of #1368 which I closed by accident.

### DIFF
--- a/plugins/utils/ccc.nix
+++ b/plugins/utils/ccc.nix
@@ -5,6 +5,7 @@
   pkgs,
   ...
 }:
+with lib;
 helpers.neovim-plugin.mkNeovimPlugin config {
   name = "ccc";
   originalName = "ccc.nvim";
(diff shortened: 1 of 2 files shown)


plugins/utils/ccc.nix
```diff
@@ -5,50 +5,218 @@
...
+        The default color used when a color cannot be picked. It must be HEX format.
       '';
 
+      inputs = mkOption {
+        type = listOfRawLua;
+        default = [
+          "ccc.input.rgb"
+          "ccc.input.hsl"
+          "ccc.input.cmyk"
+        ];
```
traxys, line 33: Why are we setting an explicit default? If this is important we should add a comment why

plugins/utils/ccc.nix
```diff
@@ -5,50 +5,218 @@
...
+        default = [
+          "ccc.picker.hex"
+          "ccc.picker.css_rgb"
+          "ccc.picker.css_hsl"
+          "ccc.picker.css_hwb"
+          "ccc.picker.css_lab"
+          "ccc.picker.css_lch"
+          "ccc.picker.css_oklab"
+          "ccc.picker.css_oklch"
+        ];
```
traxys, line 100: Same

plugins/utils/ccc.nix
```diff
@@ -5,50 +5,218 @@
...
+      };
+
+      outputs = mkOption {
+        type = listOfRawLua;
+        default = [
+          "ccc.output.hex"
+          "ccc.output.hex_short"
+          "ccc.output.css_rgb"
+          "ccc.output.css_hsl"
+        ];
```
traxys, line 65: Same

plugins/utils/ccc.nix
```diff
@@ -5,50 +5,218 @@
...
+        '';
+      };
+
+      convert = mkOption {
+        type = listOfRawLua;
+        # TODO
+        default = [
+          "{ ccc.picker.hex, ccc.output.css_rgb }"
+          "{ ccc.picker.css_rgb, ccc.output.css_hsl }"
+          "{ ccc.picker.css_hsl, ccc.output.hex }"
```
traxys, line 172: I think this option should be a list of raw lua pairs, or at least a list of list of raw lua

traxys: It could also be an attr set, as this represents nicely pairings (where the keys & values would be raw lua)

traxys (changes requested):

plugins/utils/ccc.nix
MattSturgeon: [Looking upstream](https://github.com/uga-rosa/ccc.nvim/blob/f388f1981d222967c741fe9927edf9ba5fa3bcbe/doc/ccc.txt#L180-L203), that's the default used there.

My understanding is that we should be using a [default-null option](https://github.com/nix-community/nixvim/blob/c05fba47c733445098847399b61ddfd81ecbde55/lib/options.nix#L60-L72), which document upstream's default, as used when _our_ option is null. I.e. the `default` passed to `lib.mkOption` should be `null`.

traxys: Yes exactly

plugins/utils/ccc.nix
GaetanLepage: The latter would not be RFC-42 style compliant, right ?
I went with the `listOf (listOf strLua)` approach.

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
-        '';
-
-    lsp = helpers.defaultNullOpts.mkBool true ''
-      Whether to enable LSP support. The color information is updated in the
-      background and the result is used by `:CccPick` and highlighter.
-    '';
+      inputs = mkOption {
+        type = listOfRawLua;
+        apply = mapToRawLua;
+        default = [ ];
```
MattSturgeon, line 29: If this option is left as the default empty list, will an empty lua table be passed to `setup`? If so, will that override the plugins default `inputs`?

We would normally have the default be `null`, which (I assume) leads to the option being omitted from the lua table.

Also, if passing a value (that doesn't explicitly include the defaults) _unsets_ said defaults, we should probably make a note of that in the docs.

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+              "ccc.input.cmyk"
+            ]
+          ```
+        '';
+      };
+
+      outputs = mkOption {
+        type = listOfRawLua;
+        apply = mapToRawLua;
+        default = [ ];
```
MattSturgeon, line 65: Same question as `inputs`

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+              "ccc.output.css_hsl"
+            ]
+          ```
+        '';
+      };
+
+      pickers = mkOption {
+        type = listOfRawLua;
+        apply = mapToRawLua;
+        default = [ ];
```
MattSturgeon, line 101: Same question as `inputs`

MattSturgeon (commented):
Looking good, just need to clarify whether some stuff should default  to `null` or if empty list is ok

plugins/utils/ccc.nix
MattSturgeon: Is there a reason this can't use `helpers.defaultNullOpts.mkListOf`?

GaetanLepage: > If this option is left as the default empty list, will an empty lua table be passed to `setup`? If so, will that override the plugins default `inputs`?
> 
> We would normally have the default be `null`, which (I assume) leads to the option being omitted from the lua table.
> 
> Also, if passing a value (that doesn't explicitly include the defaults) _unsets_ said defaults, we should probably make a note of that in the docs.

Your concern is perfectly valid. But here, it is OK:
- Empty lists/attrs are indeed ignored by `toLuaObject`, so it has the same behavior as a `null` default.
- Now, why not `null` ? Well, just to simplify the `apply` function and not having to check if the value is `null` or a list.

GaetanLepage: With `mkListOf`, you cannot have an option where the user fills a list with lua strings that will get converted to raw lua automatically.

MattSturgeon: Thanks for explaining 👍

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+          List of color system to be activated.
+          `ccc-action-toggle_input_mode` toggles in this order.
+          The first one is the default used at the first startup.
+          Once activated, it will keep the previous input mode.
 
-    highlighter = {
-      auto_enable = helpers.defaultNullOpts.mkBool false ''
-        Whether to enable automatically on `BufEnter`.
-      '';
+          default:
```
MattSturgeon, line 51: This should updated to match #1530

```suggestion
          Plugin default:
```

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+          "ccc.output.hex.setup({ uppercase = true })"
+          "ccc.output.hex_short.setup({ uppercase = true })"
+        ];
+        description = ''
+          List of output format to be activated.
+          `ccc-action-toggle_ouotput_mode` toggles in this order.
+          The first one is the default used at the first startup.
+          Once activated, it will keep the previous output mode.
+
+          default:
```
MattSturgeon, line 86:
```suggestion
          Plugin default:
```

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+          "ccc.picker.css_lab"
+          "ccc.picker.css_lch"
+          "ccc.picker.css_oklab"
+          "ccc.picker.css_oklch"
+          "ccc.picker.css_name"
+        ];
+        description = ''
+          List of formats that can be detected by `:CccPick` to be activated.
+
+          default:
```
MattSturgeon, line 116:
```suggestion
          Plugin default:
```

plugins/utils/ccc.nix
```diff
@@ -5,59 +5,252 @@
...
+            "ccc.picker.css_oklch"
+            "ccc.output.hex"
+          ]
+        ];
+        description = ''
+          Specify the correspondence between picker and output.
+          The default setting converts the color to `css_rgb` if it is in hex format, to `css_hsl`
+          if it is in `css_rgb` format, and to hex if it is in `css_hsl` format.
+
+          default:
```
MattSturgeon, line 209:
```suggestion
          Plugin default:
```

MattSturgeon (changes requested):
After a quick scan, everything seems resolved. Just need to update some manually crafted descriptions to match #1530.

MattSturgeon (approved):

GaetanLepage:
@traxys we need your approval to merge ;)

traxys (approved):

PR #1490, opened 2024-05-05
plugins/nixd: Adapt to new options

traxys:


tests/test-sources/plugins/lsp/language-servers/nixd.nix
```diff
@@ -7,30 +7,22 @@
...
-                "foo"
-                "bar"
-              ];
-              installable = "";
-            };
-            depth = 0;
-            workers = 3;
-          };
+          nixpkgs.expr = ''
+            import (builtins.getFlake "/home/lyc/workspace/CS/OS/NixOS/flakes").inputs.nixpkgs { }
```
MattSturgeon, line 11: Where does the referenced flake come from? Seems strange to include this in a test.

tests/test-sources/plugins/lsp/language-servers/nixd.nix
```diff
@@ -7,30 +7,22 @@
...
-            target = {
-              args = [
-                "yes"
-                "no"
-                "maybe"
-              ];
-              installable = "";
-            };
+            nixos.expr = ''
+              (builtins.getFlake "/home/lyc/flakes").nixosConfigurations.adrastea.options
```
MattSturgeon, line 20: Same

tests/test-sources/plugins/lsp/language-servers/nixd.nix
```diff
@@ -7,30 +7,22 @@
...
+            '';
+
+            home-manager.expr = ''
+              (builtins.getFlake "/home/lyc/flakes").homeConfigurations."lyc@adrastea".options
```
MattSturgeon, line 24: Same

plugins/lsp/language-servers/nixd.nix
```diff
@@ -10,52 +10,52 @@ let
...
+      in
+      helpers.mkNullOrOption (helpers.nixvimTypes.maybeRaw provider) ''
+        This expression will be interpreted as "nixpkgs" toplevel
+        Nixd provides package, lib completion/information from it.
+      '';
   };
 
   config = mkIf cfg.enable {
-    extraPackages = optional (cfg.settings.formatting.command == "nixpkgs-fmt") pkgs.nixpkgs-fmt;
+    extraPackages = optional (cfg.settings.formatting.command == [ "nixpkgs-fmt" ]) pkgs.nixpkgs-fmt;
```
MattSturgeon, line 59: Maybe not on scope for this PR, but it'd be nice to do this for other nix formatters too. Maybe reference an attr set mapping name to package, then do `extraPackages = set.${cfg.settings.formatting.command} or null;`, although that doesn't work for lists...

plugins/lsp/language-servers/default.nix
```diff
@@ -660,10 +663,11 @@ in
...
     ./ccls.nix
     ./efmls-configs.nix
-    ./nixd.nix
     ./pylsp.nix
     ./rust-analyzer.nix
     ./svelte.nix
     ./vls.nix
   ];
+
+  config = lib.mkMerge [ nixdSettings.config ];
```
MattSturgeon, line 672: Why use `mkMerge` here? Couldn't this just be `inherit (nixdSettings) config;`?

MattSturgeon (commented):
On my phone, so just my initial thoughts.

I'm also interested in whether there's anything that can be incorporated from #1502

plugins/lsp/language-servers/default.nix
traxys: It could, but this is in case we need the same for other serverq

tests/test-sources/plugins/lsp/language-servers/nixd.nix
traxys: Took the upstream documentation

traxys:
> On my phone, so just my initial thoughts.
> 
> I'm also interested in whether there's anything that can be incorporated from #1502

I don't think so

tests/test-sources/plugins/lsp/language-servers/nixd.nix
GaetanLepage: resolved then ?

traxys: @MattSturgeon what do you think?

MattSturgeon: I was reading it as a nix expression that'd actually get evaluated and was therefore very confused!

If it's just a string being assigned to an option then I guess it's fine.

traxys: It is simply a string yes! `nixd` would evaluate that expression on LspAttach I believe

plugins/lsp/language-servers/nixd.nix
traxys: Yeah definitively, I didn't want to clutter this PR with that

GaetanLepage (approved):
LGTM

PR #1496, opened 2024-05-07
expose withRuby option

### LINKED ISSUE #1495: [BUG] Ruby development broken
| Plugin       | -  | 
| Nixpkgs      | `unstable` | 
| Home Manager | - | 

## Description

Modern Ruby development defines dependencies in a project's committed `Gemfile` and `Gemfile.lock`. The dependencies are then installed by `bundler` into a default location in `$(HOME)/.local/share`. However, because nixvim uses `pkgs.neovimUtils.makeNeovimConfig`, and [nixpkgs' default for `withRuby` is true](https://github.com/NixOS/nixpkgs/blob/24.05-pre/nixos/modules/programs/neovim.nix#L51), the environment variable `GEM_HOME` is set for the neovim process, pointing somewhere into the Nix store at `/nix/store/...-neovim-ruby-env/lib/ruby/gems`. `GEM_HOME` is recognized by `bundler`, which will try to write into that path, which is naturally forbidden. So, commands like `bundle install` cannot be run from inside neovim, only from outside, which is inconvenient at best, but mostly confusing and concerning.

A pristine Ruby environment is really important for Ruby software development, and setting GEM_HOME is a big no-no in regards to that (I know that this doesn't come from nixvim). So, currently a user will have to decide between mutually exclusive use cases: 1) Use neovim's Ruby provider for neovim's purposes (`withRuby = true`) or 2) do Ruby software development in a pristine Ruby environment (`withRuby = false`).

I'll create a Pull request to allow configuration of `withRuby`, what has already been foreseen here: https://github.com/nix-community/nixvim/pull/1198#issuecomment-1975460528

```sh
mkdir /tmp/test
cd /tmp/test
nix-shell -p ruby
bundle init
echo "gem 'solargraph', '0.49.0'" >> Gemfile
bundle install # works
nvim +'! bundle install' # broken:  Bundler::GenericSystemCallError Errno::EROFS: Read-only file system @ dir_s_mkdir
```

michaelfranzl:
fixes #1495

### DIFF
--- a/wrappers/modules/output.nix
+++ b/wrappers/modules/output.nix
@@ -24,6 +24,12 @@ with lib;
       '';
     };
 
+    withRuby = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Enable Ruby provider.";
+    };
+
     withNodeJs = mkOption {
       type = types.bool;
       default = false;
@@ -88,6 +94,7 @@ with lib;
             extraPython3Packages
             viAlias
             vimAlias
+            withRuby
             withNodeJs
             ;
           # inherit customRC;


GaetanLepage (approved):
LGTM !

traxys (approved):

PR #1512, opened 2024-05-12
plugins/arrow: init

hmajid2301:
Add the [arrow-nvim](https://github.com/otavioschwanck/arrow.nvim) plugin.

We are currently ignoring the option `treesitter_context`, which is not  currently  usable, for detail see https://github.com/otavioschwanck/arrow.nvim/pull/43#issue-2236320268

### DIFF
--- a/plugins/utils/arrow.nix
+++ b/plugins/utils/arrow.nix
@@ -1,0 +1,220 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "arrow";
+  originalName = "arrow.nvim";
+  defaultPackage = pkgs.vimPlugins.arrow-nvim;
+
+  maintainers = [ maintainers.hmajid2301 ];
+
+  settingsOptions = {
+    show_icons = helpers.defaultNullOpts.mkBool false ''
+      If true will show icons.
+    '';
+
+    always_show_path = helpers.defaultNullOpts.mkBool false ''
+      If true will show path.
+    '';
+
+    separate_by_branch = helpers.defaultNullOpts.mkBool false ''
+      If true will split bookmarks by git branch.
+    '';
+
+    hide_handbook = helpers.defaultNullOpts.mkBool false ''
+      If true to hide the shortcuts on menu.
+    '';
+
+    save_path = helpers.defaultNullOpts.mkLuaFn ''
+      function()
+        return vim.fn.stdpath("cache") .. "/arrow"
+      end
+    '' "Function used to determine where to save arrow data.";
+
+    mapping = {
+      edit = helpers.defaultNullOpts.mkStr "e" ''
+        Mapping to edit bookmarks.
+      '';
+
+      delete_mode = helpers.defaultNullOpts.mkStr "d" ''
+        Mapping to go to delete mode, where you can remove bookmarks.
+      '';
+
+      clear_all_items = helpers.defaultNullOpts.mkStr "C" ''
+        Mapping to clear all bookmarks.
+      '';
+
+      toggle = helpers.defaultNullOpts.mkStr "s" ''
+        Mapping to save if `separate_save_and_remove` is true.
+      '';
+
+      open_vertical = helpers.defaultNullOpts.mkStr "v" ''
+        Mapping to open bookmarks in vertical split.
+      '';
+
+      open_horizontal = helpers.defaultNullOpts.mkStr "-" ''
+        Mapping to open bookmarks in horizontal split.
+      '';
+
+      quit = helpers.defaultNullOpts.mkStr "q" ''
+        Mapping to quit arrow.
+      '';
+
+      remove = helpers.defaultNullOpts.mkStr "x" ''
+        Mapping to remove bookmarks. Only used if `separate_save_and_remove` is true.
+      '';
+
+      next_item = helpers.defaultNullOpts.mkStr "]" ''
+        Mapping to go to next bookmark.
+      '';
+
+      prev_item = helpers.defaultNullOpts.mkStr "[" ''
+        Mapping to go to previous bookmark.
+      '';
+    };
+
+    custom_actions = {
+      open =
+        helpers.defaultNullOpts.mkLuaFn
+          ''
+            function(target_file_name, current_file_name) end
+          ''
+          ''
+            - `target_file_name`: file selected to be open
+            - `current_file_name`: filename from where this was called
+          '';
+
+      split_vertical = helpers.defaultNullOpts.mkLuaFn ''
+        function(target_file_name, current_file_name) end
+      '' "";
+
+      split_horizontal = helpers.defaultNullOpts.mkLuaFn ''
+        function(target_file_name, current_file_name) end
+      '' "";
+    };
+
+    window =
+      helpers.defaultNullOpts.mkAttrsOf types.anything
+        ''
+          {
+            relative = "editor";
+            width = "auto";
+            height = "auto";
+            row = "auto";
+            col = "auto";
+            style = "minimal";
+            border = "single";
+          }
+        ''
+        ''
+          Controls the appearance and position of an arrow window.
+          See `:h nvim_open_win()` for all options.
+        '';
+
+    per_buffer_config = {
+      lines = helpers.defaultNullOpts.mkInt 4 ''
+        Number of lines on preview.
+      '';
+
+      sort_automatically = helpers.defaultNullOpts.mkBool true ''
+        If true will sort buffer marks automatically.
+      '';
+
+      satellite = {
+        enable = helpers.defaultNullOpts.mkBool false ''
+          If true will display arrow index in scrollbar at every update.
+        '';
+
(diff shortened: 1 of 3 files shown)


plugins/utils/arrow.nix
```diff
@@ -0,0 +1,226 @@
...
+
+        priority = helpers.defaultNullOpts.mkInt 1000 '''';
+      };
+
+      zindex = helpers.defaultNullOpts.mkInt 50 ''
+        Z index of the buffer.
+      '';
+
+      # currently not usable, for detail see https://github.com/otavioschwanck/arrow.nvim/pull/43#issue-2236320268
+      # treesitter_context = helpers.defaultNullOpts.mkNullable null '''';
```
hmajid2301, line 144: wasn't sure if i should include this or just remove it as the option doesn't work. I don't think the typing is correct either :thinking:

traxys: I'd suggest not including it, you can mention in the PR/commit message that this option is _not_ handled because it is broken

hmajid2301: done

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+      delete_mode = helpers.defaultNullOpts.mkStr "d" ''
+        Mapping to go to delete mode, where you can remove bookmarks.
+      '';
+
+      clear_all_items = helpers.defaultNullOpts.mkStr "C" ''
+        Mapping to clear all bookmarks.
+      '';
+
+      toggle = helpers.defaultNullOpts.mkStr "s" ''
+        Mapping to save if separate_save_and_remove is true.
```
GaetanLepage, line 53:
```suggestion
        Mapping to save if `separate_save_and_remove` is true.
```

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+      open_horizontal = helpers.defaultNullOpts.mkStr "-" ''
+        Mapping to open bookmarks in horizontal split.
+      '';
+
+      quit = helpers.defaultNullOpts.mkStr "q" ''
+        Mapping to quit arrow.
+      '';
+
+      remove = helpers.defaultNullOpts.mkStr "x" ''
+        Mapping to remove bookmarks. Only used if separate_save_and_remove is true.
```
GaetanLepage, line 69:
```suggestion
        Mapping to remove bookmarks. Only used if `separate_save_and_remove` is true.
```

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+      open_horizontal = "-";
+      quit = "q";
+      remove = "x";
+      next_item = "]";
+      prev_item = "[";
+    };
+    custom_actions = {
+      open = ''function(target_file_name, current_file_name) end'';
+      split_vertical = ''function(target_file_name, current_file_name) end'';
+      split_horizontal = ''function(target_file_name, current_file_name) end'';
```
GaetanLepage, line 197:
```suggestion
      open = "function(target_file_name, current_file_name) end";
      split_vertical = "function(target_file_name, current_file_name) end";
      split_horizontal = "function(target_file_name, current_file_name) end";
```

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+    global_bookmarks = helpers.defaultNullOpts.mkBool false ''
+      If true arrow will save files globally (ignores separate_by_branch).
+    '';
+
+    index_keys = helpers.defaultNullOpts.mkStr "123456789zxcbnmZXVBNM,afghjklAFGHJKLwrtyuiopWRTYUIOP" ''
+      Keys mapped to bookmark index.
+    '';
+
+    full_path_list =
+      helpers.defaultNullOpts.mkNullable (with types; listOf str) ''[ "update_stuff" ]''
```
GaetanLepage, line 166:
```suggestion
      helpers.defaultNullOpts.mkListOf types.str ''[ "update_stuff" ]''
```

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+      The leader key to use for arrow. Will precede all mappings.
+      Recommended to be a single character.
+    '';
+
+    save_key = helpers.defaultNullOpts.mkStr "cwd" ''
+      What will be used as root to save the bookmarks. Can be also `git_root`.
+    '';
+
+    global_bookmarks = helpers.defaultNullOpts.mkBool false ''
+      If true arrow will save files globally (ignores separate_by_branch).
```
GaetanLepage, line 158:
```suggestion
      If true arrow will save files globally (ignores `separate_by_branch`).
```

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,223 @@
...
+      '';
+    };
+
+    custom_actions = {
+      open =
+        helpers.defaultNullOpts.mkLuaFn
+          ''
+            function(target_file_name, current_file_name) end
+          ''
+          "target_file_name = file selected to be open, current_file_name = filename from where this was called";
```
GaetanLepage, line 87:
```suggestion
          ''
            - `target_file_name`: file selected to be open
            - `current_file_name`: filename from where this was called
          '';
```

tests/test-sources/plugins/utils/arrow.nix
```diff
@@ -0,0 +1,63 @@
+{
```
GaetanLepage, line 1: Why no `defaults` configuration ? Is it the same as `examples` ?

GaetanLepage (changes requested):
Thank you for this good contribution.
Here are a few remarks.

GaetanLepage:
Could you please squash your commits into one, and while we are at it, name it `plugins/arrow: init`.

hmajid2301:
> Could you please squash your commits into one, and while we are at it, name it `plugins/arrow: init`.

Yes will do after I've resolved all of your issues.

tests/test-sources/plugins/utils/arrow.nix
GaetanLepage: Ususally, we have at least `empty` and `defaults` as test cases.
`example` is optional.
Could you please add a `defaults` test case where all of the settings are set to their default value ?

MattSturgeon (approved):
LGTM, once other feedback is addressed 👍

GaetanLepage:
@hmajid2301 I would recommend you not to use the "commit suggestions" button:
- It does not ensure formatting (which leads to useless CI failures)
- For each commit, a CI run is started and they do not get cancelled automatically. As those are quite heavy, it is better to address all of the requests locally and push once.

tests/test-sources/plugins/utils/arrow.nix
hmajid2301: of course these are the default :sweat: hhahha. i will look at how other modules do it.

hmajid2301:
> @hmajid2301 I would recommend you not to use the "commit suggestions" button:
> 
>     * It does not ensure formatting (which leads to useless CI failures)
> 
>     * For each commit, a CI run is started and they do not get cancelled automatically. As those are quite heavy, it is better to address all of the requests locally and push once.

good shout, will do that in future.

hmajid2301:
sorry ignore my bad rebase

tests/test-sources/plugins/utils/arrow.nix
hmajid2301: did you still want another use case? or is empty and defaults fine.

GaetanLepage: Yes, maybe an `example` one with this config: https://github.com/theCat69/nvim-config/blob/247a677d74aaca49f649d552da27e8051828e9eb/lua/plugins/jumping/arrow.lua#L4

hmajid2301: Not sure if you wanted comments, if not that should be done.

hmajid2301: do you know how I can type as an unsigned int or str ?  Seems width height can be an int which makes sense. Or it cna be `auto`

GaetanLepage: In this case it wouldn't be that because you want a regular string, not a raw string.
So, `helpers.defaultNullOpts.mkNullable (with types; either str ints.unsigned)`.

plugins/utils/arrow.nix
```diff
@@ -0,0 +1,224 @@
...
+      split_vertical = helpers.defaultNullOpts.mkLuaFn ''
+        function(target_file_name, current_file_name) end
+      '' "";
+
+      split_horizontal = helpers.defaultNullOpts.mkLuaFn ''
+        function(target_file_name, current_file_name) end
+      '' "";
+    };
+
+    window = {
```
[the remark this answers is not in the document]
MattSturgeon: Would this suggestion also apply to the `custom_actions` option?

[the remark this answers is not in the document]
GaetanLepage: Yes that's kind of the idea.

GaetanLepage (approved):
LGTM !
Thanks @hmajid2301 !

hmajid2301:
As ever thank you for your patience @GaetanLepage!

PR #1524, opened 2024-05-15
plugins/languages/rustaceanvim: fix executor option

wdcocq:
The old option would never convert a string to raw lua code. As it would always interpret it as an enum value, even if it wasn't one of the accepted values.

Using the already existing helper option types would either always convert it to a string or always convert it to raw lua code.  So I've added new option types that test whether it's an value that's part of the enum values and if not, convert it to raw lua: `mkNullOrEnumOrStrLua(Fn)` and `defaultNullOpts.mkEnumOrStrLua(Fn)` 

I'm pretty new to Nix, Neovim and Lua. So if this isn't very idiomatic/correct, please let me know!

Also I've added the `"neotest"` value to the possible executors.

### DIFF
--- a/plugins/languages/rustaceanvim.nix
+++ b/plugins/languages/rustaceanvim.nix
@@ -27,91 +27,123 @@ in
       example = null;
     };
 
-    tools = {
-      executor = helpers.defaultNullOpts.mkNullable (
-        with helpers.nixvimTypes;
-        either strLuaFn (enum [
+    tools =
+      let
+        executors = [
           "termopen"
           "quickfix"
           "toggleterm"
           "vimux"
-        ])
-      ) "termopen" "How to execute terminal commands.";
+          "neotest"
+        ];
+        testExecutors = executors ++ [ "background" ];
+      in
+      {
+        executor = helpers.defaultNullOpts.mkEnum executors "termopen" ''
+          `{execute_command}  (fun(cmd:string,args:string[],cwd:string|nil,opts?:RustaceanExecutorOpts))`
+          The executor to use for runnables/debuggables. 
 
-      onInitialized = helpers.mkNullOrLuaFn ''
-        `fun(health:RustAnalyzerInitializedStatus)`
-        Function that is invoked when the LSP server has finished initializing.
-      '';
+          Example:
+          ```lua
+          {
+            execute_command = function(command, args, cwd, _)
+              require('toggleterm.terminal').Terminal
+                :new({
+                  dir = cwd,
+                  cmd = require('rustaceanvim.shell').make_command_from_args(command, args),
+                  close_on_exit = false,
+                  direction = 'vertical',
+                })
+                :toggle()
+            end
+          }
+          ```
+        '';
 
-      reloadWorkspaceFromCargoToml = helpers.defaultNullOpts.mkBool true ''
-        Automatically call `RustReloadWorkspace` when writing to a `Cargo.toml` file.
-      '';
+        testExecutor = helpers.defaultNullOpts.mkEnum testExecutors "termopen" ''
+          `{execute_command}  (fun(cmd:string,args:string[],cwd:string|nil,opts?:RustaceanExecutorOpts))`
+          The executor to use for runnables that are tests/testables 
+        '';
 
-      hoverActions = {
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Hello,

Thank you very much for your interest in the project and this proposal.
So, your observation is correct and the current implementation is wrong.
Here the correct type would be `either rawType (enum [...])`.
This means that the user should be able to provide either:
- a string value. In this case, it **must** be one of the enum values.
- a raw lua value: `{ __raw = "some lua code"; }`

Your solution, although interesting, is not that robust and predictable by the user.
We do not want to make the _final type_ (i.e. normal string or raw lua) depend on the value itself.

wdcocq:
I thought I actually tested that but it was probably early in my discovery phase. Anyway, I've adopted your recommendations.

lib/options.nix
```diff
@@ -72,7 +72,7 @@ rec {
           ''
       );
 
-    mkNullableWithRaw = type: mkNullable (maybeRaw type);
+    mkNullableWithRaw = type: mkNullable (nixvimTypes.maybeRaw type);
```
GaetanLepage, line 75: Not sure that this is need.

wdcocq: A that is a leftover. but it will error out without it. But `mkNullableWithRaw` is never used anywhere so it doesn't normally turn up.

wdcocq:
```
 error: undefined variable 'maybeRaw'
        at /nix/store/p39jzjnq4qa2v2vnfdwlzv05zmn59083-source/lib/options.nix:75:43:
            74|
            75|     mkNullableWithRaw = type: mkNullable (maybeRaw type);
              |                                           ^
            76|
```

GaetanLepage: Nice catch indeed. Fixed in #1526

GaetanLepage: Merged ! Can you rebase on `main` please ?

wdcocq: done

plugins/languages/rustaceanvim.nix
```diff
@@ -27,92 +27,127 @@ in
...
+        executors = [
           "termopen"
           "quickfix"
           "toggleterm"
           "vimux"
-        ])
-      ) "termopen" "How to execute terminal commands.";
+          "neotest"
+        ];
+        test_executors = executors ++ [ "background" ];
```
GaetanLepage, line 39:
```suggestion
        testExecutors = executors ++ [ "background" ];
```
In general, we use camelCase when coding in nix.

GaetanLepage (changes requested):
This PR does not remove/rename any of the existing options right ?

wdcocq:
> This PR does not remove/rename any of the existing options right ?

No it adds the enum value `"neotest"` to `tools.executor` and adds two new options; `tools.test_executor` and `tools.crate_test_executor`. Which also get the added enum value of `"background"` a top of the normal executor options.
Besides the obvious raw lua code now ;)

As per documention of rustaceanvim:
```
        {executor?}                          (RustaceanExecutor|executor_alias)           The executor to use for runnables/debuggables
        {test_executor?}                     (RustaceanExecutor|test_executor_alias)      The executor to use for runnables that are tests / testables
        {crate_test_executor?}               (RustaceanExecutor|test_executor_alias)      The executor to use for runnables that are crate test suites (--all-targets)

...

executor_alias                                                  *executor_alias*

    Type: ~
        "termopen"|"quickfix"|"toggleterm"|"vimux"|"neotest"


test_executor_alias                                        *test_executor_alias*

    Type: ~
        executor_alias|"background"
```

The rest is just added documentation

So I don't think this could break backwards compatibility, as it's a superset of the old possibilities.

plugins/languages/rustaceanvim.nix
wdcocq: I see, the rustaceanvim snake_case must have thrown me off..

GaetanLepage:
Thanks for addressing all my comments.
Could you please squash all your commits into one: `plugins/rusteceanvim: update options"

wdcocq:
Yeah something went wrong with the squash there, should be ok now

GaetanLepage (approved):

traxys (approved):

PR #1530, opened 2024-05-17
lib/options: change "default: ..." to "Plugin default: ..."

GaetanLepage:
I hope that this change will heavily reduce people's legitimate confusion on having to "Default: " In the documentation.


GaetanLepage:
Or maybe `Plugin's default` or `Upstream default` ?

MattSturgeon (approved):
Nice!

Either "Plugin default" or "Plugin's default" are fine IMO.

PR #1531, opened 2024-05-17
lib/options: introduce new mkPackageOption for dependencies

GaetanLepage:
This new helper is now used to the package options exposed to the user (that can be set to `null` by the user).


MattSturgeon:
I wonder if instead of renaming the existing function to `mkPluginPackageOption`, we could change the description of that to be more generic? That way the only distinction between the two helpers is that one produces a nullable option.

Both helpers could share a `mkDesc` function, which the nullable version concatenates with the "can be null to disable installation" note.

GaetanLepage:
> I wonder if instead of renaming the existing function to `mkPluginPackageOption`, we could change the description of that to be more generic? That way the only distinction between the two helpers is that one produces a nullable option.
> 
> Both helpers could share a `mkDesc` function, which the nullable version concatenates with the "can be null to disable installation" note.

Well, indeed we could "generalize" it even more to have a single one.
But wouldn't it be too much here ? I don't feel like this implementation is super heavy...
What is your opinion @traxys ?

MattSturgeon:
> But wouldn't it be too much here ? I don't feel like this implementation is super heavy...

I'm not really concerned about it being too heavy, per-se... Rather I just don't see why the plugin-specific variant needs to be explicitly "plugin" related.

From my perspective both helpers make a `package` option with a uniform description, the only difference is one is nullable.

Saying "Plugin to use" instead of "Package to use" seems unnecessarily specific, as it should be clear from the context that it is a plugin package?

MattSturgeon:
> To me those are enough differences to justify two functions.

Agreed. Apologies if I wasn't clear; I'm not proposing we only have one function.

My suggestion was something along these lines:

```nix
# private/local helper
mkDesc = name: "Which package to use for `${name}`";

# Existing public function,
# but now using mkDesc
mkPackageOption = /* omitted */;

# New public function, also using mkDesc
mkNullOrPackageOption = /* omitted */;
```

MattSturgeon:
> Creating a function to output a 6 words sentence and using it twice is not really worth it.
> 
> I acknowledge that we are 100% in subjective-coding-style land here ^^.

Sure, TBH I'd still have the same preference\* even if we just duplicated the string _without_ the extra `mkDesc` function.

\* as in I'd prefer both functions talk about "packages" rather than "plugins".

I just thought having that string defined in one place would make maintenance easier if things are changed in the future.

As an aside, not renaming the original `mkPackageOption` would significantly reduce the size of the diff...

GaetanLepage:
> * as in I'd prefer both functions talk about "packages" rather than "plugins".

I think that you are definitely right on this. I changed the phrasing to be more accurate.

MattSturgeon (approved):
Overall, I'm in favour of this PR; the new nullable package option helper is a definite improvement to existing option definitions.

I think this implementation is fine, _if_ there's no foreseeable use case for a non-null package option helper that _doesn't_ relate to plugins.

If non-plugin non-null packages _is_ something we might want, then I prefer my above suggestion of having `mkPackageOption` and `mkNullablePackageOption` instead of renaming the non-null function to `mkPluginPackageOption`.

I also feel like avoiding the rename would help reduce the scope of this PR.

No objections, but I'd like to hear @traxys thoughts before merging.

GaetanLepage:
> If non-plugin non-null packages _is_ something we might want, then I prefer my above suggestion of having `mkPackageOption` and `mkNullablePackageOption` instead of renaming the non-null function to `mkPluginPackageOption`.

This actually makes more sense to me now. I did not have understand this before, my bad.
Both solutions are fine to me. I'll let traxys choose.

GaetanLepage:
Well, for now we have no other use of the non-null package option than for plugin packages. We can then have a more precise description.

traxys (changes requested):
I'm ok with the chosen approach

traxys (approved):

GaetanLepage:
Thanks @MattSturgeon for your constructive comments !

PR #1532, opened 2024-05-17
Fix statuscol.nvim option type

jrichardsen:
The `statuscol.nvim` plugin allows for segments to not specify a `text` attribute (see [example](https://user-images.githubusercontent.com/31730729/230627808-de0b1e97-116d-4016-b4ba-cb709dfcd980.png) in [the plugin's README](https://github.com/luukvbaal/statuscol.nvim/blob/main/README.md).

This PR makes the respective nixvim option nullable and sets it to null by default.

### DIFF
--- a/plugins/ui/statuscol.nix
+++ b/plugins/ui/statuscol.nix
@@ -43,7 +43,8 @@ helpers.neovim-plugin.mkNeovimPlugin config {
           freeformType = with types; attrsOf anything;
           options = {
             text = mkOption {
-              type = with helpers.nixvimTypes; listOf (either str rawLua);
+              type = with helpers.nixvimTypes; nullOr (listOf (either str rawLua));
+              default = null;
               description = "Segment text.";
               example = [ "%C" ];
             };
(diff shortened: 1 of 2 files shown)


plugins/ui/statuscol.nix
```diff
@@ -43,7 +43,8 @@ helpers.neovim-plugin.mkNeovimPlugin config {
           freeformType = with types; attrsOf anything;
           options = {
             text = mkOption {
-              type = with helpers.nixvimTypes; listOf (either str rawLua);
+              type = with helpers.nixvimTypes; nullOr (listOf (either str rawLua));
```
GaetanLepage, line 46: In this case, directly use the nixvim helper:
```nix
text = helpers.mkNullOrOption (with helpers.nixvimTypes; listOf (either str rawLua)) ''
  Segment text.
'';
```

jrichardsen: I thought about this, but then I would have to remove the example from the option, right?

GaetanLepage: Yes true, I guess it is fine to leave it like that. I should stop nitpicking...

PR #1538, opened 2024-05-19
plugins/rustaceanvim: switch to mkNeovimPlugin

GaetanLepage:


GaetanLepage:
I included the changes from #1536 in a similar manner (using `mkDefault`.

wdcocq:
> I included the changes from #1536 in a similar manner (using `mkDefault`.

I've tested this, and as of now this doesn't work as the plugin is instantiated outside the LSP init code, where in the old plugin style  it would be instantiated within the LSP init code. Thus we get a `Undefined global: '__lspOnAttach'`

plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -0,0 +1,62 @@
...
+    };
+  };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    extraPackages = [ cfg.rustAnalyzerPackage ];
+
+    plugins.rustaceanvim.settings.server.on_attach = mkDefault "__lspOnAttach";
+
+    extraConfigLua = ''
```
MattSturgeon, line 58: Would this resolve https://github.com/nix-community/nixvim/pull/1538#issuecomment-2120256132 ?

I think this is essentially what we did before.

```suggestion
    plugins.lsp.postConfig = ''
```

traxys: I think so

GaetanLepage: What if the user has not enabled `plugins.lsp` .

GaetanLepage: We should at least do an `assertion`/`warning` to ensure that the user has enabled `plugins.lsp`.
However, is that legit to _force_ the user to use `lspconfig.nvim` when using rustaceanvim ?

MattSturgeon: [upstream](https://github.com/mrcjkb/rustaceanvim/) states "No dependency on lspconfig" so I don't _think_ we need anything more than a warning? Not sure if it really makes sense to use it without LSP enabled though.

MattSturgeon: Ah of course. In that case we should have an assertion.

If we want to support using it without `plugins.lsp.enable` then we could do things more conditionally, e.g.:

```nix
(
  setAttrByPath
  (
    if config.plugins.lsp.enable
    then [ "plugins" "lsp" "postConfig" ]
    else [ "extraConfigLua" ]
  )
  ''vim.g.rustaceanvim = ${helpers.toLuaObject cfg.settings}''
);
```

GaetanLepage: Well I don't know...
"Ideally", we could just make sure that this snippet of lua is after the **eventual** LSP snippet.
We are only trying to achieve this here.
It seems overkilled to me to enforce enabling `plugins.lsp` just to make sure that the rustacenvim's config is after (and not before) the LSP block.

MattSturgeon: > Or... `plugins.lsp.enable = mkDefault true` in `extraConfig`. But I doubt you guys would like this.

nixpkgs does those kinda "sane defaults" quite a bit, they might even use `mkForce` in this scenario... But I think for the most part we've decided to avoid "magic" in favor of users setting things explicitly?

MattSturgeon: > It seems overkilled to me to enforce enabling `plugins.lsp` just to make sure that the rustacenvim's config is after (and not before) the LSP block.

That's fair.

> "Ideally", we could just make sure that this snippet of lua is after the **eventual** LSP snippet.

I don't know how we could achieve this without some complexity.

Maybe something similar to my [earlier suggestion](https://github.com/nix-community/nixvim/pull/1538#discussion_r1607235196) but also only set `on_attach` when lsp is enabled?

If you don't like using `setAttrByPath` we could just `mkMerge` multiple `mkIf` definitions instead.

traxys: Well if plugins.lsp is not enabled then __lspOnAttach would not exist, so we could but rustacean.nvim anywhere in the configuration. I would rather go on this path, i.e:
- If plugins.lsp.enable then the onAttach default is __lspOnAttach, and we put the setup in plugins.lsp.postConfig
- If !plugins.lsp.enable then there is no onAttach default, and we put the setup in extraConfigLua

GaetanLepage: This seems totally reasonable to me.

plugins/languages/rust/rustaceanvim/settings-options.nix
```diff
@@ -0,0 +1,294 @@
...
+          ---@cast cmd string[]
+          local rs_bin = cmd[1]
+          return vim.fn.executable(rs_bin) == 1
+        end
+      ```
+    '';
+
+    on_attach = helpers.mkNullOrLuaFn ''
+      Function to call on attach.
+      By default, it will use the Nixvim global `__lspOnAttach` function.
```
MattSturgeon, line 192: This should probably be clarified; maybe something along the lines of this?

```suggestion
      If `plugins.lsp` is enabled, it defaults to the Nixvim global `__lspOnAttach` function. Otherwise it defaults to `null`.
```

plugins/languages/rust/rustaceanvim/settings-options.nix
```diff
@@ -0,0 +1,294 @@
...
+    '';
+
+    cmd = helpers.mkNullOrStrLuaFnOr (with types; listOf str) ''
+      Command and arguments for starting rust-analyzer.
+
+      This can also be the definition of a function:
+      `fun(project_root:string|nil,default_settings:table):table`
+
+      Default:
```
MattSturgeon, line 201: Should this be "Plugin default:"? (This question applies to a few options)

plugins/languages/rust/rustaceanvim/settings-options.nix
GaetanLepage: Agreed, done.

plugins/languages/rust/rustaceanvim/settings-options.nix
GaetanLepage: Fixed (I hope) everywhere.

MattSturgeon (approved):
LGTM. ~~Would be good if someone tested this still works at runtime. I.e. verify @wdcocq's issue is resolved and no other runtime issues were missed.~~

> These changes work for me and resolve my PR 👍

Good stuff 😅🎉

PR #1544, opened 2024-05-20
plugins/telescope: support non-builtin keymaps

### LINKED ISSUE #1493: [BUG] keymaps: mapping telescope's undo extension breaks custom keymaps
| Plugin       | `telescope`  | 
| Nixpkgs      | `63c3a29ca82437c87573e4c6919b09a24ea61b0f` | 
| Home Manager | `9036fe9ef8e15a819fa76f47a8b1f287903fb848` | 

## Description

Binding a key to Telescope's `undo` plugin causes custom keymaps to stop working.

## Minimal, Reproducible Example

```nix
programs.nixvim = {
  plugins.which-key.enable = true; # helps see what's happening
  plugins.telescope = {
    enable = true;
    keymaps = {

        ###### the problematic line - comment this to "fix" the bug
        "<leader>fu" = { action = "undo";         options.desc = "Telescope Undo"; };

        # These should work fine
        "<leader>ff" = { action = "builtin";      options.desc = "Telescope Builtin"; };
        "<leader>fg" = { action = "live_grep";    options.desc = "Telescope Live Grep"; };
    };
    # NOTE: the undo plugin doesn't have to be enabled
    #extensions.undo.enable = true;
  };

  # test keymaps
  globals.mapleader = " ";
  keymaps = [
    {
      action = "<cmd>help<CR>";
      key = "gh";
      mode = [ "n" ];
    }
    {
      action = "<cmd>help<CR>";
      key = "yu";
      mode = [ "n" ];
    }
    {
      action = "<cmd>help<CR>";
      key = "<leader>g";
      mode = [ "n" ];
    }
  ];
}
```
### Steps to reproduce
1. load nixvim with the config above
2. try using any of the custom keymaps (`gh`, `yu`, `<leader>g`)
	- they don't work (help page doesn't open)
	- note that `<leader>ff` and the other `telescope.keymaps` options work (as they should)
3. comment line 8 (the problematic line)
4. try using the custom keymaps from before
	- help page opens, they work
	- all other `telescope.keymaps` options still work 

<sub> it took me 2 days to find this :sob:</sub>

MattSturgeon:
Use the `:Telescope` command instead of relying directly on
`require("telescope.builtin")`.

To quote [upstream docs](https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#vim-commands):

> All `telescope.nvim` functions are wrapped in vim commands for easy access, tab completions and setting options.

Fixes #1493

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -90,7 +90,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       {
         mode = mapping.mode or "n";
         inherit key;
-        action.__raw = "require('telescope.builtin').${actionStr}";
+        action = "<cmd>Telescope ${actionStr}<cr>";
 
         options = {
           silent = cfg.keymapsSilent;
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -20,7 +20,19 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   deprecateExtraOptions = true;
   optionsRenamedToSettings = [ "defaults" ];
 
-  imports = [ ./extensions ];
+  imports = [
+    ./extensions
+
+    # TODO introduced 2024-05-24: remove 2024-08-24
+    (mkRemovedOptionModule
+      [
+        "plugins"
+        "telescope"
+        "keymapsSilent"
+      ]
+      "This option no longer has any effect now that the `plugin.telescope.keymaps` implementation uses `<cmd>`."
+    )
+  ];
 
   extraOptions = {
     keymaps = mkOption {
@@ -44,12 +56,6 @@ helpers.neovim-plugin.mkNeovimPlugin config {
           options.desc = "Telescope Git Files";
         };
       };
-    };
-
-    keymapsSilent = mkOption {
-      type = types.bool;
-      description = "Whether telescope keymaps should be silent";
-      default = false;
     };
 
     highlightTheme = mkOption {
@@ -91,10 +97,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
         mode = mapping.mode or "n";
         inherit key;
         action = "<cmd>Telescope ${actionStr}<cr>";
-
-        options = {
-          silent = cfg.keymapsSilent;
-        } // (mapping.options or { });
+        options = mapping.options or { };
       }
     ) cfg.keymaps;
 
--- a/tests/test-sources/plugins/telescope/default.nix
+++ b/tests/test-sources/plugins/telescope/default.nix
@@ -14,7 +14,6 @@
           options.desc = "Telescope Git Files";
         };
       };
-      keymapsSilent = true;
       highlightTheme = "gruvbox";
     };
   };


MattSturgeon:
`<cmd>` mappings [are not echod](https://neovim.io/doc/user/map.html#%3CCmd%3E), so the `keymapsSilent` now has no effect and could be deprecated.

traxys (approved):

GaetanLepage (changes requested):
> `<cmd>` mappings [are not echod](https://neovim.io/doc/user/map.html#%3CCmd%3E), so the `keymapsSilent` option now has no effect and could be deprecated.

Let's do that in this PR then.

MattSturgeon:
I think there's a design decision to make here.

Using using the `:Telescope` command is simpler, since both builtin and extension pickers are directly available, and options can also be passed to the picker as part of the command.

> The viml api can be called using commands like: `:Telescope find_files hidden=true layout_config={"prompt_position":"top"}`
>
> This involves setting options using an `=` and using viml syntax for lists and dictionaries when the corresponding lua function requires a table.

However [`:h telescope.command`](https://github.com/nvim-telescope/telescope.nvim/blob/5665d93988acfbb0747bdbf4f4cb583bcebc8930/doc/telescope.txt#L785) does point out some limitations:

> The viml api is more indirect, as first the command must be parsed to the
relevant lua equivalent, which brings some limitations. 
>
> One limitation of the viml api is that there can be no spaces in any of the
options.
>
> For example, if you want to use the `cwd` option for `find_files` to
specify that you only want to search within the folder `/foo bar/subfolder/`
you could not do that using the viml api, as the path name contains a space.

It's also unclear whether _all_ third-party extensions will expose their picker functions via the `:Telescope` command...

----

That said, this is already more powerful than the current approach, since the current telescope keymap wrapper doesn't allow passing any option arguments to the picker function _anyway_. Nor does it allow using third-party (extension) pickers.

If we decide we don't like the `:Telescope` approach, the only other way I can think of to (directly) support third-party picker functions is by adding an option along the lines of `keymap.extension = mkNullOrOption str "If `action` is a third-party extension function, this should be the name of the extension."`. I think this is more cumbersome from a UX perspective, though.

----

We could also augment either the old or new system by allowing `action` to be a raw string. This could be used as the body of a lua function where we've already defined some local variables (e.g. `telescope`, `builtins`, `extensions`). This way the user can use the lua API to call telescope functions.

plugins/telescope/default.nix
```diff
@@ -20,7 +20,21 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   deprecateExtraOptions = true;
   optionsRenamedToSettings = [ "defaults" ];
 
-  imports = [ ./extensions ];
+  imports = [
+    ./extensions
+
+    # TODO introduced 2024-05-24: remove 2024-08-24
+    (mkRemovedOptionModule
```
MattSturgeon, line 27: > > `<cmd>` mappings [are not echod](https://neovim.io/doc/user/map.html#%3CCmd%3E), so the `keymapsSilent` option now has no effect and could be deprecated.
> 
> Let's do that in this PR then.

Done

MattSturgeon:
@GaetanLepage could you re-review when you have chance?

plugins/telescope/default.nix
```diff
@@ -20,7 +20,21 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+      [
+        "plugins"
+        "telescope"
+        "keymapsSilent"
+      ]
+      ''
+        This option no longer has any effect now that the `plugin.telescope.keymaps` implementation uses `<cmd>`.
+      ''
```
GaetanLepage, line 35:
```suggestion
      "This option no longer has any effect now that the `plugin.telescope.keymaps` implementation uses `<cmd>`."
```

Just for the sake of requesting a change X)

GaetanLepage: (I would understand if you told me that you didn't care haha)

MattSturgeon: No problem, done

GaetanLepage (approved):
Nice ! LGTM.

PR #1546, opened 2024-05-21
modules/commands: allow lua commands, fix nargs

NickHu:
A couple of minor fixes here. Firstly, nargs needs 0 or 1, it will error if "0" or "1" are passed (as was previously the case). Secondly, `nvim_create_user_command` accepts a Lua function as its `{command}` parameter, e.g.

```lua
function(opts)
  local arg = opts.args
  ...
end
```
A simple tweak to the type of `command` is enough to fix this.

### DIFF
--- a/modules/commands.nix
+++ b/modules/commands.nix
@@ -9,7 +9,7 @@ let
   commandAttributes = types.submodule {
     options = {
       command = mkOption {
-        type = types.str;
+        type = with helpers.nixvimTypes; either str rawLua;
         description = "The command to run.";
       };
 
--- a/modules/commands.nix
+++ b/modules/commands.nix
@@ -16,8 +16,8 @@ let
       nargs =
         helpers.mkNullOrOption
           (types.enum [
-            "0"
-            "1"
+            0
+            1
             "*"
             "?"
             "+"
--- a/tests/test-sources/modules/commands.nix
+++ b/tests/test-sources/modules/commands.nix
@@ -4,6 +4,7 @@
       "W" = {
         command = ":w<CR>";
         desc = "Write file";
+        nargs = 0;
       };
       "Z" = {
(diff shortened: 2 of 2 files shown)


modules/commands.nix
```diff
@@ -8,16 +8,15 @@ with lib;
 let
   commandAttributes = types.submodule {
     options = {
-      command = mkOption {
-        type = types.str;
-        description = "The command to run.";
-      };
+      command = helpers.mkNullOrOption (with types; either str helpers.nixvimTypes.rawLua) ''
+        The command to run.
+      '';
```
GaetanLepage, line 13: The previous implementation was forcing the user to provide a value for this option.
Now, this is not the case anymore.
I would rather keep the previous behavior:
```nix
command = mkOption {
  type = with helpers.nixvimTypes; either str rawLua;
  description = "The command to run.";
};
```

GaetanLepage (changes requested):

modules/commands.nix
NickHu: Ah of course, good spot

GaetanLepage (changes requested):
Thanks, could you please add a relevant test case ?

GaetanLepage (changes requested):
Please, squash your changes into a single commit.

NickHu:
Can't you just squash-merge?

GaetanLepage:
> Can't you just squash-merge?

Yes, I can also do it myself.

However, the CI is failing.

NickHu:
It looks like a spurious CI error to me, unrelated to this PR. What do you think?

GaetanLepage (approved):
Thanks @NickHu !
Actually, it makes sense to have two separate commits for this.

PR #1554, opened 2024-05-23
plugins/lualine: fix extensions option

iSma:
Option "extensions" was incorrectly output under key "options", but it belongs to the top level of lualine's setup table.


MattSturgeon (approved):
LGTM, thanks!

PR #1555, opened 2024-05-24
plugins/lsp/language-servers: nil_ls add nix options

khaneliman:
Adding some options that were being ignored when I tried to add to config for nil_ls. 
https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -361,41 +361,8 @@ let
       name = "nil_ls";
       description = "nil for Nix";
       package = pkgs.nil;
-      settingsOptions = {
-        formatting.command = mkOption {
-          type = types.nullOr (types.listOf types.str);
-          default = null;
-          description = ''
-            External formatter command (with arguments).
-            It should accepts file content in stdin and print the formatted code into stdout.
-          '';
-        };
-        diagnostics = {
-          ignored = mkOption {
-            type = types.listOf types.str;
-            default = [ ];
-            description = ''
-              Ignored diagnostic kinds.
-              The kind identifier is a snake_cased_string usually shown together
-              with the diagnostic message.
-            '';
-          };
-          excludedFiles = mkOption {
-            type = types.listOf types.str;
-            default = [ ];
-            description = ''
-              Files to exclude from showing diagnostics. Useful for generated files.
-              It accepts an array of paths. Relative paths are joint to the workspace root.
-              Glob patterns are currently not supported.
-            '';
-          };
-        };
-      };
-      settings = cfg: {
-        nil = {
-          inherit (cfg) formatting diagnostics;
-        };
-      };
+      settingsOptions = import ./nil_ls-settings.nix { inherit lib helpers; };
+      settings = cfg: { nil = cfg; };
     }
     {
       name = "nimls";
--- a/tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
+++ b/tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
@@ -1,0 +1,33 @@
+{
+  example = {
+    plugins.lsp = {
+      enable = true;
+
+      servers.nil_ls = {
+        enable = true;
+
+        settings = {
+          diagnostics = {
+            ignored = [
+              "unused_binding"
+              "unused_with"
+            ];
+            excludedFiles = [ "Cargo.nix" ];
+          };
+          formatting = {
+            command = [ "nixfmt" ];
+          };
+          nix = {
+            binary = "nix";
+            maxMemoryMB = 2048;
+            flake = {
+              autoArchive = true;
+              autoEvalInputs = false;
+              nixpkgsInputName = "nixpkgs";
+            };
+          };
+        };
+      };
+    };
+  };
+}
--- a/plugins/lsp/language-servers/nil_ls-settings.nix
+++ b/plugins/lsp/language-servers/nil_ls-settings.nix
@@ -1,0 +1,63 @@
+{ lib, helpers }:
+# Options:
+#  - https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1#
+with lib;
+{
+  formatting = {
+    command = helpers.defaultNullOpts.mkListOf types.str "null" ''
+      External formatter command (with arguments).
+      It should accepts file content in stdin and print the formatted code into stdout.
+    '';
+  };
+
+  diagnostics = {
+    ignored = helpers.defaultNullOpts.mkListOf types.str "[]" ''
+      Ignored diagnostic kinds.
+      The kind identifier is a snake_cased_string usually shown together
+      with the diagnostic message.
+    '';
+
+    excludedFiles = helpers.defaultNullOpts.mkListOf types.str "[]" ''
+      Files to exclude from showing diagnostics. Useful for generated files.
+      It accepts an array of paths. Relative paths are joint to the workspace root.
+      Glob patterns are currently not supported.
+    '';
+  };
+
+  nix = {
+    binary = helpers.defaultNullOpts.mkStr "nix" ''
+      The path to the `nix` binary.
+    '';
+
+    maxMemoryMB = helpers.defaultNullOpts.mkInt 2560 ''
+      The heap memory limit in MiB for `nix` evaluation.
+      Currently it only applies to flake evaluation when `autoEvalInputs` is enabled, and only works
+      for Linux.
+      Other `nix` invocations may be also applied in the future.
+      `null` means no limit.
+      As a reference, `nix flake show --legacy nixpkgs` usually requires about 2GiB memory.
+    '';
+
+    flake = {
+      autoArchive = helpers.defaultNullOpts.mkBool false ''
+        Auto-archiving behavior which may use network.
+         - `null`: Ask every time.
+         - `true`: Automatically run `nix flake archive` when necessary.
+         - `false`: Do not archive. Only load inputs that are already on disk.
+      '';
+
+      autoEvalInputs = helpers.defaultNullOpts.mkBool false ''
+        Whether to auto-eval flake inputs.
+        The evaluation result is used to improve completion, but may cost lots of time and/or memory.
+      '';
+
+      nixpkgsInputName = helpers.defaultNullOpts.mkStr "nixpkgs" ''
+        The input name of nixpkgs for NixOS options evaluation.
+
+        The options hierarchy is used to improve completion, but may cost lots of time and/or memory.
+        If this value is `null` or is not found in the workspace flake's inputs, NixOS options are
+        not evaluated.
+      '';
+    };
+  };
+}


plugins/lsp/language-servers/default.nix
```diff
@@ -389,10 +389,64 @@ let
...
+              '';
+            };
+          };
+        };
       };
       settings = cfg: {
         nil = {
-          inherit (cfg) formatting diagnostics;
+          inherit (cfg) formatting diagnostics nix;
         };
```
MattSturgeon, line 450: > options that were being ignored when I tried to add to config for `nil_ls`

Should this just be `nil = cfg;`, is inheriting specific options the reason why this didn't work as-per RFC42?

@GaetanLepage @traxys

khaneliman: Makes sense, I tested that built my config still and supports new attribute sets not defined.

khaneliman:
Sorry for false push, meant to test locally before pushing that commit...

khaneliman:
Alright, seems to work same as before with new file for settings.

MattSturgeon (approved):
LGTM. It'd be great if this added a test (similar to the [`nixd` test](https://github.com/khaneliman/nixvim/blob/nil/tests/test-sources/plugins/lsp/language-servers/nixd.nix)) with example and/or defaults, but so few language servers have tests I don't think this is a blocker.

GaetanLepage (changes requested):

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+#  - https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1#
+with lib; {
+  formatting = {
+    command = helpers.defaultNullOpts.mkListOf types.str "null" ''
+      External formatter command (with arguments).
+      It should accepts file content in stdin and print the formatted code into stdout.
+    '';
+  };
+  diagnostics = {
+    ignored = helpers.defaultNullOpts.mkListOf types.str "null" ''
```
MattSturgeon, line 15:
```suggestion
    ignored = helpers.defaultNullOpts.mkListOf types.str "[]" ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      Ignored diagnostic kinds.
+      The kind identifier is a snake_cased_string usually shown together
+      with the diagnostic message.
+    '';
+    excludedFiles = helpers.defaultNullOpts.mkListOf types.str "null" ''
```
MattSturgeon, line 20:
```suggestion
    excludedFiles = helpers.defaultNullOpts.mkListOf types.str "[]" ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      Files to exclude from showing diagnostics. Useful for generated files.
+      It accepts an array of paths. Relative paths are joint to the workspace root.
+      Glob patterns are currently not supported.
+    '';
+  };
+  nix = {
+    binary = helpers.defaultNullOpts.mkStr "null" ''
```
MattSturgeon, line 27:
```suggestion
    binary = helpers.defaultNullOpts.mkStr "nix" ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      The path to the `nix` binary.
+    '';
+    maxMemoryMB = helpers.defaultNullOpts.mkInt "null" ''
```
MattSturgeon, line 30:
```suggestion
    maxMemoryMB = helpers.defaultNullOpts.mkInt 2560 ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      about 2GiB memory.
+    '';
+    flake = {
+      autoArchive = helpers.defaultNullOpts.mkBool "null" ''
+        Auto-archiving behavior which may use network.
+         - null: Ask every time.
+         - true: Automatically run `nix flake archive` when necessary.
+         - false: Do not archive. Only load inputs that are already on disk.
+      '';
+      autoEvalInputs = helpers.defaultNullOpts.mkBool "null" ''
```
MattSturgeon, line 45:
```suggestion
      autoEvalInputs = helpers.defaultNullOpts.mkBool false ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+        Whether to auto-eval flake inputs.
+        The evaluation result is used to improve completion, but may cost
+        lots of time and/or memory.
+      '';
+      nixpkgsInputName = helpers.defaultNullOpts.mkStr "null" ''
```
MattSturgeon, line 50:
```suggestion
      nixpkgsInputName = helpers.defaultNullOpts.mkStr "nixpkgs" ''
```

MattSturgeon (changes requested):
The default passed to `helpers.defaultNullOpts` is the _plugin default_, it will always produce an option with the _nix default_ set to `null`.

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
+{
+  lib,
+  helpers,
+}:
+# Options:
+#  - https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1#
+with lib; {
+  formatting = {
+    command = helpers.defaultNullOpts.mkListOf types.str null ''
```
MattSturgeon, line 9: `mkListOf str` produces a list of `maybeRaw str`, which is different from the `listOf str` we had before.

Would something like this be more correct @GaetanLepage?

```nix
mkNullable (listOf str) "null" "<description-here>"
```

Or should this type accept `__raw` lua strings @khaneliman?

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      External formatter command (with arguments).
+      It should accepts file content in stdin and print the formatted code into stdout.
+    '';
+  };
+  diagnostics = {
+    ignored = helpers.defaultNullOpts.mkListOf types.str [] ''
```
MattSturgeon, line 15: Same

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      Ignored diagnostic kinds.
+      The kind identifier is a snake_cased_string usually shown together
+      with the diagnostic message.
+    '';
+    excludedFiles = helpers.defaultNullOpts.mkListOf types.str [] ''
```
MattSturgeon, line 20: Same

plugins/lsp/language-servers/nil-ls-settings.nix
khaneliman: I believe it'd just be the executable path or name.

MattSturgeon:
CI is failing on formatting, if you want you can install the pre-commit hook by entering the devshell at least once (you can exit again and use your normal shell if you like, entering once is enough to install the hook). Enter using `nix develop`.

`nix fmt` should fix the formatting, at least.

khaneliman:
> CI is failing on formatting, if you want you can install the pre-commit hook by entering the devshell at least once (you can exit again and use your normal shell if you like, entering once is enough to install the hook). Enter using `nix develop`.
> 
> Installing the hook will check the formatting of modified files before each commit.
> 
> `nix fmt` should fix the formatting, at least.

The pre-commit hook was running on my commits. I'll let it format the full file, but I tried to keep changes to just my code.

plugins/lsp/language-servers/default.nix
```diff
@@ -275,19 +274,18 @@ let
...
       # Use the old name of the lua LS if the user is on a stable branch of nixpkgs
       # Rename occurred here: https://github.com/NixOS/nixpkgs/pull/215057
       package =
-        if (hasAttr "lua-language-server" pkgs) then
-          pkgs.lua-language-server
-        else
-          pkgs.sumneko-lua-language-server;
+        if (hasAttr "lua-language-server" pkgs)
+        then pkgs.lua-language-server
+        else pkgs.sumneko-lua-language-server;
```
MattSturgeon, line 288: Strange that unrelated code needs to be formatted...

Was this definitely `nix fmt` and not some additional formatter you have in your editor/workflow? The result looks like [Alejandra](https://github.com/kamadorueda/alejandra).

If this is coming from `nix fmt`, maybe rebase into a separate commits; one doing `nix fmt` and another adding this PR's changes.

MattSturgeon: > Looks like the hook is fighting the nix fmt..

Interesting

khaneliman: I had to direnv reload so it would reinstall the pre-commit-hook. It must have changed.

plugins/lsp/language-servers/nil-ls-settings.nix
khaneliman: Gonna be hitting the road for a few hours. If this needs to change I can address later.

plugins/lsp/language-servers/default.nix
MattSturgeon: I can't see why there'd be an issue; The only difference is `nix fmt` is wrapped with `treefmt` while `pre-commit` is directly using `statix` and `nixfmt-rfc-style`, which `treefmt` is _also_ configured to use... ([here](https://github.com/nix-community/nixvim/blob/9d858de2e9ab136d1c53d92af62fed8fccf492ab/flake-modules/dev/default.nix#L12-L36))

MattSturgeon: > Was the repo using Alejandra before and switched to new nixfmt-rfc-style?

Yeah, it was

> I had to direnv reload so it would reinstall the pre-commit-hook.

Makes sense, didn't consider you might have had an old hook installed :smile:

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,58 @@
+{ lib, helpers }:
+# Options:
+#  - https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1#
+with lib;
+{
+  formatting = {
+    command = helpers.defaultNullOpts.mkNullable (types.listOf types.str) null ''
```
MattSturgeon, line 7: I was under the impression `mkNullable` needed the plugin-default to be a string (`"null"`), but it looks like it gets passed through `toString`.

I guess we only need to pass in a string when we want to enforce formatting or display nested elements. 🤔

MattSturgeon (approved):
CI is passing and the diff looks good 👍 

Thanks for your patience!

plugins/lsp/language-servers/nil-ls-settings.nix
GaetanLepage: Hmmm, we should be careful here.
Indeed, `toString null` is `""` and not `"null"`.

MattSturgeon: Off-topic, but maybe `mkNullable` should use a (bespoke or third-party) pretty-stringify instead of `builtins.toString` to allow us to use actual `nix` and strings interchangeably?

For now, @khaneliman, could you switch to passing string defaults to `mkNullable` matching how it's done elsewhere?

GaetanLepage: > Off-topic, but maybe `mkNullable` should use a (bespoke or third-party) pretty-stringify instead of `builtins.toString` to allow us to use actual `nix` and strings interchangeably?

I agree, this could be nice !

khaneliman: Sounds good, updated to use the "null" string instead of null type.

plugins/lsp/language-servers/nil-ls-settings.nix
MattSturgeon: I'd like to run this by @traxys as his recent comments on one of my PRs are making me second guess this feedback.

Raw strings are really just lua code; i.e. no quotes. So a user might want to use a raw string here to reference some other lua string-variable defined elsewhere.

_I guess_ the only time we'd want to avoid them is when the value _won't_ end up directly in lua code, or we're going to do some string manipulation on it (and assume it is a string literal)

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,58 @@
+{ lib, helpers }:
+# Options:
+#  - https://github.com/oxalica/nil/blob/main/docs/configuration.md?plain=1#
+with lib;
+{
+  formatting = {
+    command = helpers.defaultNullOpts.mkNullable (types.listOf types.str) "null" ''
```
GaetanLepage, line 7:
```suggestion
    command = helpers.defaultNullOpts.mkListOf types.str "null" ''
```
Use `mkListOf` where possible.

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,58 @@
...
+      External formatter command (with arguments).
+      It should accepts file content in stdin and print the formatted code into stdout.
+    '';
+  };
+  diagnostics = {
```
GaetanLepage, line 12:
```suggestion
  };

  diagnostics = {
```

Please, add a blank line between each option declaration.

GaetanLepage (changes requested):
A few minor requests.
Thanks for your patience to address all of those comments.

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+    '';
+  };
+
+  diagnostics = {
+    ignored = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      Ignored diagnostic kinds.
+      The kind identifier is a snake_cased_string usually shown together
+      with the diagnostic message.
+    '';
+    excludedFiles = helpers.defaultNullOpts.mkListOf types.str [ ] ''
```
GaetanLepage, line 19:
```suggestion
    '';

    excludedFiles = helpers.defaultNullOpts.mkListOf types.str [ ] ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      It accepts an array of paths. Relative paths are joint to the workspace root.
+      Glob patterns are currently not supported.
+    '';
+  };
+
+  nix = {
+    binary = helpers.defaultNullOpts.mkStr "nix" ''
+      The path to the `nix` binary.
+    '';
+    maxMemoryMB = helpers.defaultNullOpts.mkInt 2560 ''
```
GaetanLepage, line 30:
```suggestion
    '';

    maxMemoryMB = helpers.defaultNullOpts.mkInt 2560 ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      The heap memory limit in MiB for `nix` evaluation.
+      Currently it only applies to flake evaluation when `autoEvalInputs` is
+      enabled, and only works for Linux. Other `nix` invocations may be also
+      applied in the future. `null` means no limit.
+      As a reference, `nix flake show --legacy nixpkgs` usually requires
+      about 2GiB memory.
+    '';
+    flake = {
```
GaetanLepage, line 38:
```suggestion
    '';

    flake = {
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+      autoArchive = helpers.defaultNullOpts.mkBool false ''
+        Auto-archiving behavior which may use network.
+         - null: Ask every time.
+         - true: Automatically run `nix flake archive` when necessary.
+         - false: Do not archive. Only load inputs that are already on disk.
+      '';
+      autoEvalInputs = helpers.defaultNullOpts.mkBool false ''
```
GaetanLepage, line 45:
```suggestion
      '';

      autoEvalInputs = helpers.defaultNullOpts.mkBool false ''
```

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,60 @@
...
+        Whether to auto-eval flake inputs.
+        The evaluation result is used to improve completion, but may cost
+        lots of time and/or memory.
+      '';
+      nixpkgsInputName = helpers.defaultNullOpts.mkStr "nixpkgs" ''
```
GaetanLepage, line 50:
```suggestion
      '';

      nixpkgsInputName = helpers.defaultNullOpts.mkStr "nixpkgs" ''
```

tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
```diff
@@ -0,0 +1,35 @@
...
+        enable = true;
+
+        settings = {
+          diagnostics = {
+            ignored = [
+              "unused_binding"
+              "unused_with"
+            ];
+
+            excludedFiles = [ "Cargo.nix" ];
```
GaetanLepage, line 16:
```suggestion
            ];
            excludedFiles = [ "Cargo.nix" ];
```

tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
```diff
@@ -0,0 +1,35 @@
...
+          };
+          formatting = {
+            command = [ "nixfmt" ];
+          };
+          nix = {
+            binary = "nix";
+            maxMemoryMB = 2048;
+
+            flake = {
```
GaetanLepage, line 25:
```suggestion
            maxMemoryMB = 2048;
            flake = {
```

GaetanLepage (changes requested):

plugins/lsp/language-servers/nil-ls-settings.nix
```diff
@@ -0,0 +1,65 @@
...
+{
+  formatting = {
+    command = helpers.defaultNullOpts.mkListOf types.str "null" ''
+      External formatter command (with arguments).
+      It should accepts file content in stdin and print the formatted code into stdout.
+    '';
+  };
+
+  diagnostics = {
+    ignored = helpers.defaultNullOpts.mkListOf types.str [ ] ''
```
GaetanLepage, line 14:
```suggestion
    ignored = helpers.defaultNullOpts.mkListOf types.str "[]" ''
```
Otherwise, this would translate in an empty default in the documentation as `toString [ ]` is `""`.

khaneliman: Ah, makes sense. Forgot that was for documentation.

GaetanLepage (approved):

MattSturgeon (approved):

GaetanLepage:
Thanks @khaneliman !

PR #1558, opened 2024-05-24
plugins/telescope: make `enabledExtensions` public

### LINKED ISSUE #1488: Provide custom telescope extension
I use `nixvim.homeManagerModules.nixvim` in my dotfiles `homeManager`-managed repo.
I'd like to be able to create a telescope extension and enable it.

My idea was to use [mkExtension](https://github.com/nix-community/nixvim/blob/c05fba47c733445098847399b61ddfd81ecbde55/plugins/telescope/extensions/_helpers.nix) (which I didn't find a way to access from my dotfiles repo, but I was happy to copy it in my repo) to create `my-telescope-extension` with `defaultPackage` being a local `.lua` file as the actual telescope extension I want to port to nixvim, and then pass it in `telescope.settings.extension.{my-telescope-extension.enable=true;}`.

But [this](https://github.com/nix-community/nixvim/blob/c05fba47c733445098847399b61ddfd81ecbde55/plugins/telescope/extensions/_helpers.nix#L21) suggests that the extension needs to actually be part of `nixvim` repo, is that correct?
Because if so, it feels like the approach used for extensions is conceptually different than for other configs e.g. mappings, i.e. you can provide your own key mappings, but not your extensions, they must be part of the repo itself.

If the above is true, could I instead override in my `flake.nix` (below) the `nixvim` input to essentially inject `plugins/telescope/extensions/my-telescope-extension.nix` ?
I understand this one is a rather nix question not nixvim per se, but still I thought I can ask here.
Also I know I could probably fork `nixvim` repo, add `my-telescope-extension.nix` file, then supply my fork in the `inputs.nixvim.url`, but I don't like that pattern.

```nix
inputs = {
	# snip

	# can I override this and essentially inject a file inside the plugins/telescope/extensions dir ?
	nixvim = {  url = "github:nix-community/nixvim/main"; inputs.nixpkgs.follows = "nixpks"; };
}
```

MattSturgeon:
Allow users to add telescope extensions we don't support yet. Replace marking the option as "internal" with explicit warnings and suggestions in the description.

I thought about adding an assertion for `(enabledExtensions != []) -> plugins.telescope.enable`, but this seemed excessive and would need to be added via `imports` to bypass how `mkNeovimPlugin` wraps `extraConfig`.

May resolve #1488

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -60,10 +60,19 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
     enabledExtensions = mkOption {
       type = types.listOf types.str;
-      description = "A list of enabled extensions. Don't use this directly";
       default = [ ];
-      internal = true;
-      visible = false;
+      description = ''
+        A list of enabled extensions.
+
+        You should only use this option directly if the Telescope extension you wish to enable is not natively supported
+        by nixvim.
+
+        Most extensions are available as `plugins.telescope.extensions.<name>.enable`, although some plugins that do
+        more than just provide Telescope extensions may use `plugins.<name>.enableTelescope` instead.
+
+        If you add an extension to this list manually, it is your responsibility to ensure the relevant plugin is also
+        added to `extraPackages`.
+      '';
     };
   };
 


tests/test-sources/plugins/telescope/default.nix
```diff
@@ -16,6 +16,7 @@
       };
       keymapsSilent = true;
       highlightTheme = "gruvbox";
+      enabledExtensions = [ "some_extension" ];
```
MattSturgeon, line 19: Can't test this because the extension won't exist...

```
E5113: Error while calling lua chunk: .../start/telescope.nvim/lua/telescope/_extensions/init.lua:10: 'some_extension' extension doesn't exist or isn't installed
```

traxys (approved):

PR #1559, opened 2024-05-24
plugins: standardise telescope extension assertions

MattSturgeon:
Attempt to standardise how telescope extensions are enabled throughout the project.

- Use `mkIf` over `optional`
- Use `enabledExtensions` over manually calling `load_extension`
- Use assertions over warnings when checking telescope is enabled
- Don't implicitly enable telescope

The commit message could probably be worded better, or split into multiple commits?

Some of this could be refactored into helper functions to improve maintenance?

Is the assertion for `plugins.telescope.extensions.*.enable` overkill? Will that assertion even run when `plugins.telescope.enable = false`?


plugins/git/git-worktree.nix
```diff
@@ -71,13 +74,10 @@ in
 
       extraPackages = [ pkgs.git ];
 
-      extraConfigLua =
-        let
-          telescopeCfg = ''require("telescope").load_extension("git_worktree")'';
-        in
-        ''
-          require('git-worktree').setup(${helpers.toLuaObject setupOptions})
-          ${if cfg.enableTelescope then telescopeCfg else ""}
```
MattSturgeon, line 80: I assume this wasn't done to ensure correct setup order? If so this PR may break that.

plugins/utils/harpoon.nix
```diff
@@ -182,20 +182,20 @@ in
...
 
       extraPlugins = [ cfg.package ];
 
-      extraConfigLua =
-        let
-          telescopeCfg = ''require("telescope").load_extension("harpoon")'';
-        in
-        ''
-          require('harpoon').setup(${helpers.toLuaObject setupOptions})
-          ${if cfg.enableTelescope then telescopeCfg else ""}
```
MattSturgeon, line 197: Same

plugins/utils/yanky.nix
```diff
@@ -188,10 +188,16 @@ in
...
+        {
+          assertion = cfg.picker.telescope.enable -> config.plugins.telescope.enable;
+          message = ''
+            Nixvim: You have enabled the `telescope` integration with yanky.
+            However, you have not enabled the `telescope` plugin itself (`plugins.telescope.enable = true`).
+          '';
+        }
       ];
 
-      plugins.telescope.enable = true;
```
MattSturgeon, line 193: Should transitioning from implicit-enable to assertion have a deprecation/warning period?

plugins/utils/project-nvim.nix
```diff
@@ -90,9 +90,15 @@ in
   };
 
   config = mkIf cfg.enable {
-    warnings = optional (cfg.enableTelescope && (!config.plugins.telescope.enable)) ''
```
MattSturgeon, line 93: Is directly updating from warning to assertion ok, or should there be a transition period?

khaneliman:
I came across an auto enable behavior in[ todo-comments.nvim](https://github.com/nix-community/nixvim/blob/4852f94f8ccae551514df0092a077014bafb95ca/plugins/utils/todo-comments.nix#L287) during testing a recent plugin.

plugins/utils/project-nvim.nix
khaneliman: I feel like this could easily be upgraded, now. We added the warning about 5 months ago so its had a good warning period.

PR #1565, opened 2024-05-25
modules/keymaps: add "buffer" option to mapConfigOptions

iSma:
This allows to define buffer-local keymaps, which is specially useful when combined with `keymapsOnEvents`.

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -22,6 +22,8 @@ rec {
     remap = nixvimOptions.defaultNullOpts.mkBool false "Make the mapping recursive. Inverses \"noremap\"";
 
     desc = nixvimOptions.mkNullOrOption types.str "A textual description of this keybind, to be shown in which-key, if you have it.";
+
+    buffer = nixvimOptions.defaultNullOpts.mkBool false "Make the mapping buffer-local. Equivalent to adding <buffer> to a map.";
   };
 
   modes = {


GaetanLepage:
Is this supposed to work ?
`:h nvim_set_keymap` states that `      • {opts}  Optional parameters map: Accepts all |:map-arguments| as keys
                except <buffer>, values are booleans (default false). Also:

GaetanLepage:
And `To set a buffer-local mapping, use |nvim_buf_set_keymap()|.`

iSma:
I tested the change with `vim.keymap.set`, where `:h vim.keymap.set()` states:
```
Also accepts:
• {buffer}? (`integer|boolean`) Creates buffer-local mapping,
  `0` or `true` for current buffer.
```

A quick grep through the repo found no usages of `nvim_set_keymap`, only `vim.keymap.set`, but I could have missed something.

(I've thought about typing the option `either bool int`, but I don't see a use-case where setting a specific buffer number would be useful.)

GaetanLepage:
> A quick grep through the repo found no usages of `nvim_set_keymap`, only `vim.keymap.set`, but I could have missed something.

Oh, yes, you are completely right. I had forgotten about this. We used to use `nvim_set_keymap`.

> (I've thought about typing the option `either bool int`, but I don't see a use-case where setting a specific buffer number would be useful.)

True, I don't see whether that would be useful...
Any opinion @traxys @MattSturgeon ?

lib/keymap-helpers.nix
```diff
@@ -22,6 +22,8 @@ rec {
     remap = nixvimOptions.defaultNullOpts.mkBool false "Make the mapping recursive. Inverses \"noremap\"";
 
     desc = nixvimOptions.mkNullOrOption types.str "A textual description of this keybind, to be shown in which-key, if you have it.";
+
+    buffer = nixvimOptions.defaultNullOpts.mkBool false "Make the mapping buffer-local. Equivalent to adding <buffer> to a map.";
```
GaetanLepage, line 26:
```suggestion
    buffer = nixvimOptions.defaultNullOpts.mkBool false "Make the mapping buffer-local. Equivalent to adding `<buffer>` to a map.";
```

MattSturgeon:
> I've thought about typing the option `either bool int`, but I don't see a use-case where setting a specific buffer number would be useful.

I assume they accept int for backwards compatibility reasons? Or do they allow specifying _which_ buffer to create the mapping in using a buffnr?

I personally don't see the use-case, but it'd be better to match upstream typing _if_ there's actual functionality there. Otherwise I agree, keep it simple.

GaetanLepage:
> I assume they accept int for backwards compatibility reasons? Or do they allow specifying _which_ buffer to create the mapping in using a buffnr?

For me it is the latter, i.e. to give a buffer id.

> I personally don't see the use-case, but it'd be better to match upstream typing _if_ there's actual functionality there. Otherwise I agree, keep it simple.

Although it makes no sense to me to set a binding for buffer x _at nix build time_, I don't mind if we type it the same way as upstream. Maybe @traxys will have a stronger opinion on that.

traxys:
I'd rather not expose the "int", people would be able to access it using rawLua, and it does not make sense in nixvim I think

lib/keymap-helpers.nix
iSma: There are no backticks around <silent>, <script>, etc. in the strings above

MattSturgeon: Fair point. There probably should be, but that should be done in a separate commit.

Personally I don't mind if that's tackled in this PR or not. Otherwise, I'm happy for this option to be consistent with the others and not use backticks for now.

GaetanLepage: Either way

GaetanLepage (approved):

PR #1567, opened 2024-05-26
modules/keymaps: deprecate `lua` option

MattSturgeon:
This PR deprecates the `lua` keymap submodule option, in favour of using a "raw string" directly on the `action` option.

Since we're not ready to remove it outright just yet the existing option declaration needs to stay, but it can be marked `visible = false`.

A warning will be displayed if the option is defined anyware in a nixvim config, even if explicitly setting `lua = false`.

This warning is printed when using this branch with my config:

```
trace: warning: Nixvim (keymaps): the `lua` keymap option is deprecated.

This option will be removed in 24.11. You should use a "raw" `action` instead;
e.g. `action.__raw = "<lua code>"` or `action = helpers.mkRaw "<lua code>"`.

Found 1 use in /nix/store/7lsacy6wira1l2y6xpxq1nxcigrkxaqp-source/nvim/config/main.nix

```

This PR builds on #1004 and #1010

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -2,6 +2,7 @@
   lib,
   helpers,
   config,
+  options,
   ...
 }:
 with lib;
@@ -26,13 +27,11 @@ with lib;
         "InsertEnter" = [
           {
             key = "<C-y>";
-            action = ''require("cmp").mapping.confirm()'';
-            lua = true;
+            action.__raw = ''require("cmp").mapping.confirm()'';
           }
           {
             key = "<C-n>";
-            action = ''require("cmp").mapping.select_next_item()'';
-            lua = true;
+            action.__raw = ''require("cmp").mapping.select_next_item()'';
           }
         ];
       };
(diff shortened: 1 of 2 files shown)


modules/keymaps.nix
```diff
@@ -52,6 +51,36 @@ with lib;
       };
     in
     {
+      # Deprecate `lua` keymap option
+      # TODO upgrade to an assertion (removal notice) in 24.11
+      # TODO remove entirely in 25.05?
```
MattSturgeon, line 56: Not sure what the best deprecation strategy is, open to suggestions and alternative TODO wording.

I'm working on the assumption we'll want to give users upgrading from one stable nixos version to another at least one chance to see a deprecation warning without their config breaking.

GaetanLepage: If we merge this right now, it will get into 24.05.
So, we should be able to remove it definitely in 24.11.

GaetanLepage (approved):
Very solid patch.
Thanks for taking the time to clearly state the stakes of such a change. This is greatly appreciated.

modules/keymaps.nix
GaetanLepage: We can even backport the "assertion" in 24.05 in a few months.

MattSturgeon: > So, we should be able to remove it definitely in 24.11.

When we remove the option, should we keep a custom assertion message for a while or just rely on the module system's "no such option" assertion?

> We can even backport the "assertion" in 24.05 in a few months.

Should I add some datestamps to the TODO comments?

traxys (approved):

PR #1569, opened 2024-05-26
actions: Update both stable & unstable flake.lock

traxys:
When triggering manually we can update:
- both branches
- Only the main branch (unstable)
- Only the chosen stable branch

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,11 +1,30 @@
 name: update-flake-lock
 on:
   workflow_dispatch: # allows manual triggering
+    inputs:
+      branch:
+        description: "Branch to update"
+        type: choice
+        options:
+          - 'stable & unstable'
+          - 'main'
+          - 'nixos-23.11'
   schedule:
     - cron: "0 12 * * SAT" # runs weekly on Saturday at noon
 
 jobs:
   lockfile:
+    strategy:
+      matrix:
+        # This allows to update both stable & unstable branches, but not both when triggered
+        # manually
+        branch: ['main', 'nixos-23.11']
+        selectedBranch: ["${{ inputs.branch }}"]
+        exclude:
+          - selectedBranch: main
+            branch: 'nixos-23.11'
+          - selectedBranch: 'nixos-23.11'
+            branch: main
     name: Update the flake inputs and generate options
     runs-on: ubuntu-latest
     timeout-minutes: 40
@@ -14,6 +33,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          ref: '${{ matrix.branch }}'
 
       - name: Install Nix
         uses: cachix/install-nix-action@v26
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Is there a reason we need an input? `workflow_dispatch` already has a ref selection built in.

For the rest API [`ref` sits alongside `inputs`](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#running-a-workflow-using-the-rest-api), for the `gh` CLI you can use `--ref`, for the web UI there should be a drop-down box.

.github/workflows/update.yml
```diff
@@ -25,5 +45,6 @@ jobs:
         id: update
         uses: DeterminateSystems/update-flake-lock@v21
         with:
-          pr-title: "Update flake.lock"
+          pr-title: "Update flake.lock (${{ matrix.branch }})"
           token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          branch: "update_flake_lock_action-${{ matrix.branch }}"
```
MattSturgeon, line 50: Does the update-flake-lock action use `github.ref` or the _actually_ checked-out HEAD ref as it's target/base branch when opening the PR?

MattSturgeon: Upstream [claims](https://github.com/DeterminateSystems/update-flake-lock/blob/222f041780a88f7f379bd7a3c3327585e435db5b/action.yml#L17) it's the latter, but I haven't tested or looked at the implementation to be sure.

traxys:
> Is there a reason we need a `branch` input? `workflow_dispatch` already has a ref selection built in.
> 


I believe this is for the _workflow_ itself, not the branch it targets I believe. The main reason for the branch trigger is to allow to manually update only a single branch, instead of both stable & unstable branches

.github/workflows/update.yml
traxys: I tested, it uses the currently checkout out ref

traxys:
> Also, is there a reason not to automatically run the CI for the stable branch(s) too, even if less frequently?

I believe this workflow will run the update every saturday at noon for both the stable & unstable branch, though I have yet to test it

MattSturgeon:
> believe this is for the workflow itself, not the branch it targets I believe

Correct, though the checkout action _defaults_ to that workflow ref.

If you want to be able to matrix **multiple** branches that's not useful, though.

.github/workflows/update.yml
```diff
@@ -1,11 +1,30 @@
...
+          - 'main'
+          - 'nixos-23.11'
   schedule:
     - cron: "0 12 * * SAT" # runs weekly on Saturday at noon
 
 jobs:
   lockfile:
+    strategy:
+      matrix:
+        # This allows to update both stable & unstable branches, but not both when triggered
```
MattSturgeon, line 19: Excluding this way seems a bit backwards, and won't scale if we ever want to add a third branch.

But it's probably not worth thinking up a better solution until we actually decide to do that.

MattSturgeon (approved):

.github/workflows/update.yml
traxys: Yeah I agree, but that seems a bit of a limitation of github actions

MattSturgeon: Can't look properly right now, but I think it'd be possible to use `${{ }}` expressions and `&&` `||` operators together to form a ternary.

PR #1576, opened 2024-05-27
plugin/lsp-status: init

b3nb5n:
added plugin options for [lsp-status.nvim](https://github.com/nvim-lua/lsp-status.nvim)

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -61,4 +61,10 @@
     name = "Nate Smith";
     keys = [ { fingerprint = "DE6F E362 7DF7 EEA0 79CB  38C1 F208 9547 768B 6CCA"; } ];
   };
+  b3nb5n = {
+    email = "benbaldwin000@gmail.com";
+    github = "b3nb5n";
+    githubId = 67534814;
+    name = "Ben";
+  };
 }
--- a/plugins/lsp/lsp-status.nix
+++ b/plugins/lsp/lsp-status.nix
@@ -1,0 +1,88 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lsp-status";
+  originalName = "lsp-status.nvim";
+  defaultPackage = pkgs.vimPlugins.lsp-status-nvim;
+  maintainers = [ helpers.maintainers.b3nb5n ];
+
+  settingsOptions =
+    let
+      mkIndicatorOption =
+        default:
+        helpers.defaultNullOpts.mkStr default ''
+          The string to show as diagnostics.
+          If you don't have Nerd/Awesome Fonts you can replace defaults with ASCII chars.
+        '';
+    in
+    {
+      kind_labels = helpers.defaultNullOpts.mkAttrsOf types.str "{}" ''
+        An optional map from LSP symbol kinds to label symbols. Used to decorate the current function name.
+      '';
+
+      select_symbol = helpers.defaultNullOpts.mkStr "" ''
+        An optional handler of the form `function(cursor_pos, document_symbol)` that should return
+        `true` if `document_symbol` (a `DocumentSymbol`) should be accepted as the symbol currently
+        containing the cursor.
+      '';
+
+      current_function = helpers.defaultNullOpts.mkBool true ''
+        True if the current function should be updated and displayed in the default statusline component.
+      '';
+
+      show_filename = helpers.defaultNullOpts.mkBool true ''
+        True if the current function should be updated and displayed in the default statusline component.
+      '';
+
+      indicator_ok = mkIndicatorOption "";
+      indicator_errors = mkIndicatorOption "";
+      indicator_warnings = mkIndicatorOption "";
+      indicator_info = mkIndicatorOption "🛈";
+      indicator_hint = mkIndicatorOption "❗";
+
+      indicator_separator = helpers.defaultNullOpts.mkStr " " ''
+        A string which goes between each diagnostic group symbol and its count.
+      '';
+
+      component_separator = helpers.defaultNullOpts.mkStr " " ''
+        A string which goes between each "chunk" of the statusline component (i.e. different diagnostic groups, messages).
+      '';
+
+      diagnostics = helpers.defaultNullOpts.mkBool true ''
+        If false, the default statusline component does not display LSP diagnostics.
+      '';
+    };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = config.plugins.lsp.enable;
+        message = ''
+          Nixvim (plugins.lsp-status): `plugins.lsp` must be enabled to use lsp-status
+        '';
+      }
+    ];
+
+    plugins.lsp = {
+      preConfig = ''
+        do
+          local lsp_status = require('lsp-status')
+          lsp_status.config(${helpers.toLuaObject cfg.settings})
+          lsp_status.register_progress()
+        end
+      '';
+
+      # the lsp status plugin needs to hook into the on attach and capabilities
+      # fields of the lsp setup call to track the progress of initializing the lsp
+      onAttach = "require('lsp-status').on_attach(client)";
+      capabilities = "capabilities = vim.tbl_extend('keep', capabilities or {}, require('lsp-status').capabilities)";
+    };
+  };
+}
--- a/tests/test-sources/plugins/lsp/lsp-status.nix
+++ b/tests/test-sources/plugins/lsp/lsp-status.nix
@@ -1,0 +1,32 @@
+{
+  empty = {
+    plugins = {
+      lsp.enable = true;
+      lsp-status.enable = true;
+    };
+  };
+
+  defaults = {
+    plugins = {
+      lsp.enable = true;
+      lsp-status = {
+        enable = true;
+
+        settings = {
+          kind_labels = { };
+          select_symbol = "";
+          current_function = true;
+          show_filename = true;
+          indicator_ok = "OK";
+          indicator_errors = "E";
+          indicator_warnings = "W";
+          indicator_info = "i";
+          indicator_hint = "?";
+          indicator_separator = " ";
+          component_separator = " ";
+          diagnostics = true;
+        };
+      };
+    };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -98,6 +98,7 @@
     ./lsp/lspsaga.nix
     ./lsp/lsp-format.nix
     ./lsp/lsp-lines.nix
+    ./lsp/lsp-status.nix
     ./lsp/nvim-lightbulb.nix
     ./lsp/trouble.nix
     ./lsp/wtf.nix


plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lsp-status";
+  originalName = "lsp-status.nvim";
+  url = "https://github.com/nvim-lua/lsp-status.nvim";
```
MattSturgeon, line 11: This can be removed, since the nixpkgs package has the URL defined.

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+  defaultPackage = pkgs.vimPlugins.lsp-status-nvim;
+  maintainers = [ helpers.maintainers.b3nb5n ];
+  settingsOptions = with lib; {
```
MattSturgeon, line 14: `with lib;` can be applied to the whole file, i.e. just before the call to `mkNeovimPlugin` (fairly common practice in nixvim).

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+    plugins.lsp-status = rec {
+      enable = mkEnableOption "lsp-status.nvim";
```
MattSturgeon, line 16: Settings options are exclusively for config passed to the `setup` lua function. You don't need `enable` and `package` here, `mkNeovimPlugin` will create those options automatically _outside_ of the `settings` option.

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+
+      package = helpers.mkPluginPackageOption "lsp-status.nvim" pkgs.vimPlugins.lsp-status-nvim;
+
+      kind_labels = mkOption {
+        type = types.attrsOf types.str;
```
MattSturgeon, line 21: Use `      helpers.defaultNullOpts.mkAttrsOf types.str` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = { };
+        description = "An optional map from LSP symbol kinds to label symbols. Used to decorate the current function name.";
+      };
+
+      select_symbol = mkOption {
+        type = types.nullOr types.str;
```
MattSturgeon, line 27: Use `helpers.defaultNullOpts.mkStr` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = null;
+        description = "An optional handler of the form function(cursor_pos, document_symbol) that should return true if document_symbol (a DocumentSymbol) should be accepted as the symbol currently containing the cursor.";
+      };
+
+      current_function = mkOption {
+        type = types.bool;
```
MattSturgeon, line 33: Use `helpers.defaultNullOpts.mkBool` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = true;
+        description = "true if the current function should be updated and displayed in the default statusline component.";
+      };
+
+      show_filename = mkOption {
+        type = types.bool;
```
MattSturgeon, line 39: Use `helpers.defaultNullOpts.mkBool` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = true;
+        description = "true if the current function should be updated and displayed in the default statusline component.";
+      };
+
+      indicator_ok = mkOption {
+        type = types.nullOr types.str;
```
MattSturgeon, line 45: Use `helpers.defaultNullOpts.mkStr` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = null;
+        description = "string to show as diagnostics. If you don't have Nerd/Awesome Fonts you can replace defaults with ASCII chars.";
+      };
+
+      indicator_errors = indicator_ok;
```
MattSturgeon, line 50: Maybe declare common/shared descriptions in a `let` block instead of using `rec`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+      indicator_warnings = indicator_ok;
+      indicator_info = indicator_ok;
+      indicator_hint = indicator_ok;
+
+      indicator_separator = mkOption {
+        type = types.str;
```
MattSturgeon, line 56: Use `helpers.defaultNullOpts.mkStr` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = " ";
+        description = "a string which goes between each diagnostic group symbol and its count.";
+      };
+
+      component_separator = mkOption {
```
MattSturgeon, line 61: Use `helpers.defaultNullOpts.mkStr` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        type = types.str;
+        default = " ";
+        description = ''a string which goes between each "chunk" of the statusline component (i.e. different diagnostic groups, messages).'';
+      };
+
+      diagnostics = mkOption {
+        type = types.bool;
```
MattSturgeon, line 68: Use `helpers.defaultNullOpts.mkBool` instead of `mkOption`

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+        default = true;
+        description = "If false, the default statusline component does not display LSP diagnostics.";
+      };
+    };
+  };
+
+  extraConfig = cfg: {
+    extraConfigLua = "require('lsp-status').register_progress()";
```
MattSturgeon, line 76: If this needs to be done before/after calling `setup`, then `callSetup = false` should be passed to `mkNeovimPlugin` and `setup` should be called manually here in `extraConfigLua`

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,5 @@
+_: {
+  empty = {
+    plugins.lsp-status.enable = true;
+  };
```
MattSturgeon, line 4: Adding a `defaults` test case would be good

MattSturgeon (changes requested):
Most of this feedback comes down to [this part](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md#declaring-plugin-options) of `CONTRIBUTING.md`.

Also, CI appears to be failing for an unrelated reason due to nixos-unstable transitioning to 24.11. Rebasing on `main` again _might_ help.

lib/maintainers.nix
```diff
@@ -54,4 +54,10 @@
     githubId = 66408983;
     name = "svl";
   };
+  b3nb5n = {
+    discord = "b3nb5n";
+    github = "b3nb5n";
+    githubId = 67534814;
+    name = "b3nb5n";
```
MattSturgeon, line 61: Up to you, but usually `name` would be a "real" name, i.e. "Ben".

Also, you should include an email; the email you use in git commits should be fine, since that's public anyway.

b3nb5n:
updated everything you commented on. this plugin uses "config" instead of "setup" just to double check, is there some internal setup function that also needs to be called?

MattSturgeon:
> updated everything you commented on

Great, thanks! I'll re-review shortly

> this plugin uses "config" instead of "setup" just to double check, is there some internal setup function that also needs to be called?

No, "setup" is just so ubiquitous that `mkNeovimPlugin` assumes the plugin uses it unless you pass `callSetup = false`

GaetanLepage:
Please, squash all your changes in a single commit `plugins/lsp-status: init`.

GaetanLepage:
Well, 2 commits:
- `maintainers: add b3nb5n`
- `plugins/lsp-status: init`

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,23 @@
...
+      current_function = true;
+      show_filename = true;
+      indicator_ok = "OK";
+      indicator_errors = "E";
+      indicator_warnings = "W";
+      indicator_info = "i";
+      indicator_hint = "?";
+      indicator_separator = " ";
+      component_separator = " ";
+      diagnostics = true;
```
MattSturgeon, line 20:
```suggestion
      settings = {
        kind_labels = { };
        select_symbol = "";
        current_function = true;
        show_filename = true;
        indicator_ok = "OK";
        indicator_errors = "E";
        indicator_warnings = "W";
        indicator_info = "i";
        indicator_hint = "?";
        indicator_separator = " ";
        component_separator = " ";
        diagnostics = true;
      };
```

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,23 @@
+_: {
```
GaetanLepage, line 1:
```suggestion
{
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,57 @@
...
+      diagnostics = helpers.defaultNullOpts.mkBool true "If false, the default statusline component does not display LSP diagnostics.";
+    };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    extraConfigLuaPre = ''
+      do
+        local lsp_status = require('lsp-status')
+        lsp_status.config(${helpers.toLuaObject cfg})
+        require('lsp-status').register_progress()
```
GaetanLepage, line 48:
```suggestion
        lsp_status.register_progress()
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,57 @@
...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lsp-status";
+  originalName = "lsp-status.nvim";
+  defaultPackage = pkgs.vimPlugins.lsp-status-nvim;
+  maintainers = [ helpers.maintainers.b3nb5n ];
+  settingsOptions =
+    let
+      indicator_opt = helpers.defaultNullOpts.mkStr "" "string to show as diagnostics. If you don't have Nerd/Awesome Fonts you can replace defaults with ASCII chars.";
```
GaetanLepage, line 16:
```suggestion
      indicator_opt = helpers.defaultNullOpts.mkStr "" ''
        string to show as diagnostics.
        If you don't have Nerd/Awesome Fonts you can replace defaults with ASCII chars.
      '';
```

Start each sentence with a capital letter.
Use 2 single quotes for descriptions, with a line break.

GaetanLepage (changes requested):

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,57 @@
...
+    extraConfigLuaPre = ''
+      do
+        local lsp_status = require('lsp-status')
+        lsp_status.config(${helpers.toLuaObject cfg.settings})
+        lsp_status.register_progress()
+      end
+    '';
+
+    plugins.lsp = {
+      onAttach = "require('lsp-status').on_attach(client)";
```
MattSturgeon, line 56: Is this how the plugin's `on_attach` is supposed to be used?

This is calling lsp-status's `on_attach` from lsp-config's `onAttach`.

Hard to tell from upstream's readme, but I get the impression the plugin handles calling `on_attach`

plugins/lsp/lsp-status.nix
MattSturgeon: Also, don't be afraid of longer multi-line descriptions. We often include examples or longer explanations this way.

Markdown syntax and code blocks are also supported.

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,60 @@
...
+      capabilities = "capabilities = vim.tbl_extend('keep', capabilities or {}, require('lsp-status').capabilities)";
```
MattSturgeon, line 57: As with `on_attach`, this is confusing me slightly. If this is correct could you add comments explaining what is being done and why?

plugins/lsp/lsp-status.nix
b3nb5n: yep, towards the bottom of that readme theres a complete example where they show passing the on attach function directly to the lsp setup call
```lua
lspconfig.rust_analyzer.setup({
  on_attach = lsp_status.on_attach,
  capabilities = lsp_status.capabilities
})
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,60 @@
...
+    let
+      indicator_opt = helpers.defaultNullOpts.mkStr "" ''
+        string to show as diagnostics.
+        If you don't have Nerd/Awesome Fonts you can replace defaults with ASCII chars.
+      '';
+    in
+    {
+      kind_labels =
+        helpers.defaultNullOpts.mkAttrsOf types.str ""
+          "An optional map from LSP symbol kinds to label symbols. Used to decorate the current function name.";
```
MattSturgeon, line 24: I think @GaetanLepage meant to use multiline strings for all option descriptions, not just `indicator_opt`.

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,25 @@
+{
+  empty = {
+    plugins.lsp-status.enable = true;
+  };
+
+  defaults = {
+    plugins.lsp-status = {
+      enable = true;
+      settings = {
```
GaetanLepage, line 9:
```suggestion
      enable = true;

      settings = {
```

plugins/lsp/lsp-status.nix
MattSturgeon: Thanks for explaining, I glossed over that bit!

Maybe a short comment along the lines of "lsp-status requires us to call `on_attach` and set `capabilities` in lspconfig". Could probably be worded better...

plugins/lsp/lsp-status.nix
b3nb5n: updated all the description strings

plugins/lsp/lsp-status.nix
b3nb5n: for sure, added a comment explaining a bit

GaetanLepage:
@MattSturgeon we have to ensure that `plugins.lsp` is enabled when using this plugin.
Not sure on whether a warning or assertion is better here. Maybe warning is good enough.

MattSturgeon:
> we have to ensure that `plugins.lsp` is enabled when using this plugin. Not sure on whether a warning or assertion is better here. Maybe warning is good enough.

I'm happy either way. TBH I think we still need to standardize across the board when we use assertions over warnings?

I'd lean towards an assertion when there's no conceivable way the config would work as intended unless the condition is met. I think warnings are more for when something is deprecated or the "usually" way of doing something isn't being followed?

I don't think we should get hung up discussing that on this PR though.

plugins/lsp/lsp-status.nix
MattSturgeon: Sorry to nit-picked, but we usually prefer to format the description as follows:

```nix
helpers.defaultNullOpts.mkAttrsOf types.str "" ''
  An optional map from LSP symbol kinds to label symbols. Used to decorate the current function name.
'';
```

i.e. with the description text on a separate line to the quotes.

b3nb5n:
added it as an assertion, just copied the existing pattern

MattSturgeon:
> added it as an assertion, just copied the existing pattern

You'll have to update the tests to also enable `plugins.lsp`

I think [this](https://github.com/nix-community/nixvim/pull/1576#discussion_r1616413327) and other minor formatting stuff is all that remains now?

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,69 @@
...
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = cfg.plugins.lsp.enable;
+        message = ''
+          Nixvim (plugins.lsp-status): plugins.lsp.enable must be true to use lsp-status
+        '';
+      }
+    ];
+    extraConfigLuaPre = ''
```
MattSturgeon, line 55:
```suggestion

    extraConfigLuaPre = ''
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,69 @@
...
+      diagnostics = helpers.defaultNullOpts.mkBool true ''If false, the default statusline component does not display LSP diagnostics.'';
+    };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = cfg.plugins.lsp.enable;
+        message = ''
+          Nixvim (plugins.lsp-status): plugins.lsp.enable must be true to use lsp-status
```
MattSturgeon, line 51:
```suggestion
          Nixvim (plugins.lsp-status): `plugins.lsp` must be enabled to use lsp-status
```

Or just 

```suggestion
          Nixvim (plugins.lsp-status): `plugins.lsp.enable` must be true to use lsp-status
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,69 @@
...
+    extraConfigLuaPre = ''
+      do
+        local lsp_status = require('lsp-status')
+        lsp_status.config(${helpers.toLuaObject cfg.settings})
+        lsp_status.register_progress()
+      end
+    '';
+
+    plugins.lsp = {
+      # the lsp status plugin needs to hook into the on attach and capabilities fields of the lsp setup call to track the progress of initializing the lsp
```
MattSturgeon, line 64: Hard to tell on my phone screen, but this looks like it's read better over multiple lines (I think we normally aim for 120 char long lines?)

```suggestion
      # the lsp status plugin needs to hook into the on attach and capabilities
      # fields of the lsp setup call to track the progress of initializing the lsp
```

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,26 @@
+{
+  empty = {
```
MattSturgeon, line 2:
```suggestion
  empty = {
    plugins.lsp.enable = true;
    
```

(Same for `defaults` test)

MattSturgeon (changes requested):
Looking good

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,82 @@
...
+      diagnostics = helpers.defaultNullOpts.mkBool true ''
+        If false, the default statusline component does not display LSP diagnostics.
+      '';
+    };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = cfg.plugins.lsp.enable;
```
MattSturgeon, line 61:
```suggestion
        assertion = config.plugins.lsp.enable;
```

`cfg` is _this plugin's_ part of the config

MattSturgeon:
I haven't looked at what we do in other lsp-related plugins, but do we need `extraConfigLuaPre` here or will `extraConfigLua` also work?

b3nb5n:
I just used pre because the docs for this plugin said to call register progress before calling the lsp setup functions

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lsp-status";
+  originalName = "lsp-status.nvim";
+  defaultPackage = pkgs.vimPlugins.lsp-status-nvim;
+  maintainers = [ helpers.maintainers.b3nb5n ];
+  settingsOptions =
+    let
+      indicator_opt = helpers.defaultNullOpts.mkStr "" ''
+        string to show as diagnostics.
```
MattSturgeon, line 17:
```suggestion
        The string to show as diagnostics.
```

If you wanted, you could do something along the lines of

```nix
mkIndicatorOpt = name: helpers.defaultNullOpts.mkStr "" ''
  The string to show for ${name} diagnostics.
'';
```

```nix
indicator_ok = mkIndicatorOpt "ok";
```

MattSturgeon: Actually, looking at your defaults test, you'll need to do something like that in order to correctly set the plugin-default for each option:

```nix
mkIndicatorOpt = default: name: helpers.defaultNullOpts.mkStr default ''
  The string to show for ${name} diagnostics.
'';
```

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,83 @@
...
+    assertions = [
+      {
+        assertion = config.plugins.lsp.enable;
+        message = ''
+          Nixvim (plugins.lsp-status): `plugins.lsp` must be enabled to use lsp-status
+        '';
+      }
+    ];
+
+    extraConfigLuaPre = ''
```
MattSturgeon, line 68: > I just used pre because the docs for this plugin said to call register progress before calling the lsp setup functions

```suggestion
    plugins.lsp.preConfig = ''
```

plugins/lsp/lsp-status.nix
MattSturgeon: Marking unresolved because the second comment regarding plugin defaults wasn't addressed

b3nb5n: the defaults should be unset because this option is used to override the defaults set by the plugin. is there a nullable string option that you use?

MattSturgeon: The "default" passed to `defaultNullOpts` functions is not the nix option default (that's always`null`), rather it is the "plugin default" which is used for our documentation only.

plugins/lsp/lsp-status.nix
MattSturgeon: I saw you pushed & reverted this. Guessing you had a "defined twice" error since `plugin.lsp` is defined below.

You could just move the `preConfig` definition to be within the `plugins.lsp` attrset

plugins/lsp/lsp-status.nix
b3nb5n: updated to use the nerdfont icons for the defaults

plugins/lsp/lsp-status.nix
b3nb5n: nope, mustve just been weird timing with pushing/pulling. I added it to the main definition

MattSturgeon:
https://github.com/nix-community/nixvim/pull/1576#discussion_r1616431411

> updated to use the nerdfont icons for the defaults

Hm, I hadn't considered that the plugin default might be something that can't be rendered on the documentation website.

@GaetanLepage do you think we should set those defaults to something like `"<nerdfont-icon-${iconName}>"`? Unless we wanted to add a nerdfont to the docs via CSS?

MattSturgeon (approved):
Although there's one unresolved comment to discuss https://github.com/nix-community/nixvim/pull/1576#issuecomment-2134149270, I'm happy with these changes.

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,28 @@
...
+    plugins.lsp.enable = true;
+    plugins.lsp-status.enable = true;
```
GaetanLepage, line 4:
```suggestion
    plugins = {
      lsp.enable = true;
      lsp-status.enable = true;
    };
```

tests/test-sources/plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,28 @@
...
+        indicator_ok = "OK";
+        indicator_errors = "E";
+        indicator_warnings = "W";
+        indicator_info = "i";
+        indicator_hint = "?";
+        indicator_separator = " ";
+        component_separator = " ";
+        diagnostics = true;
+      };
+    };
```
GaetanLepage, line 26:
```suggestion
    plugins = {
      lsp.enable = true;
      plugins.lsp-status = {
        enable = true;
  
        settings = {
          kind_labels = { };
          select_symbol = "";
          current_function = true;
          show_filename = true;
          indicator_ok = "OK";
          indicator_errors = "E";
          indicator_warnings = "W";
          indicator_info = "i";
          indicator_hint = "?";
          indicator_separator = " ";
          component_separator = " ";
          diagnostics = true;
        };
      };
    };
```

GaetanLepage (changes requested):

plugins/lsp/lsp-status.nix
```diff
@@ -0,0 +1,85 @@
...
+        '';
+    in
+    {
+      kind_labels = helpers.defaultNullOpts.mkAttrsOf types.str "{}" ''
+        An optional map from LSP symbol kinds to label symbols. Used to decorate the current function name.
+      '';
+
+      select_symbol = helpers.defaultNullOpts.mkStr "" ''
+            An optional handler of the form function(cursor_pos, document_symbol) that should return true if document_symbol (a DocumentSymbol) should be accepted as the symbol currently containing the cursor.
+        	'';
```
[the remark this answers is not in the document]
MattSturgeon: Actually, we also need to wrap any code snippets in backticks (especially snippets containing underscores, since that's markdown syntax for "italic"):

```suggestion
      select_symbol = helpers.defaultNullOpts.mkStr "" ''
            An optional handler of the form `function(cursor_pos, document_symbol)` that should return
            `true` if `document_symbol` (a `DocumentSymbol`) should be accepted as the symbol currently
            containing the cursor.
       '';
```

GaetanLepage (approved):
LGTM !

PR #1582, opened 2024-05-28
Add the Modus colorscheme

nwjsmith:
Add the Modus colorscheme

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -54,4 +54,11 @@
     githubId = 66408983;
     name = "svl";
   };
+  nwjsmith = {
+    email = "nate@theinternate.com";
+    github = "nwjsmith";
+    githubId = 1348;
+    name = "Nate Smith";
+    keys = [ { fingerprint = "DE6F E362 7DF7 EEA0 79CB  38C1 F208 9547 768B 6CCA"; } ];
+  };
 }
--- a/plugins/colorschemes/modus.nix
+++ b/plugins/colorschemes/modus.nix
@@ -1,0 +1,107 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "modus";
+  luaName = "modus-themes";
+  originalName = "modus-themes.nvim";
+  defaultPackage = pkgs.vimPlugins.modus-themes-nvim;
+  isColorscheme = true;
+
+  maintainers = [ helpers.maintainers.nwjsmith ];
+
+  settingsOptions = {
+    style =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "modus_operandi"
+          "modus_vivendi"
+        ]
+        ''
+          The theme comes in a light `modus_operandi` style and a dark `modus_vivendi` style.
+        '';
+
+    variant =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "default"
+          "tinted"
+          "deuteranopia"
+          "tritanopia"
+        ]
+        ''
+          Styles come in four variants:
+
+          - `default` is the plugin's main style designed to cover a broad range of needs.
+          - `tinted` tones down intensity and provides more color variety.
+          - `deuteranopia` is optimized for users with red-green color deficiency.
+          - `tritanopia` is optimized for users with blue-yellow color deficiency.
+        '';
+
+    transparent = helpers.defaultNullOpts.mkBool false ''
+      Disable setting the background color.
+    '';
+
+    dim_inactive = helpers.defaultNullOpts.mkBool false ''
+      Dims inactive windows.
+    '';
+
+    styles = {
+      comments = helpers.defaultNullOpts.mkHighlight "{italic = true;}" "" ''
+        Define comments highlight properties.
+      '';
+
+      keywords = helpers.defaultNullOpts.mkHighlight "{italic = true;}" "" ''
+        Define keywords highlight properties.
+      '';
+
+      functions = helpers.defaultNullOpts.mkHighlight "{}" "" ''
+        Define functions highlight properties.
+      '';
+
+      variables = helpers.defaultNullOpts.mkHighlight "{}" "" ''
+        Define variables highlight properties.
+      '';
+    };
+
+    on_colors = helpers.defaultNullOpts.mkLuaFn "function(colors) end" ''
+      Override specific color groups to use other groups or a hex color.
+      Function will be called with a `ColorScheme` table.
+
+      ```
+      @param colors ColorScheme
+      ```
+    '';
+
+    on_highlights = helpers.defaultNullOpts.mkLuaFn "function(highlights, colors) end" ''
+      Override specific highlights to use other groups or a hex color.
+      Function will be called with a `Highlights` and `ColorScheme` table.
+
+      ```
+      @param highlights Highlights
+      @param colors ColorScheme
+      ```
+    '';
+  };
+
+  settingsExample = {
+    style = "auto";
+    variant = "default";
+    transparent = false;
+    dim_inactive = false;
+    styles = {
+      comments.italic = true;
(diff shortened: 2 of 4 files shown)


lib/maintainers.nix
```diff
@@ -54,4 +54,11 @@
     githubId = 66408983;
     name = "svl";
   };
+  nwjsmith = {
+    email = "nate@theinternate.com";
+    github = "nwjsmith";
+    githubId = 1348;
+    name = "Nate Smith";
+    keys = [ { fingerprint = "DE6F E362 7DF7 EEA0 79CB  38C1 F208 9547 768B 6CCA"; } ];
+  };
```
MattSturgeon, line 63: We prefer this to be done in a separate commit titled "`maintainers: add nwjsmith`"

plugins/colorschemes/modus.nix
MattSturgeon, line 1: This commit should be titled "`colorschemes/modus: init`"

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,93 @@
...
+    styles = {
+      comments.italic = true;
+      keywords.italic = true;
+      functions = { };
+      variables = { };
+    };
+    on_colors = "function(colors) end";
+    on_highlights = "function(highlights, colors) end";
+  };
+}
```
MattSturgeon, line 93: Most colorschemes set `opts.termguicolors = mkDefault true` in `extraConfig`. Does this plugin need `termguicolors`?

nwjsmith:
@MattSturgeon thanks for the quick review. Addressed all of your feedback.

plugins/colorschemes/modus.nix
nwjsmith: Good catch, it is required. It's based on tokyonight which also requires it. Fixed.

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,95 @@
...
+      comments.italic = true;
+      keywords.italic = true;
+      functions = { };
+      variables = { };
+    };
+    on_colors = "function(colors) end";
+    on_highlights = "function(highlights, colors) end";
+  };
+
+  extraConfig = cfg: { opts.termguicolors = lib.mkDefault true; };
```
MattSturgeon, line 94: I can't find any mention of `termguicolors` on the upstream readme, so maybe it isn't needed for this plugin. Someone more knowledgeable than me should probably weigh in here.

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,95 @@
...
+      Override specific color groups to use other groups or a hex color.
+      Function will be called with a `ColorScheme` table.
+      `@param colors ColorScheme`
+    '';
+
+    on_highlights = helpers.defaultNullOpts.mkLuaFn "function(highlights, colors) end" ''
+      Override specific highlights to use other groups or a hex color.
+      Function will be called with a `Highlights` and `ColorScheme` table.
+      `@param highlights Highlights`
+      `@param colors ColorScheme`
```
MattSturgeon, line 75:
```suggestion
      Override specific highlights to use other groups or a hex color.
      Function will be called with a `Highlights` and `ColorScheme` table.

      ```
        @param highlights Highlights
        @param colors ColorScheme
      ```
```

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,95 @@
...
+
+      variables = helpers.defaultNullOpts.mkHighlight "{}" "" ''
+        Define variables highlight properties.
+      '';
+    };
+
+    on_colors = helpers.defaultNullOpts.mkLuaFn "function(colors) end" ''
+      Override specific color groups to use other groups or a hex color.
+      Function will be called with a `ColorScheme` table.
+      `@param colors ColorScheme`
```
MattSturgeon, line 68:
```suggestion
      Override specific color groups to use other groups or a hex color.
      Function will be called with a `ColorScheme` table.

      ```
        @param colors ColorScheme
      ```
```

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,95 @@
...
+
+  settingsOptions = {
+    style =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "modus_operandi"
+          "modus_vivendi"
+        ]
+        ''
+          The theme comes in two styles, `modus_operandi`, a dark variant `modus_vivendi`.
```
MattSturgeon, line 24: It feels a bit redundant to list the available styles here when the mkEnum function will do that anyway...

Maybe this and `variant`'s description should instead try to describe what the option actually does and/or what things it affects?

Doesn't look like there's anything to quote in upstream's readme though, so I appreciate if you don't want to do this.

MattSturgeon (commented):
Looking good to me, just some minor comments/suggestions

MattSturgeon:
[Looks like](https://buildbot.nix-community.org/api/v2/logs/58322/raw_inline) something is going wrong when attempting to load modus from the runtimepath?

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,95 @@
...
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "modus";
+  isColorscheme = true;
+  originalName = "modus-themes-nvim";
+  defaultPackage = pkgs.vimPlugins.modus-themes-nvim;
```
MattSturgeon, line 12: Add `luaName = "modus-themes"` to fix requiring the plugin.

Personally, I'd re-order these as follows, but OTTOMH IDK if that's consistent with what we do in other plugins:


```suggestion
  name = "modus";
  luaName = "modus-themes";
  originalName = "modus-themes-nvim";
  defaultPackage = pkgs.vimPlugins.modus-themes-nvim;
  isColorscheme = true;
```

plugins/colorschemes/modus.nix
nwjsmith: Good call. I've cribbed new descriptions [from documentation on the original Emacs themes](https://protesilaos.com/emacs/modus-themes#h:f0f3dbcb-602d-40cf-b918-8f929c441baf).

plugins/colorschemes/modus.nix
nwjsmith: The theme is based on the Tokyo Night, and the colors [are specified as 24-bit RGB values](https://github.com/miikanissi/modus-themes.nvim/blob/master/lua/modus-themes/colors.lua) so I believe `termguicolors` is required.

MattSturgeon (changes requested):
A bit of further feedback on the new variant description. Otherwise looks good.

Any additional thoughts @GaetanLepage?

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,105 @@
...
+          "default"
+          "tinted"
+          "deuteranopia"
+          "tritanopia"
+        ]
+        ''
+          Styles come in four variants. `default` is the project's main style designed to cover a
+          broad range of needs. `tinted` tone down intensity and provide more color variety.
+          `deuteranopia` are optimized for users with red-green color deficiency. `tritanopia` are
+          optimized for users with blue-yellow color deficiency.
```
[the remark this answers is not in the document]
nwjsmith: I like it. Switched to a bulleted list and fixed the grammar. Thank you!

plugins/colorschemes/modus.nix
```diff
@@ -0,0 +1,107 @@
...
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "modus";
+  luaName = "modus-themes";
+  originalName = "modus-themes-nvim";
```
MattSturgeon, line 11: My bad for not spotting this sooner:

```suggestion
  originalName = "modus-themes.nvim";
```

MattSturgeon (approved):
Thanks for your efforts and perseverance!

PR #1584, opened 2024-05-28
colorschemes/nightfox: add flavor option

GaetanLepage:
Let's add an option to enhance discoverability.

### DIFF
--- a/plugins/colorschemes/nightfox.nix
+++ b/plugins/colorschemes/nightfox.nix
@@ -14,6 +14,25 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   maintainers = [ maintainers.GaetanLepage ];
 
+  colorscheme = null;
+  extraOptions = {
+    flavor = mkOption {
+      type = types.enum [
+        "carbonfox"
+        "dawnfox"
+        "dayfox"
+        "duskfox"
+        "nightfox"
+        "nordfox"
+        "terafox"
+      ];
+      example = "dayfox";
+      default = "nightfox";
+      description = "Which palette/flavor to use as the colorscheme.";
+    };
+  };
+  extraConfig = cfg: { colorscheme = mkDefault cfg.flavor; };
+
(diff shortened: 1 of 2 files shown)


plugins/colorschemes/nightfox.nix
```diff
@@ -14,6 +14,25 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        "nightfox"
+        "nordfox"
+        "terafox"
+      ];
+      example = "dayfox";
+      default = "nightfox";
+      description = "Which palette/flavor to use as the colorscheme.";
+    };
+  };
+  extraConfig = cfg: { colorscheme = mkDefault cfg.flavor; };
```
MattSturgeon, line 34: Will this `mkDefault` reliably override the other `mkDefault`?

GaetanLepage: The other `mkDefault` won't happen because I set `colorscheme = null` a few lines above.

plugins/colorschemes/nightfox.nix
```diff
@@ -14,6 +14,25 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        "dawnfox"
+        "dayfox"
+        "duskfox"
+        "nightfox"
+        "nordfox"
+        "terafox"
+      ];
+      example = "dayfox";
+      default = "nightfox";
+      description = "Which palette/flavor to use as the colorscheme.";
```
MattSturgeon, line 31: Is it worth adding an "equivalent to doing xyz" note here or do you think that's overkill and/or more confusing than helpful?

GaetanLepage: I would say that it would bring confusion.
When you think about it, most Nix modules are hierarchical: options most often only set the value of other options. In this sense, 90% of the options amount to "just setting this `foo` and `bar` options to 42".

MattSturgeon (approved):

PR #1585, opened 2024-05-28
lib/options: multi-line plugin default strings

MattSturgeon:
Detect whether a plugin default is multiline using `hasInfix "\n"`.

This works fairly well, but without trimming the default string we may run into some false-positives. We could borrow the implementation from https://github.com/NixOS/nixpkgs/pull/315411 or wait and see if/when that gets merged. Personally, I don't see this FIXME as a blocker, unless we find it affects a significant portion of the docs.

I also noticed some default strings seem to have excessive indentation, IDK how easily we could remove that or if we should even bother. ~~I left a TODO comment.~~

Here's a screenshot of docs built with this change:

<details><summary>Screenshot</summary>
<p>

image

</p>
</details> 

Note how 2/3 options have inline plugin defaults, while the first option has a multiline plugin default.

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -64,12 +64,13 @@ rec {
           # Assume a string default is already formatted as intended,
           # historically strings were the only type accepted here.
           # TODO consider deprecating this behavior so we can properly quote strings
-          if isString defaultValue then
-            defaultValue
-          else
-            generators.toPretty { multiline = false; } defaultValue;
+          if isString defaultValue then defaultValue else generators.toPretty { } defaultValue;
       in
-      "Plugin default: `${default}`";
+      "Plugin default:"
+      + (
+        # Detect whether `default` is multiline or inline:
+        if hasInfix "\n" default then "\n\n```nix\n${default}\n```" else " `${default}`"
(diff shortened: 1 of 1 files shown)


lib/options.nix
```diff
@@ -69,7 +69,12 @@ rec {
           else
             generators.toPretty { multiline = false; } defaultValue;
       in
-      "Plugin default: `${default}`";
+      "Plugin default:"
+      + (
+        # Detect multiline plugin defaults:
+        # FIXME we should trim leading/trailing newlines, to avoid false-positives
```
MattSturgeon, line 75: On second thought, this may be a non-issue.

If "false-positives" exist, they're a problem with the calling code, not the implementation here.

GaetanLepage:
I have checked locally and it looks great !

lib/options.nix
GaetanLepage: I agree. I guess you can remove the comment.

MattSturgeon: Done

GaetanLepage (approved):

Issue #1299, opened 2024-03-21
Add a configuration repository for inspiration?


no-mood:
### Discussed in https://github.com/nix-community/nixvim/discussions/1295

<div type='discussions-op-text'>

<sup>Originally posted by **no-mood** March 20, 2024</sup>
Hi guys, I was starting to use nixvim (and nvim in general) and I was taking a look at the following [configuration](https://github.com/GaetanLepage/nix-config/tree/master/home/modules/tui/neovim).
Do you think it's a good idea to add a configuration repository, in the form of nix flakes, so that they can be grouped together and be used as inspiration?
Thank you very much</div>

MattSturgeon:
I was thinking it might make sense to take inspiration from [kickstart.nvim](https://github.com/nvim-lua/kickstart.nvim), which is popular because it manages to cover the basics while remaining small & simple, but also including a ton of inline documentation explaining pretty much everything.

Maybe even directly port kickstart.nvim to nixvim more or less one-for-one (kickstart.nixvim 😂)?

It would probably make sense to include a kickstart style example in this repo (as a nix template), and have `nix check` test that the config is valid. That way it'd always be up to date with the latest breaking changes.

hbjydev:
kickstart.nvim.nix 🤣

no-mood:
I'd open to try and port it to nixvim, even without experience. What do you think? I could take less time for someone more experienced to do it.
If I start this project, what is the best approach?

GaetanLepage:
I plan to give it a shot (maybe on stream) this WE. Your feedback, as a new user, will be very valuable of course.

no-mood:
I'd love it! Keep us update, I'll participate for sure.

PR #1588, opened 2024-05-29
docs/user-guide: add configuration examples

GaetanLepage:
Fixes #1295 #1299 (kickstart.nixvim is still a WIP though).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ lightline plugin:
 When we do this, lightline will be set up to a sensible default, and will use
 gruvbox as the colorscheme, no extra configuration required!
 
+Check out [this list of real world nixvim configs](https://nix-community.github.io/nixvim/user-guide/config-examples.html)!
+
 ## Support/Questions
 If you have any question, please use the [discussions page](https://github.com/nix-community/nixvim/discussions/categories/q-a)! Alternatively, join the Matrix channel at [#nixvim:matrix.org](https://matrix.to/#/#nixvim:matrix.org)!
 
--- a/docs/user-guide/config-examples.md
+++ b/docs/user-guide/config-examples.md
@@ -1,0 +1,43 @@
+# User configurations
+
+Want some help with starting your configuration?
+Or maybe you're curious how others have done things?
+Take a look at these configuration examples below.
+
+<!-- TODO: add info about `kickstart.nixvim` -->
+
+**Note:**\
+Most of those configurations are using a [standalone build](../modules/standalone.html), however,
+all of the nixvim options are accessible no matter how you are using it (flake, NixOS/HM module, nix-darwin...).
+
+| Config | Comment |
+|-|-|
+| [ahwxorg/nixvim-config] | |
+| [alisonjenkins/neovim-nix-flake] | |
+| [GaetanLepage/nix-config] | Home-manager |
+| [gwg313/nvim-nix] | |
+| [hbjydev/hvim] | |
+| [nicolas-goudry/nixvim-config] | Heavily inspired by AstroNvim |
+| [pete3n/nixvim-flake] | |
+| [redyf/Neve] | meticulously crafted custom configuration for Nixvim |
+| [siph/nixvim-flake] | |
+| [Tanish2002/neovim-config] | |
+| [traxys/Nixfiles] | |
+
+[ahwxorg/nixvim-config]: https://github.com/ahwxorg/nixvim-config
+[alisonjenkins/neovim-nix-flake]: https://github.com/alisonjenkins/neovim-nix-flake
+[GaetanLepage/nix-config]: https://github.com/GaetanLepage/nix-config/tree/master/home/modules/tui/neovim
+[gwg313/nvim-nix]: https://github.com/gwg313/nvim-nix
+[hbjydev/hvim]: https://github.com/hbjydev/hvim
+[nicolas-goudry/nixvim-config]: https://github.com/nicolas-goudry/nixvim-config
+[pete3n/nixvim-flake]: https://github.com/pete3n/nixvim-flake
+[redyf/Neve]: https://github.com/redyf/Neve
+[siph/nixvim-flake]: https://github.com/siph/nixvim-flake
+[Tanish2002/neovim-config]: https://github.com/Tanish2002/neovim-config
+[traxys/Nixfiles]: https://github.com/traxys/Nixfiles/tree/master/neovim
+
+## Share your config !
+
+To add a configuration to this list, either:
+- Edit [this file](https://github.com/nix-community/nixvim/blob/main/docs/user-guide/config-examples.md) and make a PR on the [nixvim repo](https://github.com/nix-community/nixvim).
+- Send a message to the [matrix _Documentation_ room](https://matrix.to/#/#nixvim-documentation:matrix.org)
--- a/docs/man/default.nix
+++ b/docs/man/default.nix
@@ -39,6 +39,7 @@ let
           ${./nixvim-header-start.5} \
           ${mkMDSection ../user-guide/helpers.md} \
           ${mkMDSection ../user-guide/faq.md} \
+          ${mkMDSection ../user-guide/config-examples.md} \
           ${./nixvim-header-end.5} \
           >$out/nixvim-header.5
       '';
--- a/docs/mdbook/SUMMARY.md
+++ b/docs/mdbook/SUMMARY.md
@@ -1,10 +1,11 @@
-# Structure for nixvim docs 
+# Structure for nixvim docs
 
 # User guide
 
 - [Installation](./user-guide/install.md)
 - [Helpers](./user-guide/helpers.md)
 - [FAQ](./user-guide/faq.md)
+- [Configuration examples](./user-guide/config-examples.md)
 
 # Platforms
 


MattSturgeon:
Looks fancy! I'm going to hold off reviewing while I'm on my phone though, I'll look properly later.

README.md
```diff
@@ -35,6 +35,8 @@ lightline plugin:
 When we do this, lightline will be set up to a sensible default, and will use
 gruvbox as the colorscheme, no extra configuration required!
 
+Check out [this list of real world nixvim configs](https://nix-community.github.io/nixvim/user-guide/config-examples.html) !
```
MattSturgeon, line 38:
```suggestion
Check out [this list of real world nixvim configs](https://nix-community.github.io/nixvim/user-guide/config-examples.html)!
```

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,30 @@
+# User configurations
+
+Want some help with starting your configuration ?
+Have a look at this catalog of configuration examples.
```
MattSturgeon, line 4: Subjective (other than space before `?`), feel free to ignore or iterate:

```suggestion
Want some inspiration for your configuration?
Or maybe you're curious how others have done things?
Take a look at these configuration examples below.
```

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,30 @@
...
+
+<!-- TODO: add info about `kickstart.nixvim` -->
+
+**Note:**\
+Most of those configurations are using a [standalone flake](../modules/standalone.html).\
```
MattSturgeon, line 9: Might be misleading to conflate standalone builds with seperate flakes? E.g. I use a standalone build, but it's just one flake output on my main system configs flake. While maybe not in these examples, someone _could_ do a standalone build without using flakes at all... Or maybe I'm just being pedantic?

```suggestion
Most of those configurations are using a [standalone build](../modules/standalone.html).\
```

MattSturgeon: I'd probably also replace the period with a comma and the next line's "of course" with a "however" to join the two sentences into a single thought.

I'm getting very picky and subjective here though.

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,30 @@
...
+| [hbjydev/hvim](https://github.com/hbjydev/hvim)                                                           |                                                       |
+| [nicolas-goudry/nixvim-config](https://github.com/nicolas-goudry/nixvim-config)                           | Heavily inspired by AstroNvim                         |
+| [pete3n/nixvim-flake](https://github.com/pete3n/nixvim-flake)                                             |                                                       |
+| [redyf/Neve](https://github.com/redyf/Neve)                                                               | meticulously crafted custom configuration for Nixvim  |
+| [siph/nixvim-flake](https://github.com/siph/nixvim-flake)                                                 |                                                       |
+| [Tanish2002/neovim-config](https://github.com/Tanish2002/neovim-config)                                   |                                                       |
+| [traxys/Nixfiles](https://github.com/traxys/Nixfiles/tree/master/neovim)                                  |                                                       |
```
MattSturgeon, line 24: Have you considered whether tables are suitable when reading the docs on mobile (or a small desktop window)?

MattSturgeon: If we go with a table (they do look nice), it may be wise to _either_ increase the column width a bit, _or_ don't align the column dividers at all (`|`). That way adding a new row with a wider column in the future shouldn't affect _all the other rows_ as well.

I'd also suggest using footnote-style markdown link syntax here, to make the source code a little more readable and less wide.

E.g. with un-aligned columns and footnote links:

```markdown
| Config | Comment |
|-|-|
| [siph/nixvim-flake] | |
| [Tanish2002/neovim-config] | |
| [traxys/Nixfiles] | |

[siph/nixvim-flake]: https://github.com/siph/nixvim-flake
[Tanish2002/neovim-config]: https://github.com/Tanish2002/neovim-config
[traxys/Nixfiles]: https://github.com/traxys/Nixfiles/tree/master/neovim
```

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Share your config !
+
+To add a configuration to this list, either:
+- Edit [this file](https://github.com/nix-community/nixvim/blob/main/docs/user-guide/config-examples.md) and make a PR on the [nixvim repo](https://github.com/nix-community/nixvim).
+- Send a message to the [matrix _Documentation_ room](https://matrix.to/#/#nixvim-documentation:matrix.org)
```
MattSturgeon, line 30: In addition to PR and Matrix, should we suggest GH Issues and GH Discussion threads?

docs/user-guide/config-examples.md
GaetanLepage: I agree, this is better !

docs/user-guide/config-examples.md
GaetanLepage: I didn't know about this syntax. Done !

docs/user-guide/config-examples.md
GaetanLepage: Hmm, for me those are the two best ways of "adding an entry to this list".

docs/user-guide/config-examples.md
GaetanLepage: > I'd probably also replace the period with a comma and the next line's "of course" with a "however" to join the two sentences into a single thought.

Done



> I'm getting very picky and subjective here though.

No, this is useful I think. As a native English speaker, you definitely have your say in this !

GaetanLepage: > Might be misleading to conflate standalone builds with seperate flakes? E.g. I use a standalone build, but it's just one flake output on my main system configs flake. While maybe not in these examples, someone _could_ do a standalone build without using flakes at all... Or maybe I'm just being pedantic?

While I completely get your point, I would stay with "standalone flake" to ensure consistency with the rest of our documentation (website, README, issues, ...). I agree that the term is not technically accurate.
Any opinion @traxys ?

traxys: I think I agree with @MattSturgeon, we should try to dissociate flake & build

GaetanLepage: Sure: `flake` -> `build` done

docs/user-guide/config-examples.md
MattSturgeon: I don't have a strong opinion here. Just throwing out the idea. @traxys what's your thoughts?

docs/user-guide/config-examples.md
GaetanLepage: Is that fine for you like this @MattSturgeon ?

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,43 @@
...
+[alisonjenkins/neovim-nix-flake]: https://github.com/alisonjenkins/neovim-nix-flake
+[GaetanLepage/nix-config]: https://github.com/GaetanLepage/nix-config/tree/master/home/modules/tui/neovim
+[gwg313/nvim-nix]: https://github.com/gwg313/nvim-nix
+[hbjydev/hvim]: https://github.com/hbjydev/hvim
+[nicolas-goudry/nixvim-config]: https://github.com/nicolas-goudry/nixvim-config
+[pete3n/nixvim-flake]: https://github.com/pete3n/nixvim-flake
+[redyf/Neve]: https://github.com/redyf/Neve
+[siph/nixvim-flake]: https://github.com/siph/nixvim-flake
+[Tanish2002/neovim-config]: https://github.com/Tanish2002/neovim-config
+[traxys/Nixfiles]: https://github.com/traxys/Nixfiles/tree/master/neovim
```
MattSturgeon, line 37: Traditionally, the "footnote" part of the link would go at the end of the document. But I don't think that's actually a requirement and I can see why you might prefer to keep it "local".

Personally, I'd have it at the EOF though.

docs/user-guide/config-examples.md
MattSturgeon: > Is that fine for you like this?

Yep, that wording lgtm

docs/user-guide/config-examples.md
GaetanLepage: It seems to work fine like that.
Keeping it local makes more sense to me (as long as it works ofc).

MattSturgeon: No strong opinion here. @traxys?

docs/user-guide/config-examples.md
```diff
@@ -0,0 +1,43 @@
+# User configurations
+
+Want some help with starting your configuration?
+Or maybe you're curious how others have done things?
+Take a look at these configuration examples below.
+
+<!-- TODO: add info about `kickstart.nixvim` -->
+
+**Note:**\
```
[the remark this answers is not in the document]
GaetanLepage: Indeed, But I was finding it clearer to have this header against the text block than breaking a line.
But I can add a blankine if you prefer.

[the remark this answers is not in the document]
MattSturgeon: > But I was finding it clearer to have this header against the text block than breaking a line.

No, I was trying to say that most markdown flavors would render this the same with or without escaping that LF.

Is that not the case with mdbook?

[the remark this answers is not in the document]
MattSturgeon: Actually, reading the spec having the `\` does the opposite from what I thought it did (I thought it was a redundant escape sequence):

> For a more visible alternative, a backslash before the [line ending](https://github.github.com/gfm/#line-ending) may be used instead of two spaces:\
> [Example 659](https://github.github.com/gfm/#example-659)
>
> ```markdown
> foo\
> baz
> ```
>
> ```html
> <p>foo<br />
> baz</p>
> ```



> But I was finding it clearer to have this header against the text block than breaking a line.

I think I misinterpreted this. Were you saying you preferred rendering a linebreak after the **Note:** header, but wanted to avoid a full _paragraph_ break?

Your current code should render something like this:

```html
<p>
<b>Note:<b>
<br>
Most of those configurations are using a <a href="../modules/standalone.html">standalone build</a>, however,
all of the nixvim options are accessible no matter how you are using it (flake, NixOS/HM module, nix-darwin...).
</p>
```

or

**Note:**\
Most of those configurations are using a [standalone build](../modules/standalone.html), however, all of the nixvim options are accessible no matter how you are using it (flake, NixOS/HM module, nix-darwin...).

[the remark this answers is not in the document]
MattSturgeon: (This is resolved, assuming this is rendering as you intended)

docs/user-guide/config-examples.md
traxys: I think I prefer what @GaetanLepage suggested

docs/user-guide/config-examples.md
traxys: No strong opinions from me either

GaetanLepage:
Any additional requests ?

MattSturgeon (approved):
LGTM!

GaetanLepage:
Thanks for the review guys :)

PR #1589, opened 2024-05-29
plugins/edgy: init

### LINKED ISSUE #1568: [PLUGIN REQUEST] edgy.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `edgy`  | 
| Homepage | `https://github.com/folke/edgy.nvim`     | 
| Nixpkgs  | `https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/editors/vim/plugins/generated.nix#L3385` |

GaetanLepage:
Add support for [edgy.nvim](https://github.com/folke/edgy.nvim).

Fixes #1568

### DIFF
--- a/plugins/ui/edgy.nix
+++ b/plugins/ui/edgy.nix
@@ -1,0 +1,321 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "edgy";
+  originalName = "edgy.nvim";
+  defaultPackage = pkgs.vimPlugins.edgy-nvim;
+
+  maintainers = [ maintainers.GaetanLepage ];
+
+  extraConfig = cfg: {
+    # Those options are strongly recommended by the plugin author:
+    # https://github.com/folke/edgy.nvim?tab=readme-ov-file#-installation
+    opts = {
+      laststatus = mkDefault 3;
+      splitkeep = mkDefault "screen";
+    };
+  };
+
+  settingsOptions =
+    let
+      viewOpts = {
+        ft = helpers.mkNullOrStr ''
+          File type of the view.
+        '';
+
+        filter = helpers.mkNullOrLuaFn ''
+          Optional function to filter buffers and windows.
+
+          `fun(buf:buffer, win:window)`
+        '';
+
+        title = helpers.mkNullOrStr ''
+          Optional title of the view.
+          Defaults to the capitalized filetype.
+        '';
+
+        size = helpers.mkNullOrOption types.ints.unsigned ''
+          Size of the short edge of the edgebar.
+          For edgebars, this is the minimum width.
+          For panels, minimum height.
+        '';
+
+        pinned = helpers.mkNullOrOption types.bool ''
+          If true, the view will always be shown in the edgebar even when it has no windows.
+        '';
+
+        open = helpers.mkNullOrStr ''
+          Function or command to open a pinned view.
+        '';
+
+        wo = helpers.mkNullOrOption (with types; attrsOf anything) ''
+          View-specific window options.
+        '';
+      };
+
+      mkViewOptsOption =
+        name:
+        helpers.defaultNullOpts.mkListOf (
+          with types;
+          either str (submodule {
+            options = viewOpts;
+          })
+        ) [ ] "List of the ${name} edgebar configurations.";
+    in
+    {
+      left = mkViewOptsOption "left";
+      bottom = mkViewOptsOption "bottom";
+      right = mkViewOptsOption "right";
+      top = mkViewOptsOption "top";
+
+      options =
+        mapAttrs
+          (_: defaultSize: {
+            size = helpers.defaultNullOpts.mkUnsignedInt defaultSize ''
+              Size of the short edge of the edgebar.
+              For edgebars, this is the minimum width.
+              For panels, minimum height.
+            '';
+
+            wo = helpers.mkNullOrOption (with types; attrsOf anything) ''
+              View-specific window options.
+            '';
+          })
+          {
+            left = 30;
+            bottom = 10;
+            right = 30;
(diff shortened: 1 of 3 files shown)


MattSturgeon (commented):
Can't see any issues. Hesitant to approve until I've looked properly on a bigger screen.

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,321 @@
...
+
+      keys = mkOption {
+        type = with helpers.nixvimTypes; attrsOf (either strLuaFn (enum [ false ]));
+        default = { };
+        apply = mapAttrs (_: v: if isString v then helpers.mkRaw v else v);
+        description = ''
+          Buffer-local keymaps to be added to edgebar buffers.
+          Existing buffer-local keymaps will never be overridden.
+          Set to `false` to disable a builtin.
+
```
MattSturgeon, line 199: Not sure how we usually do this, but it'd be nice to document the expected function signature, i.e. a void function that accepts one parameter "`win`" which is a `Edgy.Window`

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,321 @@
...
+                "⠦"
+                "⠧"
+                "⠇"
+                "⠏"
+              ];
+              interval = 80;
+            }}
+          '';
+        };
+      };
```
MattSturgeon, line 171: This would benefit from some explanatory comments. I'd also like @traxys' thoughts on this before merging, since I'm struggling to follow the what & why here.

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,321 @@
...
+                };
+              })
+            );
+          default = null;
+          example = "require('noice.util.spinners').spinners.circleFull";
+          apply = v: if isString v then helpers.mkRaw v else v;
+          description = ''
+            Spinner for pinned views that are loading.
+
+            ${helpers.defaultNullOpts.mkDefaultDesc {
```
MattSturgeon, line 154: Personally, I think this would be better than using `mkDefaultDesc` directly:

```nix
description = helpers.defaultNullOpts.mkDesc { /* ... */ } ''
  Spinner for pinned views that are loading.
'';
```

plugins/ui/edgy.nix
GaetanLepage: Like so ?

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,323 @@
...
+              "⠙"
+              "⠹"
+              "⠸"
+              "⠼"
+              "⠴"
+              "⠦"
+              "⠧"
+              "⠇"
+              "⠏"
+            ];
```
MattSturgeon, line 165: I'd be tempted to put the `frames` default in the let-block, but as it's only used twice this is fine if preferred.

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,323 @@
...
+        winbar = true;
+        winfixwidth = true;
+        winfixheight = false;
+        winhighlight = "WinBar:EdgyWinBar,Normal:EdgyNormal";
+        spell = false;
+        signcolumn = "no";
+      } "Global window options for edgebar windows.";
+
+      keys = mkOption {
+        type = with helpers.nixvimTypes; attrsOf (either strLuaFn (enum [ false ]));
```
MattSturgeon, line 190: I'd be tempted to move `with helpers.nixvimTypes` so that it applies to the whole of `settingsOptions`.

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,323 @@
...
+      wo = helpers.defaultNullOpts.mkAttrsOf types.anything {
+        winbar = true;
+        winfixwidth = true;
+        winfixheight = false;
+        winhighlight = "WinBar:EdgyWinBar,Normal:EdgyNormal";
+        spell = false;
+        signcolumn = "no";
+      } "Global window options for edgebar windows.";
+
+      keys = mkOption {
```
MattSturgeon, line 189:
```suggestion
      # This option accepts an attrs or a lua string.
      # Hence, we use `mkOption` to convert the string to raw lua in `apply`.
      keys = mkOption {
```

plugins/ui/edgy.nix
```diff
@@ -0,0 +1,323 @@
...
+        description = ''
+          Buffer-local keymaps to be added to edgebar buffers.
+          Existing buffer-local keymaps will never be overridden.
+
+          Each value is either:
+          - A function declaration (as a raw lua string)
+            -> `fun(win:Edgy.Window)`
+          - `false` to disable this mapping.
+
+          ${helpers.defaultNullOpts.mkDefaultDesc {
```
MattSturgeon, line 202: Use `description = mkDesc` instead

MattSturgeon (changes requested):
Minor comments

plugins/ui/edgy.nix
GaetanLepage: That's a good idea. Done.

plugins/ui/edgy.nix
GaetanLepage: Specifically for this plugin or in general ?

MattSturgeon: In general, if I found I was referencing types/nixvimTypes a fair amount throughout `settingsOptions`.

MattSturgeon (approved):
Can't see any issues :+1:

plugins/ui/edgy.nix
MattSturgeon: > However, this practice (of relying on `with` statements) is being discouraged in nixpkgs as it can lead to subtle and nasty errors.

I agree with that principal, but I think using a `with types` statement on an attrset exclusively containing option definitions is a reasonable exception.

MattSturgeon: As I say, just my preference. No strong opinions.

PR #1591, opened 2024-05-30
lib/options: delegate all non-string default rendering to `mkDefaultDesc`

MattSturgeon:
- **lib/options: `mkEnum` quote default if string**
- **lib/options: `defaultNullOpts` delegate string rendering**

Other than gradually transitioning away from passing in large pre-rendered strings to `mkNullable` (etc), and then subsequently being able to _also_ delegate string-type rendering in `mkEnum` and `mkStr`, this should conclude this series of PRs adding `toPretty` support to `mkDefaultDesc`.

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -104,20 +104,20 @@ rec {
 
     mkLuaFn = default: desc: mkNullOrLuaFn (mkDesc default desc);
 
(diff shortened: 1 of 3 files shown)


GaetanLepage (approved):
Very nice. This is much better.

PR #1601, opened 2024-05-31
helpers/toLuaObject: add support for raw keys

GaetanLepage:
This allows to have "raw lua" for table keys.

This:
```nix
{
  __raw__foo = 1;
  bar = 2;
}
```

Would translate into
```lua
{
  [foo] = 1;
  ['bar'] = 2;
}
```

### DIFF
--- a/lib/to-lua.nix
+++ b/lib/to-lua.nix
@@ -15,12 +15,17 @@ rec {
         + (concatStringsSep "," (
           mapAttrsToList (
             n: v:
-            if (builtins.match "__unkeyed.*" n) != null then
-              toLuaObject v
+            let
+              valueString = toLuaObject v;
+            in
+            if hasPrefix "__unkeyed" n then
+              valueString
+            else if hasPrefix "__rawKey__" n then
+              "[${n}] = " + valueString
             else if n == "__emptyString" then
-              "[''] = " + (toLuaObject v)
+              "[''] = " + valueString
             else
-              "[${toLuaObject n}] = " + (toLuaObject v)
+              "[${toLuaObject n}] = " + valueString
           ) (filterAttrs (n: v: v != null && (toLuaObject v != "{}")) args)
         ))
         + "}"
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -11,6 +11,47 @@ with lib;
     "__empty" = null;
   };
 
+  /**
+    Turn all the keys of an attrs into raw lua.
+
+    # Example
+
+    ```nix
+    toRawKeys { foo = 1; bar = 2; }
+    => { __rawKey__foo = 1; __rawKey__bar = 2; }
+    ```
+
+    # Type
+
+    ```
+    toRawKeys :: AttrSet -> AttrSet
+    ```
+  */
+  toRawKeys = mapAttrs' (n: v: nameValuePair "__rawKey__${n}" v);
+
+  /**
+    Create a 1-element attrs with a raw lua key.
+
+    # Example
+
+    ```nix
+    mkRawKey "foo" 1
+    => { __rawKey__foo = 1; }
+    ```
+
+    # Type
+
+    ```
+    mkRawKey :: String -> String -> AttrSet
+    ```
+
+    # Arguments
+
+    - [n] The attribute name (raw lua)
+    - [v] The attribute value
+  */
+  mkRawKey = n: v: toRawKeys { "${n}" = v; };
+
   /*
        Convert a string from camelCase to snake_case
     Type: string -> string
--- a/docs/user-guide/helpers.md
+++ b/docs/user-guide/helpers.md
@@ -40,3 +40,18 @@ A certain number of helpers are defined that can be useful:
 
 - `helpers.enableExceptInTests`: Evaluates to `true`, except in `mkTestDerivationFromNixvimModule`
   where it evaluates to `false`. This allows to skip instantiating plugins that can't be run in tests.
+
+- `helpers.toRawKeys attrs`: Convert the keys of the given `attrs` to raw lua.
+  ```nix
+    toRawKeys {foo = 1; bar = "hi";}
+  ```
+  will translate in lua to:
+  ```lua
+    {[foo] = 1, [bar] = 2,}
+  ```
+  Otherwise, the keys of a regular `attrs` will be interpreted as lua string:
+  ```lua
+    {['foo'] = 1, ['bar'] = 2,}
+    -- which is the same as
+    {foo = 1, bar = 2,}
+  ```


GaetanLepage:
Hmm, what do you think about the prefix ? Maybe `__rawKey__` would be better no ?

lib/utils.nix
```diff
@@ -11,6 +11,9 @@ with lib;
     "__empty" = null;
   };
 
+  # Turns all the keys of an attrs to raw lua.
+  rawKeysAttrs = mapAttrs' (n: v: nameValuePair "__rawKey__${n}" v);
```
GaetanLepage, line 15:
```suggestion
  toRawKeysAttrs = mapAttrs' (n: v: nameValuePair "__rawKey__${n}" v);
```
?

MattSturgeon: Agreed. I'm unsure on `toRawKeysAttrs` vs `toRawKeyAttr` though

GaetanLepage: Hmm, maybe `toRawKeys` ? X) I can't choose.

MattSturgeon:
So to set a raw type value with a raw key, you'd use this?

```nix
{
  __rawKey__someLuaHere.__raw = "moreLuaHere";
}
```

I agree we should support raw keys. Prefixes feel a little messy though.

Would something along these lines work (better/worse)?

```nix
{
  __key.__raw = "someLua";
  __value = "a string";
}
```

GaetanLepage:
> So to set a raw type value with a raw key, you'd use this?
> 
> ```nix
> {
>   __rawKey__someLuaHere.__raw = "moreLuaHere";
> }
> ```
> 
> I agree we should support raw keys. Prefixes feel a little messy though.
> 
> Would something along these lines work (better/worse)?
> 
> ```nix
> # a __key __value pair:
> {
>   __key.__raw = "someLua";
>   __value = "a string";
> }
> ```

I am not sure to get the proposition.
What is important is to support any combination of raw/normal key and raw/normal value.

```nix
{
  normal_key_1 = "normal value";
  normal_key_2.__raw = "raw lua value";
  __raw__some_raw_key_1 = "normal value";
  __raw__some_raw_key_2.__raw = "raw lua value";
}
```
Apart from prefixes, I don't see any option (but I could miss some !).
Important note: raw keys will most likely never be popular. Indeed, it is the first time that we encounter a need for them.
So, the usability is not crucial for users.

MattSturgeon:
> > Would something along these lines work (better/worse)?
> > ```nix
> > # a __key __value pair:
> > {
> >   __key.__raw = "someLua";
> >   __value = "a string";
> > }
> > ```
> 
> I am not sure I get the proposition. [...] Apart from prefixes, I don't see any option (but I could miss some !).

My "pair" proposal was intended to allow for slightly easier to read (if more verbose) definitions.

In hindsight it'd be a little clunky as it'd have to be mapped from a list of pairs (`[ {pair} ]`) into a lua table (string).

_I guess_ that could be an alternative way to define mixed list/dict tables though (a list of normal types alongside pairs).

As with any `[pair]->{dict}` mapping, you'd have key duplication issues too.

```nix
luaTableOption = [
  {
    __key.__raw = "luaKey";
    __value = "string";
  }
  {
    __key = "normal_key";
    __value.__raw = "lua code";
  }
  {
    __key = "another_key";
    __value = "string";
  }
  "un-keyed table entry"
  {
    __raw = "print('un-keyed table entry with raw lua')";
  }
];
```

> What is important is to support any combination of raw/normal key and raw/normal value.

Agreed

> Important note: raw keys will most likely never be popular. Indeed, it is the first time that we encounter a need for them. So, the usability is not crucial for users.

Agreed, it'll likely only be used internally, but some users may find it useful too.

---

I think my `__rawKey__` prefix legibility concerns would be addressed better by the `toRawKeyAttr` function you've already added, and/or possibly a `mkRawKey` style function.

Consider:

```nix
{
  someOption =
    { /* normal attrs */ }
    // (helpers.mkRawKey "someLuaKey" "some value");
  # vs
  otherOption =
    { /* normal attrs */ }
    // (helpers.toRawKeyAttrs {
      "someLuaKey" = "some value";
    });
  # vs
  lowLevelOption = {
    /* normal attrs */
    "__rawKey__someLuaKey" = "some value";
  };
```

lib/utils.nix
MattSturgeon: I think either `toRawKeys`, `toRawKeyAttrs` (one `s` total), or `attrsToRawKeys`.

We could also have a companion function `mkRawKey = key: value: toRawKey { ${name} = value; }`

lib/utils.nix
```diff
@@ -11,6 +11,12 @@ with lib;
     "__empty" = null;
   };
 
+  # Turn all the keys of an attrs to raw lua.
+  toRawKey = mapAttrs' (n: v: nameValuePair "__rawKey__${n}" v);
```
MattSturgeon, line 15: I think this should be a plural:

```suggestion
  toRawKeys = mapAttrs' (n: v: nameValuePair "__rawKey__${n}" v);
```

(also update usage in `mkRawKey`, of course)

lib/utils.nix
```diff
@@ -11,6 +11,12 @@ with lib;
...
+
+  # Create a 1-element attrs with a raw lua key
```
MattSturgeon, line 17:
```nix
  /**
    Create a 1-element attrs with a raw lua key.

    # Example

    ```nix
    mkRawKey "foo" 1
    => { __rawKey__foo = 1; }
    ```

    # Type

    ```
    mkRawKey :: String -> String -> AttrSet
    ```

    # Arguments

    - [n] The attribute name (raw lua)
    - [v] The attribute value
  */
```

lib/utils.nix
```diff
@@ -11,6 +11,12 @@ with lib;
     "__empty" = null;
   };
 
+  # Turn all the keys of an attrs to raw lua.
```
MattSturgeon, line 14: Use [RFC145](https://github.com/nixos/rfcs/blob/master/rfcs/0145-doc-strings.md#migration-example)?

```nix
  /**
    Turn all the keys of an attrs into raw lua.

    # Example

    ```nix
    toRawKeys { foo = 1; bar = 2; }
    => { __rawKey__foo = 1; __rawKey__bar = 2; }
    ```

    # Type

    ```
    toRawKeys :: AttrSet -> AttrSet
    ```
  */
```

lib/to-lua.nix
```diff
@@ -15,12 +15,17 @@ rec {
         + (concatStringsSep "," (
           mapAttrsToList (
             n: v:
-            if (builtins.match "__unkeyed.*" n) != null then
-              toLuaObject v
+            let
+              valueString = toLuaObject v;
```
MattSturgeon, line 19:
```suggestion
              lua = toLuaObject v;
```
What do you think?

MattSturgeon (commented):
Final suggestions, use or ignore as you wish.

Also consider whether we should add anything to the user-guide.

lib/utils.nix
GaetanLepage: Done

lib/to-lua.nix
GaetanLepage: I think that `valueString` is clearer, but again, no strong opinion X)

MattSturgeon: I think `lua` implies the string has already been rendered/formatted as lua, but I also don't have a strong opinion.

`luaValue` and `luaString` could also work, though `lua` has the advantage of also being less verbose.

GaetanLepage:
Sure, we have to wait for traxys' ACK before merging?

MattSturgeon:
> Sure, we have to wait for traxys' ACK before merging.

Unless you think that's overkill

GaetanLepage:
@traxys can you have a quick look at this ?
I would like to get the barbar refactoring merged before 24.05.

PR #1602, opened 2024-05-31
flake-modules/wrappers: add default module

Prince213:
This allows using `nixvim.homeManagerModules.default` for example.


MattSturgeon (approved):
Thanks for this!

PR #1603, opened 2024-05-31
lib/options: introduce "prime" fn variants (alt API)

MattSturgeon:
This PR adds new variants for all low-lever option factory funcs (`mkNullable`, `mkNullOrOption`, etc), as well as a handful of higher-level factories (`mkAttributeSet`, `mkAttrsOf`, & `mkListOf`).

This is done so that we have easy access to the underlying `lib.mkOption` attributes, such as `example`.

All attributes are passed through blindly, unless they are intentionally modified along the way.

Currently this PR doesn't make any effort to document _required_ arg-attributes.

I made (more or less) every function its own commit because I needed to use `git bisect`.\
This will also make it easier to drop/revert any specific function if we decide it's overkill to have a prime variant of it.\
I'll happily squash once we're ready to merge - how would you like the history structured?

Unrelated change: combined `mkDesc` and `mkDefaultDesc` into a single function, since `mkDefaultDesc` was not really intended to be used directly. The same functionality can be achieved by passing an empty `description` to `mkDesc`.

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -54,14 +54,16 @@ rec {
     );
   mkNullOrLuaFn = description: mkNullOrLua' { inherit description; };
 
-  mkNullOrStrLuaOr =
-    ty: desc:
-    lib.mkOption {
-      type = lib.types.nullOr (types.either nixvimTypes.strLua ty);
-      default = null;
-      description = desc;
-      apply = v: if builtins.isString v then mkRaw v else v;
-    };
+  mkNullOrStrLuaOr' =
+    { type, ... }@args:
+    mkNullOrOption' (
+      args
+      // {
+        type = with nixvimTypes; either strLua type;
+        apply = v: if isString v then mkRaw v else v;
+      }
+    );
+  mkNullOrStrLuaOr = type: description: mkNullOrStrLuaOr' { inherit type description; };
 
   mkNullOrStrLuaFnOr =
     ty: desc:
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -171,9 +171,22 @@ rec {
       default:
       assert default == null || isString default;
       mkNullableWithRaw types.str (generators.toPretty { } default);
-    mkAttributeSet = mkNullable nixvimTypes.attrs;
-    mkListOf = ty: default: mkNullable (with nixvimTypes; listOf (maybeRaw ty)) default;
-    mkAttrsOf = ty: default: mkNullable (with nixvimTypes; attrsOf (maybeRaw ty)) default;
+
+    mkAttributeSet' = args: mkNullable' (args // { type = nixvimTypes.attrs; });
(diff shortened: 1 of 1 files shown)


GaetanLepage:
> I'll happily squash once we're ready to merge - how would you like the history structured?

I don't think that a squash is necessary. I am fine with one commit per function.

> Unrelated change: combined `mkDesc` and `mkDefaultDesc` into a single function, since `mkDefaultDesc` was not really intended to be used directly. The same functionality can be achieved by passing an empty `description` to `mkDesc`.

Makes sense indeed.

GaetanLepage:
My opinion on this quite mixed.
On the one hand, I find it quite appealing to get those new helpful functions.
At the same time, this "bloats" (maybe a bit too strong of a word) our already quite heavy `helpers` function set. I am afraid that this could "scare" even more potential new contributors. But this is just a feeling.

So,  no _strong reject_ from my side. I would be curious to see what @traxys thinks about that.

MattSturgeon:
> At the same time, this "bloats" (maybe a bit too strong of a word) our already quite heavy `helpers` function set. I am afraid that this could "scare" even more potential new contributors. But this is just a feeling.

In principal, I don't think we should be afraid of growing `helpers`. Although if it does grow larger, we might consider putting less stuff at the top-level (same principal as https://github.com/NixOS/nixpkgs/issues/266103). If we avoid adding functions for this reason alone, we will end up making our lives more difficult in the long-run.

I think the helpers library becoming intimidating is best addressed via documentation, examples, and guides. In particular, it may be nice to have automated docs similar to the [Nixpkgs Manual](https://nixos.org/manual/nixpkgs/stable/#chap-functions), [Noogle](https://noogle.dev), etc.

That said, if there are specific functions that are so simple and/or "single use" that it makes no sense to have a prime variant, we can drop them on a case-by case basis while reviewing this PR (or even revert them after it's merged).

GaetanLepage:
> If we avoid adding functions for this reason alone, we will end up making our lives more difficult in the long-run.

Sure.



> I think the helpers library becoming intimidating is best addressed via documentation, examples, and guides. In particular, it may be nice to have automated docs similar to the [Nixpkgs Manual](https://nixos.org/manual/nixpkgs/stable/#chap-functions), [Noogle](https://noogle.dev), etc.

Definitely true.

> That said, if there are specific functions that are so simple and/or "single use" that it makes no sense to have a prime variant, we can drop them on a case-by case basis while reviewing this PR (or even revert them after it's merged).

I agree. Also, even though we export the helpers publicly, I don't think we should be too worried of breaking users' workflow when changing them. `mkRaw` is probably used by some people, but apart from this, I doubt many are used outside the project. Especially the option declaration stuff.

traxys (approved):

traxys:
Yeah I don't really have a problem with a growing number of helpers, as long as we know which one to apply

GaetanLepage (approved):

PR #1605, opened 2024-05-31
modules/keymaps: fix false-positive `lua` warning

MattSturgeon:
It appears that submodule options for list-of submodules are considered "defined" when _actually_ defined, while "plane" submodules (i.e. `plugins.wtf.keymaps`) seem to behave differently...

To work around the issue, I've made the (hidden and deprecated) `lua` keymap option "`nullOr bool`", so that we can assume "`null`" means "unset"...

This could cause an edge-case if a user is _reading_ the `lua` value in their config and assuming it is a non-null boolean though.

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -80,6 +80,9 @@ rec {
 
   mkMapOptionSubmodule =
     defaults:
+    # TODO remove assert once `lua` option is gone
+    # This is here to ensure no uses of `mkMapOptionSubmodule` set a `lua` default
+    assert !(defaults ? lua);
     (
       with types;
       submodule {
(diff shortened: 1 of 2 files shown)


GaetanLepage:
> To work around the issue, I've made the (hidden and deprecated) `lua` keymap option "`nullOr bool`", so that we can assume "`null`" means "unset"...

I'm OK with that.

> This could cause an edge-case if a user is _reading_ the `lua` value in their config and assuming it is a non-null boolean though.

Probably no-one, and certainly a very small amount of people. So I wouldn't care much. (Good to point it out nonetheless).

GaetanLepage (approved):

PR #1606, opened 2024-05-31
plugins/lsp: add ruby-lsp

### LINKED ISSUE #1604: [PLUGIN REQUEST] ruby-lsp support
| Field    | Description |
|----------|-------------|
| Plugin   | `ruby-lsp`  | 
| Homepage | https://github.com/Shopify/ruby-lsp     | 
| Nixpkgs  | [`ruby-lsp`](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/top-level/ruby-packages.nix#L3294) | 

## Extra Information

Hi, I'm wondering if by any chance you can add support for ruby-lsp, it is a newer alternative to solargraph that is supposed to be more accurate. I'm very new to nixos and nixvim, so apologies if I'm missing details in the issue description.

GaetanLepage:
Add support for [ruby-lsp](https://github.com/Shopify/ruby-lsp).

Fixes #1604

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -503,6 +503,11 @@ let
       package = null;
     }
     {
+      name = "ruby-lsp";
+      description = "ruby-lsp for Ruby";
+      serverName = "ruby_lsp";
+    }
+    {
       name = "ruff";
       description = "Official ruff language server (Rust) for Python";
     }
--- a/tests/test-sources/plugins/lsp/_lsp.nix
+++ b/tests/test-sources/plugins/lsp/_lsp.nix
@@ -174,6 +174,7 @@
             pylsp.enable = true;
             pylyzer.enable = true;
             pyright.enable = true;
+            ruby-lsp.enable = true;
             ruff.enable = true;
             ruff-lsp.enable = true;
             rust-analyzer = {


plugins/lsp/language-servers/default.nix
```diff
@@ -502,6 +502,11 @@ let
       serverName = "rnix";
       package = null;
     }
+    {
+      name = "ruby-lsp";
+      description = "ruby-lsp for Ruby";
+      serverName = "ruby_lsp";
```
MattSturgeon, line 508: Should this define a package (`pkgs.ruby-lsp`) or is that handled automatically elsewhere?

MattSturgeon (approved):
I'm not confident  in this part of the codebase, but this looks consistent with outher definitions. If CI passes I'm happy.

plugins/lsp/language-servers/default.nix
GaetanLepage: This is done automatically yes: https://github.com/nix-community/nixvim/blob/0ba2ea5416485d3f6552a108710711acd688cfef/plugins/lsp/language-servers/_mk-lsp.nix#L12

GaetanLepage:
> I'm not confident in this part of the codebase, but this looks consistent with outher definitions. If CI passes I'm happy.

Yes, we do this quite consistently in a straight forward way.

Issue #603, opened 2023-09-26
Expose keymap options for all keymaps defined by nixvim


traxys:
Nixvim defines keymaps for some  plugins, we should expose all the possible options for those plugins (like desc, silent, ....) instead of only allowing to specify a key.

trueNAHO:
Related: https://github.com/nix-community/nixvim/blob/b3fb1c4c8189bc873911da3f31d18082a0721fa9/lib/keymap-helpers.nix

MattSturgeon:
I think the best approach is to add a `mkMapOptionSubmodule` variant that omitts either `key` or `action`.

Maybe we could do something like:

```nix
mkMapOptionSubmodule = defaults: mkMapOptionSubmodule' {
  hasKey = true;
  hasAction = true;
  inherit defaults;
};

mkMapOptionSubmodule' = {
  hasKey ? false,
  hasAction ? false,
  defaults ? {}
}: submodule {
  options = (
	  (optionalAttrs hasKey {
	    key = mkOption (
	      {
	        type = str;
	        description = "The key to map.";
	        example = "<C-m>";
	      }
	      // (optionalAttrs (defaults ? key) { default = defaults.key; })
	    );
      })
      // (optionalAttrs hasAction {
	    action = mkOption (
	      {
	        type = nixvimTypes.maybeRaw str;
	        description = "The action to execute.";
	      }
	      // (optionalAttrs (defaults ? action) { default = defaults.action; })
	    );
      })
      // {
	    mode = mkModeOption defaults.mode or "";
	    options = mapConfigOptions;
        # Deprecated `lua` option omitted
	  }
  );
}

MattSturgeon:
Or since `mkMapOptionSubmodule` is only used by `wtf.nvim` and internally in `keymap-helpers`, we could change its interface instead of introducing a variant.

PR #1608, opened 2024-05-31
plugins/barbar: switch to mkNeovimPlugin + update options

GaetanLepage:
This new implementation does not bring along the weirdly recursive definition of the `icons` sub-options.
Hence, I had to fall back to `mkRemovedOption` for some of those. Not a big deal IMO.


plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,566 @@ let
...
+      mapAttrsToList (
+        optionName: funcName:
+        let
+          key = cfg.keymaps.${optionName};
+        in
+        optional (key != null) {
+          mode = "n";
+          inherit key;
+          action = "<Cmd>Buffer${funcName}<CR>";
+          options.silent = cfg.keymaps.silent;
```
MattSturgeon, line 207: `<cmd>` is always silent anyway

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,566 @@ let
...
+      ]
+    );
+
+  extraOptions = {
+    keymaps =
+      {
+        silent = mkEnableOption "silent keymaps for barbar";
+      }
+      // (mapAttrs (
+        optionName: funcName: helpers.mkNullOrOption types.str "Keymap for function Buffer${funcName}"
```
MattSturgeon, line 192: Could this be type `(either str keymap-submodule)`? Maybe the user wants a different `mode` or a custom `options.desc`, etc

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,566 @@ let
...
   };
 in
-{
-  options.plugins.barbar = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "barbar.nvim";
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "barbar";
+  originalName = "barbar.nvim";
+  defaultPackage = pkgs.vimPlugins.barbar-nvim;
+  extraPlugins = [ pkgs.vimPlugins.nvim-web-devicons ];
```
MattSturgeon, line 52: How do you decide if this dependency should be included? Upstream recommendation? (Should I do the same in the `dashboard` PR?)

MattSturgeon (commented):
Sorry, I don't have time to properly review right now. Here's some initial comments though.

plugins/bufferlines/barbar.nix
GaetanLepage: Good to know. I have removed the option.

plugins/bufferlines/barbar.nix
GaetanLepage: > How do you decide if this dependency should be included? Upstream recommendation? (Should I do the same in the `dashboard` PR?)

Well, this plugin is not mandatory per say. However, I feel like making an option to opt-out from this quite overkilled.
Also, this was the behavior for our `barbar` module prior to this refactor. Hence, I kept the same approach.

plugins/bufferlines/barbar.nix
MattSturgeon: `either str` or this:

```nix
(submodule {
  options = {
    mode = keymap-helpers.mkModeOption "n";
    options = keymap-helpers.mapConfigOptions;
  };
})
```

?

GaetanLepage: Are we sure that the plugin accepts this ?

GaetanLepage: > Are we sure that the plugin accepts this ?

This is stupid, it is not a plugin option.

MattSturgeon: We could also do this first: https://github.com/nix-community/nixvim/issues/603#issuecomment-2143472314

GaetanLepage: As this PR is quite big, I would have such a change in a dedicated one.

MattSturgeon: https://github.com/nix-community/nixvim/pull/1616

GaetanLepage: Done.

MattSturgeon:
`error: The option `plugins.barbar.keymaps.silent' can no longer be used since it's been removed. Keymaps will be silent anyways. This option has always been useless.`

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+      optionName: funcName:
+      helpers.mkNullOrOption (helpers.keymaps.mkMapOptionSubmodule {
+        defaults = {
+          mode = "n";
+          action = "<Cmd>Buffer${funcName}<CR>";
+        };
+        hasKey = true;
+        hasAction = true;
+      }) "Keymap for function Buffer${funcName}"
+    ) keymapsActions;
```
MattSturgeon, line 204: One limitation of `keymaps` being an attrset is that you can only define _one_ keymap per action.

At least one other plugin uses a list of submodules, with the `action` type being an enum of `optionNames` (`attrNames keymapsActions`): https://github.com/nix-community/nixvim/blob/f4ce7dad0e9ba16692783a70ed7f5b79e41d6c58/plugins/utils/improved-search.nix#L38-L54

That said _most_ plugin keymaps so far are using attrsets like this one.

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+  };
 
-    package = helpers.mkPluginPackageOption "barbar" pkgs.vimPlugins.barbar-nvim;
+  extraConfig = cfg: {
+    keymaps = filter (keymap: keymap != null) (
+      # TODO: switch to `attrValues cfg.keymaps` when removing the deprecation warnings above:
+      attrValues (filterAttrs (n: v: n != "silent") cfg.keymaps)
```
MattSturgeon, line 210: I don't think the `filterAttrs` is needed, my understanding is that `mkRemovedOptionModule` uses an _assertion_, so `keymaps.silent` should never be defined.

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+  };
 
-    animation = helpers.defaultNullOpts.mkBool true "Enable/disable animations";
+  settingsOptions = {
+    animation = helpers.defaultNullOpts.mkBool true ''
+      Enable/disable animations.
+    '';
 
-    autoHide = helpers.defaultNullOpts.mkBool false "Enable/disable auto-hiding the tab bar when there is a single buffer.";
+    auto_hide = helpers.defaultNullOpts.mkNullable (with types; either int (enum [ false ])) (-1) ''
```
MattSturgeon, line 219:
```suggestion
    auto_hide = helpers.defaultNullOpts.mkNullableWithRaw (with types; either int (enum [ false ])) (-1) ''
```

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
-    maximumPadding =
-      helpers.defaultNullOpts.mkInt 4
-        "Sets the maximum padding width with which to surround each tab";
+    highlight_visible = helpers.defaultNullOpts.mkBool true ''
+      Enables highlighting of visible buffers.
+    '';
+
+    icons = {
+      buffer_index =
+        helpers.defaultNullOpts.mkNullable
```
MattSturgeon, line 293:
```suggestion
        helpers.defaultNullOpts.mkNullableWithRaw
```

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+            with types;
+            either bool (enum [
+              "superscript"
+              "subscript"
+            ])
+          )
+          false
+          "If `true`, show the index of the buffer with respect to the ordering of the buffers in the tabline.";
+
+      buffer_number = helpers.defaultNullOpts.mkNullable (
```
MattSturgeon, line 304:
```suggestion
      buffer_number = helpers.defaultNullOpts.mkNullableWithRaw (
```

MattSturgeon: Maybe `defaultNullTypes` should get a `mkEither`...

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+        ])
+      ) false "If `true`, show the `bufnr` for the associated buffer.";
+
+      button = helpers.defaultNullOpts.mkNullable (with types; either str (enum [ false ])) "" ''
+        The button which is clicked to close / save a buffer, or indicate that it is pinned.
+        Use `false` to disable it.
+      '';
+
+      diagnostics = mkOption {
+        type = types.submodule {
```
MattSturgeon, line 318: Should this submodule have a `freeformType`, or is that already handled by it being part of `settings`?

Or is there some other reason why that's unnecessary or undesirable?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+          };
+          "vim.diagnostic.severity.INFO" = {
+            enabled = false;
+            icon = " ";
+          };
+          "vim.diagnostic.severity.WARN" = {
+            enabled = false;
+            icon = " ";
+          };
+        } "Set the icon for each diagnostic level.";
```
MattSturgeon, line 357: Maybe add a note to this option's description explaining that the attribute keys will be treated as raw lua key expressions?

Especially given how unusual this in nixvim.

This might be best demonstrated with an example along the lines of 

> The following config:
> ```nix
> {
>   "vim.diagnostic.severity.INFO".enabled = true;
>   "vim.diagnostic.severity.WARN".enabled = true;
> }
> ```
> Will result in the following lua:
> ```lua
> {
>   -- Note the table keys are not string literals:
>   [vim.diagnostic.severity.INFO] = { ['enabled'] = true },
>   [vim.diagnostic.severity.WARN] = { ['enabled'] = true },
> }
> ```

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+      };
+
+      gitsigns =
+        helpers.defaultNullOpts.mkAttrsOf
+          (
+            with types;
+            submodule {
```
MattSturgeon, line 364: Same question re: `freeformType`

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+        '';
+
+        separator_at_end = helpers.defaultNullOpts.mkBool true ''
+          If true, add an additional separator at the end of the buffer list.
+          Can be used to create a visual separation when the inactive buffer background color is the
+          same as the fill region background color.
+        '';
+      };
+
+      modified = helpers.defaultNullOpts.mkAttrsOf types.anything { button = "●"; } ''
```
MattSturgeon, line 432: Why `anything` and not `str`?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+            button = false;
+            filename = false;
+            separator.right = " ";
+          }
+          ''
+            The icons which should be used for a pinned buffer.
+            Supports all the base options (e.g. `buffer_index`, `filetype.enabled`, etc).
+          '';
+
+      alternate = helpers.mkNullOrOption (with types; attrsOf anything) ''
```
MattSturgeon, line 449: Why not use `mkAttrsOf`?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+        The icons which should be used for the `|alternate-file|`.
+        Supports all the base options (e.g. `buffer_index`, `filetype.enabled`, etc) as well as
+        `modified` and `pinned`.
+      '';
+
+      current = helpers.mkNullOrOption (with types; attrsOf anything) ''
```
MattSturgeon, line 455: Why not use `mkAttrsOf`?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
+              right = "";
+            };
+          }
+          ''
+            The icons which should be used for `|hidden-buffer|`s and `|inactive-buffer|`s.
+            Supports all the base options (e.g. `buffer_index`, `filetype.enabled`, etc) as well as
+            `modified` and `pinned`.
+          '';
+
+      visible = helpers.mkNullOrOption (with types; attrsOf anything) ''
```
MattSturgeon, line 475: Why not use `mkAttrsOf`?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,567 @@ let
...
-        require('barbar').setup(${helpers.toLuaObject setupOptions})
-      '';
+      Otherwise (or in case all letters are already assigned), the behavior is to assign letters in
+      the order of provided to `letters`.
+    '';
+
+    sidebar_filetypes = helpers.defaultNullOpts.mkAttrsOf (
+      with types;
+      either (enum [ true ]) (submodule {
+        options = {
```
MattSturgeon, line 547: Same question re: `freeformType`

MattSturgeon (changes requested):
Some changes, mostly questions and discussion.

plugins/bufferlines/barbar.nix
GaetanLepage: Because I was not able to guess the plugin default value. So I preferred to have nothing.

plugins/bufferlines/barbar.nix
GaetanLepage: Because this can litterally be any of the surrounding options.
The way to declare this cleanly would be to express this strange recursivity. That is what was done prior to this rewrite.

MattSturgeon: Ah ok, IMO this is fine then

plugins/bufferlines/barbar.nix
MattSturgeon: Makes sense. In that case, maybe add a comment to each place where this applies?

plugins/bufferlines/barbar.nix
MattSturgeon: Maybe a comment would be useful for future reference though

plugins/bufferlines/barbar.nix
GaetanLepage: Yes, I would keep it like this for this plugin. Again, the user can always fall back to the top-level `keymaps` option.

plugins/bufferlines/barbar.nix
GaetanLepage: This is what I needed to do to solve https://github.com/nix-community/nixvim/pull/1608#issuecomment-2143498632

MattSturgeon: Strange, but ok. I guess attempting to read the value here causes it to be evaluated? I figured those assertions would only affect _defining_ the option, not _reading_ it :/

GaetanLepage: But the option does exist. It's hidden but it's there. So when you loop over `cfg.keymaps` it's there with its default.
So it is as it was used. Well it **is** used.

plugins/bufferlines/barbar.nix
traxys: This needs a `maybeRaw` no?

plugins/bufferlines/barbar.nix
MattSturgeon: I was under the impression the module system made a distinction between an option existing and a configuration being _defined_. 

I assumed the assertion was just for _defining_ the option, but I guess it's also for _reading_ its value. Makes sense, considering you'd want to throw if someone tries to read a removed option.

Just a shame it also affects reading the (entire) parent attrset, but I can't see how the module system would avoid that.

plugins/bufferlines/barbar.nix
MattSturgeon: `maybeRaw (attrsOf anything)`

plugins/bufferlines/barbar.nix
GaetanLepage: Yes why not.

plugins/bufferlines/barbar.nix
GaetanLepage: Done

plugins/bufferlines/barbar.nix
GaetanLepage: Done. The documentation already mentions "Supports all the base options (e.g. `buffer_index`, `filetype.enabled`, etc) as well as `modified` and `pinned`.".

plugins/bufferlines/barbar.nix
GaetanLepage: Done

plugins/bufferlines/barbar.nix
GaetanLepage: You are right, it doesn't hurt. Otherwise, the user can't set arbitrary sub-options.
Done.

plugins/bufferlines/barbar.nix
GaetanLepage: https://github.com/nix-community/nixvim/pull/1608#discussion_r1623296130

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,592 @@ let
...
+                icon = " ";
+              };
+              "vim.diagnostic.severity.WARN" = {
+                enabled = false;
+                icon = " ";
+              };
+            }
+            ''
+              Set the icon for each diagnostic level.
+              The keys will be automatically translated to raw lua:
```
MattSturgeon, line 365: I'd add a paragraph break, personally:
```suggestion

              The keys will be automatically translated to raw lua:
```

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,592 @@ let
...
+            with types;
+            submodule {
+              freeformType = with types; attrsOf anything;
+              options = {
+                enabled = mkOption {
+                  type = types.bool;
+                  description = helpers.defaultNullOpts.mkDesc true ''
+                    Enables showing git changes of this type.
+                    Requires `|gitsigns.nvim|`.
+                  '';
```
MattSturgeon, line 395: `enabled` and `icon` aren't nullable or maybeRaw. Is that intentional?

Why not just use `defaultNullOpts.mkBool true` here, instead of manually using `mkOption` and `mkDesc`?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,592 @@ let
...
+                };
 
-    maximumLength = helpers.defaultNullOpts.mkInt 30 "Sets the maximum buffer name length.";
+                icon = mkOption {
+                  type = types.str;
+                  example = "";
+                  description = ''
+                    The icon which accompanies the number of git status.
```
MattSturgeon, line 402: Using `mkOption` here to avoid setting `Plugin default`? If so could you add a comment?

Should the type not be `nullOr (maybeRaw str)`? If so, use `helpers.mkNullOrStr`.

plugins/bufferlines/barbar.nix
MattSturgeon: Could you add a comment that you're using `anything` to avoid needing to implement recursive options?

plugins/bufferlines/barbar.nix
```diff
@@ -69,249 +24,592 @@ let
...
-      extraConfigLua = ''
-        require('barbar').setup(${helpers.toLuaObject setupOptions})
-      '';
+      Otherwise (or in case all letters are already assigned), the behavior is to assign letters in
+      the order of provided to `letters`.
+    '';
+
+    sidebar_filetypes = helpers.defaultNullOpts.mkAttrsOf (
+      with types;
+      either (enum [ true ]) (submodule {
```
MattSturgeon, line 570: I think this'll be affected by #1618. I'm not implying you should change the option though.

MattSturgeon (changes requested):

plugins/bufferlines/barbar.nix
GaetanLepage: Yes, this is intentional. Each element of this attrs should be a module with those two keys.

plugins/bufferlines/barbar.nix
GaetanLepage: Yes, I don't see a better alternative anyway...

plugins/bufferlines/barbar.nix
MattSturgeon: In `diagnostics` you use `defaultNullOpts` submodule options:

```nix
mkDiagnosticIconOptions = iconDefault: {
  enabled = helpers.defaultNullOpts.mkBool false ''
    Enable showing diagnostics of this `|diagnostic-severity|` in the 'tabline'.
  '';

  icon = helpers.defaultNullOpts.mkStr iconDefault ''
    The icon which accompanies the number of diagnostics.
  '';
};
```

Why is `gitsigns` different?

If it's just that when all sub-options must be non-null, then we could still support raw types via `maybeRaw`.

Unless there's some other reason not to?

MattSturgeon: Either way, I think there should be a comment to explain why the submodule's options aren't `defaultNullOpts` and/or aren't `maybeRaw`

MattSturgeon: Also, if the option _must_ be present then surely `mkDesc true` makes no sense, since logically a required option can't have a plugin-default.

plugins/bufferlines/barbar.nix
GaetanLepage: I didn't want it to be nullable, but now I'm not sure...

plugins/bufferlines/barbar.nix
GaetanLepage: Is it now OK @MattSturgeon ?

MattSturgeon (approved):
Can't see any remaining issues :+1:

GaetanLepage:
> Can't see any remaining issues 👍

Thanks for the detailed review !

PR #1613, opened 2024-06-01
plugins/lsp/lua-ls: refactor settings options

GaetanLepage:


traxys:
```
 error: An option declaration for `plugins.lsp.servers.lua-ls.settings.runtime.nonstandardSymbol' has type
`lambda' rather than an attribute set.
Did you mean to define this outside of `options'?
```

MattSturgeon:
```
 error: The option `plugins.lsp.servers.lua_ls' does not exist. Definition values:
- In `<unknown-file>':
{
enable = true;
settings = {
addonManager = {
enable = true;
...
```

MattSturgeon (approved):
Can't see any issues

PR #1615, opened 2024-06-01
lib/options: remove `mkStr` assert

MattSturgeon:
Drop assert because it is no longer needed, and also blocks us from passing in raw types.

If a non-string is passed in, it'll be formatted here. This is okay
because `mkDesc` will not re-format strings anyway.

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -168,9 +168,7 @@ rec {
     mkStr =
       # TODO we should delegate rendering quoted string to `mkDefaultDesc`,
       # once we remove its special case for strings.
-      default:
-      assert default == null || isString default;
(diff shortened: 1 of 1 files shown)


lib/options.nix
```diff
@@ -168,9 +168,7 @@ rec {
     mkStr =
       # TODO we should delegate rendering quoted string to `mkDefaultDesc`,
       # once we remove its special case for strings.
```
GaetanLepage, line 170: Should the TODO stay ?

MattSturgeon: Yes. That relates to `mkString` calling `toPretty` instead of delegating that to `mkDesc`.

GaetanLepage (approved):

PR #1614, opened 2024-06-01
plugins/lsp/nil-ls: refactor options

GaetanLepage:

PR #1619, opened 2024-06-02
lib/options: add prime variants for the rest of `defaultNullOpts`

MattSturgeon:
Follow up to #1603

- **plugins: normalise `null` plugin-defaults**
  This first commit was needed to satisfy the assertion I added to `mkEnum'`.
  I could move this to another PR if needed.\
  We could also check each of the files touched to see if other changes ought to be made.

- **lib/options: add more `defaultNullOpts` 'variants**
  The meat of this PR - add a prime variant for each `defaultNullOpts` option factory function.

- **lib/options: add `mkEnum'` argument assertions**
  I also added some assertions to ensure `mkEnum'` is always used correctly. This could be kept or dropped (after running it through CI) depending on preference.

@GaetanLepage It'd be great if you could run CI with these changes on #1614.

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -213,6 +213,9 @@ rec {
           default ? head values,
           ...
         }@args:
+        # `values` is a list and `default` is one of the values (or null)
+        assert isList values;
+        assert default == null || elem default values;
         mkNullableWithRaw' (
           (filterAttrs (n: v: n != "values") args)
           // {
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -76,173 +76,235 @@ rec {
     );
   mkNullOrStrLuaFnOr = type: description: mkNullOrStrLuaFnOr' { inherit type description; };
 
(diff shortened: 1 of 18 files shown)


lib/options.nix
```diff
@@ -76,173 +76,238 @@ rec {
...
+      convertArgs =
+        { default, description, ... }@args:
+        (
+          args
+          // {
+            default = null;
+            description = defaultNullOpts.mkDesc default description;
+          }
+        );
+    in
```
MattSturgeon, line 92: Putting this in a `let` block has caused the rest of the file to be indented. Maybe review with whitespace changes hidden...

Or we could come up with a better name for the helper and expose it publicly (removing the let-block).

MattSturgeon: That said, almost all lines were touched _anyway_, so the extra indent probably isn't a big deal :thinking:

traxys (approved):
LGTM

PR #1620, opened 2024-06-02
plugins/lsp: normalise names & add `serverName` alias

MattSturgeon:
- **plugins/lsp: remove `installLanguageServer` assertion**
- **plugins/lsp: allow aliasing/renaming servers**
- **plugins/lsp: automatically add `serverName` alias**
- **plugins/lsp: normalise kebab-case-names**


MattSturgeon:
I don't think adding an alias to `plugins.lsp.servers.<name>` has any effect on sub-options:

```
cannot find attribute `plugins.lsp.servers.nil_ls.installLanguageServer''
```

GaetanLepage:
This needs to be rebased on `main`.

MattSturgeon:
> I don't think adding an alias to `plugins.lsp.servers.<name>` has any effect on sub-options:
> 
> ```
> cannot find attribute `plugins.lsp.servers.nil_ls.installLanguageServer''
> ```

The error comes from here: https://github.com/nix-community/nixvim/blob/c0a4c2ef6ff7c7869c22d6faf0ab97929f7ac949/plugins/lsp/default.nix#L17-L28

So removing `"nil_ls"` from the `serverName` list works around the issue, but it still leaves me with little confidence the rename is working as intended.

Perhaps the only issue is that the rename isn't done _yet_ when trying to create other rename/removed modules _at `import` time_?

plugins/lsp/default.nix
```diff
@@ -69,7 +69,6 @@ in
         "lua-ls"
         "marksman"
         "metals"
-        "nil_ls"
```
MattSturgeon, line 72: This removal notice is overdue for removal anyway (2024-03-14); should we remove the whole thing instead of just the `nil_ls` case?

Or maybe it's better to wait until after 24.05 before removing any warnings/assertions?

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -7,6 +7,8 @@
 { pkgs, ... }:
 {
   name,
+  aliases ? [ ],
+  oldNames ? [ ],
```
MattSturgeon, line 11: Do you _want_ support for both aliases (no-warning) and renames (warnings), or would you rather we only did renames with warnings?

tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
```diff
@@ -1,5 +1,37 @@
 {
   example = {
+    plugins.lsp = {
+      enable = true;
+
+      servers.nil-ls = {
```
MattSturgeon, line 6: ERROR: [lspconfig] Cannot access configuration for nil-ls. Ensure this server is listed in `server_configurations.md` or added as a custom server.

I guess we need an `upstreamName` or `lspconfigName` option (similar to `mkNeovimPlugin`'s `luaName`).

MattSturgeon: Or, if we want our `plugins.lsp.servers.<name>`s to exactly match upstream's `server_configurations.md` names, perhaps we should drop this PR (until we have an upstream rename to deal with)?

tests/test-sources/plugins/lsp/language-servers/nil-ls.nix
```diff
@@ -1,5 +1,37 @@
...
+              autoEvalInputs = false;
+              nixpkgsInputName = "nixpkgs";
+            };
+          };
+        };
+      };
+    };
+  };
+
+  withAlias = {
```
MattSturgeon, line 34: This test was useful for debugging, but may be overkill to actually be merged?

plugins/lsp/default.nix
MattSturgeon: Removing for now, let me know if you'd prefer a different approach

MattSturgeon:
I've expanded the scope of this PR to rename all snake_case servers to kebab-case.

An alias is now automatically added for any server where `serverName != name`.

plugins/lsp/language-servers/_mk-lsp.nix
GaetanLepage: Right now you have them both enabled ? What's the behavior ?

MattSturgeon: I have both _available_, but only `aliases` is in use. Even that is only used indirectly, via `serverName` automatically getting an alias.

This means users are free to use `name` and `serverName` interchangeably without a warning (docs will still show `name` of course).

Having the `alias` and/or `oldNames` arguments available means a server _could_ explicitly specify additional aliases and/or renames, however this is currently unused and may be overkill.

The automatic `serverName` alias could also be switched to a rename, if a warning is desired.

GaetanLepage: All right. I would go for a rename then. We eventually want to get rid of the rename/warning and maintain a single name.

MattSturgeon: > All right. I would go for a rename then.

Fine with me. Should I simplify `mkLsp` to _only_ do this for the `serverName` case, or should the `alias`/`oldNames` arguments stay for future use?

> We eventually want to get rid of the rename/warning and maintain a single name.

Personally, I don't see the harm in maintaining the rename indefinitely - especially if it's automatic. New users _guessing_ the module name based on lspconfig docs instead of ours may appreciate the warning?

I guess we could replace the warning with an assertion (in the future), but that'd mean implementing our own custom version of `mkRenamedOptionModule`.

GaetanLepage: > Fine with me. Should I simplify mkLsp to only do this for the serverName case, or should the alias/oldNames arguments stay for future use?

I would drop the alias code if it is unused, yes.

> Personally, I don't see the harm in maintaining the rename indefinitely - especially if it's automatic. New users guessing the module name based on lspconfig docs instead of ours may appreciate the warning?

Just leave a TODO for us not to forget that it is there, but let's keep it at least until 24.11.

MattSturgeon: Done

MattSturgeon:
@GaetanLepage I think this should be good to go now

GaetanLepage:
Sorry, I didn't understand your rename. I don't think we should ever remove it. I had not grasped that you have made it automatic regarding the `serverName`.

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -113,20 +113,27 @@ in
...
     [
       (mkRemovedOptionModule (
         basePluginPath ++ [ "extraSettings" ]
       ) "You can use `${basePluginPathString}.extraOptions.settings` instead.")
       (extraConfig cfg)
-    ];
+    ]
+    # Add an alias for the lspconfig server name, if different to `name`.
+    # TODO Should we keep this indefinitely? Added 2024-06-03.
+    # Note: users may use lspconfig's docs to guess the `plugins.lsp.servers.*` name
```
GaetanLepage, line 135:
```suggestion
    # Add an alias for the lspconfig server name, if different to `name`.
    # Note: users may use lspconfig's docs to guess the `plugins.lsp.servers.*` name
```

GaetanLepage (approved):

PR #1622, opened 2024-06-03
plugins/lsp/efmls-configs: cosmetic-only refactoring

GaetanLepage:


traxys (approved):

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -217,19 +214,19 @@ in
   config =
     let
       cfg = config.plugins.efmls-configs;
-      toolAsList = tools: if builtins.isList tools then tools else [ tools ];
+      toolAsList = tools: if isList tools then tools else [ tools ];
```
MattSturgeon, line 217: FYI for future reference:

This can be simplified to `toolsAsList = lib.toList;`

[See docs](https://noogle.dev/f/lib/toList)

PR #1629, opened 2024-06-04
treewide: switch back to alejandra formatting

MattSturgeon:
- **flake-modules: treefmt is formatter by default**
- **flake-modules: switch fmt to alejandra**
- **treewide: reformat with alejandra**
- **.git-blame-ignore-revs: init**

This PR switches back to using Alejandra to format the project, as discussed on matrix.

A git blame ignore-revs-file is included using a filename [recognised by GitHub](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).
To use this locally, you should configure your git config as described in `.git-blame-ignore-revs`'s header comments. Or see [this section of the git manual](https://www.git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt).

In the future, it'd be nice to automatically add that config, e.g. via git-hooks and/or a devshell hook.

I kept the change in flake-modules and the actual treewide re-formatting in separate commits for ease of review, however they can be squashed if preferred.

_Note to self: when squashing or rebasing, remember to update the sha in `.git-blame-ignore-revs`!!_


GaetanLepage:
The changes look correct to me. The `.git-blame-ignore-revs` is a nice addition.

GaetanLepage (approved):

GaetanLepage:
Of course, let's wait for @traxys's confirmation before merging such a change.

lib/types.nix
```diff
@@ -13,73 +15,72 @@ let
...
-    "trace"
-  ];
+    logLevel = types.enum [
+      "off"
+      "error"
+      "warn"
+      "info"
+      "debug"
+      "trace"
+    ];
```
MattSturgeon, line 45: This kind of multi-line list was one of @GaetanLepage's biggest complaints with nixfmt, however alejandra doesn't appear to be any different :confounded:

lib/utils.nix
```diff
@@ -11,69 +12,78 @@ with lib;
     "__empty" = null;
   };
 
-  /**
-    Turn all the keys of an attrs into raw lua.
+  /*
```
MattSturgeon, line 15: Alejandra doesn't support [RFC145](https://github.com/nixos/rfcs/blob/master/rfcs/0145-doc-strings.md#migration-example) docstrings!

plugins/colorschemes/modus.nix
```diff
@@ -12,35 +12,35 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        "tritanopia"
+      ]
+      ''
+        Styles come in four variants:
+
+        - `default` is the plugin's main style designed to cover a broad range of needs.
+        - `tinted` tones down intensity and provides more color variety.
+        - `deuteranopia` is optimized for users with red-green color deficiency.
+        - `tritanopia` is optimized for users with blue-yellow color deficiency.
+      '';
```
MattSturgeon, line 43: Another example of relatively short lists being formatted essentially the same.

MattSturgeon (commented):
After scrolling through a decent portion of the diff and observing formatting differences, I'm less convinced this is a change worth making.

I _do_ think we should discuss it further, but many of the improvements we were expecting from switching back to Alejandra don't seem to have happened.

There are a few cases where the formatting is better, but just as many where it is worse.

GaetanLepage:
This might need a deeper look indeed.
In terms of padding in lists/attrs, I don't really care.
I am not sure whereas the downgrades can be made better (while staying alejandra-conformant).
I know that it is possible to have (not too big) lists displayed on a single line.

Alejandra is probably less strict than nixfmt, so we might have some margin to adapt code here and there. To be confirmed though.

MattSturgeon:
> Alejandra is probably less strict than nixfmt, so we might have some margin to adapt code here and there. To be confirmed though.

I'd be surprised if that is the case, but for sure worth investigating: Alejandra was originally marketed as a fully AST-based (and therefore deterministic) formatter.

MattSturgeon:
> > Alejandra is probably less strict than nixfmt, so we might have some margin to adapt code here and there. To be confirmed though.
> 
> I'd be surprised if that is the case, but for sure worth investigating: Alejandra was [originally marketed](https://discourse.nixos.org/t/the-uncompromising-nix-code-formatter/17385/2?u=kamadorueda) as a fully AST-based (and therefore deterministic) formatter, but that may have changed I guess.

Specifically on lists, I stand corrected. If your input code has the list on a single line, alejandra will avoid splitting it up.

So you could have two identical lists, formatted differently in alejandra.

As an example, this is valid formatted code output by alejandra:

```nix
{
  foo = ["default" "tinted" "deuteranopia" "tritanopia"];
  bar = [
    "default"
    "tinted"
    "deuteranopia"
    "tritanopia"
  ];
}
```

Whether or not this non-determinism is a good or a bad thing, I'm undecided.

MattSturgeon:
As for practical ways we could apply that formatting to 450 files, I'm at a loss.

For files with no changes since 62f32bfc711f0a3c8a52af4c0790345b4b3d2346, we could simply checkout a copy of the file from the parent commit [`62f32bf^`](https://github.com/nix-community/nixvim/commit/c6281260dc074012b54f22f6dba7d6bef72813b6) using a bash script. This should cover most files.

For other files, we'd either have to review every list on a case-by-case basis or write a small single-purpose custom formatting tool that decides based on list length and/or resulting line-length whether to format a list on one line or not.

_If_ we do that, I think it should be done as part of the treewide reformat commit, so that it can be a single entry in the ignore-revs file.

That said, I'm not 100% convinced it is worth the effort.

[This issue](https://github.com/kamadorueda/alejandra/issues/390) in particular is annoying, because it means we can't use [RFC145](https://github.com/nixos/rfcs/blob/master/rfcs/0145-doc-strings.md#migration-example) docstrings.

lib/utils.nix
MattSturgeon: There is a [patch](https://github.com/kamadorueda/alejandra/pull/401) for this issue, though we probably want to avoid compiling a rust program each time we format our code

traxys:
That's one of the reasons I don't like alejandra much, it leaves a huge user choice in the mix, making it difficukt to enforce a style & having loads of PRs with only style reviews.
I'm not really favourable for it, but I am not against it if you decide otherwise

traxys:
For me the _biggest_ use of a formatter is to avoid _any_ style comments in PRs (or at least try to have _almost_ none). I feel like with nixfmt we have had much less of those, and I'd like that to continue, so I guess the options I'd like are either we stay on nixfmt, or we go back to `alejandra` but we mostly don't question how people format their code in accordance to alejandra

GaetanLepage:
> That's one of the reasons I don't like alejandra much, it leaves a huge user choice in the mix, making it difficukt to enforce a style & having loads of PRs with only style reviews.
> I'm not really favourable for it, but I am not against it if you decide otherwise

I agree on the fact that a "strict" formatter is preferable. However, in this case, the question was worth considering as I find the nixfmt style very less readable than alejandra.

But not having to discuss style in PR reviews is a significant benefit of nixfmt... Not easy...

MattSturgeon:
I've marked this as draft since discussion is ongoing and any decisions would likely involve changes to the PR.

I think we're all agreed that a strict formatter is preferable, however we sometimes find specific nixfmt choices less readable than manual formatting.

Frankly, to some degree this is inevitable; the stricter a formatter is the more edge case you'll run into where it doesn't work well for that specific scenario.

For short non-singleton lists, we may be able to patch nixfmt instead of resorting to an entirely different formatter... This would give us the "best of both worlds".

Are short non-singleton lists the only solid example where you prefer alejandra @GaetanLepage?

GaetanLepage:
I am willing to put my preferences aside considering the importance of strict formatting. This is indeed the most important feature of the formatter.
I am OK to let it go and focus on more important matters.
Thank you anyway for the in-depth exploration of the subject.

MattSturgeon:
I'll close this as I think we're all agreed we don't want to do this. Fingers crossed https://github.com/NixOS/nixfmt/issues/206 is accepted upstream or we are able to maintain our own patch.

PR #1632, opened 2024-06-04
fix(hmts): correct error message on no `treesitter`

JarvisCraft:
# Description

This fixes the invalid message in `hmts` plugin whose declaration probably derived from that of the other plugin.

### DIFF
--- a/plugins/languages/treesitter/hmts.nix
+++ b/plugins/languages/treesitter/hmts.nix
@@ -20,7 +20,7 @@ in
 
   config = mkIf cfg.enable {
     warnings = optional (!config.plugins.treesitter.enable) [
-      "Nixvim: treesitter-refactor needs treesitter to function as intended"
+      "Nixvim: hmts needs treesitter to function as intended"
     ];
 
     extraPlugins = [ cfg.package ];


MattSturgeon (approved):
Thanks!

MattSturgeon:
My bad, I should've spotted before merging (no big deal):

Just for future reference, we don't use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), rather we use scope/path-based commit titles. See the [history](https://github.com/nix-community/nixvim/commits/main/) for examples.

JarvisCraft:
> Thanks!

You are welcome :rocket:

> Just for future reference, we don't use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), rather we use scope/path-based commit titles. See the [history](https://github.com/nix-community/nixvim/commits/main/) for examples.
> 
> So the PR/commit should've been something like "plugins/hmts: fix error message on no treesitter"

My bad, I will consider this in my future contributions.

Fell free to reword my commit, if it does not match the scheme of the repository)

MattSturgeon:
> My bad, I will consider this in my future contributions.
> 
> Fell free to reword my commit, if it does not match the naming scheme of the repository)

Not a big deal, I should've checked as part of my review. Can't change it now without force-pushing to main (which is a big no-no).

Thanks again for your contribution!

PR #1634, opened 2024-06-04
flake-modules: minor tweaks from #1629

MattSturgeon:
This PR includes the less controversial commits from #1629, as well as a couple new commits:

- flake-modules: drop flake-root
- flake-modules: explicitly set the same nixfmt for treefmt + pre-commit


GaetanLepage (approved):
Perfect !

PR #1643, opened 2024-06-07
plugins/luasnip: add SnipMate loader

nawordar:
This PR adds the `plugins.luasnip.fromSnipmate` option. Because there was lots of code duplication, I decided to separate loader types from the configuration. Hopefully, that is not too much for a single commit.

### DIFF
--- a/plugins/snippets/luasnip/default.nix
+++ b/plugins/snippets/luasnip/default.nix
@@ -8,6 +8,46 @@
 with lib;
 let
   cfg = config.plugins.luasnip;
+
+  loaderSubmodule = types.submodule {
+    options = {
+      lazyLoad = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether or not to lazy load the snippets
+        '';
+      };
+
+      # TODO: add option to also include the default runtimepath
+      paths =
+        helpers.mkNullOrOption
+          (
+            with helpers.nixvimTypes;
+            oneOf [
+              str
+              path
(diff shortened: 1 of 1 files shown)


plugins/snippets/luasnip/default.nix
```diff
@@ -8,6 +8,51 @@
...
+        '';
+      };
+
+      include = mkOption {
+        type = types.nullOr (types.listOf types.str);
+        default = null;
+        description = ''
+          List of languages to include, by default is not set.
+        '';
+      };
```
GaetanLepage, line 53:
```suggestion
      include = helpers.mkNullOrOption (with helpers.nixvimTypes; maybeRaw (listOf str)) ''
        List of languages to include, by default is not set.
      '';
```

MattSturgeon: Should it be `maybeRaw (listOf (maybeRaw str))`?

GaetanLepage: Haha I knew you would say that !
(it's fine for it to be only a `listOf str`, but yes, it could also accept raw element).

MattSturgeon (commented):
If I'm nitpicking, there's a few options that could use `mkNullOrOption` or similar, but I can't see any major issues. LGTM. Thanks!

GaetanLepage (approved):
LGTM

MattSturgeon (approved):
LGTM! Thanks

PR #1646, opened 2024-06-07
docs/lsp: Display server URLs

MattSturgeon:
- **meta: extend `meta.nixvimInfo` support treewide**
- **docs/lsp: Show LSP server homepage URLs**

image

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -67,17 +67,18 @@ let
           filterAttrs (_: component: component.isOption && (isVisible component)) opts;
       path = removeWhitespace (concatStringsSep "/" path);
       moduleDoc =
-        if builtins.length path >= 2 && lib.hasAttrByPath path nixvimInfo then
-          let
-            info = lib.getAttrFromPath path nixvimInfo;
-            maintainers = lib.unique (options.config.meta.maintainers.${info.file} or [ ]);
-            maintainersNames = builtins.map (m: m.name) maintainers;
-          in
+        let
+          info = optionalAttrs (hasAttrByPath path nixvimInfo) (getAttrFromPath path nixvimInfo);
+          maintainers = lib.unique (options.config.meta.maintainers.${info.file} or [ ]);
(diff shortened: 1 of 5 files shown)


docs/mdbook/default.nix
```diff
@@ -67,12 +67,13 @@ let
...
-            maintainers = lib.unique (options.config.meta.maintainers.${info.file} or [ ]);
-            maintainersNames = builtins.map (m: m.name) maintainers;
-          in
+        let
+          info = optionalAttrs (hasAttrByPath path nixvimInfo) (getAttrFromPath path nixvimInfo);
+          maintainers = lib.unique (options.config.meta.maintainers.${info.file} or [ ]);
+          maintainersNames = builtins.map (m: m.name) maintainers;
+        in
+        # Validate this actually is an info attrset
+        if info ? file && info ? description && info ? url then
```
MattSturgeon, line 76: Alternative: slightly more explicit:

```suggestion
        if attrNames info == [ "file" "description" "url" ] then
```

traxys, line 76: without this you encountered false positives ?

MattSturgeon: Yes, because the _final_ nixvimInfo contains nested attrs, each step along the path will check if it is an info set.

GaetanLepage (approved):
LGTM

docs/mdbook/default.nix
traxys: Yeah I agree that it would be more readable. The other way would be to use a `__nixvim` option that would mark this, but it is uglier

MattSturgeon: On second thoughts, nixfmt breaks this over several lines **and** the list order is actually semantically significant, which could easily turn into a bug one day...

```nix
# nixfmt:
if
  attrNames info == [
    "description"
    "file"
    "url"
  ]
then
```

I think I'll leave it as it was after all.

traxys (approved):

PR #1652, opened 2024-06-08
plugins/wakatime: init

### LINKED ISSUE #1649: [PLUGIN REQUEST] wakatime
| Field    | Description |
|----------|-------------|
| Plugin   | wakatime | 
| Homepage | https://github.com/wakatime/vim-wakatime   | 
| Nixpkgs  | https://github.com/NixOS/nixpkgs/blob/da127f212e0cc14069b0928e196246cddaf2326f/pkgs/applications/editors/vim/plugins/generated.nix#L16268 | 

Context: I am using https://github.com/redyf/Neve/tree/main and it has `wakatime` [installed](https://github.com/redyf/Neve/blob/main/config/utils/wakatime.nix). Actually, I'm using `nixvimExtend` based on Neve where I simply add the stuff I need. I also remove plugins I don't need (e.g. `plugins.neocord.enable = pkgs.lib.mkForce false;` within `nixvimExtend`). I do not need wakatime at all but I can't disable it because it's not a nixvim plugin. So, if nixvim had that plugin, than Neve could use that plugin (`plugin.wakatime = ...` instead of `extraPlugins`), and I could disable it ...

I understand there are various other ways to get what I want but that would be my preferred one and I feel like this could serve the community anyways.

I am happy to provide a minimal PR that only allows to add (i.e. enable) `vim-wakatime` as a plugin. Without any further configuration (which, from how I understand `wakatime` is probably good enough anyways).

GaetanLepage:
Add support for [vim-wakatime](https://github.com/wakatime/vim-wakatime).
This module is quite barebone, but at the same time, there is no need to do much.

Fixes #1649

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -218,6 +218,7 @@
     ./utils/vim-bbye.nix
     ./utils/vim-css-color.nix
     ./utils/vim-matchup.nix
+    ./utils/wakatime.nix
     ./utils/which-key.nix
     ./utils/wilder.nix
     ./utils/yanky.nix
--- a/plugins/utils/wakatime.nix
+++ b/plugins/utils/wakatime.nix
@@ -1,0 +1,15 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.vim-plugin.mkVimPlugin config {
+  name = "wakatime";
+  originalName = "vim-wakatime";
+  defaultPackage = pkgs.vimPlugins.vim-wakatime;
+
+  maintainers = [ maintainers.GaetanLepage ];
+}
--- a/tests/test-sources/plugins/utils/wakatime.nix
+++ b/tests/test-sources/plugins/utils/wakatime.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.wakatime.enable = true;
+  };
+}


GaetanLepage:
Note: the `wakatime-cli` dependency gets downloaded automatically by the plugin.
This does not even pose problem on NixOS as it is a statically linked binary.

MattSturgeon:
> Note: the `wakatime-cli` dependency gets downloaded automatically by the plugin. This does not even pose problem on NixOS as it is a statically linked binary.

I'm sure you're aware already, but the cli [is available in nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/wa/wakatime-cli/package.nix#L52), we could include that version if there is a way to disable the vim plugin's autodownload.

If it's entirely statically linked that's not a big deal though.

GaetanLepage:
I agree, we could, but is it worth the hassle ?
I have checked, and the plugin does work fine with the auto-install.

GaetanLepage:
Maybe we could have an option to let the user choose...

MattSturgeon (approved):
> I agree, we could, but is it worth the hassle ?
> I have checked, and the plugin does work fine with the auto-install.

If everything works fine then this isn't a major concern for now.

> Maybe we could have an option to let the user choose...

Some purists may like to have _as much as possible_ installed in a deterministic way at build time, however that's something we could revisit later.

LGTM!

GaetanLepage:
> Some purists may like to have _as much as possible_ installed in a deterministic way at build time, however that's something we could revisit later.

Definitely !

PR #1657, opened 2024-06-09
lib/helpers: add `deprecation.mkTransitionOptionModule`

MattSturgeon:
`helpers.deprecation.mkTransitionOptionModule` is similar to `lib.mkRenamedOptionModule`, however it instead warns users that the option was renamed to avoid confusion, and will eventually be transitioned to the new option.

I've tested this by adding the following to surround.vim's `imports` and changing `name` to `"surround-vim"`:

```nix
imports = [
  (helpers.deprecation.mkTransitionOptionModule {
    from = [ "plugins" "surround" ];
    to = [ "plugins" "surround-vim" ];
    future = [ "plugins" "surround-nvim" ];
    takeover = "2024-09-09";
  })
];
```

This produces the following warning when used with a config that contains `plugins.surround.enable = true`:
```
trace: warning: The option `plugins.surround' defined in `/nix/store/lylx2xpibp09k24ihrkf8pyn22kv0kcf-source/nvim/config/main.nix' has been renamed to `plugins.surround-vim'.
This has been done to avoid confusion with the new option `plugins.surround-nvim'.

At some point after 2024-09-09, `plugins.surround-nvim' will be moved to `plugins.surround'.

```

It's a shame multiline trace warnings are so ugly, so I'd appreciate suggestions for making it more readable. The first line of the warning is based on the warning used by `lib.doRename`.

### DIFF
--- a/lib/deprecation-helpers.nix
+++ b/lib/deprecation-helpers.nix
@@ -1,0 +1,82 @@
+{ lib, nixvimUtils, ... }:
+{
+  /**
+    Produce a NixOS Module that warns users the option path is transitioning from one option to another.
+
+    The primary use case is for adding a plugin variant,
+    with the intention of it using the "primary" option path for that plugin.
+
+    It is recommended that after transitioning, `lib.mkAliasOptionModule` is used,
+    to avoid breaking configs that used the `to` path.
+
+    If there's no plan to transition, just use `lib.mkRenamedOptionModule`.
+
+    # Example
+    ```nix
+    mkTransitionOptionModule {
+      from = [ "plugins" "surround" ];
+      to = [ "plugins" "surround-vim" ];
+      future = [ "plugins" "surround-nvim" ];
+      takeover = "24.11";
+    }
+    => <nixos-module>
+    ```
+
+    # Type
+    ```
+    mkTransitionOptionModule :: AttrSet -> NixosModule
+    ```
+  */
+  mkTransitionOptionModule =
+    {
+      # The path `to` used to be found at and `future` will eventually be moved to.
+      from,
+      # The option the alias points to, previously found at `from`.
+      to,
+      # The new option; will takeover `from` in the future
+      future,
+      # The date or release after which `future` will be moved to `from`.
+      takeover ? lib.trivial.release,
+    }:
+    # Return a module
+    {
+      lib,
+      options,
+      config,
+      ...
+    }:
+    let
+      inherit (lib)
+        setAttrByPath
+        getAttrFromPath
+        attrByPath
+        showFiles
+        showOption
+        optional
+        ;
+      opt = getAttrFromPath from options;
+      toOpt = getAttrFromPath to options;
+    in
+    {
+      options = setAttrByPath from (
+        lib.mkOption {
+          description = "Alias of `${showOption to}`.";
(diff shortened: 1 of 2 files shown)


traxys:
Maybe we should use NixOS releases as takeovers globally in nixvim?

MattSturgeon:
> Maybe we should use NixOS releases as takeovers globally in nixvim?

I agree, I think that'll make things easier for stable users upgrading and we'll also have an obvious reminder of _when_ we need to go through and remove old deprecations.

Although, personally I'm in favor of _not_ removing old deprecations where possible.\* Instead we could move them somewhere less obnoxious, like `./warnings.nix` files?

\* e.g. you can't avoid a _transition_ warning needing to be removed when `future` takes over `from`'s path.
\* or if we can't find a less obnoxious way to maintain the warning (e.g. `keymaps.*.lua`), then removal is fine.

lib/helpers.nix
```diff
@@ -15,6 +15,7 @@ in
   maintainers = import ./maintainers.nix;
   keymaps = import ./keymap-helpers.nix { inherit lib nixvimOptions nixvimTypes; };
   autocmd = import ./autocmd-helpers.nix { inherit lib nixvimOptions nixvimTypes; };
+  deprecation = import ./deprecation-helpers.nix { inherit lib nixvimUtils; };
```
MattSturgeon, line 18: I've scoped this for now. Would you rather it be a top-level helper?

lib/deprecation-helpers.nix
```diff
@@ -0,0 +1,82 @@
...
+        setAttrByPath
+        getAttrFromPath
+        attrByPath
+        showFiles
+        showOption
+        optional
+        ;
+      opt = getAttrFromPath from options;
+      toOpt = getAttrFromPath to options;
+    in
```
MattSturgeon, line 59: This should produce useful error messages when used incorrectly:

```suggestion
      get = path: attrByPath path (abort "mkTransitionOptionModule: can't find option `${showOption path}`") options;
      opt = get from;
      toOpt = get to;
    in
    # Also assert `future` exists
    # get will abort before the assert does, only used to force it to evaluate
    assert (get future) != null;
```

lib/deprecation-helpers.nix
```diff
@@ -0,0 +1,82 @@
...
+  */
+  mkTransitionOptionModule =
+    {
+      # The path `to` used to be found at and `future` will eventually be moved to.
+      from,
+      # The option the alias points to, previously found at `from`.
+      to,
+      # The new option; will takeover `from` in the future
+      future,
+      # The date or release after which `future` will be moved to `from`.
```
MattSturgeon, line 38: See https://github.com/nix-community/nixvim/pull/1657#issuecomment-2158232702
```suggestion
      # The nixos release after which `future` will be moved to `from`.
```
?

MattSturgeon (commented):
Left some comments for discussion.

I don't think this should be merged until _after_ we all agree on how we should handle the kinda name conflicts this helper attempts to resolve (i.e. #1371 and #1352). Otherwise we've just got dead code laying around!

Happy to discuss here or on matrix.

lib/helpers.nix
traxys: I think scoped is nicer

PR #1661, opened 2024-06-09
lib/options: refactor and rename (non-breaking)

MattSturgeon:
#### lib/types: add `nixvimTypes.freeformModule`
We often have submodules in `settings` options, most of these should be freeform although many aren't.

#### lib/options: add `rawOpts` helpers
New helpers for making non-null options that support the raw type.

#### lib/options: move top-level helpers to `nullableOpts`
Most of the top-level option helpers were actually creating nullable options. Move them into their own scope.

#### lib/options: standardise raw-type support in `defaultNullOpts`
`mkAttrsOf` (etc) didn't inject `maybeRaw` everywhere they should. Also begin transitioning `mkNullable` and `mkNullableWithRaw` by introducing `mkNullableNoRaw` alias.

#### lib/options: merge `mkNullOrStrLuaOr'` (etc)
This is the only breaking change, although I don't think anyone ever used these.

#### Other
Do we still need `mkCompositeOption`? Should we make it freeform?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,8 @@ These option declarations should be in `settingsOptions` and their names should 
 There are a number of helpers to help you correctly implement them:
 
 - `helpers.defaultNullOpts.{mkBool,mkInt,mkStr,...}`: This family of helpers takes a default value and a description, and sets the Nix default to `null`. These are the main functions you should use to define options.
-- `helpers.defaultNullOpts.mkNullable`: This takes a type, a default and a description. This is useful for more complex options.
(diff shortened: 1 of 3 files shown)


GaetanLepage:
> Do we still need `mkCompositeOption`? Should we make it freeform?

We should first look for all the occurences in the code base and then, if none of this use cases is legit, we should remove it.

GaetanLepage:
Although your description is very clear, I would prefer if this set of changes is spread accross a few individual PRs.

MattSturgeon:
Closing as stale

PR #1665, opened 2024-06-09
treewide: transition away from plugin-default strings

MattSturgeon:
- Replace all instances of pre-formatted plugin-default string values
  with actual nix values.\*
- Replace many instances of `mkNullable` (etc) with more appropriate
  helpers.
- Move some option examples to use `mkOption`s `example` attribute.
- Some minor cleanup in edited files.

\* Hopefully all of them, anyway!

I can only apologize for the size of the diff :sob:


plugins/utils/better-escape.nix
```diff
@@ -28,19 +28,17 @@ in
...
-          Example (recommended):
+      Example (recommended):
 
-          keys.__raw = \'\'
-            function()
-              return vim.api.nvim_win_get_cursor(0)[2] > 1 and '<esc>l' or '<esc>'
-            end
-          \'\';
-        '';
+      keys.__raw = '''
```
GaetanLepage, line 36:
```suggestion
      keys.__raw = '''
```
Why 3 simple quotes here ?

plugins/ui/noice.nix
```diff
@@ -132,34 +180,30 @@ with lib;
...
             ''
               Lsp Progress is formatted using the builtins for lsp_progress
             '';
         formatDone =
-          helpers.defaultNullOpts.mkNullable (types.either types.str types.anything) ''"lsp_progress"''
+          helpers.defaultNullOpts.mkNullable (types.either types.str types.anything) "lsp_progress"
             "";
 
-        throttle = helpers.defaultNullOpts.mkNum "1000 / 30" "frequency to update lsp progress message";
+        throttle = helpers.defaultNullOpts.mkNum (1000 / 30) "frequency to update lsp progress message";
```
GaetanLepage, line 191:
```suggestion
        throttle = helpers.defaultNullOpts.mkNum (1000 / 30) "frequency to update lsp progress message";
```
Wouldn't this appear as the result of 1000 / 30 (i.e. 100) ?

plugins/lsp/lspsaga.nix
```diff
@@ -201,19 +205,19 @@ in
...
           `max_height` of the finder window (float layout).
         '';
 
-        leftWidth = helpers.defaultNullOpts.mkNullable (types.numbers.between 0.0 1.0) "0.3" ''
+        leftWidth = helpers.defaultNullOpts.mkNullable (types.numbers.between 0.0 1.0) 0.3 ''
           Width of the left finder window (float layout).
         '';
 
-        rightWidth = helpers.defaultNullOpts.mkNullable (types.numbers.between 0.0 1.0) "0.3" ''
+        rightWidth = helpers.defaultNullOpts.mkNullable (types.numbers.between 0.0 1.0) 0.3 ''
```
GaetanLepage, line 216: Maybe we should add a `mkNumber01` helper at some point...

plugins/utils/harpoon.nix
```diff
@@ -144,9 +144,16 @@ in
         Menu window height
       '';
 
-      borderChars = helpers.defaultNullOpts.mkNullable (
-        with types; listOf str
-      ) ''["─" "│" "─" "│" "╭" "╮" "╯" "╰"]'' "Border characters";
+      borderChars = helpers.defaultNullOpts.mkNullable (with types; listOf str) [
```
GaetanLepage, line 147:
```suggestion
      borderChars = helpers.defaultNullOpts.mkListOf types.str [
```
?

MattSturgeon: Good catch.

Eventually I lost motivation for changing _all_ cases of `mkNullable listOf` to `mkListOf`, but I can give this another pass if needed.

plugins/lsp/lspsaga.nix
MattSturgeon: And/or `mkBetween`?

plugins/ui/noice.nix
MattSturgeon: Yes. Without adding "pluginDefaultText", or some other way to pass pre-formatted text, I couldn't see an alternative.

plugins/utils/better-escape.nix
MattSturgeon: Within a double-quote indented string, triple-quotes is the correct way to render a double-quote.

The previous code rendered as `\'\'`.

MattSturgeon: Well, [it was ok here](https://nix-community.github.io/nixvim/plugins/better-escape.html#pluginsbetter-escapekeys), but this was definitely necessary elsewhere where I made this change. Maybe being within a markdown code block makes a difference.

plugins/ui/noice.nix
MattSturgeon: Maybe we could add a special case in `mkDesc` for `literExpression` types.

That might be easier to use with our existing helpers rather than forcing the use of `defaultText` 🤔

plugins/languages/rust/rust-tools.nix
```diff
@@ -58,26 +58,44 @@ in
...
 
-      maxWidth =
-        helpers.defaultNullOpts.mkNullable types.int null
-          "Maximal width of the hover window. null means no max.";
+      maxWidth = helpers.defaultNullOpts.mkInt null "Maximal width of the hover window. null means no max.";
 
-      maxHeight =
-        helpers.defaultNullOpts.mkNullable types.int null
-          "Maximal height of the hover window. null means no max.";
+      maxHeight = helpers.defaultNullOpts.mkInt null "Maximal height of the hover window. null means no max.";
```
GaetanLepage, line 98:
```suggestion
      maxHeight = helpers.defaultNullOpts.mkInt null ''
        Maximal height of the hover window. null means no max.
      '';
```

MattSturgeon: This was only moved onto one line because we no longer hit the max line length (`mkInt` is shorter than `mkNullable types.int`).

I know we usually use indented strings, but this file predominantly uses ordinary string literals and I was trying to minimize changes.

plugins/bufferlines/bufferline.nix
```diff
@@ -104,26 +104,25 @@ in
...
-            ```
+          "ordinal"
+          "buffer_id"
+          "both"
+        ];
+        default = "none";
+        description = ''
+          Customize the styling of numbers.
+
+          Either one of ${concatStringsSep " " (map (v: ''`"${v}"`'') values)} or a lua function.
```
MattSturgeon, line 118: Could probably drop this line, the example should make things clear:
image

plugins/completion/cmp/options/settings-options.nix
```diff
@@ -290,10 +302,16 @@ with lib;
...
-            "";
+        border = helpers.defaultNullOpts.mkBorder [
+          ""
+          ""
+          ""
+          " "
+          ""
+          ""
+          ""
+          " "
```
MattSturgeon, line 313: Are the spaces erroneous?

plugins/completion/cmp/sources/cmp-git.nix
MattSturgeon, line 1: I added some FIXMEs to this file as I was unsure if some types should change to `strLuaFn`

MattSturgeon:
Making this a draft. I'll try and tackle specific parts of this PR individually if feasible.

PR #1666, opened 2024-06-09
misc: ensure all options have a description

GaetanLepage:


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -55,6 +55,12 @@ in
       cmd = mkOption {
         type = with types; nullOr (listOf str);
         default = if (cfg.package or null) != null then cmd cfg else null;
+        description = ''
+          A list where each entry corresponds to the blankspace delimited part of the command that
+          launches the server.
+          The first entry is the binary used to run the language server.
+          Additional entries are passed as arguments.
```
MattSturgeon, line 62:
```suggestion
          A list where each entry corresponds to the blankspace delimited part of the command that
          launches the server.

          The first entry is the binary used to run the language server.
          Additional entries are passed as arguments.
```

plugins/utils/hydra/hydras-option.nix
```diff
@@ -112,6 +112,10 @@ in
 mkOption {
   type = types.listOf hydraType;
   default = [ ];
+  description = ''
+    A list of hydra configurations.
+    See [here](https://github.com/nvimtools/hydra.nvim?tab=readme-ov-file#creating-a-new-hydra).
```
MattSturgeon, line 117: :heart:

plugins/utils/nix-develop.nix
```diff
@@ -16,13 +16,26 @@ in
...
+        SHELL = true;
+        TMP = true;
+      };
     };
 
     separatedVariables = mkOption {
-      type = types.attrsOf types.str;
+      type = with types; attrsOf str;
       default = { };
+      description = "An attrs specifying which separator to use for concatanating path values.";
```
MattSturgeon, line 34:
```suggestion
      description = "An attrs specifying the separator to use for particular environment variables.";
```

modules/filetype.nix
```diff
@@ -17,7 +17,20 @@ let
...
+            priority = mkOption {
+              type = ints.unsigned;
+              description = ''
+                Filename patterns can specify an optional priority to resolve cases when a file path
+                matches multiple patterns.
+
+                Higher priorities are matched first.
+                When omitted, the priority defaults to 0.
+
+                A pattern can contain environment variables of the form "${SOME_VAR}" that will be
```
MattSturgeon, line 29:
```suggestion
                A pattern can contain environment variables of the form `"''${SOME_VAR}"` that will be
```
This will be interpolated as a nix expression without escaping the `$` as `''$`. Not sure why CI isn't complaining though, am I missing something?

Also, I'd add backticks, personally.

MattSturgeon (approved):
Looks good!

PR #1667, opened 2024-06-10
lib/options: move "plugin default" into `defaultText`

MattSturgeon:
### Summary

Introduce a **new** `helpers.pluginDefaultText` and **deprecate** `helpers.defaultNullOpts.mkDesc`.

### Motivation

Displaying the "plugin default" within `defaultText` ensures that both defaults are grouped together in the docs.

### Related changes
This blocks #1665:
- That PR could also do a treewide migration away from `mkDesc`
- Pre-formatted defaults (e.g. `1000 / 30`) could be passed as plugin-`defaultText`
- That PR should also remove the special case `pluginDefaultText` inherited from `mkDesc`

### Rendered docs
<details><summary>Options <strong>affected</strong> by this PR</summary>
<p>

image

</p>
</details>

<details><summary>Option manually adding "plugin default" (currently <strong>not</strong> affected)</summary>
<p>

image

</p>
</details> 

<details><summary>Option using deprecated <code>mkDesc</code> (currently <strong>not</strong> affected)</summary>
<p>

image

</p>
</details>

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -6,6 +6,65 @@
 with lib;
 with nixvimUtils;
 rec {
+  # Render a plugin default string
+  pluginDefaultText =
+    let
+      # Assume a string `default` is already formatted as intended,
+      # TODO: remove this behavior so we can quote strings properly
+      # historically strings were the only type accepted by mkDesc.
+      legacyRenderOptionValue =
+        v:
+        literalExpression (
+          if isString v then
+            v
+          else
+            generators.toPretty {
+              allowPrettyValues = true;
+              multiline = true;
+            } v
+        );
+    in
+    {
+      # plugin default: any value or literal expression
+      pluginDefault,
+      # nix option default value, used if `nixDefaultText` is missing
+      default ? null,
+      # nix option default string or literal expression
+      defaultText ? (options.renderOptionValue default) // {
+        __lang = "nix";
+      },
+      ...
+    }:
+    let
+      # Only add `__lang` if `pluginDefault` is not already a literal type
+      pluginDefaultText =
+        if pluginDefault ? _type && pluginDefault ? text then
+          pluginDefault
+        else
+          (legacyRenderOptionValue pluginDefault) // { __lang = "nix"; };
+
+      # Format text using markdown code block or inline code
+      # Handle `v` being a literalExpression or literalMD type
+      toMD =
+        v:
+        let
+          value = options.renderOptionValue v;
+          multiline = hasInfix "\n" value.text;
+          lang = value.__lang or ""; # `__lang` is added internally when parsed in argument defaults
+        in
+        if value._type == "literalMD" then
+          if multiline then "\n${value.text}" else " ${value.text}"
+        else if multiline then
+          "\n```${lang}\n${value.text}\n```"
+        else
+          " `${value.text}`";
+    in
+    literalMD ''
+      ${toMD defaultText}
+
+      _Plugin default:_${toMD pluginDefaultText}
+    '';
+
   # Creates an option with a nullable type that defaults to null.
   mkNullOrOption' =
     {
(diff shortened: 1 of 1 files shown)


lib/options.nix
```diff
@@ -6,6 +6,34 @@
...
+      nixDefaultText ? generators.toPretty { } nixDefault,
+      ...
+    }:
+    let
+      # Format text using markdown code block or inline code
+      # TODO support `s` being a literalExpression or literalMD type
+      toMD = s: if hasInfix "\n" s then "\n\n```nix\n${s}\n```" else "`${s}`";
+    in
+    literalMD ''
+      ${toMD nixDefaultText}
```
MattSturgeon, line 32: Upstream's docs rendering would actually pass `defaultText` through `toPretty` and then wrap it in `literalExpression`. (Unless `defaultText` is already a `literalExpression` or `literalMD` type).

This means passing `defaultText` will usually render as a quoted string, unlike my implementation.

It might be better to more closely replicate the upstream implementation, so that the behavior isn't surprising. I guess set the fallback value to something like:
```nix
# legacy isString special-case omitted for clarity
defaultText ? literalExpression (generators.toPretty { } default),
nixDefaultText ? literalExpression (generators.toPretty { } nixDefault),
```

And then correctly handle literal expressions in the main implementation.

MattSturgeon: I also noticed upstream pass `allowPrettyValues = true` to `toPretty`, this enables support for embedded custom rendering. Any value of the type `{ __pretty = fn; val = …; }` will use the `__pretty` function to pretty print `val`.

e.g. you could nest `{ __pretty = v: v; val = "(raw nix expression)"; }` somewhere in a list/set, and it'd be pretty-printed as `(raw nix expression)`!

lib/options.nix
```diff
@@ -6,6 +6,34 @@
...
+      # pre-formatted plugin default string
+      #
+      # Assume a string `default` is already formatted as intended,
+      # TODO: remove this behavior so we can quote strings properly
+      # historically strings were the only type accepted by mkDesc.
+      defaultText ? if isString default then default else generators.toPretty { } default,
+      # nix option default value, used if `nixDefaultText` is missing
+      nixDefault ? null,
+      # pre-formatted nix option default string
+      nixDefaultText ? generators.toPretty { } nixDefault,
```
MattSturgeon, line 23: I think these names would be clearer if we used `pdefault` or `pluginDefault`:

- `default`: the actual value used wen option is missing (e.g. `null`)
- `defaultText`: any value, or literal expression, to display instead of `default`
- `pluginDefault`: any value, or literal expression, to display as **"plugin default"**

lib/options.nix
```diff
@@ -79,55 +138,39 @@ rec {
...
         (
-          args
+          # TODO filter pluginDefault
+          (filterAttrs (
+            n: _:
+            !(elem n [
+              "pluginDefault"
+              "defaultText"
+            ])
+          ) args)
```
MattSturgeon, line 154: Once `defaultNullOpts` migrates `default` -> `pluginDefault`, this can be simplified:

```suggestion
          assert !(args ? default || args ? defaultText);
          (filterAttrs (n: _: n != "pluginDefault") args)
```

lib/options.nix
MattSturgeon: This is resolved, but for now I'll leave the thread visible for context.

lib/options.nix
MattSturgeon: Partially resolved: `defaultNullOpts` still uses `default`, but `convertArgs` translates that into `pluginDefaults`.

Eventually, I think we should migrate `defaultNullOpts`'s `default` -> `pluginDefault` too.

GaetanLepage: > I think these names would be clearer if we used `pdefault` or `pluginDefault`:
> 
>     * `default`: the actual value used wen option is missing (e.g. `null`)
> 
>     * `defaultText`: any value, or literal expression, to display instead of `default`
> 
>     * `pluginDefault`: any value, or literal expression, to display as **"plugin default"**

I agree with this naming.



> I think trying to use `default` to mean "plugin default" in the `defaultNullOpts` context has caused unnecessary confusion.

Yes, that's not optimal.

GaetanLepage (approved):
This is a very detailed and careful work.
I won't lie, this added complexity is quite frightening though. Also, the drift from nix's "default" tools is not amazingly comfortable.
At the same time, _it is already_ too late, and those helpers do bring value, so... let's do it.

The diff looks good to me and I trust you in maintaining those helpers.

GaetanLepage:
Let's wait for @traxys' review though.

traxys (approved):

PR #1668, opened 2024-06-10
plugins/specs: switch to mkNeovimPlugin

GaetanLepage:
That was an ancient implementation.


plugins/ui/specs.nix
```diff
@@ -0,0 +1,139 @@
...
+      ) "Please, use `settings.popup.fader` directly.")
+      (mkRemovedOptionModule (
+        basePluginPath ++ [ "resizer" ]
+      ) "Please, use `settings.popup.resizer` directly.")
+      (mkRemovedOptionModule (
+        basePluginPath ++ [ "ignored_filetypes" ]
+      ) "Please, use `settings.ignore_filetypes` instead.")
+      (mkRemovedOptionModule (
+        basePluginPath ++ [ "ignored_buffertypes" ]
+      ) "Please, use `settings.ignore_buftypes` instead.")
```
MattSturgeon, line 57: Are these all incompatible with the old options? _At first glance_ some of them look like they could be rename warnings instead of removal errors.

plugins/ui/specs.nix
```diff
@@ -0,0 +1,139 @@
...
+      settingsPath = basePluginPath ++ [ "settings" ];
+      renameToPopup =
+        old: new:
+        mkRenamedOptionModule (basePluginPath ++ [ old ]) (
+          settingsPath
+          ++ [
+            "popup"
+            new
+          ]
+        );
```
MattSturgeon, line 36:
```suggestion
        mkRenamedOptionModule
        (basePluginPath ++ [ old ])
        (settingsPath ++ [ "popup" new ]);
```

:sob:

plugins/ui/specs.nix
```diff
@@ -0,0 +1,139 @@
...
+      delay_ms = helpers.defaultNullOpts.mkUnsignedInt 10 ''
+        Delay before popup displays.
+      '';
+
+      inc_ms = helpers.defaultNullOpts.mkUnsignedInt 5 ''
+        Time increments used for fade/resize effects.
+      '';
+
+      blend = helpers.defaultNullOpts.mkUnsignedInt 10 ''
+        Starting blend, between 0-100 (fully transparent), see `:h winblend`.
```
MattSturgeon, line 79:
```suggestion
        Starting blend, between 0 (opaque) and 100 (transparent), see `:h winblend`.
```

MattSturgeon (changes requested):
Looks good. I've only left questions and nits.

plugins/ui/specs.nix
GaetanLepage: Don't start me on that ^^

plugins/ui/specs.nix
GaetanLepage: Done

plugins/ui/specs.nix
GaetanLepage: No, they were all declared in a custom way before. Switching to a direct (RFC-42) style is breaking backward compatibility.

MattSturgeon (approved):
Looks good!

PR #1670, opened 2024-06-11
Allow `pluginDefault` arg outside of `defaultNullOpts` + remove `mkDesc`

MattSturgeon:
Following up on #1667, this PR allows the `pluginDefault` argument to be passed to any of nixvim'd option factory helpers.

Taking advantage of this where possible, the 3 internal uses of `mkDesc` have been migrated either to `pluginDefault` or calling `helpers.pluginDefaultText` directly.

It might make sense to have a `helpers.mkOption`, which simply extends `lib.mkOption` with support for custom nixvim args like `pluginDefault`:

```nix
mkOption = args: lib.mkOption (processNixvimArgs args);
```

Let me know if you'd like that included in this PR, and I'll be able to simplify the plugins/barbar implementation.
Maybe we'd also consider no longer exposing `pluginDefaultText` publicly, if all use-cases are accommodated by a `mkOption` wrapper?

- **lib/options: remove `defaultNullOpts.mkDesc`**
- **lib/options: allow `pluginDefault` in any helper**
- **plugins/edgy: switch from `mkDesc` to `pluginDefault`**
- **plugins/barbar: switch from `mkDesc` to `pluginDefaultText`**

I was originally going to tackle this as part of #1665, however that PR is already _way_ too big!

### DIFF
--- a/plugins/ui/edgy.nix
+++ b/plugins/ui/edgy.nix
@@ -120,7 +120,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
         '' "Callback for the ending of animations.";
 
         # This option accepts an attrs or a lua string.
-        # Hence, we use `mkOption` to convert the string to raw lua in `apply`.
+        # Hence, we convert the string to raw lua in `apply`.
         spinner =
           let
             defaultFrames = [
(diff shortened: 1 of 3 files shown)


plugins/ui/edgy.nix
```diff
@@ -182,84 +181,82 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
 
       # This option accepts an attrs or a lua string.
-      # Hence, we use `mkOption` to convert the string to raw lua in `apply`.
-      keys = mkOption {
+      # Hence, we convert the string to raw lua in `apply`.
+      keys = helpers.mkNullOrOption' {
         type = with helpers.nixvimTypes; attrsOf (either strLuaFn (enum [ false ]));
-        default = { };
+        default = null;
         apply = mapAttrs (_: v: if isString v then helpers.mkRaw v else v);
```
MattSturgeon, line 188: I believe this is causing the CI failure, since I made the type nullable...
```suggestion
        apply =
          x:
          if x == null then null
          else mapAttrs (_: v: if isString v then helpers.mkRaw v else v);
```
Or just revert to non-nullable and call `helpers.pluginDefaultText` directly\
or implement the `mkOption` wrapper mentioned in the PR description.

GaetanLepage:
> It might make sense to have a `helpers.mkOption`, which simply extends `lib.mkOption` with support for custom nixvim args like `pluginDefault`:

That would be a good idea indeed, but for another PR.

GaetanLepage (approved):
All the commits look fine to me.

PR #1676, opened 2024-06-11
colorschemes/base16: add settings + refactor to fix #1675

### LINKED ISSUE #1675: [BUG] List of available base16 colorschemes is not up to date
| Plugin       | `base16`  |
| Nixpkgs      | `unstable` |
| Home Manager | `unstable` |

## Description

Hey all,

The base16 plugin allows users to select from an already existing set of colorschemes. Nixvim then checks if this is a valid choice by verifying that this choice is inside the [theme-list.nix](https://github.com/nix-community/nixvim/blob/main/plugins/colorschemes/base16/theme-list.nix) file. 

This list is outdated, and several choices that are available through the plugin will be rejected by nixvim even if they are available, for example the ``selenized`` colorscheme family (example will use ``selenized-light``).

The "missing" colorschemes themselves will be loaded without error and can even be selected in Nixvim after it has been loaded through a simple
```vim
:colorscheme base16-selenized-light 
```

However, the MRE will produce the following error (list of colorschemes shortened)

```
error:

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/rljkm6h18fjvavm6l17jiil8glphc44f-source/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'nativeBuildInputs' of derivation 'nix-shell'

         at /nix/store/rljkm6h18fjvavm6l17jiil8glphc44f-source/pkgs/stdenv/generic/make-derivation.nix:375:7:

          374|       depsBuildBuild              = elemAt (elemAt dependencies 0) 0;
          375|       nativeBuildInputs           = elemAt (elemAt dependencies 0) 1;
             |       ^
          376|       depsBuildTarget             = elemAt (elemAt dependencies 0) 2;

       error: A definition for option `colorschemes.base16.colorscheme' is not of type `one of "3024", "apathy", "apprentice", "ashes", [...], "xcode-dusk", "zenburn" or (submodule)'. Definition values:
       - In `<unknown-file>': "selenized-light"
```

despite ``selenized-light`` being in the [plugins repository](https://github.com/RRethy/base16-nvim/blob/master/colors/base16-selenized-light.vim).

The solution is to simply update the themes-list.nix file. I have prepared an updated version of this file and uploaded it [here](https://pastebin.com/mua2uhVe).

```nix
programs.nixvim = {
  colorschemes.base16 = {
    enable = true;
    colorscheme = "selenized-light";
  };
}
```

MattSturgeon:
Removed the enum restriction on the `colorscheme` option, allowing anything to be passed through, including raw lua. This fixes #1675

## Settings

Added support for the plugin's "advanced" config settings and simplified the `extraConfig` implementation.

As some of their API is undocumented, I've opened https://github.com/RRethy/base16-nvim/pull/102. Should we wait for acknowledgement?

## Enum

Previously, we listed all builtin base16 colorschemes in [theme-list.nix](https://github.com/nix-community/nixvim/blob/cc9023fb1d74fad3b7b704a1c161a2ce9f378431/plugins/colorschemes/base16/theme-list.nix), however this means we have to maintain that list, ensuring it tracks any upstream changes.

Also worth noting that upstream don't always have their docs up to date. I've opened a PR to hopefully address this: https://github.com/RRethy/base16-nvim/pull/103.

Having a ~240 value enum type wasn't particularly readable though, anyway...

<details><summary>Current documentation (screenshot)</summary>
<p>

image

</p>
</details> 

## Refactor

I've cleaned up some of the implementation. In particular, I found a more readable way to define `customColorschemeType` ~~using [`stringToCharacters`](https://noogle.dev/f/lib/strings/stringToCharacters) and [`genAttrs`](https://noogle.dev/f/lib/attrsets/genAttrs)~~ _`mapAttrs` on the example attrs_.

## Rename

EDIT: this section is irrelevant now that we know the lua API (`setup`) doesn't use the `base16-` prefix.

<p>

I [toyed with](https://github.com/nix-community/nixvim/commit/f260687acc517f446d7e5dc07695a742b72ee4c9) renaming the option to `colors` to match upstream's internal name, and to allow automatically apply the prefix to be done [_only_ when the _old name_ was used](https://github.com/nix-community/nixvim/commit/f260687acc517f446d7e5dc07695a742b72ee4c9#diff-53c30d83d98f9af579a2b24e65b5be5697d30adcf478c7cda416f0db4b53ef79R170).\
In the end I decided the rename wasn't worth it. Let me know if you disagree.

</p>
</details>


plugins/colorschemes/base16/default.nix
```diff
@@ -37,94 +42,147 @@ helpers.vim-plugin.mkVimPlugin config {
...
+    '';
+  };
+
   extraOptions = {
     colorscheme =
       let
+        names = pipe "0123456789ABCDEF" [
+          stringToCharacters
+          (map (i: "base0" + i))
+        ];
```
MattSturgeon, line 90: As nice as this is, it's made totally redundant by declaring `examples` in the let block :sob: 

We can remove `names` and then `mapAttrs` the example instead

plugins/colorschemes/base16/default.nix
```diff
@@ -24,7 +29,7 @@ helpers.vim-plugin.mkVimPlugin config {
     let
```
MattSturgeon, line 29: I figured we can keep these renames for now. Should I bump the date in the TODO, e.g. for removal after 24.11?

MattSturgeon:
> vimPackages/start/base16-nvim/lua/base16-colorscheme.lua:705: module 'colors.base16-gruvbox-dark-hard' not found:

Turns out the `base16-` prefix only applies when using the vimscript api (`:colorscheme`), the [lua implementation has no prefix](https://github.com/RRethy/base16-nvim/blob/master/lua/colors/init.lua).

This is good, because we don't need any new warnings or changed behaviour!

plugins/colorschemes/base16/theme-list.nix
MattSturgeon, line 1: This list is probably still too long for the docs. Feel free to suggest more removals.

plugins/colorschemes/base16/default.nix
```diff
@@ -37,94 +39,134 @@ helpers.vim-plugin.mkVimPlugin config {
...
+
+  extraConfig = cfg: {
+    plugins.airline.settings.theme = mkIf cfg.setUpBar (mkDefault name);
+    plugins.lualine.theme = mkIf cfg.setUpBar (mkDefault name);
+    plugins.lightline.colorscheme = mkDefault null;
+
+    opts.termguicolors = mkDefault true;
+    # FIXME: is it still ok to use `colorscheme` after using `setup`?
+    # It would be redundant, but maybe users want to read `config.colorscheme` in their nix config?
+    colorscheme = mkForce null;
```
MattSturgeon, line 163: I don't think this needs a FIXME comment. TBH I'm not sure we need to set `colorscheme to `null` anyway, since that's the default.

If a user really wants to enable `colorschemes.base16` and still wants to call `:colorscheme` anyway, they've probably got a reason?

plugins/colorschemes/base16/default.nix
```diff
@@ -37,94 +39,134 @@ helpers.vim-plugin.mkVimPlugin config {
...
+
+    # `settings` can either be passed to `with_config` before calling `setup`,
+    # or it can be passed as `setup`'s 2nd argument.
+    # See https://github.com/RRethy/base16-nvim/blob/6ac181b5733518040a33017dde654059cd771b7c/lua/base16-colorscheme.lua#L107-L125
+    extraConfigLuaPre = ''
+      require('base16-colorscheme').setup(${helpers.toLuaObject cfg.colorscheme}, ${helpers.toLuaObject cfg.settings})
```
MattSturgeon, line 169: This is also valid:
```suggestion
      require('base16-colorscheme').with_config(${helpers.toLuaObject cfg.settings})
      require('base16-colorscheme').setup(${helpers.toLuaObject cfg.colorscheme})
```
Which do you think is better?

GaetanLepage: I don't think it makes a huge difference, but having the `settings` option not feeding `require(..).setup` is quite weird. Pay attention to the description of the `settings` option itself.

MattSturgeon: > having the `settings` option not feeding `require(..).setup` is quite weird.

Yep. Their `setup` function _can_ take the config as a second argument, but they also provide `with_config` which is used separately. I.e. _even weirder_ than a plugin that has some alternative function to `setup`!

> Pay attention to the description of the `settings` option itself.

```
Options provided to the `require('${luaName}').setup` function.
```

You're right, that's an issue. I think it likely already affects other plugins though? Unless we have some way of overriding the description... Maybe your upcoming mk*Plugin refactor should take this into account?

plugins/colorschemes/base16/default.nix
```diff
@@ -37,94 +39,131 @@ helpers.vim-plugin.mkVimPlugin config {
...
+    telescope_borders = true;
+    indentblankline = false;
+    dapui = false;
+  };
+
+  settingsOptions = with helpers.defaultNullOpts; {
+    telescope = mkBool true ''
+      Whether to enable telescope integration.
+    '';
+    telescope_borders = mkBool false ''
```
GaetanLepage, line 52:
```suggestion
  settingsOptions = {
    telescope = helpers.defaultNullOpts.mkBool true ''
      Whether to enable telescope integration.
    '';

    telescope_borders = helpers.defaultNullOpts.mkBool false ''
```

Two reasons:
- Consistence with the rest of the code base (very valuable IMO for new contributors)
- Following nixpkgs' recommendation of discouraging the use of `with`.

MattSturgeon: Agreed. Not sure why I was using `with` there in the first place!

TBH, even the way we usually put `with lib;` at the start of a file goes against best practices and nixd will be shouting `sema-escaping-with` all over the place unless you suppress that warning. But it does make the code less verbose and is consistent with the rest of the codebase...

GaetanLepage: > TBH, even the way we usually put `with lib;` at the start of a file goes against best practices and nixd will be shouting `sema-escaping-with` all over the place unless you suppress that warning. But it does make the code less verbose and is consistent with the rest of the codebase...

Sure, We'd have to assess the relevance of an alternative approach. But this is out of scope for this PR.

plugins/colorschemes/base16/default.nix
GaetanLepage: yes, let's do that.

plugins/colorschemes/base16/default.nix
GaetanLepage: Ok ^^
Marking this as solved then.

plugins/colorschemes/base16/default.nix
GaetanLepage: > You're right, that's an issue. I think it likely already affects other plugins though? Unless we have some way of overriding the description... Maybe your upcoming mk*Plugin refactor should take this into account?

Yes, this might happen to be inaccurate for certain plugins.
We could add a `settingsDescription` argument that defaults to "... provided to the require blablabla".

MattSturgeon: This was resolved in #1758

GaetanLepage (approved):
Ok, this is looking good !
Sorry for taking so long...

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1655, opened 2024-06-08
plugins/lsp: add inlay-hint option

MattSturgeon:
Following up on #1313, this PR adds a the `plugins.lsp.inlayHints` option, which enables the capability in `plugins.lsp.onAttach`.

Unlike the original PR, no attempt is made to detect the neovim version used by `finalPackage`.

I still haven't considered if/how this should impact any existing server-specific options:
- [x] Basic inlay hint support (`plugins.lsp.inlayHints`)
- [ ] Consider how/if this can be configured for a specific LSP
- [ ] Consider whether this impacts existing options (should they be deprecated?)
  - [ ] [`plugins.clangd-extensions.inlayHints`](https://nix-community.github.io/nixvim/plugins/clangd-extensions/inlayHints.html)
  - [ ] [`plugins.rust-tools.inlayHints`](https://nix-community.github.io/nixvim/plugins/rust-tools/inlayHints.html)
  - [ ] [`plugins.lsp.servers.rust-analyzer.settings.inlayHints`](https://nix-community.github.io/nixvim/plugins/lsp/servers/rust-analyzer/settings/inlayHints/closingBraceHints.html)
  - [ ] Alternatively, maybe those LSPs having an option is preferable to having a global option?
  - [ ] (This can be moved to its own issue)
- [ ] Consider how/if we should expose more advanced functionality
  - [ ] Enable only in specific modes
  - [ ] Enable for the current line
  - [ ] Etc
  - [ ] Maybe use an `enum ([ true false "line" ] ++ modes)`?
  - [ ] (This can be moved to its own issue)

Fixes #1306

PR #1680, opened 2024-06-12
plugins/lsp fix: incorrect function call

darkdarcool:
The new lsp inlay hint option in Nixvim (#1655) wasn't enabling inlay hints, and was producing the following error message (apologies for the poor screenshot, the message dissapears quickly with Noice)

image

The error is caused by an incorrect function call to [`vim.lsp.inlay_hint.enable`](https://github.com/neovim/neovim/blob/72155121006bca884e154e935640054f2e090367/runtime/lua/vim/lsp/inlay_hint.lua#L408). I wasn't sure what to put for the filter parameter (maybe that should be it's own option?), so I just left it as an empty table for the time being.

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -212,7 +212,7 @@ in
       plugins.lsp.onAttach = mkIf cfg.inlayHints ''
         -- LSP Inlay Hints {{{
         if client.server_capabilities.inlayHintProvider and vim.lsp.inlay_hint then
-          vim.lsp.inlay_hint.enable(bufnr, true)
+          vim.lsp.inlay_hint.enable(true, { bufnr = bufnr })
         end
         -- }}}
       '';


plugins/lsp/default.nix
```diff
@@ -212,7 +212,7 @@ in
       plugins.lsp.onAttach = mkIf cfg.inlayHints ''
         -- LSP Inlay Hints {{{
         if client.server_capabilities.inlayHintProvider and vim.lsp.inlay_hint then
-          vim.lsp.inlay_hint.enable(bufnr, true)
+          vim.lsp.inlay_hint.enable(true, {})
```
MattSturgeon, line 215:
```suggestion
          vim.lsp.inlay_hint.enable(true)
```

[Docs](https://neovim.io/doc/user/lsp.html#lsp-inlay_hint) say we can pass `nil` if we want filter to allow all.

MattSturgeon (commented):
My bad, I should've tested it still worked before merging. The API must've changed between when I initially worked on the feature and 0.10 being released...

plugins/lsp/default.nix
MattSturgeon: Actually, they reference `bufnr` indented under `filters`... Maybe `bufnr` itself is now a filter (additional to [kwargs](https://neovim.io/doc/user/lua.html#kwargs))?

MattSturgeon: Testing now, `:lua vim.lsp.inlay_hint.enable(true, {}, 1)` seems to work:

```suggestion
          vim.lsp.inlay_hint.enable(true, {}, bufnr)
```

I guess the extra indentation is just a bug in the docs.

MattSturgeon (approved):
Thanks for this!

PR #1686, opened 2024-06-13
plugins/toggleterm: allow lua functions in `float_opts` keys which support them

michaelhthomas:
Updates the option schema for the `toggleterm` plugin to allow lua functions as values for the `float_opts` options `width`, `height`, `row`, and `col`. These properties are described as being either a number or lua function [in the plugin documentation](https://github.com/akinsho/toggleterm.nvim?tab=readme-ov-file#setup).


MattSturgeon (approved):
Great, thanks!

GaetanLepage (approved):

PR #1691, opened 2024-06-14
lib/options: migrate to `pluginDefault` arg-name

MattSturgeon:
I was expecting this PR to be much bigger. Luckily we only use the `defaultNullOpts` prime variants in a few locations and many of those uses don't actually set a plugin default.

The non-prime variants are only affected internally, because renaming arguments does not change the function signature.

See original discussion here: https://github.com/nix-community/nixvim/pull/1667#discussion_r1632607368

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -9,7 +9,7 @@ let
   # Render a plugin default string
   pluginDefaultText =
     let
-      # Assume a string `default` is already formatted as intended,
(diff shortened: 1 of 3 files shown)


MattSturgeon:
`git-octo` failing CI _again_

GaetanLepage (approved):
Everything looks right !

PR #1703, opened 2024-06-15
plugins/lsp/lsp-lines: add option to disable virtual text

PowerUser64:
lsp-lines was disabling virtual_text by default, which is something the base plugin doesn't do, so we shouldn't do it here. This adds an option to disable it (`disableVirtualText`) and initializes it to `false`.

### DIFF
--- a/plugins/lsp/lsp-lines.nix
+++ b/plugins/lsp/lsp-lines.nix
@@ -19,17 +19,26 @@ in
       currentLine = mkOption {
         type = types.bool;
         default = false;
-        description = "Show diagnostics only on current line";
+        description = "Show diagnostics only on current line.";
+      };
+
+      disableVirtualText = mkOption {
+        type = types.bool;
+        default = false;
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Thanks for highlighting (and fixing) this issue. TBH I would rather resolve the issue by refactoring the plugin to use an rfc42 style settings option (i.e. switch to using our mkNeovimPlugin helper).

Is that something you'd like to tackle? If not one of us should be able to find time.

If you are willing to refactor the plugin: you'll find our CONTRIBUTING.md and plugins/TEMPLATE.nix files helpful, along with looking at any recently merged "switch to mkNeovimPlugin" PRs.

MattSturgeon:
Would need to check if our usual `settings` still model makes sense here, seeing as the plugin's config is actually passed to a vim built-in: `vim.diagnostic.config(diagnosticConfig)`

Thoughts @nix-community/nixvim ?

PR #1704, opened 2024-06-15
plugins/utils/nvim-osc52: add desc option

PowerUser64:
The keymaps had no description, so I added this option to set it. Also, I added periods to the ends of the option descriptions.

### DIFF
--- a/plugins/utils/nvim-osc52.nix
+++ b/plugins/utils/nvim-osc52.nix
@@ -12,34 +12,40 @@ with lib;
 
     package = helpers.mkPluginPackageOption "nvim-osc52" pkgs.vimPlugins.nvim-osc52;
 
-    maxLength = helpers.defaultNullOpts.mkInt 0 "Maximum length of selection (0 for no limit)";
-    silent = helpers.defaultNullOpts.mkBool false "Disable message on successful copy";
-    trim = helpers.defaultNullOpts.mkBool false "Trim text before copy";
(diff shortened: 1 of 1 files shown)


plugins/utils/nvim-osc52.nix
```diff
@@ -8,38 +8,44 @@
 with lib;
 {
   options.plugins.nvim-osc52 = {
-    enable = mkEnableOption "nvim-osc52, a plugin to use OSC52 sequences to copy/paste";
+    enable = mkEnableOption "nvim-osc52, a plugin to use OSC52 sequences to copy/paste.";
```
MattSturgeon, line 11: mkEnableOption adds its own period. This will render as "Whether to enable nvim-osc52, a plugin to use OSC52 sequences to copy/paste.."

MattSturgeon (changes requested):
Thanks! I'll look properly later, but here's some initial feedback

plugins/utils/nvim-osc52.nix
PowerUser64: Thanks for the pointer! Fix'd.

MattSturgeon (commented):
I think this PR highlights another plugin that needs converting to mkNeovimPlugin and some keymap options that ought to be standardized.

It also looks like we're currently setting keymaps for this plugin by default, which goes against our general principle of minimal magic and explicit opt-in configuration.

That said, this PRs changes have a low maintenance cost. Unless someone is happy to work on refactoring the plugin, I think we can go ahead with this PR?

I left some general thoughts and comments below. Apologies for rambling.

PR #1709, opened 2024-06-16
lib: make lua types + `mkRaw` helper more accommodating

MattSturgeon:
Motivated by needing to migrate some existing options from maybeRaw types to strLua types, this PR refactors the types and mkRaw to accommodate that use case, plus others.

To be specific:
- strLua types now also accept rawLua
- mkRaw will "pass through" rawLua untouched
- mkRaw now `throw`s when passed a non-null, non-string, non-raw value
  - previously _any_ invalid value resulted in `null`

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -14,7 +14,7 @@ let
     };
 in
 rec {
-  isRawType = v: lib.isAttrs v && lib.hasAttr "__raw" v && lib.isString v.__raw;
+  isRawType = v: v ? __raw && isString v.__raw;
 
   rawLua = mkOptionType {
     name = "rawLua";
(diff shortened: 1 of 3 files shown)


lib/types.nix
```diff
@@ -9,12 +9,13 @@ let
...
       descriptionClass = "noun";
-      check = lib.isString;
+      check = v: isString v || isRawType v;
       merge = lib.options.mergeEqualOption;
     };
+  isRawType = v: v ? __raw && isString v.__raw;
 in
 rec {
-  isRawType = v: lib.isAttrs v && lib.hasAttr "__raw" v && lib.isString v.__raw;
+  inherit isRawType;
```
MattSturgeon, line 18: This wouldn't be needed if we built the `helpers` lib recursively, passing in `helpers` itself to `helpers`, like what nixpkgs/lib does.

GaetanLepage (approved):
LGTM

PR #1674, opened 2024-06-11
treewide: remove pre-formatted string defaults

MattSturgeon:
Superseding #1665, this is a subset of the changes there. I've kept this in a separate branch so that there is a record of the other changes that may be nice to have.

I've filtered the diff for changes that directly relate to stringified plugin-defaults and the diff has more than halved in the process. It's still a mammoth diff though; 127 files and around 2k additions _and_ removals...

That said I wasn't able to completely separate the core change from _all_ of the tangential changes, though most of them should be fairly obvious.

The changes are spread over 15 commits, though this was just to allow me to use git bisect to find some issues. I thing they should be squashed into one or two commits before merging.

Merging https://github.com/nix-community/nixvim/pull/1670 first may also (slightly) simplify the diff, at least within `lib/options`.

#### Not covered in this PR:
- Remove `mkDesc` (that's in https://github.com/nix-community/nixvim/pull/1670)
- Remove manually inserted "Plugin default"s (big job)
- Move manually inserted "Examble"s (also pretty big)
- Switch most `mkNullable` uses to `mkNullableWIthRaw`
- Validating _any_ of the touched options are still up to date
- 100% ensuring all stringified defaults are gone.
  It's a very manual process to verify whether each default should or should not be a string.
- Replacing `\'\'` with `'''` where appropriate
- Anything I flagged with a TODO or FIXME

#### Blocked by:
- [x] #1670
- [x] #1710
- [x] #1711
- [x] #1716
- [ ] #1714

PR #1710, opened 2024-06-16
plugins/cmp-git: general cleanup of options

MattSturgeon:
Prompted by (and extracted from) #1674, this PR is a general cleanup of some options in the cmp-git plugin.

- Move several options to `strLuaFn`
- Replace all pre-formatted plugin defaults
- Fix some incorrect examples/tests

This PR **depends on** #1709, otherwise existing `__raw` values in users' configs will produce an incorrect type error:
```
error: A definition for option `plugins.cmp-git.settings.github.issues.format' is not of type `null or lua function string'. Definition values:
- In `<unknown-file>':
{
__raw = ''
function(_, issue)
local icon = ({
open = '',
...
```

### DIFF
--- a/plugins/completion/cmp/sources/cmp-git.nix
+++ b/plugins/completion/cmp/sources/cmp-git.nix
@@ -7,18 +7,26 @@
 with lib;
 let
   cfg = config.plugins.cmp-git;
+
+  mkAction =
+    action: target:
+    helpers.defaultNullOpts.mkLuaFn "require('cmp_git.${action}').git.${target}" ''
+      Function used to ${action} the ${replaceStrings [ "_" ] [ " " ] target}.
+    '';
 in
 {
   options.plugins.cmp-git.settings = helpers.neovim-plugin.mkSettingsOption {
     pluginName = "cmp_git";
     options = {
-      filetypes = helpers.defaultNullOpts.mkListOf types.str ''["gitcommit" "octo"]'' ''
-        Filetypes for which to trigger.
-      '';
+      filetypes = helpers.defaultNullOpts.mkListOf types.str [
+        "gitcommit"
+        "octo"
+      ] "Filetypes for which to trigger.";
 
-      remotes = helpers.defaultNullOpts.mkListOf types.str ''["upstream" "origin"]'' ''
-        List of git remotes.
-      '';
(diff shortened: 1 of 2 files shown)


plugins/completion/cmp/sources/cmp-git.nix
```diff
@@ -7,18 +7,36 @@
...
-      '';
+      filetypes =
+        helpers.defaultNullOpts.mkListOf types.str
+          [
+            "gitcommit"
+            "octo"
+          ]
+          ''
+            Filetypes for which to trigger.
+          '';
```
MattSturgeon, line 29:
```suggestion
          "Filetypes for which to trigger.";
```

This might cause nixfmt to attach the list brackets to the adjacent lines?

```nix
helpers.defaultNullOpts.mkListOf types.str [
            "gitcommit"
            "octo"
          ] "Filetypes for which to trigger.";
```

Even if not, I think its more readable for short descriptions like this?

MattSturgeon:
The `filter_fn` appears to be undocumented upstream, so I haven't added an option for it. This means it needs to use `__raw` in the test & example though.

plugins/completion/cmp/sources/cmp-git.nix
GaetanLepage: Yes, I would prefer a "normal" string as well.

plugins/completion/cmp/sources/cmp-git.nix
```diff
@@ -7,18 +7,36 @@
...
-      '';
+      remotes =
+        helpers.defaultNullOpts.mkListOf types.str
+          [
+            "upstream"
+            "origin"
+          ]
+          ''
+            List of git remotes.
+          '';
```
GaetanLepage, line 39:
```suggestion
          "List of git remotes.";
```

GaetanLepage (changes requested):

plugins/completion/cmp/sources/cmp-git.nix
MattSturgeon: Done

plugins/completion/cmp/sources/cmp-git.nix
MattSturgeon: Done

PR #1714, opened 2024-06-17
plugins/flash: switch to mkNeovimPlugin

MattSturgeon:
Prompted by changes in #1674, I've gone ahead and updated https://github.com/folke/flash.nvim to use mkNeovimPlugin and rfc42 settings.

The "modes" use the same options as top-level settings, so I've maintained a similar system to the existing implementation where `mkModeOption` combines all the common settings with any extras.

Rather than documenting mode-specific defaults in the option itself, I pass the defaults through to the sub-options themselves. This makes the docs slightly more accurate.

The `search.automatic` option is removed. If my interpretation of upstream's docs is correct it was never needed?

### DIFF
--- a/plugins/utils/flash.nix
+++ b/plugins/utils/flash.nix
@@ -6,589 +6,664 @@
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.flash;
-in
-{
-  options.plugins.flash =
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "flash";
+  originalName = "flash.nvim";
+  defaultPackage = pkgs.vimPlugins.flash-nvim;
+
+  maintainers = with maintainers; [
+    traxys
+    MattSturgeon
+  ];
+
+  # TODO: Added 2024-06-17; remove 2024-09-17
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings =
     let
-      configOpts = {
-        labels = helpers.defaultNullOpts.mkStr "asdfghjklqwertyuiopzxcvbnm" ''
-          Labels appear next to the matches, allowing you to quickly jump to any location. Labels are
-          guaranteed not to exist as a continuation of the search pattern.
-        '';
+      # Combine base option paths with nested mode option paths
+      merged = flatten [
+        opts
+        (mapAttrsToList (mode: extras: map (opt: "${mode}.${opt}") (opts ++ extras)) modes)
+      ];
 
+      # Option paths that should be renamed
+      opts = [
+        "labels"
+        "search.multiWindow"
+        "search.forward"
+        "search.wrap"
+        "search.mode"
+        "search.incremental"
+        "search.exclude"
+        "search.trigger"
+        "search.maxLength"
+        "jump.jumplist"
+        "jump.pos"
+        "jump.history"
+        "jump.register"
+        "jump.nohlsearch"
+        "jump.autojump"
+        "jump.inclusive"
+        "jump.offset"
+        "label.uppercase"
+        "label.exclude"
+        "label.current"
+        "label.after"
+        "label.before"
+        "label.style"
+        "label.reuse"
+        "label.distance"
+        "label.minPatternLength"
+        "label.rainbow.enabled"
+        "label.rainbow.shade"
+        "label.format"
+        "highlight.backdrop"
+        "highlight.matches"
+        "highlight.priority"
+        "highlight.group.match"
+        "highlight.group.current"
+        "highlight.group.backdrop"
+        "highlight.group.label"
+        "action"
+        "pattern"
+        "continue"
+        "config"
+        "prompt.enabled"
+        "prompt.prefix"
+        "prompt.winConfig.relative"
+        "prompt.winConfig.width"
+        "prompt.winConfig.height"
+        "prompt.winConfig.row"
+        "prompt.winConfig.col"
+        "prompt.winConfig.zindex"
+        "remoteOp.restore"
+        "remoteOp.motion"
+      ];
+
+      # "Modes" that should also be renamed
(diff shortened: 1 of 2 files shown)


plugins/utils/flash.nix
```diff
@@ -6,536 +6,669 @@
...
+  name = "flash";
+  originalName = "flash.nvim";
+  defaultPackage = pkgs.vimPlugins.flash-nvim;
+
+  maintainers = with maintainers; [
+    traxys
+    MattSturgeon
+  ];
+
+  # TODO: Added 2024-06-17; remove 2024-09-17
```
GaetanLepage, line 19: You need to deprecate the `extraOptions` option.

plugins/utils/flash.nix
```diff
@@ -6,536 +6,669 @@
...
-        labels = helpers.defaultNullOpts.mkStr "asdfghjklqwertyuiopzxcvbnm" ''
-          Labels appear next to the matches, allowing you to quickly jump to any location. Labels are
-          guaranteed not to exist as a continuation of the search pattern.
-        '';
-
+      # Combine base option paths with nested mode option paths
+      merged = flatten [
+        opts
+        (mapAttrsToList (mode: extras: map (opt: "${mode}.${opt}") (opts ++ extras)) modes)
+      ];
```
GaetanLepage, line 26: Why do you have those variables "out of order" ? I would personally prefer an "imperative-friendly" way.

MattSturgeon: Could you clarify what you mean? Do you mean having `merged` declared before `opts`?

I did that to make it obvious up front what the goal of the let block is, rather than having rather large attr sets/lists being the first thing you see.

Happy to change and/or remedy with comments.

GaetanLepage: Yes, I meant
```
a = ...
b = f(a)
c = f(b)
d = f(a, c)
...
```

I find it more readable. Otherwise you are wondering what these `opts` are. But this is personal preference of course.

MattSturgeon: > I find it more readable. Otherwise you are wondering what these `opts` are. But this is personal preference of course.

Generally, I agree. It was only the size of the other variable definitions that changed my mind this time.

I think because the `opts` is _right there_ on the next line it is ok (this time)?

GaetanLepage: Yes sure, either way is fine. We should stop nit picking (I am looking at myself).

MattSturgeon: > Yes sure, either way is fine. We should stop nit picking (I am looking at myself).

Nothing wrong with being thorough and discussing the best approach in nuanced cases.

Maybe avoid nit-picking new contributes if it is likely to put them off contributing, but that doesn't mean let sloppy code in :smiley:

plugins/utils/flash.nix
```diff
@@ -6,536 +6,670 @@
...
-            before
-            style
-            reuse
-            distance
-            ;
-          min_pattern_length = c.label.minPatternLength;
-          rainbow = {
-            inherit (c.label.rainbow) enabled shade;
+
+            char_actions = helpers.defaultNullOpts.mkLuaFn defaults.char_actions ''
```
MattSturgeon, line 620: Think this type is wrong? Upstream docs say:

> alias Flash.CharActions table<string, "next" | "prev" | "right" | "left">
> The direction for `prev` and `next` is determined by the motion.
> `left` and `right` are always left and right.

MattSturgeon: I guess it is right, the [only usage](https://github.com/folke/flash.nvim/blob/43f67935d388fbb540f8b40e8cbfd80de54f978a/lua/flash/plugins/char.lua#L200) calls it as a function.

plugins/utils/flash.nix
```diff
@@ -6,536 +6,670 @@
...
+    map (splitString ".") merged;
+
+  imports = [
+    # TODO: check automatic search config still works
+    (mkRemovedOptionModule [
+      "plugins"
+      "flash"
+      "search"
+      "automatic"
+    ] "You can configure `plugins.flash.settings.search.*` directly.")
```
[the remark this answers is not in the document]
traxys: I'm not very familiar with the plugin, I don't remembrer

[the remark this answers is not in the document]
MattSturgeon: I'm also not too familiar, so just trying to interpret the docs.

I think we can go with this and just hope if there's any issues they get reported.

MattSturgeon:
> Is this blocked by anything ?

No, don't think so

plugins/utils/flash.nix
```diff
@@ -6,589 +6,664 @@
...
-    // {
-      enable = mkEnableOption "flash.nvim";
+          prompt = {
+            enabled = helpers.defaultNullOpts.mkBool defaults.prompt.enabled ''
+              Options for the floating window that shows the prompt, for regular jumps.
+            '';
 
-      package = helpers.mkPluginPackageOption "flash.nvim" pkgs.vimPlugins.flash-nvim;
+            # Not sure what the type is...
+            # Think it's listOf (maybeRaw (listOf (maybeRaw str)))?
```
GaetanLepage, line 458: This needs to be solved.

MattSturgeon: The comment is not mine, it is already present in `main`.

Same as https://github.com/nix-community/nixvim/pull/1714#discussion_r1647191887, I think I'd have to play with actually using/configuring the plugin to understand to correct solution.

The current type will at least allow users who know how to configure the plugin to do so, though they won't get build-time type checking.

GaetanLepage: Ok, I didn't pay attention.

GaetanLepage (approved):
LGTM

PR #1717, opened 2024-06-17
lib/helpers: refactor mkPlugin helpers

GaetanLepage:


lib/plugin/neovim-plugin.nix
```diff
@@ -0,0 +1,102 @@
...
+    {
+      name,
+      defaultPackage, # make this parameter mandatory for neovim plugins
+      # deprecations
+      deprecateExtraOptions ? false,
+      # colorscheme
+      isColorscheme ? false,
+      # options
+      settingsOptions ? { },
+      settingsExample ? null,
```
MattSturgeon, line 46: This could generalize both the settings option description and the callSetup implementation.

```suggestion
      settingsExample ? null,
      settingsDesc ? "Options provided to the `require('${luaName}').${setupFn}` function.",
      setupFn ? "setup",
```

I think it should be possible to override the description too, not just setup name, because some lua plugins may not use a function at all.

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+    let
+      cfg = config.${namespace}.${name};
+    in
+    mkIf cfg.enable (mkMerge [
+      {
+        extraPlugins = extraPlugins ++ optional (defaultPackage != null) cfg.package;
+        inherit extraPackages;
+      }
+      (extraConfig cfg)
+      (optionalAttrs (isColorscheme && (colorscheme != null)) { colorscheme = mkDefault colorscheme; })
```
MattSturgeon, line 88: Should this handle opt.termguicolors too?

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+      basePluginPath = [
+        namespace
+        name
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    map (
+      option:
+      let
+        optionPath = if isString option then [ option ] else option; # option is already a path (i.e. a list)
```
MattSturgeon, line 42:
```suggestion
        optionPath = toList option;
```

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+  description ? null,
+  # deprecations
+  optionsRenamedToSettings ? [ ],
+  # colorscheme
+  isColorscheme ? false,
+  colorscheme ? name,
+  # options
+  originalName ? name,
+  defaultPackage ? null,
+  extraOptions ? { },
```
MattSturgeon, line 22: I wonder if there's a way we can pass in a (curried, wrapped, or custom) settings-option-factory function? The only difference between nvim and vim's settings is the description, right?

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+        optionPath = if isString option then [ option ] else option; # option is already a path (i.e. a list)
+
+        optionPathSnakeCase = map nixvimUtils.toSnakeCase optionPath;
+      in
+      mkRenamedOptionModule (
+        [
+          namespace
+          name
+        ]
+        ++ optionPath
```
MattSturgeon, line 51:
```suggestion
        basePluginPath ++ optionPath
```

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+in
+{
+  meta = {
+    inherit maintainers;
+    nixvimInfo = {
+      inherit description url;
+      path = [
+        namespace
+        name
+      ];
```
MattSturgeon, line 63:
```suggestion
      inherit description url path;
```

Let's define this in a wider scope or _as a default-arg_ so we can use it here **and** in the deprecations.

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+    };
+  };
+
+  imports = optionsRenamedToSettingsWarnings ++ imports;
+
+  options.${namespace}.${name} =
```
MattSturgeon, line 69:
```suggestion
  options = setAttrByPath path (
```

lib/plugin/mk-plugin.nix
```diff
@@ -0,0 +1,90 @@
...
+      enable = mkEnableOption originalName;
+    }
+    // (optionalAttrs (defaultPackage != null) {
+      package = nixvimOptions.mkPluginPackageOption originalName defaultPackage;
+    })
+    // extraOptions;
+
+  config =
+    let
+      cfg = config.${namespace}.${name};
```
MattSturgeon, line 80:
```suggestion
      cfg = getAttrFromPath path config;
```

lib/plugin/neovim-plugin.nix
```diff
@@ -0,0 +1,102 @@
...
+          settings = mkSettingsOption {
+            pluginName = name;
+            options = settingsOptions;
+            example = settingsExample;
+          };
+        } // (args.extraOptions or { });
+
+        extraConfig =
+          let
+            extraConfigNamespace = if isColorscheme then "extraConfigLuaPre" else "extraConfigLua";
```
MattSturgeon, line 87: Make this a default arg instead?

`extraConfigNamespace ? if isColorscheme then "extraConfigLuaPre" else "extraConfigLua",`

lib/plugin/vim-plugin.nix
```diff
@@ -0,0 +1,87 @@
...
+            basePluginPath = [
+              namespace
+              name
+            ];
+            settingsPath = basePluginPath ++ [ "settings" ];
+          in
+          (args.imports or [ ])
+          ++ (optional (deprecateExtraConfig && createSettingsOption) (
+            mkRenamedOptionModule (basePluginPath ++ [ "extraConfig" ]) settingsPath
+          ));
```
MattSturgeon, line 75: Isn't this the same as nvim's deprecate extraConfig? Could you just pass `deprecateExtraConfig = deprecateExtraConfig && createSettingsOption` through to mkPlugin?

EDIT: The option name is different too (extraOptions vs extraConfig). I guess not worth it then.

MattSturgeon (commented):
Apologies for the number of comments, you've probably already considered most of them.

Let's discuss tomorrow!

lib/plugin/vim-plugin.nix
GaetanLepage: > EDIT: The option name is different too (extraOptions vs extraConfig). I guess not worth it then.

indeed...

lib/plugin/mk-plugin.nix
GaetanLepage: You find this clearer ?

MattSturgeon: I think it's more flexible, if we ever support having a nested namespace/path

lib/plugin/mk-plugin.nix
MattSturgeon: If the description is the only real difference, maybe have `mkPlugin` accept a `settingsDesc` argument?

lib/plugin/neovim-plugin.nix
MattSturgeon: Ok

lib/plugin/neovim-plugin.nix
```diff
@@ -0,0 +1,102 @@
...
+        if pluginName != null then
+          "Options provided to the `require('${pluginName}').setup` function."
+        else
+          throw "mkSettingsOption: Please provide either a `pluginName` or `description`.",
+      example ? null,
+    }:
+    nixvimOptions.mkSettingsOption { inherit options description example; };
+
+  # TODO: DEPRECATED: use the `settings` option instead
+  extraOptionsOptions = {
```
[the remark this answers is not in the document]
GaetanLepage: Yup. Once the rest of the internal refactors will be over (soon as hope), we can focus on modernizing the rest of the codebase.

lib/plugin/vim-plugin.nix
```diff
@@ -0,0 +1,87 @@
...
+
+        extraOptions = settingsOption // (args.extraOptions or { });
```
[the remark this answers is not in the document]
GaetanLepage: Indeed !

PR #1721, opened 2024-06-18
tests/plugins/sniprun: disable old irrelevant test for sniprun

GaetanLepage:
The relevant one is in [`tests/test-sources/plugins/languages/sniprun.nix`](https://github.com/nix-community/nixvim/blob/bf109a3e59845fd7806dfdd386edfa3ac1baa998/tests/test-sources/plugins/languages/sniprun.nix).


MattSturgeon (approved):
Good spot!

PR #1722, opened 2024-06-18
plugins/cmp-tabby: switch to RFC-42 style

GaetanLepage:


plugins/completion/cmp/sources/cmp-tabby.nix
```diff
@@ -11,37 +11,59 @@ in
...
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    [
+      (mkRenamedOptionModule (basePluginPath ++ [ "extraOptions" ]) settingsPath)
+      (mkRenamedOptionModule (basePluginPath ++ [ "host" ]) (settingsPath ++ [ "host" ]))
+      (mkRenamedOptionModule (basePluginPath ++ [ "maxLines" ]) (settingsPath ++ [ "max_lines" ]))
+      (mkRenamedOptionModule (basePluginPath ++ [ "runOnEveryKeyStroke" ]) (
+        settingsPath ++ [ "run_on_every_keystroke" ]
+      ))
+      (mkRenamedOptionModule (basePluginPath ++ [ "stop" ]) (settingsPath ++ [ "stop" ]))
```
MattSturgeon, line 30: This could be mapped through `helpers.toSnakeCase` like mkNeovimPlugin does. This is fine though as it is small, simple and readable.

Off topic: Isn't this a good use-case for mk*Plugin accepting a path instead of a namespace?

MattSturgeon (approved):
Looks good! One less file to deal with in #1674 :grin:

PR #1724, opened 2024-06-18
plugins/none-ls: use rfc42 style

MattSturgeon:
- **plugins/none-ls: switch to mkNeovimPlugin**
- **plugins/none-ls: switch sources to rfc42 style**
- **plugins/none-ls: refactor withArgs deprecation**


MattSturgeon:
```
 error: function 'anonymous lambda' called without required argument 'options'
at /nix/store/s2b5bnzqsqs5qiyin7l72ck9wwq3j01d-source/plugins/none-ls/servers.nix:1:1:
1| {
| ^
2| pkgs,
```

I'm confused, the 'anonymous lambda' looks to be the servers.nix module. Why would a module be called without the `options` argument?

MattSturgeon:
> I'm confused, the 'anonymous lambda' looks to be the servers.nix module. Why would a module be called without the `options` argument?

In case anyone stumbles on this: the module system doesn't pass all args to module functions; instead it checks which args the module expects and only provides those ones. If you use `@args`, `args` will only contain your explicitly declared args (plus some _special_ args).

In this case, I was probably doing something like capturing the module `args` _without_ declaring that I need the `options` arg, and forwarding `args` to `server.nix`.

PR #1726, opened 2024-06-19
modules: use real nix expressions for option examples

GaetanLepage:


traxys (approved):

traxys:
I  think we need to disable the octo tests, they fail too often

MattSturgeon (approved):

Issue #1705, opened 2024-06-15
none-ls: updating + refactioring


MattSturgeon:
Overview issue to track specific issues, PRs, and ideas related to updating and refactoring none-ls.

- [ ] Switch to `mkNeovimPlugin` (move options to `settings`)
- [ ] Move `sources.*.withArgs` to `sources.*.settings`
  - [ ] Use `either strLua submodule` to allow a rename-alias
  - [ ] Include ["filetypes"](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md#filetypes) from `builtins.json` in option types/defaults/descriptions?
- [ ] Consider how `enableLspFormat` should work with rfc42 `settings`?
  - [ ] When enabled, it sets `on_attach` _default to_ `require('lsp-format').on_attach`.
  - [ ] Maybe we should just suggest users include this in their `on_attach` manually?
  - [ ] Current description is misleading: #1690
- [ ] Support "custom" sources submodules
  - [ ] Custom sources need to be [registered](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/SOURCES.md#registerto_register) first
  - [ ] Submodule [shape](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/MAIN.md#sources): `{ method, generator, filetypes[], name?, condition? }`.
- [ ] General refactoring
  - [ ] Move "builtin sources packages" list into a dedicated file
  - [ ] Move `builtinPackages` out of sync with `builtins.json` errors out of module and into a CI check
- [ ] IFD
  - [ ] Check if the IFD is contributing to #1154 
  - [ ] Weigh pros/cons of IFD vs a separate "update" script

General notes and discussion can be had on this issue, but I'd encourage opening [discussions](https://github.com/nix-community/nixvim/discussions) or [issues](https://github.com/nix-community/nixvim/issues) for specific issues. These can be linked in the relevant checklist above.

Tagging @traxys because he's done most of the work on the none-ls module so far.

traxys:
>  Consider how enableLspFormat should work with rfc42 settings? 
I'd say we leave it as a top level option, top level options are options that are defined by nixvim, not the wrapped plugin

MattSturgeon:
> > Consider how enableLspFormat should work with rfc42 settings?
> 
> I'd say we leave it as a top level option, top level options are options that are defined by nixvim, not the wrapped plugin

[...]

PR #1727, opened 2024-06-20
plugins/none-ls: switch to mkNeovimPlugin

### LINKED ISSUE #1690: [BUG] `lsp-format` should get enabled if enabling `enableLspFormat`
| Plugin       | `none-ls`  | 
| Nixpkgs      | `unstable` | 
```nix
{
  plugins.none-ls = {
    enable = true;
    enableLspFormat = true;
  };
}
```
image

according to the docs it should enable `lsp-format` plugin by self.

MattSturgeon:
Part one refactoring none-ls (#1705): switch the top-level plugin options to use mkNeovimPlugin and rfc42-style settings.

### lsp-format

This PR maintains the existing option `enableLspFormat`, a legacy option that sets none-ls's `on_attach` to `require('lsp-format').on_attach`. There are some issues with this option, but removing or refactoring it is not a focus for this initial PR.

I did update the option description though: fixes #1690.

### `with()` args

The various sources have a `withArgs` option (`plugins.none-ls.sources.*.*.withArgs`), which is a lua-string type. Switching this to rfc42-style settings would be nice, but is not in scope for this PR. ([Upstream docs](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md#configuration), [experimental draft PR](https://github.com/nix-community/nixvim/pull/1724)).

### Non-builtin sources

The `settings.sources` option (previously `sourcesItems`) can list builtin sources, along with custom sources and "third-party" sources. We're currently only considering builtin sources (as listed in `builtins.json`). Properly supporting other sources is not in scope for this PR.

### DIFF
--- a/plugins/none-ls/default.nix
+++ b/plugins/none-ls/default.nix
@@ -2,287 +2,154 @@
   lib,
   helpers,
   config,
+  options,
   pkgs,
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.none-ls;
-in
-{
-  imports = [
-    ./servers.nix
-    (mkRenamedOptionModule
-      [
-        "plugins"
-        "null-ls"
-      ]
-      [
-        "plugins"
-        "none-ls"
-      ]
-    )
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "none-ls";
+  originalName = "none-ls.nvim";
+  luaName = "null-ls";
+  defaultPackage = pkgs.vimPlugins.none-ls-nvim;
+
+  maintainers = [ maintainers.MattSturgeon ];
+
+  # TODO: introduced 2024-06-18, remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    "border"
+    "cmd"
+    "debounce"
+    "debug"
+    "defaultTimeout"
+    "diagnosticConfig"
+    "diagnosticsFormat"
+    "fallbackSeverity"
+    "logLevel"
+    "notifyFormat"
+    "onAttach"
+    "onInit"
+    "onExit"
+    "rootDir"
+    "shouldAttach"
+    "tempDir"
+    "updateInInsert"
   ];
 
-  options.plugins.none-ls = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "none-ls";
-
-    package = mkOption {
-      type = types.package;
-      default = pkgs.vimPlugins.none-ls-nvim;
-      description = "Plugin to use for none-ls";
-    };
-
-    enableLspFormat = mkOption {
-      type = types.bool;
-      default = config.plugins.lsp-format.enable;
-      description = ''
-        Automatically enable the `lsp-format` plugin and configure `none-ls` accordingly.
-      '';
-      example = true;
-    };
-
-    border = helpers.defaultNullOpts.mkBorder null "`:NullLsInfo` UI window." ''
-      Uses `NullLsInfoBorder` highlight group (see [Highlight Groups](#highlight-groups)).
-    '';
-
-    cmd = helpers.defaultNullOpts.mkListOf types.str [ "nvim" ] ''
-      Defines the command used to start the null-ls server. If you do not have an
-      `nvim` binary available on your `$PATH`, you should change this to an absolute
-      path to the binary.
-    '';
-
-    debounce = helpers.defaultNullOpts.mkInt 250 ''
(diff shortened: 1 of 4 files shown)


plugins/none-ls/default.nix
```diff
@@ -2,287 +2,128 @@
...
+            require('lsp-format').on_attach
+          '';
+        };
+
+      # Summarize why lsp-format integration is enabled (used in warnings)
+      enableLspFormatStatus =
+        if options.plugins.none-ls.enableLspFormat.isDefined then
+          "You have enabled the lsp-format integration with none-ls."
+        else
+          "The lsp-format integration with none-ls was enabled automatically, because `plugins.lsp-format` is enabled.";
```
MattSturgeon, line 101: Thinking about it, I'm not actually sure how `isDefined` handles option defaults. I have a feeling it'll always be true, since the _default_ counts as a definition.

We might be able to do `length definitions == 1`, assuming the default is still listed in `definitions` when an actual definition exists?

MattSturgeon: Nope, just tested. `definitions` is always length 1, regardless of whether it's using the default value or an explicit definition.

I guess this `if` branch is pointless then. Seems a shame there's no way to test if an option is explicitly defined :thinking:

plugins/none-ls/default.nix
```diff
@@ -2,287 +2,121 @@
...
+      ./servers.nix
+      (mkRenamedOptionModule oldPluginPath basePluginPath)
+      (mkRenamedOptionModule (basePluginPath ++ [ "sourcesItems" ]) (settingsPath ++ [ "sources" ]))
+    ];
 
-  options.plugins.none-ls = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "none-ls";
+  # TODO:
+  # settingsExample = {
+  # };
```
GaetanLepage, line 61: Just a reminder

plugins/none-ls/settings.nix
```diff
@@ -0,0 +1,254 @@
...
+    #   Defines a list of sources for null-ls to register.
+    #   Users can add built-in sources (see [BUILTINS]) or custom sources (see [MAIN]).
+    #
+    #   If you've installed an integration that provides its own sources and aren't
+    #   interested in built-in sources, you don't have to define any sources here. The
+    #   integration will register them independently.
+    #
+    #   [BUILTINS]: https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTINS.md
+    #   [MAIN]: https://github.com/nvimtools/none-ls.nvim/blob/main/doc/MAIN.md
+    # '';
```
GaetanLepage, line 224: Same, this should be dealt with here I think.

tests/test-sources/plugins/none-ls.nix
```diff
@@ -21,34 +21,45 @@
       lsp-format.enable = true;
       none-ls = {
         enable = true;
-        enableLspFormat = true;
+        # This is implied:
+        # enableLspFormat = true;
```
GaetanLepage, line 25: Just remove it entirely then, no ?

GaetanLepage (changes requested):

plugins/none-ls/settings.nix
MattSturgeon: This is for future reference, currently that option should only be used by the `sources.**.enable` option implementation, hence keeping the previous description.

In the future this option may be useful for custom and/or third-party sources, so we'll likely take upstream's description into account.

I could remove the commented out upstream description if you like?

tests/test-sources/plugins/none-ls.nix
MattSturgeon: Could do, but since this is kind of a weird option (`default = plugins.lsp-format.enable`) and also an option we might remove/rework in the future, I thought it was useful to note the implication.

Up to you.

plugins/none-ls/default.nix
MattSturgeon: Done

tests/test-sources/plugins/none-ls.nix
GaetanLepage: No it's fine like this.

GaetanLepage (approved):
LGTM

plugins/none-ls/settings.nix
MattSturgeon: Since the option is `visible = false` anyway, I've included both in the description and removed the commented out one.

PR #1729, opened 2024-06-20
flake/dev: add list-plugins script/command

GaetanLepage:
image

### DIFF
--- a/flake-modules/dev/devshell.nix
+++ b/flake-modules/dev/devshell.nix
@@ -63,6 +63,11 @@
                 ${pkgs.lib.getExe pkgs.python3} -m http.server -d "$doc_derivation"/share/doc
               '';
             }
+            {
+              name = "list-plugins";
+              command = "${pkgs.python3.interpreter} ${./list-plugins.py}";
+              help = "List plugins and get implementation infos";
+            }
           ];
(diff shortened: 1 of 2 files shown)


flake-modules/dev/devshell.nix
```diff
@@ -63,6 +63,11 @@
                 ${pkgs.lib.getExe pkgs.python3} -m http.server -d "$doc_derivation"/share/doc
               '';
             }
+            {
+              name = "list-plugins";
+              command = "${pkgs.python3.interpreter} ${./list-plugins.py}";
+              help = "List plugins and get implementation infos";
```
MattSturgeon, line 69: The help text could try to imply this prints mostly maintenance related info?

MattSturgeon (approved):
LGTM!

flake-modules/dev/devshell.nix
GaetanLepage: I thought so and remembered that the devshell was actually made for... well.. development. So I felt that it was not needed.

If you have a better phrasing, I will be happy to use it though !

MattSturgeon: I think this is fine :+1:

PR #1735, opened 2024-06-21
plugins/lsp-lines: switch to mkNeovimPlugin

MattSturgeon:
~~It is a neovim plugin, but it has no setup options so use mkVimPlugin.~~

The plugin recommends several defaults and examples for `vim.diagnostic.config`, some of which we've historically set. This PR makes that a little more configurable, closes #1703.

IMO we could probably handle this better by having a `diagnostics` module (similar to `opts`, `keymaps`, etc); a freeform submodule option that gets passed directly to `vim.diagnostic.config()`. Any thoughts on this approach?

If we had a diagnostics module, would we deprecate the `extraOptions` here, or would we just `lib.mkDefault` what we need to and document other examples?

### DIFF
--- a/plugins/lsp/lsp-lines.nix
+++ b/plugins/lsp/lsp-lines.nix
@@ -6,40 +6,35 @@
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.lsp-lines;
-in
-{
-  options = {
-    plugins.lsp-lines = {
-      enable = mkEnableOption "lsp_lines.nvim";
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lsp-lines";
+  luaName = "lsp_lines";
+  originalName = "lsp_lines.nvim";
+  defaultPackage = pkgs.vimPlugins.lsp_lines-nvim;
 
-      package = helpers.mkPluginPackageOption "lsp_lines.nvim" pkgs.vimPlugins.lsp_lines-nvim;
+  # This plugin has no settings; it is configured via vim.diagnostic.config
+  hasSettings = false;
 
-      currentLine = mkOption {
-        type = types.bool;
-        default = false;
-        description = "Show diagnostics only on current line";
-      };
-    };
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Tagging @PowerUser64

traxys:
I think a vim.diagnostics module makes sense

MattSturgeon:
Blocked on #1737

plugins/lsp/lsp-lines.nix
```diff
@@ -6,40 +6,36 @@
...
+    # Strongly recommended by the plugin, to avoid duplication.
+    diagnostics.virtual_text = mkDefault false;
 
-          vim.diagnostic.config(${helpers.toLuaObject diagnosticConfig})
-        end
-      '';
-    };
+    extraConfigLua = ''
+      require("lsp_lines").setup()
+    '';
```
MattSturgeon, line 39: IMO, mkNeovimPlugin should support this usage. Maybe add a `hasSettings ? true` argument which decides whether to create a `settings` option?

GaetanLepage: Yes, I am facing the same issue with md-tui.
I propose to hold this while waiting for the `mk[Neo]VimPlugin` refactoring.

MattSturgeon: Hope I'm not stepping on your toes, but for this specific issue I've made [a small PR](https://github.com/nix-community/nixvim/pull/1758) allowing `settingsOptions` to be set to `null`.

MattSturgeon: This was resolved in https://github.com/nix-community/nixvim/pull/1758

GaetanLepage (approved):

MattSturgeon:
@PowerUser64 this now uses a `mkDefault` on the new `diagnostics` module, so to achieve what you wanted in #1703 you would just set `diagnostics.virtual_text = null` or `diagnostics.virtual_text = true`.

PR #1736, opened 2024-06-21
plugins/nvim-osc52: mark as obsolete

MattSturgeon:
As per the notice on the [plugin's readme](https://github.com/ojroques/nvim-osc52/blob/main/README.md)

Tagging @PowerUser64

Closes #1704

### DIFF
--- a/plugins/utils/nvim-osc52.nix
+++ b/plugins/utils/nvim-osc52.nix
@@ -1,14 +1,31 @@
+# TODO: As of nvim 0.10 this plugin is obsolete.
+# warning added 2024-06-21, remove after 24.11.
 {
   lib,
   helpers,
   pkgs,
   config,
+  options,
   ...
 }:
 with lib;
 {
   options.plugins.nvim-osc52 = {
-    enable = mkEnableOption "nvim-osc52, a plugin to use OSC52 sequences to copy/paste";
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to enable nvim-osc52, a plugin to use OSC52 sequences to copy/paste.
+
+        Note: this plugin is obsolete and will be removed after 24.11.
+        As of Neovim 0.10 (specifically since [this PR][1]), native support for OSC52 has been added.
+        Check [`:h clipboard-osc52`][2] for more details.
+
+        [1]: https://github.com/neovim/neovim/pull/25872
+        [2]: https://neovim.io/doc/user/provider.html#clipboard-osc52
+      '';
+    };
 
     package = helpers.mkPluginPackageOption "nvim-osc52" pkgs.vimPlugins.nvim-osc52;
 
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Should the warnings/desc-notes reference our [clipboard.providers](https://nix-community.github.io/nixvim/clipboard/providers/index.html#clipboardproviders) option?

PowerUser64:
I wouldn't say this plugin is obsolete, even though its core functionality is integrated into neovim. It still provides the option to explicitly copy to/from OSC 52 without using neovim's way of setting the provider to OSC 52 (see just above [here](https://neovim.io/doc/user/provider.html#paste) in the docs for the recommended method). That recommended method isn't compatible with which-key, which used to freeze when trying to read from OSC, so they had to disable it entirely https://github.com/folke/which-key.nvim/blob/0099511294f16b81c696004fa6a403b0ae61f7a0/lua/which-key/plugins/registers.lua#L61. I think being able to explicitly copy over OSC while having a clipboard provider set is still a useful thing to be able to do.

For example, if I started a tmux session on my desktop and then tried to access it over SSH from another computer, neovim wouldn't know I'm connected over SSH and would use the built-in clipboard provider for "+, which would send the clipboard contents to the host machine, rather than the machine I'm connected from. In this case, I would need to have a way to explicitly copy over OSC 52, like what is provided by this plugin. (Because the tmux session is started from the normal desktop, the ssh environment variables won't be set within the tmux session, so neovim won't auto-enable the OSC 52 clipboard provider like it's supposed to.)

MattSturgeon:
> I wouldn't say this plugin is obsolete, even though its core functionality is integrated into neovim.

Thanks for your feedback! I've read your response and you have a point. I won't reply to each part to try and keep this brief.

Just to clarify, _I'm_ not calling it obsolete; upstream added a note stating that their plugin is obsolete 7 months ago: https://github.com/ojroques/nvim-osc52/commit/5e0210990b3c809ec58bbf830e0dabd4bda3a949

Since then, the only contributions (two PRs) have been merged in the last month, which might imply the project is no longer particularly active.

If you don't think it's obsolete, maybe they'd accept a PR clarifying the wording slightly? Something along the lines of "Note: neovim 0.10 added osc52 support, however this plugin provides an alternative implementation with xyz differences".

@nix-community/nixvim do you think it is still worth the effort to update this plugin to use mkNeovimPlugin (instead of adding this warning), given that some users still find it useful despite upstream declaring it "obsolete"?

Maybe add the warning but provide a way to silence it?

PowerUser64:
Been thinking about this. I think that we should leave the job of making any warnings about this to the plugin itself. If the authors think it's necessary to notify people, they can do it themselves. Otherwise we can just leave it up to people who use the plugin to pay attention to `:help news` and the nvim release notes to learn about things like this. So long as it doesn't cause problems in peoples' setups or mounds of issues for us, it probably doesn't matter. Less "magic" for us. That's my take, anyway.

Maybe we could have a Nixvim news channel somewhere for things like this? Something like an issue titled "News - Subscribe for updates" or something that gets added to the underlying neovim installation that the user could do `:help nixvim-news` and get info on things like this.

MattSturgeon:
In principle I agree we should do less. But in practice, we can't expect deprecated and/or abandoned plugins to come back and add a notice at runtime, we're lucky if they add one to their readme (speaking more generally than just this plugin).

The more pressing issues here is a maintenance one: we don't want old non-rfc42 modules sitting around indefinitely. Therefore, this plugin module should either be updated to use mkNeovimPlugin, or be deprecated with a view for removal. (Or some combination of the two).

I don't mind which approach we take. I initially went down the deprecation route after finding the notice on upstream's readme when I was starting to do the mkNeovimPlugin update.

I'd like some input from @GaetanLepage or @traxys on this one.

traxys:
If I understood the documentation it can be replaced with `require('vim.ui.clipboard.osc52').copy/paste('<register>')`, so I think deprecating it is fine, by linking to the nvim doc (https://neovim.io/doc/user/provider.html#clipboard-osc52) or adding some snippets in our doumentation or maybe deperecation warning

GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

PR #1739, opened 2024-06-21
plugins/auto-save: switch to mkNeovimPlugin + update options

braindefender:
This PR switches the plugin to `mkNeovimPlugin` and updates plugin options according to [okuuva/auto-save.nvim](https://github.com/okuuva/auto-save.nvim?tab=readme-ov-file#%EF%B8%8F-configuration) configuration options.

### DIFF
--- a/plugins/utils/auto-save.nix
+++ b/plugins/utils/auto-save.nix
@@ -6,30 +6,62 @@
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.auto-save;
-in
-{
-  options.plugins.auto-save = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "auto-save";
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "auto-save";
+  originalName = "auto-save.nvim";
+  defaultPackage = pkgs.vimPlugins.auto-save-nvim;
 
-    package = helpers.mkPluginPackageOption "auto-save" pkgs.vimPlugins.auto-save-nvim;
+  maintainers = [ helpers.maintainers.braindefender ];
 
-    keymaps = {
-      silent = mkOption {
-        type = types.bool;
-        description = "Whether auto-save keymaps should be silent.";
-        default = false;
-      };
+  # TODO: introduced 2024-06-21, remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    [
+      "executionMessage"
+      "message"
+    ]
+    [
+      "executionMessage"
+      "dim"
+    ]
+    [
+      "executionMessage"
+      "cleaningInterval"
+    ]
+    "triggerEvents"
+    "writeAllBuffers"
+    "debounceDelay"
+  ];
 
-      toggle = helpers.mkNullOrOption types.str "Keymap for running auto-save.";
-    };
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "auto-save"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    [
+      (mkRenamedOptionModule (basePluginPath ++ [ "enableAutoSave" ]) (settingsPath ++ [ "enabled" ]))
+      (mkRemovedOptionModule (basePluginPath ++ [ "keymaps" ]) ''
+        Use the top-level `keymaps` option to create a keymap that runs :ASToggle
 
-    enableAutoSave = helpers.defaultNullOpts.mkBool true ''
+        keymaps = [
+          { key = "<leader>s"; action = "<cmd>ASToggle<CR>"; }
+        ];
+      '')
+    ];
+
+  settingsOptions = {
+    enabled = helpers.defaultNullOpts.mkBool true ''
       Whether to start auto-save when the plugin is loaded.
     '';
 
-    executionMessage = {
+    execution_message = {
+      enabled = helpers.defaultNullOpts.mkBool true ''
+        Show execution message after successful auto-save.
+      '';
+
       message =
         helpers.defaultNullOpts.mkStr
           {
@@ -40,7 +72,7 @@ in
             '';
           }
           ''
-            The message to print en save.
+            The message to print on save.
             This can be a lua function that returns a string.
           '';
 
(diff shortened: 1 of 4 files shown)


MattSturgeon:
Thanks for highlighting and fixing this issue!

Would you be willing to do this as part of a refactor to have the plugin use our newer `mkNeovimPlugin` helper? This would mean there would be a "freeform" `settings` option (passed directly to `setup`), so not having all options wouldn't limit users as they can define any config.

You can reference the [template](https://github.com/nix-community/nixvim/blob/main/plugins/TEMPLATE.nix), the [contributing guide](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md), and [recently merged PRs](https://github.com/nix-community/nixvim/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged+mkNeovimPlugin). Or one of us can take over when we get chance.

plugins/utils/auto-save.nix
```diff
@@ -6,65 +6,94 @@
...
-    };
-
-    enableAutoSave = helpers.defaultNullOpts.mkBool true ''
-      Whether to start auto-save when the plugin is loaded.
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "auto-save";
+  originalName = "auto-save.nvim";
+  defaultPackage = pkgs.vimPlugins.auto-save-nvim;
+
+  maintainers = [ maintainers.braindefender ];
```
MattSturgeon, line 14:
```suggestion
  maintainers = [ helpers.maintainers.braindefender ];
```

Also, could you move adding yourself to the list to a separate commit, titled "maintainers: add braindefender"

plugins/utils/auto-save.nix
```diff
@@ -6,65 +6,94 @@
...
-
-    keymaps = {
-      silent = mkOption {
-        type = types.bool;
-        description = "Whether auto-save keymaps should be silent.";
-        default = false;
-      };
-
-      toggle = helpers.mkNullOrOption types.str "Keymap for running auto-save.";
-    };
```
MattSturgeon, line 26: The keymaps option can be maintained by putting it in `extraOptions` (as opposed to `settingsOptions`)

plugins/utils/auto-save.nix
```diff
@@ -73,72 +102,57 @@ in
...
-      '';
-
-      keymaps =
-        with cfg.keymaps;
-        optional (toggle != null) {
-          mode = "n";
-          key = toggle;
-          action = ":ASToggle<CR>";
-          options.silent = cfg.keymaps.silent;
-        };
```
MattSturgeon, line 142: This could be kept, using an `extraConfig` attribute:

```nix
extraConfig = cfg: {
  keymaps = mkIf (toggle != null) {
    mode = "n";
    key = cfg.keymaps.toggle;
    action = ":ASToggle<CR>";
    options.silent = cfg.keymaps.silent;
  };
};
```

(Note: prefer `mkIf` over `optional` for config definitions)

plugins/utils/auto-save.nix
```diff
@@ -6,65 +6,94 @@
...
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "auto-save";
+  originalName = "auto-save.nvim";
+  defaultPackage = pkgs.vimPlugins.auto-save-nvim;
+
+  maintainers = [ helpers.maintainers.braindefender ];
+
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    "enableAutoSave"
```
MattSturgeon, line 18:
```suggestion
```

This will need a custom rename warning, since the old name is different to the new name.

You can add this into an `imports = [ ]` next to `optionsRenamedToSettings`:

```nix
imports = let
  basePluginPath = ["plugins" "auto-save"];
  settingsPath = basePluginPath ++ [ "settings" ];
in
  [
    (mkRenamedOptionModule  (basePluginPath ++ [ "enableAutoSave" ]) (settingsPath ++ [ "enabled" ]))
  ];
```

(If you have a few of those, we usually declare a `basePluginPath` and `settingsPath` in a let block, see [this, for example](https://github.com/nix-community/nixvim/blob/4e224d27ae0db889f92ef228ad6cb984b91777fb/plugins/utils/rest.nix#L31-L47))

plugins/utils/auto-save.nix
```diff
@@ -6,65 +6,94 @@
...
-    enableAutoSave = helpers.defaultNullOpts.mkBool true ''
-      Whether to start auto-save when the plugin is loaded.
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "auto-save";
+  originalName = "auto-save.nvim";
+  defaultPackage = pkgs.vimPlugins.auto-save-nvim;
+
+  maintainers = [ helpers.maintainers.braindefender ];
+
+  deprecateExtraOptions = true;
```
MattSturgeon, line 16:
```suggestion
    # TODO: introduced 2024-06-21, remove after 24.11
    deprecateExtraOptions = true;
```

tests/test-sources/plugins/utils/auto-save.nix
```diff
@@ -3,47 +3,71 @@
...
 
-          if fn.getbufvar(buf, "&modifiable") == 1 and utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
-            return true -- met condition(s), can save
+  defaults = {
+    plugins.auto-save = {
+      enable = true;
+
+      settings = {
+        enabled = true;
+        executionMessage = {
```
MattSturgeon, line 34: Settings names use the same casing as upstream (i.e. snake_case here: `execution_message`)

tests/test-sources/plugins/utils/auto-save.nix
```diff
@@ -3,47 +3,71 @@
...
+        };
+        condition = ''
+          function(buf)
+            local fn = vim.fn
+            local utils = require("auto-save.utils.data")
+
+            if utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
+              return true
+            end
+            return false
```
MattSturgeon, line 63: Upstream default is `nil`, so this should be `null`.

plugins/utils/auto-save.nix
MattSturgeon: `optionsRenamedToSettings` is [smart enough](https://github.com/nix-community/nixvim/blob/4e224d27ae0db889f92ef228ad6cb984b91777fb/lib/neovim-plugin.nix#L90-L95) to translate camelCase into snake_case, but for anything more complex we need to use [`mkRenamedOptionModule`](https://noogle.dev/f/lib/mkRenamedOptionModule) or [`mkRemovedOptionModule`](https://noogle.dev/f/lib/mkRemovedOptionModule) explicitly.

MattSturgeon (changes requested):
Thanks for persisting with this! I've left some comments below, don't hesitate to ask if you're not sure on anything in particular :smiley:

tests/test-sources/plugins/utils/auto-save.nix
MattSturgeon: The function could be moved to an example though

plugins/utils/auto-save.nix
```diff
@@ -73,72 +102,57 @@ in
...
 
-            if
-              fn.getbufvar(buf, "&modifiable") == 1 and utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
-              return true -- met condition(s), can save
+            if utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
+              return true
             end
-            return false -- can't save
+            return false
           end
```
MattSturgeon, line 109: This first value should be the plugin default. (same feedback as the test)

You could do this to define an example, though:

```nix
condition = helpers.defaultNullOpts.mkLuaFn' {
  pluginDefault = null;
  description = ''
    Function that determines whether to save the current buffer or not.
    - return true: if buffer is ok to be saved
    - return false: if it's not ok to be saved

    In this example, the second argument of `utils.not_in(..., {})`
    determines which filetypes will be ignored by auto-save plugin.
    Buffers that are `nomodifiable` are not saved by default.
  '';
  example = ''
    function(buf)
      local fn = vim.fn
      local utils = require("auto-save.utils.data")
      if
        fn.getbufvar(buf, "&modifiable") == 1 and utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
        return true -- met condition(s), can save
      if utils.not_in(fn.getbufvar(buf, "&filetype"), {}) then
        return true
      end
      return false -- can't save
      return false
    end
  '';
};
```

plugins/utils/auto-save.nix
braindefender: As for keymaps, there is a [section in the plugin repo](https://github.com/okuuva/auto-save.nvim/?tab=readme-ov-file#-usage) that describes how to add keymap. But it is not default or anything so I decided to remove them.

plugins/utils/auto-save.nix
braindefender: Yes. This will be better. Done.

plugins/utils/auto-save.nix
MattSturgeon: If you don't want to keep the `keymaps` option, you should add a [mkRemovedOptionModule](https://noogle.dev/f/lib/mkRemovedOptionModule) assertion to `imports`, with migration instructions.

Something along these lines:

```nix
(mkRemovedOptionModule (basePluginPath ++ [ "keymaps" ]) ''
  Use the top-level `keymaps` option to create a keymap that runs :ASToggle

  keymaps = [
    { key = "<leader>s"; action = "<cmd>ASToggle<CR>"; }
  ];
'')
```

(note the warning is printed as stderr, so no fancy markdown formatting)

MattSturgeon: On a technical level, though, we _can_ continue to support the option (to avoid breaking people's configs).

`extraOptions` sit alongside the settings options; unlike settings, these don't have to map directly to upstream's options and can be anything we think is useful enough to continue maintaining.

I do agree that this particular keymap is pretty basic though, and we probably wouldn't add it if it were a new plugin. So up to you.

plugins/utils/auto-save.nix
```diff
@@ -6,139 +6,173 @@
...
       message =
-        helpers.defaultNullOpts.mkStr
-          {
-            __raw = ''
-              function()
-                return ("AutoSave: saved at " .. vim.fn.strftime("%H:%M:%S"))
-              end
-            '';
-          }
+        helpers.defaultNullOpts.mkLuaFn
```
MattSturgeon, line 66: I think this change is wrong, [in the source](https://github.com/okuuva/auto-save.nvim/blob/5fe9ab0c42f0457f2a973e814a6352b8eeb04730/lua/auto-save/config.lua#L8) it is annotated as `string|fun():string`, so we want `mkStr` with a `__raw` plugin-default (& test), as it was before.

plugins/utils/auto-save.nix
```diff
@@ -40,105 +72,109 @@ in
...
+    trigger_events = {
+      immediate_save =
+        helpers.defaultNullOpts.mkListOf types.str
+          [
+            "BufLeave"
+            "FocusLost"
+          ]
+          ''
+            Vim events that trigger an immediate save.
+            See `:h events` for events description.
```
MattSturgeon, line 99: Just an FYI in case you're not aware: this will (usually) render as one line of text. That's fine if the linebreak is only there to make the code more readable and not needed in the actual docs.

If you _did_ want a "real" line break, markdown supports two kings.

A line break\
like this\
is done by ending the line with a `\`:

```md
A line break\
like this\
is done by ending the line with a `\`:
```

Or a full paragraph break

is done by including one-or-more blank lines:

```md
Or a full paragraph break

is done by including one-or-more blank lines:
```

tests/test-sources/plugins/utils/auto-save.nix
```diff
@@ -3,47 +3,61 @@
...
+    plugins.auto-save = {
+      enable = true;
+
+      settings = {
+        enabled = true;
+        execution_message = {
+          enabled = true;
+          dim = 0.18;
+          cleaning_interval = 1250;
+          message = ''
```
MattSturgeon, line 38:
```suggestion
          message.__raw = ''
```

plugins/utils/auto-save.nix
braindefender: > I do agree that this particular keymap is pretty basic though, and we probably wouldn't add it if it were a new plugin.

Yep, I think it should be dropped for consistency with other/new plugins.

tests/test-sources/plugins/utils/auto-save.nix
braindefender: Oh, the tests.. Fixed!

plugins/utils/auto-save.nix
braindefender: Hmm. Thanks for the information. I didn't know about `\`.
It'll be better with line breaks, of course.

MattSturgeon: I think it's fine as-is, just wanted to make sure it was a deliberate decision either way.

Most of our descriptions reserve line breaks for fairly significant, well, breaks

MattSturgeon (approved):
LGTM. Thanks for working on this and persevering through the review.

I'll give @GaetanLepage or @traxys chance to take a look, then merge tomorrow assuming no further changes requested from them.

If you're able, could you `git fetch upstream`\* and then rebase on `upstream/main`

This will let us merge _your_ original commit, which is nice but not necessary.

\*or whatever you've saved our repo as in your git remotes

MattSturgeon:
One more _tiny_ (one-letter) nit-pick, can you change `plugin/` to `plugins/` in your commit message when you rebase?

braindefender:
> LGTM. Thanks for working on this and persevering through the review.

Thanks for the review and suggestions.

> If you're able, could you `git fetch upstream`* and then rebase on `upstream/main`
> This will let us merge _your_ original commit, which is nice but not necessary.
> *or whatever you've named our repo in your git remotes

Done.

plugins/utils/auto-save.nix
```diff
@@ -6,30 +6,62 @@
...
+
+        keymaps = [
+          { key = "<leader>s"; action = "<cmd>ASToggle<CR>"; }
+        ];
+      '')
+    ];
+
+  settingsOptions = {
+    enabled = helpers.defaultNullOpts.mkBool true ''
+      Whether to start auto-save when th plugin is loaded.
```
GaetanLepage, line 57:
```suggestion
      Whether to start auto-save when the plugin is loaded.
```

GaetanLepage (approved):
LGTM ! Thank you for this very good first PR :)

PR #1744, opened 2024-06-22
colorschemes/everforest: init

Sheemap:
Adding the everforest colorscheme https://github.com/sainnhe/everforest

Fairly new to nix, I would super appreciate any feedback yall have!

I pulled all the settings info strings directly out of `:help everforest`

### DIFF
--- a/plugins/colorschemes/everforest.nix
+++ b/plugins/colorschemes/everforest.nix
@@ -1,0 +1,265 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+let
+  mkEverforestBool =
+    pluginDefault: desc:
+    helpers.defaultNullOpts.mkEnum [
+      0
+      1
+    ] pluginDefault desc;
+
+in
+helpers.vim-plugin.mkVimPlugin config {
+  name = "everforest";
+  isColorscheme = true;
+  defaultPackage = pkgs.vimPlugins.everforest;
+  globalPrefix = "everforest_";
+
+  maintainers = [ helpers.maintainers.sheemap ];
+
+  settingsOptions = {
+    background =
+      helpers.defaultNullOpts.mkEnum
+        [
+          "hard"
+          "medium"
+          "soft"
+        ]
+        "medium"
+        ''
+          The background contrast used in this color scheme.
+        '';
+
+    enable_italic = mkEverforestBool 0 ''
+      To enable italic in this color scheme, set this option to `1`.
+    '';
+
+    disable_italic_comment = mkEverforestBool 0 ''
+      By default, italic is enabled in `Comment`. To disable italic in `Comment`, set this option to `1`
+    '';
+
+    cursor =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "auto"
+          "red"
+          "orange"
+          "yellow"
+          "green"
+          "aqua"
+          "blue"
+          "purple"
+        ]
+        ''
+          Customize the cursor color, only works in GUI clients.
+        '';
+
+    transparent_background =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          0
+          1
+          2
+        ]
+        ''
+          To use transparent background, set this option to `1`.
+
+          If you want more ui components to be transparent (for example, status line
+          background), set this option to `2`.
+        '';
+
+    dim_inactive_windows = mkEverforestBool 0 ''
+      Dim inactive windows. Only works in Neovim currently.
+
+      When this option is used in conjunction with |g:everforest_show_eob| set to 0,
+      the end of the buffer will only be hidden inside the active window. Inside
+      inactive windows, the end of buffer filler characters will be visible in
+      dimmed symbols. This is due to the way Vim and Neovim handle |hl-EndOfBuffer|.
+    '';
+
+    sign_column_background =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "none"
+          "grey"
+        ]
+        ''
+          By default, the color of sign column background is the same as normal text
+          background, but you can use a grey background by setting this option to `'grey'`.
+        '';
+
+    spell_foreground =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "none"
+          "colored"
+        ]
+        ''
+          By default, this color scheme won't color the foreground of |spell|, instead
+          colored under curls will be used. If you also want to color the foreground,
+          set this option to `'colored'`.
+        '';
+
+    ui_contrast =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          "low"
+          "high"
+        ]
+        ''
+          The contrast of line numbers, indent lines, etc.
+        '';
+
+    show_eob = mkEverforestBool 1 ''
+      Whether to show |hl-EndOfBuffer|.
+    '';
+
+    float_style =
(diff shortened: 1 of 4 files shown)


GaetanLepage (changes requested):
This looks very solid to me at a first glance.
Could you please add a test ? (Refer to similar PRs/files to draw inspiration from).

GaetanLepage:
Looks like there is something wrong with the tests.
You can debug locally by entering the devshell (`nix develop` or using direnv).
Then, simply run the `tests` command.

Sheemap:
Got it, thank you!

So it looks like the issue is due to the `better_performance` flag being set to 1. This causes the everforest plugin to write stuff to the disk for caching or something.

Problem is Nix's immutability is not allowing that, and it fails. Theres no setting I can find to tweak the cache location.

Seems to me an upstream problem. What would you recommend I do to resolve this? Is there way to resolve this within Nixvim? Or perhaps just modifying the docs and raising an assertion error if that setting is set?

MattSturgeon:
> Problem is Nix's immutability is not allowing that, and it fails. Theres no setting I can find to tweak the cache location.

If so, that should only be affecting the test and should be fine when running normally? Unless the plugin is trying to write its cache to the nix store.

If it is just the test that's affected, you can disable whatever option is needed. Along with a comment, explaining why said option causes tests to fail.

Sheemap:
Unfortunately no its trying to write to the nix store. Can turn it off in the test, but this feature will not work in a real environment

```
line 5: Vim(call):E739: Cannot create directory /nix/store/0y403ixxd7jkdxvsjwl6alkayf7qxsza-vimplugin-everforest-2024-06-02/after: read-only file system                                                                                                                                                                                                                                         
```

MattSturgeon:
> Unfortunately no, its trying to write to the nix store. Can turn it off in the test, but this feature will not work in a real environment

Frustrating. We'll just have to disable the test and report an issue upstream, recommending they use `XDG_CACHE_HOME`. If an issue is open before this PR is done, we can reference it in a comment.

As for the option, we don't actually have to explicitly declare all availabile settings options anyway, since `settings` is a freeform option, users can define any config they like, even stuff we haven't added options for.

In some respects it's better not to define all available options; once we've added them we then have to try and keep them up to date.

Sheemap:
Cool sounds good! Ive removed the option from the settings. I'm happy to remove any others, though I'm not sure which ones are important enough to be kept in vs removed

tests/test-sources/plugins/colorschemes/everforest.nix
```diff
@@ -0,0 +1,42 @@
...
+        show_eob = 0;
+        float_style = "dim";
+        diagnostic_text_highlight = 1;
+        diagnostic_line_highlight = 1;
+        diagnostic_virtual_text = "colored";
+        current_word = "bold";
+        inlay_hints_background = "dimmed";
+        disable_terminal_colors = 1;
+        lightline_disable_bold = 1;
+        colors_override = {
```
MattSturgeon, line 29:
```suggestion
        # This option is broken, because the plugin tries to write its cache in /nix/store
        # better_performance = 1;
        colors_override = {
```

I'd rather have a comment rather than a silently missing option, but this is fine.

Just a preference not a request.

MattSturgeon: If removing it is enough to fix the test, it makes me think that `1` _is not_ the default value anyway, making me second guess some of the other defaults listed here.

plugins/colorschemes/everforest.nix
```diff
@@ -0,0 +1,266 @@
...
+  pkgs,
+  ...
+}:
+let
+  mkEverforestBool =
+    pluginDefault: desc:
+    helpers.defaultNullOpts.mkEnum [
+      0
+      1
+    ] pluginDefault desc;
```
MattSturgeon, line 14: I think a type alias and/or option helper for `enum [0 1]` would be helpful in all of nixvim. Maybe in a future PR we'll do that

plugins/colorschemes/everforest.nix
```diff
@@ -0,0 +1,266 @@
...
+
+      Currently, the following plugins are supported:
+
+      - neovim's built-in language server client
+      - https://github.com/neoclide/coc.nvim
+      - https://github.com/prabirshrestha/vim-lsp
+      - https://github.com/ycm-core/YouCompleteMe
+      - https://github.com/dense-analysis/ale
+      - https://github.com/neomake/neomake
+      - https://github.com/vim-syntastic/syntastic
```
MattSturgeon, line 151: I'd rather these be `[markdown links](url)`:

```suggestion
      - neovim's built-in language server client
      - [coc.nvim](https://github.com/neoclide/coc.nvim)
      - [vim-lsp](https://github.com/prabirshrestha/vim-lsp)
      - [YouCompleteMe](https://github.com/ycm-core/YouCompleteMe)
      - [ale](https://github.com/dense-analysis/ale)
      - [neomake](https://github.com/neomake/neomake)
      - [syntastic](https://github.com/vim-syntastic/syntastic)
```

Same for links in other option descriptions.

plugins/colorschemes/everforest.nix
```diff
@@ -0,0 +1,266 @@
...
+    '';
+
+    lightline_disable_bold = mkEverforestBool 0 ''
+      By default, bold is enabled in lightline color scheme. To disable bold in
+      lightline color scheme, set this option to `1`.
+    '';
+
+    colors_override = helpers.defaultNullOpts.mkAttrsOf (lib.types.listOf lib.types.str) { } ''
+      Override color palette. The available keys can be found in
+      `https://github.com/sainnhe/everforest/blob/master/autoload/everforest.vim`
```
MattSturgeon, line 245:
```suggestion
      The plugin's [source code](https://github.com/sainnhe/everforest/blob/master/autoload/everforest.vim).
```

MattSturgeon:
This is an excellent first time contribution! I hope you don't mine, I left some further feedback and discussion points.

- I'd prefer links to use markdown link syntax, so they are actually clickable in the docs and render nicely.
- I'd prefer a comment be left about the broken option (up to you)
- I'm confused as to why removing a default value has any effect; the whole point of the "plugin defaults" is they should match the plugin's defaults, as used when the option is not defined

Sheemap:
Thank you! :) Ive addressed all those concerns.

The `better_performance = 1` value is not the default, I actually intentionally made _all_ the test values non-default to potentially catch any that I didnt find on my mostly unmodified install (which worked in this case, wahoo :smile:) 

Do you expect the tests to be testing the default values? Is there a way I can create multiple tests in this file to test both the default and non-default cases?

tests/test-sources/plugins/colorschemes/everforest.nix
Sheemap: Done! Thanks for the suggestion, definitely agree with your preference here.

Sheemap:
OH, I did not realize the test values are literally named as `plugin.defaults`! Only after re-reading your comment did I understand. Ill update those to be default values haha

MattSturgeon:
> Thank you! :) Ive addressed all those concerns.

Brilliant 

> Do you expect the tests to be testing the default values? Is there a way I can create multiple tests in this file to test both the default and non-default cases?

Yes you can create as many test as you like, generally "empty" and "defaults" are the bare minimum, but "example" is also commonly used to test non-defaults.

Tests can be named anything, those are just the most common names we use.

MattSturgeon (approved):
Great! Thanks for your persistence and patience!

PR #1750, opened 2024-06-24
plugins/gitsigns: remove deprecated options

MattSturgeon:
- **lib/deprecation: init with mkRemovedSubmoduleOptionModule**
- **plugins/gitsigns: remove deprecated options**

Added a clone of `mkRemovedOptionModule` that supports nested sub-options. This is exactly the same, except is uses our custom `getOptionRecursive` instead of `getAttrFromPath`.

Like `mkRemovedOptionModule`, this is an assertion. Maybe a warning would make sense instead, seeing as `settings` is a freeform type?

I've tested the functions a little in the repl and by checking out the old test file (before the options were removed from the test).

Should we have some lib-tests? If so do you have any suggestions for test-cases?

### DIFF
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -1,0 +1,49 @@
+{ lib, ... }:
+with lib;
+rec {
+  # Get a (sub)option by walking the path,
+  # checking for submodules along the way
+  getOptionRecursive =
+    opt: prefix: optionPath:
+    if optionPath == [ ] then
+      opt
+    else if isOption opt then
+      getOptionRecursive (opt.type.getSubOptions prefix) prefix optionPath
+    else
+      let
+        name = head optionPath;
+        opt' = getAttr name opt;
+        prefix' = prefix ++ [ name ];
+        optionPath' = drop 1 optionPath;
+      in
+      getOptionRecursive opt' prefix' optionPath';
+
+  # Like mkRemovedOptionModule, but has support for nested sub-options
+  # and uses warnings instead of assertions.
+  mkDeprecatedSubOptionModule =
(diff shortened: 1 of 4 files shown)


traxys:
I feel like we should mostly stick to warnings for RFC-42 style plugins when options are removed by upstream, but I don't feel that very strongly

MattSturgeon:
This isn't working. It looks like `isDefined` is always true. Maybe the freeformType is causing this?

MattSturgeon:
All issues now resolved (see above edits).

I wasn't able to avoid the option having a value in `cfg.settings`. Doing so would result in option used but not defiend, because the option value is evaluated when (e.g.) `toLuaObject` uses `mapAttrs` on the settings submodule.

This happens even when the opt is `isDefined = false`.

Therefore, any undefiend option will show up in `cfg.settings` as `null`.

traxys:
We can set visible=false to hide them from the documentation

GaetanLepage (approved):
LGTM, thanks @MattSturgeon !

One thought: It wouldn't shock me to have `mkDeprecatedSubOptionModule` under top-level `helpers`. The name is clear enough to not have to put it under `helpers.deprecation`.

traxys (approved):

MattSturgeon:
> One thought: It wouldn't shock me to have `mkDeprecatedSubOptionModule` under top-level `helpers`. The name is clear enough to not have to put it under `helpers.deprecation`.

Ok, I'll do that.

I think at some point it'd be nice to restructure our lib closer to how nixpkgs/lib does it, with everything scoped but some top-level aliases added where appropriate.

PR #1758, opened 2024-06-25
lib/neovim-plugin: more configurable settings

MattSturgeon:
### Motivation
Sometimes we use `mkVimPlugin` for neovim plugins just because we want to avoid creating the `settings` option. This isn't ideal at the best of times, but makes even less sense when we still need to call `setup` anyway.

### Implementation
Allow callers to explicitly set `settingsOptions = null` to disable creating the `settings` option.
It still defaults to `{ }`, so existing plugins are not affected.

`toLuaObject cfg.settings` is now an optional string.

### Descriptions

I've incorporated changes from #1676 that felt more suited to this PR than that one.

Namely:
- Allow overriding the settings description
- Allow overriding the `setup` function name (used in config & description)

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -5,20 +5,7 @@
   nixvimUtils,
 }:
 with lib;
-rec {
-  mkSettingsOption =
-    {
-      pluginName ? null,
-      options ? { },
-      description ?
-        if pluginName != null then
-          "Options provided to the `require('${pluginName}').setup` function."
-        else
-          throw "mkSettingsOption: Please provide either a `pluginName` or `description`.",
-      example ? null,
-    }:
-    nixvimOptions.mkSettingsOption { inherit options description example; };
-
+{
   # TODO: DEPRECATED: use the `settings` option instead
   extraOptionsOptions = {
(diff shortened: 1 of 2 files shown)


lib/neovim-plugin.nix
```diff
@@ -95,17 +95,19 @@ rec {
...
-          options = settingsOptions;
-          example = settingsExample;
-        };
-      } // extraOptions;
+      options.${namespace}.${name} =
+        {
+          enable = mkEnableOption originalName;
+          package = nixvimOptions.mkPluginPackageOption originalName defaultPackage;
+        }
+        // optionalAttrs (settingsOptions != null) {
```
MattSturgeon, line 103: We could instead have a `hasSettings ? true` argument, I just figured you'd prefer making the existing arg nullable.

GaetanLepage: I don't know. Maybe the `hasSettings` will be less prone to unwanted behavior.

MattSturgeon: Done

GaetanLepage (approved):
Nothing to add !

PR #1759, opened 2024-06-26
add overseer.nvim (WIP)

PowerUser64:
This is a dependency of compiler.nvim and a plugin of its own. Not done yet.

TODO:
- [ ] Finish translating options
- [ ] Write a test using all options

### DIFF
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -192,6 +192,7 @@
     ./utils/obsidian
     ./utils/oil.nix
     ./utils/ollama.nix
+    ./utils/overseer.nix
     ./utils/persistence.nix
     ./utils/presence-nvim.nix
     ./utils/project-nvim.nix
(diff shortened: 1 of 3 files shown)


MattSturgeon:
Looks good at first glance, thanks!

Let us know if there's anything in particular you need help or guidance with.

> - Finish translating options

Don't feel you need to have nix options for _all_ plugin options; remember `settings` is "freeform", so anything can be configured even if there's no corresponding option.

Having an option just provides documentation and type-checking.

> - Write a test using all options

It's nice to have example test cases and/or tests for specific usages. Generally "defaults" is the only one that usually covers _all_ options.

Note that tests (and `settingsExample`) _can_ contain config that you haven't created an explicit option for (because `settings` is freeform).

GaetanLepage:
@PowerUser64 do you want further help with this ?

GaetanLepage:
Followed up in #2249

PR #1761, opened 2024-06-26
lib/lua: refactor toLuaObject, now configurable

MattSturgeon:
Refactor toLuaObject, providing a `lua.toLua'` version that takes configuration options.

`lua.toLua` is the same as `lua.toLua' { }` (default options), and is aliased at the top-level as `toLuaObject` to avoid breaking changes.

Heavily based on nixpkgs [`lib.generators.toPretty`](https://github.com/NixOS/nixpkgs/blob/00ba4c2c35f5e450f28e13e931994c730df05563/lib/generators.nix#L326-L398), which has the added benefit of improving the output format (more space padding) and supporting indented multiline output.

Removal of empty/null attrs has been refactored as a recursive first-pass or pre-proccess. Before we called `toLuaObject` an additional time in the filter function, which was inefficient.

All tests pass, but I should probably add new tests for the configuration options.

I deprecated `__emptyString`, because it is the same as attr name `""`.

### DIFF
--- a/lib/to-lua.nix
+++ b/lib/to-lua.nix
@@ -32,51 +32,193 @@ rec {
   # and contain only letters, digits, and underscores.
   isIdentifier = s: !(isKeyword s) && (match "[A-Za-z_][0-9A-Za-z_]*" s) == [ ];
 
+  # toLua' with default options, aliased as toLuaObject at the top-level
+  toLua = toLua' { };
+
   # Black functional magic that converts a bunch of different Nix types to their
   # lua equivalents!
-  toLuaObject =
-    args:
-    if builtins.isAttrs args then
-      if hasAttr "__raw" args then
-        args.__raw
-      else if hasAttr "__empty" args then
-        "{ }"
-      else
-        "{"
-        + (concatStringsSep ", " (
-          mapAttrsToList (
+  toLua' =
+    {
+      /**
+        If this option is true, attrsets like { __pretty = fn; val = …; }
+        will use fn to convert val to a lua representation.
+      */
+      allowPrettyValues ? false,
+
+      /**
+        If this option is true, values like { __raw = "print('hi)"; }
+        will render as print('hi')
+      */
+      allowRawValues ? true,
+
+      /**
+        If this option is true, attrsets like { "__rawKey__print('hi')" = "a"; }
+        will render as { [print('hi')] = "a"] }
+      */
+      allowRawAttrKeys ? true,
+
+      /**
+        If this option is true, attrsets like { __unkeyed.1 = "a"; b = "b"; }
+        will render as { "a", b = "b" }
+      */
+      allowUnkeyedAttrs ? true,
+
+      /**
+        If this option is true, attrsets like { a = "a"; b = "b"; }
+        will render as { a, b = "b" } instead of { ["a"] = "a", ["b"] = "b" }
+      */
+      allowUnquotedAttrKeys ? true,
+
+      # TODO: smartQuoteStyle: use ' or " based on string content
+      # TODO: allowRawStrings: use [=[ ]=] strings for multiline strings
+
+      /**
+        If this option is true, attrsets like { a = null; b = ""; }
+        will render as { ["b"] = "" }
+      */
+      removeNullAttrValues ? true,
+
+      /**
+        If this option is true, attrsets like { a = { }; b = [ ]; c = ""; }
+        will render as { ["c"] = "" }
+      */
+      removeEmptyAttrValues ? true,
+
+      /**
+        If this option is true, attrsets like { a.__empty = null; }
+        will render as { ["a"] = { } }, ignoring removeEmptyAttrValues and removeNullAttrValues.
+      */
+      allowExplicitEmpty ? true,
+
+      /**
+        If this option is true, attrsets like { __emptyString = "foo"; }
+        will render as { [""] = "foo" }.
+
+        This is deprecated: use an attrset like { "" = "foo"; } instead.
+      */
+
+      allowLegacyEmptyStringAttr ? true,
+      /**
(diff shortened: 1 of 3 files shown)


lib/to-lua.nix
```diff
@@ -32,51 +32,184 @@ rec {
...
+        in
+        if v == null then
+          "nil"
+        else if isInt v then
+          toString v
+        # toString loses precision on floats, so we use toJSON instead. This isn't perfect
+        # as the resulting string may not parse back as a float (e.g. 42, 1e-06), but for
+        # pretty-printing purposes this is acceptable.
+        else if isFloat v then
+          builtins.toJSON v
```
MattSturgeon, line 172: IDK what the best approach is for lua. Previously we just did `toString v`, but the precision notice copied from `toPretty` is kinda scary :confused:

traxys: Not a problem for lua I think, the only data type is "number"

MattSturgeon: Ok, I'll amend the comment to something along the lines of "toString can reduce precision, so we use toJSON. Although this can convert to int, that shouldn't matter for lua."

lib/to-lua.nix
```diff
@@ -32,51 +32,184 @@ rec {
...
+        else if allowUnquotedAttrKeys && isIdentifier s then
+          s
+        else if allowLegacyEmptyStringAttr && s == "__emptyString" then
+          trace ''nixvim(toLua): __emptyString is deprecated, just use an attribute named "".'' (
+            toTableKey ""
+          )
+        else
+          "[${go "" s}]";
+
+      go =
```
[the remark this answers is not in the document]
MattSturgeon: Follows the convention from nixpkgs, where private recursive closures are called "go", as in go to x value and print it with y indentation.

More specifically, this is the name used in `lib.generators.toPretty`.

We can name it anything we like, but I figure this was both clear and terse.

Could probably benefit from a comment though.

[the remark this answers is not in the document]
GaetanLepage: Ok, I had no idea about this. Thanks for the clarification.
No need to add a comment, don't worry.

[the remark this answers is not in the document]
GaetanLepage: I mean, no need to add a comment about why it is named `go`.
However, having a short comment describing what the function does could be helpful.

lib/to-lua.nix
MattSturgeon: I've done some experimenting in the nix and lua repls, it seems toJSON is indeed better and lua seems to understand the exponetised floats, although I can't find any reference to that in the [5.1 ref manual](https://www.lua.org/manual/5.1/manual.html).

```
nix-repl> builtins.toJSON 400034582305840239.3
"4.0003458230584026e+17"

lua> print (400034582305840239.3)
4.0003458230584e+17

lua> print (4.0003458230584026e+17)
4.0003458230584e+17

nix-repl> builtins.toJSON (4 * 0.000001)
"4e-06"

lua> print (4e-06)
4e-06

nix-repl> builtins.toJSON (4 * 0.000001 + 0.01)
"0.010004"

lua> print (4e-06 + 0.01)
0.010004
```

The comment from `toPretty` seems to be referencing how _nix_ can't parse exponentiatized float literals:

```
nix-repl> builtins.toJSON (4e-06)
error: undefined variable 'e-06'
       at «string»:1:19:
            1| builtins.toJSON (4e-06)
             |                   ^
```

GaetanLepage (approved):
LGTM.

Just one question, what is the goal of having a configurable `toLuaObject` ?
Would you like to use different variants in specific places of the code base ?

MattSturgeon:
> Just one question, what is the goal of having a configurable `toLuaObject` ? Would you like to use different variants in specific places of the code base ?

The main goal is to be able to change functionality (add options) without worrying about breaking existing usage.

Basing the code on the patten used by `toPretty` also allows for indented multiline output, which may be desirable in some parts of the codebase more than others?

Many of the more specific options are really targeted at end-users and/or unforseen scenarios.

For instance, users stringifying their own attrs may prefer _not_ to remove empty elements.

GaetanLepage:
> > Just one question, what is the goal of having a configurable `toLuaObject` ? Would you like to use different variants in specific places of the code base ?
> 
> The main goal is to be able to change functionality (add options) without worrying about breaking existing usage.
> 
> Basing the code on the patten used by `toPretty` also allows for indented multiline output, which may be desirable in some parts of the codebase more than others?
> 
> Many of the more specific options are really targeted at end-users and/or unforseen scenarios.
> 
> For instance, users stringifying their own attrs may prefer _not_ to remove empty elements.

Very clear !
All good, feel free to merge ;)

andrevmatos:
I think this may not be working as intended:
started noticing due to lualine's sections now showing (some) square separators;
when looking at the generated `init.lua`, this seems to be caused by several `nil` values now being included in the generated sections, e.g.:
```lua
require("lualine").setup({
  ...
  sections = {
    lualine_c = {
      {
        "filename",
        color = nil,
        fmt = nil,
        icon = nil,
        icons_enabled = nil,
        newfile_status = true,
        padding = nil,
        path = 1,
        separator = { left = nil, right = nil },
        shorting_target = 60,
      },
    },
  }
```
Usually, `nil` values are fine, but that object `separator` is overriding the global section separators, and making them square.
Maybe `removeNullAttrValues` option is not working as expected? Or not recursively?

MattSturgeon:
@andrevmatos thanks for reporting! Do you mind opening a bug report issue so I don't forget to look into this?

GaetanLepage:
Why do we have `nil` now ? They should still be ignored no ?

andrevmatos:
@MattSturgeon #1804

PR #1760, opened 2024-06-26
lib/lua: more readable tables

MattSturgeon:
- **lib/lua: add `isKeyword` and `isIdentifier`**
- **lib/lua: only quote table keys when needed**
- **lib/lua: pad table `,` with a space**

## Identifiers
Valid identifiers as [defined in the lua manual](https://www.lua.org/manual/5.1/manual.html#2.1) do not need to be quoted in lua tables, so we can make most tables less verbose and more readable by checking if the key is a valid identifier.

## Separators
The table separators are more readable when padded with whitespace.

## Future work
I think toLuaObject could be improved by taking some inspiration from generators.toPretty.

In particular, we could add optional support for multiline output and the empty attr values filter could be more efficient by not calling toLuaObject recursively in its condition.

PR #1762, opened 2024-06-26
lib/lua: workaround `builtins.match` alias issue

MattSturgeon:
A user has reported (https://github.com/nix-community/nixvim/pull/1760#issuecomment-2192066615) that `with lib` isn't bringing `match` into scope on in their flake. Odd, but we can fix by explicitly using `builtins.match`

### DIFF
--- a/lib/to-lua.nix
+++ b/lib/to-lua.nix
@@ -30,7 +30,7 @@ rec {
 
   # Valid lua identifiers are not reserved keywords, do not start with a digit,
   # and contain only letters, digits, and underscores.
-  isIdentifier = s: !(isKeyword s) && (match "[A-Za-z_][0-9A-Za-z_]*" s) == [ ];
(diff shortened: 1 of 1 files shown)


traxys (approved):

khaneliman:
I feel like an idiot... I still had my nixpkgs.url set to a fork from something I was doing a week ago to resolve a package build issue. I changed it back to unstable and it was fine...i only noticed it when i was adding the new codeium module and it was complaining about the pkg missing.

MattSturgeon:
> I changed it back to unstable and it was fine...

Thanks for confirming 😀 

I'll close this then

MattSturgeon:
I'm re-opening this because we've had additional reports _and_ I can reproduce the issue with my own config.

I don't understand why, but this PR does fix the issue.

khaneliman:
> I'm re-opening this because we've had additional reports _and_ I can reproduce the issue with my own config.
> 
> I don't understand why, but this PR does fix the issue.

It was confusing to me why it would have just magically went away with a different nixpkgs url too… lol

PR #1764, opened 2024-06-27
Adding new example nixvim configuration

JMartJonesy:
Adding new configuration JMartJonesy/kickstart.nixvim

### DIFF
--- a/docs/user-guide/config-examples.md
+++ b/docs/user-guide/config-examples.md
@@ -19,6 +19,7 @@ all of the nixvim options are accessible no matter how you are using it (flake, 
 | [GaetanLepage/nix-config] | Home-manager |
 | [gwg313/nvim-nix] | |
 | [hbjydev/hvim] | |
+| [JMartJonesy/kickstart.nixvim] | An implementation of kickstart.nvim using nixvim |
 | [MikaelFangel/nixvim-config] | An easy-setup configuration for NixVim, focused on straightforward customization |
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Hi ! Thanks for this.
Could you please respect the alphabetical order please ?

JMartJonesy:
Oops sure thing did not even notice the alphabetic ordering. I'll fix it

JMartJonesy:
Ok it has been updated

MattSturgeon (commented):
Thanks for adding your config.

My only concern is that the name will be confusing when [GaetanLepage/kickstart.nixvim](https://github.com/GaetanLepage/kickstart.nixvim) is also ready to be included.

On the other hand, I do value having multiple interpretations of kickstart.nvim in nix. There's also [kickstart-nix.nvim](https://github.com/nix-community/kickstart-nix.nvim), which doesn't use nixvim.

Adding "an" to the description helps a little, but I'm still unsure. I think @GaetanLepage should have the final say.

GaetanLepage:
> As per my earlier comment [#1764 (review)](https://github.com/nix-community/nixvim/pull/1764#pullrequestreview-2144886099), I think @GaetanLepage should have the final say due to the name conflict.

Thanks for pointing this out.
I thought about it.
I will have a look at this project later and maybe use it in my kickstart.
I don't mind having this one listed here anyway.

PR #1768, opened 2024-06-27
plugins/otter: init

PerchunPak:
https://github.com/jmbuhr/otter.nvim

Unfortunately, cannot run tests for this. After 33 minutes, it exits with SIGSTOP with no error message (both times I tried)

### DIFF
--- a/plugins/languages/otter.nix
+++ b/plugins/languages/otter.nix
@@ -1,0 +1,86 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "otter";
+  originalName = "otter.nvim";
+  defaultPackage = pkgs.vimPlugins.otter-nvim;
+
+  maintainers = [ lib.maintainers.perchun ];
+
+  settingsOptions = {
+    lsp = {
+      hover = {
+        border = helpers.defaultNullOpts.mkListOf lib.types.str [
+          "╭"
+          "─"
+          "╮"
+          "│"
+          "╯"
+          "─"
+          "╰"
+          "│"
+        ] "";
+      };
+
+      diagnostic_update_events = helpers.defaultNullOpts.mkListOf' {
+        type = lib.types.str;
+        pluginDefault = [ "BufWritePost" ];
+        description = ''
+          `:h events` that cause the diagnostics to update.
+
+          See example for less performant but more instant diagnostic updates.
+        '';
+        example = [
+          "BufWritePost"
+          "InsertLeave"
+          "TextChanged"
+        ];
+      };
+    };
+
+    buffers = {
+      set_filetype = helpers.defaultNullOpts.mkBool false ''
+        If set to true, the filetype of the otterbuffers will be set.
+        Otherwise only the autocommand of lspconfig that attaches
+        the language server will be executed without setting the filetype.
+      '';
+
+      write_to_disk = helpers.defaultNullOpts.mkBool false ''
+        Write `<path>.otter.<embedded language extension>` files
+        to disk on save of main buffer.
+        Useful for some linters that require actual files,
+        otter files are deleted on quit or main buffer close.
+      '';
+    };
+
+    strip_wrapping_quote_characters = helpers.defaultNullOpts.mkListOf lib.types.str [
+      "'"
+      "\""
+      "\`"
+    ] "";
+
+    handle_leading_whitespace = helpers.defaultNullOpts.mkBool false ''
+      Otter may not work the way you expect when entire code blocks are indented (eg. in Org files).
+      When true, otter handles these cases fully. This is a (minor) performance hit.
+    '';
+  };
+
+  extraOptions = {
+    addCmpSources = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Automatically add otter to cmp sources, as it is required for some functionality.
+      '';
+    };
+  };
+
+  extraConfig = cfg: {
+    plugins.cmp.settings.sources = lib.mkIf cfg.addCmpSources [ { name = "otter"; } ];
+  };
+}
--- a/tests/test-sources/plugins/languages/otter.nix
+++ b/tests/test-sources/plugins/languages/otter.nix
@@ -1,0 +1,39 @@
+{
+  empty = {
+    plugins.otter.enable = true;
+  };
+
+  defaults = {
+    plugins.otter = {
+      enable = true;
+
+      settings = {
+        lsp = {
+          hover = {
+            border = [
+              "╭"
+              "─"
+              "╮"
+              "│"
+              "╯"
+              "─"
+              "╰"
+              "│"
+            ];
+          };
+          diagnostic_update_events = [ "BufWritePost" ];
+        };
+        buffers = {
+          set_filetype = false;
+          write_to_disk = false;
+        };
+        strip_wrapping_quote_characters = [
+          "'"
+          "\""
+          "\`"
+        ];
+        handle_leading_whitespace = false;
+      };
+    };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -71,6 +71,7 @@
     ./languages/nix.nix
     ./languages/nvim-jdtls.nix
     ./languages/openscad.nix
+    ./languages/otter.nix
     ./languages/parinfer-rust.nix
     ./languages/plantuml-syntax.nix
     ./languages/python/jupytext.nix


plugins/languages/otter.nix
```diff
@@ -0,0 +1,58 @@
...
+  name = "otter";
+  originalName = "otter.nvim";
+  defaultPackage = pkgs.vimPlugins.otter-nvim;
+
+  maintainers = [ lib.maintainers.perchun ];
+
+  settingsOptions = {
+    lsp = {
+      hover = {
+        border = helpers.defaultNullOpts.mkListOf lib.types.str [
```
PerchunPak, line 18: I am unsure that it is possible to set type to `list of N elements`

plugins/languages/otter.nix
```diff
@@ -0,0 +1,58 @@
...
+          "│"
+          "╯"
+          "─"
+          "╰"
+          "│"
+        ] "";
+      };
+      diagnostic_update_events = helpers.defaultNullOpts.mkListOf lib.types.str [ "BufWritePost" ] ''
+        `:h events` that cause the diagnostics to update. Set to:
+        [ "BufWritePost" "InsertLeave" "TextChanged" ] for less performant
```
MattSturgeon, line 31:
```suggestion
        `[ "BufWritePost" "InsertLeave" "TextChanged" ]` for less performant
```

This could also be done as an option example:

```nix
diagnostic_update_events = helpers.defaultNullOpts.mkListOf' {
  type = lib.types.str;
  pluginDefault = [ "BufWritePost" ];
  description = ''
    `:h events` that cause the diagnostics to update.

    See example for less performant but more instant diagnostic updates.
  '';
  example = [ "BufWritePost" "InsertLeave" "TextChanged" ];
};
```

plugins/languages/otter.nix
```diff
@@ -0,0 +1,58 @@
...
+        the language server will be executed without setting the filetype
+      '';
+      write_to_disk = helpers.defaultNullOpts.mkBool false ''
+        write <path>.otter.<embedded language extension> files
+        to disk on save of main buffer.
+        useful for some linters that require actual files
+        otter files are deleted on quit or main buffer close
+      '';
+    };
+    strip_wrapping_quote_characters = helpers.defaultNullOpts.mkListOf lib.types.str [
```
MattSturgeon, line 48:
```suggestion
    };

    strip_wrapping_quote_characters = helpers.defaultNullOpts.mkListOf lib.types.str [
```

Generally, we prefer a blank line between each option, unless it's (e.g.) a small submodule with so few options that it's more readable without.

Not a big deal, I wouldn't have mentioned it if I wasn't already asking for changes :smile:

tests/test-sources/plugins/languages/otter.nix
MattSturgeon, line 1: Thanks for adding a test!

You're welcome to add additional tests, such as examples, but this is more than enough.

plugins/languages/otter.nix
```diff
@@ -0,0 +1,58 @@
...
+      '';
+    };
+    buffers = {
+      set_filetype = helpers.defaultNullOpts.mkBool false ''
+        if set to true, the filetype of the otterbuffers will be set.
+        otherwise only the autocommand of lspconfig that attaches
+        the language server will be executed without setting the filetype
+      '';
+      write_to_disk = helpers.defaultNullOpts.mkBool false ''
+        write <path>.otter.<embedded language extension> files
```
MattSturgeon, line 42:
```suggestion
        write `<path>.otter.<embedded language extension>` files
```

plugins/languages/otter.nix
```diff
@@ -0,0 +1,58 @@
...
+        to disk on save of main buffer.
+        useful for some linters that require actual files
+        otter files are deleted on quit or main buffer close
```
MattSturgeon, line 45:
```suggestion
        Useful for some linters that require actual files
        otter files are deleted on quit or main buffer close.
```

Generally, try to capitalize & punctuate correctly (I know this can be painful when copying upstream docs)

MattSturgeon (changes requested):
This is an excellent first PR. Thanks!

I feel like I'm nit-picking, but I've left some minor comments.

Most of them are picking out examples of things that can be applied to the whole diff.

plugins/languages/otter.nix
PerchunPak: Your example is much better than upstream docs, thanks

tests/test-sources/plugins/languages/otter.nix
PerchunPak: I don't really know what else to add, tbh I didn't even had an opportunity to use the plugin yet

plugins/languages/otter.nix
MattSturgeon: I think we'd need to make a custom type for that, using [`mkOptionType`](https://github.com/NixOS/nixpkgs/blob/69bee9866a4e2708b3153fdb61c1425e7857d6b8/lib/types.nix#L115-L195).

I don't think we go to this extent for other options like this though. IMO `listOf str` is fine.

MattSturgeon: We'd need to do something similar to [`nonEmptyListOf`](https://github.com/NixOS/nixpkgs/blob/69bee9866a4e2708b3153fdb61c1425e7857d6b8/lib/types.nix#L570-L576), i.e. a custom `check` function that checks the length.

Again, I don't think this is needed. It is something that _could_ be tackled in another PR targeting any other options that'd also benefit.

MattSturgeon (approved):
LGTM! Thanks for this.

I'll leave this open a while in case other maintainers want to take a look before merging.

If you're able to rebase, we can do a fast-forward merge with _your_ commit rather than our rebase.

plugins/languages/otter.nix
PerchunPak: I am pretty sure it is not the only one option for border characters in NixVim, and so it would be useful tree-wide. So I will leave this for someone else in another PR.

PerchunPak:
> If you're able to rebase, we can do a fast-forward merge with your commit rather than our rebase.

It doesn't really matter

PerchunPak:
Just noticed that I forgot to add otter as cmp source, will this work?

```nix
config = lib.mkIf cfg.enable {
  plugins.cmp = {
    settings.sources = [ { name = "otter"; } ];
  };
}
```

MattSturgeon:
> Just noticed that I forgot to add otter as cmp source, will this work?

We normally do it the other way around, by listing the source in [this file](https://github.com/nix-community/nixvim/blob/main/plugins/completion/cmp/sources.nix). This makes the [`plugins.cmp.autoEnableSources`](https://nix-community.github.io/nixvim/plugins/cmp/index.html#pluginscmpautoenablesources) aware of the plugin.

Is otter.nvim more than just a cmp source? Otherwise we'd probably define it in https://github.com/nix-community/nixvim/tree/main/plugins/completion/cmp/sources instead of directly in `plugins`.

I'm not 100% familiar with how we do nvim-cmp, so probably best if @GaetanLepage chimes in here.

GaetanLepage:
> We normally do it the other way around, by listing the source in [this file](https://github.com/nix-community/nixvim/blob/main/plugins/completion/cmp/sources.nix). This makes the [`plugins.cmp.autoEnableSources`](https://nix-community.github.io/nixvim/plugins/cmp/index.html#pluginscmpautoenablesources) aware of the plugin.
> 
> Is otter.nvim more than just a cmp source? Otherwise we'd probably define it in https://github.com/nix-community/nixvim/tree/main/plugins/completion/cmp/sources instead of directly in `plugins`.
> 
> I'm not 100% familiar with how we do nvim-cmp, so probably best if @GaetanLepage chimes in here.

Yes, no matter what, just add it to the source-plugin attrs.

MattSturgeon:
> error: The option `plugins.otter.enable' in `plugins/languages/otter.nix' is already declared in `plugins/completion/cmp/sources'.

I guess that explains why the other cmp sources in https://github.com/nix-community/nixvim/tree/main/plugins/completion/cmp/sources?rgh-link-date=2024-06-27T14%3A57%3A02Z don't directly use `mkNeovimPlugin` and instead manually use `helpers.mkSettingsOption`.

I'm afraid you'll have to take a look over those files and re-work your plugin using the same approach. Also move from `plugins/languages/` to `plugins/completion/cmp/sources` and check how we normally handle tests for cmp sources.

PerchunPak:
Had to rename the plugin to `otter-nvim`, because `plugins/completion/cmp/sources.nix` uses later that name for `pkgs.vimPlugin.otter-nvim`.

Here is the proper link for the force push: https://github.com/nix-community/nixvim/compare/7dfadb6421cbce87c6ecd3196947192efa888e70..b059fa8d6eebbc12273e23451448610b4d695a4d

I also tested the plugin on my nixos install and it works

MattSturgeon:
> Had to rename the plugin to `otter-nvim`, because `plugins/completion/cmp/sources.nix` uses later that name for `pkgs.vimPlugin.otter-nvim`.

The documentation might be a bit misleading (referring to plugins), it actually means the plugin's nixvim module name:

https://github.com/nix-community/nixvim/blob/35e837c10b767c3dbf580f3cade5084f376f7c70/plugins/completion/cmp/cmp-helpers.nix#L48-L56

MattSturgeon:
It's ok if `autoEnableSources` doesn't work for otter in this PR, so I think we should just merge 7dfadb6421cbce87c6ecd3196947192efa888e70 as is rather than attempting to integrate with the cmp systems.

Users can still manually enable `plugins.otter.enable` _and_ add `"otter"` to nvim-cmp sources.

I'll look into what refactoring can be done to make our cmp implementation more flexible, but that shouldn't block this PR.

PerchunPak:
> Apologies again for suggesting you rewrite your perfectly good PR as a cmp-plugin.

It's okay, I was happy to work on this code.

I also added an option (`plugins.otter.addCmpSources`) to automatically do
```nix
plugins.cmp = {       
  settings.sources = [ { name = "otter"; } ];
};
```
and it is enabled by default.

PerchunPak:
I also noticed that if you define a plugin using `plugins/completion/cmp/sources.nix`, it doesn't define `package` option

Compare https://nix-community.github.io/nixvim/plugins/cmp-async-path.html and https://nix-community.github.io/nixvim/plugins/easyescape.html

PerchunPak:
Just again noticed that I overlooked that otter has dependencies. How are they should be specified in NixVim?

https://github.com/jmbuhr/otter.nvim?tab=readme-ov-file#dependencies

GaetanLepage:
> Just again noticed that I overlooked that otter has dependencies. How are they should be specified in NixVim?

They should be dealt with in nixpkgs.

GaetanLepage:
https://github.com/NixOS/nixpkgs/pull/323107

plugins/languages/otter.nix
```diff
@@ -0,0 +1,90 @@
...
+    };
+  };
+
+  extraConfig =
+    cfg:
+    lib.mkIf (cfg.enable && cfg.addCmpSources) {
+      plugins.cmp = {
+        settings.sources = [ { name = "otter"; } ];
+      };
+    };
```
MattSturgeon, line 89: This can just be:

```nix
extraConfig = cfg: {
  plugins.cmp.settings.sources = lib.mkIf cfg.addCmpSources [ { name = "otter"; } ];
};
```

Since extraConfig gets wrapped in `mkIf cfg.enable` by mkNeovimPlugin.

plugins/languages/otter.nix
```diff
@@ -0,0 +1,90 @@
...
+    handle_leading_whitespace = helpers.defaultNullOpts.mkBool false ''
+      Otter may not work the way you expect when entire code blocks are indented (eg. in Org files).
+      When true, otter handles these cases fully. This is a (minor) performance hit.
+    '';
+  };
+
+  extraOptions = {
+    addCmpSources = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
```
MattSturgeon, line 76: I think normally this would be default false, and then we'd have an assertion for `cfg.addCmpSources -> config.plugins.cmp.enable`.

Although most users probably want this, we're trying to prefer explicit opt-in config over having automation and "magic". For completion sources in particular, users may want to set a priority, or add it to a different sources list.

plugins/languages/otter.nix
PerchunPak: I wrote it in that way, because though it would be more readable and if in the future something will change, you won't have a bug where otter will be added to source even when it is disabled, but sure.

plugins/languages/otter.nix
PerchunPak: So should I set it default to false? I would expect it automatically be added to cmp sources (because this is one of the main features of otter), and then if I want it to tweak, disable the option and add it manually.

MattSturgeon: I think it's fine in this instance

MattSturgeon (approved):
Thanks!

GaetanLepage:
> Thanks!

I trust you on this one, merge it if you think that it's ready :)

PR #1773, opened 2024-06-28
plugins/surround: stop using alias

SuperSandro2000:
This alias exists already in 23.11, so it should be safe to backport.

https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/applications/editors/vim/plugins/aliases.nix


GaetanLepage (approved):
Indeed, thank you !

PR #1776, opened 2024-06-29
plugins/cmp: refactor source-plugin name association

MattSturgeon:
- **plugins/cmp: fix secondary settings descriptions**
- **plugins/cmp: refactor source->plugin association**
- **plugins/cmp: refactor sources list & mkCmpSourcePlugin**
- **plugins/otter: switch to nvim-cmp association**

## Motivation

Currently, one list (well, attrset) is used to define both _source->plugin association_ __and__ _declaring cmp plugins to be created_.

This means we cannot add a nvim-cmp source-name association for _other_ plugins, not created via `mkCmpSourcePlugin`.

The general plugin creation code is also relatively dated compared to other parts of nixvim.

## Strategy

Continue to list cmp plugins to be created using `mkCmpSourcePlugin` in a nix file, but move source-name association to an (internal) nix option; `cmpSourcePlugins`.

## mkCmpSourcePlugin refactor

Inspired by https://github.com/nix-community/nixvim/blob/main/plugins/lsp/language-servers/default.nix

Changed `sources.nix` to a _list of_ `mkCmpSourcePlugin` arguments, allowing for things like `extraConfig` or other `mkVimPlugin` arguments to be defined directly in `sources.nix`.

Although, TBH more complex plugins could ditch using `mkCmpSourcePlugin` and just define the association themseves (using `{ cmpSourcePlugins.${sourceName} = name; }`), so maybe this flexibility is unnecessary?

## plugins.otter

As an example of how this can be used, I've transitioned otter to use a nvim-cmp source->plugin assiciation, removing its `addCmpSources` option.

I'm not 100% convinced this is better, because unlike most cmp plugins, otter feels like something that would normally be configured independently, via `plugins.otter.*` 🤔


plugins/completion/cmp/sources/_mk-cmp-plugin.nix
```diff
@@ -0,0 +1,32 @@
...
+  # Could be set to mkNeovimPlugin
+  pluginFn ? helpers.vim-plugin.mkVimPlugin,
+  ...
+}@args:
+pluginFn config (
+  builtins.removeAttrs args [
+    "package"
+    "pluginFn"
+    "sourceName"
+  ]
```
MattSturgeon, line 23: TBH, I think being able to do more "complex" plugin definitions in sources.nix is probably overkill.

I think the most we'd need there is `package` and `maintainers`?

Anything more complex could be moved _out of_ sources.nix and declare its cmp-source association itself.

Though this would mean we'd need to rethink how the "all sources" _test_ works...

MattSturgeon:
One thing I haven't considered in this PR is how declaring source associations this way affects the all-sources test.

Currently we assume all sources are listed in sources.nix, but the main point of this PR is to _not_ do that and instead allow sources to be defined anywhere.

tests/test-sources/plugins/completion/cmp-all-sources.nix
```diff
@@ -21,7 +21,7 @@
 
             filterFunc = sourceName: !(elem sourceName disabledSources);
 
-            sourceNames = filter filterFunc (attrNames cmp-sources);
+            sourceNames = filter filterFunc (map (getAttr "name") cmp-sources);
```
MattSturgeon, line 24: Would it be ok to just do:

```nix
additionalSources = [
  "otter"
  "some-other-source"
  "something-else"
  "not-in-sources.nix"
];

sourceNames = 
  filter filterFunc (map (getAttr "name") cmp-sources)
  ++ additionalSources;
```

This is similar to what we do in other tests where we don't have one definitive list to lookup.

MattSturgeon: I can come up with a better solution if I tweak the test system to allow test-cases to be fully blown modules rather than only allowing plain-attrs configs.

tests/fetch-tests.nix
```diff
@@ -57,7 +57,10 @@ let
             inherit pkgs lib helpers;
             config = { };
           };
-          cmp-sources = import ../plugins/completion/cmp/sources/sources.nix;
+          cmp-sources = import ../plugins/completion/cmp/sources/sources.nix {
+            inherit pkgs lib helpers;
+            config = { };
```
GaetanLepage, line 62: Why do we need those ?
It is not obvious from the content of `sources/sources.nix`.

plugins/completion/cmp/sources/sources.nix
```diff
@@ -1,49 +1,185 @@
...
-  ultisnips = "cmp-nvim-ultisnips";
-  vim_lsp = "cmp-vim-lsp";
-  vimwiki-tags = "cmp-vimwiki-tags";
-  vsnip = "cmp-vsnip";
-  yanky = "cmp_yanky";
-  zsh = "cmp-zsh";
-}
+[
+  {
+    name = "cmp-async-path";
```
GaetanLepage, line 6: Definitely `pluginName`. At least I feel it would severely enhance clarity.

tests/test-sources/plugins/completion/cmp-all-sources.nix
GaetanLepage: For me, it's fine to only test those extra sources in their respective test file. Because, they will most likely have one as they are "standalone" plugins.

tests/fetch-tests.nix
MattSturgeon: The idea was that entries within `sources.nix` would be able to declare (e.g.) `settingsOptions` and `extraConfig` inline, so would need access to `pkgs`, `lib`, & `helpers`.

The more I think about it the more I lean towards moving any cmp plugin that needs to have extra options _out of_ `sources.nix` and into its own file. If we take that approach, this change is unnecessary.

tests/test-sources/plugins/completion/cmp-all-sources.nix
MattSturgeon: #1781 will allow me to do the same thing in the test as we do in `auto-enable.nix`, i.e. something like `attrNames config.cmpSourcePlugins`. No manual listing required.

GaetanLepage: Okay, then why not !

tests/fetch-tests.nix
GaetanLepage: Yes, got you.
Anyway, as it is not used now, let's adopt the minimal approach and start from there.

plugins/completion/cmp/sources/_mk-cmp-plugin.nix
MattSturgeon: Although not currently in use, I've kept passing in the module args, as even basic stuff like setting a different maintainer or package would need that.

plugins/completion/cmp/sources/_mk-cmp-plugin.nix
```diff
@@ -0,0 +1,31 @@
...
+{
+  name ? pluginName,
+  pluginName,
+  sourceName,
+  defaultPackage ? pkgs.vimPlugins.${pluginName},
+  maintainers ? [ lib.maintainers.GaetanLepage ],
+  imports ? [ ],
+  ...
+}@args:
+helpers.vim-plugin.mkVimPlugin config (
```
MattSturgeon, line 17: Not sure if this is best using the upcoming `mkPlugin`, or perhaps using `mkNeovimPlugin` but with `callSetup ? false` and `hasSettings ? false` as defaults

plugins/completion/cmp/sources/_mk-cmp-plugin.nix
MattSturgeon: Actually, just moving the list into `sources/default.nix` makes more sense IMO. This is what we to for language-servers, and since we don't need to read the sources list in the test anymore it simplifies things IMO.

plugins/completion/cmp/auto-enable.nix
```diff
@@ -0,0 +1,92 @@
...
+        value.enable = mkIf (elem source enabledSources) true;
+      }) config.cmpSourcePlugins;
+    }
+
+    # FIXME: nesting this in `mkIf autoEnableSources` feels correct, but is different to what we did before:
+    {
+      plugins.lsp.capabilities = mkIf (elem "nvim_lsp" enabledSources) ''
+        capabilities = vim.tbl_deep_extend("force", capabilities, require('cmp_nvim_lsp').default_capabilities())
+      '';
+    }
```
MattSturgeon, line 90: @GaetanLepage the FIXME comment is really meant for the PR review. If we agree it isn't an issue, let's remove the comment.

MattSturgeon: Before LSP capabilities were always added even when the user set `autoEnableSources = false`. I think they should either depend on `autoEnableSources` or have their own toggle option?

plugins/completion/cmp/sources/sources.nix
MattSturgeon: I've done this, but I'm not convinced it is actually better. In the list it makes sense, but in `mkCmpSourcePlugin` it does not.

Remember `sources` is supposed to be a list of `mkCmpSourcePlugin` arguments, and other than a few special args, they get passed through to `mkVimPlugin`.

plugins/completion/cmp/sources/_mk-cmp-plugin.nix
GaetanLepage: I propose to revisit this once the refactor is done no ?

plugins/completion/cmp/auto-enable.nix
GaetanLepage: IMO, we should indeed put it behind a `mkIf`.

plugins/completion/cmp/sources/_mk-cmp-plugin.nix
MattSturgeon: Sure

plugins/completion/cmp/auto-enable.nix
MattSturgeon: Is `autoEnableSources` suitable or should it have its own option?

GaetanLepage: think `autoEnableSources` is fine.

MattSturgeon: Done

GaetanLepage (approved):

PR #1781, opened 2024-06-29
tests/test-derivation: allow tests to be modules

MattSturgeon:
## tests/test-derivation: allow tests to be modules
This ended up being a bigger cleanup/refactor than expected, as I encountered a lot of cruft and redundancy.

The core motivation of these changes is to allow for more complex tests, that need access to module args such as `options` or `config`. In particular #1776.

I've removed a few functions, which may look scary at first, however I've checked for any places where they may have been exposed publicly and they appear to be 100% internal.

## flake/legacyPackages: simplify `makeNixvim`, making it more powerful
While working in this area I was question why `makeNixvim` is deliberately restrictive in only accepting a config attrset.
I guess this made sense when we had a name conflict with `options`, but now that that's resolved I see no reason why _any_ module form shouldn't be accepted and passed directly through to `makeNixvimWithModule`.

Incidentally, this makes me want to find a better name for `makeNixvimWithModule`, maybe `makeNixvim'`? Regardless, out of scope for this PR.

### DIFF
--- a/docs/user-guide/install.md
+++ b/docs/user-guide/install.md
@@ -75,9 +75,10 @@ If using the home-manager module, see [Home Manager Usage](./modules/hm.md) for 
 ### Standalone usage
 
 When using nixvim as a standalone derivation you can use the following functions, located in `<nixvim>.legacyPackages.${system}`:
-- `makeNixvim`: This function takes an attribute set of options values as arguments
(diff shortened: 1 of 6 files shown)


GaetanLepage (approved):
LGTM

MattSturgeon:
@mergifyio queue

Issue #1456, opened 2024-04-24
[Feature Request] pylsp option to change python version


purepani:
Currently there is no easy way to change the version python used to get pylsp. It seems to use pkgs.python3Packages by default, but changing the package from the default packages fails since there are a bunch of overrides done to the package by default in order to make the plugin work. 

I'm hoping for something like this to work :
```
plugins.lsp.pylsp.package = pkgs.python312Packages.python-language-server;
```

At the moment, not being able to easily do this means that python 3.12 language features(like generic types in class and function definitions) won't be detected, regardless of whether you change the mypy python executable or not (at least I wasn't able to get it working).

Currently these are my settings for pylsp:
```	  
    plugins.lsp.servers.pylsp = {
	  	enable = true;
		package = pkgs.python312Packages.python-lsp-server; 
		settings = {
			plugins = {
				#black.enabled=true;
				#flake8.enabled=true;
				isort.enabled=true;
				jedi_completion.enabled=true;
				jedi_hover.enabled=true;
				jedi_references.enabled=true;
				#jedi_signature_help=true;
				#jedi_symbols=true;
				pylsp_mypy = {
					enabled=true;
					report_progress=true;
					overrides = ["--python-executable" (helpers.mkRaw "py_path") true] ;
				};
				
				rope.enabled=true;
				rope_autoimport.enabled=true;
				rope_completion.enabled=true;
				ruff.enabled=true;
			};
		};
	  };
    extraConfigLuaPre = ''
	local venv_path = os.getenv('VIRTUAL_ENV') -- or vim.fn.exepath('python')
	local path_python = vim.fn.exepath('python')
	local py_path = nil
	if venv_path ~= nil then
	  py_path = venv_path .. "/bin/python3"
	elseif path_python ~=nil then
	  py_path = path_python
	else
	  py_path = vim.g.python3_host_prog
	end
	print(py_path)
	'';


```
The code in extraConfigLuaPre does correctly change the python executable for mypy(when the path changes in a dev shell), but despite this, it won't work for 3.12 features due to originating from the 3.11 package set(as far as I can tell).

PR #1785, opened 2024-07-01
plugins/lsp/pylsp: Add python option to generate obtain the pylsp from.

purepani:
Currently, pkgs.python3Packges is used by default and is unable to be changed. Without this, using features from newer python versions in mypy(and maybe other plugins) is impossible since mypy uses it's own python interpreter that it's installed under for getting typing hints afaik.
Fixes https://github.com/nix-community/nixvim/issues/1456

### DIFF
--- a/plugins/lsp/language-servers/pylsp.nix
+++ b/plugins/lsp/language-servers/pylsp.nix
@@ -10,485 +10,492 @@ let
   cfg = config.plugins.lsp.servers.pylsp;
 in
 {
-  # All settings are documented here:
-  # https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md
-
-  options.plugins.lsp.servers.pylsp.settings = {
-    configurationSources = mkOption {
-      type = lib.types.nullOr (
-        types.enum [
-          "pycodestyle"
-          "flake8"
-        ]
-      );
-      description = "List of configuration sources to use.";
-      default = null;
-      apply = value: if (value != null) then [ value ] else null;
+  options.plugins.lsp.servers.pylsp = {
+    pythonPackage = mkOption {
+      type = lib.types.package;
+      description = "Python package to use for lsp.";
+      default = pkgs.python3;
     };
 
-    plugins = {
-      autopep8 = {
-        enabled = helpers.defaultNullOpts.mkBool true ''
-          Enable or disable the plugin.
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/pylsp.nix
```diff
@@ -12,7 +12,11 @@ in
 {
   # All settings are documented here:
   # https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md
-
+  options.plugins.lsp.servers.pylsp.python = mkOption {
```
GaetanLepage, line 15:
```suggestion
  options.plugins.lsp.servers.pylsp.pythonPackage = mkOption {
```
This is a clearer name IMO. cc @MattSturgeon

GaetanLepage (changes requested):
This is very useful indeed !

plugins/lsp/language-servers/pylsp.nix
MattSturgeon: Agreed. Defining `options.plugins.lsp.servers.pylsp = {` once would also be an improvement imo, but that's a nit.

purepani: Not sure how to make the diff be smaller, but both of these are implemented.

plugins/lsp/language-servers/pylsp.nix
```diff
@@ -12,483 +12,490 @@ in
 {
   # All settings are documented here:
   # https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md
```
MattSturgeon, line 14: Can you move these comments down, to above the `settings = {` line?

plugins/lsp/language-servers/pylsp.nix
MattSturgeon: > Not sure how to make the diff be smaller

Can't be helped due to the auto-formatting.

You can view the diff with whitespace ignored though: https://github.com/nix-community/nixvim/pull/1785/files?w=1

plugins/lsp/language-servers/pylsp.nix
purepani: Done!

MattSturgeon (approved):
LGTM, thanks!

GaetanLepage:
@Mergifyio queue

PR #1792, opened 2024-07-01
standalone: rename `nixvimExtend` to `extend`

MattSturgeon:
This is our scope, so there's no need to be explicit.

This also follows the precedent set by `lib.extend`, although that takes an overlay function.

The old name still exists, but it will print the following when used:

> trace: warning: <nixvim>.nixvimExtend has been renamed to <nixvim>.extend


MattSturgeon:
@mergifyio queue

PR #1799, opened 2024-07-02
wrappers: simplify modules

MattSturgeon:
I'm going through the wrapper + modules trying to clean things up a bit. This is some of the first low-hanging fruit.

### TL;DR
Moved all inline wrapper-specific modules to their own file in `wrappers/modules/`

IMO the consistency makes things much clearer.


MattSturgeon:
@mergifyio queue

PR #1800, opened 2024-07-02
modules: refactor module bootstrapping

MattSturgeon:
## Summary

Continuing to try and simplify things, I've refactored the way we bootstrap the nixvim modules to be passed to `lib.evalModules`.

First, ensure all modules are defined in `./plugins`, `./modules` and `./wrappers/modules`.
I.e. none in `./flake-modules`.

Then, instead of [currying `pkgs` into a bootstrapping module](https://github.com/nix-community/nixvim/blob/3d969603481c745f8faa411f1e8b7c97517c67a3/flake-modules/modules.nix#L5-L19), we can require `defaultPkgs` be provided as a special arg.

This makes a few things easier to reason about, as all modules are now just module files, rather than being functions that return modules.

### Evaluated module flake output

It'd be nice to have an empty/default module evaluation as a flake output. This should make it easier to debug new options (etc) as you can inspect the evaluated module's `config` and `options` in the nix repl.

```nix
perSystem = {pkgs, lib, ...}: {
  packages.evaluatedModule = lib.evalModules {
    modules = [ ../modules ];
    specialArgs = {
      inherit helpers;
      defaultPackages = pkgs;
    };
  };
}
```

### Modules bootstrap their own `helpers`
IDK if this is possible, because we might need helpers to be a specialArg to use it to create `imports`, but it'd be nice if the modules could bootstrap their own `helpers` lib.

Alternatively, if we find we rely on specialArgs, maybe have a function that returns a standardized `specialArgs` attrs:

```nix
# modules/_special-args.nix
defaultPkgs: {
  inherit defaultPkgs;
  inherit (defaultPkgs) lib;
  helpers = inherit (import ../lib {pkgs = defaultPkgs;}) helpers;
}
```

### Move more stuff to `modules/`

I suspect some things in `wrappers/modules/` can be moved to the main `modules/` tree? Or is it there because we _don't_ want to include it in `files` submodules?

### DIFF
--- a/modules/misc/nixpkgs.nix
+++ b/modules/misc/nixpkgs.nix
@@ -1,24 +1,16 @@
-{
-  defaultPkgs,
-  pkgs,
-  lib,
-  ...
-}:
+{ defaultPkgs, lib, ... }:
 let
   # TODO: https://github.com/nix-community/nixvim/issues/1784
   finalPackage = defaultPkgs;
 in
 {
   config = {
-    _module.args = {
-      # We explicitly set the default override priority, so that we do not need
-      # to evaluate finalPkgs in case an override is placed on `_module.args.pkgs`.
-      # After all, to determine a definition priority, we need to evaluate `._type`,
-      # which is somewhat costly for Nixpkgs. With an explicit priority, we only
-      # evaluate the wrapper to find out that the priority is lower, and then we
-      # don't need to evaluate `finalPkgs`.
-      pkgs = lib.mkOverride lib.modules.defaultOverridePriority finalPackage;
-      inherit (pkgs) lib;
-    };
+    # We explicitly set the default override priority, so that we do not need
+    # to evaluate finalPkgs in case an override is placed on `_module.args.pkgs`.
+    # After all, to determine a definition priority, we need to evaluate `._type`,
+    # which is somewhat costly for Nixpkgs. With an explicit priority, we only
+    # evaluate the wrapper to find out that the priority is lower, and then we
+    # don't need to evaluate `finalPkgs`.
+    _module.args.pkgs = lib.mkOverride lib.modules.defaultOverridePriority finalPackage;
   };
 }


modules/misc/meta.nix
```diff
@@ -0,0 +1,54 @@
+{ defaultPkgs, lib, ... }:
+with lib;
+{
+  imports = [
+    # TODO: to have modulesPath we'd have to define it using specialArgs
+    # (Can't use config._module.args to define imports)
+    (defaultPkgs.path + "/nixos/modules/misc/meta.nix")
```
MattSturgeon, line 7: Doing this means `defaultPkgs` must be evaluated even if `_module.args.pkgs` is overridden, but I don't see an alternative.

It's no different to what we do currently though, so finding a solution doesn't block this PR.

MattSturgeon:
Because `evalModules` always sets `specialArgs.lib`, if we added a `nixosModulesPath` special arg we would be able to make `defaultPkgs` an option instead of a special arg.

Ideally we wouldn't have any special args, to make the modules as portable as possible, but if we decide to extend `lib` (future work) we'll need to do that via `specialArgs` anyway.

The only way I can think of to avoid needing `nixosModulesPath` is if we use a `fetch` function to download the NixOS modules we need as input sources.

GaetanLepage:
> I suspect some things in `wrappers/modules/` can be moved to the main `modules/` tree? Or is it there because we _don't_ want to include it in `files` submodules?

Not sure about this.
This stuff is actually quite old and the motivation for those choices might not be completely valid.

MattSturgeon:
Looks like the way home-manager handles this is to set the `pkgsPath` special arg, and use that to construct `pkgs`: `defaultPkgs = import pkgsPath config.nixpkgs`.

Meaning they use a fresh instance, with its own config/overlays, but matching the same locked channel.

Unless these nixos module options are used:

`home-manager.useGlobalPkgs`

> Whether to enable using the system configuration’s `pkgs` argument in Home Manager. This disables the Home Manager options `nixpkgs.*`.

`home-manager.useUserPackages`

> Whether to enable installation of user packages through the `users.users.<name>.packages` option.

I believe when home-manager is used standalone, it sets `pkgsPath` to `<nixpkgs>` when not using flakes or home-manager's `inputs.nixpkgs` otherwise.

It might do the same thing when it's a NixOS module, or it might use `pkgs.path` (I'm not sure).

I think we should have some idea how we want to approach this before finalising this PR, but I don't want to block it as we can always re-work things again if needed.

MattSturgeon:
@mergifyio queue

PR #1801, opened 2024-07-02
plugins/treesitter: switch to mkNeovimPlugin

khaneliman:
Had a setting I wanted to take advantage of that wasn't supported in its current iteration... figured instead of changing that specific setting I'd help migrate it to the new schema.

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -6,210 +6,309 @@
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.treesitter;
-in
-{
-  options = {
-    plugins.treesitter = {
-      enable = mkEnableOption "tree-sitter syntax highlighting";
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "treesitter";
+  originalName = "nvim-treesitter";
+  luaName = "nvim-treesitter.configs";
+  defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
-      package = mkOption {
-        type = types.package;
-        default = pkgs.vimPlugins.nvim-treesitter;
-        description = "Plugin to use for nvim-treesitter. If using nixGrammars, it should include a `withPlugins` function";
-      };
+  maintainers = [ lib.maintainers.khaneliman ];
 
-      nixGrammars = mkOption {
-        type = types.bool;
-        default = true;
-        description = "Install grammars with Nix";
-      };
+  # TODO introduced 2024-07-06: remove after 24.11
+  optionsRenamedToSettings = [
+    "ensureInstalled"
+    "ignoreInstall"
+    "parserInstallDir"
+    [
+      "incrementalSelection"
+      "enable"
+    ]
+    [
+      "incrementalSelection"
+      "keymaps"
+      "initSelection"
+      "nodeDecremental"
+    ]
+    [
+      "incrementalSelection"
+      "keymaps"
+      "initSelection"
+      "nodeIncremental"
+    ]
+    [
+      "incrementalSelection"
+      "keymaps"
+      "initSelection"
+      "scopeIncremental"
+    ]
+  ];
 
-      ensureInstalled = mkOption {
-        type =
-          with types;
-          oneOf [
-            (enum [ "all" ])
-            (listOf str)
-          ];
-        default = "all";
-        description = "Either \"all\" or a list of languages";
-      };
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "treesitter"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    [
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "moduleConfig" ]) settingsPath)
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "customCaptures" ]) (
+        settingsPath
+        ++ [
+          "highlight"
+          "custom_captures"
+        ]
+      ))
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "disabledLanguages" ]) (
+        settingsPath
+        ++ [
+          "highlight"
+          "disable"
+        ]
+      ))
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "indent" ]) (
+        settingsPath
+        ++ [
+          "indent"
+          "enable"
+        ]
+      ))
+    ];
 
-      gccPackage = helpers.mkPackageOption {
-        default = if cfg.nixGrammars then null else pkgs.gcc;
-        description = ''
-          Which package (if any) to be added as the GCC compiler.
-          This is required to build grammars if you are not using `nixGrammars`.
-          To disable the installation of GCC, set this option to `null`.
-        '';
-      };
+  settingsOptions = {
+    auto_install = helpers.defaultNullOpts.mkBool false ''
+      Whether to automatically install missing parsers when entering a buffer.
+    '';
 
-      parserInstallDir = mkOption {
-        type = types.nullOr types.str;
-        default = if cfg.nixGrammars then null else "$XDG_DATA_HOME/nvim/treesitter";
-        description = ''
-          Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
-          This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
-        '';
-      };
+    highlight = {
+      additional_vim_regex_highlighting =
+        helpers.defaultNullOpts.mkNullableWithRaw
+          (with helpers.nixvimTypes; either bool (listOf (maybeRaw str)))
+          false
+          ''
+            Setting this to true will run `syntax` and tree-sitter at the same time. \
+            Set this to `true` if you depend on 'syntax' being enabled (e.g. for indentation). \
+            See `:h syntax`.
 
-      ignoreInstall = mkOption {
-        type = types.listOf types.str;
-        default = [ ];
-        description = "List of parsers to ignore installing (for \"all\")";
-      };
+            Using this option may slow down your editor, and you may see some duplicate highlights. \
+            Instead of true, it can also be a list of languages.
+          '';
 
-      disabledLanguages = mkOption {
-        type = types.listOf types.str;
-        default = [ ];
-        description = "A list of languages to disable";
-      };
+      enable = helpers.defaultNullOpts.mkBool false ''
+        Whether to enable treesitter highlighting.
+      '';
 
-      customCaptures = mkOption {
-        type = types.attrsOf types.str;
-        default = { };
-        description = "Custom capture group highlighting";
-      };
+      disable =
+        helpers.defaultNullOpts.mkStrLuaFnOr (with helpers.nixvimTypes; listOf (maybeRaw str)) null
+          ''
+            Can either be a list of the names of parsers you wish to disable or
+            a lua function that returns a boolean indicating the parser should be disabled.
+          '';
 
-      incrementalSelection =
+      custom_captures = helpers.defaultNullOpts.mkAttrsOf types.str { } ''
+        Custom capture group highlighting.
+      '';
+    };
+
+    incremental_selection = {
+      enable = helpers.defaultNullOpts.mkBool false ''
+        Incremental selection based on the named nodes from the grammar.
+      '';
+
+      keymaps =
         let
-          keymap =
-            default:
-            mkOption {
-              type = types.str;
-              inherit default;
-              description = "Key shortcut";
-            };
+          mkKeymap = default: helpers.defaultNullOpts.mkStr default "Key shortcut";
         in
         {
-          enable = mkEnableOption "incremental selection based on the named nodes from the grammar";
-          keymaps = {
-            initSelection = keymap "gnn";
-            nodeIncremental = keymap "grn";
-            scopeIncremental = keymap "grc";
-            nodeDecremental = keymap "grm";
-          };
+          init_selection = mkKeymap "gnn";
+          node_incremental = mkKeymap "grn";
+          scope_incremental = mkKeymap "grc";
+          node_decremental = mkKeymap "grm";
         };
+    };
 
-      indent = mkEnableOption "tree-sitter based indentation";
+    indent = {
+      enable = helpers.defaultNullOpts.mkBool false ''
+        Whether to enable treesitter indentation.
+      '';
+    };
 
-      folding = mkEnableOption "tree-sitter based folding";
+    ensure_installed = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      Either "all" or a list of languages to ensure installing.
+    '';
 
-      languageRegister = mkOption {
-        type = with types; attrsOf (either str (listOf str));
-        description = ''
-          This is a wrapping of the `vim.treesitter.language.register` function.
-          Register specific parsers to one or several filetypes.
-          The keys are the parser names and the values are either one or several filetypes.
-        '';
-        default = { };
-        example = {
-          cpp = "onelab";
-          python = [
-            "myFiletype"
-            "anotherFiletype"
-          ];
-        };
+    ignore_install = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      List of parsers to ignore installing. Used when `ensure_installed` is set to "all".
+    '';
+
+    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
+      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
+
+      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
+      Otherwise, change it to something that will work for you!
+    '';
+
+    sync_install = helpers.defaultNullOpts.mkBool false ''
+      Install parsers synchronously (only applied to `ensure_installed`).
+    '';
+  };
+
+  settingsExample = {
+    additional_vim_regex_highlighting = true;
+    ensure_installed = true;
+    ignore_install = [ "rust" ];
+    indent = true;
+    parser_install_dir = "$XDG_DATA_HOME/nvim/treesitter";
+
+    highlight = {
+      enable = true;
+      disable = [ "rust" ];
+      custom_captures = { };
+    };
+
+    incremental_selection = {
+      enable = true;
+
+      keymaps = {
+        init_selection = "gnn";
+        node_incremental = "grn";
+        scope_incremental = "grc";
+        node_decremental = "grm";
       };
-
-      grammarPackages = mkOption {
-        type = with types; listOf package;
-        default = cfg.package.passthru.allGrammars;
-        description = "Grammar packages to install";
-      };
-
-      moduleConfig = mkOption {
-        type = types.attrsOf types.anything;
-        default = { };
-        description = "This is the configuration for extra modules. It should not be used directly";
-      };
-
-      nixvimInjections = mkEnableOption "nixvim specific injections, like lua highlighting in extraConfigLua";
     };
   };
 
-  config =
-    let
-      tsOptions = {
-        highlight = {
-          inherit (cfg) enable;
-          disable = if (cfg.disabledLanguages != [ ]) then cfg.disabledLanguages else null;
+  extraOptions = {
+    folding = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = "Whether to enable treesitter folding.";
+    };
 
-          custom_captures = if (cfg.customCaptures != { }) then cfg.customCaptures else null;
-        };
+    gccPackage = helpers.mkPackageOption {
+      name = "gcc";
+      default = pkgs.gcc;
+      defaultText = literalExpression "pkgs.gcc";
+      example = literalExpression "pkgs.gcc14";
+      description = ''
+        Which package (if any) to be added as the GCC compiler.
 
-        incremental_selection =
-          if cfg.incrementalSelection.enable then
-            {
-              enable = true;
-              keymaps = {
-                init_selection = cfg.incrementalSelection.keymaps.initSelection;
-                node_incremental = cfg.incrementalSelection.keymaps.nodeIncremental;
-                scope_incremental = cfg.incrementalSelection.keymaps.scopeIncremental;
-                node_decremental = cfg.incrementalSelection.keymaps.nodeDecremental;
-              };
-            }
-          else
-            null;
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of GCC, set this option to `null`.
+      '';
+    };
 
-        indent = if cfg.indent then { enable = true; } else null;
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = config.plugins.treesitter.package.passthru.allGrammars;
+      example = literalExpression "pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars";
+      defaultText = literalExpression "config.plugins.treesitter.package.passthru.allGrammars";
+      description = "Grammar packages to install";
+    };
 
-        ensure_installed = if cfg.nixGrammars then [ ] else cfg.ensureInstalled;
-        ignore_install = cfg.ignoreInstall;
-        parser_install_dir = cfg.parserInstallDir;
-      } // cfg.moduleConfig;
-    in
-    mkIf cfg.enable {
-      extraConfigLua =
-        (optionalString (cfg.parserInstallDir != null) ''
-          vim.opt.runtimepath:append("${cfg.parserInstallDir}")
-        '')
-        + ''
-          require('nvim-treesitter.configs').setup(${helpers.toLuaObject tsOptions})
-        ''
-        + (optionalString (cfg.languageRegister != { }) ''
-          __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
+    # TODO: Implement rawLua support to be passed into extraConfigLua.
+    languageRegister = mkOption {
+      type = with types; attrsOf (coercedTo str toList (listOf str));
+      default = { };
+      example = {
+        cpp = "onelab";
+        python = [
+          "foo"
+          "bar"
+        ];
+      };
+      description = ''
+        This is a wrapping of the `vim.treesitter.language.register` function.
+
+        Register specific parsers to one or several filetypes.
+
+        The keys are the parser names and the values are either one or several filetypes.
+      '';
+    };
+
+    nixGrammars = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = "Whether to install grammars defined in `grammarPackages`.";
+    };
+
+    nixvimInjections = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = "Whether to enable Nixvim injections, e.g. highlighting `extraConfigLua` as lua.";
+    };
+
+    nodejsPackage = helpers.mkPackageOption {
+      name = "nodejs";
+      default = pkgs.nodejs;
+      defaultText = literalExpression "pkgs.nodejs";
+      example = literalExpression "pkgs.nodejs_22";
+      description = ''
+        Which package (if any) to be added as the nodejs package.
+
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of NodeJS, set this option to `null`.
+      '';
+    };
+
+    treesitterPackage = helpers.mkPackageOption {
+      name = "tree-sitter";
+      default = pkgs.tree-sitter;
+      defaultText = literalExpression "pkgs.tree-sitter";
+      description = ''
+        Which package (if any) to be added as the tree-sitter binary.
+
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of tree-sitter, set this option to `null`.
+      '';
+    };
+  };
+
+  # NOTE: We call setup manually below.
+  callSetup = false;
+  # NOTE: We install cfg.package manually so we can install grammars using it.
+  installPackage = false;
+
+  extraConfig = cfg: {
+    extraConfigLua =
+      # NOTE: Upstream state that the parser MUST be at the beginning of runtimepath.
+      # Otherwise the parsers from Neovim takes precedent, which may be incompatible with some queries.
+      (optionalString (cfg.settings.parser_install_dir != null) ''
+        vim.opt.runtimepath:prepend("${cfg.settings.parser_install_dir}")
+      '')
+      + ''
+        require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings})
+      ''
+      + (optionalString (cfg.languageRegister != { }) ''
+        do
+          local __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
 
           for parser_name, ft in pairs(__parserFiletypeMappings) do
             require('vim.treesitter.language').register(parser_name, ft)
           end
-        '');
+        end
+      '');
 
-      extraFiles = mkIf cfg.nixvimInjections {
-        "queries/nix/injections.scm".text = ''
-          ;; extends
+    extraFiles = mkIf cfg.nixvimInjections { "queries/nix/injections.scm".source = ./injections.scm; };
 
-          (binding
-            attrpath: (attrpath (identifier) @_path)
-            expression: [
-              (string_expression (string_fragment) @lua)
-              (indented_string_expression (string_fragment) @lua)
-            ]
-            (#match? @_path "^extraConfigLua(Pre|Post)?$"))
+    extraPlugins = mkIf (cfg.package != null) [
+      (mkIf cfg.nixGrammars (cfg.package.withPlugins (_: cfg.grammarPackages)))
+      (mkIf (!cfg.nixGrammars) cfg.package)
+    ];
 
-          (binding
-            attrpath: (attrpath (identifier) @_path)
-            expression: [
-              (string_expression (string_fragment) @vim)
-              (indented_string_expression (string_fragment) @vim)
-            ]
-            (#match? @_path "^extraConfigVim(Pre|Post)?$"))
-        '';
-      };
+    extraPackages = [
+      cfg.gccPackage
+      cfg.nodejsPackage
+      cfg.treesitterPackage
+    ];
 
-      extraPlugins =
-        if cfg.nixGrammars then [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ] else [ cfg.package ];
-      extraPackages = with pkgs; [
-        tree-sitter
-        nodejs
-        cfg.gccPackage
-      ];
-
-      opts = mkIf cfg.folding {
-        foldmethod = "expr";
-        foldexpr = "nvim_treesitter#foldexpr()";
-      };
-    };
+    opts = mkIf cfg.folding (mkDefault {
+      foldmethod = "expr";
+      foldexpr = "nvim_treesitter#foldexpr()";
+    });
+  };
 }
--- a/tests/test-sources/plugins/languages/treesitter/treesitter.nix
+++ b/tests/test-sources/plugins/languages/treesitter/treesitter.nix
@@ -1,18 +1,70 @@
+{ pkgs, ... }:
 {
+  default = {
+    plugins.treesitter = {
+      enable = true;
+
+      settings = {
+        auto_install = false;
+        ensure_installed = [ ];
+        ignore_install = [ ];
+        parser_install_dir = null;
+        sync_install = false;
+
+        highlight = {
+          additional_vim_regex_highlighting = false;
+          enable = false;
+          custom_captures = { };
+          disable = null;
+        };
+
+        incremental_selection = {
+          enable = false;
+          keymaps = {
+            init_selection = "gnn";
+            node_incremental = "grn";
+            scope_incremental = "grc";
+            node_decremental = "grm";
+          };
+        };
+
+        indent = {
+          enable = false;
+        };
+      };
+    };
+  };
+
   empty = {
     plugins.treesitter.enable = true;
   };
 
-  nonix = {
-    # TODO: See if we can build parsers (legacy way)
-    tests.dontRun = true;
+  empty-grammar-packages = {
     plugins.treesitter = {
       enable = true;
-      nixGrammars = false;
+
+      grammarPackages = [ ];
     };
   };
 
-  nixvimInjections = {
+  highlight-disable-function = {
+    plugins.treesitter = {
+      enable = true;
+
+      settings = {
+        highlight = {
+          enable = true;
+          disable = ''
+            function(lang, bufnr)
+              return api.nvim_buf_line_count(bufnr) > 50000
+            end
+          '';
+        };
+      };
+    };
+  };
+
+  nixvim-injections = {
     plugins.treesitter = {
       enable = true;
       nixvimInjections = true;
@@ -27,18 +79,43 @@
     };
   };
 
-  # This needs a custom input
-  # custom = {
-  #   plugins.treesitter = {
-  #     enable = true;
-  #     nixGrammars = true;
-  #     grammarPackages = [
-  #       (build-ts.lib.buildGrammar pkgs {
-  #         language = "gleam";
-  #         version = "0.25.0";
-  #         source = gleam;
-  #       })
-  #     ];
-  #   };
-  # };
+  no-nix = {
+    # TODO: See if we can build parsers (legacy way)
+    tests.dontRun = true;
+    plugins.treesitter = {
+      enable = true;
+      nixGrammars = false;
+    };
+  };
+
+  specific-grammars = {
+    plugins.treesitter = {
+      enable = true;
+
+      grammarPackages = with pkgs.vimPlugins.nvim-treesitter.builtGrammars; [
+        bash
+        git_config
+        git_rebase
+        gitattributes
+        gitcommit
+        gitignore
+        json
+        jsonc
+        lua
+        make
+        markdown
+        meson
+        ninja
+        nix
+        readline
+        regex
+        ssh-config
+        toml
+        vim
+        vimdoc
+        xml
+        yaml
+      ];
+    };
+  };
 }
--- a/plugins/languages/treesitter/injections.scm
+++ b/plugins/languages/treesitter/injections.scm
@@ -1,0 +1,17 @@
+;; extends
+
+(binding
+  attrpath: (attrpath (identifier) @_path)
+  expression: [
+    (string_expression (string_fragment) @lua)
+    (indented_string_expression (string_fragment) @lua)
+  ]
+  (#match? @_path "^extraConfigLua(Pre|Post)?$"))
+
+(binding
+  attrpath: (attrpath (identifier) @_path)
+  expression: [
+    (string_expression (string_fragment) @vim)
+    (indented_string_expression (string_fragment) @vim)
+  ]
+  (#match? @_path "^extraConfigVim(Pre|Post)?$"))
--- a/plugins/languages/treesitter/treesitter-refactor.nix
+++ b/plugins/languages/treesitter/treesitter-refactor.nix
@@ -144,7 +144,7 @@ with lib;
 
       extraPlugins = [ cfg.package ];
 
-      plugins.treesitter.moduleConfig.refactor = {
+      plugins.treesitter.settings.refactor = {
         highlight_definitions = {
           inherit (cfg.highlightDefinitions) enable disable;
           clear_on_cursor_move = cfg.highlightDefinitions.clearOnCursorMove;
--- a/plugins/languages/treesitter/treesitter-textobjects.nix
+++ b/plugins/languages/treesitter/treesitter-textobjects.nix
@@ -193,7 +193,7 @@ with lib;
 
       extraPlugins = [ cfg.package ];
 
-      plugins.treesitter.moduleConfig.textobjects =
+      plugins.treesitter.settings.textobjects =
         with cfg;
         let
           processKeymapsOpt =
--- a/plugins/utils/vim-matchup.nix
+++ b/plugins/utils/vim-matchup.nix
@@ -178,7 +178,7 @@ with lib;
     mkIf cfg.enable {
       extraPlugins = [ cfg.package ];
 
-      plugins.treesitter.moduleConfig.matchup = mkIf cfg.treesitterIntegration.enable {
+      plugins.treesitter.settings.matchup = mkIf cfg.treesitterIntegration.enable {
         inherit (cfg.treesitterIntegration) enable disable;
         disable_virtual_text = cfg.treesitterIntegration.disableVirtualText;
         include_match_words = cfg.treesitterIntegration.includeMatchWords;


plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
+      '';
+    };
 
-      moduleConfig = mkOption {
-        type = types.attrsOf types.anything;
-        default = { };
-        description = "This is the configuration for extra modules. It should not be used directly";
-      };
+    # TODO: default null nixgrammars
+    ensure_installed = helpers.defaultNullOpts.mkListOf types.str [ "all" ] ''
```
MattSturgeon, line 71: Is this an upstream treesitter option or a custom nixvim option?

`settingsOptions` is only for stuff that should be passed to the plugins `setup` function; anything specific to nixvim should go in `extraOptions` instead.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
+
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = cfg.package.passthru.allGrammars;
+      description = "Grammar packages to install";
+    };
 
-          custom_captures = if (cfg.customCaptures != { }) then cfg.customCaptures else null;
-        };
+    languageRegister = helpers.defaultNullOpts.mkAttrsOf (either str (listOf str)) { } ''
```
MattSturgeon, line 104: I think using `types.coercedTo str toList (listOf str)` would be better.

This will allow defining strings or lists, but will merge multiple definitions as lists.

Although I don't think we do this elsewhere currently, it would allow multiple definitions to be merged when one def is a str and another is a list.

This is probably something we should consider in other places where we use `either`.

plugins/languages/treesitter/treesitter.nix
khaneliman: It's upstream

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
-        foldexpr = "nvim_treesitter#foldexpr()";
-      };
+    moduleConfig = helpers.defaultNullOpts.mkAttrsOf types.anything { } ''
+      This is the configuration for extra modules. It should not be used directly.
+    '';
+
+    nixGrammars = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Install grammars with Nix";
```
MattSturgeon, line 117: You can set `internal = true` and `visible = false` on options that end-users shouldn't use.

You can also set `readOnly = true` if you want to enforce having exactly one definition, though I don't think that applies here.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
-        tree-sitter
-        nodejs
-        cfg.gccPackage
-      ];
-
-      opts = mkIf cfg.folding {
-        foldmethod = "expr";
-        foldexpr = "nvim_treesitter#foldexpr()";
-      };
+    moduleConfig = helpers.defaultNullOpts.mkAttrsOf types.anything { } ''
```
MattSturgeon, line 110: Generally, unless you specifically want to have `nullOr`, you don't usually need `defaultNullOpts` outside of `settingsOptions`.

I'm on my phone, so I can't really tell if it is needed here or not though.

plugins/languages/treesitter/treesitter.nix
khaneliman: I believe this is a user option to enable / disable the installation of the grammar packages through nix. So a user would use this option if they want nixpkgs managed grammars. This is related to a few TODOs about how we should handle that since the old derivation just changed the defaults when this option was provided but I was wondering if we wanted to add warnings/assertions to those settings if this is enabled.

plugins/languages/treesitter/treesitter.nix
khaneliman: I'm actually kinda thinking now that this might not even be needed with the new mkNeovimPlugin... I believe this was supposed to be a generic setting to merge settings in that weren't specifically called out..

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
+    nixvimInjections = mkEnableOption "nixvim specific injections, like lua highlighting in extraConfigLua";
+  };
+
+  extraConfig = cfg: {
+    extraConfigLua =
+      (optionalString (cfg.parserInstallDir != null) ''
+        vim.opt.runtimepath:append("${cfg.parserInstallDir}")
+      '')
+      + ''
+        require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings // cfg.moduleConfig})
```
MattSturgeon, line 129: This feels wrong (`settings // moduleConfig`).

It'd be better if whatever your other options need to change in the `setup` opts did so by configuring the nix settings options, e.g. with mkDefault, or mkForce.

This allows end-users to override the definitions.

Can't tell from my initial phone review what `moduleConfig` is for though, so can't offer more specific advice atm.

khaneliman: Yeah, I was just commenting on this on that other thread, too... I agree it feels weird after looking back at the derivation.  It seemingly was used to support other modules modifying treesitter from an internal option, which shouldn't be needed with the new settings attribute set.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
+          expression: [
+            (string_expression (string_fragment) @vim)
+            (indented_string_expression (string_fragment) @vim)
+          ]
+          (#match? @_path "^extraConfigVim(Pre|Post)?$"))
+      '';
+    };
+
+    extraPlugins =
+      if cfg.nixGrammars then [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ] else [ cfg.package ];
```
MattSturgeon, line 162: As a general best practice, prefer using mkMerge and mkIf in _config definitions_, unless there's a specific reason to use a "real" branch such as `if`, `optional`, etc.

This allows the module system to be a bit smarter; e.g. it'll check the definition is valid even when the condition is false

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
-
-      incrementalSelection =
+helpers.neovim-plugins.mkNeovimPlugin config {
+  name = "treesitter";
+  originalName = "nvim-treesitter";
+  defaultPackage = pkgs.vimPlugins.nvim-treesitter;
+
+  optionsRenamedToSettings = [
+
+  ];
```
MattSturgeon, line 16: You can probably copy-paste the _defaults_ test into here, then run a vim macro to format as follows:

```nix
[
  "someOption"
  [ "some" "nested" "option" ]
  "anotherTopLevelSetting"
]
```

Note, use the "old" camelCase names; mkNeovimPlugin will convert the names to snake\_case automatically.

Check the renames work with the tests _before_ updating the tests to the new option names.

khaneliman: Good suggestion, I was wondering how to best tackle that, too.

MattSturgeon: Don't be afraid to use `mkRemovedOptionModule` or `mkRenamedOptionModule` for anything that `optionsRenamedToSettings` can't handle.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Also, I think we'll need to add a makeNeovimPlugin argument that disables adding `cfg.package` to `extraPlugis`

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,169 @@
...
+
+    extraPlugins =
+      if cfg.nixGrammars then [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ] else [ cfg.package ];
+    extraPackages = with pkgs; [
+      tree-sitter
+      nodejs
+      cfg.gccPackage
+    ];
+
+    opts = mkIf cfg.folding {
```
MattSturgeon, line 169: This should probably be `mkIf cfg.folding (mkDefault {`, to allow users to more easily override if needed.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Do you think having a rename alias `moduleConfig -> settings` would work as a graceful deprecation then?

Other than the old option being nullable the types should match up, I think.

khaneliman: Makes sense, renaming instead of just removing it. I was curious how many plugins actually tied into the moduleConfig, but hadn't scoured the repo, yet.

MattSturgeon: Was thinking in case end-users were using the option, but if it was an internal option I think it's fair to use a `mkRemovedOptionModule` assertion or even just remove it with no warnings

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,237 @@
...
+    # TODO: helper?
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = cfg.package.passthru.allGrammars;
+      description = "Grammar packages to install";
+    };
 
-          custom_captures = if (cfg.customCaptures != { }) then cfg.customCaptures else null;
-        };
+    languageRegister = types.coercedTo str toList (listOf str) { } ''
```
MattSturgeon, line 175:
```suggestion
    languageRegister = helpers.defaultNullOpts.mkAttrsOf (with types; coercedTo str toList (listOf str)) { } ''
```

khaneliman: Thanks, lol

plugins/languages/treesitter/treesitter.nix
GaetanLepage: Yes, this option should belong in `extraOptions`.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,253 @@
...
+            (indented_string_expression (string_fragment) @vim)
+          ]
+          (#match? @_path "^extraConfigVim(Pre|Post)?$"))
+      '';
     };
+
+    extraPlugins = mkMerge [
+      (mkIf cfg.nixGrammars [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ])
+      (mkIf (!cfg.nixGrammars) [ cfg.package ])
+    ];
```
MattSturgeon, line 245: We will need to add an option to `mkNeovimPlugin` to disable it also installing `cfg.package` (similar to options like `callSetup`). I can do that in a separate PR.

MattSturgeon (changes requested):
Here's another round of feedback.

I appreciate this is a fairly complex PR so please ask if there's anything you'd like additional guidance on. Maybe reach out on matrix if that's easier?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,253 @@
...
+
+    # TODO: default null nixgrammars
+    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
+      Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
+      This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
+    '';
+  };
+
+  settingsExample = {
+    enable = true;
```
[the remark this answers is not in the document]
khaneliman: Ah, yeah.. sorry was thinking program configuration example..

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,253 @@
...
 
-      moduleConfig = mkOption {
-        type = types.attrsOf types.anything;
-        default = { };
-        description = "This is the configuration for extra modules. It should not be used directly";
+    gccPackage = pkgs.gcc;
+    grammarPackages = pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars;
+    nixvimInjections = true;
+
+    settings = {
```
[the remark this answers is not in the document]
khaneliman: Ah, yeah.. sorry was thinking program configuration example..

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,253 @@
...
   };
 
-  config =
-    let
-      tsOptions = {
-        highlight = {
-          inherit (cfg) enable;
-          disable = if (cfg.disabledLanguages != [ ]) then cfg.disabledLanguages else null;
+  extraOptions = {
+    # TODO: default null nixgrammars
```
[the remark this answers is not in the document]
khaneliman: I was just making note that in the previous version of this derivation, they had 2 defaults. One for when nixGrammars = false (manage the grammar installation outside nix) and true (we install the grammars through nixpkgs).

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,253 @@
...
-    in
-    mkIf cfg.enable {
-      extraConfigLua =
-        (optionalString (cfg.parserInstallDir != null) ''
-          vim.opt.runtimepath:append("${cfg.parserInstallDir}")
-        '')
-        + ''
-          require('nvim-treesitter.configs').setup(${helpers.toLuaObject tsOptions})
+    languageRegister =
+      helpers.defaultNullOpts.mkAttrsOf (with types; coercedTo str toList (listOf str)) { }
```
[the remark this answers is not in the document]
khaneliman: Should probably be fine to be an empty attrsOf since we control the logic below of whether to use it

plugins/languages/treesitter/treesitter.nix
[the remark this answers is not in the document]
MattSturgeon: We tackle this in a few different ways throughout nixvim.

- For LSP, we have `plugins.lsp.servers.*.enable`
- For nvim-cmp, we have [`plugins.cmp.autoEnableSources`](https://nix-community.github.io/nixvim/plugins/cmp/index.html#pluginscmpautoenablesources) (default `true`
  - This filters `settings.sources` for "known sources" and enables the associated `plugins.*.enable`
- I'm sure there's other examples, but not OTTOMH

I think here we could have something like `autoInstallGrammers`, which filters some list in `settings` for known grammers and installs the associated packages?

khaneliman:
> Here's another round of feedback.
> 
> I appreciate this is a fairly complex PR so please ask if there's anything you'd like additional guidance on. Maybe reach out on matrix if that's easier?

Yeah, but it's been nice to get more familiar with the way a lot of these helper methods and library functions work in this project. Sorry if I'm slow with this one, on vacation and just doing bits here and there. I'll reach out on Matrix when I run into some stuff I forgot you all have that setup, too.

plugins/languages/treesitter/treesitter.nix
khaneliman: Also wondering why it was done this way in the derivation... it seems like the intent is just to override the package with the withPlugins when we have nixGrammars enabled. The package being included when false seems confusing since that'll already be provided by the cfg.package with the package option. Would it make more sense to just do this?

```suggestion
    extraPlugins = optionals cfg.nixGrammars [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ];
```

MattSturgeon: > The package being included when false seems confusing since that'll already be provided by the cfg.package with the package option

The intention is to install _either_ `cfg.package` as-is, _or_ install a modified version of `cfg.package` that bundles the `grammarPackages`.

Your suggestion would end up installing _two_ treesitter packages when `cfg.nixGrammars` is true. This is also why we need to disable `mkNeovimPlugin` from installing the package.

---

As a side note:

It's _usually_ better to use `mkIf` instead of `optionals` in module configs

This is because `mkIf` returns a `{ _type = "if"; condition = true/false; content = []; }` wrapper that the module system can merge into the final config if needed, but it can still validate `content` is correct even when `condition = false`. This approach can also avoid some inf-recursive conditions that you'd run into with a "real" nix `if then else` expression.

But sometimes `if then else` is more readable:

```nix
extraPlugins = [
  (if cfg.nixGrammars then cfg.package.withPlugins (_: cfg.grammarPackages) else cfg.package)
]
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
+        ]
+      ))
+    ];
+
+  settingsOptions = {
+    additional_vim_regex_highlighting = helpers.defaultNullOpts.mkBool false ''
+      Setting this to true will run `:h syntax` and tree-sitter at the same time.
+      Set this to `true` if you depend on 'syntax' being enabled (like for indentation).
+      Using this option may slow down your editor, and you may see some duplicate highlights.
+      Instead of true it can also be a list of languages
```
MattSturgeon, line 76: Inconvenient, but this means the type should be `either bool (listOf str)`

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
-        default = if cfg.nixGrammars then null else pkgs.gcc;
-        description = ''
-          Which package (if any) to be added as the GCC compiler.
-          This is required to build grammars if you are not using `nixGrammars`.
-          To disable the installation of GCC, set this option to `null`.
-        '';
-      };
+      keymaps =
+        let
+          keymap = default: helpers.defaultNullOpts.mkStr default ''Key shortcut'';
```
MattSturgeon, line 101:
```suggestion
          mkKeymap = default: helpers.defaultNullOpts.mkStr default "Key shortcut";
```

MattSturgeon: We should double check that upstream only accepts string values; some plugins accept either strings or tables for keymaps

MattSturgeon: If it's likely that other keymaps get added upstream, would it be simpler to just make an `attrsOf` keymaps option instead of defining individual keymap options?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
 
-  config =
-    let
-      tsOptions = {
-        highlight = {
-          inherit (cfg) enable;
-          disable = if (cfg.disabledLanguages != [ ]) then cfg.disabledLanguages else null;
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars;
```
MattSturgeon, line 175: I wonder if it's better to default to `cfg.package.passthru.allGrammars`? That'd probably work better if users (e.g.) wanted a very specific version of treesitter.

Should also consider if we need an assertion or something for when these options are used alongside a null `cfg.package`.

`package` is nullable to allow disabling installing the package, sometimes users do that in order to use the package from the environment.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
+  };
 
-      opts = mkIf cfg.folding {
-        foldmethod = "expr";
-        foldexpr = "nvim_treesitter#foldexpr()";
-      };
+  callSetup = false;
+  extraConfig = cfg: {
+    extraConfigLua =
+      (optionalString (cfg.settings.parser_install_dir != null) ''
```
MattSturgeon, line 206: This makes me think `parser_install_dir` is our option, not an upstream setting?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
+      + (optionalString (cfg.languageRegister != { }) ''
+        __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
+
+        for parser_name, ft in pairs(__parserFiletypeMappings) do
+          require('vim.treesitter.language').register(parser_name, ft)
+        end
+      '');
+
+    extraFiles = mkIf cfg.nixvimInjections {
+      "queries/nix/injections.scm" = readFile ./queries/nix/injections.scm;
```
MattSturgeon, line 221: Nice.

IMO nesting the file in a directory is unnecessary, it can just be a sibling to this file.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Should also be wrapped in a `mkIf (cfg.package != null)`.

We probably also want assertions for any options that need a non-null cfg.package being used.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,235 @@
...
+    extraPlugins = mkMerge [
+      (mkIf cfg.nixGrammars [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ])
+      (mkIf (!cfg.nixGrammars) [ cfg.package ])
+    ];
+
+    extraPackages = with pkgs; [
+      tree-sitter
+      nodejs
+      cfg.gccPackage
+    ];
```
MattSturgeon, line 233: @GaetanLepage should this be upstreamed as nixpkgs dependencies?

I.e. `pkgs.tree-sitter` and `pkgs.nodejs`.

MattSturgeon:
CI error:
> error: build of '/nix/store/pqzsfmbgw1izanmx1vr4hqffjz81mvg3-neovim-0.10.0.drv' on 'ssh-ng://nix@darwin03.nix-community.org' failed: error: path '/nix/store/25gm2jy676a7j1ynn5bw1ann1jy9z0xp-init.lua' is not valid

https://buildbot.nix-community.org/#/builders/190/builds/1211

Not sure why it only shows up on one specific system though?

---

As the rename aliases seem to be working, can probably update the these to use the new `.settings` option paths now.

---

Might not feel like it, but this PR is coming to gether nicely and I think you're on the home stretch now!

plugins/languages/treesitter/treesitter.nix
khaneliman: https://github.com/nvim-treesitter/nvim-treesitter/blob/a8c082b4f5b3b70bde2ffb08cc112dfd7e515fb3/README.md?plain=1#L550 Just requires some extra vim runtime config according to upstream docs.

MattSturgeon: Ah ok. In that case we should have a comment to that effect.

plugins/languages/treesitter/treesitter.nix
khaneliman: > We should double check that upstream only accepts string values; some plugins accept either strings or tables for keymaps

https://github.com/nvim-treesitter/nvim-treesitter/blob/a8c082b4f5b3b70bde2ffb08cc112dfd7e515fb3/lua/nvim-treesitter/incremental_selection.lua#L138 looks like it is just the string. It uses the string you pass in with the function name to create the keymap.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
-        type =
-          with types;
-          oneOf [
-            (enum [ "all" ])
-            (listOf str)
-          ];
-        default = "all";
-        description = "Either \"all\" or a list of languages";
-      };
+  optionsRenamedToSettings = [
```
GaetanLepage, line 17:
```suggestion
  # TODO introduced 2024-07-06: remove after 24.11
  optionsRenamedToSettings = [
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
-        default = if cfg.nixGrammars then null else "$XDG_DATA_HOME/nvim/treesitter";
-        description = ''
-          Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
-          This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
-        '';
-      };
+  settingsOptions = {
+    additional_vim_regex_highlighting =
+      with types;
+      helpers.defaultNullOpts.mkNullable (either bool (listOf str)) false ''
```
MattSturgeon, line 74:
```suggestion
      with helpers.nixvimTypes;
      helpers.defaultNullOpts.mkNullableWithRaw (either bool (listOf (maybeRaw str))) false ''
```

`mkListOf` would wrap the elemType with `maybeRaw`. We should replicate that as it allows users to set list items to raw lua code (e.g. use a string defined in a local variable somewhere)

mkNullableWithRaw allows to option to be rawLua at it's top-level. `mkListOf` doesn't do that currently, but that's a bug.

You could also do this if preferred, it's effectively the same thing:

```suggestion
      with helpers.nixvimTypes;
      helpers.defaultNullOpts.mkNullable (oneOf [
        bool
        (listOf (maybeRaw str))
        rawLua
      ]) false ''
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
-            attrpath: (attrpath (identifier) @_path)
-            expression: [
-              (string_expression (string_fragment) @vim)
-              (indented_string_expression (string_fragment) @vim)
-            ]
-            (#match? @_path "^extraConfigVim(Pre|Post)?$"))
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = config.plugins.treesitter.package.passthru.allGrammars;
+      defaultText = literalExpression "pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars";
```
MattSturgeon, line 178:
```suggestion
      default = config.plugins.treesitter.package.passthru.allGrammars or [ ];
      defaultText = literalExpression "config.plugins.treesitter.package.passthru.allGrammars";
```

Not sure, but I _think_ the default may still be evaluated even when some of those assertions fail. If so, we should add `or [ ]` to the value.

Maybe add a test for a null `plugins.treesitter.package` or play around with this in another nixvim config that has `flake lock --override-input` pointing to your branch?

Also, we should update the defaultText to match (don't need the `or [ ]` there though, as it would never apply in real usage)

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+    };
+  };
+
+  callSetup = false;
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = cfg.nixGrammars -> cfg.package != null;
+        message = ''
+          Nixvim (plugins.treesitter): You have defined nixGrammars, but `plugins.treesitter.package` is `null`.
```
MattSturgeon, line 211:
```suggestion
          Nixvim (plugins.treesitter): You have enabled `nixGrammars`, but `plugins.treesitter.package` is `null`.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+          This package is required to include grammars to install.
         '';
-      };
+      }
+      {
+        assertion =
+          ((builtins.length cfg.grammarPackages > 0) || cfg.grammarPackages == "all") -> cfg.package != null;
+        message = ''
+          Nixvim (plugins.treesitter): You have defined grammarPackages, but `plugins.treesitter.package` is `null`
```
MattSturgeon, line 219:
```suggestion
          Nixvim (plugins.treesitter): You have enabled `grammarPackages`, but `plugins.treesitter.package` is `null`.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+        assertion = cfg.nixGrammars -> cfg.package != null;
+        message = ''
+          Nixvim (plugins.treesitter): You have defined nixGrammars, but `plugins.treesitter.package` is `null`.
+          This package is required to include grammars to install.
         '';
-      };
+      }
+      {
+        assertion =
+          ((builtins.length cfg.grammarPackages > 0) || cfg.grammarPackages == "all") -> cfg.package != null;
```
MattSturgeon, line 217:
```suggestion
          (cfg.grammarPackages != [ ] || cfg.grammarPackages == "all") -> cfg.package != null;
```

Better to compare to the empty list, checking `length > 0` requires nix to count the list elements.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+        message = ''
+          Nixvim (plugins.treesitter): You have defined grammarPackages, but `plugins.treesitter.package` is `null`
+          This package is required to choose which grammars to install.
+        '';
+      }
+      {
+        assertion = !cfg.nixGrammars -> cfg.gccPackage != null;
+        message = ''
+          Nixvim (plugins.treesitter): You have disabled nixGrammars, but `plugins.treesitter.gccPackage` is `null`
```
MattSturgeon, line 226:
```suggestion
          Nixvim (plugins.treesitter): You have disabled `nixGrammars`, but `plugins.treesitter.gccPackage` is `null`.
```

I don't mind too much if we use backticks in warnings assertions (they print to the console, so don't get any formatting applied). But we should try and be consistent with style. :smile:

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+      # NOTE: It MUST be at the beginning of runtimepath. Otherwise the parsers from Neovim itself
+      # is loaded that may not be compatible with the queries from the 'nvim-treesitter' plugin.
+      (optionalString (cfg.settings.parser_install_dir != null) ''
+        vim.opt.runtimepath:prepend("${cfg.settings.parser_install_dir}")
+      '')
+      + ''
+        require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings})
+      ''
+      + (optionalString (cfg.languageRegister != { }) ''
+        __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
```
MattSturgeon, line 242:
```suggestion
        local __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
```

Also, lets wrap the treesitter config in a `do ... end` block, to avoid polluting the global scope.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+      '');
+
+    extraFiles = mkIf cfg.nixvimInjections {
+      "queries/nix/injections.scm" = readFile ./injections.scm;
     };
+
+    extraPlugins = mkIf (cfg.package != null) (mkMerge [
+      (mkIf cfg.nixGrammars [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ])
+      (mkIf (!cfg.nixGrammars) [ cfg.package ])
+    ]);
```
MattSturgeon, line 256: Because it's a list, we can simplify this and avoid using mkMerge:
```suggestion
    extraPlugins = mkIf (cfg.package != null) [
      (mkIf cfg.nixGrammars (cfg.package.withPlugins (_: cfg.grammarPackages)))
      (mkIf (!cfg.nixGrammars) cfg.package)
    ];
```

MattSturgeon (commented):
Coming together nicely! Do you need any help updating the test cases?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,264 @@
...
+    };
+
+    nixvimInjections = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''Whether to enable Nixvim injections, such as recognising extraConfigLua as lua.'';
+    };
+  };
+
+  callSetup = false;
```
MattSturgeon, line 205: #1823 added an `installPackage` option, we should disable that because the package is installed manually.
```suggestion
  callSetup = false;
  installPackage = false;
```

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Had a [discussion on matrix](https://matrix.to/#/!iBrKeqhFRZaBBGodLH:matrix.org/$YTqJrZwgfeGRswYV1W9TWDoCkVE4K-4vr09jAwU96dw?via=matrix.org&via=vhack.eu&via=gandom.nl)

- `pkgs.tree-sitter` shouldn't be needed as it is already bundled with neovim.
- `pkgs.nodejs` should be configurable by a nullable package option (like `cfg.gccPackage` is)
- `cfg.nodejsPackage` and `cfg.gccPackage` should probably default to null when `cfg.nixGrammars` is enabled? (This is a breaking change though, not 100% decided if we should do it) 

```nix
extraPackages = [
  cfg.nodejsPackage
  cfg.gccPackage
];
```

khaneliman: If nodejs is just a requirement at runtime to build the grammar packages then I think it makes sense to follow in the same lines as the gccPackage.

MattSturgeon: Yes, exactly

plugins/languages/treesitter/treesitter.nix
khaneliman: Just setting up a test for a null package doesn't seem like it's even accepted. 
```nix
    plugins.treesitter = {
      enable = true;

      package = null;
    };
```

```bash
error: A definition for option `plugins.treesitter.package' is not of type `package'. Definition values:
       - In `<unknown-file>': null
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,303 @@
...
+      type = types.bool;
+      default = true;
+      example = true;
+      description = "Whether to enable tree-sitter based folding.";
+    };
+
+    gccPackage = helpers.mkPackageOption {
+      name = "gcc";
+      default = pkgs.gcc;
+      example = pkgs.gcc;
```
MattSturgeon, line 170: The example will need to use `literalExpression` otherwise it will render as `<derivation gcc>` or similar.

Same applies to the default:

```suggestion
      defaultText = literalExpression "pkgs.gcc";
      example = literalExpression "pkgs.gcc";
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,303 @@
...
+        Which package (if any) to be added as the GCC compiler.
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of GCC, set this option to `null`.
+      '';
+    };
+
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = config.plugins.treesitter.package.passthru.allGrammars or [ ];
+      example = pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars;
```
MattSturgeon, line 181: Same (use `literalExpression`)

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,303 @@
...
+      type = types.bool;
+      default = true;
+      example = true;
+      description = ''Whether to enable Nixvim injections, such as recognising extraConfigLua as lua.'';
+    };
+
+    nodejsPackage = helpers.mkPackageOption {
+      name = "nodejs";
+      default = pkgs.nodejs;
+      example = pkgs.nodejs;
```
MattSturgeon, line 220: Same: use `literalExpression`, add a `defaultText` and consider changing/removing the example

plugins/languages/treesitter/treesitter.nix
MattSturgeon: In this specific case, I'm not sure the example is useful, since it is the same as the default.

Either remove the example, or set it to something non-default.

MattSturgeon (changes requested):

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,303 @@
...
+      # is loaded that may not be compatible with the queries from the 'nvim-treesitter' plugin.
+      (optionalString (cfg.settings.parser_install_dir != null) ''
+        do
+          vim.opt.runtimepath:prepend("${cfg.settings.parser_install_dir}")
+        end
+      '')
+      + ''
+        do
+          require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings})
+        end
```
MattSturgeon, line 277: I think it's fine if the `do .. end` block encloses the _entire_ `extraConfigLua` definition.

i.e. one `do .. end` block, not multiple

khaneliman: Alright, just saw another example in repo do it this way and thought it was preferred. Can change it back.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Interesting. mkNeovimPlugin [is using](https://github.com/nix-community/nixvim/blob/843fb302ebaa2559b7c50ad69ba9a00ae1a5836d/lib/neovim-plugin.nix#L93) `mkPluginPackageOption`, which is [apparently not nullable](https://github.com/nix-community/nixvim/blob/843fb302ebaa2559b7c50ad69ba9a00ae1a5836d/lib/options.nix#L333-L339).

@GaetanLepage is this what we want? It was my interpretation that our package options should be nullable.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: I haven't looked too closely at other examples, so IDK if other maintainers have a preference.

To me, it makes sense to just have one scope unless there's a specific reason not to.

GaetanLepage: I don't think we really need a `do .. end` block here as there is no issue with local variable leaking.
As we don't use that for most plugins, I would just not have it.
(Have I missed something) ?

MattSturgeon: We have the local `__parserFiletypeMappings`. Not a big deal as it's double underscored, but I think having a `do .. end` is better

plugins/languages/treesitter/treesitter.nix
khaneliman: >     * `pkgs.tree-sitter` shouldn't be needed as it is already bundled with neovim.

Seeing when we remove this package. Are we sure it's not needed for !nixGrammars?

```bash
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
```

MattSturgeon: Based on that warning I'd say it is required. In that case I think we should handle it the same as nodejs and gcc.

plugins/languages/treesitter/treesitter.nix
MattSturgeon: I think for now let's just trust that `cfg.package` is never null, so remove any assertions (etc) related to it being nullable.

If we change `mkPluginPackageOption` in the future we'll have to consider how it affects plugins like this one.

MattSturgeon:
Looks like our current test doesn't have test-cases for any of the settings options.

Normally plugins of this scale have an "empty" test (with just `enable = true`), a "defaults" test (where all settings are explicitly set to their plugin-defaults), an "example" test (usually similar to `settingsExample`) and zero-or-more more _specific_ tests that test special or unusual aspects of the plugin (or integration with other plugins).

- [obsidian](https://github.com/nix-community/nixvim/blob/6674dea8403747827431d4d8497c34023f93d047/tests/test-sources/plugins/utils/obsidian.nix) has two examples (but no "defaults")
- [oil](https://github.com/nix-community/nixvim/blob/6674dea8403747827431d4d8497c34023f93d047/tests/test-sources/plugins/utils/oil.nix) is the same
- [ccc](https://github.com/nix-community/nixvim/blob/6674dea8403747827431d4d8497c34023f93d047/tests/test-sources/plugins/utils/ccc.nix) has an extensive "defaults" case and an example case too
- [yanky](https://github.com/nix-community/nixvim/blob/6674dea8403747827431d4d8497c34023f93d047/tests/test-sources/plugins/utils/yanky.nix) has all the standard tests, and also a "with-telescope" integration test

To make the "defaults" case, I usually start by copying the defaults documented upstream and then reformatting it as nix code using some vim macros and/or `:s` commands.

---

Looks like the "custom" test-case has always been commented out, ever since the treesitter test was introduced by @GaetanLepage in #235. (The [old tests](https://github.com/nix-community/nixvim/tree/89f54be73a201b874bbbf3427b2181e6d589ce79/tests/plugins) before that refactor didn't have a treesitter test).

I think we should either remove it or attempt to uncomment it.

---

I think once the test is up to scratch and the last couple outstanding threads are resolved, this PR should be finished, baring any small nit-picks in the final review.

Well done for persisting with this one! :tada:

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,306 @@
...
+      + ''
+        end
+      '';
 
-      opts = mkIf cfg.folding {
-        foldmethod = "expr";
-        foldexpr = "nvim_treesitter#foldexpr()";
-      };
+    extraFiles = mkIf cfg.nixvimInjections {
+      "queries/nix/injections.scm" = readFile ./injections.scm;
```
MattSturgeon, line 293:
```suggestion
      "queries/nix/injections.scm".source = ./injections.scm;
```
See #1828

GaetanLepage:
> Looks like the "custom" test-case has always been commented out, ever since the treesitter test was introduced by @GaetanLepage in #235. (The [old tests](https://github.com/nix-community/nixvim/tree/89f54be73a201b874bbbf3427b2181e6d589ce79/tests/plugins) before that refactor didn't have a treesitter test).
> 
> I think we should either remove it or attempt to uncomment it.

I agree,  we should clarify this before merging. Of course, uncommenting is the thing to try first.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,316 @@
...
-  config =
-    let
-      tsOptions = {
-        highlight = {
-          inherit (cfg) enable;
-          disable = if (cfg.disabledLanguages != [ ]) then cfg.disabledLanguages else null;
+  # NOTE: We call require('nvim-treesitter.configs').setup in extraConfigLua.
+  callSetup = false;
+  # NOTE: We dont want to try installing the provided package because we install it in extraPlugins
+  installPackage = false;
```
MattSturgeon, line 245:
```suggestion
  # We call setup manually below.
  callSetup = false;
  # We install cfg.package manually so we can install grammars using it.
  installPackage = false;
```

Maybe something along these lines would be bit less verbose?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,316 @@
...
-        };
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = ((cfg.grammarPackages != [ ]) || cfg.grammarPackages == "all") -> cfg.package != null;
+        message = ''
+          Nixvim (plugins.treesitter): You have enabled `grammarPackages`, but `plugins.treesitter.package` is `null`.
+          This package is required to choose which grammars to install.
+        '';
+      }
```
MattSturgeon: Turns out cfg.package isn't nullable atm, so we can drop this assertion.

@GaetanLepage unless you think we should keep it in case we change `package`'s type to be nullable?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,316 @@
...
-      extraConfigLua =
-        (optionalString (cfg.parserInstallDir != null) ''
-          vim.opt.runtimepath:append("${cfg.parserInstallDir}")
+    extraConfigLua =
+      ''
+        do
+      ''
+      +
+        # NOTE: It MUST be at the beginning of runtimepath. Otherwise the parsers from Neovim itself
+        # is loaded that may not be compatible with the queries from the 'nvim-treesitter' plugin.
```
MattSturgeon, line 269:
```suggestion
        # NOTE: Upstream state that the parser MUST be at the beginning of runtimepath.
        # Otherwise the parsers from Neovim takes precedent, which may be incompatible with some queries.
```

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -1,14 +1,55 @@
...
+          keymaps = {
+            init_selection = "gnn";
+            node_incremental = "grn";
+            scope_incremental = "grc";
+            node_decremental = "grm";
+          };
+        };
+        indent = false;
+      };
+    };
```
MattSturgeon, line 24: Is this an exhaustive list of defaults?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,316 @@
...
+        Which package (if any) to be added as the nodejs package.
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of NodeJS, set this option to `null`.
+      '';
+    };
+
+    treesitterPackage = helpers.mkPackageOption {
+      name = "treesitter";
+      default = pkgs.tree-sitter;
+      defaultText = literalExpression "pkgs.treesitter";
```
MattSturgeon, line 233:
```suggestion
      defaultText = literalExpression "pkgs.tree-sitter";
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,316 @@
...
+        The keys are the parser names and the values are either one or several filetypes.
+      '';
+    };
 
-      nixvimInjections = mkEnableOption "nixvim specific injections, like lua highlighting in extraConfigLua";
+    nixGrammars = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = "Whether to automatically install grammar packages.";
```
MattSturgeon, line 208:
```suggestion
      description = "Whether to install grammars defined in `grammarPackages`.";
```

?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,300 @@
...
-        '';
-      };
+  settingsOptions = {
+    additional_vim_regex_highlighting =
+      with helpers.nixvimTypes;
+      helpers.defaultNullOpts.mkNullableWithRaw (either bool (listOf (maybeRaw str))) false ''
+        Setting this to true will run `:h syntax` and tree-sitter at the same time.
+        Set this to `true` if you depend on 'syntax' being enabled (like for indentation).
+        Using this option may slow down your editor, and you may see some duplicate highlights.
+        Instead of true it can also be a list of languages
```
MattSturgeon, line 79: Not saying this is wrong, but just checking if you're aware of how commonmark markdown handles linebreaks?

TL;DR, a fully blank line is needed for a paragraph break.\
Or you can render a `<br>` by ending a line with `\`.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,300 @@
...
-          ];
-        };
-      };
+    ignore_install = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      List of parsers to ignore installing (for "all")
+    '';
+
+    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
+      Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
+      This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
```
MattSturgeon, line 130:
```suggestion
      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).

      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default. Otherwise, change it to something that will work for you!
```

MattSturgeon (commented):
Just some small comments and nits

tests/test-sources/plugins/languages/treesitter/treesitter.nix
khaneliman: This is just what I saw used in their https://github.com/nvim-treesitter/nvim-treesitter/blob/a8c082b4f5b3b70bde2ffb08cc112dfd7e515fb3/lua/nvim-treesitter/configs.lua#L47-L76 lua config

MattSturgeon: Thanks for the link!

Looks like `highlight.custom_captures` is deprecated now, and they recommend it is set imperatively with a function.

Maybe we should drop the option now then?

MattSturgeon: Also, their `indent` default apperears to be:

```lua
indent = {
  module_path = "nvim-treesitter.indent",
  enable = false,
  is_supported = queries.has_indents,
}
```

Do they document it as `false` elsewhere?

plugins/languages/treesitter/treesitter.nix
khaneliman: I think I just copied and pasted this from their documentation. I can force the linebreaks to increase readability.

tests/test-sources/plugins/languages/treesitter/treesitter.nix
khaneliman: The readme mentions they default every module to false and force you to enable the ones you want to use.

MattSturgeon: Looking at some of their other docs, I think some optional options are missing from the table you linked.

E.g. `highlight.disable` (func(lang, buf) -> bool) or (table; listOf lang)

khaneliman: Yeah, looks like they have some sync_install and auto_install stuff in the man

plugins/languages/treesitter/treesitter.nix
MattSturgeon: @khaneliman you're welcome to add back the `NOTE:` prefix if you prefer.

tests/test-sources/plugins/languages/treesitter/treesitter.nix
khaneliman: Going through and cleaning these defaults up, sorry.. taking a little bit of time lol

MattSturgeon: No rush!

khaneliman: Alright, think I got it rearranged and updated. Looked like I had a property at top-level instead of under highlight, too.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+      '';
 
-      disabledLanguages = mkOption {
-        type = types.listOf types.str;
-        default = [ ];
-        description = "A list of languages to disable";
-      };
+      disable =
+        with types;
+        helpers.defaultNullOpts.mkStrLuaFnOr (listOf str) "null" ''
```
MattSturgeon, line 99:
```suggestion
        helpers.defaultNullOpts.mkStrLuaFnOr (listOf str) null ''
```

Unless you mean the _string_ `"null"`

khaneliman: No, sorry... I just thought I remembered a PR where we had to do a null string

MattSturgeon: Yes this changed only a few weeks ago

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -1,14 +1,66 @@
+{ pkgs, ... }:
 {
+  default = {
+    plugins.treesitter = {
+      enable = true;
+
+      settings = {
+        auto_install = false;
+        ensure_installed = [ ];
```
MattSturgeon, line 9: Good spot. I assumed the default was `"all"` but this [seems](https://github.com/nvim-treesitter/nvim-treesitter/blob/64f6f0ab4e3f613aa682eb5fe29c5025db500ddd/lua/nvim-treesitter/configs.lua#L23) [correct](https://github.com/nvim-treesitter/nvim-treesitter/blob/64f6f0ab4e3f613aa682eb5fe29c5025db500ddd/lua/nvim-treesitter/configs.lua#L609-L614).

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -1,14 +1,66 @@
...
+  };
+
   empty = {
     plugins.treesitter.enable = true;
   };
 
-  nonix = {
-    # TODO: See if we can build parsers (legacy way)
-    tests.dontRun = true;
+  emptyGrammarPackages = {
```
MattSturgeon, line 42:
```suggestion
  empty-grammar-packages = {
```

We normally kebab-case our test-case names

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -1,14 +1,66 @@
...
     plugins.treesitter = {
       enable = true;
-      nixGrammars = false;
+
+      grammarPackages = [ ];
+    };
+  };
+
+  highlightDisableFunction = {
```
MattSturgeon, line 50: Same

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -27,18 +79,43 @@
...
+  nonix = {
+    # TODO: See if we can build parsers (legacy way)
+    tests.dontRun = true;
+    plugins.treesitter = {
+      enable = true;
+      nixGrammars = false;
+    };
+  };
+
+  specificGrammars = {
```
MattSturgeon, line 91: Same

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-              (indented_string_expression (string_fragment) @vim)
-            ]
-            (#match? @_path "^extraConfigVim(Pre|Post)?$"))
-        '';
-      };
+      + ''
+        require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings})
+      ''
+      + (optionalString (cfg.languageRegister != { }) ''
+        local __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}
```
MattSturgeon, line 294: Sorry to change my mind on this, but I think the `do .. end` scope only needs to wrap this optional string:

```nix
(optionalString (cfg.languageRegister != { }) ''
  do
    local __parserFiletypeMappings = ${helpers.toLuaObject cfg.languageRegister}

    for parser_name, ft in pairs(__parserFiletypeMappings) do
      require('vim.treesitter.language').register(parser_name, ft)
    end
  end
'')
```

Because the only variable that could leak to the rest of the init.lua scope is `__parserFiletypeMappings`

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-        };
+  extraConfig = cfg: {
+    assertions =
+      map
+        (name: {
+          assertion = !cfg.nixGrammars -> cfg.${name} != null;
+          message = ''
+            Nixvim (plugins.treesitter): You have disabled `nixGrammars`, but `plugins.treesitter.${name}` is `null`.
+            This package is required to build the grammars at runtime.
+          '';
```
MattSturgeon, line 272: Does this need to be an assertion or is a warning enough?

Is it foreseeable that a user would want to set these packages to null and also disable `nixGrammars`, but not compile any grammars at runtime?

If so, I think we want a warning, otherwise the assertion is correct.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+    };
+
+    treesitterPackage = helpers.mkPackageOption {
+      name = "tree-sitter";
+      default = pkgs.tree-sitter;
+      defaultText = literalExpression "pkgs.tree-sitter";
+      description = ''
+        Which package (if any) to be added as the treesitter binary.
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of treesitter, set this option to `null`.
```
MattSturgeon, line 254:
```suggestion
        Which package (if any) to be added as the treesitter binary.

        This is required to build grammars if you are not using `nixGrammars`.
        To disable the installation of treesitter, set this option to `null`.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-      nixvimInjections = mkEnableOption "nixvim specific injections, like lua highlighting in extraConfigLua";
+    nodejsPackage = helpers.mkPackageOption {
+      name = "nodejs";
+      default = pkgs.nodejs;
+      defaultText = literalExpression "pkgs.nodejs";
+      example = literalExpression "pkgs.nodejs_22";
+      description = ''
+        Which package (if any) to be added as the nodejs package.
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of NodeJS, set this option to `null`.
```
MattSturgeon, line 243:
```suggestion
        Which package (if any) to be added as the nodejs package.

        This is required to build grammars if you are not using `nixGrammars`.
        To disable the installation of NodeJS, set this option to `null`.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+      default = true;
+      example = false;
+      description = "Whether to install grammars defined in `grammarPackages`.";
+    };
+
+    nixvimInjections = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = ''Whether to enable Nixvim injections, such as recognising extraConfigLua as lua.'';
```
MattSturgeon, line 232:
```suggestion
      description = ''
        Whether to enable Nixvim injections, e.g. highlighting `extraConfigLua` as lua.
      '';
```

Or use `"` if being on one line is preferred.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+        cpp = "onelab";
+        python = [
+          "foo"
+          "bar"
+        ];
       };
+      description = ''
+        This is a wrapping of the `vim.treesitter.language.register` function.
+        Register specific parsers to one or several filetypes.
+        The keys are the parser names and the values are either one or several filetypes.
```
MattSturgeon, line 217:
```suggestion
        This is a wrapping of the `vim.treesitter.language.register` function.

        Register specific parsers to one or several filetypes.

        The keys are the parser names and the values are either one or several filetypes.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = config.plugins.treesitter.package.passthru.allGrammars or [ ];
+      example = literalExpression "pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars";
+      defaultText = literalExpression "config.plugins.treesitter.package.passthru.allGrammars";
+      description = "Grammar packages to install";
+    };
+
+    languageRegister = mkOption {
+      type = with types; attrsOf (coercedTo str toList (listOf str));
```
MattSturgeon, line 205: I assume we don't need rawLua support? If that'd be beneficial I think a TODO comment is fine for now.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+      '';
+    };
 
-      moduleConfig = mkOption {
-        type = types.attrsOf types.anything;
-        default = { };
-        description = "This is the configuration for extra modules. It should not be used directly";
+    grammarPackages = mkOption {
+      type = with types; listOf package;
+      default = config.plugins.treesitter.package.passthru.allGrammars or [ ];
```
MattSturgeon, line 198:
```suggestion
      default = config.plugins.treesitter.package.passthru.allGrammars;
```
Turns out this wasn't needed

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+
+    gccPackage = helpers.mkPackageOption {
+      name = "gcc";
+      default = pkgs.gcc;
+      defaultText = literalExpression "pkgs.gcc";
+      example = literalExpression "pkgs.gcc14";
+      description = ''
+        Which package (if any) to be added as the GCC compiler.
+        This is required to build grammars if you are not using `nixGrammars`.
+        To disable the installation of GCC, set this option to `null`.
```
MattSturgeon, line 192:
```suggestion
        Which package (if any) to be added as the GCC compiler.

        This is required to build grammars if you are not using `nixGrammars`.
        To disable the installation of GCC, set this option to `null`.
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-        };
-      };
+    ignore_install = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      List of parsers to ignore installing (for "all")
+    '';
+
+    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
+      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
+
+      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default. Otherwise, change it to something that will work for you!
```
MattSturgeon, line 143:
```suggestion
      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
      Otherwise, change it to something that will work for you!
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-        example = {
-          cpp = "onelab";
-          python = [
-            "myFiletype"
-            "anotherFiletype"
-          ];
-        };
-      };
+    ignore_install = helpers.defaultNullOpts.mkListOf types.str [ ] ''
+      List of parsers to ignore installing (for "all")
```
MattSturgeon, line 137:
```suggestion
      List of parsers to ignore installing. Used when `ensure_installed` is set to "all".
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+          node_incremental = mkKeymap "grn";
+          scope_incremental = mkKeymap "grc";
+          node_decremental = mkKeymap "grm";
         };
+    };
 
-      indent = mkEnableOption "tree-sitter based indentation";
+    indent = {
+      enable = helpers.defaultNullOpts.mkBool false ''
+        Whether to enable tree-sitter based indentation.
```
MattSturgeon, line 128:
```suggestion
        Whether to enable treesitter based indentation.
```

I think we don't have the hyphen elsewhere?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
+      '';
 
-      disabledLanguages = mkOption {
-        type = types.listOf types.str;
-        default = [ ];
-        description = "A list of languages to disable";
-      };
+      disable =
+        with types;
+        helpers.defaultNullOpts.mkStrLuaFnOr (listOf str) null ''
```
MattSturgeon, line 99:
```suggestion
        with helpers.nixvimTypes;
        helpers.defaultNullOpts.mkStrLuaFnOr (listOf (maybeRaw str)) null ''
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,317 @@
...
-          Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
-          This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
+    highlight = {
+      additional_vim_regex_highlighting =
+        with helpers.nixvimTypes;
+        helpers.defaultNullOpts.mkNullableWithRaw (either bool (listOf (maybeRaw str))) false ''
+          Setting this to true will run `:h syntax` and tree-sitter at the same time. \
+          Set this to `true` if you depend on 'syntax' being enabled (like for indentation). \
+          Using this option may slow down your editor, and you may see some duplicate highlights. \
+          Instead of true it can also be a list of languages
```
MattSturgeon, line 90:
```suggestion
          Setting this to true will run `syntax` and tree-sitter at the same time. \
          Set this to `true` if you depend on 'syntax' being enabled (e.g. for indentation). \
          See `:h syntax`.

          Using this option may slow down your editor, and you may see some duplicate highlights. \
          Instead of true, it can also be a list of languages.
```

MattSturgeon (changes requested):
Everything looks good. I think this is my final nit-picks:

I'd appreciate if @GaetanLepage or @traxys also give this a once-over, if they have time, as these upgrade PRs are always tricky to review. But if not I'll merge after a short wait.

plugins/languages/treesitter/treesitter.nix
khaneliman: I think the hyphen is for the actual binary. https://tree-sitter.github.io/tree-sitter/

MattSturgeon: I'm find with whatever you think is correct

plugins/languages/treesitter/treesitter.nix
khaneliman: Hmm.. maybe along the same lines of thinking that they managed the packages outside of nixvim. Like we talked about with the main package binary being null.

MattSturgeon: Unless you (or anyone) disagrees, I'm fine with the assertion. We can always re-visit if someone complains.

If you did want to do a warning, the schema is slightly different; it is a simple list of warning strings.\
Therefore, we need to use `optional`, `filter` or `mkIf` to decide whether or not each warning should be set.

MattSturgeon (approved):
LGTM, thanks for your effort, dedication and patience! :tada: 

I'll leave this open to give others a chance to review, then merge if there's no additional feedback.

khaneliman:
> LGTM, thanks for your effort, dedication and patience! 🎉
> 
> I'll leave this open to give others a chance to review, then merge if there's no additional feedback.

Pushed a quick update for the assertion -> warning

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,321 @@
...
+          optional (!cfg.nixGrammars && cfg.${name} == null) ''
+            Nixvim (plugins.treesitter): You have disabled `nixGrammars`, but `plugins.treesitter.${name}` is `null`.
+            This package is required to build the grammars at runtime.
+          ''
+        )
+        [
+          "gccPackage"
+          "nodejsPackage"
+          "treesitterPackage"
+        ];
```
traxys, line 287: I don't think we should have those assertions, there are very good reasons to have them null with nixGrammars = false, I'd say it's even the reason they are nullable.

The main reason is that you may want to use the C compiler on your PATH, to find required libraries for example. We had an issue that required this change in the first place, as using this gcc will make other stuff break, for example the `cc` in the neovim terminal may use a bad compiler.

traxys: (Because setting them to null is an explicit action from the user if I understand the code correctly)

MattSturgeon: > I don't think we should have those assertions

As these are warnings I don't think it's so bad? We did have assertions before but they were switched to warnings for the kind've reasons you describe.

Is your issue that the user cannot silence the warnings?

traxys: Yeah exactly

MattSturgeon: Although I think I was the one to suggest assertions/warnings in the first place, after discussing on matrix we've concluded they aren't worth it in this case:
- They will only show for users who have explicitly disabled these packages
- Such users should know what they're doing
- If we had warnings, there'd be no practical way for users to silence them
Therefore the cons outweigh the pros.

GaetanLepage: I agree with this conclusion (in case someone cares).

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,306 @@
...
+  originalName = "nvim-treesitter";
+  luaName = "nvim-treesitter.configs";
+  defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
-      nixGrammars = mkOption {
-        type = types.bool;
-        default = true;
-        description = "Install grammars with Nix";
-      };
+  maintainers = [ ];
```
GaetanLepage, line 15: We need a maintainer.

GaetanLepage (changes requested):

plugins/languages/treesitter/treesitter.nix
MattSturgeon:
```suggestion
  maintainers = [ lib.maintainers.khaneliman ];
```

Not sure if I missed this or it got mistakenly reverted?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,306 @@
...
-      parserInstallDir = mkOption {
-        type = types.nullOr types.str;
-        default = if cfg.nixGrammars then null else "$XDG_DATA_HOME/nvim/treesitter";
-        description = ''
-          Location of the parsers to be installed by the plugin (only needed when nixGrammars is disabled).
-          This default might not work on your own install, please make sure that $XDG_DATA_HOME is set if you want to use the default. Otherwise, change it to something that will work for you!
+    highlight = {
+      additional_vim_regex_highlighting =
+        with helpers.nixvimTypes;
+        helpers.defaultNullOpts.mkNullableWithRaw (either bool (listOf (maybeRaw str))) false ''
```
GaetanLepage, line 86: That's more a nit pick but we would have the `with helpers.nixvimTypes;` within the type expression.
Not a big deal, but this would be better for consistency.

GaetanLepage (changes requested):

plugins/languages/treesitter/treesitter.nix
```diff
@@ -6,210 +6,308 @@
...
-      };
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "treesitter"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    [
```
MattSturgeon, line 54:
```suggestion
    [
      (lib.mkRenamedOptionModule (basePluginPath ++ [ "moduleConfig" ]) settingsPath)
```

khaneliman: I'm so confused where that change disappeared... i just added that lol

MattSturgeon (approved):
Thanks again :+1:

Any final comments @nix-community/nixvim?

GaetanLepage (approved):
Let's go !

Congrats on this very tedious PR @khaneliman and @MattSturgeon !

traxys (approved):
My only concern was resolved :D

MattSturgeon:
@mergifyio queue

gkze:
This broke highlighting for me
<details>
<summary>before</summary>

Screenshot from 2024-07-08 11-17-20
</details>
<details>
<summary>after</summary>

Screenshot from 2024-07-08 11-16-42
</details>

Not sure how to debug...

MattSturgeon:
> This broke highlighting for me

Can you link your config? Ideally the minimal treesitter config to reproduce the issue.

khaneliman:
Looks like the module defaulted the highlight setting to enabled when the plugin was enabled instead of respecting the plugins default of everything disabled by default. Just add settings.highlight.enable = true.

gkze:
Nice that did it thank you

PR #1805, opened 2024-07-03
lib/to-lua: fix removing empties nested in lists

### LINKED ISSUE #1804: [BUG] `toLua` does not filter null values for attrs nested in lists
| Plugin       | `lualine`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

lualine sections which are overridden by user end up with square separators, instead of the default;

While looking at the generated `init.lua`,  it looks like `toLua` is failing to filter out the `null` attributes (as per `removeNullAttrValues` default option), and the section ends up with `separator = { left = nil, right = nil}` property, which is interpreted by lualine as overriding the plugin-wide default separators.

May have been caused by #1761.
lualine is just how I identified the issue, but if nixvim is failing to properly clean up nil values, this could have much wider unintended effects.

```nix
programs.nixvim = {
  plugins.lualine = {
    enable = true;
    sections.lualine_b = [
      {name = "filename";}
    ];
  };
};
```

Generated init.lua snippet:
```lua
require("lualine").setup({
    options = { icons_enabled = true },
    sections = {
        lualine_b = {
            {
                "filename",
                color = nil,
                fmt = nil,
                icon = nil,
                icons_enabled = nil,
                padding = nil,
                separator = { left = nil, right = nil },
            },
        },
    },
})
```
Expected:
```lua
require("lualine").setup({
    options = { icons_enabled = true },
    sections = {
        lualine_b = {
            {
                "filename"
            },
        },
    },
})
```

MattSturgeon:
- Add options for removing empty/null list entries (default false)
- Fix recursion into attrs that are themselves list entries
- Added some tests
  - One test for #1804
  - Two tests for the new list filters

Fixes #1804

### DIFF
--- a/lib/to-lua.nix
+++ b/lib/to-lua.nix
@@ -85,8 +85,20 @@ rec {
       removeEmptyAttrValues ? true,
 
       /**
+        If this option is true, lists like [ { } [ ] "" ]
+        will render as { "" }
+      */
+      removeEmptyListEntries ? false,
+
+      /**
+        If this option is true, lists like [ null "" ]
+        will render as { "" }
+      */
+      removeNullListEntries ? false,
+
+      /**
         If this option is true, attrsets like { a.__empty = null; }
-        will render as { ["a"] = { } }, ignoring removeEmptyAttrValues and removeNullAttrValues.
+        will render as { ["a"] = { } }, ignoring other filters such as removeEmptyAttrValues.
       */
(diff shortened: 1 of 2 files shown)


GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

PR #1806, opened 2024-07-03
plugins/lsp: Autogenerate rust-analyzer options

traxys:
This is the approach we used to take, but we had switched to another method of generating at runtime the rust-analyzer options.
This is problematic for a few reasons:
- It introduces IFD
- It makes nixvim _highly_ dependent on the exact version of nixpkgs used, in pervasive ways.

By hardcoding the rust-analyzer options we should avoid both of these issues


GaetanLepage (approved):
Looks good !

(I haven't read the entire generated file though)

flake-modules/dev/default.nix
```diff
@@ -24,6 +24,46 @@
         };
       };
 
+      packages = {
```
MattSturgeon, line 27: I think it's better to use `apps` for stuff that we want to _run_ but don't normally need to _build_.

Flake-parts [takes care of](https://flake.parts/options/flake-parts.html#opt-perSystem.apps) the weirdness with the `apps` schema, so we can just set `apps.<name>.program` to a derivation (so long as it has a `meta.mainProgram`).

MattSturgeon (changes requested):
Thanks for investigating and working on this!

Sorry if I'm sounding negative, but I'm not convinced python is the best approach here...

I _am_ 100% on board with moving stuff that'd otherwise need an IFD into the update CI.

Unless I'm missing something, the [old nix script](https://github.com/nix-community/nixvim/blob/9b25eaaa6f64a584ffccdd90b23d0962d9138352/plugins/lsp/language-servers/rust-analyzer-config.nix) is not really any more complex than the python implementation, so I don't fully see the benefit in translating so much into python.

Something like this would also allow us to update the file in CI instead of at eval time:

```nix
let
  jsonFile = pkgs.rust-analyzer.src + "/editors/code/package.json";
in
  pkgs.writeFile "rust-analyzer-options" (''
    # WARNING This file is auto-generated and should not be edited!
  '' + (lib.generators.toPretty { } (lib.importJSON jsonFile)))
```

(Even simpler if we just save the JSON and use `importJSON` at eval time).

.github/workflows/update.yml
```diff
@@ -48,3 +48,11 @@ jobs:
...
           token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
           branch: "update_flake_lock_action-${{ matrix.branch }}"
+
+      - name: Update autogenerated files
+        run: |
+          git checkout update_flake_lock_action-${{ matrix.branch }}
+          nix run .#generate-files
+          git add ./generated
+          git commit -m "generated: Update files"
+          git push
```
MattSturgeon, line 58: Maybe leave a TODO comment about improving the CI workflow?

flake-modules/dev/default.nix
```diff
@@ -24,6 +24,23 @@
...
+        # Rust-Analyzer
+        nix build .\#rust-analyzer-options
+        cat ./result >"$generated_dir"/rust-analyzer.nix
+
+        nix fmt
+      '';
+
+      packages = {
+        rust-analyzer-options = pkgs.callPackage ./updates/rust-analyzer.nix { };
+      };
```
MattSturgeon, line 42: This can be in `apps` too, seeing as it's really something to _run_ not _build_.

flake-modules/dev/default.nix
MattSturgeon, line 1: I'd still rather have this in a `generate-files.nix` module (e.g name) rather than `default.nix`.

Or maybe `updates/default.nix`?

flake-modules/dev/updates/rust-analyzer.nix
MattSturgeon, line 1: Nit: if this is a package shouldn't it be outside of flake-modules? E.g. a top-level `packages` directory tree?

plugins/languages/rust/rustaceanvim/settings-options.nix
```diff
@@ -257,8 +257,7 @@ with lib;
     settings =
       helpers.mkNullOrStrLuaFnOr
         (types.submodule {
-          # options = import ../../../lsp/language-servers/rust-analyzer-config.nix lib pkgs;
-          options = { };
+          options = import ../../../lsp/language-servers/rust-analyzer-config.nix lib;
```
MattSturgeon, line 260: If `rust-analyzer-config.nix` was a module, then this could be:

```nix
imports = [ ../../../lsp/language-servers/rust-analyzer-config.nix ];
```

And `lib` would be provided by evalModules without needing to import and pass in lib

MattSturgeon (changes requested):
This is much better! Thanks for re-visiting. :heart: 

I've left some minor comments that'd be nice to have addressed, but I don't have any real objections to this PR anymore.

I'll review again when I have time to read and understand the implementation properly.

flake-modules/dev/default.nix
MattSturgeon: Nevermind. It is just a derivation, not a derivation that produces a script so `packages` is correct.

Assuming it needs to be a flake output at all (for debugging you could just callPackage it with nix-build?

traxys: I find it easier to debug by using it as a flake output, because
- I never remember the right invocation of nix-build
- Rust-analyzer from our flake inputs will be used

flake-modules/dev/updates/rust-analyzer.nix
traxys: I'd say not really, as it is purely internal

MattSturgeon: It's only a nit, so feel free to ignore

plugins/languages/rust/rustaceanvim/settings-options.nix
traxys: Yeah but this won't be as pretty for the other includes (rust-tools & rustacean.nvim). I'm not sure we would really be gaining anything

flake-modules/dev/default.nix
traxys: Moved everything to `flake-modules/updates`

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+            lib.lists.zipLists values descriptions
+          );
+        in
+        ''
+          ${markdownDescription}
+
+          Values:
+          ${builtins.concatStringsSep "\n" valueDesc}
+
+          default:
```
MattSturgeon, line 139: `pluginDefault` should be separate from description.

The shim should pass `pluginDefault` through `helpers.pluginDefaultText` to be consistent with our defaultNullOpts.

```nix
mkOption {
  defaultText = helpers.pluginDefaultText {
    inherit pluginDefault;
    default = null;
  };
}
```

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+      };
+      description =
+        if
+          enum == null && (anyOf == null || builtins.all (subProp: !(lib.hasAttr "enum" subProp)) anyOf)
+        then
+          ''
+            ${markdownDescription}
+
+            default:
+            ```nix
```
MattSturgeon, line 165: Same, keep pluginDefault separate.

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+        assert lib.length props == 1;
+        lib.head props;
+    in
+    {
+      "${prop.name}" = mkRustAnalyzerOption prop.name prop.value;
+    }
+  ) rustAnalyzerProperties;
+in
+writeText "rust-analyzer-options.nix" (
+  "# WARNING: this file is autogenerated\n"
```
MattSturgeon, line 199: Maybe:

WARNING: DO NOT EDIT!\
This file is generated by packages.<system>.ra-update, which is run automatically by CI.

Or something along those lines? Helpful to know "what, why and where" at a glance.

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+  + (lib.generators.toPretty { } (lib.foldl (a: b: a // b) { } rustAnalyzerOptions))
```
MattSturgeon, line 200: IMO this is more readable:

```suggestion
  + (lib.generators.toPretty { } (lib.foldl lib.updateAttrs { } rustAnalyzerOptions))
```

(Or `updateAttrsRecursive` if there's a chance we'd need that?)

plugins/lsp/language-servers/rust-analyzer-config.nix
```diff
@@ -1,167 +1,65 @@
...
+        lib.types.int
+    else if kind == "object" then
+      lib.types.attrsOf lib.types.anything
+    else if kind == "string" then
+      lib.types.str
+    else if kind == "boolean" then
+      lib.types.bool
     else
-      throw "unhandled type `${builtins.toJSON type}` (in ${opt})";
+      throw "Unknown type: ${kind}";
```
MattSturgeon, line 44: Doesn't look like any of these have `rawLua` support?

MattSturgeon: (can be a TODO if this isn't a regression)

plugins/lsp/language-servers/rust-analyzer-config.nix
```diff
@@ -1,167 +1,65 @@
...
-          ''
-            ${markdownDescription}
-
-            default:
-            ```nix
-            ${generators.toPretty { } default}
-            ```
-          '';
+      inherit description;
+      type = lib.types.nullOr (mkRustAnalyzerType type);
```
MattSturgeon, line 51: Handle plugin default outside of the description:

```suggestion
      type = lib.types.nullOr (mkRustAnalyzerType type);
      defaultText = helpers.pluginDefaultText {
        inherit pluginDefault;
        default = null;
      };
```

MattSturgeon: I know we didn't do this before, but this PR is a good opportunity to fix that.

MattSturgeon: If you used `helpers.defaultNullOpts.mkNullable'` instead of `lib.mkOption`, then you could pass in `pluginDefault` directly, without needing to use `helpers.pluginDefaultText`.

plugins/lsp/language-servers/rust-analyzer-config.nix
```diff
@@ -1,167 +1,65 @@
...
-  nixOptions = builtins.mapAttrs mkRustAnalyzerOption rustAnalyzerOptions;
-
   nestOpt =
     opt: value:
     let
-      parts = strings.splitString "." opt;
+      parts = lib.strings.splitString "." opt;
     in
-    builtins.foldl' (current: segment: { ${segment} = current; }) value (lists.reverseList parts);
+    builtins.foldl' (current: segment: { ${segment} = current; }) value (lib.lists.reverseList parts);
```
MattSturgeon, line 59: Two observations:

- You could use `foldr` instead of reversing the list
- This is just a re-implementation of [`lib.setAttrByPath`](https://noogle.dev/f/lib/setAttrByPath)

MattSturgeon (changes requested):
I've had a read through the actual implementation now.

It's looking pretty good, just a few minor comments:

flake-modules/updates/rust-analyzer.nix
MattSturgeon: Better yet, use [`mergeAttrsList`](https://noogle.dev/f/lib/mergeAttrsList):

```nix
(lib.mergeAttrsList rustAnalyzerOptions)
```

flake-modules/updates/default.nix
```diff
@@ -0,0 +1,22 @@
+{
+  perSystem =
+    { pkgs, ... }:
+    {
+      apps.generate-files.program = pkgs.writeShellScriptBin "generate-files" ''
```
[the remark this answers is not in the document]
traxys: was not aware of that one!

flake-modules/updates/default.nix
```diff
@@ -0,0 +1,22 @@
...
+        set -e
+
+        repo_root=$(git rev-parse --show-toplevel)
```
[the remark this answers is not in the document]
traxys: Does not matter for assignations, even if there are spaces it won't be split

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+        if lib.head type == "null" then
+          assert lib.length type == 2;
+          let
+            innerType = property // {
+              type = lib.elemAt type 1;
+            };
+            inner = mkRustAnalyzerOptionType nullable "${property_name}-inner" innerType;
+          in
+          # If this assert is false we need to emit a nullOr type
+          assert nullable;
```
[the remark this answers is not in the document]
traxys: https://noogle.dev/f/lib/assertMsg  There is this too

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+    else if
+      lib.elem type [
+        "object"
+        "string"
+        "boolean"
+      ]
+    then
+      { kind = type; }
+    else
+      throw "Unhandled value in ${property_name}: ${lib.generators.toPretty { } property}";
```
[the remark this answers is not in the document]
traxys: well, type is not always defined, for example for `enum` or for `anyOf`, so an unknown case may be the same

flake-modules/updates/rust-analyzer.nix
traxys: I tried to find such a function, but did not suceed! Thanks

plugins/lsp/language-servers/rust-analyzer-config.nix
traxys: Yeah I forgot to add that TODO

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,201 @@
...
+# }
+#
+{
+  lib,
+  rust-analyzer,
+  writeText,
+}:
+let
+  packageJSON = "${rust-analyzer.src}/editors/code/package.json";
+  options = (lib.importJSON packageJSON).contributes.configuration;
```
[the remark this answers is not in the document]
traxys: This does _not_ result in an IFD for `nix flake show` (though it is completly an IFD! But we can allow this IFD only in the CI)

flake-modules/updates/default.nix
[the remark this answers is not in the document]
MattSturgeon: Didn't know that, thanks for the information

flake-modules/updates/rust-analyzer.nix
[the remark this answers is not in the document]
MattSturgeon: If type is not provided then you'll get _"function `mkRustAnalyzerOptionType` called without required argument `type`"_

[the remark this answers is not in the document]
MattSturgeon: Doesn't matter too much, since you're printing the actual args provided though (`property`), I just thought it'd make the error _slightly_ nicer.

generated/rust-analyzer.nix
```diff
@@ -0,0 +1,1990 @@
+# WARNING: DO NOT EDIT
+# This file is generated with packages.<system>.rust-analyzer-options, which is run automatically by CI
```
MattSturgeon, line 2: Looks like you've re-run the script :+1:

flake-modules/updates/rust-analyzer.nix
```diff
@@ -0,0 +1,195 @@
...
+
+  # Extract only the generated properties, removing vscode specific options
+  rustAnalyzerProperties = lib.lists.sublist (generatedStart + 1) (
+    generatedEnd - generatedStart - 1
+  ) options;
+
+  mkRustAnalyzerOptionType =
+    nullable: property_name:
+    {
+      type,
```
MattSturgeon, line 40: You mentioned `type` can be undefined, but you have it as a required argument?

MattSturgeon (approved):
Everything looks addressed, LGTM! Thanks for working on this :heart: 

I flagged one potential issue based on your comment regarding `type` being potentially undefined, but everything works so I'm happy either way.

flake-modules/updates/rust-analyzer.nix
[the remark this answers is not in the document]
traxys: (by undefined I'd meant `null` sorry)

Issue #1794, opened 2024-07-01
Combine `files` and `extraFiles`


MattSturgeon:
Currently we have `files` for building a nixvim config as an extra file and `extraFiles` for directly writing text as an extra file.

There is also no way to add a path as an extra file (without using `readFile` to get the text).

If we make an option with the type `attrsOf (oneOf [ str path fileModuleType ])`, then we could combine the two existing options by making one of them into a rename-alias.

However it may be better to have a new option with a slightly different type, more in line with what nixos & home-manager use:

```nix
type = attrsOf (submodule (
  { name, config, options, ... }:
  {
    options = {
      target = /* default to `name` */;
      source = /* path to a file */;
      text = /* alternatively, file content */;
      module = /* alternatively, a nixvim module to build */;
    };
    config = mkIf (config.source == null) (mkMerge [
      { source = mkIf (config.text != null) (pkgs.writeText "" config.text); }
      { source = mkIf (config.module != null) config.module.plugin; }
    ]);
  }
))
```

This would always have a `.source` attribute that could be used as a file path in other config, along with `.target` that is _where_ the file should be symlinked from when used in xdg/etc.

Heavily based on nixos's `environment.etc.*` option type: https://github.com/NixOS/nixpkgs/blob/7dca15289a1c2990efbe4680f0923ce14139b042/nixos/modules/system/etc/etc.nix#L123-L215

MattSturgeon:
I think we can make `extraFiles` into an authority of a nixvim module's outputs.

Submodules can set put their `content` into `extraFiles` and the host module will then propagate that into its `extraFiles`.

Eventually, at the top-level, `init.lua` is also added to `extraFiles` and the wrapper implementation doesn't have to know about `init.lua` _specifically_.

daniellaing:
Could we also allow directories to be recursively included?
This is useful if someone has lots of extra lua files.

MattSturgeon:
> Could we also allow directories to be recursively included? This is useful if someone has lots of extra lua files.

Does this not work already when setting `source` to a directory?

daniellaing:
Unfortunately it doesn't. Fails with this error:

```
error: Cannot build '/nix/store/g6zhk1l7bc5w6rmmj9l0fbh7w1i0j23z-nvim-config.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/pdlwp52il52kgjkcl36yf03rffbzh166-nvim-config
       Last 1 log lines:
[...]

PR #1807, opened 2024-07-03
modules/top-level: move out of wrappers/modules

MattSturgeon:
Introduces /modules/top-level, where top-level modules common to all wrappers can be defined.

/modules/top-level/default.nix imports /modules, but the reverse is not true. See the comments I added to the head of these files.

Refactored how the `files` option's submodule is omitted from the docs:

- Instead of making a "fake" files option in the docs, we supply `specialArgs.isDocs = true`.
- The _real_ files option can then use `specialArgs.isDocs or false` to make the submodule's modules an optional list.
- This means the specialArg doesn't normally need to be set, except in the docs.

This PR should make it easier to approach #1794.

<details><summary>Old description</summary>
<p>

_This PR started life as a draft for making the `files` option available to submodules too. That was a flawed concept, but the original description is below:_

This makes the option available to submodules too, although there's no actual implementation for that yet.

Added a custom `__docsRecurse` option attr that can be set to false to prevent the docs recursing into submodules.

This is partially cleanup of `wrappers`, but also preparation for #1794 and other refactoring such as removing the IFD caused by `initContent`.

</p>
</details>


MattSturgeon:
I think this doesn't work because the `files` submodule is infinitely recursive.

This explains why it was a top-level module before.

modules like `plugins.telescope` rely on `extraFiles`, so if we want to combine `files` and `extraFiles` we either need to use a different type at the top level, or stop offering `extraFiles` in submodules.

I'm not sure the implementation even works for nested `extraFiles`, so maybe it should be stubbed with an assertion?

MattSturgeon:
Here's a docs render, just to confirm everything still works the same:

<p>


image


</p>
</details>

modules/top-level/files/default.nix
```diff
@@ -3,43 +3,42 @@
...
-            plugin = pkgs.writeTextDir config.path config.content;
-          };
-        }
-      )
+    # Don't include the modules in the docs, as that'd be redundant
+    modules = lib.optionals (!isDocs) [
+      ../../.
+      ./submodule.nix
     ];
+    description = "Nixvim configuration";
```
MattSturgeon, line 24:
```suggestion
    description = "Nixvim module";
```
This would be more accurate, but IDK which terminology is less confusing to users?

MattSturgeon: Nevermind, you don't really assign a _module_ when using a submodule in a config definition, so `config` or `configuration` _is_ more accurate.

traxys (approved):
Be carefull, I know that at least one plugins uses `extraFiles` (tree-sitter I believe?)

MattSturgeon:
> Be carefull, I know that at least one plugins uses `extraFiles` (tree-sitter I believe?)

Yes, I spotted that. I believe that's the only reason `extraFiles` isn't a top-level only option?

_This_ PR makes no changes to `extraFiles`, though, only `files`.

traxys:
> > Be carefull, I know that at least one plugins uses `extraFiles` (tree-sitter I believe?)
> 
> Yes, I spotted that. I believe that's the only reason `extraFiles` isn't a top-level only option?
> 
> _This_ PR makes no changes to `extraFiles`, though, only `files`.

Yeah it was a warning for the next PR

PR #1812, opened 2024-07-05
plugins/efmls-configs: Use a generated file

traxys:
This allows to remove a source of IFD, and reduces the complexity in building the options

### DIFF
--- a/flake-modules/updates/efmls-configs.nix
+++ b/flake-modules/updates/efmls-configs.nix
@@ -1,0 +1,69 @@
+{
+  lib,
+  vimPlugins,
+  writeText,
+  pkgs,
+}:
+let
+  tools = lib.trivial.importJSON "${vimPlugins.efmls-configs-nvim.src}/doc/supported-list.json";
+  languages = lib.attrNames tools;
+
+  toLangTools' = lang: kind: lib.map (lib.getAttr "name") (tools.${lang}.${kind} or [ ]);
+
+  miscLinters = toLangTools' "misc" "linters";
+  miscFormatters = toLangTools' "misc" "formatters";
+
+  sources =
+    (lib.listToAttrs (
+      lib.map (
+        lang:
+        let
+          toLangTools = toLangTools' lang;
+        in
+        {
+          name = lang;
+          value = {
+            linter = {
+              inherit lang;
+              possible = (toLangTools "linters") ++ miscLinters;
+            };
+            formatter = {
+              inherit lang;
+              possible = (toLangTools "formatters") ++ miscFormatters;
+            };
+          };
+        }
+      ) languages
+    ))
+    // {
+      all = {
+        linter = {
+          lang = "all languages";
+          possible = miscLinters;
+        };
+        formatter = {
+          lang = "all languages";
+          possible = miscFormatters;
+        };
+      };
+    };
+
+  inherit (import ../../plugins/lsp/language-servers/efmls-configs-pkgs.nix pkgs) packaged unpackaged;
+
+  toolList = lib.lists.unique (
+    lib.concatLists (
+      lib.map ({ linter, formatter }: linter.possible ++ formatter.possible) (lib.attrValues sources)
+    )
+  );
+
+  unknownTools = lib.filter (tool: !(lib.hasAttr tool packaged || lib.elem tool unpackaged)) toolList;
+in
+assert lib.assertMsg (lib.length unknownTools == 0)
+  "The following tools are neither marked as unpackaged nor as packaged: ${
+    lib.generators.toPretty { } unknownTools
+  }";
+writeText "efmls-configs-sources.nix" (
+  "# WARNING: DO NOT EDIT\n"
+  + "# This file is generated with packages.<system>.efmls-configs-sources, which is run automatically by CI\n"
+  + (lib.generators.toPretty { } sources)
+)
--- a/plugins/lsp/language-servers/efmls-configs-pkgs.nix
+++ b/plugins/lsp/language-servers/efmls-configs-pkgs.nix
@@ -1,0 +1,140 @@
+pkgs: {
+  # efmls-configs tools that have no corresponding nixpkgs package
+  unpackaged = [
+    "blade_formatter"
+    "cspell"
+    "cljstyle"
+    "dartanalyzer"
+    "debride"
+    "deno_fmt"
+    "fecs"
+    "fixjson"
+    "forge_fmt"
+    "gersemi"
+    "js_standard"
+    "pint"
+    "prettier_eslint"
+    "prettier_standard"
+    "redpen"
+    "reek"
+    "rome"
+    "slim_lint"
+    "solhint"
+    "sorbet"
+    "swiftformat"
+    "swiftlint"
+    "xo"
+  ];
+
+  # Mapping from a efmls-configs tool name to the corresponding nixpkgs package
+  packaged = with pkgs; {
+    inherit
+      actionlint
+      alejandra
+      ameba
+      astyle
+      bashate
+      beautysh
+      biome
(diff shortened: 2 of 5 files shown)


plugins/lsp/language-servers/efmls-configs-pkgs.nix
```diff
@@ -0,0 +1,138 @@
+pkgs: {
+  unpackaged = [
```
MattSturgeon, line 2: This would benefit from a comment

plugins/lsp/language-servers/efmls-configs-pkgs.nix
```diff
@@ -0,0 +1,138 @@
...
+    "rome"
+    "slim_lint"
+    "solhint"
+    "sorbet"
+    "swiftformat"
+    "swiftlint"
+    "xo"
+  ];
+
+  packaged = with pkgs; {
```
MattSturgeon, line 28: Same

flake-modules/updates/efmls-configs.nix
```diff
@@ -0,0 +1,74 @@
...
+          possible = miscLinters;
+        };
+        formatter = {
+          lang = "all languages";
+          possible = miscFormatters;
+        };
+      };
+    };
+
+  toolPackagingStatus = import ../../plugins/lsp/language-servers/efmls-configs-pkgs.nix pkgs;
```
MattSturgeon, line 53: What about:

```nix
inherit (import ../../plugins/lsp/language-servers/efmls-configs-pkgs.nix pkgs) packaged unpackaged;
```

?

flake-modules/updates/efmls-configs.nix
```diff
@@ -0,0 +1,74 @@
...
+  lib,
+  vimPlugins,
+  writeText,
+  pkgs,
+}:
+let
+  tools = lib.trivial.importJSON "${vimPlugins.efmls-configs-nvim.src}/doc/supported-list.json";
+  languages = lib.attrNames tools;
+
+  languageTools =
```
MattSturgeon, line 11: `toLanguageTools` would be clearer imo

Or `toLangTools'`, showing it is a more abstract version of the curried function below

flake-modules/updates/efmls-configs.nix
```diff
@@ -0,0 +1,74 @@
...
+
+  miscLinters = languageTools "misc" "linters";
+  miscFormatters = languageTools "misc" "formatters";
+
+  sources =
+    (lib.listToAttrs (
+      lib.map (
+        lang:
+        let
+          langTools = languageTools lang;
```
MattSturgeon, line 23: Again `toLangTools` or similar

flake-modules/updates/efmls-configs.nix
```diff
@@ -0,0 +1,74 @@
...
+  writeText,
+  pkgs,
+}:
+let
+  tools = lib.trivial.importJSON "${vimPlugins.efmls-configs-nvim.src}/doc/supported-list.json";
+  languages = lib.attrNames tools;
+
+  languageTools =
+    lang: kind:
+    lib.map (v: v.name) (if lib.hasAttr kind tools.${lang} then tools.${lang}.${kind} else [ ]);
```
MattSturgeon, line 13:
```nix
lib.map (getAttr "name") (tools.${lang}.${kind} or [ ]);
```

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -161,71 +50,45 @@ in
...
-            // {
-              all = {
-                linter = mkChooseOption "all languages" "linter" miscLinters;
-                formatter = mkChooseOption "all languages" "formatter" miscFormatters;
-              };
-            };
-        };
-        description = "Configuration for each filetype. Use `all` to match any filetype.";
-        default = { };
+    setup = lib.mkOption {
```
MattSturgeon, line 53: Is this a good opportunity to transition to using `settings` (and maybe mkPlugin or at least mkSettingsOption)?

If not, let's add a TODO comment

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -5,144 +5,33 @@
...
       default = [ ];
     };
 
-    toolPackages = attrsets.mapAttrs (
+    toolPackages = lib.attrsets.mapAttrs (
       tool: pkg:
-      mkOption {
-        type = types.package;
+      lib.mkOption {
+        type = lib.types.package;
```
MattSturgeon, line 34: Can we use `helpers.mkPluginPackageOption`?

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -236,43 +99,42 @@ in
...
     in
-    mkIf cfg.enable {
+    lib.mkIf cfg.enable {
       extraPlugins = [ cfg.package ];
 
-      warnings = optional ((length nixvimPkgs.wrong) > 0) ''
+      warnings = lib.optional ((lib.length nixvimPkgs.wrong) > 0) ''
         Nixvim (plugins.efmls-configs): Following tools are not handled by nixvim, please add them to externallyManagedPackages to silence this:
-          ${concatStringsSep " " nixvimPkgs.wrong}
+          ${lib.concatStringsSep " " nixvimPkgs.wrong}
```
MattSturgeon, line 137: Could you add a TODO comment about showing the offending `files` or `definitions`?

(Assuming you don't want to do that in this PR)

MattSturgeon (changes requested):
Looks good, thanks! Just some minor comments

plugins/lsp/language-servers/efmls-configs.nix
traxys: Well, not really, because this plugin is not really configured with a setup option, it plugs directly into nvim-lspconfig & we do some mapping of the options.
So I'd say this is a plugin that does not fit at all in the RFC-42 schema

plugins/lsp/language-servers/efmls-configs.nix
traxys: I don't think I want to do it here, it does seem so easily to reverse the steps to find where those tools have been used, I'd have to refactor a lot of code I think. Adding a todo

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -5,147 +5,31 @@
...
 
-    toolPackages = attrsets.mapAttrs (
-      tool: pkg:
-      mkOption {
-        type = types.package;
-        default = pkg;
-        description = "Package for ${tool}";
-      }
+    toolPackages = lib.attrsets.mapAttrs (
+      tool: pkg: helpers.mkPluginPackageOption "${tool}" pkg
```
MattSturgeon, line 32: My bad, I think we want `helpers.mkPackageOption` instead, since this isn't the _plugin's_ package.

Also `"${tools}"` can just be `tools`, right?

plugins/lsp/language-servers/efmls-configs.nix
MattSturgeon: I might be missing something, but I think doing `plugins.lsp.servers.efm.extraOptions.settings.languages = cfg.settings` would be fairly idiomatic?

We'd just have to make sure the settings option's description talked about that option instead of `require(..).setup`.

plugins/lsp/language-servers/efmls-configs.nix
traxys: That's hasty refactorings for you....

MattSturgeon (approved):
LGTM, thanks!

traxys:
@Mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1813, opened 2024-07-05
plugins/nvim-osc52: hide warnings in tests

MattSturgeon:
#1736 introduced deprecation warnings, which now print when we run the tests. ([Example](https://buildbot.nix-community.org/api/v2/logs/141658/raw_inline)).

We don't care about them in the tests, so just disable all warnings using `mkForce`. This might also hide other warnings we do care about, but I don't see a better approach?

### DIFF
--- a/tests/test-sources/plugins/utils/nvim-osc52.nix
+++ b/tests/test-sources/plugins/utils/nvim-osc52.nix
@@ -1,10 +1,16 @@
+{ pkgs, ... }:
 {
   empty = {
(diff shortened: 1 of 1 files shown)


traxys (approved):
I don't think it's too bad to do this for a deprecated plugin

MattSturgeon:
@mergifyio queue

GaetanLepage (approved):

MattSturgeon:
@mergifyio queuu

PR #1814, opened 2024-07-05
plugin/none-ls: Use a generated file

traxys:
This removes the last use of IFD from the plugins. It also removes the runtime check on none-ls sources, moving it at the file generation


GaetanLepage (approved):
Diff LGTM

GaetanLepage:
Is it me or the CI has not triggered ?

traxys:
It seems like....

flake-modules/updates/none-ls.nix
```diff
@@ -0,0 +1,32 @@
...
+  );
+in
+assert lib.assertMsg (lib.length undeclaredTool == 0)
+  "Undeclared tools: ${lib.generators.toPretty { } undeclaredTool}";
+assert lib.assertMsg (lib.length uselesslyDeclaredTool == 0)
+  "Tool is not supported upstream: ${lib.generators.toPretty { } uselesslyDeclaredTool}";
+writeText "efmls-configs-sources.nix" (
+  "# WARNING: DO NOT EDIT\n"
+  + "# This file is generated with packages.<system>.none-ls-builtins, which is run automatically by CI\n"
+  + (lib.generators.toPretty { } builtinSourceNames)
```
MattSturgeon, line 31: Glad to have these out of the module eval! :grin:

plugins/none-ls/packages.nix
```diff
@@ -0,0 +1,222 @@
...
+    "swiftformat"
+    "swiftlint"
+    "textidote"
+    "textlint"
+    "twigcs"
+    "vacuum"
+  ];
+
+  # builtin sources that don't require a package
+  noPackage = [
```
MattSturgeon, line 45: Also an improvement over the old nullable package _or_ false system!

plugins/none-ls/servers.nix
```diff
@@ -5,273 +5,48 @@
...
-  hasBuiltinPackage =
-    source:
-    if builtins.hasAttr source builtinPackages then
-      !(builtins.isBool builtinPackages.${source})
-    else
-      true;
-
-  builtinPackage = source: builtinPackages.${source} or null;
+  # Does this builitin require a package ?
+  hasBuiltinPackage = source: lib.hasAttr source packaged || lib.elem source unpackaged;
```
MattSturgeon, line 14: The name is a little confusing. Would `builtinNeedsPackage` or `requiresPackage` (etc) be better?

MattSturgeon (approved):
LGTM! I left one comment regarding a name that I think came from the old impl, but if you don't care I won't block on a nit.

plugins/none-ls/servers.nix
traxys: Definitively! I tried finding a better name, but didn't have enough inspiration

traxys:
@Mergifyio queue

traxys:
@Mergifyio queue

PR #1816, opened 2024-07-05
github/update: refactor to use create-pull-request action

MattSturgeon:
Refactor the update CI workflow to update the lock file and generated files _and then_ create the PR in a subsequent step.

I added an action summary, which will link to the created/updated PR and describe what was done on the Action summary page.

I updated the generate script with a `--commit` option, when used the changes are committed, using a commit message that summaries the changes.

The created/updated PR will use the commit messages in its body to match what is done by DeterminateSystems/update-flake-lock.

Instead of a Personal Access Token, an SSH Deploy Key is now used (@GaetanLepage can you add this to the repo please?).
This means the PR can be created as `github-actions[bot]` and `on: push` events will still trigger.\
**NOTE:** `on: pull_request` won't trigger, but this shouldn't affect us.

<details><summary>Old description</summary>
<p>

I haven't tested this yet (because testing CI workflows on a fork is painful).

@GaetanLepage We should also consider using SSH Deploy Keys instead of a PAT to work-around bot account PRs not triggering CI.\
See https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

I haven't included a PR body, because creating one with details of the commits is non-trivial. This isn't that important, but I left a TODO comment anyway.

I can remove the TODO if preferred.

</p>
</details>


.github/workflows/update.yml
```diff
@@ -25,35 +25,76 @@ jobs:
...
+        run: nix run .#generate-files
+
+      - name: Create Pull Request
+        id: pr
+        uses: peter-evans/create-pull-request@v6
         with:
-          pr-title: "Update flake.lock (${{ matrix.branch }})"
+          # TODO: use an SSH deploy key instead
+          # See https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
+          # and https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-using-ssh-deploy-keys
```
GaetanLepage, line 61: This one we could leave for now.

.github/workflows/update.yml
```diff
@@ -25,35 +25,76 @@ jobs:
...
+          delete-branch: true
+          add-paths: |
+            generated/**
+          commit-message: |
+            generated: Update files
+          title: |
+            [${{ matrix.branch }}] Update flake.lock & generated files
+          # TODO: include commit details in the PR body
+          # Can also use body-path if it's easier to generate a file
+          # body: |
```
GaetanLepage, line 75: We can also leave it for another PR.

GaetanLepage (approved):

.github/workflows/update.yml
MattSturgeon: I'm not going to test the PR works until over the weekend, so I don't mind including it in this PR. Doesn't matter too much either way though, don't mind waiting until it's merged.

.github/workflows/update.yml
MattSturgeon: TBH, I'm not sure it's worth the effort. Should I just remove the comments?

GaetanLepage: I don't know. The action we currently use does it automatically right ?
But I agree, not a huge deal.

MattSturgeon: It shouldn't be too hard to print the flake.lock commit message and then also the files changed by the auto-gen script, I'll try and find motivation to come up with something.

MattSturgeon: We can get the commit body without title using:
```shell
git show --no-patch --format=%b
```
(or `%B` for full body with title)

As for the auto-generated files, we can get the unstaged changes with:
```shell
git diff --name-status
# or
git diff --name-only
```

Although this will only list _tracked_ files with unstaged changes.

To get any file that `git add -A` would add, we could use the `--dry-run` option, e.g:
```shell
$ git add -A --dry-run ./generated
add 'generated/efmls-configs.nix'
add 'generated/foo'
```

GaetanLepage: Ok ! That looks promising !
No need to go crazy about it either. I guess it's a cool bonus.

.github/workflows/update.yml
```diff
@@ -25,35 +25,96 @@ jobs:
...
+      contents: write
+      pull-requests: write
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
-          ref: "${{ matrix.branch }}"
+          ssh-key: ${{ secrets.UPDATE_CI_KEY }}
```
MattSturgeon, line 41: Repo secret `UPDATE_CI_KEY` needs to contain the SSH private key.

The corresponding public key should be added as a "Deploy Key" with **write** access.

See [instructions here](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-using-ssh-deploy-keys).

MattSturgeon:
I've now tested this on my fork.

See https://github.com/MattSturgeon/nixvim/actions/runs/9862728221 and https://github.com/MattSturgeon/nixvim/actions/runs/9862812497 for example runs.

[This PR](https://github.com/MattSturgeon/nixvim/pull/3) was created/updated by those runs.

.github/workflows/update.yml
GaetanLepage: Can you rename it to `SSH_DEPLOY_KEY` ?

GaetanLepage (approved):
LGTM !

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1818, opened 2024-07-05
modules/output: Remove the initContent option

traxys:
This (internal) option introduces IFD, and can be substituted internally with the `initPath` option easily.
If the content is required somewhere users can use readFile on the initPath, though it will result in an IFD in their project.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -61,13 +61,6 @@ with lib;
       visible = false;
     };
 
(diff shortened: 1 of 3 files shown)


GaetanLepage (approved):
Perfect !

MattSturgeon (approved):
❤️

traxys:
@Mergifyio queue

PR #1819, opened 2024-07-05
modules/context: init

MattSturgeon:
### Summary
Added `isTopLevel` and `isDocs` internal options. This replaces `specialArgs.isDocs` that was added in #1807.

### specialArgs vs options
We should only use `specialArgs` when absolutely necessary, a module or `_module.args` is preferred, but those aren't available until `config` is evaluated so they can't be used for `imports`.

The main problem with `specialArgs` is it makes our modules less portable, but it also means the arg can't be overridden with `mkForce` (etc).

Luckily that shouldn't be an issue for these context flags.

### What is `isTopLevel` even for??
The `isTopLevel` option is currently unused; I can drop it from this PR if preferred.

It is _intended_ to allow `files` or `extraFiles` to have a different _type_ when at the top-level, since we can't have a submodule option that recursively includes itself in its modules.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -58,7 +58,10 @@ let
       ) opt.declarations;
     };
 
-  modules = [ ../modules/top-level ];
+  modules = [
+    ../modules/top-level
+    { isDocs = true; }
+  ];
 
   hmOptions = builtins.removeAttrs (lib.evalModules {
     modules = [ (import ../wrappers/modules/hm.nix { inherit lib; }) ];
@@ -73,7 +76,6 @@ rec {
           specialArgs = {
             inherit helpers;
             defaultPkgs = pkgsDoc;
-            isDocs = true;
           };
         })
         options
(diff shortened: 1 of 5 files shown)


GaetanLepage:
Thanks Matt for this !

> ### What is `isTopLevel` even for??
> 
> The `isTopLevel` option is currently unused; I can drop it from this PR if preferred.
> 
> It is _intended_ to allow `files` or `extraFiles` to have a different _type_ when at the top-level, since we can't have a submodule option that recursively includes itself in its modules.

If this option is not used, I would rather not have it.
Generally, I think we should only introduce new features when they are actually needed.

MattSturgeon:
> > 
> 
> If this option is not used, I would rather not have it. Generally, I think we should only introduce new features when they are actually needed.

Sure, done.

I'll stash it in a local branch and grab it if/when I need it.

GaetanLepage (approved):
LGTM

MattSturgeon:
@mergifyio queue

PR #1823, opened 2024-07-06
lib/neovim-plugin: allow disabling `installPackage`

MattSturgeon:
Some plugins may wish to handle package installation themselves, so make installing `cfg.package` optional.

I used a "real" branch instead of `mkIf` because we don't know _why_ the plugin doesn't want us to install it - it's possible `cfg.package` shouldn't be evaluated.

This is needed for #1801

@GaetanLepage sorry to step on your toes again re: the the mkPlugin refactor!

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -47,6 +47,7 @@ with lib;
       extraPlugins ? [ ],
       extraPackages ? [ ],
       callSetup ? true,
+      installPackage ? true,
     }:
     let
       namespace = if isColorscheme then "colorschemes" else "plugins";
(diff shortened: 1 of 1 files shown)


lib/neovim-plugin.nix
```diff
@@ -107,7 +108,7 @@ with lib;
         in
         mkIf cfg.enable (mkMerge [
           {
-            extraPlugins = [ cfg.package ] ++ extraPlugins;
+            extraPlugins = optional installPackage cfg.package ++ extraPlugins;
```
GaetanLepage, line 111:
```suggestion
            extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
```

Probably useless here, but I would rather have them (or swap the order and have the `optional` in second).

MattSturgeon: Yeah, `++` and `//` have lower priority than the function call, hence why we have to do things like `(args // { override = "hi"; })` elsewhere.

I agree it's easy to forget though and your suggestions are more readable.

MattSturgeon: Done

GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

PR #1826, opened 2024-07-06
modules/output: fix extraLuaPackages

stasjok:
It seams `extraLuaPackages` doesn't work or am I missing something?

This PR ensures `extraLuaPackages` are made available on the runtimepath.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -85,6 +85,7 @@ with lib;
         {
           inherit (config)
             extraPython3Packages
+            extraLuaPackages
             viAlias
             vimAlias
             withRuby
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -1,0 +1,7 @@
+{
+  extraLuaPackages = {
+    extraLuaPackages = ps: [ ps.jsregexp ];
+    # Make sure jsregexp is in LUA_PATH
+    extraConfigLua = ''require("jsregexp")'';
+  };
+}


modules/top-level/output.nix
```diff
@@ -85,6 +85,7 @@ with lib;
         {
           inherit (config)
             extraPython3Packages
+            extraLuaPackages
```
MattSturgeon, line 88: Can you explain the change and why it is specific to top-level outputs (rather than /modules/outputs.nix)?

stasjok: I've added a test, but I can't run it on my PC. Too much RAM is required. Is there a way to run only a single test? I want to check if it fails before this change and passes after.

MattSturgeon: > Is there a way to run only a single test?

Not that I'm aware of. That'd be nice though, my laptop takes around an hour to run the basic tests.

Letting buildbot do the work is fine if your hardware can't handle it.

MattSturgeon:
Doing this globally in `outputs` probably isn't the right approach. Do you know what plugin(s) this fixed?

stasjok:
> Doing this globally in `outputs` probably isn't the right approach. Do you know what plugin(s) this fixed?

Why is that? It doesn't have anything to do with plugins, it's about global option [extraLuaPackages](https://nix-community.github.io/nixvim/NeovimOptions/index.html#extraluapackages). It should be propagated to finalPackage. Similar to `extraPython3Packages`.

But if you want to check plugins, you can check any plugin that modify `extraLuaPackages` option, e.g. luasnip. Something like:

```nix
  programs.nixvim = {
    enable = true;
    plugins.luasnip.enable = true;
  };
```

Then `:checkhealth luasnip`.

modules/top-level/output.nix
stasjok: > > Is there a way to run only a single test?
> 
> Not that I'm aware of. That'd be nice though, my laptop takes around an hour to run the basic tests.

I found it.

```
nix build .#checks.x86_64-linux.tests.passthru.entries.modules-output-extraLuaPackages
```

will run only this test. So I checked that it fails before the fix (`require` returns `module 'jsregexp' not found` error).

MattSturgeon: Thanks for the tip using `tests.passthru`, that'll come in handy!



> Can you explain the change and why it is specific to top-level outputs (rather than /modules/outputs.nix)?

I've edited the description to clarify the goal of the PR, please comment/edit further if my changes are inaccurate.

MattSturgeon (approved):
> > Doing this globally in `outputs` probably isn't the right approach. Do you know what plugin(s) this fixed?
> 
> Why is that? It doesn't have anything to do with plugins, it's about global option [extraLuaPackages](https://nix-community.github.io/nixvim/NeovimOptions/index.html#extraluapackages). It should be propagated to finalPackage. Similar to `extraPython3Packages`.

Sorry, I was trying to review this PR without giving it my full attention.

Your right, if we want `extraLuaPackages` to be available on the rtp then this PR fixes that.

LGTM, but I'd like an second opinion from someone more familiar with this part of the project.

traxys (approved):
Well it seems inline with the extraPythonPackages, seems to be good

MattSturgeon:
@mergifyio queue

GaetanLepage (approved):

PR #1827, opened 2024-07-07
flake: disable `allow-import-from-derivation`

MattSturgeon:
Unfortunately this will print a confirmation when first using the flake:

```
do you want to allow configuration setting 'allow-import-from-derivation' to be set to 'false' (y/N)?
```

This is no different to the cachix configs though.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -71,6 +71,7 @@
     extra-trusted-public-keys = [
       "nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs="
     ];
+    allow-import-from-derivation = false;
   };
 
   outputs =
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -23,19 +23,19 @@
       mkConfigAssertions = name: value: [
         {
           assertion = lib.hasInfix "extraConfigLuaPre1" value;
-          message = "Configuration file ${name} does not contain extraConfigLuaPre.";
+          message = "Configuration file ${name} should contain extraConfigLuaPre.";
         }
         {
           assertion = lib.hasInfix "extraConfigLua2" value;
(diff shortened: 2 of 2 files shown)


MattSturgeon:
Looks like a number of tests rely on IFD.

If we can find a way to do them without IFD, I think we should.

I think some of the packages used by the update CI are being included in `checks`? We probably _don't_ want that?

traxys:
Yeah I agree we probably don't want that, we may need to move them to a different section than packages

GaetanLepage (changes requested):
- What do you mean by "the first time using the flake" ? The first time running `nix run` with the nixvim flake as an input of your flake ?
- `nocheckPackages` -> `noCheckPackages` ?

MattSturgeon:
> * What do you mean by "the first time using the flake" ? The first time running `nix run` with the nixvim flake as an input of your flake ?

The first time running any `nix` command with this flake checked out locally.

I don't _think_ our flake's nixConfig is inherited by downstream flakes (haven't tested, could be wrong).

>     * `nocheckPackages` -> `noCheckPackages` ?

Sure, no strong opinion.

---

Just for the record, this PR doesn't currently do what it intends to. Passing the `--allow-import-from-derivation` flag to `nix` commands _does not_ override the flake's nix config, so the update CI will not work if we merge this in its current state.

MattSturgeon:
Maybe we can configure nix in GitHub CI to not allow the flake to set nixConfig?

MattSturgeon:
Since #1980, we're now able to disable IFD in the flake. I've therefore rebased this and removed the outdated implementation.

Just need to fix/remove some of the new tests added by https://github.com/nix-community/nixvim/commit/299d0406bb339283969907d1f171f0eec7d84c43

Tests that need to read the content of files will have to be custom test derivations so that we can read the files during the build/check phase rather than importing the file into the eval.

GaetanLepage (approved):
Ok let's go !

MattSturgeon:
@mergifyio queue

PR #1828, opened 2024-07-07
modules: refactor `extraFiles`

MattSturgeon:
### Summary
Refactored `extraFiles` to use a file-type submodule.

This submodule has the following options:
- `source`: the file path to use
- `text`: (optional) text to write and (potentially) override `source`
- `target`: (default attr name) where the file should be installed
- `enable` (default true) can be used to disable a file added by another module

Users should now assign text to a `text` attribute, however they could also assign a file path to a `source` attribute instead.

### Type coercion
The type conversion is a non-breaking change, as I was able to deprecate the old type using coercion.

I've introduced a tweaked version of `types.coerceTo` (`helpers.transitionType`) that a) only shows the new type's description and b) prints a warning when the old type is used in a definition.

### Tests
I've added an additional check derivation that tests extra files exist in the correct location within `config.filesPlugin`.

### Breaking changes
The `files` option's submodule has an (internal) `plugin` option, which is a derivation containing the resulting file.

Previously this was built using `pkgs.writeTextDir` (i.e. the file was nested in its dir path), however now the file is written directly to the derivation location, using `pkgs.writeText`.

This is done so that `files` can be easily added directly to `extraFiles`, another step towards #1794.

### text vs sources
This means we now pass all files to hm/nixos as filepath "sources" instead of text. This enabled me to use `helpers.writeLua` to format submodule files like we do with `initPath`.

#### init in extraFiles?
It'd be nice if we could say "everything is in `extraFiles`" (maybe we'd look to rename it at some point?)

To that end, I think it should be a goal to also include `init.lua` in `extraFiles` instead of handling it separately. IDK if there's any complications to this idea though?

#### Print extraFiles
Even if that's not practical, I'd still like to make a way for nixvim-print-init to be able to print extra files too.

#### Nested extraFiles
I believe that currently `extraFiles` nested in a `files` submodule have no effect, i.e. they do not propagate to the top-level `extraFiles`. I haven't tested this though.

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -1,0 +1,82 @@
+{
+  lib,
+  helpers,
+  pkgs,
+  ...
+}:
+let
+  fileType = lib.types.submodule (
+    {
+      name,
+      config,
+      options,
+      ...
+    }:
+    {
+      options = {
+        enable = lib.mkOption {
+          type = lib.types.bool;
+          default = true;
+          description = ''
+            Whether this file should be generated.
+            This option allows specific files to be disabled.
+          '';
+        };
+
+        target = lib.mkOption {
+          type = lib.types.str;
+          defaultText = lib.literalMD "the attribute name";
+          description = ''
+            Name of symlink, relative to nvim config.
+          '';
+        };
+
+        text = lib.mkOption {
+          type = with lib.types; nullOr lines;
+          default = null;
+          description = "Text of the file.";
+        };
+
+        source = lib.mkOption {
+          type = lib.types.path;
+          description = "Path of the source file.";
+        };
+      };
+
+      config =
+        let
+          derivationName = "nvim-" + lib.replaceStrings [ "/" ] [ "-" ] name;
+        in
+        {
+          target = lib.mkDefault name;
+          source = lib.mkIf (config.text != null) (
+            # mkDerivedConfig uses the option's priority, and calls our function with the option's value.
+            # This means our `source` definition has the same priority as `text`.
+            lib.mkDerivedConfig options.text (pkgs.writeText derivationName)
+          );
+        };
+    }
+  );
+
+  # TODO: Added 2024-07-07, remove after 24.11
+  # Before we had a fileType, we used types.str.
+  coercedFileType = helpers.transitionType lib.types.str (text: { inherit text; }) fileType;
+in
+{
+  options = {
+    extraFiles = lib.mkOption {
+      type = lib.types.attrsOf coercedFileType;
+      description = "Extra files to add to the runtime path";
+      default = { };
+      example = lib.literalExpression ''
+        {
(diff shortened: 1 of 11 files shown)


modules/top-level/files/default.nix
```diff
@@ -59,12 +60,37 @@ in
...
+        ${lib.concatMapStringsSep "\n" (
+          { target, source, ... }:
+          lib.escapeShellArgs [
+            "makeEntry"
+            # Force local source paths to be added to the store
+            "${source}"
+            target
+          ]
+        ) extraFiles}
+      '';
```
MattSturgeon, line 94: This implementation is taken from here: https://github.com/NixOS/nixpkgs/blob/7dca15289a1c2990efbe4680f0923ce14139b042/nixos/modules/system/etc/etc.nix#L275-L293

Should probably have a reference/attribution comment?

modules/top-level/files/default.nix
```diff
@@ -59,12 +60,37 @@ in
...
+      );
+
       # A directory with all the files in it
-      filesPlugin = pkgs.buildEnv {
-        name = "nixvim-config";
-        paths =
-          (lib.mapAttrsToList (_: file: file.plugin) files)
-          ++ (lib.mapAttrsToList pkgs.writeTextDir config.extraFiles);
-      };
+      filesPlugin = pkgs.runCommandLocal "nvim-config" { } ''
```
MattSturgeon, line 74: Not convinced `filesPlugin` is the best name for this.

modules/files.nix
```diff
@@ -0,0 +1,103 @@
...
+      # Override merge to print a warning when oldType is used
+      merge =
+        opt: defs:
+        let
+          coerceVal =
+            val:
+            if oldType.check val then
+              lib.warn ''
+                Passing a ${oldType.description} for ${lib.showOption opt} is deprecated, use ${newType.description} instead. Definitions: ${lib.options.showDefs defs}
+              '' (fn val)
```
MattSturgeon, line 79: I don't see this warning when running:
```
nix build .#checks.x86_64-linux.tests.passthru.entries.modules-extra-files-query
```
So I suspect something isn't working...

MattSturgeon: Resolved: overriding attributes on the type returned by `types.coercedTo` wasn't working, but copying its source and making the appropriate changes does work.

modules/top-level/files/default.nix
MattSturgeon: I don't want to rename it in this PR though.

GaetanLepage:
The very clear PR description helped a lot with that !

traxys (approved):
We should investigate if `extraFiles` in  `files` work, and either fix that or remove the extraFiles option in non-top-level modules, but let's do that in a followup PR

MattSturgeon:
@mergifyio queue

PR #1830, opened 2024-07-07
plugins/glow: init

getchoo:
Adds support for https://github.com/ellisonleao/glow.nvim

### DIFF
--- a/plugins/languages/markdown/glow.nix
+++ b/plugins/languages/markdown/glow.nix
@@ -1,0 +1,87 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "glow";
+  originalName = "glow.nvim";
+  defaultPackage = pkgs.vimPlugins.glow-nvim;
+
+  maintainers = [ lib.maintainers.getchoo ];
+
+  settingsOptions = {
+    glow_path = helpers.defaultNullOpts.mkStr (helpers.mkRaw "vim.fn.exepath('glow')") ''
+      Path to `glow` binary.
+
+      If null or `""`, `glow` in your `$PATH` with be used if available.
+
+      Using `glowPackage` is the recommended way to make `glow` available in your `$PATH`.
+    '';
+
+    install_path = helpers.defaultNullOpts.mkStr "~/.local/bin" ''
+      Path for installing `glow` binary if one is not found at `glow_path` or in your `$PATH`.
+
+      Consider using `glowPackage` instead.
+    '';
+
+    border = helpers.defaultNullOpts.mkEnumFirstDefault [
+      "shadow"
+      "none"
+      "double"
+      "rounded"
+      "solid"
+      "single"
+    ] "Style of the floating window's border.";
+
+    style = helpers.defaultNullOpts.mkEnum [
+      "dark"
+      "light"
+    ] (helpers.mkRaw "vim.opt.background") "Glow style.";
+
+    pager = helpers.defaultNullOpts.mkBool false ''
+      Display output in a pager style.
+    '';
+
+    width = helpers.defaultNullOpts.mkInt 100 ''
+      Width of the floating window.
+    '';
+
+    height = helpers.defaultNullOpts.mkInt 100 ''
+      Height of the floating window.
+    '';
+
+    width_ratio = helpers.defaultNullOpts.mkNum 0.7 ''
+      Maximum width of the floating window relative to the window size.
+    '';
+
+    height_ratio = helpers.defaultNullOpts.mkNum 0.7 ''
+      Maximum height of the floating window relative to the window size.
+    '';
+  };
+
+  settingsExample = {
+    border = "shadow";
+    style = "dark";
+    pager = false;
+    width = 80;
+    height = 100;
+    width_ratio = 0.7;
+    height_ratio = 0.7;
+  };
+
+  extraOptions = {
+    glowPackage = helpers.mkPackageOption {
+      description = ''
+        Which package to use for `glow` in your `$PATH`.
+        Set to `null` to disable its automatic installation.
+      '';
+      default = pkgs.glow;
+      defaultText = lib.literalExpression "pkgs.glow";
+    };
+  };
+
+  extraConfig = cfg: { extraPackages = [ cfg.glowPackage ]; };
+}
--- a/tests/test-sources/plugins/languages/markdown/glow.nix
+++ b/tests/test-sources/plugins/languages/markdown/glow.nix
@@ -1,0 +1,26 @@
+{ pkgs, ... }:
+{
+  empty = {
+    plugins.glow.enable = true;
+  };
+
+  defaults = {
+    plugins.glow = {
+      enable = true;
+
+      glowPackage = pkgs.glow;
+
+      settings = {
+        glow_path.__raw = "vim.fn.exepath('glow')";
+        install_path = "~/.local/bin";
+        border = "shadow";
+        style = "dark";
+        pager = false;
+        width = 80;
+        height = 100;
+        width_ratio = 0.7;
+        height_ratio = 0.7;
+      };
+    };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -68,6 +68,7 @@
     ./languages/ledger.nix
     ./languages/lint.nix
     ./languages/ltex-extra.nix
+    ./languages/markdown/glow.nix
     ./languages/markdown/markdown-preview.nix
     ./languages/markdown/preview.nix
     ./languages/nix.nix


tests/test-sources/plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.glow.enable = true;
+  };
```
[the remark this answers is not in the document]
MattSturgeon: i.e. a test-case with all settings explicitly set to the plugin's default

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,52 @@
...
+
+  settingsOptions = {
+    glow_path = helpers.mkNullOrStr' {
+      description = ''
+        Path to `glow` binary
+
+        If set to an empty string or `null`, `glow` will be automatically 
+        installed to `plugins.glow.settings.install_path`
+      '';
+      default = lib.getExe pkgs.glow;
```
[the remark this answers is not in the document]
MattSturgeon: This is probably what you want:
```suggestion
      pluginDefault = lib.literalExpression "lib.getExe pkgs.glow";
```

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    style =
+      helpers.defaultNullOpts.mkEnumFirstDefault
+        [
+          null
+          "dark"
+          "light"
+        ]
+
```
MattSturgeon, line 35:
```suggestion
```

Extra linebreak

plugins/languages/markdown/glow.nix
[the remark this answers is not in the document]
GaetanLepage: Well we have two choices here:
- Set `lib.getExe pkgs.glow` as the default value for this option
- Create a `plugins.glow.glowPackage` option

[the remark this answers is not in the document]
MattSturgeon: Yes, I read it as `pkgs.glow` is the plugin default, but of course how would a plugin know about nix packages by default?

I think something like:

```nix
settingsOptions = {
  glow_path = helpers.defaultNullOpts.mkStr "" ''
    Path to `glow` binary.

    Consider using the `glowPackage` option instead.
  '';
};

extraOptions = {
  glowPackage = helpers.mkPackageOption {
    name = "glow";
    default = pkgs.glow;
  };
};

extraConfig = cfg: {
  plugins.glow.settings.glow_path = lib.mkIf (cfg.glowPackage != null) (lib.getExe cfg.glowPackage);
};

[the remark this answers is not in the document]
Comment (getchoo) [reply plugins/languages/markdown/glow.nix:23]: i had actually tried making a package option originally but hadn't come across extraOptions and extraConfig yet. thanks!

File: plugins/languages/markdown/glow.nix [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+          "shadow"
+          "none"
+          "double"
+          "rounded"
+          "solid"
+          "single"
+        ]
+        ''
+          Style of the floating window's border
+        '';
```
MattSturgeon, line 40: I think nixfmt may format this nicer if you use a `"` string.

We do _normally_ prefer `''`-indented strings for option descriptions, but in this case I think `"` would be better.

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,96 @@
...
+    style =
+      helpers.defaultNullOpts.mkEnum
+        [
+          "dark"
+          "light"
+        ]
+        (lib.literalExpression "vim.opt.background")
+        ''
+          Glow style
+        '';
```
MattSturgeon, line 51: Same

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,96 @@
...
+    width = 80;
+    height = 100;
+    width_ratio = 0.7;
+    height_ratio = 0.7;
+  };
+
+  extraOptions = {
+    glowPackage = helpers.mkPackageOption {
+      name = "glow";
+      default = pkgs.glow;
```
MattSturgeon, line 87: Optional, but adding `defaultText = lib.literalExpression "pkgs.glow"` would improve the docs slightly.

We don't currently do this much (at all?), but I think it'd be nice to going forward.

MattSturgeon (commented):
Looking good, just a few minor comments

plugins/languages/markdown/glow.nix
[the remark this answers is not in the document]
MattSturgeon: > I thought that those helpers would enforce `default = null`.

Just for the record, it is only `helpers.defaultNullOpts` that enforce `default = null`.

plugins/languages/markdown/glow.nix
getchoo: good catch. i was just thinking about it when looking at the `.package` options in the docs 😆

tests/test-sources/plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,21 @@
...
+
+  defaults = {
+    plugins.glow = {
+      enable = true;
+
+      settings = {
```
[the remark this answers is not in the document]
getchoo: i'll assume yes on this for now

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,86 @@
...
+      defaultText = lib.literalExpression "pkgs.glow";
+    };
+  };
+
+  extraConfig = cfg: {
+    plugins.glow.settings = {
+      glow_path = lib.mkIf (cfg.glowPackage != null) (lib.getExe cfg.glowPackage);
+    };
```
GaetanLepage, line 84: Or we can simply `extraPackages = [cfg.glowPackage]`. The plugin will be able to find it just fine.

MattSturgeon: If so, we could reword the `glow_path` desc to something like:

"If left set to null, `config.glow.glowPackage` will be used via PATH"

(Could be worded better)

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,86 @@
...
+      "double"
+      "rounded"
+      "solid"
+      "single"
+    ] "Style of the floating window's border";
+
+    style = helpers.defaultNullOpts.mkEnum [
+      "dark"
+      "light"
+    ] (lib.literalExpression "vim.opt.background") "Glow style";
```
MattSturgeon, line 40:
```suggestion
    ] (helpers.mkRaw "vim.opt.background") "Glow style";
```

GaetanLepage (changes requested):
Please add a period (`.`) at the end of each option documentation sentence.

tests/test-sources/plugins/languages/markdown/glow.nix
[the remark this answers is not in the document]
MattSturgeon: Once https://github.com/nix-community/nixvim/pull/1830#discussion_r1668690255 is resolved, we won't need to do anything special with `glowPackage` here; it too can be set to it's default `pkgs.glow`.

Also, `glow_path` should be `.__raw = "vim.fn.exepath('glow')"`

plugins/languages/markdown/glow.nix
getchoo: that seems to trigger [this](https://github.com/nix-community/nixvim/blob/97fa47376b73d774b081c8e5dd54fcfd0ad278cb/lib/options.nix#L223) assertion. any good way around it?

tests/test-sources/plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,26 @@
...
+  };
+
+  defaults = {
+    plugins.glow = {
+      enable = true;
+
+      glowPackage = pkgs.glow;
+
+      settings = {
+        glow_path = helpers.mkRaw "vim.fn.exepath('glow')";
```
GaetanLepage, line 14:
```suggestion
        glow_path.__raw = "vim.fn.exepath('glow')";
```

plugins/languages/markdown/glow.nix
MattSturgeon: Looks like you've found a bug! 🪲

The assertion doesn't handle raw types correctly.

Just double checking, the type is definitely an enum right?

getchoo: > Just double checking, the type is definitely an enum right?

yep! you can find the definition of it upstream [here](https://github.com/ellisonleao/glow.nvim/blob/238070a686c1da3bccccf1079700eb4b5e19aea4/lua/glow.lua#L14) and `vim.opt.background` being used as a default [here](https://github.com/ellisonleao/glow.nvim/blob/238070a686c1da3bccccf1079700eb4b5e19aea4/lua/glow.lua#L32). this seems accurate to what's described as recognized values of `background` in [neovim's documentation](https://neovim.io/doc/user/options.html#'background')

tests/test-sources/plugins/languages/markdown/glow.nix
getchoo: are these not equivalent? or is helpers.mkRaw not preferred?

MattSturgeon: They are equivalent, so it's a style nit.

We usually use `.__raw =` when it can be set as a nested attr like this.

plugins/languages/markdown/glow.nix
MattSturgeon: #1837 should resolve the issue

tests/test-sources/plugins/languages/markdown/glow.nix
traxys: (Yeah I'd say it's a nit because I would personally prefer the helpers.mkRaw, as I find it less confusing to read)

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,87 @@
...
+      description = ''
+        Which package to use for `glow` in your $PATH.
+        Set to `null` to disable its automatic installation.
+      '';
+      default = pkgs.glow;
+      defaultText = lib.literalExpression "pkgs.glow";
+    };
+  };
+
+  extraConfig = cfg: { extraPackages = lib.optional (cfg.glowPackage != null) cfg.glowPackage; };
```
MattSturgeon, line 86:
```suggestion
  extraConfig = cfg: { extraPackages = [ cfg.glowPackage; ]; };
```

1. In config defs, `mkIf` is generally preferred to a "real" branch (i.e. `if then else` or wrappers like `optional`)
2. We don't need any conditions here, because our `extraPackages` type filters null items anyway.

plugins/languages/markdown/glow.nix
```diff
@@ -0,0 +1,87 @@
...
+
+  maintainers = [ lib.maintainers.getchoo ];
+
+  settingsOptions = {
+    glow_path = helpers.defaultNullOpts.mkStr (helpers.mkRaw "vim.fn.exepath('glow')") ''
+      Path to `glow` binary.
+
+      If null or `""`, `glow` in your $PATH with be used if available.
+
+      Consider using `glowPackage` instead.
```
MattSturgeon, line 21:
```suggestion
      Using `plugins.glow.glowPackage` is the recommended way to make `glow` available on your `PATH`.
```

My suggestion can be improved a lot, but I think _something_ like this would be better?

GaetanLepage: That's good

GaetanLepage (approved):
LGTM !

traxys (commented):
Did not see any glaring issues, I trust your reviews

GaetanLepage:
@Mergifyio queue

PR #1831, opened 2024-07-08
wrappers: make `_shared` return a module

### LINKED ISSUE #1220: [BUG] docs wrongly claim `helpers` is accessible anywhere
| Plugin       | `N/A`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

The docs state that I should be able to access the `helpers` attribute as a module parameter, however `nix flake check` states it is missing.

Not very minimal I'm afraid, but [this branch](https://github.com/MattSturgeon/nix_config/tree/bug/nixvim-helpers) demonstrates the issue when running `nix flake check .`

_(I'll try and make a proper MRE if this proves necessary)_

## Speculation

My first thought was this is flake related, however I would have expected this to have come up already if so. My next guess is is could be related to [the way I add](https://github.com/MattSturgeon/nix_config/blob/f3696a3eaf2d9e3700949e8d9b5af2ef82f33d53/lib/system.nix#L16-L22) my flake inputs to the nixos/home-manager `extraArgs`?

MattSturgeon:
## Make `_shared.nix` return a module
We can consolidate some duplication by propagating config files directly in a shared module, we just need to know the option name and the name of the init file.

## Make `helpers` available via `lib` option
**NOTE: this change has been dropped from this PR!**

<p>

Currently we make helpers available via `config.nixvim.helpers`, however NixOS, nix-darwin and home-manager all provide a `config.lib` option (type `attrsOf attrs`).

This `lib` option is intended to be used by modules to define their custom functions in a global scope. It is **unrelated** to the `lib` available via module args.

I've used `mkRenameOptionModule` to rename `nixvim.helpers` to `lib.nixvim`. This prints a warning when `config.nixvim.helpers` is evaluated:
```
trace: Obsolete option `nixvim.helpers' is used. It was renamed to `lib.nixvim'.
```

</p>
</details> 

I've also updated the docs. Previously they said that `helpers` is available as a module arg in all wrappers, however this was only true for the standalone build. Fixes #1220.

Maybe we should fix the nixos-24.05 docs too.

## Bootstrap "helpers" directly
In order to simplify the wrappers some more, and reduce the number of arguments needed to be passed in, we can bootstrap `helpers` directly from `../lib/helpers.nix` instead of using a `getHelpers` function passed in from the flake.

We do this in the shared module returned by `_shared.nix`.

## Remove `getHelpers` function
It was now only used by the docs, so we may as well remove it.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -1,7 +1,7 @@
 {
-  lib,
   pkgs,
-  _nixvimTests,
+  lib ? pkgs.lib,
+  _nixvimTests ? false,
   ...
 }:
 let
--- a/wrappers/_shared.nix
+++ b/wrappers/_shared.nix
@@ -1,6 +1,4 @@
 {
-  # Our helpers
-  helpers,
   # Option path where extraFiles should go
   filesOpt ? null,
   # Filepath prefix to apply to extraFiles
@@ -9,7 +7,12 @@
   # Is prefixed with `filesPrefix`
   initName ? "init.lua",
 }:
-{ lib, config, ... }:
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
 let
   inherit (lib)
     isAttrs
@@ -39,7 +42,7 @@ in
 
   config = mkMerge [
     # Make our lib available to the host modules
-    { nixvim.helpers = lib.mkDefault helpers; }
+    { nixvim.helpers = lib.mkDefault (import ../lib/helpers.nix { inherit pkgs lib; }); }
     # Propagate extraFiles to the host modules
     (optionalAttrs (filesOpt != null) (
       mkIf (!cfg.wrapRc) (
--- a/wrappers/_shared.nix
+++ b/wrappers/_shared.nix
@@ -1,37 +1,65 @@
-helpers:
+{
+  # Our helpers
+  helpers,
+  # Option path where extraFiles should go
+  filesOpt ? null,
+  # Filepath prefix to apply to extraFiles
+  filesPrefix ? "nvim/",
+  # Filepath to use when adding `cfg.initPath` to `filesOpt`
+  # Is prefixed with `filesPrefix`
+  initName ? "init.lua",
+}:
 { lib, config, ... }:
 let
   inherit (lib)
     isAttrs
     listToAttrs
     map
+    mkIf
+    mkMerge
     mkOption
     mkOptionType
+    optionalAttrs
+    setAttrByPath
     ;
   cfg = config.programs.nixvim;
   extraFiles = lib.filter (file: file.enable) (lib.attrValues cfg.extraFiles);
 in
 {
-  helpers = mkOption {
-    type = mkOptionType {
-      name = "helpers";
-      description = "Helpers that can be used when writing nixvim configs";
-      check = isAttrs;
+  options = {
+    nixvim.helpers = mkOption {
+      type = mkOptionType {
+        name = "helpers";
+        description = "Helpers that can be used when writing nixvim configs";
+        check = isAttrs;
+      };
+      description = "Use this option to access the helpers";
     };
-    description = "Use this option to access the helpers";
-    default = helpers;
   };
 
-  # extraFiles, but nested under "nvim/" for use in etc/xdg config
-  configFiles = listToAttrs (
-    map (
-      { target, source, ... }:
-      {
-        name = "nvim/" + target;
-        value = {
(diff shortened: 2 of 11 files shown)


wrappers/_shared.nix
```diff
@@ -1,37 +1,66 @@
...
-        name = "nvim/" + target;
-        value = {
-          inherit source;
-        };
-      }
-    ) extraFiles
-  );
+  config = mkMerge [
+    # Add our lib to `config.lib`
+    { lib.nixvim = import ../lib/helpers.nix { inherit pkgs lib; }; }
```
MattSturgeon, line 43: Do we want to just use `lib/helpers.nix`, or the entire `lib/default.nix` (which includes `helpers` as an attr)?

I guess functions outside of `helpers` may be useful in some cases, or useful ones may be added in the future?
It is also more consistent with the flake `outputs.lib.<system>`.

wrappers/_shared.nix
```diff
@@ -1,37 +1,59 @@
-helpers:
+{
+  helpers,
+  filesOpt ? null,
+  filesPrefix ? "nvim/",
+  initName ? "init.lua",
+}:
```
GaetanLepage: Maybe documenting those (inline) would be helpful.
Especially `filesOpt`.

lib/helpers.nix
```diff
@@ -1,7 +1,7 @@
 {
   lib,
-  pkgs,
-  _nixvimTests,
+  pkgs ? pkgs.lib,
```
GaetanLepage, line 3: Why does `pkgs` default to `pkgs.lib` ?

wrappers/_shared.nix
GaetanLepage: That's a good point.
Would it break a user config ? I mean, would the path of the existing helper functions change ?

GaetanLepage (changes requested):
Very good idea !
Overall looks good to me.
I just need some clarification;

lib/helpers.nix
MattSturgeon: Throughout docs, wrappers, etc, if we do not have a better way of getting `lib`, we get it from `pkgs.lib`.

This is somewhat redundant in most cases, but it matches what is done in `lib/default.nix`.

GaetanLepage: Sure, but shouldn't it be `lib ? pkgs.lib` ?

wrappers/_shared.nix
MattSturgeon: > Would it break a user config ? I mean, would the path of the existing helper functions change ?

Long-term, having what is now `helpers` just be a subset of our nixvim lib, may be better. You'd end up with `lib.nixvim.checks`, `lib.nixvim.options`, `lib.nixvim.types`, with a few top-level aliases.

As for now, we can either have `lib.nixvim.{checks,helpers}` or we can do something like:

```nix
# Copy .helpers to the lib's: top-level
let
  nixvimLib = import ../lib { inherit lib pkgs; };
in
lib.nixvim = nixvimLib // nixvimLib.helpers;
```

Either way, we can make it non-breaking with the `mkRenamedOptionModule` option. We either rename to `lib.nixvim` or `lib.nixvim.helpers`, depending on our long-term plan.

lib/helpers.nix
MattSturgeon: I'm blind, you're right lol

wrappers/_shared.nix
MattSturgeon: Sure, done

wrappers/_shared.nix
MattSturgeon: If preferred, I can drop the rename to `config.lib.*` from this PR and tackle that once we have a firmer roadmap for our lib & helpers?

GaetanLepage: Yes, this would need further discussion.

MattSturgeon: I think we all agree it's better to use the `lib` option provided by host modules.

So the reason to wait is to decide exactly how we want to use the option and attempt to get it right first try?

MattSturgeon: I've dropped the option rename from the PR now

GaetanLepage (approved):
OK for me.
@traxys could you please double check this PR ?

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1832, opened 2024-07-08
lib/types: allow emptyTable as a valid rawLua value

GaetanLepage:


lib/types.nix
```diff
@@ -12,7 +12,7 @@ let
       check = v: isString v || isRawType v;
       merge = lib.options.mergeEqualOption;
     };
-  isRawType = v: v ? __raw && isString v.__raw;
+  isRawType = v: (v ? __raw && isString v.__raw) || (v ? __empty);
```
MattSturgeon, line 15: Might be better to do this in `rawLua.check`? I think we use `isRawType` in a few functions and IDK if this will change their behaviour.

GaetanLepage: Sure, done.

GaetanLepage:
@Mergifyio queue

Issue #1618, opened 2024-06-01
[BUG] maybeRaw submodules do not generate suboption documentation


MattSturgeon:
## Description

- Options with the type `listOf (maybeRaw submodule)` or `attrsOf (maybeRaw submodule)` do not generate documentation for the submodule's options.
- `maybeRaw (listOf submodule)` and `maybeRaw (attrsOf submodule)` have the same issue.
- Switching to the type `listOf submodule` will cause documentation to generate as expected.

## Minimal, Reproducible Example (MRE)

```nix
{
  working_option = helpers.mkNullOrOption' {
    type = with lib.types; listOf (submodule {
      options = {
        test_suboption = helpers.defaultNullOpts.mkStr "" ''
          This description **will** appear in the docs
        '';
      };
    });
  };

  broken_option1 = helpers.mkNullOrOption' {
    type = with helpers.nixvimTypes; listOf (maybeRaw (submodule {
      options = {
        test_suboption = helpers.defaultNullOpts.mkStr "" ''
          This description will **not** appear in the docs
        '';
      };
    }));
  };

  broken_option2 = helpers.mkNullOrOption' {
    type = with helpers.nixvimTypes; maybeRaw (listOf (submodule {
      options = {
        test_suboption = helpers.defaultNullOpts.mkStr "" ''
          This description will **not** appear in the docs
        '';
      };
    }));
  };
}
```

### Options known to be affected

- [plugins.lsp.servers.nixd.settings.nixpkgs](https://nix-community.github.io/nixvim/plugins/lsp/servers/nixd/settings/index.html#pluginslspserversnixdsettingsnixpkgs)
- [plugins.lsp.servers.nixd.settings.options](https://nix-community.github.io/nixvim/plugins/lsp/servers/nixd/settings/index.html#pluginslspserversnixdsettingsoptions)

### Related:

- #929
- #1839

MattSturgeon:
`freeformType` seems to also trigger the issue:

```nix
{
  broken_option3 = helpers.mkNullOrOption' {
    type = with helpers.nixvimTypes; listOf (submodule {
      options = {
        freeformType = attrsOf anything;
        test_suboption = helpers.defaultNullOpts.mkStr "" ''
          This description will not appear in the docs
        '';
      };
    });
  };
}
```

MattSturgeon:
[...]

PR #1839, opened 2024-07-08
[DEBUGGING] investigate docs: maybeRaw submodule not rendering

MattSturgeon:
Investigating #1618

@nix-community/nixvim you should be able to push directly to this branch. For now I've just added some calls to `trace`.

Build using
```
nix build --no-eval-cache .#docs
```

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -4,20 +4,72 @@
 let
   # Extend nixpkg's lib, so that we can handle recursive leaf types such as `either`
   lib = pkgs.lib.extend (
-    final: prev: {
+    final: prev:
+    let
+      prettyEval =
+        value:
+        builtins.tryEval (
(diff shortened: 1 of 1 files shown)


MattSturgeon:
The current diff only prints two items:

```nix
prefix = plugins.telescope.extensions.frecency.settings.workspace_scan_cmd
t1.name = "rawLua"
t1.desc = "raw lua code"
t1Ty = { }
t2.name = "listOf"
t2.desc = "list of string"
t2Ty = { }

trace: either.getsuboptions:
prefix = plugins.telescope.extensions.frecency.settings.workspace_scan_cmd
t1.name = "rawLua"
t1.desc = "raw lua code"
t1Ty = { }
t2.name = "listOf"
t2.desc = "list of string"
t2Ty = { }
```

PR #1846, opened 2024-07-09
tests: add check for nixpkgs maintainers

MattSturgeon:
Would it make sense for this to be a custom CI workflow instead of a flake check?
I'm wondering if there's an out-of-sync nixpkgs lock scenario where it being a check could cause issues.

If we can't think of any, then this is fine and we can deal with issues as they come up.

### DIFF
--- a/tests/maintainers.nix
+++ b/tests/maintainers.nix
@@ -1,0 +1,21 @@
+{
+  pkgs ? import <nixpkgs> { },
+  lib ? pkgs.lib,
+}:
+let
+  inherit (lib) attrNames filter length;
+  nixvimList = import ../lib/maintainers.nix;
+  nixpkgsList = lib.maintainers;
+  duplicates = filter (name: nixpkgsList ? ${name}) (attrNames nixvimList);
+  count = length duplicates;
+in
+pkgs.runCommand "maintainers-test" { inherit count duplicates; } ''
+  if [ $count -gt 0 ]; then
+    echo "$count nixvim maintainers are also nixpkgs maintainers:"
+    for name in $duplicates; do
+      echo "- $name"
+    done
+    exit 1
+  fi
+  touch $out
+''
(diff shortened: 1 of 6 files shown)


tests/maintainers.nix
```diff
@@ -0,0 +1,21 @@
...
+pkgs.runCommand "maintainers-test" { inherit count duplicates; } ''
+  if [ $count -gt 0 ]; then
+    echo "$count nixvim maintainers are also nixpkgs maintainers:"
+    for name in $duplicates; do
+      echo "- $name"
+    done
+    exit 1
+  fi
+  touch $out
+''
```
GaetanLepage, line 21: Isn't this possible using a pure nix check ?
Something like an assertion ?

GaetanLepage: I don't know, couldn't it also support a "classic test". I think nixpkgs has a notion of tests...
Anyway, we could still use `assert` which would cause a dumb derivation to fail.

I just feel like this bash snippet is a bit overkilled.

traxys: We do have some Nix tests in lib-tests

MattSturgeon: > We do have some Nix tests in lib-tests

Would you rather this be done in `lib-tests`? I felt it made more sense to keep it separate.

Keeping it separate allows:
- Better fail messages
- Future expansion (e.g. check sorting)

MattSturgeon: For the record, lib-tests does essentially the same thing:

https://github.com/nix-community/nixvim/blob/abc409c16fd029f673de194c660df083ba727bd7/tests/lib-tests.nix#L309-L325

GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio refresh
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1851, opened 2024-07-11
flake/dev/formatting: run ruff format in isolated mode

GaetanLepage:
This prevents ruff to look for personal config files when running `nix fmt`.


MattSturgeon (approved):
Diff looks good. I trust you've tested ❤️

MattSturgeon:
@GaetanLepage do you think we should upstream this to the module in treefmt-nix? Maybe have a default-true option that adds the cli flag when enabled?

PR #1853, opened 2024-07-11
plugins/none-ls: refactor using `mkSourcePlugin`

MattSturgeon:
This wasn't possible before because we were using IFD, which apparently cannot be used to evaluate module `imports`.

Now we can use a "mkModule" style helper, allowing the implementation to be simplified overall. This should also simplify future refactoring, such as an rfc42 "withArgs" equivalent.

I was able to clean the test up a little too, now we don't need to list all "unpackaged" sources in the test; only sources we want to _disable_ from being tested.


plugins/none-ls/_mk-source-plugin.nix
```diff
@@ -0,0 +1,53 @@
...
+{
+  options.plugins.none-ls.sources.${sourceType}.${sourceName} =
+    {
+      enable = lib.mkEnableOption "the ${sourceName} ${sourceType} source for none-ls";
+      withArgs = helpers.mkNullOrOption helpers.nixvimTypes.strLua ''
+        Raw Lua code passed as an argument to the source's `with` method.
+      '';
+    }
+    # Only declare a package option if a package is required
+    // lib.optionalAttrs (packaged ? ${sourceName} || lib.elem sourceName unpackaged) {
```
MattSturgeon, line 25: @traxys can you clarify why it is useful to have a separate `unpackaged` list?

If we made `packaged` into a nullable list, none of _this_ implementation (below) would have to change, but this line could be simplified:

```suggestion
    // lib.optionalAttrs (packaged ? ${sourceName}) {
```

If you just like the _list syntax_ and not needing to spam `null` repeatedly, we can use `genAttrs`:

```nix
packaged =
  {
    inherit (pkgs)
      actionlint
      alejandra
      # etc
      ;
  }
  # Unpackaged
  // genAttrs [
    "blade_formatter"
    "bsfmt"
    # etc
  ] (_: null);
```

(I agree the _package or null or boolean_ we had before was clunky, so I see the value in the separate `noPackage` list).

MattSturgeon: (I've pushed up a change implementing this, let me know if you'd rather I drop/revert it)

tests/test-sources/plugins/none-ls.nix
```diff
@@ -100,95 +99,70 @@
...
+                  "verilator"
+                  "verible_verilog_format"
+                ])
+                ++ (lib.optionals pkgs.stdenv.isAarch64 [
+                  "semgrep"
+                  "smlfmt"
+                  # As of 2024-03-11, swift-format is broken on aarch64
+                  # TODO: re-enable this test when fixed
+                  "swift_format"
+                ]);
```
traxys, line 150: Have you checked that all the packages marked as broken are still broken ?

plugins/none-ls/sources.nix
```diff
@@ -0,0 +1,22 @@
...
+    ));
+
+  config =
+    let
+      cfg = config.plugins.none-ls;
+      gitsignsEnabled = cfg.sources.code_actions.gitsigns.enable;
+    in
+    lib.mkIf cfg.enable {
+      # Enable gitsigns if the gitsigns source is enabled
+      plugins.gitsigns.enable = lib.mkIf gitsignsEnabled true;
```
traxys, line 20: Do you think there is a way to have a warning if gitsigns was not already enabled? I think we will end up with an infinite recursion, but maybe you know a way to to that?
If we can add a warning we could remove this magic in 24.11

plugins/none-ls/_mk-source-plugin.nix
```diff
@@ -0,0 +1,50 @@
...
+{
+  config,
+  pkgs,
+  lib,
+  helpers,
+  ...
+}:
+let
+  inherit (import ./packages.nix pkgs) packaged;
+  pkg = packaged.${sourceName};
```
traxys, line 12: Lazy evaluation is nice :D

plugins/none-ls/_mk-source-plugin.nix
```diff
@@ -0,0 +1,50 @@
...
+    }
+    # Only declare a package option if a package is required
+    // lib.optionalAttrs (packaged ? ${sourceName}) {
+      package = lib.mkOption (
+        {
+          type = lib.types.nullOr lib.types.package;
+          description =
+            "Package to use for ${sourceName} by none-ls. "
+            + (lib.optionalString (pkg == null) ''
+              Not handled in nixvim, either install externally and set to null or set the option with a derivation.
```
traxys, line 33: maybe in `nixpkgs` instead of `nixvim`? As we try to be up-to-date. Though this means that we sort-of garantee to check the unpackaged sources, I'm not sure if we want that.
In the end, as you wish I think

plugins/none-ls/_mk-source-plugin.nix
```diff
@@ -0,0 +1,50 @@
...
+  pkg = packaged.${sourceName};
+
+  cfg = config.plugins.none-ls;
+  cfg' = config.plugins.none-ls.sources.${sourceType}.${sourceName};
+in
+{
+  options.plugins.none-ls.sources.${sourceType}.${sourceName} =
+    {
+      enable = lib.mkEnableOption "the ${sourceName} ${sourceType} source for none-ls";
+      withArgs = helpers.mkNullOrOption helpers.nixvimTypes.strLua ''
```
traxys, line 21: It should be straighforward (in another PR) to have structured options here no? We'd need some kind of coercion from string to rawLua though

traxys (commented):
Some nits/questions, but globally nothing much to say

plugins/none-ls/_mk-source-plugin.nix
traxys: I don't have any issues with that, it was just simpler to write

tests/test-sources/plugins/none-ls.nix
MattSturgeon: No I haven't. Is the best way to just empty the list and see what buildbot does?

plugins/none-ls/sources.nix
MattSturgeon: I think this is possible by checking the `options.plugins.gitsigns.enable.definitions` length.

If the definitions only has one element, then that's our mkIf. If there's multiple definitions then the user has defined it themselves.

Would we also want the warning to check if the final merged value would be `true` as well?

MattSturgeon: While I think it's possible, I don't really want to tackle it in this PR.

plugins/none-ls/_mk-source-plugin.nix
MattSturgeon: I just copy-pasted the existing description. I agree we can improve the wording.

Maybe:

> Currently not packaged in nixpkgs.
> Either set this to `null` and install ${sourceName} outside of nix, or set this to a custom or third-party nix package.

plugins/none-ls/_mk-source-plugin.nix
MattSturgeon: Agreed. #1705 is tracking the idea.

tests/test-sources/plugins/none-ls.nix
traxys: Yeah I think... We can punt that to another PR

plugins/none-ls/sources.nix
traxys: Yeah no problem

plugins/none-ls/_mk-source-plugin.nix
traxys: Looks good!

MattSturgeon: Done

traxys (approved):
CI fail seems unrelated

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

Issue #1849, opened 2024-07-11
[BUG] Treesitter trying to create directory in nix store


RoastBeefer00:
| Field        | Description |
|--------------|-------------|
| Plugin       | `treesitter`  | 
| Nixpkgs      | `nixos-unstable` | 
| Home Manager | `24.05` | 

- [x ] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Error
I get the following error every time I launch neovim after the latest updates:

```
Error detected while processing /nix/store/5y36fg53rjbivbrxih547n5qjlsl46q4-init.lua:
Could not create parser dir ' /nix/store/ak1knrp12ckr8axic4g6ndfifgfmqzpl-vim-pack-dir/pack/m
yNeovimPackages/start/nvim-treesitter/parser ':  Vim:E739: Cannot create directory /nix/store
/ak1knrp12ckr8axic4g6ndfifgfmqzpl-vim-pack-dir/pack/myNeovimPackages/start/nvim-treesitter/pa
rser: read-only file system
Press ENTER or type command to continue
```

[Here is my config](https://github.com/RoastBeefer00/nixos-config/tree/main/nixvim)

elohmeier:
Setting `plugins.treesitter.nixGrammars = false;` fixes this for me.

elohmeier:
I've fixed this by switching to grammarPackages instead of the `plugins.treesitter.ensure_installed` setting.

Example:
```nix
{ pkgs, ... }:

{
  plugins.treesitter = {
    enable = true;
  
    grammarPackages = with pkgs.vimPlugins.nvim-treesitter.builtGrammars; [
      bash
      git_config
      git_rebase
      gitattributes
      gitcommit
      gitignore
      json
      jsonc
      lua
      make
      markdown
      meson
      ninja
      nix
      readline
      regex
      ssh-config
      toml
      vim
      vimdoc
      xml
      yaml
    ];
  };
}
```

RoastBeefer00:
Thank you @elohmeier this resolved the issue for me as well.

darkylein:
I run into the same bug. I don't think this issue should be closed. Maybe @khaneliman could have a look over it.

ForsakenHarmony:
@RoastBeefer00 that's a workaround not a fix, should probably reopen this issue

MattSturgeon:
[...]

PR #1855, opened 2024-07-11
plugins/treesitter: fix parser_install_dir default

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1849

We default to null for settingsOptions to respect plugin defaults. The nvim-treesitter plugin also has a null default for parser_install_dir, but the fallback of package path isn't a writeable directory on NixOS. This provides a sane default for NixOS users.

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -142,12 +142,19 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       List of parsers to ignore installing. Used when `ensure_installed` is set to "all".
     '';
 
-    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
-      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
+    parser_install_dir = helpers.mkNullOrOption' {
+      type = with helpers.nixvimTypes; maybeRaw str;
+      # Backport the default from nvim-treesitter 1.0
+      # The current default doesn't work on nix, as it is readonly
+      default.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+      pluginDefault = lib.literalMD "the plugin's package directory";
+      description = ''
+        Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
 
-      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
-      Otherwise, change it to something that will work for you!
-    '';
+        By default, parsers are installed to the "site" dir.
+        If set to `null` the _plugin default_ is used, which will not work on nix.
+      '';
+    };
 
     sync_install = helpers.defaultNullOpts.mkBool false ''
       Install parsers synchronously (only applied to `ensure_installed`).
@@ -158,7 +165,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     auto_install = false;
     ensure_installed = [ "all" ];
     ignore_install = [ "rust" ];
-    parser_install_dir = "$XDG_DATA_HOME/nvim/treesitter";
+    parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'treesitter')";
     sync_install = false;
 
     highlight = {
@@ -284,7 +291,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       # NOTE: Upstream state that the parser MUST be at the beginning of runtimepath.
       # Otherwise the parsers from Neovim takes precedent, which may be incompatible with some queries.
       (optionalString (cfg.settings.parser_install_dir != null) ''
-        vim.opt.runtimepath:prepend("${cfg.settings.parser_install_dir}")
+        vim.opt.runtimepath:prepend(${helpers.toLuaObject cfg.settings.parser_install_dir})
       '')
       + ''
         require('nvim-treesitter.configs').setup(${helpers.toLuaObject cfg.settings})
--- a/tests/test-sources/plugins/languages/treesitter/treesitter.nix
+++ b/tests/test-sources/plugins/languages/treesitter/treesitter.nix
@@ -8,7 +8,9 @@
         auto_install = false;
         ensure_installed = [ ];
         ignore_install = [ ];
-        parser_install_dir = null;
+        # NOTE: This is our default, not the plugin's
+        parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+
         sync_install = false;
 
         highlight = {


alesauce:
I noticed this issue a few days ago as well because it blocked my auto-update workflow on [my standalone flake](https://github.com/alesauce/nixvim-flake). I'm trying to pare down my config to the MRE, but pulling the changes made to this PR didn't fix the issue for me. [Branch where I switched NixVim inputs.](https://github.com/alesauce/nixvim-flake/tree/experiment/test-treesitter-fix-pr) [Here is the failed check](https://github.com/alesauce/nixvim-flake/actions/runs/9901015974/job/27352720121).

Let me know if there's anything else I can do to help with this!

khaneliman:
> I noticed this issue a few days ago as well because it blocked my auto-update workflow on [my standalone flake](https://github.com/alesauce/nixvim-flake). I'm trying to pare down my config to the MRE, but pulling the changes made to this PR didn't fix the issue for me. [Branch where I switched NixVim inputs.](https://github.com/alesauce/nixvim-flake/tree/experiment/test-treesitter-fix-pr) [Here is the failed check](https://github.com/alesauce/nixvim-flake/actions/runs/9901015974/job/27352720121).
> 
> Let me know if there's anything else I can do to help with this!

This is just reverting the parser_install_dir default back to how it was before since the plugin's default doesn't work on NixOS. Looks like your CI doesn't have XDG_DATA_HOME so it prints as " /nvim/treesitter" and that root path doesn't work in the container. I see you dont disable nixGrammars and the behavior before was to unset a few options for you if was enabled (it is by default). One of which would have set the parser_install_dir to null before so the plugin's default looking for the package path would have kicked in instead of the path that its trying now.

alesauce:
Ahhhhhh, I understand now - thank you for taking the time to explain!

plugins/languages/treesitter/treesitter.nix
```diff
@@ -142,12 +142,17 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+      default = "$XDG_DATA_HOME/nvim/treesitter";
+      defaultText = "$XDG_DATA_HOME/nvim/treesitter";
+      description = ''
+        Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
 
-      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
-      Otherwise, change it to something that will work for you!
-    '';
+        This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
+        Otherwise, change it to something that will work for you!
```
MattSturgeon, line 153: Depending on the final implementation, this note will likely need re-writing.

MattSturgeon (changes requested):
Thanks for figuring this out!

Instead of just reverting to what we had before, this is an opportunity to improve the implementation.

- Upstream are apparently aware of nix, as `get_parser_install_dir` [has a note](https://github.com/nvim-treesitter/nvim-treesitter/blob/972aa544efb56e2f2f53c5f3c2537e43467dd5cb/lua/nvim-treesitter/configs.lua#L577-L581) regarding /nix/store being readonly.
- They [also state](https://github.com/nvim-treesitter/nvim-treesitter/blob/972aa544efb56e2f2f53c5f3c2537e43467dd5cb/lua/nvim-treesitter/configs.lua#L577-L581) that ["site" dir](https://github.com/nvim-treesitter/nvim-treesitter/blob/972aa544efb56e2f2f53c5f3c2537e43467dd5cb/lua/nvim-treesitter/utils.lua#L143-L147) from the runtimepath is used as a fallback.
  - We should investigate why this isn't helping us - is the note worded poorly or is there a bug upstream?
  - I can't see any code implementing a "site dir" fallback, maybe the note is just outdated?

If we do end up setting our own default:
- We should still try to document the plugin default
- We should try to set a sane default for both linux & [darwin](https://stackoverflow.com/a/5084892).
- We should implement the XDG fallback:
  - `${XDG_DATA_HOME:=~/.local/share}`
  - or the equivalent as a lua function.

Also upstream [state that](https://github.com/nvim-treesitter/nvim-treesitter/blob/972aa544efb56e2f2f53c5f3c2537e43467dd5cb/doc/nvim-treesitter.txt#L42-L46) `parser_install_dir` should be in the rtp, we should probably document that (and implement it if we set a default).

traxys:
The other option is to warn the user when we detect that we need a parser install dir, and none was supplied

MattSturgeon:
In nvim-treesitter's `main` branch[^1] `parser_install_dir` has been renamed to `install_dir`, and the default is now:
```lua
vim.fs.joinpath(vim.fn.stdpath('data'), 'site')
```
i.e. `$XDG_DATA_HOME/nvim/site`

It seems the note on the `master` branch about falling back to "site" was incorrect.

---

Seeing as we will remove the `parser_install_dir` option when 1.0 makes it to nixos-unstable, I think it is fine to have it default to `vim.fs.joinpath(vim.fn.stdpath('data'), 'site')` until then, even though doing so isn't really in-line with RFC42.

Using `vim.fn.stdpath` also resolves my concerns with darwin compatibility and XDG vars being undefined.

[^1]: `main` is their WIP branch for working on 1.0. The _current_ default branch, used in nixpkgs, is `master`.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -142,12 +142,17 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       List of parsers to ignore installing. Used when `ensure_installed` is set to "all".
     '';
 
-    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
-      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
+    parser_install_dir = mkOption {
+      type = with types; nullOr str;
+      default = "$XDG_DATA_HOME/nvim/treesitter";
+      defaultText = "$XDG_DATA_HOME/nvim/treesitter";
```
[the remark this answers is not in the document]
MattSturgeon:
```suggestion
      # Backport the default from nvim-treesitter 1.0
      # The current default doesn't work on nix, as it is readonly
      default.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
      pluginDefault = lib.literalMD "the plugin's package directory";
```

plugins/languages/treesitter/treesitter.nix
MattSturgeon:
```suggestion

        If left as default, parsers are installed to the "site" dir.

        If set to null the _plugin default_ is used, which will not work on nix.

        If a custom path is used it must be added to the `runtimepath`, e.g. using:
        ```lua
        vim.opt.runtimepath:append('/some/path/to/store/parsers')
        ```
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -158,7 +167,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     auto_install = false;
     ensure_installed = [ "all" ];
     ignore_install = [ "rust" ];
-    parser_install_dir = "$XDG_DATA_HOME/nvim/treesitter";
+    parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
```
MattSturgeon, line 170:
```suggestion
```

No need to include default values in the example.

Could also do this:

```suggestion
    parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'treesitter')";
```
But IMO it distracts from more useful parts of the example. Especially since the user _also_ needs to add the path to the rtp and we can't really show that in the example

tests/test-sources/plugins/languages/treesitter/treesitter.nix
```diff
@@ -8,7 +8,8 @@
         auto_install = false;
         ensure_installed = [ ];
         ignore_install = [ ];
-        parser_install_dir = null;
+        parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+
```
MattSturgeon, line 12:
```suggestion
        # NOTE: This is our default, not the plugin's
        parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -142,12 +142,21 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       List of parsers to ignore installing. Used when `ensure_installed` is set to "all".
     '';
 
-    parser_install_dir = helpers.defaultNullOpts.mkStr "$XDG_DATA_HOME/nvim/treesitter" ''
-      Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
+    parser_install_dir = helpers.mkNullableWithRaw' {
+      type = types.str;
```
MattSturgeon, line 146:
```suggestion
    parser_install_dir = helpers.mkNullOrOption' {
      type = with helpers.nixvimTypes; maybeRaw str;
```

My mistake (as discussed on matrix)

MattSturgeon (changes requested):

plugins/languages/treesitter/treesitter.nix
khaneliman: We automatically add that path to rtp in the extraConfig, though. Do you want me to leave that example as it was prior to show how to set to the XDG path then?

MattSturgeon: I don't mind having an example, but it should be different to the default.

Ideally examples should be a useful "real world" example. If that's not possible an obvious "psudo" example that users know not to copy verbatim is ok too. But we don't need examples just for the sake of having them.

If we add the setting to the rtp in extraConfig then we can probably drop the note I suggested in the option desc too?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -284,7 +289,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       # NOTE: Upstream state that the parser MUST be at the beginning of runtimepath.
       # Otherwise the parsers from Neovim takes precedent, which may be incompatible with some queries.
       (optionalString (cfg.settings.parser_install_dir != null) ''
-        vim.opt.runtimepath:prepend("${cfg.settings.parser_install_dir}")
+        vim.opt.runtimepath:prepend("${helpers.toLuaObject cfg.settings.parser_install_dir}")
```
MattSturgeon, line 292:
```suggestion
        vim.opt.runtimepath:prepend(${helpers.toLuaObject cfg.settings.parser_install_dir})
```

`toLuaObject` will handle the quotes

MattSturgeon (approved):
LGTM, thanks again for your efforts!

Could you give the commit message a quick once over, as mentioned on matrix?

@traxys do you want to have a final look before we merge?

plugins/languages/treesitter/treesitter.nix
```diff
@@ -142,12 +142,17 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
-      This default might not work on your own install, please make sure that `$XDG_DATA_HOME` is set if you want to use the default.
-      Otherwise, change it to something that will work for you!
-    '';
+    parser_install_dir = helpers.mkNullOrOption' {
+      type = with helpers.nixvimTypes; maybeRaw str;
+      # Backport the default from nvim-treesitter 1.0
+      # The current default doesn't work on nix, as it is readonly
+      default.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+      pluginDefault = lib.literalMD "the plugin's package directory";
+      description = ''
```
MattSturgeon, line 151: Looks like this line was removed by mistake:

```
Location of the parsers to be installed by the plugin (only needed when `nixGrammars` is disabled).
```

traxys (approved):

MattSturgeon:
@metgifyio queue

khaneliman:
> @metgifyio queue

typo :P

MattSturgeon:
@mergifyio refresh

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1858, opened 2024-07-12
plugins/treesitter: fix folding

khaneliman:
opts getting wiped out by setting entire set as default instead of each attribute

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -319,9 +319,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       cfg.treesitterPackage
     ];
 
-    opts = mkIf cfg.folding (mkDefault {
-      foldmethod = "expr";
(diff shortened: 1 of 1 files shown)


MattSturgeon (approved):
I'm surprised moving the mkDefault to the specific attrs makes a difference, but if it works then today I've learnt something!

khaneliman:
> I'm surprised moving the mkDefault to the specific attrs makes a difference, but if it works then today I've learnt something!

My assumption was that it basically treated the entire attribute set as the default definition for opts and then providing another opts attribute overrode it? 

I have library functions in my repo for doing this on nested attributes I think for this reason.
```nix
default-attrs = mapAttrs (_key: mkDefault);

nested-default-attrs = mapAttrs (_key: default-attrs);
```

MattSturgeon:
@mergifyio queue

MattSturgeon:
Thanks again! 👍

PR #1862, opened 2024-07-14
github/update: run on current branch

MattSturgeon:
This will be paired with a PR refactoring the workflow on the main branch, but nothing should block merging this early since we do not currently run the workflow _on_ the stable branch.

This PR syncs the update workflow with `main`'s, but doesn't include the parts where we generate files.

It also removes the inputs and schedule parts of the workflow, as the intention is a new `call-update.yml` workflow will be added to call each branch's `update.yml` workflow.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,39 +1,22 @@
-name: update-flake-lock
+name: update
 on:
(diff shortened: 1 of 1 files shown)


traxys (approved):

MattSturgeon:
@mergifyio queue

PR #1864, opened 2024-07-14
plugins/treesitter: disable folding by default

SuperSandro2000:
I think this got erroneously changed in #1801.

Having a fold at every function is rather annoying to work with and having to unfold many things in every file.

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -203,12 +203,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   };
 
   extraOptions = {
-    folding = mkOption {
-      type = types.bool;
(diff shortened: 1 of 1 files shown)


MattSturgeon:
> I think this got erroneously changed in #1801.

Yes, it used to be:

```nix
folding = mkEnableOption "tree-sitter based folding"
```

I missed that in the review

traxys:
I think `mkEnableOption` is the right function to declare this option

SuperSandro2000:
> Yes, it used to be:
> 
> ```nix
> folding = mkEnableOption "tree-sitter based folding"
> ```

Force pushed that

MattSturgeon (approved):
LGTM, thanks!
(CI is unrelated)

PR #1866, opened 2024-07-15
lib/options: lib/neovim-plugin: Add lazyLoad option to plugins

psfloyd:
This PR works on a solution for the issue #421.

# Already implemented
- Adding `mkLazyLoadOption` to generate lazyLoad options for plugins.
- Adding and implementing a lazy-loading mechanism using [lz.n](https://github.com/nvim-neorocks/lz.n)
- Conditionally enabling lazy or regular loading depending on user settings in `mkNeovimPlugin`

# Planned
- For plugins that do not callSetup, extract `extraConfigLua` from extraConfig.
- Integrate with `mkVimPlugin`


lib/neovim-plugin.nix
```diff
@@ -48,9 +48,12 @@ with lib;
       extraPackages ? [ ],
       callSetup ? true,
       installPackage ? true,
+      allowLazyLoad ? true,
```
MattSturgeon, line 51: If the current implementation depends on `callSetup`, maybe do `allowLazyLoad ? callSetup`?

lib/neovim-plugin.nix
```diff
@@ -99,21 +102,36 @@ with lib;
             example = settingsExample;
           };
         }
+        // optionalAttrs allowLazyLoad {
+          lazyLoad = nixvimOptions.mkLazyLoadOption {
+            inherit originalName cfg;
+            optionsForPlugin = true;
+            lazyLoad = {
+              # TODO: extract extraConfigLua and extraConfigLuaPre from extraConfig
```
MattSturgeon, line 110: IDK if this is a practical goal, due to the way the module system works.

In theory, `extraConfig` functions could be returning a `mkIf`, `mkDefault`, `mkMerge`, etc assuming that the module system will be able to merge it.

It's also possible that plugins define some of their config in other imported modules, e.g. cmp does this, as does none-ls, and several others.

lib/neovim-plugin.nix
```diff
@@ -99,21 +102,36 @@ with lib;
...
+              after = optionalString callSetup ''
+                require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
+              '';
+              colorscheme = if (isColorscheme && (colorscheme != null)) then [ colorscheme ] else null;
```
MattSturgeon, line 114: Could the `colorscheme` arg be merged in with `optionalAttrs` rather than explicitly passing in `null`

```nix
// (optionalAttrs (isColorscheme && colorscheme != null) { colorscheme = [ colorscheme ]; })
```

Also, why does mkLazyLoadOption take colorscheme as a list? If it accepted string values (or either using `lib.toList`) then we could do `{ inherit colorscheme; }`.

lib/neovim-plugin.nix
```diff
@@ -99,21 +102,36 @@ with lib;
...
+            } // lazyLoad;
+          };
+        }
         // extraOptions;
 
       config =
         let
-          cfg = config.${namespace}.${name};
           extraConfigNamespace = if isColorscheme then "extraConfigLuaPre" else "extraConfigLua";
+          lazyLoaded = if (cfg ? lazyLoad) then cfg.lazyLoad.enable else false;
```
MattSturgeon, line 123: This can just be:

```nix
lazyLoaded = cfg.lazyLoad.enable or false;
```

`or` defines a fallback for when an attribute doesn't exist.

lib/neovim-plugin.nix
```diff
@@ -99,21 +102,36 @@ with lib;
...
           {
             extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
             inherit extraPackages;
 
-            ${extraConfigNamespace} = optionalString callSetup ''
+            ${extraConfigNamespace} = optionalString (callSetup && !lazyLoaded) ''
               require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
             '';
+
+            plugins.lz-n.plugins = mkIf lazyLoaded [ cfg.lazyLoad ];
```
MattSturgeon, line 134: I think we also need to have an assertion to ensure `plugins.lz-n` is enabled:

```nix
assertions = [
  {
    assertion = cfg.lazyLoad -> config.plugins.lz-n.enable;
    message = ''
      Some message about lazyloading bein configured for the plugin but not enabled globally.
    '';
  }
];
```

MattSturgeon (commented):
Thanks for working on this, I'm sure many people will be pleased to see progress here!

I haven't taken a proper look yet, but I left some comments below.

- Would it be possible to extract parts of this into smaller PRs?
  - E.g. initializing `plugins.lz-n` could be its own PR.
- Would it be possible for `mkLazyLoadOption` to work with either `plugins.lz-n` or `plugins.lazy` based on user preference?

lib/neovim-plugin.nix
traxys: One way to handle this would be to add internal options to each plugin of the form `<pluginName>Config(Pre|Post)?`, and have the mkNeovimPlugin put the code in the correct places according to the specified options

MattSturgeon: > One way to handle this would be to add internal options to each plugin of the form `<pluginName>Config(Pre|Post)?`, and have the mkNeovimPlugin put the code in the correct places according to the specified options

IMO having `plugins.*.{pre,post,}config` options would be useful, not just internally but to users too.

Updating the codebase to actually make use of such options internally would be a significant chore though.

psfloyd:
Thank you for your review!

> * Would it be possible to extract parts of this into smaller PRs?
>   
>   * E.g. initializing `plugins.lz-n` could be its own PR.

Yes, I am going to make a PR for lz-n, and use that PR as base for the lazyLoad options

> * Would it be possible for `mkLazyLoadOption` to work with either `plugins.lz-n` or `plugins.lazy` based on user preference?

Yes, I think it should be possible. I could add a `backend` option to select which one to use. But I don't think it should be set on a per plugin basis.
Should we add a global `config.lazyLoader` option with global parameters for lazy-loading?
On that namespace we could add an `enable` option and apply it to the corresponding backend.

lib/neovim-plugin.nix
psfloyd: Even if plugins do not require a setup function, they are still lazy-loaded by lz.n. Their modules are not loaded until the conditions are reached.

lib/neovim-plugin.nix
psfloyd: You are right, this adds some complexity that may be out of scope for now.

lib/neovim-plugin.nix
psfloyd: Great idea, I will implement it on the new PR.

lib/neovim-plugin.nix
psfloyd: `optionalAttrs` 

> Could the `colorscheme` arg be merged in with `optionalAttrs` rather than explicitly passing in `null`
> 
> ```nix
> // (optionalAttrs (isColorscheme && colorscheme != null) { colorscheme = [ colorscheme ]; })
> ```
Yes.

> Also, why does mkLazyLoadOption take colorscheme as a list? If it accepted string values (or either using `lib.toList`) then we could do `{ inherit colorscheme; }`.
The colorscheme option is a list because of merging. 
It could be of type `(either str (listOf str))`. But if I understand correctly, when it is defined as a `str`, then it cannot be merged with other values set as `str` or `list`.
In that case, if plugin mantainers set the option as a `str`, then user cannot extend it.

MattSturgeon: > The colorscheme option is a list because of merging.

Ah, I read this as passing in arguments to mkLazyLoadOption rather than defining config.

It's still possible to do that with the type system, though whether we need to in this case is another matter!

Various non-list types can be merged (e.g. types.lines), however the simplist option here would be a type that coerces to a list:

```nix
type =
  with lib.types;
  coercedTo str lib.toString (listOf str);
```

psfloyd: I didn't know about `coercedTo`. I am going to implement it that way, and also the other `str or listOf str`. Thanks for the help

PR #1874, opened 2024-07-16
plugins/lz-n: init

psfloyd:
This is the first PR of splitting PR #1866.

This PR adds:
- `plugins.lz-n` and its options for lazy-loading plugins using [lz.n](https://github.com/nvim-neorocks/lz.n).
- Tests for `plugins.lz-n`.

### DIFF
--- a/plugins/pluginmanagers/lz-n.nix
+++ b/plugins/pluginmanagers/lz-n.nix
@@ -1,0 +1,200 @@
+{
+  lib,
+  options,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "lz-n";
+  originalName = "lz.n";
+  maintainers = [ maintainers.psfloyd ];
+  defaultPackage = pkgs.vimPlugins.lz-n;
+
+  settingsDescription = ''
+    Options provided to `vim.g.lz_n`.
+
+    `{ load = "fun"; }` -> `vim.g.lz_n = { load = fun, }`
+  '';
+
+  settingsOptions = {
+    load = defaultNullOpts.mkLuaFn "vim.cmd.packadd" ''
+      Function used by `lz.n` to load plugins.
+    '';
+  };
+
+  callSetup = false; # Does not use setup
+
+  extraOptions =
+    let
+      lzPluginType = types.submodule {
+        freeformType = types.attrsOf types.anything;
+        options = {
+          __unkeyed-1 = mkOption {
+            type = types.str;
+            description = ''
+              The "unkeyed" attribute is the plugin's name.
+              This is passed to `load` function and should normally match the repo name of the plugin.
+
+              More specifically, this is the name of the folder in `/pack/opt/{name}` that is loaded with `load` (`packadd` by default).
+              See `:h packadd`.
+            '';
+          };
+
+          enabled = nixvim.defaultNullOpts.mkStrLuaFnOr types.bool true ''
+            When false, or if the function returns false, then this plugin will not be included in the spec.
+            This option corresponds to the `enabled` property of lz.n.
+          '';
+
+          beforeAll = nixvim.mkNullOrLuaFn ''
+            Always executed before any plugins are loaded.
+          '';
+
+          before = nixvim.mkNullOrLuaFn ''
+            Executed before this plugin is loaded.
+          '';
+
+          after = nixvim.mkNullOrLuaFn ''
+            Executed after this plugin is loaded.
+          '';
+
+          load = nixvim.mkNullOrLuaFn ''
+            Can be used to override the `vim.g.lz_n.load()` function for this plugin.
+          '';
+
+          priority = nixvim.defaultNullOpts.mkUnsignedInt (literalMD "`50` (or `1000` if `colorscheme` is set)") ''
+            Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
+          '';
+
+          event = nixvim.mkNullOrOption' {
+            type = types.anything;
+            description = ''
+              Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua
+            '';
+            example = [
+              "BufEnter *.lua"
+              "DeferredUIEnter"
+            ];
+          };
+
+          cmd = nixvim.mkNullOrOption' {
+            type = types.anything;
+            description = ''
+              Lazy-load on command.
+            '';
+            example = [
+              "Neotree"
+              "Telescope"
+            ];
+          };
+
+          ft = nixvim.mkNullOrOption' {
+            type = types.anything;
+            description = ''
+              Lazy-load on filetype.
+            '';
+            example = [ "tex" ];
+          };
+
+          colorscheme = nixvim.mkNullOrOption' {
+            type = types.anything;
+            description = ''
+              Lazy-load on colorscheme.
+            '';
+            example = "onedarker";
+          };
+
+          keys = nixvim.mkNullOrOption' {
+            type = types.listOf types.anything;
+            description = ''
+              Lazy-load on key mapping. Mode is `n` by default.
+            '';
+            example = [
+              "<C-a>"
+              [
+                "<C-x>"
+                "g<C-x>"
+              ]
+              {
+                __unkeyed-1 = "<leader>fb";
+                __unkeyed-2 = "<CMD>Telescope buffers<CR>";
+                desc = "Telescope buffers";
+              }
+              { __raw = "{ '<leader>ft', '<CMD>Neotree toggle<CR>', desc = 'NeoTree toggle' }"; }
+            ];
+          };
+        };
+      };
+    in
+    {
+      plugins = mkOption {
+        description = ''
+          List of plugin specs provided to the `require('lz.n').load` function.
+          Plugin specs can be ${nixvim.nixvimTypes.rawLua.description}.
+        '';
+        default = [ ];
+        type = types.listOf lzPluginType;
+        example = [
+          {
+            __unkeyed-1 = "neo-tree.nvim";
+            enabled = ''
+              function()
+              return true
+              end
+            '';
+            keys = [
+              {
+                __unkeyed-1 = "<leader>ft";
+                __unkeyed-2 = "<CMD>Neotree toggle<CR>";
+                desc = "NeoTree toggle";
+              }
+            ];
+            after = ''
+              function()
+                require("neo-tree").setup()
+              end
+            '';
+          }
+          {
+            __unkeyed-1 = "telescope.nvim";
+            cmd = [ "Telescope" ];
+            keys = [
+              {
+                __unkeyed-1 = "<leader>fa";
+                __unkeyed-2 = "<CMD>Telescope autocommands<CR>";
+                desc = "Telescope autocommands";
+              }
+              {
+                __unkeyed-1 = "<leader>fb";
+                __unkeyed-2 = "<CMD>Telescope buffers<CR>";
+                desc = "Telescope buffers";
+              }
+            ];
+          }
+          {
+            __unkeyed-1 = "onedarker.nvim";
+            colorscheme = [ "onedarker" ];
+          }
+          {
+            __raw = ''
+              {
+                  "crates.nvim",
+                  ft = "toml",
+              },
+            '';
+          }
+        ];
+      };
+    };
+
+  extraConfig = cfg: {
+    globals.lz_n = modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
+    extraConfigLua = mkIf (cfg.plugins != [ ]) ''
+      require('lz.n').load( ${nixvim.toLuaObject cfg.plugins})
+    '';
+  };
+}
--- a/tests/test-sources/plugins/pluginmanagers/lz-n.nix
+++ b/tests/test-sources/plugins/pluginmanagers/lz-n.nix
@@ -1,0 +1,170 @@
+let
+  /*
+    Transform plugins into attrset and set optional to true
+    This installs the plugin files to {runtimepath}/pack/opt, instead of {runtimepath}/pack/start.
+    Plugins in pack/opt are not loaded on startup, but can be later loaded with `:packadd {name}` or `require("lz.n").load({name})`
+
+    See `nixpkgs/pkgs/applications/editors/neovim/utils.nix`
+    See `nixpkgs/pkgs/applications/editors/vim/plugins/vim-utils.nix`
+  */
+  optionalPlugins = map (x: (if x ? plugin then x else { plugin = x; }) // { optional = true; });
+in
+{
+  # Empty configuration
+  empty = {
+    plugins.lz-n.enable = true;
+  };
+
+  # Empty configuration
+  defaults = {
+    plugins.lz-n = {
+      enable = true;
+      settings = {
+        load.__raw = "vim.cmd.packadd";
+      };
+    };
+  };
+
+  # single-plugin and priority of plugins.lz-n.settings to globals.lz-n
+  example-single-plugin = {
+    module =
+      { pkgs, lib, ... }:
+      {
+        extraPlugins = optionalPlugins [ pkgs.vimPlugins.neo-tree-nvim ];
+
+        plugins.lz-n = {
+          enable = true;
+          settings = {
+            load = lib.mkDefault "vim.cmd.packadd";
+          };
+          plugins = [
+            # enabled, on keys as rawLua
+            {
+              __unkeyed-1 = "neo-tree.nvim";
+              enabled = ''
+                function()
+                return true
+                end
+              '';
+              keys = [
+                {
+                  __unkeyed-1 = "<leader>ft";
+                  __unkeyed-2 = "<CMD>Neotree toggle<CR>";
+                  desc = "NeoTree toggle";
+                }
+              ];
+              after = # lua
+                ''
+                  function()
+                    require("neo-tree").setup()
+                  end
+                '';
+            }
+          ];
+        };
+      };
+  };
+
+  example-multiple-plugin = {
+    module =
+      { pkgs, lib, ... }:
+      {
+        extraPlugins =
+          with pkgs.vimPlugins;
+          [ onedarker-nvim ]
+          ++ (optionalPlugins [
+            neo-tree-nvim
+            dial-nvim
+            vimtex
+            telescope-nvim
+            nvim-biscuits
+            crates-nvim
+          ]);
+
+        plugins.treesitter.enable = true;
+
+        plugins.lz-n = {
+          enable = true;
+          plugins = [
+            # enabled, on keys
+            {
+              __unkeyed-1 = "neo-tree.nvim";
+              enabled = ''
+                function()
+                return true
+                end
+              '';
+              keys = [
+                {
+                  __unkeyed-1 = "<leader>ft";
+                  __unkeyed-2 = "<CMD>Neotree toggle<CR>";
+                  desc = "NeoTree toggle";
+                }
+              ];
+              after = # lua
+                ''
+                  function()
+                    require("neo-tree").setup()
+                  end
+                '';
+            }
+            # on keys as list of str and rawLua
+            {
+              __unkeyed-1 = "dial.nvim";
+              keys = [
+                "<C-a>"
+                { __raw = "{ '<C-x>'; mode = 'n' }"; }
+              ];
+            }
+            # beforeAll, before, on filetype
+            {
+              __unkeyed-1 = "vimtex";
+              ft = [ "plaintex" ];
+              beforeAll = # lua
+                ''
+                  function()
+                    vim.g.vimtex_compiler_method = "latexrun"
+                  end
+                '';
+              before = # lua
+                ''
+                  function()
+                    vim.g.vimtex_compiler_method = "latexmk"
+                  end
+                '';
+            }
+            # On event
+            {
+              __unkeyed-1 = "nvim-biscuits";
+              event.__raw = "{ 'BufEnter *.lua' }";
+              after.__raw = ''
+                function()
+                require('nvim-biscuits').setup({})
+                end
+              '';
+            }
+            # On command no setup function, priority
+            {
+              __unkeyed-1 = "telescope.nvim";
+              cmd = [ "Telescope" ];
+              priority = 500;
+            }
+            # On colorschme
+            {
+              __unkeyed-1 = "onedarker.nvim";
+              colorscheme = [ "onedarker" ];
+            }
+            # raw value
+            {
+              __raw = ''
+                {
+                    "crates.nvim",
+                    ft = "toml",
+                }
+              '';
+            }
+          ];
+        };
+      };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -119,6 +119,7 @@
 
     ./pluginmanagers/packer.nix
     ./pluginmanagers/lazy.nix
+    ./pluginmanagers/lz-n.nix
 
     ./snippets/friendly-snippets.nix
     ./snippets/luasnip


plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+            map (
+              keymap:
+              {
+                __unkeyed_1 = keymap.key;
+                __unkeyed_2 = keymap.action;
+                inherit (keymap) mode;
+              }
+              // keymap.options
+            ) keymaps;
+        pluginToLua = plugin: {
```
MattSturgeon, line 121: IMO this name is confusing, since it returns normalised attrs rather than lua code.

Maybe `processPlugin` or `normalisePlugin`?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+            beforeAll
+            before
+            after
+            event
+            cmd
+            ft
+            colorscheme
+            priority
+            load
+            ;
```
MattSturgeon, line 133: Would it be simpler to do something like

```nix
removeAttrs plugin [ "name" "enabledInSpec" ] // {
  __unkeyed = plugin.name;
  enabled = plugin.enabledInSpec;
  keys = processKeymap plugin.keys;
};
```

?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+          enabled = plugin.enabledInSpec;
+          keys = processKeymap plugin.keys;
+        };
+        pluginListToLua = map pluginToLua;
+        plugins = pluginListToLua cfg.plugins;
```
MattSturgeon, line 138: I think this is readable enough when inlined:

```suggestion
        plugins = map pluginToLua cfg.plugins;
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+        pluginSpecs = if length plugins == 1 then head plugins else plugins;
```
MattSturgeon, line 139: Is this important? Would it still work fine to pass a singleton list as a list?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+              or null "Can be used to override the vim.g.lz_n.load() function for ${originalName}.";
+          };
+        };
+      default = pluginDefault;
+    };
+in
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  inherit name originalName;
+  maintainers = with helpers.maintainers; [ psfloyd ];
```
MattSturgeon, line 70: Nit: no need for `with` on a singleton list
```suggestion
  maintainers = [ helpers.maintainers.psfloyd ];
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+  '';
+
+  settingsOptions = {
+    load = mkNullOrLuaFn' {
+      description = ''
+        Function used by `lz.n` to load plugins.
+      '';
+      default = null;
+      pluginDefault = "vim.cmd.packadd";
+    };
```
MattSturgeon, line 86:
```suggestion
    load = helpers.defaultNullOpts.mkStrLuaFnOr "vim.cmd.packadd" ''
      Function used by `lz.n` to load plugins.
    '';
```

MattSturgeon (commented):
Thanks for working on this!

It's looking good, I've just left a few comments below to hopefully improve things further

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,151 @@
...
+      pluginDefault = {
+        name = originalName;
+      };
+    in
+    mkOption {
+      description = "Lazy-load settings for ${originalName}.";
+      type =
+        with nixvimTypes;
+        let
+          nullStrToList = coercedTo str toList (listOf str);
```
MattSturgeon, line 34: I'm _unsure_ a coerced type makes sense here, _ideally_ we want `plugins` to closely reflect the lz.n's config schema without too much magic on our end.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,147 @@
...
+            priority = mkNullable number pluginDefault.priority or null ''
+              Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
+              Default priority is 50 (or 1000 if colorscheme is set).
+            '';
+            load = mkLuaFn pluginDefault.load
+              or null "Can be used to override the vim.g.lz_n.load() function for ${originalName}.";
+          };
+        };
+      default = pluginDefault;
+    };
```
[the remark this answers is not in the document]
psfloyd: Ok, i am going to simplify `lzPluginType`, to only describe the options required by lz-n

plugins/pluginmanagers/lz-n.nix
psfloyd: When there is only one plugin in the spec, lz-n does not process it properly if it is nested.

psfloyd:
I have made the changes suggested by the review.

A question more about GitHub ettiquette. Should I keep these changes in its own commit, or should i squash them to the original commit?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+        after = helpers.mkNullOrLuaFn "Executed after this plugin is loaded.";
+        event = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
+        cmd = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on command.";
+        ft = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on filetype.";
+        colorscheme = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on colorscheme.";
+        keys = helpers.mkNullOrOption (listOf helpers.keymaps.mapOptionSubmodule) "Lazy-load on key mapping. Use the same format as `config.keymaps`.";
+        priority = helpers.mkNullOrOption number ''
+          Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
+          Default priority is 50 (or 1000 if colorscheme is set).
+        '';
```
MattSturgeon, line 33:
```suggestion
        priority = helpers.defaultNullOpts.mkUnsignedInt (lib.literalMD "`50` (or `1000` if `colorscheme` is set)") ''
          Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
        '';
```

We should try and use `defaultNullOpts` and provide a plugin default when defining options that directly map to upstream's options.

Even if the plugin default is unknown, you can use the `'` variant (e.g. `mkUnsignedInt'`) and not provide `pluginDefault`, though this is _effectively_ the same as using the non-`defaultNullOpts` option helpers.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+    load = helpers.defaultNullOpts.mkLuaFn "vim.cmd.packadd" ''
+      Function used by `lz.n` to load plugins.
+    '';
+  };
+
+  callSetup = false; # Does not use setup
+
+  extraOptions = {
+    plugins = mkOption {
+      description = "List of plugins processed by lz.n";
```
MattSturgeon, line 60:
```suggestion
      description = "Plugin specs provided to the `require('lz.n').load` function.";
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+      default = [ ];
+      type = types.listOf lzPluginType;
```
MattSturgeon, line 62: This would benefit from an (ideally real) example:

```nix
example = [
  {
    name = "some_plugin.nvim";
    # etc
  }
];
```

MattSturgeon, line 62: It'd be good if this accepted "rawLua" types, e.g. using `helpers.nixvimTypes.maybeRaw` or by adding `helpers.nixvimTypes.rawLua` to a `lib.types.oneOf`.

MattSturgeon: I think it'd also be fine to have `coercedTo str toList (listOf str)` here, since:

a) upstream accept either str or table of str
b) we need a list type for merging

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+          "name"
+          "enabledInSpec"
+        ]
+        // {
+          __unkeyed = plugin.name;
+          enabled = plugin.enabledInSpec;
+          keys = processKeymaps plugin.keys;
+        };
+      plugins = map processPlugin cfg.plugins;
+      pluginSpecs = if length plugins == 1 then head plugins else plugins;
```
MattSturgeon, line 95: We can avoid _this_ specific bit of magic here by moving it to the option's `apply` function:

```nix
apply = v: if length plugins == 1 then head plugins else plugins;
```

Although, IDK if that's a good idea... It would affect what users see when they lookup `config.plugins.lz-n.plugins`.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+        plugin:
+        removeAttrs plugin [
+          "name"
+          "enabledInSpec"
+        ]
+        // {
+          __unkeyed = plugin.name;
+          enabled = plugin.enabledInSpec;
+          keys = processKeymaps plugin.keys;
+        };
```
MattSturgeon, line 93: I'm not a big fan of having this processing step; it de-couples our options from what upstream actually expects.

I appreciate the mixed list/attrs table format is awkward to represent in nix, so I understand why you've done it this way. If this were a `settings` option though, I'd be saying "this doesn't comply with rfc42".

The per-plugin options added in #1875 could have its own schema that doesn't relate to lz.n's, because that will be _our_ options that we map internally to whatever plugin is implementing lazy-loading.

Here, however, we are attempting to represent a plugin's (lz.n's) upstream options, so I think it is best to follow rfc42.

@traxys @GaetanLepage what's your view?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+        else
+          map (
+            keymap:
+            {
+              __unkeyed_1 = keymap.key;
+              __unkeyed_2 = keymap.action;
+              inherit (keymap) mode;
+            }
+            // keymap.options
+          ) keymaps;
```
MattSturgeon, line 82: Again, while it's nice to use our normal keymap submodule, I think it is better to follow rfc42 and represent the plugin's options directly.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+let
+  lzPluginType =
+    with helpers.nixvimTypes;
+    submodule {
+      options = {
```
MattSturgeon, line 13:
```suggestion
    submodule {
      freeformType = attrsOf anything;
      options = {
```

To be truly RFC42, and to allow us to omit options for anything unimportant and/or likely to change upstream, we should make this a _freeform_ submodule.

This allows users to define config even if we haven't declared a corresponding option (so long as their config matches either the freeform type or one of our declared options, it is valid).

MattSturgeon (changes requested):
> I have made the changes suggested by the review.

Thanks!

> A question more about GitHub ettiquette. Should I keep these changes in its own commit, or should i squash them to the original commit?

Personally, it doesn't bother me either way.

Having separate commits can be beneficial since that way you can more easily tell what has changed, although it's still possible to diff force-pushes.

The most important thing is that it is easy to squash the changes once the PR has moved on enough that having separate commits is no longer needed.

On more complex PRs you can end up with conflicts when squashing lots of small changes that were committed in a different order.

Also, if you're not aware, `git commit --fixup` and `git rebase -i --autosquash` can be very useful.

---

I've focused this review on the `plugins` option and attempting to make it fit better with RFC42.

See [this section](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md#declaring-plugin-options) of our contributing guide, or [the RFC itself](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md).

For some of the discussion points I'd appreciate input from the other maintainers @nix-community/nixvim

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,106 @@
...
+        name = mkOption {
+          type = str;
+          description = "The plugin's name (not the module name). This is what is passed to the load(name) function.";
+        };
+        enabledInSpec = helpers.mkNullOrStrLuaFnOr bool ''
```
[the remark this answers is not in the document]
psfloyd: I had changed the name to avoid confusion with the `enable` option on the other PR, but here it is not necesary.

```suggestion
        enabled = helpers.defaultNullOpts.mkStrLuaFnOr bool true ''
```
Upstream accepts a lua function that returns a boolean for this option. Wouldn't `mkStrLuaFnOr` be more apropiate?

plugins/pluginmanagers/lz-n.nix
psfloyd: > It'd be good if this accepted "rawLua" types, e.g. using `helpers.nixvimTypes.maybeRaw` or by adding `helpers.nixvimTypes.rawLua` to a `lib.types.oneOf`.

The problem I found with using `maybeRaw` o `oneOf` is that documentation does not get generated for `lz-n.plugins.*.option`.

This is what I tested:

```nix
type = with helpers.nixvimTypes; maybeRaw (coercedTo str toList (listOf lzPluginType));
```

```nix
type = with helpers.nixvimTypes; oneOf [ rawLua (coercedTo str toList (listOf lzPluginType))];
```

plugins/pluginmanagers/lz-n.nix
psfloyd: When removing magic for `enabledInSpec` and `keys`, the only thing that is left is to set the name as unkeyed.

psfloyd:
Thank you for your review and guidance!

I made most of the changes suggested. Some things still need input from reviewers.

plugins/pluginmanagers/lz-n.nix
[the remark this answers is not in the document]
traxys: I think it would be better if upstream accepts that

plugins/pluginmanagers/lz-n.nix
traxys: I'd prefer to avoid having any processing that changes the shape of arguments passed to plugins, it's confusing when people look at examples that are in lua initially.
Though our documentation & tooling on unkeyed tables may be lacking, but we should strive to fix that if it's the case

plugins/pluginmanagers/lz-n.nix
```diff
@@ -27,7 +27,22 @@ let
         cmd = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on command.";
         ft = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on filetype.";
         colorscheme = helpers.mkNullOrOption (either str (listOf str)) "Lazy-load on colorscheme.";
-        keys = helpers.mkNullOrOption (listOf helpers.keymaps.mapOptionSubmodule) "Lazy-load on key mapping. Use the same format as `config.keymaps`.";
+        keys =
+          with helpers.nixvimTypes;
+          let
+            singleStrToLuaObj =
+              s: if hasInfix "{" s then helpers.mkRaw s else helpers.toLuaObject (helpers.mkRaw s);
+            keysType = coercedTo str singleStrToLuaObj strLua;
```
MattSturgeon, line 35: Not sure coerced is appropriate here, unless you're using it for a particular reason?

If we're matching upstream's spec they either accept `str` or they don't. Either way we shouldn't need to do any coercion. The `either` or `oneOf` type should be ok if `str` is accepted.

MattSturgeon: Ah, I see. You're using it because you want `strLua` **and** `str`, and you are detecting lua based on table syntax being used.

While clever, and helpful in most scenarios, I think we should use `maybeRaw str` here (equivalent to `either str rawLua`).

Unfortunately this will be affected by #1618, but that should be fixed separately.

1. Detecting `"{"` will fail if that char is used as part of an actual keymapping.
2. We're already debating amongst maintainers whether `strLua` is too magic to use at all
3. If there's any confusion with actual string types, we should not use `strLua`

plugins/pluginmanagers/lz-n.nix
MattSturgeon: If we're using a coercedTo, then we're already modifying the user's definition. But it's probably better if we can keep things simple and document the option as always being a list (in the final merged config)?

Ideally, we'd get upstream to accept singleton list tables, however for now I think this one processing step is ok.

plugins/pluginmanagers/lz-n.nix
MattSturgeon: #1618 is the issue in question. We're struggling to get to the bottom of it, but if a type really should be a maybeRaw submodule, we should make it such.

One caveat: if a type is `attrs` or `attrsOf anything` then `{__raw = "";}` will pass the `check` function. While it'd be undocumented, it would be supported to use raw lua with such a type even without wrapping it in `maybeRaw`.

Another workaround is to use a custom type with a custom `check` function, but this is cumbersome.

plugins/pluginmanagers/lz-n.nix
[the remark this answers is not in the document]
MattSturgeon: Well, `mkBool` will already return a `maybeRaw bool` type, so I don't think it's really necessary to support `strLuaFn` here.

This may just be my personal opinion, but the consistency of always using `__raw` or `helpers.mkRaw` is good for knowing what is lua code and what is a string.

plugins/pluginmanagers/lz-n.nix
MattSturgeon: I often mull over different ways to represent mixed list-attrs tables in nix, there's pros/cons to various approaches but none that map _super_ intuitively.

Something like this may be nice, but I don't think it's worth the effort migrating away from `__unkeyed-N`:

```nix
{
  _type = "lua-table";
  items = [ ];
  named = { };
}
```

---

Either way, while it is a little ugly, I agree we should avoid the processing step and let users define exactly what upstream expects via our option.

Generally speaking, having users come across `__raw` and `__unkeyed` in documentation and examples is good for exposure, so they can later figure out how to translate lua into nix.

For this plugin specifically, it's likely users won't have to interact with it directly, since #1875 will be more user-facing.

psfloyd:
I think I resolved most of the concerns. All the magic for setting options was removed. 
All trigger options now accept rawLua.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,126 @@
...
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+let
+  lzPluginType =
+    with helpers.nixvimTypes;
+    let
+      maybeRawStrOrList = either (maybeRaw str) (listOf (maybeRaw str));
```
MattSturgeon, line 13:
```suggestion
      maybeRawStrOrList =
        oneOf [
          str
          (listOf (maybeRaw str))
          rawLua
        ];
```

This is the same thing, but with fewer nested `either`s (`maybeRaw` is just `t: either t rawLua`)

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,126 @@
...
+    in
+    submodule {
+      freeformType = attrsOf anything;
+      options = {
+        __unkeyed = mkOption {
+          type = str;
+          description = "The plugin's name (not the module name). This is what is passed to the load(name) function.";
```
MattSturgeon, line 20: This description feels a little clunky? Can't quite put my finger on it though...

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,126 @@
...
+          },
+        '')
+      ];
+    };
+  };
+
+  extraConfig =
+    cfg:
+    let
+      pluginSpecs = if length cfg.plugins == 1 then head cfg.plugins else cfg.plugins;
```
MattSturgeon, line 116: Could you add a comment explaining why we do this and that it'd be nice if upstream didn't require us to?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,126 @@
...
+    in
+    {
+      globals.lz_n = cfg.settings;
```
MattSturgeon, line 119: It'd be nice if we could copy over any option priorities too (e.g. `mkDefault` or `mkForce`).

[`mkDerivedConfig`](https://noogle.dev/f/lib/modules/mkDerivedConfig) does this for a single option, but here I think we need [`mkAliasAndWrapDefsWithPriority`](https://github.com/NixOS/nixpkgs/blob/585a077fd2b54233843f824f0660599185de1528/lib/modules.nix#L1074-L1078).

That's the function [used by](https://github.com/NixOS/nixpkgs/blob/585a077fd2b54233843f824f0660599185de1528/lib/modules.nix#L1364) `doRename` (which is in-turn used by `mkRenamedOptionModule` and `mkAliasOptionModule`).

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+    in
+    submodule {
+      freeformType = attrsOf anything;
+      options = {
+        __unkeyed = mkOption {
+          type = str;
+          description = ''
+            The plugin's name (not the lua module name). This is what is passed to the load(name) function.
+            This is the name of the folder in /pack/*/opt/{name}.
+            See :h packadd
```
MattSturgeon, line 28: Minor rewording and formatting suggestions:

```suggestion
            The "unkeyed" attribute is the plugin's name.
            This is passed to the `load` function and should match what would normally be passed to `require`.
            
            More specifically, this is the name of the folder in `/pack/*/opt/{name}`.
            See `:h packadd`.
```

Of course please iterate further if you want to!

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+      maybeRawStrOrList = oneOf [
+        str
+        (listOf (maybeRaw str))
+        rawLua
+      ];
+    in
+    submodule {
+      freeformType = attrsOf anything;
+      options = {
+        __unkeyed = mkOption {
```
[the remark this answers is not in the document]
MattSturgeon: I don't think we've ever had an "unkeyed" option definition before... So far all unkeyed attrs have been in freeform definitions.

I can't see a better alternative, it just feels as though we're now more publicly exposing the messy side of trying to map nix values to lua tables. And in doing so committing more solidly to our current strategy? :face_with_diagonal_mouth: 

@traxys are you happy with this approach?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+        enabled = helpers.defaultNullOpts.mkStrLuaFnOr bool true ''
+          When false, or if the function returns false, then this plugin will not be included in the spec.
+          This option corresponds to the `enabled` property of lz.n.
+        '';
+        beforeAll = helpers.mkNullOrLuaFn "Always executed before any plugins are loaded.";
+        before = helpers.mkNullOrLuaFn "Executed before this plugin is loaded.";
+        after = helpers.mkNullOrLuaFn "Executed after this plugin is loaded.";
+        event = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
+        cmd = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on command.";
+        ft = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on filetype.";
```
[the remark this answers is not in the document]
MattSturgeon: Also, remember that this is a freeform submodule: we don't need to have options for _all_ upstream options since they can be defined via the freeform type.

I don't mind having them, so long as we are confident they won't become a maintenance burden down the line. This happens when options are removed, renamed or re-typed upstream and we have to update accordingly.

The alternative is to document them as "examples" instead of as sub-options. I'm **not pushing** for you to change this, just prompting the discussion!

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+            (helpers.mkRaw ''{ "<leader>ft", "<CMD>Neotree toggle<CR>", desc = "NeoTree toggle" }'')
+          ];
+        };
+        priority = helpers.defaultNullOpts.mkUnsignedInt (lib.literalMD "`50` (or `1000` if `colorscheme` is set)") ''
+          Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
+        '';
+        load = helpers.mkNullOrLuaFn "Can be used to override the `vim.g.lz_n.load()` function for this plugin.";
+      };
+    };
+in
```
MattSturgeon, line 56: Not sure, but maybe the file would be more readable with this let-block on the `extraOptions` attr instead of at the head of the file?

Not too important since I think we need _some_ parts of the let block up here anyway.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+              require("neo-tree").setup()
+            end
+          '';
+        }
+        {
+          __unkeyed = "telescope.nvim";
+          cmd = [ "Telescope" ];
+          keys = [
+            (helpers.mkRaw ''{ "<leader>fa", "<CMD>Telescope autocommands<CR>", desc = "Telescope autocommands" }'')
+            (helpers.mkRaw ''{ "<leader>fb", "<CMD>Telescope buffers<CR>", desc = "Telescope buffers" }'')
```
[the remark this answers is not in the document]
MattSturgeon: Side note: no need for a `''` string here, unless you felt strongly about using `"` instead of `'` in the example.

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,140 @@
+{ pkgs, helpers, ... }:
+let
+  inherit (pkgs) lib;
```
MattSturgeon, line 3: This could be moved to "module" args in the respective test-cases to avoid needing to do `inherit (pkgs) lib`.

To do this you add a `module` attr in the test case which contains the actual test case's module function:

```nix
single-plugin = {
  module =
    {
      config,
      options,
      pkgs,
      lib,
      helpers,
      ...
    }:
    {
      # Test config module
    };
};
```

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,140 @@
...
+  optionalPlugins = map (plugin: {
+    inherit plugin;
+    optional = true;
+  });
```
MattSturgeon, line 7: This function would benefit from a comment regarding it's purpose: what it is for, why it is needed and (if not obvious) how it works.

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,140 @@
...
+            {
+                "crates.nvim",
+                ft = "toml",
+            },
+          ''
+        )
+      ];
+    };
+  };
+
```
MattSturgeon, line 139:
```suggestion
```

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,140 @@
...
+          after = # lua
+            ''
+              function()
+                require("neo-tree").setup()
+              end
+            '';
+        }
+      ];
+    };
+
```
MattSturgeon, line 44:
```suggestion
```

MattSturgeon (changes requested):
A few nits, but mostly just discussion points.

The main blocker now is discussing the best way to handle documenting unkeyed attrs as part of a well-defined submodule.

The PR itself is in good shape, well done! :tada:

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+      ];
+    };
+  };
+
+  extraConfig = cfg: {
+    globals.lz_n = modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
+    extraConfigLua = mkIf (cfg.plugins != [ ]) ''
+      require('lz.n').load(
+        ${helpers.toLuaObject cfg.plugins}
+      )
```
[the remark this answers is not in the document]
MattSturgeon: Did it turn out we don't need to handle singleton lists? Or is this now handled elsewhere?

plugins/pluginmanagers/lz-n.nix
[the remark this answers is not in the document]
MattSturgeon: (We could even go to the extreme of not having a submodlue and instead doing `maybeRaw (attrsOf anything)`)

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,130 @@
...
+        '';
+        beforeAll = helpers.mkNullOrLuaFn "Always executed before any plugins are loaded.";
+        before = helpers.mkNullOrLuaFn "Executed before this plugin is loaded.";
+        after = helpers.mkNullOrLuaFn "Executed after this plugin is loaded.";
+        event = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
+        cmd = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on command.";
+        ft = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on filetype.";
+        colorscheme = helpers.mkNullOrOption maybeRawStrOrList "Lazy-load on colorscheme.";
+        keys = helpers.mkNullOrOption' {
+          type = maybeRawStrOrList;
```
MattSturgeon, line 43: This type being `listOf str` instead of `listOf anything` is the only reason your examples need to use raw lua. I don't see why users shouldn't be able to include attrs or lists in the keys list without resorting to raw lua?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,170 @@
...
+      plugins = mkOption {
+        description = ''
+          List of plugin specs provided to the `require('lz.n').load` function.
+          Plugin specs can be ${helpers.nixvimTypes.rawLua.description}.
+        '';
+        default = [ ];
+        type = types.listOf lzPluginType;
+        example = [
+          {
+            __unkeyed = "neo-tree.nvim";
```
MattSturgeon, line 115: Same 
```suggestion
            __unkeyed-1 = "neo-tree.nvim";
```

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,161 @@
...
+let
+  /*
+    Transform plugins into attrset and set optional to true
+    This installs the plugin files to {runtimepath}/pack/opt, instead of {runtimepath}/pack/start.
+    Plugins in pack/opt are not loaded on startup, but can be later loaded with `:packadd {name}` or `require("lz.n").load({name})`
+
+    See `nixpkgs/pkgs/applications/editors/neovim/utils.nix`
+    See `nixpkgs/pkgs/applications/editors/vim/plugins/vim-utils.nix`
+  */
+  optionalPlugins = map (x: (if (x ? plugin) then x else { plugin = x; }) // { optional = true; });
```
MattSturgeon, line 11:
```suggestion
  optionalPlugins = map (x: (if x ? plugin then x else { plugin = x; }) // { optional = true; });
```

Sorry to nit

tests/test-sources/plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,161 @@
...
+          before = # lua
+            ''
+              function()
+                vim.g.vimtex_compiler_method = "latexmk"
+              end
+            '';
+        }
+        # On event
+        {
+          __unkeyed = "nvim-biscuits";
```
MattSturgeon, line 130:
```suggestion
          __unkeyed-1 = "nvim-biscuits";
```

Same as option declaration and settings-example

MattSturgeon (commented):
Thanks, I think this is pretty much done. There's just a couple thinks I'd like a second opinion on from another maintainer.

Other than that it's just nits below. Now the PR is pretty much done, could you try and keep the commits squashed to just the two "add maintainer" and "add plugin" commits?

Also, sorry this took a while for me to look at. If it's ready for review and you have been waiting, don't hesitate to ping one of us, since it's easy for individual PRs to get forgotten about otherwise!

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,170 @@
...
+  options,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "lz-n";
+  originalName = "lz.n";
+  maintainers = [ helpers.maintainers.psfloyd ];
```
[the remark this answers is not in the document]
MattSturgeon: On a related note: `helpers` should now be accessed as `lib.nixvim` (`helpers` still works, but new code should use `lib.nixvim`)

psfloyd:
Great, thanks for taking a look at it again. I applied all the changes and some other style changes suggested by GaetanLepage on #1979. I sqashed all comits as requested.

No worries about the wait, I took that time to improve my personal nixvim config ;).

Regarding the `strLua` options, if it is necesary I can reafactor the options with another type. I prefer to have a stable interface for #1875.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,200 @@
+{
+  lib,
+  options,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+let
+  inherit (lib.nixvim) defaultNullOpts;
```
MattSturgeon, line 10: Good idea! If we're gonna inherit `defaultNullOpts`, maybe also add `inherit (lib) types`? I think that's pretty much the only thing that is currently justifying the `with lib`?

Not important, just a suggestion. Feel free to ignore or wait for feedback from the other maintainers.

MattSturgeon:
> Regarding the `strLua` options, if it is necesary I can reafactor the options with another type. I prefer to have a stable interface for #1875.

I think it's fine. Nixvim is _full_ of strLua, so this won't make any eventual deprecation notably worse.

I think other maintainers will take a look over the next few days :crossed_fingers:

MattSturgeon:
> Great, thanks for taking a look at it again. I applied all the changes and some other style changes suggested by GaetanLepage on #1979. I sqashed all comits as requested.

Thanks for your persistent work & dedication on these PRs!

MattSturgeon:
```
cannot build on 'ssh-ng://nix@build04.nix-community.org': error: failed to start SSH connection to 'nix@build04.nix-community.org'
Failed to find a machine for remote build!
derivation: 39zgfnw2zax5d7qyr7894a99dizldi7i-plugins-utils-mark-radar.drv
required (system, features): (aarch64-linux, [])
2 available machines:
(systems, maxjobs, supportedFeatures, mandatoryFeatures)
([aarch64-linux], 80, [benchmark, big-parallel, gccarch-armv8-a, kvm, nixos-test], [])
([aarch64-darwin, x86_64-darwin], 8, [big-parallel], [])
error: a 'aarch64-linux' with features {} is required to build '/nix/store/39zgfnw2zax5d7qyr7894a99dizldi7i-plugins-utils-mark-radar.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

Seems like a buildbot issue? Just strange it's only happening on this PR. Will try rebuilding.

psfloyd:
> I think other maintainers will take a look over the next few days 🤞

Great!

> Thanks for your persistent work & dedication on these PRs!

Thanks to all of you for maintaining this awesome project.

> Seems like a buildbot issue? Just strange it's only happening on this PR. Will try rebuilding.

I saw the error, but since it was not in the lz-n tests I didn't know how to troubleshoot it.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -0,0 +1,200 @@
...
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
```
GaetanLepage, line 12:
```suggestion
nixvim.neovim-plugin.mkNeovimPlugin config {
```

GaetanLepage (changes requested):
Minor nit, otherwise good to me !

MattSturgeon (approved):
LGTM, are you happy @GaetanLepage?

GaetanLepage (approved):
Yes, let's go !

GaetanLepage:
@mergifyio queue

GaetanLepage:

mrcjkb:
@psfloyd just wanted to let you know that I've added a new `lazy` field to the `lz.n.PluginSpec`.
It's meant to allow to trigger loading of plugins from within the `before` or `after` hooks of other plugins.

PR #1875, opened 2024-07-16
lib/neovim-plugin: Add lazyLoad option to plugins

psfloyd:
This PR is the second split of PR #1866, it works on a solution for the issue #421.
This PR depends on PR #1874.
All suggestions from the original PR are included.

# Design description
The idea is to automatically create `lazyLoad` option for all plugins that are defined with `mkNeovimPlugin`.
`lazyLoad` is conformed by triggers (e.g. filetype, event, key, command) and actions (e.g. functions to run before/after loading the plugin). `mkNeovimPlugin` would also generate default values for some of these options (e.g. the setup).

Plugin maintainers can decide to not create the `lazyLoad` option, by setting `allowLazyLoading` to false when defining the plguin.

Plugin maintainers can provide sensible default values to `lazyLoad` when defining a plugin. For example, if defining `telescope.nvim`, set `Telescope` as a command trigger. Also defining a custom setup function if they are setting `callSetup` as false.

Finnally users can :
- Enable lazy-loading (globally or on a per-plugin basis)
- Add to these triggers (or override them with `mkForce`) on a per-plugin basis.
- Override actions on a per-plugin basis.

# Progress
- [x] Adding `mkLazyLoadOption` to generate lazyLoad options for plugins.
- [x] Conditionally enabling lazy or regular loading depending on user settings in `mkNeovimPlugin`.
- [x] Adding and implementing a lazy-loading mechanism using [lz.n](https://github.com/nvim-neorocks/lz.n).
- [ ]  Adding and implementing a lazy-loading mechanism using [lazy.nvim](https://github.com/folke/lazy.nvim).
- [ ] Assert that only one of the backends is enabled.
- [ ] Conditionally enabling lazy or regular loading depending on user settings in `mkVimPlugin`.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -44,6 +44,7 @@ let
       defaultNullOpts
       mkCompositeOption
       mkCompositeOption'
+      mkLazyLoadOption
       mkNullOrLua
       mkNullOrLua'
       mkNullOrLuaFn
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -43,9 +43,14 @@ with lib;
       extraPackages ? [ ],
       callSetup ? true,
       installPackage ? true,
+      # lazyLoad
+      allowLazyLoad ? true,
+      packageName ? originalName, # Name of the package folder created in {runtimepath}/pack/start or {runtimepath}/pack/opt
+      lazyLoadDefaults ? { },
     }:
     let
       namespace = if isColorscheme then "colorschemes" else "plugins";
+      cfg = config.${namespace}.${name};
     in
     {
       meta = {
@@ -94,25 +99,90 @@ with lib;
             example = settingsExample;
           };
         }
+        // optionalAttrs allowLazyLoad {
+          lazyLoad = helpers.mkLazyLoadOption {
+            inherit originalName;
+            lazyLoadDefaults =
+              (optionalAttrs (isColorscheme && colorscheme != null) { inherit colorscheme; }) // lazyLoadDefaults;
+          };
+        }
         // extraOptions;
 
       config =
         let
-          cfg = config.${namespace}.${name};
           extraConfigNamespace = if isColorscheme then "extraConfigLuaPre" else "extraConfigLua";
+          lazyLoaded = cfg.lazyLoad.enable or false;
+
+          # lz-n lazyLoad backend
+          # Transform plugin into attrset and set optional to true
+          # See `tests/test-sources/plugins/pluginmanagers/lz-n.nix`
+          optionalPlugin =
+            x:
+            if isColorscheme then x else ((if x ? plugin then x else { plugin = x; }) // { optional = true; });
+          mkFn = str: if str != "" then "function()\n" + str + "end" else null;
         in
         mkIf cfg.enable (mkMerge [
-          {
-            extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
-            inherit extraPackages;
-          }
-          (optionalAttrs callSetup {
-            ${extraConfigNamespace} = ''
-              require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
-            '';
-          })
+          # Always set
           (optionalAttrs (isColorscheme && (colorscheme != null)) { colorscheme = mkDefault colorscheme; })
-          (extraConfig cfg)
+
+          # Normal loading
+          (mkIf (!lazyLoaded) (mkMerge [
+            (extraConfig cfg)
+            {
+              extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
+              inherit extraPackages;
+            }
+            (optionalAttrs callSetup {
+              ${extraConfigNamespace} = ''
+                require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
+              '';
+            })
+          ]))
+
+          # Lazy loading with lz-n
+          (mkIf (lazyLoaded && config.plugins.lz-n.enable) (mkMerge [
+            (lib.removeAttrs (extraConfig cfg) [
+              "extraConfigLua"
+              "extraConfigLuaPre"
+            ])
+            {
+              extraPlugins = (optional installPackage (optionalPlugin cfg.package)) ++ extraPlugins;
+              inherit extraPackages;
+            }
+            {
+              plugins.lz-n.plugins = [
+                {
+                  __unkeyed-1 = packageName;
+                  after =
+                    if cfg.lazyLoad.after == null then
+                      mkFn (
+                        (extraConfig cfg).extraConfigLuaPre or ""
+                        + optionalString callSetup ''require('${luaName}')${setup}(${
+                          optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)
+                        }) ''
+                        + (extraConfig cfg).extraConfigLua or ""
+                      )
+                    else
+                      cfg.lazyLoad.after;
+                  inherit (cfg.lazyLoad)
+                    enabled
+                    priority
+                    before
+                    beforeAll
+                    # after defined above
+                    event
+                    cmd
+                    ft
+                    keys
+                    colorscheme
+                    extraSettings
+                    ;
+                }
+              ];
+            }
+          ]))
+
+          # Lazy loading with lazy.nvim
         ]);
     };
 }
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -372,11 +372,6 @@ rec {
       originalName,
       lazyLoadDefaults ? { },
     }:
-    let
-      pluginDefault = {
-        enable = false;
-      } // lazyLoadDefaults;
-    in
     mkOption {
       description = ''
         Lazy-load settings for ${originalName}.
@@ -393,65 +388,71 @@ rec {
         submodule {
           options = with defaultNullOpts; {
 
-            enable = mkOption {
-              type = bool;
-              default = pluginDefault.enable;
-              description = ''
-                Enable lazy-loading for ${originalName}
-              '';
-            };
+            enable = mkEnableOption ''
+              lazy-loading for ${originalName}
+            '';
 
             # Spec loading:
-            enabled = mkStrLuaFnOr bool pluginDefault.enabledInSpec or null ''
+            enabled = mkStrLuaFnOr bool (lazyLoadDefaults.enabledInSpec or null) ''
               When false, or if the function returns false, then ${originalName} will not be included in the spec.
+
               Equivalence: lz.n => enabled; lazy.nvim => enabled
             '';
 
-            priority = mkNullable number pluginDefault.priority or null ''
+            priority = mkNullable number (lazyLoadDefaults.priority or null) ''
               Only useful for start plugins (not lazy-loaded) to force loading certain plugins first.
+
               Equivalence: lz.n => priority; lazy.nvim => priority
             '';
 
             # Spec setup
             # Actions
-            beforeAll = mkLuaFn pluginDefault.beforeAll or null ''
+            beforeAll = mkLuaFn (lazyLoadDefaults.beforeAll or null) ''
               Always executed before any plugins are loaded.
+
               Equivalence: lz.n => beforeAll; lazy.nvim => init
             '';
 
-            before = mkLuaFn pluginDefault.before or null ''
+            before = mkLuaFn (lazyLoadDefaults.before or null) ''
               Executed before ${originalName} is loaded.
+
               Equivalence: lz.n => before; lazy.nvim => None
             '';
 
-            after = mkLuaFn pluginDefault.after or null ''
+            after = mkLuaFn (lazyLoadDefaults.after or null) ''
               Executed after ${originalName} is loaded.
+
               Equivalence: lz.n => after; lazy.nvim => config
             '';
 
             # Triggers
-            event = mkNullable triggerType pluginDefault.event or null ''
+            event = mkNullable triggerType (lazyLoadDefaults.event or null) ''
               Lazy-load on event. Events can be specified as `BufEnter` or with a pattern like `BufEnter *.lua`
+
               Equivalence: lz.n => event; lazy.nvim => event
             '';
 
-            cmd = mkNullable triggerType pluginDefault.cmd or null ''
+            cmd = mkNullable triggerType (lazyLoadDefaults.cmd or null) ''
               Lazy-load on command.
+
               Equivalence: lz.n => cmd; lazy.nvim => cmd
             '';
 
-            ft = mkNullable triggerType pluginDefault.ft or null ''
+            ft = mkNullable triggerType (lazyLoadDefaults.ft or null) ''
               Lazy-load on filetype.
+
               Equivalence: lz.n => ft; lazy.nvim => ft
             '';
 
-            keys = mkNullable (listOf helpers.keymaps.mapOptionSubmodule) pluginDefault.keys or null ''
+            keys = mkNullable (listOf helpers.keymaps.mapOptionSubmodule) (lazyLoadDefaults.keys or null) ''
               Lazy-load on key mapping. Use the same format as `config.keymaps`.
+
               Equivalence: lz.n => keys; lazy.nvim => keys
             '';
 
-            colorscheme = mkNullable triggerType pluginDefault.colorscheme or null ''
+            colorscheme = mkNullable triggerType (lazyLoadDefaults.colorscheme or null) ''
               Lazy-load on colorscheme.
+
               Equivalence: lz.n => colorscheme; lazy.nvim => None
             '';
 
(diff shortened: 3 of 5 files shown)


plugins/languages/qmk.nix
```diff
@@ -15,7 +15,8 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   settingsOptions = {
     name = mkOption {
-      type = types.str;
+      type = with types; nullOr str;
+      default = null;
```
MattSturgeon, line 19: This seems unrelated?

plugins/languages/qmk.nix
```diff
@@ -24,7 +25,8 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     };
 
     layout = mkOption {
-      type = with types; listOf str;
+      type = with types; nullOr (listOf str);
+      default = null;
```
MattSturgeon, line 29: Same; unrelated?

lib/options.nix
```diff
@@ -368,4 +369,63 @@ rec {
...
+            keys = mkNullable (listOf nixvimKeymaps.mapOptionSubmodule) pluginDefault.keys
+              or null "Lazy-load on key mapping. Use the same format as `config.keymaps`.";
+            colorscheme = mkNullable (listOf str) pluginDefault.colorscheme or null "Lazy-load on colorscheme.";
+            priority = mkNullable number pluginDefault.priority or null ''
+              Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
+              Default priority is 50 (or 1000 if colorscheme is set).
+            '';
+            load = mkLuaFn pluginDefault.load
+              or null "Can be used to override the vim.g.lz_n.load() function for ${originalName}.";
+          };
```
MattSturgeon, line 427: How closely tied are these options to lz.n? Ideally the per-plugin options would be common to both lz.n, lazy.nvim and any other lazyload implementation we end up supporting in the future.

This could mean not making all options available per-plugin (users could still define them via the top-level option) and/or adding warnings/documentation indicating that some options only apply to specific implementation(s).

lib/neovim-plugin.nix
```diff
@@ -99,21 +103,47 @@ with lib;
...
-            ${extraConfigNamespace} = optionalString callSetup ''
+            ${extraConfigNamespace} = optionalString (callSetup && !lazyLoaded) ''
               require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
             '';
+
+            plugins.lz-n.plugins = mkIf lazyLoaded [ cfg.lazyLoad ];
+            assertions = [
+              {
+                assertion = config.plugins.lz-n.enable || !lazyLoaded;
+                # TODO: replace with global option after adding backend
```
MattSturgeon, line 140: Maybe instead of configuring a backend, we could just use whichever supported `plugins.*` is enabled?

With this approach, we might want to warn/assert that only _one_ such implementation is enabled?

lib/helpers.nix
```diff
@@ -8,13 +8,21 @@ let
...
   nixvimUtils = import ./utils.nix { inherit lib nixvimTypes _nixvimTests; };
-  nixvimOptions = import ./options.nix { inherit lib nixvimTypes nixvimUtils; };
+  nixvimOptions = import ./options.nix {
+    inherit
+      lib
+      nixvimTypes
+      nixvimUtils
+      nixvimKeymaps
+      ;
+  };
```
MattSturgeon, line 18: I really need to get around to refactoring this; it'd be much simpler if `helpers` itself was passed in (recursively)

MattSturgeon (commented):
Just my initial thoughts. Again, thanks for working on this!

lib/options.nix
psfloyd: > How closely tied are these options to lz.n? Ideally the per-plugin options would be common to both lz.n, lazy.nvim and any other lazyload implementation we end up supporting in the future.

Most options are shared by both lz.n and lazy.nvim, just with different names that would have to be adapted when setting them for each backend.
The differences are:
- `before`: is only defined by lz.n. lazy.nvim only allows `beforeAll`.
- `colorscheme`: lazy.nvim lazy-loads colorschemes automatically, so it wouldn't be set for it.
- `load`: This is specific for lz.n.




> This could mean not making all options available per-plugin (users could still define them via the top-level option) and/or adding warnings/documentation indicating that some options only apply to specific implementation(s).

What do you mean by "define them via the top-level option"?

Another idea I had is to have the subset of optrons that apply to all implementations at top-level, and a FreeForm option that is passed to the specific implementation being used. Just like `plugins.*.settings` is passed to `require(*).setup`.
What do you think fits best?

lib/neovim-plugin.nix
```diff
@@ -99,21 +103,47 @@ with lib;
...
 
-            ${extraConfigNamespace} = optionalString callSetup ''
+            ${extraConfigNamespace} = optionalString (callSetup && !lazyLoaded) ''
               require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
             '';
+
+            plugins.lz-n.plugins = mkIf lazyLoaded [ cfg.lazyLoad ];
+            assertions = [
+              {
+                assertion = config.plugins.lz-n.enable || !lazyLoaded;
```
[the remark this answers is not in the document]
psfloyd: You are right, I hadn't considered that.

plugins/languages/qmk.nix
psfloyd: Because of the current implementation, documentation can't be built if any `plugins.*.settings.option` is missing a default value. This was the only error I found.
Now that I think of it, this limitation is invasive to all other plugin definitions. I should try to remove it.

plugins/languages/qmk.nix
psfloyd: Same as above

psfloyd:
To keep things organized i am going to wait for PR #1874 to be stabilised to keep working on this.
Also I updated the top comment with a design description, to review the general concept.

lib/options.nix
MattSturgeon: > What do you mean by "define them via the top-level option"?

I mean that if `plugins.*.lazyload` doesn't provide the option a user needs, they can directly use `plugins.lazy` or `plugins.lz-n`.

> Most options are shared by both lz.n and lazy.nvim, just with different names that would have to be adapted when setting them for each backend.

My view here is that we should define our own options here, without trying to match lazy.nvim or lz.n _exactly_. We can then map whatever options we decided on to the implementing plugin's settings internally.

This should keep our per-plugin option de-coupled from the implementing plugins' options, allowing for future changes both on our end and in upstream plugins.

MattSturgeon: > Another idea I had is to have the subset of optrons that apply to all implementations at top-level, and a FreeForm option that is passed to the specific implementation being used. Just like `plugins.*.settings` is passed to `require(*).setup`.
> What do you think fits best?

I'm not sure I follow 100%, but freeform and rfc42 are usually the way to go when dealing directly with a plugin's options.

lib/neovim-plugin.nix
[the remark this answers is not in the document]
MattSturgeon: It was my suggestion to add assertions, so I'm the idiot here :joy:

MattSturgeon:
> Add to these triggers (or override them with `mkForce`) on a per-plugin basis.

Probably better to have our defaults set with `mkDefault`, allowing users to override without needing to resort to `mkForce`.

That's only a guideline though, not a rule. There may be times where a default is so sensible that it makes sense not to use `mkDefault`?

arunoruto:
Seems like #1874 got morgen! I am hyped for this feature 🙌🏻
I hope I can move from helix back to neovim when this is implemented!

lib/helpers.nix
MattSturgeon: This was resolved in #1948

psfloyd:
Now that #1874 has been merged I am going to rework this PR to work with that implementation. While working on #1874 I  discovered some flaws in the draft implementation of this PR, that need to be addressed.

I am also going to integrate with the new helpers implementation from #1948.

It also would be nice to integrate `lazyLoad` options with `configLua` from #1876, when/if that gets merged.

Later I am going to update the first comment to better reflect the tasks that need to be done.

psfloyd:
I refactored the `lazyLoad` options, decupling them from the specific lz-n backend.
The first commit [894e190](https://github.com/nix-community/nixvim/pull/1875/commits/894e190fb41780194564dfd2c94145afd5c76462) is just for testing purposes. It can be removed at the end.

lib/options.nix
```diff
@@ -366,4 +366,108 @@ rec {
...
+              default = pluginDefault.enable;
+              description = ''
+                Enable lazy-loading for ${originalName}
+              '';
+            };
+
+            # Spec loading:
+            enabled = mkStrLuaFnOr bool pluginDefault.enabledInSpec or null ''
+              When false, or if the function returns false, then ${originalName} will not be included in the spec.
+              Equivalence: lz.n => enabled; lazy.nvim => enabled
```
MattSturgeon, line 407:
```suggestion
              When false, or if the function returns false, then ${originalName} will not be included in the spec.

              Equivalence: lz.n => enabled; lazy.nvim => enabled
```

I think this would benefit from being in its own paragraph.

Note, in commonmark, a single linebreak is _not_ rendered in the output. You either need a blankline to cause a paragraph-break, or a `\` at the end of a line to treat the following newline as a line-break.

MattSturgeon (commented):
I know this is still a draft, so some of my feedback may be irrelevant. But here's some _early_ feedback in the hopes it helps.

Thank you for your continued efforts!

tests/test-sources/modules/lazy-load.nix
```diff
@@ -0,0 +1,11 @@
+{
+  example = {
+    plugins.nvim-snippets = {
+      enable = true;
+      lazyLoad = {
+        enable = true;
+        event = "InsertEnter";
+      };
+    };
+  };
```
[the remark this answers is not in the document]
psfloyd: Yes, the test sneaked through before the implementation. Now it is fixed

lib/options.nix
```diff
@@ -366,4 +366,108 @@ rec {
...
+
+  mkLazyLoadOption =
+    {
+      originalName,
+      lazyLoadDefaults ? { },
+    }:
+    let
+      pluginDefault = {
+        enable = false;
+      } // lazyLoadDefaults;
```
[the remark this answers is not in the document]
psfloyd: `mkEnableOption` should be better then. I agree about `enable` being always false by default

tests/test-derivation.nix
```diff
@@ -47,6 +48,8 @@ let
               dontRun ? false,
             }:
             lib.optionalString (!dontRun) ''
+              ln -s ${derivation} $out/${name}
```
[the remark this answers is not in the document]
psfloyd: The old implementation also didn't work for this kind of test. Because the linkFarm was of all the tests. 
With this change I can build a specific test and run that nixvim instance to test functionallity inside neovim.

Maybe I missed a better way to achieve this...

psfloyd:
I just pushed an implementation for lazy loading with lz-n. 
Because of the complexity and amount of lines, would it be better to split lazy loading implementations into a new file? Perhaps `lib/lazy-load`? What do you think?

lib/neovim-plugin.nix
```diff
@@ -94,25 +99,90 @@ with lib;
...
+              '';
+            })
+          ]))
+
+          # Lazy loading with lz-n
+          (mkIf (lazyLoaded && config.plugins.lz-n.enable) (mkMerge [
+            (lib.removeAttrs (extraConfig cfg) [
+              "extraConfigLua"
+              "extraConfigLuaPre"
+            ])
```
MattSturgeon, line 147: This isn't a robust solution, we really need something like #1876 so that we can assign plugin-config to an option and then decide here how to _use_ that option.

The problem is we're assuming `extraConfig` returns a "plain" attrset, instead of something like a `mkIf`, `mkOverride`, `mkMerge`, etc.

It's also kinda "pulling the rug out" from the module definition, which is expecting a config definition to actually add a definition to the respective option.

If we have an intermediate, plugin-specific option, this is more transparent and also exposes better hooks to end-users.

lib/neovim-plugin.nix
```diff
@@ -94,25 +99,90 @@ with lib;
...
+                  __unkeyed-1 = packageName;
+                  after =
+                    if cfg.lazyLoad.after == null then
+                      mkFn (
+                        (extraConfig cfg).extraConfigLuaPre or ""
+                        + optionalString callSetup ''require('${luaName}')${setup}(${
+                          optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)
+                        }) ''
+                        + (extraConfig cfg).extraConfigLua or ""
+                      )
```
MattSturgeon, line 164: Could this be the `after` option's `default`, instead of `null`? This `if then else` is _effectively_ setting a default.

Don't mind too much if it is set with `default` in mkOption, or if we use `lib.mkDefault` in a config definition. We should probably also have a `defaultText = literalExpression ""` too, for transparency.

lib/options.nix
```diff
@@ -366,4 +366,109 @@ rec {
         else
           example;
     };
+
+  mkLazyLoadOption =
+    {
+      originalName,
+      lazyLoadDefaults ? { },
```
MattSturgeon, line 373:
```suggestion
      pluginName,
      defaults ? { },
```

lib/neovim-plugin.nix
MattSturgeon: I guess this is more complex because `mkLazyLoadOption` isn't given an actual option path, so it can't (truly) know which `extraConfigLua` definitions are being used, etc. Same for `luaName`, `setup`, etc.

We can still _implement_ it as a `${namespace}.${name}.lazyload.after = mkDefault (mkFn "")`, we just would struggle to _document_ this in `defaultText`

tests/test-derivation.nix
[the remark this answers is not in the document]
MattSturgeon: > With this change I can build a specific test and run that nixvim instance to test functionality inside neovim.

I think what we want here is to be able to build a test's passthru to debug it. If the test fails, that's _because_ it can't be built.

If I understand passthru correctly, it would allow us to bypass building the test derivation and instead use the derivation that was being tested.

In the case of a linkFarm we'd probably be building a passthru of a passthru, russian doll style.

This is something we ought to tackle in another PR, especially as our test infrastructure is somewhat in flux currently: #2015 #1989 #1986 etc

lib/neovim-plugin.nix
```diff
@@ -43,9 +43,14 @@ with lib;
       extraPackages ? [ ],
       callSetup ? true,
       installPackage ? true,
+      # lazyLoad
+      allowLazyLoad ? true,
+      packageName ? originalName, # Name of the package folder created in {runtimepath}/pack/start or {runtimepath}/pack/opt
```
MattSturgeon, line 48: Is this not the same as `luaName`?

lib/neovim-plugin.nix
```diff
@@ -94,25 +99,90 @@ with lib;
...
           extraConfigNamespace = if isColorscheme then "extraConfigLuaPre" else "extraConfigLua";
+          lazyLoaded = cfg.lazyLoad.enable or false;
+
+          # lz-n lazyLoad backend
+          # Transform plugin into attrset and set optional to true
+          # See `tests/test-sources/plugins/pluginmanagers/lz-n.nix`
+          optionalPlugin =
+            x:
+            if isColorscheme then x else ((if x ? plugin then x else { plugin = x; }) // { optional = true; });
+          mkFn = str: if str != "" then "function()\n" + str + "end" else null;
```
MattSturgeon, line 122: `mkFn` feels like an over-abstraction

lib/neovim-plugin.nix
```diff
@@ -94,25 +99,90 @@ with lib;
...
+            }
+            (optionalAttrs callSetup {
+              ${extraConfigNamespace} = ''
+                require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
+              '';
+            })
+          ]))
+
+          # Lazy loading with lz-n
+          (mkIf (lazyLoaded && config.plugins.lz-n.enable) (mkMerge [
```
MattSturgeon, line 143: I don't think it actually matters whether `lz-n` is enabled; we can define the config either way, it'll just have no effect, right?

We should probably assert somewhere that _a supported lazy loading backend_ is enabled if the user has enabled lazyloading globally though.

MattSturgeon (commented):
Minor suggestions and attempts to understand the changes :slightly_smiling_face:

MattSturgeon:
> Because of the complexity and amount of lines, would it be better to split lazy loading implementations into a new file? Perhaps `lib/lazy-load`? What do you think?

I'm not sure. The stuff in `lib/options.nix` could be in its own file if you prefer.

I don't think the stuff in `mkNeovimPlugin` can be moved without major changes?

Issue #1407, opened 2024-04-12
[FEATURE] Append arbitrary lua code before or after generated plugin setup


becknik:
> This isn't a bug report, but rather a feature proposal/ discussion

## Description

In the generated `init.lua` file, I'd like to be able add some lua code before/ after (or even inside) of a plugins `setup()` call to tidy up the generated file's structure.

Adding the code to `extraCofnigLua` or `extraConfigLuaPre` just adds tons of code to the top/ bottom of the `init.lua` and therefore doesn't creates any tight coupling between it and the plugin it's relevant or even necessary for. This goes against what I saw on some neovim setup and also should help human readers to understand the `init.lua`, which might be shown as a setup-reference to plain old neovim users.

## Example

When trying to configure `plugins.cmp`, I'm in need of the following code snippet:

```lua
local symbols = {
  Variable = '𝛸',
  -- --snip--
}

local kinds = vim.lsp.protocol.CompletionItemKind
for i, kind in ipairs(kinds) do
  kinds[i] = symbols[kind] or kind
end

local signs = { Error = "󰅚 ", Warn = "󰀪 ", Hint = "󰌶 ", Info = " " }
for type, icon in pairs(signs) do
  local hl = "DiagnosticSign" .. type
  vim.fn.sign_define(hl, { text = icon, texthl = hl, numhl = hl })
end
```

I'm currently solving this as follows:

```nix
  extraConfigLuaPre = builtins.readFile ./cmp.lua;

  # https://github.com/hrsh7th/nvim-cmp
  plugins.cmp = {
    enable = true;
    autoEnableSources = true;
    settings = {
      # __raw = builtins.readFile ./cmp.lua;
      formatting.format = ''
          function(entry, vim_item)
            vim_item.kind = symbols[vim_item.kind]
            vim_item.menu = '[' .. entry.source.name .. ']'
            return vim_item
          end
      '';
      # --snip--
    };
  };
```

This results in the `symbols` map being created beginning from line 36 and `cmp` from line 243, and my config is rather small atm...

traxys:
We could introduce something like this indeed

PR #1876, opened 2024-07-16
lib/neovim-plugin: Introduce `plugins.*.luaConfig.*` options

traxys:
Resolves https://github.com/nix-community/nixvim/issues/1407

This commit adds a `plugins.<name>.config` section controlling the plugin specific configuration.
    
The section contains the internal `init` option, containing the plugin's initialization code.
    
It also contains the public `pre` and `post` options, that allow to add code before & after the `init` section

Finally, it contains the `final` option, being the concatenation of the three previous options.

I would have like to have a `location` option that allows to specify where to add the configuration option, it could be useful for rustaceanvim that needs to put the config in different places depending on `plugins.lsp.enable`, but that ended up in infinite recursions.
This option could also have been useful for lazy loading, as it may have enabled to add the configuration in a separate function to accomodate lazy loading.

The implementation is a bit less clean than I would have liked, as the `flash` plugin _already_ defines a `config` option, so I had to write some code to skip adding this new option for that plugin

## Old description

Those new options are scope to the plugin they help to configure. Those options should be preferred internally to using the raw `extraConfigLua(Pre|Post)?` option, as it allows being more flexible with how the configuration is generated.

This commit does not introduce any customization, but doing this can allow things like:
- Easily splitting the plugin in another lua file, for example for a specific filetype
- Passing the configuration to a lazy loading provider

This also allows users to define helper functions close to the plugins they are used into, allowing a more modular configuration

This improvement is only available to plugins using mkNeovimPlugin right now.

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -25,6 +25,11 @@
       # colorscheme
       isColorscheme ? false,
       colorscheme ? name,
+      # luaConfig
+      configLocation ? if isColorscheme then "extraConfigLuaPre" else "extraConfigLua",
+      # For some plugins it may not make sense to have a configuration attribute, as they are
+      # configured through some other mean, like global variables
+      hasConfigAttrs ? true,
       # options
       originalName ? name,
       # Can be a string, a list of strings, or a module option:
@@ -60,7 +65,14 @@
         let
           cfg = config.${namespace}.${name};
           opt = options.${namespace}.${name};
-          extraConfigNamespace = if isColorscheme then "extraConfigLuaPre" else "extraConfigLua";
+
+          setupCode = ''
+            require('${luaName}')${setup}(${
+              lib.optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)
+            })
+          '';
+
+          setLuaConfig = lib.setAttrByPath (lib.toList configLocation);
         in
         {
           meta = {
@@ -100,27 +112,40 @@
                 example = settingsExample;
               };
             }
+            // lib.optionalAttrs hasConfigAttrs {
+              luaConfig = lib.mkOption {
+                type = lib.nixvim.nixvimTypes.pluginLuaConfig;
+                default = { };
+                description = "The plugin's lua configuration";
+              };
+            }
             // extraOptions;
 
-          config = lib.mkIf cfg.enable (
-            lib.mkMerge [
-              {
-                extraPlugins = (lib.optional installPackage cfg.package) ++ extraPlugins;
-                inherit extraPackages;
-              }
-              (lib.optionalAttrs callSetup {
-                ${extraConfigNamespace} = ''
-                  require('${luaName}')${setup}(${
-                    lib.optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)
+          config =
+            assert lib.assertMsg (
+              callSetup -> configLocation != null
+            ) "When a plugin has no config attrs and has a setup function it must have a config location";
+            lib.mkIf cfg.enable (
+              lib.mkMerge (
+                [
+                  {
+                    extraPlugins = (lib.optional installPackage cfg.package) ++ extraPlugins;
+                    inherit extraPackages;
+                  }
+                  (lib.optionalAttrs (isColorscheme && (colorscheme != null)) {
+                    colorscheme = lib.mkDefault colorscheme;
                   })
-                '';
-              })
-              (lib.optionalAttrs (isColorscheme && (colorscheme != null)) {
-                colorscheme = lib.mkDefault colorscheme;
-              })
-              (extraConfig cfg)
-            ]
-          );
+                  (extraConfig cfg)
+                ]
+                ++ (lib.optionals (!hasConfigAttrs) [
+                  (lib.optionalAttrs callSetup (setLuaConfig setupCode))
+                ])
+                ++ (lib.optionals hasConfigAttrs [
+                  (lib.optionalAttrs callSetup { ${namespace}.${name}.luaConfig.content = setupCode; })
+                  (lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content))
+                ])
+              )
+            );
         };
     in
     {
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -113,4 +113,38 @@ rec {
         types.optionDescriptionPhrase (class: class == "noun" || class == "composite") elemType
       }";
     };
+
+  pluginLuaConfig = types.submodule (
+    { config, ... }:
+    {
+      options = {
+        pre = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = ''
+            Lua code inserted at the start of the plugin's configuration.
+            This is the same as using `lib.nixvim.utils.mkBeforeSection` when defining `content`.
+          '';
+        };
+        post = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = ''
+            Lua code inserted at the end of the plugin's configuration.
+            This is the same as using `lib.nixvim.utils.mkAfterSection` when defining `content`.
+          '';
+        };
+        content = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Configuration of the plugin";
+        };
+      };
+
+      config.content = lib.mkMerge [
+        (lib.nixvim.utils.mkBeforeSection config.pre)
+        (lib.nixvim.utils.mkAfterSection config.post)
+      ];
+    }
+  );
 }
--- a/plugins/colorschemes/ayu.nix
+++ b/plugins/colorschemes/ayu.nix
@@ -39,7 +39,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   };
 
   extraConfig = cfg: {
-    extraConfigLuaPre = ''
+    colorschemes.ayu.luaConfig.content = ''
       local ayu = require("ayu")
       ayu.setup(${toLuaObject cfg.settings})
       ayu.colorscheme()
--- a/plugins/utils/rest.nix
+++ b/plugins/utils/rest.nix
@@ -434,7 +434,8 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
       }
     ];
 
-    extraConfigLua = lib.mkIf cfg.enableTelescope ''require("telescope").load_extension("rest")'';
+    # TODO: There may be some interactions between this & telescope, maybe requiring #2292
+    plugins.rest.luaConfig.post = lib.mkIf cfg.enableTelescope ''require("telescope").load_extension("rest")'';
 
     extraPackages = [ cfg.curlPackage ];
 
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -78,6 +78,12 @@ rec {
     lib.concatStrings (map processWord words);
 
   /**
+    Those helpers control the lua sections split in `pre, content, post`
+  */
+  mkBeforeSection = lib.mkOrder 300;
+  mkAfterSection = lib.mkOrder 2000;
+
+  /**
     Capitalize a string by making the first character uppercase.
 
     # Example
--- a/plugins/by-name/ccc/default.nix
+++ b/plugins/by-name/ccc/default.nix
@@ -246,7 +246,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
     # ccc requires `termguicolors` to be enabled.
     opts.termguicolors = lib.mkDefault true;
 
-    extraConfigLua = ''
+    plugins.ccc.luaConfig.content = ''
       ccc = require('ccc')
       ccc.setup(${helpers.toLuaObject cfg.settings})
     '';
--- a/plugins/by-name/coq-nvim/default.nix
+++ b/plugins/by-name/coq-nvim/default.nix
@@ -82,7 +82,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
       coq_settings = cfg.settings;
     };
 
-    extraConfigLua = "require('coq')";
+    plugins.coq-nvim.luaConfig.content = "require('coq')";
 
     plugins.lsp = {
       preConfig = ''
--- a/plugins/by-name/hydra/default.nix
+++ b/plugins/by-name/hydra/default.nix
@@ -36,7 +36,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   callSetup = false;
   extraConfig = cfg: {
-    extraConfigLua = ''
+    plugins.hydra.luaConfig.content = ''
       hydra = require('hydra')
 
       hydra.setup(${helpers.toLuaObject cfg.settings})
--- a/plugins/by-name/mini/default.nix
+++ b/plugins/by-name/mini/default.nix
@@ -99,7 +99,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
         '';
       }
     ];
-    extraConfigLua =
+    plugins.mini.luaConfig.content =
       lib.foldlAttrs (lines: name: config: ''
         ${lines}
         require(${lib.nixvim.toLuaObject "mini.${name}"}).setup(${lib.nixvim.toLuaObject config})
--- a/plugins/by-name/rustaceanvim/default.nix
+++ b/plugins/by-name/rustaceanvim/default.nix
@@ -48,6 +48,8 @@ helpers.neovim-plugin.mkNeovimPlugin {
   };
 
   callSetup = false;
+  hasConfigAttrs = false;
+  configLocation = null;
   extraConfig =
     cfg:
     mkMerge [
--- a/plugins/by-name/telescope/default.nix
+++ b/plugins/by-name/telescope/default.nix
@@ -122,7 +122,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
       }
     ) cfg.keymaps;
 
-    extraConfigLua = ''
+    plugins.telescope.luaConfig.content = ''
       require('telescope').setup(${toLuaObject cfg.settings})
 
       local __telescopeExtensions = ${toLuaObject cfg.enabledExtensions}
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -431,7 +431,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
   installPackage = false;
 
   extraConfig = cfg: {
-    extraConfigLua =
+    plugins.treesitter.luaConfig.content =
       # NOTE: Upstream state that the parser MUST be at the beginning of runtimepath.
       # Otherwise the parsers from Neovim takes precedent, which may be incompatible with some queries.
       (optionalString (cfg.settings.parser_install_dir != null) ''
--- a/plugins/by-name/which-key/default.nix
+++ b/plugins/by-name/which-key/default.nix
@@ -596,7 +596,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
         ''
       ];
 
-      extraConfigLua = lib.optionalString opt.registrations.isDefined ''
+      plugins.which-key.luaConfig.content = lib.optionalString opt.registrations.isDefined ''
         require("which-key").register(${toLuaObject cfg.registrations})
       '';
     };
--- a/plugins/by-name/yanky/default.nix
+++ b/plugins/by-name/yanky/default.nix
@@ -373,7 +373,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
       }
     ];
 
-    extraConfigLua = ''
+    plugins.yanky.luaConfig.content = ''
       do
         local utils = require('yanky.utils')
         ${optionalString config.plugins.telescope.enable "local mapping = require('yanky.telescope.mapping')"}
--- a/plugins/cmp/default.nix
+++ b/plugins/cmp/default.nix
@@ -70,7 +70,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   callSetup = false;
   extraConfig = cfg: {
-    extraConfigLua =
+    plugins.cmp.luaConfig.content =
       ''
         local cmp = require('cmp')
         cmp.setup(${helpers.toLuaObject cfg.settings})
--- a/plugins/colorschemes/base16/default.nix
+++ b/plugins/colorschemes/base16/default.nix
@@ -179,7 +179,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
     # `settings` can either be passed to `with_config` before calling `setup`,
     # or it can be passed as `setup`'s 2nd argument.
     # See https://github.com/RRethy/base16-nvim/blob/6ac181b5733518040a33017dde654059cd771b7c/lua/base16-colorscheme.lua#L107-L125
-    extraConfigLuaPre = ''
+    colorschemes.base16.luaConfig.content = ''
       do
         local base16 = require('${luaName}')
         base16.with_config(${toLuaObject cfg.settings})
--- a/plugins/colorschemes/onedark.nix
+++ b/plugins/colorschemes/onedark.nix
@@ -34,7 +34,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   callSetup = false;
   colorscheme = null;
   extraConfig = cfg: {
-    extraConfigLuaPre = ''
+    colorschemes.onedark.luaConfig.content = ''
       _onedark = require('onedark')
       _onedark.setup(${lib.nixvim.toLuaObject cfg.settings})
       _onedark.load()
--- a/plugins/colorschemes/vscode.nix
+++ b/plugins/colorschemes/vscode.nix
@@ -31,7 +31,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   };
 
   extraConfig = cfg: {
-    extraConfigLuaPre = ''
+    colorschemes.vscode.luaConfig.content = ''
       local _vscode = require("vscode")
       _vscode.setup(${toLuaObject cfg.settings})
       _vscode.load()
--- a/plugins/none-ls/default.nix
+++ b/plugins/none-ls/default.nix
@@ -147,7 +147,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
       ];
 
       # We only do this here because of enableLspFormat
-      extraConfigLua = ''
+      plugins.none-ls.luaConfig.content = ''
         require("null-ls").setup(${helpers.toLuaObject setupOptions})
       '';
     };
--- a/plugins/pluginmanagers/lz-n.nix
+++ b/plugins/pluginmanagers/lz-n.nix
@@ -190,7 +190,7 @@ nixvim.neovim-plugin.mkNeovimPlugin {
 
   extraConfig = cfg: {
     globals.lz_n = modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
-    extraConfigLua = mkIf (cfg.plugins != [ ]) ''
+    plugins.lz-n.luaConfig.content = mkIf (cfg.plugins != [ ]) ''
       require('lz.n').load( ${nixvim.toLuaObject cfg.plugins})
     '';
   };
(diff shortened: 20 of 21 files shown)


GaetanLepage (approved):
LGTM !

lib/neovim-plugin.nix
```diff
@@ -90,6 +90,21 @@ with lib;
...
+
+          configLuaPre = mkOption {
+            type = types.lines;
+            default = "";
+          };
+
+          configLuaPost = mkOption {
+            type = types.lines;
+            default = "";
+          };
```
MattSturgeon, line 106: Do you think these should be hidden, until they will work as expected for _all_ plugins that have these options added?

Currently for many plugins these options will not actually do what they claim (add config beside the plugin's config) since the plugin implementation isn't updated to use the new options internally.

lib/neovim-plugin.nix
```diff
@@ -104,16 +119,20 @@ with lib;
...
             inherit extraPackages;
 
-            ${extraConfigNamespace} = optionalString callSetup ''
+            ${namespace}.${name}.${extraConfigNamespace} = optionalString callSetup ''
               require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
             '';
+
+            extraConfigLua = cfg.configLua;
+            extraConfigLuaPre = cfg.configLuaPre;
+            extraConfigLuaPost = cfg.configLuaPost;
```
MattSturgeon, line 135: From reading the option names, I was actually expecting to see the config grouped together in `extraConfigLua`:

```nix
extraConfigLua = concatStringsSep "\n" [
  cfg.configLuaPre
  cfg.configLua
  cfg.configLuaPost
];
```

Allowing users to add config before/with/after the plugin's internal config.

I think this is more useful for end-users, but I can see use cases for both approaches.

Maybe we should have `preConfigLua{Pre,Post,}`, `configLua{Pre,Post,}`, and `postConfigLua{Pre,Post,}`?

Though this feels a bit cumbersome...

lib/neovim-plugin.nix
traxys: Well I should have updated all plugins declared with mkNeovimPlugin! So it _should not_ be a lie

lib/neovim-plugin.nix
traxys: Yeah this feels a bit complicated. I hesitated on the two approaches, but I prefered going globaly, as it implies the local approach. We can always revisit this behaviour later, for example by adding a global option `localPluginConfig` that choses what to do.
This a bit in the spirit of what is made availaible by this PR!

lib/neovim-plugin.nix
MattSturgeon: Ok, if so then I'm fine with them being visible. The touched file count is suspiciously low, but I'll trust you :grin:

lib/neovim-plugin.nix
MattSturgeon: Ignoring any internal usage we have for a second, I think from a user perspective the fact the options are nested under `plugins.<name>` implies they are related to that plugin.\
Therefore I think the "local" pre/post makes most sense to users.

I'm hesitant to merge as-is if we're already considering how it may change going forward, given we may end up running into rename conflicts (etc).

---

I wonder if we can mitigate the confusion (and the clunky-ness of having 9 options) by renaming the "global" options.

Maybe the top-level options could be something like `extraConfig{Early,Late,}` (dropping the implied `Lua` and changing from pre/post to early/late)?

More local options would then be free to use "pre" and "post" without confusion, while falling back to "early" and "late" when they need to tie in with the more "global" config locations?

MattSturgeon:
Can we revisit this now the CI is working again?

traxys:
Yup, I thought about it a bit and I think I'll go with your suggestion, of creating a  single lua block with `pre|post`

traxys:
@MattSturgeon I re-wrote the PR and amended the description with the new implementation details!

lib/neovim-plugin.nix
```diff
@@ -79,6 +82,31 @@ with lib;
...
+              internal = true;
+              default = "";
+              description = "Initialization code (between pre & post)";
+            };
+            final = mkOption {
+              type = types.str;
+              description = "Generated lua configuration for ${name}";
+              readOnly = true;
+            };
+          };
```
MattSturgeon, line 108: I think this would be better abstracted as a submodule type, that is a little more self-contained and re-usable:

```nix
pluginConfigText = types.submodule (
  { config, ... }:
  {
    options = {
      pre = mkOption {
        type = types.lines;
        default = "";
        description = "Configuration before the initialization of ${name}";
      };
      post = mkOption {
        type = types.lines;
        default = "";
        description = "Configuration before the initialization of ${name}";
      };
      init = mkOption {
        type = types.lines;
        internal = true;
        default = "";
        description = "Initialization code (between pre & post)";
      };
      final = mkOption {
        type = types.str;
        description = "Generated lua configuration for ${name}";
        readOnly = true;
      };
    };

    config.final = lib.nixvim.concatNonEmptyLines [
      config.pre
      config.init
      config.post
    ];
  });
```

Note the submodule can define `final` internally.

This could probably go in `nixvimTypes`?

lib/neovim-plugin.nix
```diff
@@ -91,15 +119,27 @@ with lib;
...
+            '';
+          })
+          (optionalAttrs (callSetup && !skipConfigAttrs) {
+            ${namespace}.${name}.config.init = ''
+              require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
+            '';
+            ${configLocation} = cfg.config.final;
+          })
+          (optionalAttrs (callSetup && skipConfigAttrs) {
+            ${configLocation} = ''
```
MattSturgeon, line 142: This could still use `setAttrByPath` if you wrap `configLocation` in `toList`. This allows `configLocation` to either be a string or a list.

```nix
(optionalAttrs (callSetup && skipConfigAttrs) (setAttrByPath (toList configLocation) ''
  require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
''))
```

lib/neovim-plugin.nix
```diff
@@ -91,15 +119,27 @@ with lib;
...
           }
-          (optionalAttrs callSetup {
-            ${extraConfigNamespace} = ''
+          (optionalAttrs (!skipConfigAttrs) {
+            ${namespace}.${name}.config.final = ''
+              ${cfg.config.pre}
+              ${cfg.config.init}
+              ${cfg.config.post}
+            '';
+          })
```
MattSturgeon, line 134: This can be implemented internally in the `config.*` submodule (see above)

lib/neovim-plugin.nix
```diff
@@ -91,15 +119,27 @@ with lib;
...
+          (optionalAttrs (callSetup && !skipConfigAttrs) {
+            ${namespace}.${name}.config.init = ''
+              require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
+            '';
+            ${configLocation} = cfg.config.final;
+          })
```
MattSturgeon, line 140: This should be split up, assigning `config.final` to `configLocation` shouldn't be gated behind `callSetup`:

```nix
(optionalAttrs (callSetup && !skipConfigAttrs) {
  ${namespace}.${name}.config.init = ''
    require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
  '';
})
(optionalAttrs (!skipConfigAttrs) {
  ${configLocation} = cfg.config.final;
})
```

Also, we can use `setAttrByPath` if preferred, see above.

plugins/colorschemes/ayu.nix
```diff
@@ -40,10 +40,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   };
 
   extraConfig = cfg: {
-    extraConfigLuaPre = ''
+    colorschemes.ayu.config.init = ''
       local ayu = require("ayu")
       ayu.setup(${helpers.toLuaObject cfg.settings})
       ayu.colorscheme()
     '';
+    extraConfigLuaPre = cfg.config.final;
```
MattSturgeon, line 48: I'm confused, why is this needed here? Is this because assigining `final` to `configLocation` is currently gated behind `callSetup`? (I don't think it needs to be, though)

plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -82,9 +80,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
         # Ensure the plugin config is placed **after** the rest of the LSP configuration
         # (and thus after the declaration of `__lspOnAttach`)
-        plugins.lsp.postConfig = configStr;
+        plugins.lsp.postConfig = cfg.config.final;
       })
       # Else, just put the plugin config anywhere
-      (mkIf (!config.plugins.lsp.enable) { extraConfigLua = configStr; })
+      (mkIf (!config.plugins.lsp.enable) { extraConfigLua = cfg.config.final; })
```
MattSturgeon, line 86: Is this why you gated the `final` -> `configLocation` assignment behind `callSetup`? If so I think we should introduce a different way to do that.

Perhaps `configLocation` could be nullable and setting it to null prevents assigning anything to it?

MattSturgeon (commented):
Initial thoughts and questions, thanks for continuing to work on this!

lib/neovim-plugin.nix
```diff
@@ -36,6 +36,9 @@ with lib;
       hasSettings ? true,
       extraOptions ? { },
       # config
+      configLocation ? [ (if isColorscheme then "extraConfigLuaPre" else "extraConfigLua") ],
```
MattSturgeon, line 39:
```suggestion
      configLocation ? if isColorscheme then "extraConfigLuaPre" else "extraConfigLua",
```

Because we use `toList` in the implementation, this is fine as a string. Maybe just add a comment saying it can either be a string or a list?

plugins/utils/flash.nix
```diff
@@ -11,6 +11,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   originalName = "flash.nvim";
   defaultPackage = pkgs.vimPlugins.flash-nvim;
 
+  skipConfigAttrs = true;
```
MattSturgeon, line 14: I don't understand why this is needed? I can't see anything config-related in the flash module. Do you mind clarifying?

plugins/utils/hydra/default.nix
```diff
@@ -38,7 +38,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   callSetup = false;
   extraConfig = cfg: {
-    extraConfigLua = ''
+    plugins.hydra.config.init = ''
```
MattSturgeon, line 41: Have you replaced `extraConfigLua` assignments in _all_ mkNeovimPlugin call sites treewide? :eyes:

lib/neovim-plugin.nix
```diff
@@ -36,6 +36,9 @@ with lib;
...
+      # Required for plugins.flash, as it used to define config
+      skipConfigAttrs ? false,
```
MattSturgeon, line 41: As per my other comment, I don't really understand why this is needed.

plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -50,13 +50,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   };
 
   callSetup = false;
+  configLocation = null;
```
MattSturgeon, line 53: Perhaps add a note explaining why we can't just do `configLocation = if config.plugins.lsp.enable then [ "plugins" "lsp" "postConfig" ] else "extraConfigLua"` and that we'll implement it below using `mkIf`?

lib/neovim-plugin.nix
```diff
@@ -91,18 +101,27 @@ with lib;
...
+            in
+            if !skipConfigAttrs then
+              (optionalAttrs callSetup { ${namespace}.${name}.config.init = setupCode; })
+              // (optionalAttrs (configLocation != null) (setAttrByPath (toList configLocation) cfg.config.final))
+            else
+              assert assertMsg (
+                callSetup -> configLocation != null
+              ) "When a plugin has no config attrs and has a setup function it must have a config location";
+              optionalAttrs callSetup (setAttrByPath (toList configLocation) setupCode)
+          )
```
MattSturgeon, line 124: Something about this feels a little messy. Can't quite put my finger on it though... Maybe a combination of minor things?

- why use `!` in an if-then-else when we can just flip it?
- The `optionalAttrs // optionalAttrs` could fail if `configLocation` ever starts with `namespace`
  - We could use a recursive update, but probably better to just use `mkMerge`
- Having nested conditional branches _within_ a `mkMerge` feels messy, probably better to just have all variations listed directly in the `mkMerge`?

Perhaps something like:
```nix
config =
  let
    setupLine = ''
      require('${luaName}')${setup}(${optionalString (cfg ? settings) (helpers.toLuaObject cfg.settings)})
    '';
  in
  assert assertMsg (
    # FIXME: check this logic
    (skipConfigAttrs && callSetup) -> configLocation != null
  ) "When a plugin has no config attrs and has a setup function it must have a config location";
  mkMerge [
    (optionalAttrs (callSetup && !skipConfigAttrs) { ${namespace}.${name}.config.init = setupCode; })
    (optionalAttrs (configLocation != null && !skipConfigAttrs) (setAttrByPath (toList configLocation) cfg.config.final))
    # etc
  ];
```
?

plugins/utils/flash.nix
traxys: The flash plugin has a renamed option from `config` to `settings.config`

plugins/utils/hydra/default.nix
traxys: To my knowledge yes!

lib/neovim-plugin.nix
traxys: I wanted to factor the skipConfigAttrs calls, because it became quite complicated to follow the flow with everthing in the same condition, as you have a `callSetup && skipConfigAttrs`, a `callSetup && !skipConfigAttrs` and a `configLocation != null && !skipConfigAttrs`

After 24.11 we should be able to remove the skipConfigAttrs, and it would simplify a lot the module

MattSturgeon: Hm, I wonder if another way to approach this would be to concat `optionals` lists together, potentially containing optionalAttrs?

```nix
mkMerge (
  optionals skipConfigAttrs [
    (optionalAttrs callSetup (setAttrByPath (toList configLocation) setupCode)
  ]
  ++ optionals (!skipConfigAttrs) [
    (optionalAttrs callSetup { ${namespace}.${name}.config.init = setupCode; })
    (optionalAttrs (configLocation != null) {})
  ]
);
```

But I think you're right, half of the issue is just that it is a messy problem to solve, so I shouldn't be surprised that it ends up looking messy :joy:

plugins/utils/flash.nix
MattSturgeon: Ah, I see. That's unfortunate!

I wonder if `config` is the best name for the new option anyway... It's already an overloaded word in nixos modules, since we have a config attr in the module output, plus a `config` module arg, plus modules in general usually relate to _configuration_...

Not saying we should choose a different name _just because_ of this conflict though. Only if we actually think of a better name.

plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -74,6 +70,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
               Note that if you supplied an attrset and not a function you need to set this attr set in:
                 `plugins.rustaceanvim.settings.server.default_settings.rust-analyzer'.
             '';
+        plugins.rustaceanvim.config.init = ''
```
khaneliman, line 73: Will this help resolve the checkhealth error we've had ? 
```
- ERROR vim.g.rustaceanvim is set, but it was sourced after rustaceanvim was initialized.
```

plugins/utils/flash.nix
khaneliman: I was wondering this, as well... but I don't know what else you'd choose.

plugins/languages/rust/rustaceanvim/default.nix
MattSturgeon: Would need to look into what that error really means, but it sounds like rustaceanvim is unhappy that we're setting `globals.rustaceanvim` before calling their `setup` function?

khaneliman: Yeah, didn't know if adding this was setting up the global before setup, now. Or if we needed to move it to `pre` instead of init.

khaneliman: Sorry, looking at their [readme](https://github.com/mrcjkb/rustaceanvim?tab=readme-ov-file#zap-quick-setup) they don't use the setup call and just use a filetype detection to initialize. Not sure why we'd be getting that health error, can just look at it separately.

plugins/utils/flash.nix
traxys: Naming things isn't my strong suit, if you have better ideas don't hesitate!

MattSturgeon: Perhaps `luaConfig`,  `luaText`, or even just `lua`?

lib/types.nix
```diff
@@ -93,4 +93,39 @@ rec {
...
+          type = types.lines;
+          internal = true;
+          default = "";
+          description = "Initialization code (between pre & post)";
+        };
+        final = mkOption {
+          type = types.str;
+          description = "Complete plugin configuration";
+          readOnly = true;
+        };
```
MattSturgeon, line 121: Should we consider marking this as `internal`?

I'm also wondering if it makes sense to make it a `types.lines` and not `readOnly`?

We could then define the value as:
```nix
config.final = mkMerge [
  (mkIf (config.pre != "") (mkBefore config.pre))
  (mkIf (config.init != "") config.init)
  (mkIf (config.post != "") ((mkAfter config.post))
]
```

MattSturgeon (dismissed):
I don't see any issues, but it might be nice to mark `luaConfig.final` as internal?

I also think it may be more powerful/future-proof to make `final` a lines type and make use of ordered definitions, see https://github.com/nix-community/nixvim/pull/1876#discussion_r1735420626.

lib/types.nix
khaneliman: I think it makes sense to mark it internal, since it's going to be a derived value from other options. I'm not sure on benefit of readOnly vs not.

MattSturgeon: Personally I see `readOnly` as a "last resort" for when it absolutely never makes sense for users to modify the option's value.

I think `internal` makes more sense here, and if the option is mutable I think `types.lines` makes more sense than `types.str`.

traxys: The idea of it being internal is that you would be able to benefit from nixvim genrrating the setup code, but you could put it anywhere, in an autocmd for example.

In that case I would strongly prefer this to be readOnly

MattSturgeon: > The idea of it being internal 

Currently it isn't `internal`. I'm suggesting maybe it should be?

> you would be able to benefit from nixvim genrrating the setup code, but you could put it anywhere, in an autocmd for example

I don't see how it being mutable is an obstacle to any of that?

I guess we could advise users that they should normally prefer the `pre` and `post` options, but aren't we just imposing needless restrictions by making `final` read-only?

traxys: Yeah I was detailing my reasons for it _not_ to be internal.

It is indeed a bit of a "useless" restriction, but I don't like very much to have two ways to do something (add to pre/post or add to final).
I think pre/post is better for newcommers, so that's why I went with this approach

khaneliman (dismissed):
Really like where this PR has gone in giving more control over lua relevant to a plugin

lib/types.nix
khaneliman: I think internal would make it hidden from a newcomer, though. It just gives power to advanced users who are aware they might shoot themselves in the foot by using an option. But, I'm not opposed to just revisiting later if we find a desire to override the final option, at some point.

traxys: Yeah I don't want it to be internal, though my first reaction was to mark it as readOnly to be conservative in what we expose. If we judge there is a sufficient value in marking it not readOnly I could do it.

I'm not a 100% convinced it's needed with the pre/post options though

stasjok:
Maybe wrap the config in `do ... end` block? Or even add an additional option like `wrapInDo` default true. This PR is very useful, but usually I try to write self-contained blocks of code and do block ensures that I don't pollute locals or don't override something unintentionally.

plugins/utils/ccc.nix
```diff
@@ -248,7 +248,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     # ccc requires `termguicolors` to be enabled.
     opts.termguicolors = lib.mkDefault true;
 
-    extraConfigLua = ''
+    plugins.ccc.luaConfig.init = ''
       ccc = require('ccc')
```
stasjok, line 252: It's not introduced by this PR but why is it not local? Hydra later too.

plugins/utils/rest.nix
```diff
@@ -434,6 +434,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
       }
     ];
 
+    # TODO: Clarify how to handle this, add to the plugin's luaConfig or to telescope's
     extraConfigLua = lib.mkIf cfg.enableTelescope ''require("telescope").load_extension("rest")'';
```
khaneliman, line 438: This is something that recently I've been wondering how we can start ordering the generated entries with dag or something. I'd like to be able to make sure certain plugins are setup after others. I don't know how we can guarantee that telescope is available when it's added to rests config or that rest is available if we add it to telescopes config.

khaneliman: Is it enough to just assign a mkOverride priority ordering system for the luaConfigs?

traxys: The best solution in a vacuum would be to use some kind of DAG to order the plugins, but implementing it could be a bit complicated (TBH I would not know where to start)

traxys:
> Maybe wrap the config in `do ... end` block? Or even add an additional option like `wrapInDo` default true. This PR is very useful, but usually I try to write self-contained blocks of code and do block ensures that I don't pollute locals or don't override something unintentionally.

Very good idea, though I think I'll implement that in a follow-up PR

plugins/utils/rest.nix
khaneliman: Yeah, definitely warrants its own Issue tracking and is out of scope on this PR... just made me think about it with this particular note. I think it's fine to keep it in rest, for now.

lib/types.nix
```diff
@@ -113,4 +113,32 @@ rec {
...
+          description = "Configuration after the initialization of the plugin";
+        };
+        init = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Initialization code (between pre & post)";
+        };
+      };
+
+      config.init = lib.mkMerge [
```
khaneliman, line 138: Interesting, so the `mkMerge` combines `pre` and `post` with `init` instead of replacing its contents in this declaration?

plugins/utils/rest.nix
khaneliman, line 438:
```suggestion
    plugins.rest.luaConfig.init = lib.mkIf cfg.enableTelescope ''require("telescope").load_extension("rest")'';
```

lib/types.nix
traxys: Yeah, I added a test to check that it indeed works. I was inspired with what @MattSturgeon did

khaneliman (approved):
LGTM I think this will be a very useful change

lib/types.nix
```diff
@@ -113,4 +113,32 @@ rec {
...
     };
+
+  pluginLuaConfig = types.submodule (
+    { config, ... }:
+    {
+      options = {
+        pre = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Configuration before the initialization of the plugin";
```
MattSturgeon, line 124: I think this description could clarify that `pre` is actually merged into `init` with a "before" merge-order priority

lib/types.nix
```diff
@@ -113,4 +113,32 @@ rec {
...
+        };
+        post = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Configuration after the initialization of the plugin";
```
MattSturgeon, line 129: Same

lib/types.nix
```diff
@@ -113,4 +113,32 @@ rec {
...
+        };
+        init = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Initialization code (between pre & post)";
```
MattSturgeon, line 134: Same;

"between" pre&post is actually a little misleading. Anyone that _reads_ `init` will see it as _including` pre&post.

Anyone who uses `mkOrder` on init may also be confused when pre&post get merged in-between their ordered definitions.

lib/types.nix
```diff
@@ -113,4 +113,32 @@ rec {
...
+        };
+      };
+
+      config.init = lib.mkMerge [
+        (lib.mkBefore config.pre)
+        (lib.mkAfter config.post)
```
MattSturgeon, line 140: I wonder if it better to use non-standard merge priority numbers?

E.g. what would we expect to happen when someone uses:

```
pre = "foo";
init = mkBefore "bar";
```

Currently they would both be merged at priority 500, however we can't know just from reading the definition of we'll get `"foo\nbar"` or `"bar\nfoo"`.

I think users would expect `pre` to be merged earlier than mkBefore and `post` to be merged after mkAfter?

MattSturgeon (commented):
Looks good overall! I just wonder if the descriptions could be clarified?

lib/types.nix
khaneliman: Seems like an implementation level detail that would cloud the purpose of the option.

lib/types.nix
khaneliman: Hmm... this makes me not like tying them to the same mkOrder merge if we can have this interaction. It should be set that pre comes before init and post comes after.

lib/types.nix
khaneliman: Definitely a niche case, but makes sense to at least prevent default priority overrides from messing with ordering.

lib/types.nix
MattSturgeon: I don't think the fact that `pre` and `post` are merged into `init` is purely an implementation detail:

- End users will be exposed to this detail when _reading_ the final `init` value
- `mkOrder` is a user-facing utility
  - how mkOrder interacts with pre & post could be important to some users

I think the implementation is fine, so long as users aren't surprised by it.

khaneliman: >     * End users will be exposed to this detail when _reading_ the final `init` value

True, this makes sense to me as a reason to include some mention, then.

lib/types.nix
traxys: Yeah I think I should change the name, to something like `config`  or `content`

lib/types.nix
```diff
@@ -113,4 +113,38 @@ rec {
...
+  pluginLuaConfig = types.submodule (
+    { config, ... }:
+    {
+      options = {
+        pre = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = ''
+            Lua code inserted at the start of the plugin's configuration.
+            This is simply sugar around `lib.nixvim.utils.mkBeforeSection content`
```
MattSturgeon, line 126:
```suggestion
            This is the same as using `lib.nixvim.utils.mkBeforeSection` when defining `content`.
```

Or maybe something like 

```suggestion
            This is the same as `content = lib.mkOrder ${(lib.nixvim.utils.mkBeforeSection null).priority} ""`
```

?

traxys: I'd prefer the first one

lib/types.nix
```diff
@@ -113,4 +113,38 @@ rec {
...
+        content = lib.mkOption {
+          type = types.lines;
+          default = "";
+          description = "Configuration of the plugin";
+        };
+      };
+
+      config.content = lib.mkMerge [
+        (lib.nixvim.utils.mkBeforeSection config.pre)
+        (lib.nixvim.utils.mkAfterSection config.post)
```
MattSturgeon, line 146: Strictly speaking, these should each be wrapped in `mkIf (config.pre != "")`, to avoid unnecessary empty lines when no pre/post is defined.

Or we could drop their default and use `isDefined`. Or we could keep their default and use `highestPrio < 1500`.

Regardless of the approach used, it will likely end up kinda messy. Therefore I don't think we need to necessarily tackle it in this PR.

MattSturgeon (approved):
Awesome, LGTM!

traxys:
@Mergifyio queue

PR #1877, opened 2024-07-16
plugins/none-ls/sources: structured "with" settings

MattSturgeon:
Either strLua, or a structured `attrsOf anything`.

I resisted the urge to make this a submodule with options, but maybe it'd make sense to link the upstream documentation.


GaetanLepage (approved):
LGTM !

Issue #1878, opened 2024-07-17
buildbot errors


zowoq:
https://github.com/nix-community/nixvim/blob/b9ed90003273f0a75151b32948e16b44891f403c/tests/default.nix#L46

The problem seems to be the linkfarm test that was added in https://github.com/nix-community/nixvim/pull/892, it is using a lot of memory.

> This should dramatically reduce the build time on buildbot.

@GaetanLepage Do you recall the details of this problem, we may be able to find another way of addressing it?

cc @Mic92

Mic92:
The issue here is that one derivation depends on all derivations. So effectively everything gets evaluated twice

GaetanLepage:
> @GaetanLepage Do you recall the details of this problem, we may be able to find another way of addressing it?

Yes, so I made that change for the following reason. Our test suite has one derivation per configuration we want to test.
Thus, as we have 2-3 test cases per plugin, and more than 100 plugins, this corresponded to several hundreds of test derivations per architecture.
So basically,  build-bot was getting flood and couldn't keep up with that many derivations (+ it was making the UI unusable as well).

traxys:
The issue is that in the end our tests require a derivation each (as we launch neovim with the supplied configuration). Would launching each plugin (so ~100 derivations) independently be less though on buildbot?

MattSturgeon:
> (Each of the derivations would be dependent on 2 to 4 other derivations on average)

So effectively a tree of link farm derivations?

In total we'd have more derivations, but maybe having a tree structure would be easier on buildbot?

As another approach, would it be easier on buildbot if we split up our `checks.*.tests` output into several `checks.*.test-*` outputs?

Currently `checks.*.tests` is the link farm derivation that has hundreds of dependencies.

Worth noting: while we have many wrapped nvim test derivations, they should all be wrapping _the same_ nvim package derivation, so that should only need to be built/downloaded once.

GaetanLepage:
> So effectively a tree of link farm derivations?

Yes, what matters for build-bot is the number of top-level derivations. In this sense, it is tractable to have ~60 total top-level derivations (what we have now) rather than ~600+ that we would have without the linkFarm.

zowoq:
I've disabled buildbot on this repo for now.

Even after the recent changes it's still using a lot of memory, can reproduce it by running the eval manually:

```
nix run nixpkgs#nix-eval-jobs -- --workers 1 --check-cache-status --force-recurse --flake 'github:nix-community/nixvim#checks'
```

PR #1882, opened 2024-07-17
tests: Reduce the size of the test derivation link farm

traxys:
This PR reduces the number of test derivations, only creating one derivation per plugin instead of one test derivation per test configuration

#1878

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -17,30 +17,51 @@ let
   };
 
   exampleFiles = {
-    example =
+    name = "examples";
+    cases =
       let
         config = import ../example.nix { inherit pkgs; };
       in
-      builtins.removeAttrs config.programs.nixvim [
-        # This is not available to standalone modules, only HM & NixOS Modules
-        "enable"
-        # This is purely an example, it does not reflect a real usage
-        "extraConfigLua"
-        "extraConfigVim"
+      [
+        {
+          name = "main";
+          case = builtins.removeAttrs config.programs.nixvim [
+            # This is not available to standalone modules, only HM & NixOS Modules
+            "enable"
+            # This is purely an example, it does not reflect a real usage
+            "extraConfigLua"
+            "extraConfigVim"
+          ];
+        }
       ];
   };
 
   # We attempt to build & execute all configurations
-  derivationList = pkgs.lib.mapAttrsToList (name: def: {
-    inherit name;
-    path = mkTestDerivationFromNixvimModule {
-      inherit name;
-      # The module can either be the actual definition,
+  derivationList = builtins.map (
+    { name, cases }:
+    let
+      # The test case can either be the actual definition,
       # or a child attr named `module`.
-      module = def.module or (lib.removeAttrs def [ "tests" ]);
-      dontRun = def.tests.dontRun or false;
-      pkgs = pkgsUnfree;
-    };
-  }) (testFiles // exampleFiles);
+      prepareModule = case: case.module or (lib.removeAttrs case [ "tests" ]);
+      dontRunModule = case: case.tests.dontRun or false;
+    in
+    {
+      inherit name;
+      path = mkTestDerivationFromNixvimModule {
+        inherit name;
+        tests = builtins.map (
+          { name, case }:
+          {
+            inherit name;
+            module = prepareModule case;
+            dontRun = dontRunModule case;
+          }
+        ) cases;
+        # Use the global dontRun only if we don't have a list of modules
+        dontRun = dontRunModule cases;
+        pkgs = pkgsUnfree;
+      };
+    }
+  ) (testFiles ++ [ exampleFiles ]);
 in
 pkgs.linkFarm "nixvim-tests" derivationList
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -7,7 +7,7 @@
 }:
 let
   fetchTests = import ./fetch-tests.nix;
-  test-derivation = import ./test-derivation.nix { inherit pkgs makeNixvimWithModule; };
+  test-derivation = import ./test-derivation.nix { inherit pkgs makeNixvimWithModule lib; };
   inherit (test-derivation) mkTestDerivationFromNixvimModule;
 
   # List of files containing configurations
--- a/tests/test-derivation.nix
+++ b/tests/test-derivation.nix
@@ -1,40 +1,63 @@
-{ pkgs, makeNixvimWithModule, ... }:
+{
+  pkgs,
+  lib ? pkgs.lib,
+  makeNixvimWithModule,
+  ...
+}:
 let
   # Create a nix derivation from a nixvim executable.
   # The build phase simply consists in running the provided nvim binary.
   mkTestDerivationFromNvim =
     {
       name,
-      nvim,
+      nvim ? null,
+      tests ? [ ],
       dontRun ? false,
       ...
     }:
+    # At least one of these should be undefined
+    assert lib.assertMsg (tests == [ ] || nvim == null) "Both nvim & tests can't be supplied";
+    let
+      nvimAsList = lib.optional (nvim != null) {
+        derivation = nvim;
+        inherit name dontRun;
+      };
+
+      testList = tests ++ nvimAsList;
+    in
     pkgs.stdenv.mkDerivation {
       inherit name;
 
-      nativeBuildInputs = [
-        nvim
-        pkgs.docker-client
-      ];
+      nativeBuildInputs = [ pkgs.docker-client ];
 
       dontUnpack = true;
       # We need to set HOME because neovim will try to create some files
       #
       # Because neovim does not return an exitcode when quitting we need to check if there are
       # errors on stderr
-      buildPhase =
-        if !dontRun then
-          ''
-            mkdir -p .cache/nvim
+      buildPhase = lib.optionalString (!dontRun) (
+        ''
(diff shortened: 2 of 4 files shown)


tests/test-derivation.nix
```diff
@@ -9,32 +14,51 @@ let
...
+      testList =
+        if builtins.isList nvim then
+          nvim
+        else
+          [
+            {
+              derivation = nvim;
+              inherit name dontRun;
+            }
+          ];
```
MattSturgeon, line 27: Passing `nvim` as a list feels like a hack, maybe this should be its own `mkTestDerivationFromNvims` function?

Or at least handle it as a separate argument:

```nix
{
  name,
  nvim ? abort "nvim is required if `tests` is not provided.",
  tests ? [ { derivation = nvim; inherit name dontRun; } ],
  dontRun ? false,

File: tests/test-derivation.nix [hunk 2]
```diff
@@ -9,32 +14,51 @@ let
...
-            	echo "ERROR: $output"
-              exit 1
-            fi
-          ''
-        else
-          '''';
+              output=$(HOME=$(realpath .) ${lib.getExe derivation} -mn --headless "+q" 2>&1 >/dev/null)
+              if [[ -n $output ]]; then
+                echo "ERROR: $output"
+                exit 1
```
MattSturgeon, line 56: Maybe in another PR, could we collect errors and print them all instead of exiting immediately?

tests/test-derivation.nix
```diff
@@ -53,10 +77,21 @@ let
       dontRun ? false,
     }:
     let
-      nvim = makeNixvimWithModule {
-        inherit pkgs module extraSpecialArgs;
-        _nixvimTests = true;
-      };
+      moduleList = if builtins.isList module then module else [ { inherit module name; } ];
```
MattSturgeon, line 80: Same comment as for `mkTestDerivationFromNvim`

tests/test-derivation.nix
```diff
@@ -53,10 +77,21 @@ let
...
+      nvim = builtins.map (
```
MattSturgeon, line 81: If this is a list, it probably shouldn't be named `nvim` anymore.

I guess this is also the same comment as above.

MattSturgeon (commented):
Looking good overall, thanks for working on this, you just about beat me to it!

Nits, questions, ideas, comments, discussion: below.

One other thought: if one derivation per file is still too much for buildbot, we can probably do a derivation per _directory_ (i.e. a deeply nested tree of test derivations, rather than a fairly flat tree).

MattSturgeon:
Tried rebuilding but got the same error, again only on x86_64-darwin:

```
plugins-completion-coq> calling 'envHostTargetHook' function hook 'bintoolsWrapper_addLDVars' /nix/store/d8qw1hx36cb2gxlq2wf1kf7mycsijin2-libcxx-16.0.6-dev
plugins-completion-coq> calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/mn8ink3sn20bhvkkhi3a63jp08ik2pks-libcxx-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'ccWrapper_addCVars' /nix/store/mn8ink3sn20bhvkkhi3a63jp08ik2pks-libcxx-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'bintoolsWrapper_addLDVars' /nix/store/mn8ink3sn20bhvkkhi3a63jp08ik2pks-libcxx-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/wjqfklqkmjqjp19ccc9schq5la1651qf-compiler-rt-libc-16.0.6-dev
plugins-completion-coq> calling 'envHostTargetHook' function hook 'ccWrapper_addCVars' /nix/store/wjqfklqkmjqjp19ccc9schq5la1651qf-compiler-rt-libc-16.0.6-dev
plugins-completion-coq> calling 'envHostTargetHook' function hook 'bintoolsWrapper_addLDVars' /nix/store/wjqfklqkmjqjp19ccc9schq5la1651qf-compiler-rt-libc-16.0.6-dev
plugins-completion-coq> calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/fwximy8vv7h9yrh8bdv97z4aasd6mwby-compiler-rt-libc-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'ccWrapper_addCVars' /nix/store/fwximy8vv7h9yrh8bdv97z4aasd6mwby-compiler-rt-libc-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'bintoolsWrapper_addLDVars' /nix/store/fwximy8vv7h9yrh8bdv97z4aasd6mwby-compiler-rt-libc-16.0.6
plugins-completion-coq> calling 'envHostTargetHook' function hook 'sdkRootHook' /nix/store/gjb3zq5l87pfsmmps9imsz4sykhvsy9x-apple-framework-CoreFoundation
plugins-completion-coq> calling 'envHostTargetHook' function hook 'ccWrapper_addCVars' /nix/store/gjb3zq5l87pfsmmps9imsz4sykhvsy9x-apple-framework-CoreFoundation
plugins-completion-coq> calling 'envHostTargetHook' function hook 'bintoolsWrapper_addLDVars' /nix/store/gjb3zq5l87pfsmmps9imsz4sykhvsy9x-apple-framework-CoreFoundation
plugins-completion-coq> Running phase: patchPhase
plugins-completion-coq> Running phase: configurePhase
plugins-completion-coq> calling 'preConfigure' function hook '_multioutConfig'
plugins-completion-coq> no configure script, doing nothing
plugins-completion-coq> Running phase: buildPhase
plugins-completion-coq> Running test for artifacts
plugins-completion-coq> Running test for empty
plugins-completion-coq> Running test for nixvim-defaults
plugins-completion-coq> ERROR: Error detected while processing command line:
plugins-completion-coq> COQ EXITED - 137
```

Maybe this test is just broken on that system though and this isn't an issue with this PR?

traxys:
Yeah I fail to see why this PR would result in this for the test. Maybe it's flaky, and we never encountered that because tests were really isolated?

MattSturgeon:
> Yeah I fail to see why this PR would result in this for the test. Maybe it's flaky, and we never encountered that because tests were really isolated?

:+1: 

Lets just disable it for now, either on darwin or specifically on x86_64 darwin.

tests/default.nix
```diff
@@ -7,7 +7,7 @@
 }:
 let
   fetchTests = import ./fetch-tests.nix;
-  test-derivation = import ./test-derivation.nix { inherit pkgs makeNixvimWithModule; };
+  test-derivation = import ./test-derivation.nix { inherit pkgs makeNixvimWithModule lib; };
```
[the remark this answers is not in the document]
traxys: Yeah I think so, we import it in two different places, it's not very good I think

tests/test-derivation.nix
traxys: Yeah I was not of the best way to do that. I would have liked to _force_ to have a list, but because we expose this I wouldn't want to break user configurations.

tests/test-derivation.nix
traxys: I'll do that in another PR

tests/test-derivation.nix
MattSturgeon: > because we expose this I wouldn't want to break user configurations

Agreed, we should avoid breaking changes

tests/test-derivation.nix
```diff
@@ -1,40 +1,62 @@
...
   mkTestDerivationFromNvim =
     {
       name,
-      nvim,
+      nvim ? null,
+      tests ? [ ],
       dontRun ? false,
       ...
     }:
+    assert lib.assertMsg (tests == [ ] || nvim == null) "Both nvim & tests can't be supplied";
```
MattSturgeon, line 18: I think this should be `&&`?

Also need to assert that at least one is defined:

```nix
assert lib.assertMsg (tests != [ ] || nvim != null) "Either `nvim` or `tests` must be provided.";
```

---

Would it be easier to use `@args` and then do `args ? tests && args ? nvim` (etc)?

traxys: It's not a problem that none are defined, it's only a problem if both are defined

MattSturgeon: > It's not a problem that none are defined, it's only a problem if both are defined

:+1: ok

> I think this should be `&&`?

I think this is still the case though? Unless I'm missing some logic?

traxys: It should be fine, because the assert will trigger if `tests == [ ] || nvim == null` evaluates to `false`. This means that `tests != [ ]` and `nvim != null`, which is the error we are trying to detect.
(asserts are very confusing to check for error paths though --')

MattSturgeon: Yes, sorry I was getting mixed up! Maybe let's add a clarifying comment?

traxys: Done!

MattSturgeon (approved):
LGTM. Further improvements can be made in fresh PRs.

Bear in mind that we will need to backport to 24.05 as we also run buildbot on that branch. I think that branch has already diverged in a previous PR, so we'll likely need manual intervention.

MattSturgeon:
@mergifyio backport nixos-24.05

PR #1886, opened 2024-07-18
modules/performance: add an option to combine plugins to a single plugin pack

stasjok:
Hey. I'm the one from [this comment](https://github.com/nix-community/nixvim/issues/421#issuecomment-1646834969) in [[Feature] Lazy loading](https://github.com/nix-community/nixvim/issues/421) issue. I decided to give nixvim a try, but since I don't want to lose my performance optimizations, I implemented them in nixvim. Since my comment have got a couple of upvotes, here I am with a PR. I don't know if this would be merged, since this feature have caveats. But if it works for your config, you'll get a free performance boost (startup time and every time you `:e` a buffer). It's optional, so it shouldn't brake existing configs. I still think that this is more important than trying to lazy load everything (but for some specific cases lazy loading is also useful).

I'm planning to implement three performance options:

1) This one: combine plugins into a single pack
2) Optimize runtimepath further (basically set rtp to essential directories, removing everything else, mostly all system dirs). On my system with home-manager I have 14 useless items there. It's not yet implemented, but should be easy. For now I still have this in my config: https://github.com/stasjok/dotfiles/blob/36037f523185ba1409dd953999fda0f0db0dbd4f/nvim/init.lua#L1-L17, so I'm not in a hurry. It would depend on a feedback on this PR.
3) Byte compile every lua file. It's already implemented, will be a second PR.

To get you interested here are startuptime measurements (all measurements are taken from my config after a couple of tries with a hot disk cache):

1) No optimizations: 155 ms
2) All plugins combined: 95-100 ms
3) Everything byte compiled (including plugins and nvim runtime): 55-57 ms
4) rtp optimized: 53-55 ms (doesn't do much, to be honest)
5) For comparison, `vim.loader.enable()` without everything from the above: 115-120 ms

This PR adds a `performance.combinePlugins` option with two tunables: `pathsToLink` for choosing what paths are linked to plugin pack and `standalonePlugins` for specifying what plugins to exclude from combining. Last option is important, because for combining to work there must not be any file and doc tag collisions. Good news is if it builds, most of the time it would work correctly (except when some paths need to be added to `pathsToLink`).

Also this PR adds a couple of examples on how to fix errors in nixvim on a per plugin case (so most of the time user wouldn't need to configure anything). This is not all problematic plugins, but the ones I have in my config (to get this PR usable). Bad news came from a recent nixpkgs PR https://github.com/NixOS/nixpkgs/pull/321550. After that PR parser queries are colliding with treesitter plugins. It can be fixed in user overlay ([example](https://github.com/stasjok/dotfiles/blob/36037f523185ba1409dd953999fda0f0db0dbd4f/overlay/default.nix#L47-L61)), or like in this PR with nvim-treesitter (to exclude plugins containing a collection of queries from combining), or `meta.priority` can also be used. Because of this issue I even wanted to introduce a second layer of combined plugins, but for now a completely separate plugins would suffice. If the list of such exclusions will be too large, than it can be reevaluated.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -74,15 +74,54 @@ in
 
   config =
     let
+      # Plugin list extended with dependencies
+      allPlugins =
+        let
+          pluginWithItsDeps = p: [ p ] ++ builtins.concatMap pluginWithItsDeps p.dependencies or [ ];
+        in
+        lib.unique (builtins.concatMap pluginWithItsDeps config.extraPlugins);
+
+      # All plugins with doc tags removed
+      allPluginsOverridden = map (
+        plugin:
+        plugin.overrideAttrs (prev: {
+          nativeBuildInputs = lib.remove pkgs.vimUtils.vimGenDocHook prev.nativeBuildInputs or [ ];
+          configurePhase = builtins.concatStringsSep "\n" (
+            builtins.filter (s: s != ":") [
+              prev.configurePhase or ":"
+              "rm -vf doc/tags"
+            ]
+          );
+        })
+      ) allPlugins;
+
+      # Combine all plugins into a single pack
+      pluginPack = pkgs.vimUtils.toVimPlugin (
+        pkgs.buildEnv {
+          name = "plugin-pack";
+          paths = allPluginsOverridden;
+          inherit (config.performance.combinePlugins) pathsToLink;
+          # Remove empty directories and activate vimGenDocHook
+          postBuild = ''
+            find $out -type d -empty -delete
+            runHook preFixup
+          '';
+        }
+      );
+
+      # Combined plugins
+      combinedPlugins = [ pluginPack ];
+
+      # Plugins to use in finalPackage
+      plugins = if config.performance.combinePlugins.enable then combinedPlugins else config.extraPlugins;
+
       defaultPlugin = {
         plugin = null;
         config = "";
         optional = false;
       };
 
-      normalizedPlugins = map (
-        x: defaultPlugin // (if x ? plugin then x else { plugin = x; })
-      ) config.extraPlugins;
+      normalizedPlugins = map (x: defaultPlugin // (if x ? plugin then x else { plugin = x; })) plugins;
 
       neovimConfig = pkgs.neovimUtils.makeNeovimConfig (
         {
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -5,7 +5,10 @@
   helpers,
   ...
 }:
-with lib;
+let
+  inherit (lib) types mkOption;
+  inherit (lib) optional optionalString optionalAttrs;
+in
 {
   options = {
     viAlias = mkOption {
@@ -97,7 +100,7 @@ with lib;
         # Necessary to make sure the runtime path is set properly in NixOS 22.05,
         # or more generally before the commit:
         # cda1f8ae468 - neovim: pass packpath via the wrapper
-        // optionalAttrs (functionArgs pkgs.neovimUtils.makeNeovimConfig ? configure) {
+        // optionalAttrs (lib.functionArgs pkgs.neovimUtils.makeNeovimConfig ? configure) {
           configure.packages = {
             nixvim = {
               start = map (x: x.plugin) normalizedPlugins;
@@ -115,7 +118,9 @@ with lib;
       init = helpers.writeLua "init.lua" customRC;
 
       extraWrapperArgs = builtins.concatStringsSep " " (
-        (optional (config.extraPackages != [ ]) ''--prefix PATH : "${makeBinPath config.extraPackages}"'')
+        (optional (
+          config.extraPackages != [ ]
+        ) ''--prefix PATH : "${lib.makeBinPath config.extraPackages}"'')
         ++ (optional config.wrapRc ''--add-flags -u --add-flags "${init}"'')
       );
 
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -98,13 +98,26 @@ in
         in
         lib.unique (builtins.concatMap pluginWithItsDeps normalizedPlugins);
 
+      # Remove dependencies from all plugins in a list
+      removeDependencies = ps: map (p: p // { plugin = removeAttrs p.plugin [ "dependencies" ]; }) ps;
+
       # Separated start and opt plugins
       partitionedPlugins = builtins.partition (p: p.optional) allPlugins;
       startPlugins = partitionedPlugins.wrong;
       # Remove opt plugin dependencies since they are already available in start plugins
-      optPlugins = map (
-        p: p // { plugin = builtins.removeAttrs p.plugin [ "dependencies" ]; }
-      ) partitionedPlugins.right;
+      optPlugins = removeDependencies partitionedPlugins.right;
+
+      # Test if plugin shouldn't be included in plugin pack
+      isStandalone =
+        p:
+        builtins.elem p.plugin config.performance.combinePlugins.standalonePlugins
+        || builtins.elem (lib.getName p.plugin) config.performance.combinePlugins.standalonePlugins;
+
+      # Separated standalone and combined start plugins
+      partitionedStartPlugins = builtins.partition isStandalone startPlugins;
+      toCombinePlugins = partitionedStartPlugins.wrong;
+      # Remove standalone plugin dependencies since they are already available in start plugins
+      standaloneStartPlugins = removeDependencies partitionedStartPlugins.right;
 
       # Combine start plugins into a single pack
       pluginPack =
@@ -121,12 +134,12 @@ in
                 ]
               );
             })
-          ) startPlugins;
+          ) toCombinePlugins;
 
           # Python3 dependencies
           python3Dependencies =
             let
-              deps = map (p: p.plugin.python3Dependencies or (_: [ ])) startPlugins;
+              deps = map (p: p.plugin.python3Dependencies or (_: [ ])) toCombinePlugins;
             in
             ps: builtins.concatMap (f: f ps) deps;
 
@@ -149,7 +162,7 @@ in
 
           # Combined plugin configs
           combinedConfig = builtins.concatStringsSep "\n" (
-            builtins.concatMap (x: lib.optional (x.config != null && x.config != "") x.config) startPlugins
+            builtins.concatMap (x: lib.optional (x.config != null && x.config != "") x.config) toCombinePlugins
           );
         in
         normalize {
@@ -158,7 +171,7 @@ in
         };
 
       # Combined plugins
-      combinedPlugins = [ pluginPack ] ++ optPlugins;
+      combinedPlugins = [ pluginPack ] ++ standaloneStartPlugins ++ optPlugins;
 
       # Plugins to use in finalPackage
       plugins = if config.performance.combinePlugins.enable then combinedPlugins else normalizedPlugins;
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -74,64 +74,83 @@ in
 
   config =
     let
+      # Plugin normalization
+      normalize =
+        p:
+        let
+          defaultPlugin = {
+            plugin = null;
+            config = "";
+            optional = false;
+          };
+        in
+        defaultPlugin // (if p ? plugin then p else { plugin = p; });
+      normalizePluginList = plugins: map normalize plugins;
+
+      # Normalized plugin list
+      normalizedPlugins = normalizePluginList config.extraPlugins;
+
       # Plugin list extended with dependencies
       allPlugins =
         let
-          pluginWithItsDeps = p: [ p ] ++ builtins.concatMap pluginWithItsDeps p.dependencies or [ ];
+          pluginWithItsDeps =
+            p: [ p ] ++ builtins.concatMap pluginWithItsDeps (normalizePluginList p.plugin.dependencies or [ ]);
         in
-        lib.unique (builtins.concatMap pluginWithItsDeps config.extraPlugins);
+        lib.unique (builtins.concatMap pluginWithItsDeps normalizedPlugins);
 
-      # All plugins with doc tags removed
-      allPluginsOverridden = map (
-        plugin:
-        plugin.overrideAttrs (prev: {
-          nativeBuildInputs = lib.remove pkgs.vimUtils.vimGenDocHook prev.nativeBuildInputs or [ ];
-          configurePhase = builtins.concatStringsSep "\n" (
-            builtins.filter (s: s != ":") [
-              prev.configurePhase or ":"
-              "rm -vf doc/tags"
-            ]
-          );
-        })
-      ) allPlugins;
+      # Separated start and opt plugins
+      partitionedPlugins = builtins.partition (p: p.optional) allPlugins;
+      startPlugins = partitionedPlugins.wrong;
+      # Remove opt plugin dependencies since they are already available in start plugins
+      optPlugins = map (
+        p: p // { plugin = builtins.removeAttrs p.plugin [ "dependencies" ]; }
+      ) partitionedPlugins.right;
 
-      # Python3 dependencies from all plugins
-      python3Dependencies =
+      # Combine start plugins into a single pack
+      pluginPack =
         let
-          deps = map (p: p.python3Dependencies or (_: [ ])) allPluginsOverrided;
-        in
-        ps: builtins.concatMap (f: f ps) deps;
+          # Plugins with doc tags removed
+          overriddenPlugins = map (
+            plugin:
+            plugin.plugin.overrideAttrs (prev: {
+              nativeBuildInputs = lib.remove pkgs.vimUtils.vimGenDocHook prev.nativeBuildInputs or [ ];
+              configurePhase = builtins.concatStringsSep "\n" (
+                builtins.filter (s: s != ":") [
+                  prev.configurePhase or ":"
+                  "rm -vf doc/tags"
+                ]
+              );
+            })
+          ) startPlugins;
 
-      # Combine all plugins into a single pack
-      pluginPack = pkgs.vimUtils.toVimPlugin (
-        pkgs.buildEnv {
-          name = "plugin-pack";
-          paths = allPluginsOverridden;
-          inherit (config.performance.combinePlugins) pathsToLink;
-          # Remove empty directories and activate vimGenDocHook
-          postBuild = ''
-            find $out -type d -empty -delete
-            runHook preFixup
-          '';
-          passthru = {
-            inherit python3Dependencies;
-          };
-        }
-      );
+          # Python3 dependencies
+          python3Dependencies =
+            let
+              deps = map (p: p.plugin.python3Dependencies or (_: [ ])) startPlugins;
+            in
+            ps: builtins.concatMap (f: f ps) deps;
+        in
+        pkgs.vimUtils.toVimPlugin (
+          pkgs.buildEnv {
+            name = "plugin-pack";
+            paths = overriddenPlugins;
+            inherit (config.performance.combinePlugins) pathsToLink;
+            # Remove empty directories and activate vimGenDocHook
+            postBuild = ''
+              find $out -type d -empty -delete
+              runHook preFixup
+            '';
+            passthru = {
+              inherit python3Dependencies;
+            };
+          }
+        );
 
       # Combined plugins
-      combinedPlugins = [ pluginPack ];
+      combinedPlugins = [ (normalize pluginPack) ] ++ optPlugins;
 
       # Plugins to use in finalPackage
-      plugins = if config.performance.combinePlugins.enable then combinedPlugins else config.extraPlugins;
-
-      defaultPlugin = {
-        plugin = null;
-        config = "";
-        optional = false;
-      };
-
-      normalizedPlugins = map (x: defaultPlugin // (if x ? plugin then x else { plugin = x; })) plugins;
+      plugins = if config.performance.combinePlugins.enable then combinedPlugins else normalizedPlugins;
 
       neovimConfig = pkgs.neovimUtils.makeNeovimConfig (
         {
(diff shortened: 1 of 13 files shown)


stasjok:
Can someone tell me what test is failing? For me evaluation takes too much RAM that nix is getting killed by oom-killer.

stasjok:
Am I wrong thinking that buildbot instances also have not enough memory? All this strange `unexpected end-of-file` errors, looks like nix can't finish. I've added a pretty extensive testing for options here, but it still only 10 new tests.

traxys:
We have some issues currently with buildbot :/

modules/top-level/files/submodule.nix
```diff
@@ -22,6 +23,10 @@
       path = lib.mkDefault name;
       type = lib.mkDefault (if lib.hasSuffix ".vim" name then "vim" else "lua");
       # No need to use mkDerivedConfig; this option is readOnly.
-      plugin = helpers.writeLua derivationName config.content;
+      plugin =
+        let
+          writeContent = if config.type == "lua" then helpers.writeLua else pkgs.writeText;
+        in
+        writeContent derivationName config.content;
```
MattSturgeon, line 30: I think this is unrelated to this PR? (Incidentally, #1889 fixes this too)

modules/top-level/output.nix
```diff
@@ -121,7 +216,9 @@ with lib;
       init = helpers.writeLua "init.lua" customRC;
 
       extraWrapperArgs = builtins.concatStringsSep " " (
-        (optional (config.extraPackages != [ ]) ''--prefix PATH : "${makeBinPath config.extraPackages}"'')
+        (optional (
+          config.extraPackages != [ ]
+        ) ''--prefix PATH : "${lib.makeBinPath config.extraPackages}"'')
```
MattSturgeon, line 221: This is just removing `with lib`? Could that be done in a dedicated PR? It makes it more difficult to review this one

plugins/git/fugitive.nix
```diff
@@ -13,5 +13,12 @@ helpers.vim-plugin.mkVimPlugin config {
 
   maintainers = [ lib.maintainers.GaetanLepage ];
 
+  extraConfig = cfg: {
+    # mini.nvim and fugitive have duplicate doc tags
+    performance.combinePlugins.standalonePlugins = lib.mkIf (
```
MattSturgeon, line 18: Having all plugins merged by default and marked standalone on a case-by-case basis makes me nervous.

Is it possible to add test cases for the `combinePlugins` option so we can know for sure that all plugins still work when it is enabled?

plugins/telescope/default.nix
```diff
@@ -112,6 +112,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
         require('telescope').load_extension(extension)
       end
     '';
+
+    # planets picker requires files in data/memes/planets
+    performance.combinePlugins.pathsToLink = [ "/data/memes/planets" ];
```
MattSturgeon, line 117: Having to define this explicitly feels fragile... What would happen if we add another plugin module but don't notice we need to define addition paths to link?

MattSturgeon (commented):
This looks interesting! Thanks for working on it!

Unfortunately our CI is having issues atm so it's difficult to know with confidence nothing is broken. Hopefully this'll be resolved soon.

I'm pleased to see additional tests added, hopefully they cover most edge-cases?

I haven't got time to fully digest this right now, but I've left some initial thoughts below.

modules/top-level/files/submodule.nix
stasjok: That's true. I needed it for a test to pass, because it failed to create a .vim file. I already forgot that I changed it. It's better be fixed outside of this PR.

modules/top-level/output.nix
stasjok: It's in a separate first commit, changes are minimal. I did it to make language server smarter about undeclared variables and such. But yes, I can create a branch from the first commit and make a PR for it, but it's kind of strange to do it only for one file.

plugins/git/fugitive.nix
stasjok: If you mean to add a case with all plugins included, then I doubt it'll work. This option is not universal because of collisions. So it completely depends on what plugins user use. But in case of any collisions it won't build at all, so user will know instantly when he adds a new plugin to a config. User is expected to try to enable it, if it works - that's good, if not, than it need debugging. But some most often cases can be workarounded in nixvim beforehand, like I did with a couple of plugins here. So this performance option is not for all.

plugins/telescope/default.nix
stasjok: > What would happen if we add another plugin module but don't notice we need to define addition paths to link?

Additional paths will be missing from user runtime if `combinePlugins` is enabled, so it depends on what files are missing. For telescope here, it pretty much doesn't matter, because planets is a picker that is only for testing. Most plugins doesn't need any paths except standard.

I think in a description of this options should be clear, that this option is experimental, doesn't give any guaranties etc. So if it works for user's config, he'll get a performance boost, if not — sorry. I've added EXPERIMENTAL to the description for that purpose. But English is not native for me, if you have better description, we can update it.

stasjok:
> Unfortunately our CI is having issues atm so it's difficult to know with confidence nothing is broken. Hopefully this'll be resolved soon.

In a sibling PR all test are passed for two out of four arches. Those PR is branched on top of this, so I think all tests are passing. But to be honest I've noticed that most of the tests doesn't actually test functionality, only that config generated is valid and setup function is successful. In my tests I tried to test that my functionality actually work. Link to the x86_64 tests from a sibling PR: https://buildbot.nix-community.org/#/builders/146/builds/227

modules/top-level/output.nix
MattSturgeon: If it's needed elsewhere in this PR and it's a small change in its own commit then that's fine.

traxys:
@stasjok FYI we recently merged a PR that seemed to massively decrease the amount of RAM used by the tests

stasjok:
> @stasjok FYI we recently merged a PR that seemed to massively decrease the amount of RAM used by the tests

I've already rebased, but all tests are still `Broken pipe`.

modules/top-level/output.nix
traxys: Yeah at some point I'd like to remove all `with lib;`

MattSturgeon:
> I've already rebased, but all tests are still `Broken pipe`.

That's a separate issue that _should_ only affect buildbot (our CI). See https://github.com/nix-community/buildbot-nix/issues/227.



> we recently merged a PR that seemed to massively decrease the amount of RAM used by the tests

This should make it easier to run tests locally. Local tests should be unaffected by the broken pipe issue.

modules/top-level/output.nix
```diff
@@ -98,13 +98,26 @@ in
         in
         lib.unique (builtins.concatMap pluginWithItsDeps normalizedPlugins);
 
+      # Remove dependencies from all plugins in a list
+      removeDependencies = ps: map (p: p // { plugin = removeAttrs p.plugin [ "dependencies" ]; }) ps;
+
       # Separated start and opt plugins
       partitionedPlugins = builtins.partition (p: p.optional) allPlugins;
```
traxys, line 105: Maybe let's name this `partitionedOptionalPlugins`, to help with the other partition

modules/top-level/output.nix
```diff
@@ -98,13 +98,26 @@ in
...
+      optPlugins = removeDependencies partitionedPlugins.right;
+
+      # Test if plugin shouldn't be included in plugin pack
+      isStandalone =
+        p:
+        builtins.elem p.plugin config.performance.combinePlugins.standalonePlugins
+        || builtins.elem (lib.getName p.plugin) config.performance.combinePlugins.standalonePlugins;
+
+      # Separated standalone and combined start plugins
+      partitionedStartPlugins = builtins.partition isStandalone startPlugins;
```
traxys, line 117: Maybe `partitionedStandaloneStartPlugins` to help know how it's partitioned?

plugins/telescope/default.nix
traxys: I think the description is good. I think we can support this feature with less guarantees on breaking changes, i.e we won't add tests for _all_ the plugins trying to enable the option, as this might be a too large maintenance burden, but we'll gladly include workarounds if people suggest them

plugins/git/fugitive.nix
traxys: I'm not sure we want to mark them as standalone, maybe we only want to add a warning in some way, so that the user choses what to do:
- Disable fugitive
- Disable mini
- Mark one of them as standalone

We could make this check in the performance module in case we add others in the future.

This may be harder to test though, because we need to check that a warning is indeed generated

traxys:
Thank you very much for this feature, and especially thank you for the in-depth tests, they make me much more confident that we can maintain this feature without it breaking silently!

modules/top-level/output.nix
```diff
@@ -74,15 +74,54 @@ in
...
+
+      # All plugins with doc tags removed
+      allPluginsOverridden = map (
+        plugin:
+        plugin.overrideAttrs (prev: {
+          nativeBuildInputs = lib.remove pkgs.vimUtils.vimGenDocHook prev.nativeBuildInputs or [ ];
+          configurePhase = builtins.concatStringsSep "\n" (
+            builtins.filter (s: s != ":") [
+              prev.configurePhase or ":"
+              "rm -vf doc/tags"
```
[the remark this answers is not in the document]
MattSturgeon: I think the intention was to avoid the blank line being concatenated.

Using `optional` or filtering null would be better:

```nix
concatStringsSep "\n" (
  optional (prev ? configurePhase) prev.configurePhase
  ++ [ "rm -vf doc/tags" ]
)
```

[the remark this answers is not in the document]
traxys: I don't think it's too bad to have empty lines

modules/top-level/output.nix
```diff
@@ -74,64 +74,83 @@ in
...
+            plugin = null;
+            config = "";
+            optional = false;
+          };
+        in
+        defaultPlugin // (if p ? plugin then p else { plugin = p; });
+      normalizePluginList = plugins: map normalize plugins;
+
+      # Normalized plugin list
+      normalizedPlugins = normalizePluginList config.extraPlugins;
```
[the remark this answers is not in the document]
MattSturgeon: Don't have this specific example in front of me, but using `lib.pipe` often makes data-pipelines more readable without needing to resort to spamming a `let` block full of intermediate values.

modules/top-level/output.nix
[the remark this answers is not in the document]
stasjok: I usually do it this way to avoid empty lines. With your suggestion first line will be empty and it bother me when I look at a derivation with `nix derivation show`. Maybe it can be trimmed with some lib function instead.

modules/top-level/output.nix
```diff
@@ -95,6 +95,13 @@ in
         })
       ) allPlugins;
 
+      # Python3 dependencies from all plugins
+      python3Dependencies =
+        let
+          deps = map (p: p.python3Dependencies or (_: [ ])) allPluginsOverrided;
+        in
+        ps: builtins.concatMap (f: f ps) deps;
+
```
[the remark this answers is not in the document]
stasjok: Yes. We need to collect all the dependencies and provide them to `makeNeovimConfig`. I did it by attaching dependencies to a plugin pack, so that the following already existing code won't be changed, only the plugin list, but I guess those dependencies can be provided by concating with `extraPython3Packages`.

modules/top-level/output.nix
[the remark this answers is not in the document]
traxys: If you have a simpler way I'd not be against it, but I'd rather go with a bit simpler though annoying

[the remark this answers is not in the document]
traxys: I think @MattSturgeon could be good enough

modules/top-level/output.nix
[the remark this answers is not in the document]
stasjok: It is used later on, when deciding what plugin list to pass to `makeNeovimConfig`. If the option is not enabled.

modules/top-level/output.nix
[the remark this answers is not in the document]
MattSturgeon: > Maybe it can be trimmed with some lib function instead.

https://github.com/NixOS/nixpkgs/pull/315411 doesn't seem to be going anywhere :sob: 

> I usually do it this way to avoid empty lines.

See my above comment if that's important to you, although as I said @traxys' suggestion is simpler.
I don't like the idea of "magic" values like `":"`, though filtering `"" || null` would be ok too.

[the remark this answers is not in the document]
stasjok: `:` is not a magic string. Unfortunately it comes from the vimPlugins in nixpkgs. From here: https://github.com/NixOS/nixpkgs/blob/ad0b5eed1b6031efaed382844806550c3dcb4206/pkgs/applications/editors/vim/plugins/build-vim-plugin.nix#L18. So I did it long time ago and I didn't like both empty lines and redundant `:` if I just append to whatever is there. vimPlugins from nixpkgs always have configure phase defined with `:` there.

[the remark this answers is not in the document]
traxys: Then something like:
```nix
let oldPhase = if prev ? configurePhase && prev.configurePhase != ":" then configurePhase +'\n' else ""; in
   oldPhase + "rf -vf docs/tags";
```
?

It's a bit less confusing I think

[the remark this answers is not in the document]
stasjok: But I think I'll change it to the

```
configurePhase = ''
    ${prev.configurePhase or  ""}
    rm -vf doc/tags
'';
```

It was from my home config where I often bother myself with such things because of perfectionism. I pretty much distastes many generated in nix configs because of such things. But what good in nixvim, you are using stylua, so it fixes many things.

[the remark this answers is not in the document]
MattSturgeon: > `:` is not a magic string. Unfortunately it comes from the vimPlugins in nixpkgs.

In that case I think filtering it is fine, but it'd benefit from an explanatory comment.

Unless either of you disagree?

plugins/git/fugitive.nix
stasjok: Warning in this case probably doesn't make much sense, because it just won't build at all when fugitive and mini both enabled. To be honest I'm also not a big fan about splitting them by default. In my config I just patched mini and removed mini.git's doc (because I don't use mini.git and it's doc is not needed for me), thus fixing duplicate doc tags. So in a user config it's easier to mark it as standalone than to unmark it back (is it even possible without mkForce that would also disable all other fixes?). It was as an example here on how it can be fixed in nixvim itself. But I agree to add only the most common fixes by default (like nvim-treesitter, because pretty much everyone is using it).

traxys: I think there is a big difference between inter-plugin conflicts (where you have a choice of what to do) vs plugins with runtime stuff, where the fix is much obvious: make it standalone.

I know that it would error, but this does not mean that we could not generate a warning anyway, before having the conflict error I think. We may need to move some stuff around to ensure that we evaluate warnings before building neovim though, or we could have some special casing with `lib.warn` for this case

A warning is valuable because by experience people tend to open issues for such things (even if it marked experimental --')

stasjok: > I think there is a big difference between inter-plugin conflicts (where you have a choice of what to do) vs plugins with runtime stuff, where the fix is much obvious: make it standalone.

Could you rephrase it? I don't think I understand what's meant here.

> I know that it would error, but this does not mean that we could not generate a warning anyway, before having the conflict error I think.

But what would be a condition for the warning? If it's the same as here (mini.nvim + fugitive.vim both are enabled), than when user fixes the issue warning will stay... Add a check that none of the plugins are present in `standalonePlugins` to the condition? What if user fixed it some other way?

traxys: > Could you rephrase it? I don't think I understand what's meant here.

I think there is a difference between the tree-sitter situation and fugitive/mini. tree-sitter has a single "obvious" fix, being make tree-sitter a standalonePlugin, as it conflicts with something in the core neovim derivation. fugitive/mini does not have a single fix, there could be a number of resolutions, and I don't think nixvim should do something

> But what would be a condition for the warning? If it's the same as here (mini.nvim + fugitive.vim both are enabled), than when user fixes the issue warning will stay... Add a check that none of the plugins are present in `standalonePlugins` to the condition? What if user fixed it some other way?

Ah indeed did not think how to silence the warning. Then the better solution would maybe to add a new page in the user guide listing all conflicts between plugins when enabling this option

stasjok: OK. Dropped `fugitive.vim` commit.

traxys:
Could you rebase on main, we should have fixed our CI problems!

traxys:
I hate this error:
```

File: tests/test-sources/plugins/languages/treesitter/treesitter.nix [hunk 10]
```diff
@@ -120,4 +120,39 @@
       ];
     };
   };
+
+  combine-plugins = {
```
MattSturgeon, line 124: This test seems to fail on darwin (https://github.com/nix-community/nixvim/pull/1886#issuecomment-2247402248)

@traxys is `tests.dontRun = pkgs.stdenv.isDarwin` enough or should the relevant `enable` options be set to that instead?

traxys: No, we need to disable the `enable` I believe

traxys: (it's always annoying to hit cases like this --')

stasjok: What path do you want?

* It should work when splitting the tests (https://buildbot.nix-community.org/#/builders/2798/builds/1, https://buildbot.nix-community.org/#/builders/1856/builds/19).
* It probably can work when reducing the number of parsers (instead of all, select only a number of them).
* Disable test on darwin.

traxys: I'd say splitting the tests could be the best option, followed by reducing the parsers and disable the test as a last resort

stasjok: > I'd say splitting the tests could be the best option, followed by reducing the parsers and disable the test as a last resort

OK, I'll squash with treesitter commit when it finished if there won't be another opinions.

traxys:
>     2. Optimize runtimepath further (basically set rtp to essential directories, removing everything else, mostly all system dirs). On my system with home-manager I have 14 useless items there. It's not yet implemented, but should be easy. For now I still have this in my config: https://github.com/stasjok/dotfiles/blob/36037f523185ba1409dd953999fda0f0db0dbd4f/nvim/init.lua#L1-L17, so I'm not in a hurry. It would depend on a feedback on this PR.

This would a nice feature, while it does not have many performance benefits, I find annoying having so many useless entries!

traxys:
@Mergifyio queue

traxys:
It seemed to work, can you squash the test in the tree-sitter implementation?

stasjok:
> > 2. Optimize runtimepath further (basically set rtp to essential directories, removing everything else, mostly all system dirs). On my system with home-manager I have 14 useless items there. It's not yet implemented, but should be easy. For now I still have this in my config: https://github.com/stasjok/dotfiles/blob/36037f523185ba1409dd953999fda0f0db0dbd4f/nvim/init.lua#L1-L17, so I'm not in a hurry. It would depend on a feedback on this PR.
> 
> This would a nice feature, while it does not have many performance benefits, I find annoying having so many useless entries!

I've implemented it here: https://github.com/stasjok/nixvim/commits/optimize-rtp/ (last two commits), but I don't like it much, that's why I haven't submitted a PR. It turns out somewhat awkward having this in the performance namespace with the options to add extra paths to rtp:

https://github.com/stasjok/nixvim/blob/2551d1afa61c5271b08228fbf43cd6a870f7df97/modules/performance.nix#L7-L22

In the codebase I need to check `wrapRc` and if `performance.optimizeRuntimePath` is enabled. So it kind of strange to allow adding additional paths to rtp, but it works only if optimization is enabled and doesn't work otherwise...

So my thoughts are, if there is an interest in this feature, it shouldn't be `performance` options. Nixvim should do it by default with additional options to allow adding extra paths to rtp and packpath. It gives several benefits:

1. nixvim will be pure by default (only paths managed by nixvim will be in rtp by default), but user always can add additional paths with options. Some plugins, like treesitter, can add the paths there automatically when nix parser management is disabled for example.
2. filesPlugin should never be added as a plugin. This is wrong to begin with. That's because user files should always be first in rtp. When filesPlugins is in the mix with other plugins, overriding plugin files may not work (depending on sorting order).
3. I think that there should be a default internal variable with a path to user files. By default it point to filesPlugin, but other modules (wrappers) can override it with it's own path (home-manager should set it to xdg config path that is perfectly known to it without using raw lua, like I did now, nixos can override it with etc directory with configs and so on). Final rtp is: `configpath,plugins,nvimruntime,configafter`.
4. Potentially we can manage rtp partially with wrapper args, in that case we can even drop `wrapRc` completely and just place init.lua to filesPlugin. But this is debatable, maybe managing rtp paths as the first lines in init.lua is enough plus allow users to add paths with raw lua. Or it can mixed approach: user config path is added with command line args (so that the directory with init.lua is known to neovim), other paths are managed in init.lua.

In the end since all above as pretty large refactor and it should also change how things are managed right now, I put this task aside and decided to just change my lua code with a simple nix config in my config (in home-manager options I can get a static paths to my config files, to plugin path and to Nvim runtime, so no lua loops and matching needed, just plain vim.o.runtimepath assignment.

traxys:
Yeah I agree that it's something that we should do by default, we _already_ do some modifications of the runtime path, so it wouldn't be too out of place

traxys (approved):

octvs:
This (and the following PRs on the performance module) reduced more than half of my startup time, now tested on several hosts. All with a couple of lines of configuration change. Sincere applaud to @stasjok.

Mic92:
We added a couple of optimizations in buildbot-nix and it should be much faster handling large matrices by skipping over already build plugins.

PR #1887, opened 2024-07-18
modules/performance: add byteCompileLua option

stasjok:
This PR implements a second item from [this comment](https://github.com/nix-community/nixvim/issues/421#issuecomment-1646834969).

It can be merged only on top of #1886.

It introduces an option `performance.byteCompileLua`, that, when enabled, byte compiles lua files. It can be configured what to compile. By default it's user configs. But also plugins and neovim itself can be compiled.

### DIFF
--- a/lib/builders.nix
+++ b/lib/builders.nix
@@ -26,4 +26,26 @@
         --indent-width 4 \
         "$out"
     '';
+
+  /*
+    Write a byte compiled lua file to the nix store.
+
+    # Type
+
+    ```
+    writeByteCompiledLua :: String -> String -> Derivation
+    ```
+
+    # Arguments
+
+    - [name] The name of the derivation
+    - [text] The content of the lua file
+  */
+  writeByteCompiledLua =
+    name: text:
+    pkgs.runCommandLocal name { inherit text; } ''
+      echo -n "$text" > "$out"
+
+      ${lib.getExe' pkgs.luajit "luajit"} -bd -- "$out" "$out"
+    '';
 }
--- a/lib/builders.nix
+++ b/lib/builders.nix
@@ -1,5 +1,5 @@
 { lib, pkgs }:
-{
+rec {
   /*
     Write a lua file to the nix store, formatted using stylua.
 
@@ -48,4 +48,88 @@
 
       ${lib.getExe' pkgs.luajit "luajit"} -bd -- "$out" "$out"
     '';
+
+  /*
+    Get a source lua file and write a byte compiled copy of it
+    to the nix store.
+
+    # Type
+
+    ```
+    byteCompileLuaFile :: String -> String -> Derivation
+    ```
+
+    # Arguments
+
+    - [name] The name of the derivation
+    - [src] The path to the source lua file
+  */
+  byteCompileLuaFile =
+    name: src:
+    pkgs.runCommandLocal name { inherit src; } ''
+      ${lib.getExe' pkgs.luajit "luajit"} -bd -- "$src" "$out"
+    '';
+
+  # Setup hook to byte compile all lua files in the output directory.
+  # Invalid lua files are ignored.
+  byteCompileLuaHook = pkgs.makeSetupHook { name = "byte-compile-lua-hook"; } (
+    let
+      luajit = lib.getExe' pkgs.luajit "luajit";
+    in
+    pkgs.writeText "byte-compile-lua-hook.sh" # bash
+      ''
+        byteCompileLuaPostFixup() {
+            # Target is a single file
+            if [[ -f $out ]]; then
+                if [[ $out = *.lua ]]; then
+                    tmp=$(mktemp)
+                    ${luajit} -bd -- "$out" "$tmp"
+                    mv "$tmp" "$out"
+                fi
+                return
+            fi
+
+            # Target is a directory
+            while IFS= read -r -d "" file; do
+                tmp=$(mktemp -u "$file.XXXX")
+                # Ignore invalid lua files
+                if ${luajit} -bd -- "$file" "$tmp"; then
+                    mv "$tmp" "$file"
+                else
+                    echo "WARNING: Ignoring byte compiling error for '$file' lua file" >&2
+                fi
+            done < <(find "$out" -type f,l -name "*.lua" -print0)
+        }
+
+        postFixupHooks+=(byteCompileLuaPostFixup)
+      ''
+  );
+
+  /*
+    Returns an overridden derivation with all lua files byte compiled.
+
+    # Type
+
+    ```
+    byteCompileLuaDrv :: Derivation -> Derivation
+    ```
+
+    # Arguments
+
+    - [drv] Input derivation
+  */
+  byteCompileLuaDrv =
+    drv:
+    drv.overrideAttrs (
+      prev:
+      {
+        nativeBuildInputs = prev.nativeBuildInputs or [ ] ++ [ byteCompileLuaHook ];
+      }
+      // lib.optionalAttrs (prev ? buildCommand) {
+        buildCommand = ''
+          ${prev.buildCommand}
+          runHook postFixup
+        '';
+      }
+    );
 }
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -2,14 +2,16 @@
   lib,
   helpers,
   pkgs,
+  config,
   ...
 }:
 let
-  fileType = lib.types.submodule (
+  fileTypeModule =
     {
       name,
       config,
       options,
+      topConfig,
       ...
     }:
     {
@@ -41,6 +43,14 @@ let
           type = lib.types.path;
           description = "Path of the source file.";
         };
+
+        finalSource = lib.mkOption {
+          type = lib.types.path;
+          description = "Path to the final source file.";
+          readOnly = true;
+          visible = false;
+          internal = true;
+        };
       };
 
       config =
@@ -54,9 +64,29 @@ let
             # This means our `source` definition has the same priority as `text`.
             lib.mkDerivedConfig options.text (pkgs.writeText derivationName)
           );
+          finalSource =
+            # Byte compile lua files if performance.byteCompileLua option is enabled
+            if
+              lib.hasSuffix ".lua" config.target
+              && topConfig.performance.byteCompileLua.enable
+              && topConfig.performance.byteCompileLua.configs
+            then
+              if lib.isDerivation config.source then
+                # Source is a derivation
+                helpers.byteCompileLuaDrv config.source
+              else
+                # Source is a path or string
+                helpers.byteCompileLuaFile derivationName config.source
+            else
+              config.source;
         };
-    }
-  );
+    };
+
+  fileType = lib.types.submoduleWith {
+    shorthandOnlyDefinesConfig = true;
+    modules = [ fileTypeModule ];
+    specialArgs.topConfig = config;
+  };
 
   # TODO: Added 2024-07-07, remove after 24.11
   # Before we had a fileType, we used types.str.
--- a/modules/performance.nix
+++ b/modules/performance.nix
@@ -12,6 +12,12 @@ in
         default = true;
         example = false;
       };
+      configs = lib.mkOption {
+        description = "Whether to byte compile lua configuration files.";
+        type = types.bool;
+        default = true;
+        example = false;
+      };
     };
 
     combinePlugins = {
--- a/modules/performance.nix
+++ b/modules/performance.nix
@@ -21,6 +21,9 @@ in
       plugins = lib.mkEnableOption "plugins" // {
         description = "Whether to byte compile lua plugins.";
       };
+      nvimRuntime = lib.mkEnableOption "nvimRuntime" // {
+        description = "Whether to byte compile lua files in Nvim runtime.";
+      };
     };
 
     combinePlugins = {
--- a/modules/performance.nix
+++ b/modules/performance.nix
@@ -18,6 +18,9 @@ in
         default = true;
         example = false;
       };
+      plugins = lib.mkEnableOption "plugins" // {
+        description = "Whether to byte compile lua plugins.";
+      };
     };
 
     combinePlugins = {
--- a/modules/performance.nix
+++ b/modules/performance.nix
@@ -4,6 +4,16 @@ let
 in
 {
   options.performance = {
+    byteCompileLua = {
+      enable = lib.mkEnableOption "byte compiling of lua files";
+      initLua = lib.mkOption {
+        description = "Whether to byte compile init.lua.";
+        type = types.bool;
+        default = true;
+        example = false;
+      };
+    };
+
     combinePlugins = {
       enable = lib.mkEnableOption "combinePlugins" // {
         description = ''
--- a/wrappers/_shared.nix
+++ b/wrappers/_shared.nix
@@ -53,11 +53,11 @@ in
         setAttrByPath filesOpt (
           listToAttrs (
             map (
-              { target, source, ... }:
+              { target, finalSource, ... }:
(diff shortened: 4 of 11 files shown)


stasjok:
It looks like a lot of code that I traced to find any parts generating config files would be changed by #1889. Only three last commits beginning with "modules/performance: add byteCompileLua option" is related to this PR, but it'll definitely will conflict (as is the second my PR). So I guess I'll have to re-implement things.

wrappers/_shared.nix
```diff
@@ -42,18 +43,34 @@ in
 
   config = mkMerge [
     # Make our lib available to the host modules
-    { nixvim.helpers = lib.mkDefault (import ../lib/helpers.nix { inherit pkgs lib; }); }
```
traxys, line 45: @MattSturgeon wasn't this necessary?

MattSturgeon: Technically this is unchanged, the import is moved to the `let` block above.

I'd prefer to keep the import here and have `inherit (config.nixvim) helpers` in the let-block instead, though.

wrappers/_shared.nix
```diff
@@ -42,18 +43,34 @@ in
...
+    { nixvim.helpers = lib.mkDefault helpers; }
     # Propagate extraFiles to the host modules
     (optionalAttrs (filesOpt != null) (
       mkIf (!cfg.wrapRc) (
         setAttrByPath filesOpt (
           listToAttrs (
             map (
               { target, source, ... }:
+              let
+                needByteCompiling = cfg.performance.byteCompileLua.enable && cfg.performance.byteCompileLua.configs;
```
MattSturgeon, line 55: Is this the right place to be doing this? The `_shared.nix` module is only used by the home-manager/NixOS/nix-darwin modules; anything here will not affect the standalone build.

If all builds should be affected, we probably want something in `/modules/top-level`.

(Apologies if I missed the point of the changes, I'm on my phone and haven't looked properly)

stasjok: It is now done with the minimum changes to the options itself (in fact probably no changes at all). So I did it in the final stages just before final files a generated. I found it in two places: first is nixvim itself manages the files in the `nvim-config` plugin. Second place is nixvim is propagating user files to the upper module. For now it's used only with home-manager, because all other modules have `wrapRc = lib.mkForce true;`. User can inspect its config with unmodified files (content, text, source attributes).

Second approach I considered is populating all relevant places in the options itself, but it'll take a refactor of the options. In fact you already done at least part of it in #1889. So when I wrote this I thought that my approach is simpler and less intrusive to nixvim framework. But after your PR it can be re-evaluated, maybe it'll be simpler by doing it _right_.

The following is my thoughts about doing it in options on top of the current main (not your PR), I didn't done it in practice, so maybe I'm overlooking something.

1) Since my `performance` module is in main modules it is populated to submodules (`files`). So we need to make it's default value of `performance.byteCompileLua` to follow top-level value. Can an option in submodule be set with `mkDefault` to the value of top-level module? I'm sure it shouldn't be a problem. User can even disable a particular file from compiling.

2) In the `files` submodule the `plugin` option can be optionally compiled (where `writeLlua`, `writeText` and `writeContent` are).

3) In the top level I didn't find an option with `init.lua` file, it's generated in `let in` block (but your PR fixes that). Also `nixvim-print-init` should be handled (it should print a text file, not the one possibly compiled).

4) All above was easy, but what requires more changes (in the same places as in this PR: `nvim-config` and `shared` module) is `extraFiles`. `files` submodule is also goes there in the end. It has `source` option, but I can't change it, because it can be set by the user. So I need some other option introduced, `finalSource` for example. And it would be populated with source unchanged or compiled depending on the option value and whether the file is lua. `files` submodule also should populate `finalSource`.

5) Maybe `files` submodule shouldn't decide for itself if the file should be compiled. In that case it should populate `source` option in `extraFiles` (as it is no), and `finalPackage` in `extraFiles` is populated depending on the global performance option. But this should be done with care, because top level init.lua module and `files` submodules share the same configs mostly so the differences could be tricky. But maybe neither `init.lua`, nor `files` submodule should do this and it can be left only in `extraFiles`. `init.lua` can be handled specially as it is in this PR.

MattSturgeon: > 1. Can an option in submodule be set with `mkDefault` to the value of top-level module? I'm sure it shouldn't be a problem. User can even disable a particular file from compiling.

This is possible when declaring the submodule type, since submodules can contain `options` `imports` and `config` attrs just like any other module.

You could do:

```nix
lib.types.submodule {
  options = {
    # TODO
  };

  config = {
    # This will inherit the priority from the top-level definition:
    performance.byteCompile = lib.mkDerivedConfig options.performance.byteCompile (v: v);

   # Or this will do the same for all the options in `performance`
   performance = lib.mkAliasAndWrapDefsWithPriority (v: v) options.performance;
  };
}
```

> 2. In the `files` submodule the `plugin` option can be optionally compiled (where `writeLua`, `writeText` and `writeContent` are).

I'm missing the significance of this.

> 3. In the top level I didn't find an option with `init.lua` file, it's generated in `let in` block (but your PR changes that).

If it helps we can look into getting specific parts of that refactor merged separately. That'll also make the refactor easier to test/review since it'll reduce its scope.

> Also `nixvim-print-init` should be handled (it should print a text file, not the one possibly compiled).

What do you mean? Why would this need special handling? Are you pre-compiling the init.lua too?



> * All above was easy, but what requires more changes (in the same places as in this PR: `nvim-config` and `shared` module) is `extraFiles`. `files` submodule is also goes there in the end. It has `source` option, but I can't change it, because it can be set by the user. So I need some other option introduced, `finalSource` for example. And it would be populated with source unchanged or compiled depending on the option value and whether the file is lua. `files` submodule also should populate `finalSource`.
> 
> * Maybe `files` submodule shouldn't decide for itself if the file should be compiled. In that case it should populate `source` option in `extraFiles` (as it does now), and `finalSource` in `extraFiles` is populated depending on the global performance option. But this should be done with care, because top level init.lua module and `files` submodules share the same configs mostly so the differences could be tricky. But maybe neither `init.lua`, nor `files` submodule should do this and it can be left only in `extraFiles`. `init.lua` can be handled specially it `top-level/output.nix` as it is in this PR.

Honestly, I struggled to absorb all of that. But I agree we shouldn't be overriding user-definable options like `extraFiles.*.source` with a compiled version of itself.
I think that'd require infinite recursion anyway.

A `compiledFiles`, `compiledPlugin`, or some other name could certainly be introduced.

---

We also need to watch out for "Import From Derivation" when consuming other derivations (`pkgs.writeText`, produce a derivation).

It's fine to pass a derivation's path into another derivation writer/builder, but we can't import/read the derivation path's _files_ (content) into the nix evaluation itself.

stasjok: > > 3\. In the `files` submodule the `plugin` option can be optionally compiled (where `writeLua`, `writeText` and `writeContent` are).
> 
> I'm missing the significance of this.

Your original point was that I shouldn't done anything in _shared.nix. The only way I know to _not doing this in shared.nix_, is to have all the files generated in files and/or extraFiles submodules. All files should already be compiled properly and separately (without any filePlugin). And if I understand this right, the way files are propagated to extraFiles right now is that `source` of extraFiles is set to `plugin` files. So my idea was, that I just change plugin to write a compiled file and it would automatically get to extraFiles, and that to home-manager.

> > Also `nixvim-print-init` should be handled (it should print a text file, not the one possibly compiled).
> 
> What do you mean? Why would this need special handling? Are you pre-compiling the init.lua too?

If a top-level module has some e.g. `finalSource` option which would have a path to a compiled init.lua, it means that `nixvim-print-init` need to be pointed to another uncompiled copy of the init.lua. That's all I meant.

> We also need to watch out for "Import From Derivation" when consuming other derivations (`pkgs.writeText`, produce a derivation).

Ah I did it because my library function writeByteCompiledLua takes string and I only have generated file in that place. I though that it's alright.

Main question: are you planing to finish your refactor PR anytime soon? If I understand correctly, my current implementation is unsuitable (although it works fine for me). If I need to rewrite it, I need to know what is the state of that refactor PR, because we touch the same places.

MattSturgeon: > Main question: are you planing to finish your refactor PR anytime soon?

I'd like to, but we have higher priorities currently (e.g. the CI being broken, which blocks many PRs including this).

I also don't know what the other maintainers' views are on that refactor PR, like any PR it'll need their approval before merging.

As I said, if there's specific features from that PR (or features related to it) that would help you out implementing this, I could look into making a smaller PR that adds just those.

As I understand it, this PR is also blocked by #1886, so I wouldn't rush to rewrite stuff until the other blockers are resolved.

stasjok:
Rebased and changed implementation for byte compiling extraFiles. Other implementations stayed as it was (in particularly init.lua).

I hope this time it's alright, because it'll be frustrating if it needs to be rewritten again.

modules/files.nix
```diff
@@ -54,9 +64,40 @@ let
...
+                  prev:
+                  {
+                    nativeBuildInputs = prev.nativeBuildInputs or [ ] ++ [ helpers.byteCompileLuaHook ];
+                  }
+                  // lib.optionalAttrs (prev ? buildCommand) {
+                    buildCommand = ''
+                      ${prev.buildCommand}
+                      runHook postFixup
+                    '';
+                  }
```
traxys, line 86: Exposing this function in lib/builders.nix could be useful, as it provides an easy way to byte-compile selected plugins if required. It should also allow to re-use the same function here & in overriding the plugins, reducing the risk somebody else forgets to modify one if need arises.

traxys (commented):
A few question, but I'm mostly ok with the code

MattSturgeon:
> Running test for default
    > ERROR: Error detected while processing /nix/store/21ahy9d40fhvzijkdvdascnz60d4py71-nvim-config/plugin/test.lua:
    > E5113: Error while calling lua chunk: ...y9d40fhvzijkdvdascnz60d4py71-nvim-config/plugin/test.lua:16: plugin/file_text.lua is expected to be byte compiled, but it's not


`file_text.lua is expected to be byte compiled, but it's not`

modules/files.nix
```diff
@@ -54,9 +64,40 @@ let
...
+                ))
+              else
+                # Source is a path or string
+                helpers.writeByteCompiledLua derivationName (builtins.readFile config.source)
```
MattSturgeon, line 90: This will lead to IFD if the `source` is a path within a derivation, or a stringified derivation.

To avoid this we'd need a builder function that accepts a path (instead of string content) and doesn't read the path's content during evaluation (only during its build)

modules/performance.nix
```diff
@@ -4,6 +4,24 @@ let
...
+    byteCompileLua = {
+      enable = lib.mkEnableOption "byte compiling of lua files";
+      initLua = lib.mkEnableOption "initLua" // {
+        description = "Whether to byte compile init.lua.";
+        default = true;
+      };
+      configs = lib.mkEnableOption "configs" // {
+        description = "Whether to byte compile lua configuration files.";
+        default = true;
+      };
```
traxys, line 16: I think it'd be clearer to not use mkEnableOption here, as I feel that it should be reserved for option with default = false

lib/builders.nix
```diff
@@ -26,4 +26,46 @@
...
+  byteCompileLuaHook = pkgs.makeSetupHook { name = "byte-compile-lua-hook"; } (
+    let
+      luajit = lib.getExe' pkgs.luajit "luajit";
+    in
+    pkgs.writeText "byte-compile-lua-hook.sh" # bash
+      ''
+        byteCompileLuaPostFixup() {
+            while IFS= read -r -d "" file; do
+                tmp=$(mktemp -u "$file.XXXX")
+                if ${luajit} -bd -- "$file" "$tmp"; then
```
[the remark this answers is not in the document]
stasjok: Glad you asked. I want another opinion here. So in my original version in my dotfiles byte compiling was failing on any error. Usually it's the best course of action to fail fast and not ignoring errors. But there is one case pretty much anyone would stumble on: [this test file in nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/tests/indent/lua/cond.lua) is invalid lua and the build fails because of that. In my dotfiles I have the freedom to just override nvim-treesitter in overlay and remove tests from there. So to make using this option smooth I opted to ignore errors. In this case if it's encountering invalid lua file it leaves it as is.

Potential courses of action:

* Ignore errors. Errors are still there in the build logs, if someone want to look. Pros: no workarounds required. Cons: some other errors can potentially be ignored, not only invalid lua (you can see it in the Darwin test failures because of my bug).
* Add a dirty workaround for nvim-treesitter: if the option is enabled, override before adding to extraPlugins (removing tests). Pros: errors are not ignored. Cons: not pretty, other potential cases can arise where a directory has invalid .lua files for whatever reasons.
* Leave the user alone with the build failures.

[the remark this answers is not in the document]
traxys: Am I right in that not _all_ lua files are accessible from neovim, only those in select directories? Could we filter on that, or am I remembering wrong?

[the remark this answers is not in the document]
stasjok: They are all acceptable, there is no a restriction to that. But most of the time only files in lua, ftplugin, plugin, syntax and other standard directories are used. But anyone can `:runtime` any lua file, fzy-native plugin has some lua code in `deps` directory, so I wouldn't want to restrict it to only standard directories.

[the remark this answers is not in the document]
stasjok: Just to be clear: I'm quite OK with the current behavior. But maybe some of the maintainers have other opinions.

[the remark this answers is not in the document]
traxys: Yeah indeed. Maybe we could add an echo to stderr whenever we encounter an invalid file, this could be used to debug if we have any issues with bytecode compilation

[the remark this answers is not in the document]
stasjok: > Yeah indeed. Maybe we could add an echo to stderr whenever we encounter an invalid file, this could be used to debug if we have any issues with bytecode compilation

luajit is already outputting all its errors, that's why I didn't add my own echo. You can check it right now in the buildbot logs (I specifically added nvim-treesitter to the plugins in tests): https://buildbot.nix-community.org/api/v2/logs/6970/raw_inline

```
vimplugin-nvim-treesitter> luajit: ...gin-nvim-treesitter-2024-07-17/tests/indent/lua/cond.lua:15: 'end' expected (to close 'if' at line 14) near '<eof>'
vimplugin-nvim-treesitter> luajit: ...gin-nvim-treesitter-2024-07-17/tests/indent/lua/func.lua:12: 'end' expected (to close 'function' at line 11) near '<eof>'
```

But maybe I can add something else to clarity, like: `Error compiling $file. Skipping.`.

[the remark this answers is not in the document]
traxys: Yeah that would be great, users may not understand why luajit is being called

GaetanLepage (approved):
Sorry for the late review.
I had an overall look at the PR and trust @MattSturgeon and @traxys for checking the details.

Thank you very much @stasjok for this amazing work. Also the effort you have put in documentation and testing is truly appreciable !

stasjok:
Will it get merged?

GaetanLepage:
Sure !  Let's give this a shot. Thanks for the hard work !

@Mergifyio queu

GaetanLepage:

MattSturgeon:
(I'm going to re-queue this so it merges _after_ #1956, since that PR doesn't need rebasing)

MattSturgeon:
@mergifyio dequeue

MattSturgeon:
@mergifyio queue

MattSturgeon:
@stasjok thanks again for all your hard work, persistence and dedication!

PR #1891, opened 2024-07-19
modules/files: fix creating configs of vim type

stasjok:
This PR is a fix extracted from #1886. Since #1889 also fixes that, maybe this PR is redundant. But at least tests are fixed here (I think that `files.nix` was placed in the wrong directory, because I didn't find any reference to it). Also added a test case for the error this is fixing. Feel free to close if #1889 is merged earlier.


modules/top-level/files/submodule.nix
```diff
@@ -22,6 +23,10 @@
       path = lib.mkDefault name;
       type = lib.mkDefault (if lib.hasSuffix ".vim" name then "vim" else "lua");
       # No need to use mkDerivedConfig; this option is readOnly.
-      plugin = helpers.writeLua derivationName config.content;
+      plugin =
+        let
+          writeContent = if config.type == "lua" then helpers.writeLua else pkgs.writeText;
```
MattSturgeon, line 28: I think it's more readable with this in the same let block as `derivationName`.

Also: in #1889 I named the function `writer`, although that's not really important. `writeContent` is fine, just a little verbose.

MattSturgeon (changes requested):
Thanks! Definetely better to address this before dealing with the other PRs.

Would you mind addressing my nit about using just one `let` block?

I'll run tests locally, to be sure, then I'm happy to merge!

MattSturgeon:
Also, could you rebase on `main`?

MattSturgeon:
Update: `nix flake check` passed locally (x86_64-linux). Happy to merge once rebased and nits addressed.

MattSturgeon (approved):
Great, thanks!

modules/top-level/files/submodule.nix
stasjok: Maybe I'm wrong, but isn't a verb better in that case? I named it by analogy to `writeLua` and `writeText`.

PR #1893, opened 2024-07-19
plugins/lsp/pylsp: `propagatedBuildInputs` -> `dependencies`

PerchunPak:
See https://github.com/NixOS/nixpkgs/pull/327630#discussion_r1679253679

The only (patched) pylsp plugin that doesn't use `dependencies` yet is `pylsp-rope`, though it will change soon with https://github.com/NixOS/nixpkgs/pull/328410 ([pr-tracker](https://nixpk.gs/pr-tracker.html?pr=328410)).

### DIFF
--- a/plugins/lsp/language-servers/pylsp.nix
+++ b/plugins/lsp/language-servers/pylsp.nix
@@ -536,14 +536,28 @@ in
           mapAttrsToList
             (
               pluginName: nixPackage:
+              let
+                # Get rid of the python-lsp-server dependency
+                filterDependencies = dependencies: filter (dep: dep.pname != "python-lsp-server") dependencies;
+              in
               (optional (isEnabled plugins.${pluginName}) (
-                nixPackage.overridePythonAttrs (old: {
-                  # Get rid of the python-lsp-server dependency
-                  propagatedBuildInputs = filter (dep: dep.pname != "python-lsp-server") old.propagatedBuildInputs;
-
-                  # Skip testing because those naked dependencies will complain about missing pylsp
-                  doCheck = false;
-                })
+                nixPackage.overridePythonAttrs (
+                  old:
+                  (lib.genAttrs
+                    # NOTE: Depending on the exact nixpkgs rev, we may have `dependencies` or `propagatedBuildInputs`
+                    # See https://github.com/NixOS/nixpkgs/pull/327630#discussion_r1679253679
+                    # TODO: Only filter `dependencies`
+                    (lib.filter (n: old ? ${n}) [
+                      "propagatedBuildInputs"
+                      "dependencies"
+                    ])
+                    (n: filterDependencies old.${n})
+                  )
+                  // {
+                    # Skip testing because those naked dependencies will complain about missing pylsp
+                    doCheck = false;
+                  }
+                )
               ))
             )
             (


MattSturgeon (approved):
LGTM, thanks!

Does this fix anything or is it just updating to the new code style?

@GaetanLepage if you have time, can you also review?

MattSturgeon:
The build [fails](https://buildbot.nix-community.org/#/builders/146/builds/230) with:

> error: attribute 'dependencies' missing

traxys:
I think @GaetanLepage said he'd deal with this himself using the update action

MattSturgeon:
This feels like another thing that'll break anyone who's nixpkgs lock is out of sync with outs...

PerchunPak:
Yeah, I actually created this PR because of the error described in #1894. In nixpkgs' master every plugin uses `dependencies` already, so feel free to merge this when https://github.com/NixOS/nixpkgs/pull/328410 gets to unstable (this will fix the current evaluation error)

MattSturgeon:
I'm wondering if we should have a more graceful approach than handles either the old/new attr depending on which is actually present, at least for a short transition period?

PerchunPak:
Oh, yeah. I also just noticed that this rename probably won't be backported to 24.05

MattSturgeon:
> Oh, yeah. I also just noticed that this rename probably won't be backported to 24.05

That's not an issue, since our `main` branch isn't intended to work on the stable channel.

plugins/lsp/language-servers/pylsp.nix
```diff
@@ -536,14 +536,25 @@ in
...
-                  # Skip testing because those naked dependencies will complain about missing pylsp
-                  doCheck = false;
-                })
+                nixPackage.overridePythonAttrs (
+                  old:
+                  mkMerge [
+                    (mkIf (old ? propagatedBuildInputs) {
+                      propagatedBuildInputs = filterDependencies old.propagatedBuildInputs;
+                    })
+                    (mkIf (old ? dependencies) { dependencies = filterDependencies old.dependencies; })
```
MattSturgeon, line 550: This won't work, because `mkIf` is "smart" and evaluates even if the condition is false.

Since (unlike most conditional config) we _don't_ want to evaluate in this case, we should use an `if then else` wrapper such as `lib.optionalAttrs` instead.

Also, I don't think `mkMerge` and `mkIf` can be used in an override function, since they are only intended for use with module config definitions.

Something like this should work:

```nix
(
  old:
  # Note: Depending on the exact nixpkgs rev, we either need to filter `dependencies` or `propagatedBuildInputs`
  # TODO: Only filter `dependencies`
  lib.optionalAttrs (old ? propagatedBuildInputs) (filterDependencies old.propagatedBuildInputs)
  // lib.optionalAttrs (old ? dependencies) (filterDependencies old.dependencies)
)
```

MattSturgeon: Or using `genAttrs`:

```nix
(
  old:
    lib.genAttrs
    (lib.filter (n: old ? ${n}) [
      "propagatedBuildInputs"
      "dependencies"
    ])
    (n: filterDependencies old.${n})
)
```

PerchunPak: this looks so much better!

plugins/lsp/language-servers/pylsp.nix
```diff
@@ -536,14 +536,28 @@ in
...
+                  old:
+                  (lib.genAttrs
+                    # NOTE: Depending on the exact nixpkgs rev, we may have `dependencies` or `propagatedBuildInputs`
+                    # See https://github.com/NixOS/nixpkgs/pull/327630#discussion_r1679253679
+                    # TODO: Only filter `dependencies`
+                    (lib.filter (n: old ? ${n}) [
+                      "propagatedBuildInputs"
+                      "dependencies"
+                    ])
+                    (n: filterDependencies old.${n})
```
MattSturgeon, line 554: Since this is done gracefully I don't see why we can't merge now?

MattSturgeon (approved):
LGTM, but I'll let @GaetanLepage have the final say since he's dealing with the lockfile update and the relevant upstream changes?

plugins/lsp/language-servers/pylsp.nix
PerchunPak: If you are talking about merging the PR, tests were failing. I just merged `main` to here and tests are green again

MattSturgeon: Our CI is currently disabled, so the only "test" running on GitHub is `mergify` (which doesn't actually test any code).

Or are you referring to tests run locally?

PerchunPak: I referred to [this test failure](https://buildbot.nix-community.org/#/builders/43/builds/287) in 6066ff2874c3b12a2418d91ca508e18340585276, and local tests fail for me even on main with some huge and totally unrelated error.

MattSturgeon: > I referred to [this test failure](https://buildbot.nix-community.org/#/builders/43/builds/287) in [6066ff2](https://github.com/nix-community/nixvim/commit/6066ff2874c3b12a2418d91ca508e18340585276),

The linked failure is #1878 (one of the reasons CI is currently disabled).

> local tests fail for me even on main with some huge and totally unrelated error.

Strange. I've just run the test suite locally on another PR and everything passed just fine...

Are you using the `tests` devshell command or just `nix flake check`? Are you trying to use `--all-systems`?

MattSturgeon: Just ran `tests` on this branch, and everything passes for me.

PerchunPak: I used `nix flake check --all-systems -v` and got these logs:

[logs.txt.gz](https://github.com/user-attachments/files/16338335/logs.txt.gz)
(too big for pastebin, so sending as a file)

Anyway, this doesn't really matter. If all tests are passing, why not merge this PR?

MattSturgeon: > why not merge this PR?

As mentioned above, I'm delegating to @GaetanLepage here since he's dealing with these changes in nixpkgs.

GaetanLepage:
Thank you very much @PerchunPak for spinning up this fix so quickly.
I should have merged it earlier to fix it instead of waiting for the upstream fix to land on `nixos-unstable`.
Now that `pylsp-rope` also makes use of `dependencies`, I have directly used it in https://github.com/nix-community/nixvim/pull/1900/commits/9f9e20205a5631474120f2a2cd10d4ff32a6abed and the problem should now be gone.

Please, feel free to report any eventual weird behavior.

PR #1895, opened 2024-07-19
tests: Collect all failures in a test derivation

traxys:
This runs all the test cases for a test derivation, and prints out all failures at the end, instead of stopping at the first failure

### DIFF
--- a/tests/test-derivation.nix
+++ b/tests/test-derivation.nix
@@ -38,24 +38,39 @@ let
       buildPhase = lib.optionalString (!dontRun) (
         ''
           mkdir -p .cache/nvim
+          status=0
+          declare -A errors
         ''
-        + lib.concatStringsSep "\n" (
-          builtins.map (
-            {
-              derivation,
-              name,
-              dontRun ? false,
-            }:
-            lib.optionalString (!dontRun) ''
-              echo "Running test for ${name}"
+        + (
+          lib.concatStringsSep "\n" (
+            builtins.map (
+              {
+                derivation,
+                name,
+                dontRun ? false,
+              }:
+              lib.optionalString (!dontRun) ''
+                echo "Running test for ${name}"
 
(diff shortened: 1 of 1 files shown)


tests/test-derivation.nix
```diff
@@ -38,24 +38,39 @@ let
...
-              if [[ -n $output ]]; then
-                echo "ERROR: $output"
-                exit 1
-              fi
-            ''
-          ) testList
+                output=$(HOME=$(realpath .) ${lib.getExe derivation} -mn --headless "+q" 2>&1 >/dev/null)
+                if [[ -n $output ]]; then
+                  echo "  test failure"
+                  errors[${name}]=$output
```
MattSturgeon, line 58: Would it be simpler to print `$output` here (and set `$status`) instead of collecting + printing later?

MattSturgeon (commented):
Awesome! Seems to work, I just wonder if we can do this in a single loop?

tests/test-derivation.nix
traxys: I'd like to avoid that, as I find it easier to look at then end of the build!

MattSturgeon: Ah, wasn't sure if the build was full of other output or fairly empty.

Would it make sense to print twice then? Once in context and again at the end?

MattSturgeon: Not pushing for this, just trying to understand the best options

traxys: I think you are right, in that there would be little output. I'll change that

MattSturgeon: I'm happy either way, assuming the output is readable. Whatever you prefer :grin:

MattSturgeon (approved):
Not sure if you plan on changing the implementation, but as I said I'm happy either way, here's an approving review.

MattSturgeon:
Can we revisit this now the CI is working again?

MattSturgeon:
I think this is obsolete now, due to other refactoring to the test suite.

PR #1896, opened 2024-07-19
docs: Add a section of *Package options

### LINKED ISSUE #1282: Don't expose gcc on the PATH after starting nvim
The plugin `rust-analyzer` adds `gcc` to the `$PATH`.
If we open a terminal inside neovim, and run `cargo build`/`cargo test` etc, it will fail if the project depends on various dynamically linked libraries (e.g. if using `diesel`, which links dynamically the C header files) because obviously the headers of those libs are outside `/nix/store`.

Fixes I've tried and worked:
- manually remove the nix's gcc from the `$PATH` (but you have to do this in every terminal instance you open in neovim)
- write `build.rs` to inform the compiler and linker where various system libs are installed

Fix I'm proposing:
- don't expose the gcc from nix on the $PATH when starting neovim, if possible

I realise that gcc gets added to the `$PATH` by `rust-analyzer`, not by `nixvim`, but can still be removed by `nixvim` afterwards, unless there's a concern that doing so may break `rust-analyzer`.

traxys:
Documents the fix for #1282

Closes #1282

### DIFF
--- a/docs/user-guide/faq.md
+++ b/docs/user-guide/faq.md
@@ -87,3 +87,11 @@ keymaps = [
 
 [`map`]: https://nixos.org/manual/nix/stable/language/builtins#builtins-map
 [`keymaps`]: ../keymaps
+
+## How to use system provided binaries instead of nixvim provided ones
+
+There are a number of plugins that install extra packages using `nix`, but this can cause issues.
+For example enabling `plugins.treesitter` could add `gcc` to the PATH of neovim, and this could break workflows that rely on the system provided compiler.
+
+Most plugins that install packages also provide a `xxxPackage` option that can be set to `null` to skip the installation of the package.
+For example `plugin.treesitter` provides the `gccPackage` option.


docs/user-guide/faq.md
```diff
@@ -87,3 +87,11 @@ keymaps = [
...
 [`map`]: https://nixos.org/manual/nix/stable/language/builtins#builtins-map
 [`keymaps`]: ../keymaps
+
+## How do I avoid using `nix` to install dependent packages
+
+There are a number of plugins that install extra packages using `nix`, but this can cause issues.
+For example enabling `nvim-treesitter` could add `gcc` to the PATH of neovim, and this could break workflows that rely on the system provided compiler.
+
+Most plugins that install packages also provide a `xxxPackage` option that can be set to `null` to skip the installation of the package.
+For example `nvim-treesitter` provides the `gccPackage` option.
```
MattSturgeon, line 97:
```suggestion
For example `plugins.treesitter` provides the `gccPackage` option.
```

docs/user-guide/faq.md
```diff
@@ -87,3 +87,11 @@ keymaps = [
 
 [`map`]: https://nixos.org/manual/nix/stable/language/builtins#builtins-map
 [`keymaps`]: ../keymaps
+
+## How do I avoid using `nix` to install dependent packages
+
+There are a number of plugins that install extra packages using `nix`, but this can cause issues.
+For example enabling `nvim-treesitter` could add `gcc` to the PATH of neovim, and this could break workflows that rely on the system provided compiler.
```
MattSturgeon, line 94: Should this just be <code>`treesitter`</code> to match our plugin name?

docs/user-guide/faq.md
```diff
@@ -87,3 +87,11 @@ keymaps = [
 
 [`map`]: https://nixos.org/manual/nix/stable/language/builtins#builtins-map
 [`keymaps`]: ../keymaps
+
+## How do I avoid using `nix` to install dependent packages
```
MattSturgeon, line 91: Maybe this title should be more focused on the problem than the solution? Not sure...

Something in the spirit of "How to use system-provided binaries instead of packages installed by nixvim" but less verbose and overly technical?

I'm happy with the current title if we can't think of anything immediately though.

MattSturgeon (commented):
Good idea!

MattSturgeon (approved):
Great, thanks!

PR #1897, opened 2024-07-19
docs: Add an option search to our documentation

traxys:
This option search is availaible in the `/search/` sub path.
It's provided by [NuschtOS/search](https://github.com/NuschtOS/search).

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -1,5 +1,6 @@
 {
   pkgs ? import <nixpkgs> { config.enableUnfree = true; },
+  nuschtosSearch,
 }:
 let
   # Extend nixpkg's lib, so that we can handle recursive leaf types such as `either`
@@ -71,8 +72,7 @@ let
   hmOptions = builtins.removeAttrs (lib.evalModules {
     modules = [ (import ../wrappers/modules/hm.nix { inherit lib; }) ];
   }).options [ "_module" ];
-in
-rec {
+
   options-json =
     (pkgsDoc.nixosOptionsDoc {
       inherit
@@ -88,17 +88,38 @@ rec {
       inherit transformOptions;
       warningsAreErrors = false;
     }).optionsJSON;
+
+in
+{
+  inherit options-json;
+
   man-docs = pkgsDoc.callPackage ./man { inherit options-json; };
 }
-# Do not check if documentation builds fine on darwin as it fails:
-# > sandbox-exec: pattern serialization length 69298 exceeds maximum (65535)
-// lib.optionalAttrs (!pkgsDoc.stdenv.isDarwin) {
-  docs = pkgsDoc.callPackage ./mdbook {
-    inherit
-      helpers
-      modules
-      hmOptions
-      transformOptions
-      ;
-  };
-}
+// lib.optionalAttrs (!pkgsDoc.stdenv.isDarwin) (
+  let
+    mkSearch =
+      baseHref:
+      nuschtosSearch.packages.mkSearch {
+        optionsJSON = options-json + "/share/doc/nixos/options.json";
+        urlPrefix = "https://github.com/nix-community/nixvim/tree/main";
+        inherit baseHref;
+      };
+  in
+  {
+    # NuschtOS/search does not seem to work on darwin
+    search = mkSearch "/";
+
+    # Do not check if documentation builds fine on darwin as it fails:
+    # > sandbox-exec: pattern serialization length 69298 exceeds maximum (65535)
+    docs = pkgsDoc.callPackage ./mdbook {
+      inherit
+        helpers
+        modules
+        hmOptions
+        transformOptions
+        ;
+      # TODO: Find how to handle stable when 24.11 lands
+      search = mkSearch "/search/";
+    };
+  }
+)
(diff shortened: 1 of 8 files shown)


flake.nix
```diff
@@ -33,6 +33,13 @@
       ```
     */
 
+    # We need this until NixOS/nixpkgs#327670 lands in nixos-unstable & we update the lockfile
+    nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    nuschtosSearch = {
+      url = "github:NuschtOS/search";
+      inputs.nixpkgs.follows = "nixpkgs-unstable";
+    };
```
MattSturgeon, line 41: How long would the wait? Is it worth delaying this PR?

traxys: Not very long I think, a few days

MattSturgeon: Unless it gets delayed, let's hold off so we don't have to add additional inputs and additional arguments to the docs package.

docs/default.nix
```diff
@@ -1,5 +1,6 @@
 {
   pkgs ? import <nixpkgs> { config.enableUnfree = true; },
+  nuschtosSearch,
```
MattSturgeon, line 3: We can get this from pkgs now right? Or are we still waiting for https://github.com/NixOS/nixpkgs/pull/327670 to reach us?

docs/default.nix
```diff
@@ -88,7 +88,19 @@ rec {
...
+      optionsJSON = options-json + "/share/doc/nixos/options.json";
+      urlPrefix = "https://github.com/nix-community/nixvim/tree/main";
+      inherit baseHref;
+    };
+in
+{
   man-docs = pkgsDoc.callPackage ./man { inherit options-json; };
+
+  search = mkSearch "/";
 }
```
MattSturgeon, line 104: Think we need to inherit `options-json` if that was exported before?

flake-modules/dev/devshell.nix
```diff
@@ -81,10 +81,9 @@
 
                 echo -e "\n=> Documentation successfully built ('$doc_derivation')"
 
-                port=8000
-                echo -e "\n=> Now open your browser and navigate to 'localhost:$port'\n"
+                echo -e "\n=> You can then open your browser to view the doc\n"
 
-                ${pkgs.lib.getExe pkgs.python3} -m http.server -d "$doc_derivation"/share/doc
+                (cd "$doc_derivation"/share/doc && ${pkgs.lib.getExe pkgs.python3} ${./server.py})
```
MattSturgeon, line 86: Is the `()` subshell important here?

docs/default.nix
traxys: I don't think it has been packaged for nixos, I think it's meant to be consumed as a flake

flake-modules/dev/server.py
MattSturgeon, line 1: I wonder if there's a package in nixpkgs that does this without needing a custom script? Not important, probably easier to maintain this than hunt for a suitable solution

flake-modules/dev/devshell.nix
traxys: I always use a subshell whenever I have a `cd`. I think it's not _required_ here, but I prefer globally changing the directory in a script

flake.nix
```diff
@@ -33,6 +33,11 @@
       ```
     */
 
+    nuschtosSearch = {
+      url = "github:NuschtOS/search";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
```
MattSturgeon, line 39: As per other/prior feedback, can we get this from nixpkgs instead of a flake input?

traxys: As mentionned, it does not have a nixpkgs package

flake.nix
MattSturgeon: Oh, I missed that the comment was referencing needing a newer nixpkgs lock _not_ needing the nuschtos flake input. I think that invalidates some of my other feedback.

MattSturgeon:
CI:

- [Issues building `search` on darwin](https://buildbot.nix-community.org/#/builders/665/builds/2)
- Formatting issues (python)

docs/default.nix
traxys: good catch

traxys:
>     * [Issues building `search` on darwin](https://buildbot.nix-community.org/#/builders/665/builds/2)

Removed from darwin, as we _already_ don't build the docs for darwin

flake-modules/dev/devshell.nix
MattSturgeon: I don't think the `cd` can leak out of the script anyway, but no worries.

It may be simpler to pass `"$doc_derivation"/share/doc` in to the `server.py` script rather than implicitly using the current working directory, though IDK and it's not really important.

docs/default.nix
MattSturgeon: My mistake, I misinterpreted this thread: https://github.com/nix-community/nixvim/pull/1897#discussion_r1690408693

MattSturgeon (approved):
LGTM! Thanks

traxys:
We hit another problem with our 1100 checks, `twisted.internet.defer.FirstError: FirstError[#0, [Failure instance: Traceback: <class 'sqlalchemy.exc.OperationalError'>: (psycopg2.OperationalError) could not look up local user ID 995: Too many open files`

traxys:
@mergifyio queue

PR #1898, opened 2024-07-19
tests: Remove the test link farm

traxys:


MattSturgeon:
We'll have to update the `tests` dev command; maybe we could use [`gum choose --no-limit`](https://github.com/charmbracelet/gum?tab=readme-ov-file#choose) (and maybe [`--deliminator`](https://github.com/charmbracelet/gum/pull/598)) to allow users to select arbitrary tests to be run?

This could be another PR; this PR could just disable the `tests` command for now.

traxys:
Yeah I tried this PR to see what it would do, but I think we ought to wait for nix-community/buildbot-nix#217 to reduce the clutter a bit

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,69 @@
...
+	-h | --help)
+		help
+		exit 0
+		;;
+	-i | --interactive)
+		interactive=true
+		shift
+		;;
+	-t | --test)
+		test_name=$2
```
MattSturgeon, line 41: Can we add a TODO: allow specifying `--test` multiple times?

Or could we just use non-option arguments as test names?

 I.e. using the `*)` case, but after filtering out unrecognised `-*)` options.

MattSturgeon: This could be (TODO) compatible with `-i` by making the provided test-names on by default? Or by enabling them in addition to the interactively enabled tests?

MattSturgeon: (feel free to ignore me until this is undrafted)

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,69 @@
...
+	esac
+done
+
+if [[ $interactive = true && -n $test_name ]]; then
+	echo "Can't specify both --interactive and --test" >&2
+	exit 1
+fi
+
+test_list() {
+	nix flake show --json 2>/dev/null |
```
stasjok, line 54: I didn't try anything of this, but I'm sure that `nix flake show` will be very slow. Try this instead: `nix eval .#checks.x86_64-linux.tests.entries --apply builtins.attrNames --json`

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,69 @@
...
+		jq '.checks."x86_64-linux" | keys | map(select(startswith("test-")))[]' --raw-output
+}
+
+if [[ $interactive = false && -z $test_name ]]; then
+	test_list |
+		xargs -I{} sh -xc "${NIXVIM_NIX_COMMAND} build .#checks.${NIXVIM_SYSTEM}.{} || exit 255" || exit $?
```
stasjok, line 60: Also when specifying multiple `installables` to `nix build` with `.#` it evaluates them separately (loads a flake multiple time). It's a 10 second for each on my VM. It should be faster something like this:

```
build -f . checks.x86_64-linux.enable-except-in-tests checks.x86_64-linux.extend checks.x86_64-linux.extra-args-tests checks.x86_64-linux.extra-files checks.x86_64-linux.home-manager-module checks.x86_64-linux.lib-tests checks.x86_64-linux.maintainers checks.x86_64-linux.man-docs checks.x86_64-linux.nixos-module checks.x86_64-linux.no-flake
```

instead of

```
nix build .#checks.x86_64-linux.enable-except-in-tests .#checks.x86_64-linux.extend .#checks.x86_64-linux.extra-args-tests .#checks.x86_64-linux.extra-files .#checks.x86_64-linux.home-manager-module .#checks.x86_64-linux.lib-tests .#checks.x86_64-linux.maintainers .#checks.x86_64-linux.man-docs .#checks.x86_64-linux.nixos-module .#checks.x86_64-linux.no-flake
```

because in the first case it evaluates only one time, then it takes attrs of it. In the second case every `.#` evaluates a flake separately.

stasjok: Ah, I see that xargs here won't combine it to one command, so `nix build` is called serially. Maybe it can be ran in batches (need to find a balance between memory consumption and speed). Example:

```bash
test_list | xargs -n 10 sh -xc "nix build -f . \"\$@\" || exit 255" --
```

but prefix `checks.${NIXVIM_SYSTEM}` should be added in jq.

flake-modules/dev/launch-test.sh
traxys: Yup indeed nix flake show takes multiple seconds, thanks I'll try that

flake-modules/dev/devshell.nix
```diff
@@ -32,11 +32,30 @@
...
+                                jq
+                                fzf
+                              ]
+                            )
+                          }
+                      '';
+                in
+                ''
+                  NIXVIM_SYSTEM=${system} NIXVIM_NIX_COMMAND=${nix} ${launchTest}/bin/launch-test.sh "$@"
+                '';
```
GaetanLepage, line 58: Couldn't this be done with a `mkShellApplication` ? Not sure if it would be better, but at least you could provide the runtime deps.

flake-modules/dev/devshell.nix
```diff
@@ -32,11 +32,30 @@
...
+                              [
+                                jq
+                                fzf
+                              ]
+                            )
+                          }
+                      '';
+                in
+                ''
+                  NIXVIM_SYSTEM=${system} NIXVIM_NIX_COMMAND=${nix} ${launchTest}/bin/launch-test.sh "$@"
```
GaetanLepage, line 57:
```suggestion
                  NIXVIM_SYSTEM=${system} NIXVIM_NIX_COMMAND=${nix} ${lib.getExe launchTest} "$@"
```

flake-modules/dev/devshell.nix
```diff
@@ -32,11 +32,30 @@
...
               help = "Run nixvim tests";
-              command = ''
-                echo "=> Running nixvim tests for the '${system}' architecture..."
-
-                ${nix} build .#checks.${system}.tests "$@"
-              '';
+              command =
+                let
+                  launchTest =
+                    pkgs.runCommandLocal "launch-test.sh"
```
MattSturgeon, line 38: Does runCommandLocal scan `$out/bin/` to determine `meta.mainProgram` (needed for `lib.getExe`)? Some writers do this, but I think runCommand is too "low level" for that kinda magic?

flake-modules/dev/devshell.nix
```diff
@@ -32,11 +32,30 @@
...
+                        makeWrapper $script $out/bin/launch-test.sh \
+                          --prefix PATH : ${
+                            lib.makeBinPath (
+                              with pkgs;
+                              [
+                                jq
+                                fzf
+                              ]
+                            )
+                          }
```
MattSturgeon, line 53: Can't we use the env attrs for this?

e.g.

```nix
wrapper_path = with pkgs; [
  jq
  fzf
];
#...
''
  --prefix PATH : "$wrapper_path"
'';
```

flake-modules/dev/devshell.nix
traxys: Did not think about it TBH

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+help() {
+	cat <<EOF
+Usage: tests [OPTIONS] [tests...]
+
+If tests are passed on the command line only these will be launched
+
+Options:
+	-h/--help: Display this help message and exit
+	-l/--list: Display the list of tests and exit
+	-i/--interactive: Pick interactively the tests. Can't be supplied if tests where passed.
```
MattSturgeon, line 17: Better to use the `-h, --help:` format used by most CLI tools? Instead of `/`.

flake-modules/dev/devshell.nix
traxys: I need the mkShellApplication for that, but yes!

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+EOF
+}
+
+if ! OPTS=$(getopt -o "hli" -l "help,list,interactive" -- "$@"); then
+	echo "Invalid options" >&2
+	help
+	exit 1
+fi
```
MattSturgeon, line 25: This ensures only valid opts are passed?

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+		;;
+	-i | --interactive)
+		interactive=true
+		shift
+		;;
+	--)
+		shift
+		specified_tests=("$@")
+		break
+		;;
```
MattSturgeon, line 55: I think we also need to handle mixed opts+args using `*)`.

Anything handled with `*)` should be added to the `specified_tests` list, and then everything after `--)` should also be added (without overwriting).

flake-modules/dev/launch-test.sh
MattSturgeon, line 1: Not sure if it's intended, but this file is tab-indented while others are using spaces.

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+	esac
+done
+
+run_tests() {
+	# Add the prefix "checks.${system}." to each argument
+	if ! "${NIXVIM_NIX_COMMAND}" build -f . "${@/#/checks.${NIXVIM_SYSTEM}.}"; then
```
MattSturgeon, line 61: Can we use `--file` instead of `-f`? I prefer using long-form options in scripts since anyone can follow their meaning, even if unfamiliar with the specific command's options.

MattSturgeon: Also, does `nix build -f .` produce `result`? I'm wondering if we can speed up the generate-files build by doing all builds in one step?

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+		echo "Test failure" >&2
+		exit 1
+	fi
+}
+
+if [[ $interactive = true && ${#specified_tests[@]} -ne 0 ]]; then
+	echo "Can't use --interactive with tests on the command line" >&2
+	exit 1
+fi
```
MattSturgeon, line 70: Can we add <code># TODO: Decide if/how `--interactive` can be used alongside specified_tests?</code>

flake-modules/dev/launch-test.sh
```diff
@@ -0,0 +1,83 @@
...
+if [[ $interactive = true ]]; then
+	test_name=$(mk_test_list | fzf) || exit $?
+	specified_tests+=("$test_name")
+fi
+
+if [[ ${#specified_tests[@]} -eq 0 ]]; then
+	readarray -t complete_test_list < <(mk_test_list)
+	run_tests "${complete_test_list[@]}"
+else
+	echo "Running tests: ${specified_tests[*]}" >&2
```
MattSturgeon, line 81: Maybe also include the count? `Running ${#specified_tests[@]}: ${specified_tests[*]}`?

flake-modules/dev/launch-test.sh
MattSturgeon: @traxys is this resolved?

flake-modules/dev/launch-test.sh
MattSturgeon: @traxys is this resolved?

flake-modules/dev/devshell.nix
MattSturgeon: Agreed, would probably be better. Should still be able to keep the sh file separate.

flake-modules/dev/devshell.nix
MattSturgeon: This would be resolved by using `mkShellApplication` as per @GaetanLepage's feedback.

flake-modules/dev/devshell.nix
MattSturgeon: Also resolved using `mkShellApplication`

flake-modules/dev/launch-test.sh
MattSturgeon: To be fancy, we could make the `s` optional by doing something like:

```bash
s=$([ "${#specified_tests[@]}" -gt 1 ] && echo "s" || echo "")
```

flake-modules/dev/launch-test.sh
traxys: Yes

flake-modules/dev/launch-test.sh
traxys: No need to handle that, `getopt` already does it for us!

traxys: `getopt` normalizes the arguments in the form `[OPTIONS] -- [POSITIONAL ARGS]`

flake-modules/dev/launch-test.sh
traxys: I specifically don't want to allow --interactive alongside specied_tests, because I find that confusing

flake-modules/dev/launch-test.sh
traxys: I use `shfmt` to format shell scripts that defaults to tabs

flake-modules/dev/launch-test.sh
traxys: It does for each of the derivations passed to the command line.
I don't see how can be do all the builds in "one" step, I already pass all the builds to a single `nix` invocation :/

flake-modules/dev/launch-test.sh
traxys: Not sure I want to be that fancy :P

flake-modules/dev/launch-test.sh
traxys: Yup

flake-modules/dev/launch-test.sh
traxys: Yes

flake-modules/dev/launch-test.sh
MattSturgeon: Sure 👍

flake-modules/dev/launch-test.sh
MattSturgeon: Ok, wasn't sure if it was intended or not. Maybe we can add shfmt to our treefmt in a follow up PR?

flake-modules/dev/launch-test.sh
MattSturgeon: Sorry, which comment are you replying to?

MattSturgeon (approved):
No issues from me assuming everything works and tests pass etc

flake-modules/dev/launch-test.sh
traxys: > I'm wondering if we can speed up the generate-files build by doing all builds in one step?

flake-modules/dev/launch-test.sh
traxys: I added it in this PR, as it is one line

PR #1907, opened 2024-07-21
modules/lsp/servers/vls: do not add filetype extension if not enabled

stasjok:
Option `plugins.lsp.servers.vls.autoSetFiletype` adds filetype extension by default, even when `plugins.lsp.servers.vls.enable = false`. I think this is not intended behavior.

My main goal is to make `files` submodules to produce empty configs by default. When I do

```nix
files."ftplugin/nix.lua".localOpts.shiftwidth = 2;
```

I expect `ftplugin/nix.lua` to have only this option, and nothing else.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -591,7 +591,9 @@ let
           example = false;
         };
       };
-      extraConfig = cfg: { filetype.extension = mkIf cfg.autoSetFiletype { v = "vlang"; }; };
+      extraConfig = cfg: {
+        filetype.extension = mkIf (cfg.enable && cfg.autoSetFiletype) { v = "vlang"; };
+      };
     }
     {
       name = "vuels";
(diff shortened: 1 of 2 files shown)


traxys (approved):

traxys:
(You may want to rebase on main for the commit attribution)

plugins/lsp/language-servers/default.nix
```diff
@@ -591,7 +591,9 @@ let
           example = false;
         };
       };
-      extraConfig = cfg: { filetype.extension = mkIf cfg.autoSetFiletype { v = "vlang"; }; };
+      extraConfig = cfg: {
+        filetype.extension = mkIf (cfg.enable && cfg.autoSetFiletype) { v = "vlang"; };
```
MattSturgeon, line 595: I wonder if the issue here is actually `_mk-lsp.nix` not having `mkIf cfg.enable` like the other mkPlugins do in our helper lib?

PR #1908, opened 2024-07-21
modules/filetype: ensure that modules does not provide empty settings

stasjok:
`helpers.mkIfNonNull'` checks only for value = null, but I noticed that when there is anywhere something like `mkIf false config.filename = {}` (like with vlang filetype currently in nixvim, see also #1907), value of `config.filetype` is not null, but has all attrs with null values. It can be debugged with `nix repl`:

```
nix-repl> c = makeNixvim {}

nix-repl> :p c.options.filetype.definitions
[ { extension = { _type = "if"; condition = false; content = { v = "vlang"; }; }; } ]

nix-repl> :p c.config.filetype
{ extension = null; filename = null; pattern = null; }
```

Notice that `condition = false`, but the value is not null.

This PR changes mkIf condition to account for that.

NOTE: Test `checks.x86_64-linux.tests.entries.modules-filetypes` will fail until #1907 is merged.

### DIFF
--- a/modules/filetype.nix
+++ b/modules/filetype.nix
@@ -6,6 +6,8 @@
 }:
 with lib;
 let
+  cfg = config.filetype;
+
   filetypeDefinition = helpers.mkNullOrOption (
     with types;
     attrsOf (oneOf [
@@ -52,7 +54,9 @@ in
         pattern = filetypeDefinition "set filetypes matching the specified pattern";
       };
 
-  config.extraConfigLua = helpers.mkIfNonNull' config.filetype ''
-    vim.filetype.add(${helpers.toLuaObject config.filetype})
-  '';
+  config.extraConfigLua =
+    lib.mkIf (cfg != null && (builtins.any (v: v != null) (builtins.attrValues cfg)))
+      ''
+        vim.filetype.add(${helpers.toLuaObject cfg})
+      '';
 }
--- a/tests/test-sources/modules/filetypes.nix
+++ b/tests/test-sources/modules/filetypes.nix
@@ -38,4 +38,21 @@
       };
     };
   };
+
+  default-empty.module =
+    { config, ... }:
+    {
+      files.test = { };
+
+      assertions = [
+        {
+          assertion = builtins.match ".*vim\.filetype\..*" config.content == null;
+          message = "No vim.filetype definitions should be present in init.lua by default.";
+        }
+        {
+          assertion = builtins.match ".*vim\.filetype\..*" config.files.test.content == null;
+          message = "No vim.filetype definitions should be present in files submodules by default.";
+        }
+      ];
+    };
 }


MattSturgeon:
> noticed that when there is anywhere something like `mkIf false config.filename = {}`, value of config.filetype is not null, but has all attrs with null values.

This is probably related to how the module system handles things; e.g. if `filetype` contains sub-options those options will _always_ appear in the final merged `config`.

For our use-case where we often want to print entire attrs as lua this is annoying and is why `toLuaObject` filters out empty attrs by default.

modules/filetype.nix
```diff
@@ -52,7 +54,9 @@ in
         pattern = filetypeDefinition "set filetypes matching the specified pattern";
       };
 
-  config.extraConfigLua = helpers.mkIfNonNull' config.filetype ''
-    vim.filetype.add(${helpers.toLuaObject config.filetype})
-  '';
+  config.extraConfigLua =
+    lib.mkIf (cfg != null && (cfg.extension != null || cfg.filename != null || cfg.pattern != null))
```
MattSturgeon, line 58: This is ugly, but it is probably the right thing to do.

MattSturgeon: Would it make sense to do something like `all (v: v == null) (attrValues config fileType)`?

I think even if users had freeform definitions, `toLuaObject` would still ignore nulls.

Might need to also check `config.filetype ? __empty` though, in case the user explicitly wants to define an empty filetype?

(Is the option even freeform? If not some of these caveats don't apply. Sorry for the poor review, I'm on my phone atm!)

stasjok: I think it warrants a library function similar to `mkIfNonNull` and `mkIfNonNull'` that test attrs similar to `toLuaObject`. If `toLuaObject` would produce something empty (it can be configured via function params what is considered empty: only if nil, or if empty table, or if empty string), condition is false. I don't think I know all this options and combinations with __empty, __raw etc to write one though.

MattSturgeon:
```suggestion
    lib.mkIf (cfg != null && (lib.any (v: v != null) (lib.attrValues cfg)))
```

I've had a closer look, and it should be fine here to check whether any attrs are non-null using the builtins [`any`](https://noogle.dev/f/builtins/any) and [`attrValues`](https://noogle.dev/f/builtins/attrValues).

Since this isn't a freeform option we don't have to worry about unexpected attrs or `__empty`.

stasjok: Thanks, that's nicer.

MattSturgeon (approved):
Looks good, thanks!

tests/test-sources/modules/filetypes.nix
```diff
@@ -38,4 +38,21 @@
...
   };
+
+  default-empty.module =
+    { config, ... }:
+    {
+      files.test = { };
+
+      assertions = [
+        {
+          assertion = builtins.match ".*vim\.filetype\..*" config.content == null;
```
MattSturgeon, line 49: BTW you can use [`lib.hasInfix`](https://noogle.dev/f/lib/hasInfix) when you don't actually need a regex:

```suggestion
          assertion = lib.hasInfix "vim.filetype." config.content;
```

stasjok: Right. I already found out about it yesterday when working on modules/output refactoring. I intended to change it in other PRs, but looks like I forgot about this.

PR #1909, opened 2024-07-21
modules/output: produce empty files by default

stasjok:
This PR removes empty

```lua
vim.cmd([[

]])
```

block from config.

I know that this is already handled by #1889, but I've added tests to make sure that this is the behavior. Also I added the test to make sure that all extraConfigs are outputted.

This also changes various places in other modules with the intention of not setting `extraConfig*` options to empty strings (more specifically whitespace only strings).

NOTE: The tests will fail until #1906, #1907 and #1908 are merged.

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -110,11 +110,12 @@ with lib;
           {
             extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
             inherit extraPackages;
-
-            ${extraConfigNamespace} = optionalString callSetup ''
+          }
+          (optionalAttrs callSetup {
+            ${extraConfigNamespace} = ''
               require('${luaName}')${setup}(${optionalString (cfg ? settings) (toLuaObject cfg.settings)})
             '';
-          }
+          })
           (optionalAttrs (isColorscheme && (colorscheme != null)) { colorscheme = mkDefault colorscheme; })
           (extraConfig cfg)
         ]);
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -4,7 +4,14 @@
   _nixvimTests,
 }:
 with lib;
-{
+rec {
+  # Whether a string contains something other than whitespaces
+  hasContent = str: builtins.match "[[:space:]]*" str == null;
+
+  # Concatenate a list of strings, adding a newline at the end of each one,
+  # but skipping strings containing only whitespace characters
+  concatNonEmptyLines = lines: concatLines (builtins.filter hasContent lines);
+
   listToUnkeyedAttrs =
     list:
     builtins.listToAttrs (lib.lists.imap0 (idx: lib.nameValuePair "__unkeyed-${toString idx}") list);
@@ -118,4 +125,26 @@ with lib;
       ${string}
     end
   '';
+
+  # Wrap Vimscript for using in lua,
+  # but only if the string contains something other than whitespaces
+  # TODO: account for a possible ']]' in the string
+  wrapVimscriptForLua =
+    string:
+    optionalString (hasContent string) ''
+      vim.cmd([[
+      ${string}
+      ]])
+    '';
+
+  # Wrap lua script for using in Vimscript,
+  # but only if the string contains something other than whitespaces
+  # TODO: account for a possible 'EOF' if the string
+  wrapLuaForVimscript =
+    string:
+    optionalString (hasContent string) ''
+      lua << EOF
+      ${string}
+      EOF
+    '';
 }
--- a/modules/autocmd.nix
+++ b/modules/autocmd.nix
@@ -42,7 +42,7 @@ with lib;
     let
       inherit (config) autoGroups autoCmd;
     in
-    mkIf (autoGroups != { } || autoCmd != { }) {
+    mkIf (autoGroups != { } || autoCmd != [ ]) {
       # Introduced early October 2023.
       # TODO remove in early December 2023.
       assertions = [
--- a/modules/highlights.nix
+++ b/modules/highlights.nix
@@ -35,50 +35,52 @@ with lib;
     };
   };
 
-  config = {
-    extraConfigLuaPre =
-      (optionalString (config.highlight != { })
-        # lua
-        ''
-          -- Highlight groups {{
-          do
-            local highlights = ${helpers.toLuaObject config.highlight}
+  config = mkMerge [
+    {
+      extraConfigLuaPre =
+        mkIf (config.highlight != { })
+          # lua
+          ''
+            -- Highlight groups {{
+            do
+              local highlights = ${helpers.toLuaObject config.highlight}
 
-            for k,v in pairs(highlights) do
-              vim.api.nvim_set_hl(0, k, v)
+              for k,v in pairs(highlights) do
+                vim.api.nvim_set_hl(0, k, v)
+              end
             end
-          end
-          -- }}
-        ''
-      )
-      + (optionalString (config.match != { })
-        # lua
-        ''
-          -- Match groups {{
-          do
-            local match = ${helpers.toLuaObject config.match}
-
-            for k,v in pairs(match) do
-              vim.fn.matchadd(k, v)
-            end
-          end
             -- }}
-        ''
-      );
+          '';
+      extraConfigLuaPost =
+        mkIf (config.highlightOverride != { })
+          # lua
+          ''
+            -- Highlight groups {{
+            do
+              local highlights = ${helpers.toLuaObject config.highlightOverride}
 
-    extraConfigLuaPost =
-      optionalString (config.highlightOverride != { })
-        # lua
-        ''
-          -- Highlight groups {{
-          do
-            local highlights = ${helpers.toLuaObject config.highlightOverride}
-
-            for k,v in pairs(highlights) do
-              vim.api.nvim_set_hl(0, k, v)
+              for k,v in pairs(highlights) do
+                vim.api.nvim_set_hl(0, k, v)
+              end
             end
-          end
-          -- }}
-        '';
-  };
+            -- }}
+          '';
+    }
+    {
+      extraConfigLuaPre =
+        mkIf (config.match != { })
+          # lua
+          ''
+            -- Match groups {{
+            do
+              local match = ${helpers.toLuaObject config.match}
+
+              for k,v in pairs(match) do
+                vim.fn.matchadd(k, v)
+              end
+            end
+              -- }}
+          '';
+    }
+  ];
 }
--- a/modules/opts.nix
+++ b/modules/opts.nix
@@ -55,31 +55,35 @@ in
   };
 
   config = {
-    extraConfigLuaPre = concatLines (
-      mapAttrsToList (
-        optionName:
-        {
-          prettyName,
-          luaVariableName,
-          luaApi,
-          ...
-        }:
-        let
-          varName = "nixvim_${luaVariableName}";
-          optionDefinitions = config.${optionName};
-        in
-        optionalString (optionDefinitions != { }) ''
-          -- Set up ${prettyName} {{{
-          do
-            local ${varName} = ${helpers.toLuaObject optionDefinitions}
+    extraConfigLuaPre =
+      let
+        content = helpers.concatNonEmptyLines (
+          mapAttrsToList (
+            optionName:
+            {
+              prettyName,
+              luaVariableName,
+              luaApi,
+              ...
+            }:
+            let
+              varName = "nixvim_${luaVariableName}";
+              optionDefinitions = config.${optionName};
+            in
+            optionalString (optionDefinitions != { }) ''
+              -- Set up ${prettyName} {{{
+              do
+                local ${varName} = ${helpers.toLuaObject optionDefinitions}
 
-            for k,v in pairs(${varName}) do
-              vim.${luaApi}[k] = v
-            end
-          end
-          -- }}}
-        ''
-      ) optionsAttrs
-    );
+                for k,v in pairs(${varName}) do
+                  vim.${luaApi}[k] = v
+                end
+              end
+              -- }}}
+            ''
+          ) optionsAttrs
+        );
+      in
+      mkIf (content != "") content;
   };
 }
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -107,16 +107,10 @@ with lib;
         }
       );
 
-      customRC =
-        let
-          hasContent = str: (builtins.match "[[:space:]]*" str) == null;
-        in
-        (optionalString (hasContent neovimConfig.neovimRcContent) ''
-          vim.cmd([[
-            ${neovimConfig.neovimRcContent}
-          ]])
-        '')
-        + config.content;
+      customRC = helpers.concatNonEmptyLines [
+        (helpers.wrapVimscriptForLua neovimConfig.neovimRcContent)
+        config.content
+      ];
 
       init = helpers.writeLua "init.lua" customRC;
 
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -140,7 +140,7 @@ with lib;
         '';
       };
 
-      extraConfigLuaPre = lib.optionalString config.wrapRc ''
+      extraConfigLuaPre = lib.mkIf config.wrapRc ''
         -- Ignore the user lua configuration
         vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
         vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -4,4 +4,90 @@
     # Make sure jsregexp is in LUA_PATH
     extraConfigLua = ''require("jsregexp")'';
   };
+
+  # Test that all extraConfigs are present in output
+  all-configs.module =
+    {
+      config,
+      pkgs,
+      lib,
+      ...
+    }:
+    let
+      configs = {
+        extraConfigLuaPre = "string.format('extraConfigLuaPre1')";
+        extraConfigLua = "string.format('extraConfigLua2')";
+        extraConfigLuaPost = "string.format('extraConfigLuaPost3')";
+        extraConfigVim = "let g:var = 'extraConfigVim4'";
+      };
+      mkConfigAssertions = name: value: [
+        {
+          assertion = lib.hasInfix "extraConfigLuaPre1" value;
+          message = "Configuration file ${name} does not contain extraConfigLuaPre.";
+        }
+        {
+          assertion = lib.hasInfix "extraConfigLua2" value;
+          message = "Configuration file ${name} does not contain extraConfigLua.";
+        }
+        {
+          assertion = lib.hasInfix "extraConfigLuaPost3" value;
+          message = "Configuration file ${name} does not contain extraConfigLuaPost.";
+        }
+        {
+          assertion = lib.hasInfix "extraConfigVim4" value;
+          message = "Configuration file ${name} does not contain extraConfigVim.";
+        }
+      ];
+    in
+    configs
+    // {
+      files = {
+        "test.lua" = configs;
+        "test.vim" = configs;
+      };
+
+      # Plugin configs
+      extraPlugins = [
+        {
+          plugin = pkgs.emptyDirectory;
+          config = "let g:var = 'neovimRcContent5'";
+        }
+      ];
+
+      assertions =
+        mkConfigAssertions "init.lua" config.content
+        ++ mkConfigAssertions "test.lua" config.files."test.lua".content
+        ++ mkConfigAssertions "test.vim" config.files."test.vim".content
+        # Check the final generated init.lua too
+        ++ mkConfigAssertions "initPath" (builtins.readFile config.initPath)
+        ++ [
+          # Only init.lua contains configuration from plugin definitions
+          {
+            assertion = lib.hasInfix "neovimRcContent5" (builtins.readFile config.initPath);
+            message = "Configuration file init.lua does not contain plugin configs";
+          }
+        ];
+    };
+
+  files-default-empty.module =
+    { config, helpers, ... }:
+    {
+      files = {
+        # lua type
+        "test.lua" = { };
+        # vim type
+        "test.vim" = { };
+      };
+
+      assertions = [
+        {
+          assertion = !helpers.hasContent config.files."test.lua".content;
+          message = "Default content of test.lua file is expected to be empty.";
+        }
+        {
+          assertion = !helpers.hasContent config.files."test.vim".content;
+          message = "Default content of test.vim file is expected to be empty.";
+        }
+      ];
+    };
 }
--- a/modules/diagnostics.nix
+++ b/modules/diagnostics.nix
@@ -19,7 +19,7 @@ with lib;
   };
 
   config = {
-    extraConfigLuaPre = optionalString (config.diagnostics != { }) ''
+    extraConfigLuaPre = mkIf (config.diagnostics != { }) ''
       vim.diagnostic.config(${helpers.toLuaObject config.diagnostics})
     '';
   };
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -90,7 +90,7 @@ with lib;
           ${concatStringsSep "\n" luaDefs}
         '';
 
-      extraConfigLua = optionalString (config.keymaps != [ ]) ''
+      extraConfigLua = mkIf (config.keymaps != [ ]) ''
         -- Set up keybinds {{{
         do
           local __nixvim_binds = ${helpers.toLuaObject (map normalizeMapping config.keymaps)}
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -1,4 +1,9 @@
-{ lib, config, ... }:
+{
+  lib,
+  config,
+  helpers,
+  ...
+}:
 with lib;
 let
   pluginWithConfigType = types.submodule {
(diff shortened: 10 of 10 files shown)


MattSturgeon:
Thanks! I'd already looked at doing this in https://github.com/nix-community/nixvim/pull/1889, however since that one is larger in scope and unfinished it's better to have this merged sooner than later!

It's also good to have tests, which I hadn't implemented in https://github.com/nix-community/nixvim/pull/1889.

Would you mind looking over how I'd done it in that other PR to see if there's any elements you'd like to incorporate in this one?

E.g. it probably makes sense to move `hasContent` into our helpers lib and maybe come up with some kinda wrapper function to reduce boilerplate when wrapping potentially empty lua/vim strings?

Not too important as those things may expand the size/scope of this PR unnecessarily making it harder to review. But worth a quick look.

stasjok:
> Would you mind looking over how I'd done it in that other PR to see if there's any elements you'd like to incorporate in this one?

Your implementation is better, because it doesn't output empty lines. I like it more. But I intentionally didn't rip it off, because I wanted something simple to get started (and functional) + tests. And since you already implemented a better version, I thought that you'll easily rebase to it later. But if you think that it's better to include your version here, I probably can extract it from there.

MattSturgeon:
> But if you think that it's better to include your version here, I probably can extract it from there.

I don't mind too much either way. There's certainly value in keeping this PR simple as you said.

lib/neovim-plugin.nix
```diff
@@ -111,7 +111,7 @@ with lib;
             extraPlugins = (optional installPackage cfg.package) ++ extraPlugins;
             inherit extraPackages;
 
-            ${extraConfigNamespace} = optionalString callSetup ''
+            ${extraConfigNamespace} = mkIf callSetup ''
```
MattSturgeon, line 114: This is intentionally not using mkIf

modules/autocmd.nix
```diff
@@ -42,7 +42,7 @@ with lib;
     let
       inherit (config) autoGroups autoCmd;
     in
-    mkIf (autoGroups != { } || autoCmd != { }) {
+    mkIf (autoGroups != { } || autoCmd != [ ]) {
```
MattSturgeon, line 45: Unrelated fix

modules/highlights.nix
```diff
@@ -35,50 +35,52 @@ with lib;
...
-      (optionalString (config.highlight != { })
-        # lua
-        ''
-          -- Highlight groups {{
-          do
-            local highlights = ${helpers.toLuaObject config.highlight}
+  config = mkMerge [
+    {
+      extraConfigLuaPre =
+        mkIf (config.highlight != { })
```
MattSturgeon, line 41: Is using `mkIf` really so important that we must also use it when concatenating strings?

IMO concatenating optional strings is fine here and in similar cases, @traxys what do you think?

modules/opts.nix
```diff
@@ -55,31 +55,35 @@ in
...
-          varName = "nixvim_${luaVariableName}";
-          optionDefinitions = config.${optionName};
-        in
-        optionalString (optionDefinitions != { }) ''
-          -- Set up ${prettyName} {{{
-          do
-            local ${varName} = ${helpers.toLuaObject optionDefinitions}
+    extraConfigLuaPre =
+      let
+        content = helpers.concatNonEmptyStringsNewline (
```
MattSturgeon, line 60: Is this a new helper? `concatNonEmptyLines` would be more concise.

modules/top-level/output.nix
```diff
@@ -140,7 +140,7 @@ with lib;
         '';
       };
 
-      extraConfigLuaPre = lib.optionalString config.wrapRc ''
+      extraConfigLuaPre = lib.mkIf config.wrapRc ''
```
MattSturgeon, line 143: `extraPlugins` lower down in this file can also be a `mkIf`

MattSturgeon (changes requested):
(Just reviewing the treewide mkIf commit)

- This would be better off in a dedicated PR, IMO.
  I don't see how it relates to this one?
- There's a typo in the commit message (`treeside` -> `treewide`)
- Some stuff is either deliberately not using mkIf or is less readable, especially string concatenation.

stasjok:
> I don't mind too much either way. There's certainly value in keeping this PR simple as you said.

I rewrote it completely. See if it's OK.

Also I replaced a couple of global instances of `optionalString` with `mkIf`. It's in a separate commit, I can drop it if needed.

Before:

```
nix-repl> c.options.extraConfigLuaPre.definitions      
[ "\n\n\n\n" "" "" "-- Ignore the user lua configuration\nvim.opt.runtimepath:remove(vim.fn.stdpath('config'))         
     -- ~/.config/nvim\nvim.opt.runtimepath:remove(vim.fn.stdpath('config') .. \"/after\")  -- ~/.config/nvim/after\nvi
m.opt.runtimepath:remove(vim.fn.stdpath('data') .. \"/site\")     -- ~/.local/share/nvim/site\n" ]

nix-repl> c.options.extraConfigLuaPost.definitions
[ "" "" ]
```

After:

```
nix-repl> c.options.extraConfigLuaPre.definitions  
[ "-- Ignore the user lua configuration\nvim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config
/nvim\nvim.opt.runtimepath:remove(vim.fn.stdpath('config') .. \"/after\")  -- ~/.config/nvim/after\nvim.opt.runtimepath
:remove(vim.fn.stdpath('data') .. \"/site\")     -- ~/.local/share/nvim/site\n" ]

nix-repl> c.options.extraConfigLuaPost.definitions
[ "" ]
```

Notice that there is no empty definitions by default.

But similar cases with empty definitions are still present in some of the plugins, although they are guarded by `mkIf cfg.enable`, if plugin is enabled, they appear. But I definitely not going to edit them now.

modules/highlights.nix
stasjok: There was two way to do it:

1) Concat strings in a variable, then in mkIf check if it empty.
2) Separate with mkMerge.

I can't decide what's best, I though that this way looks better than when two options are concated with variables and third as is. This way all three options looks the similar.

lib/neovim-plugin.nix
stasjok: Why? If `callSetup` is false, plugin produces empty an string value. Later on all this empty strings are concatenated with newlines. If it can be fixed in other places, why not here?

modules/autocmd.nix
stasjok: All this fixes was to remove empty extra empty lines from output. This fixes a condition that is guarding `extraConfigLuaPost` from producing empty string.

tests/test-sources/modules/output.nix
```diff
@@ -1,7 +1,91 @@
...
+  all-configs.module =
+    { config, ... }:
+    let
+      configs = {
+        extraConfigLuaPre = "string.format('extraConfigLuaPre1')";
+        extraConfigLua = "string.format('extraConfigLua2')";
+        extraConfigLuaPost = "string.format('extraConfigLuaPost3')";
+        extraConfigVim = "let g:var = 'extraConfigVim4'";
+      };
+      configAssertion = name: config: [
```
MattSturgeon, line 22: I think `name: value:` or `n: v:` is better, to avoid confusion with the overloaded `config` from the module args.

MattSturgeon: Also lets name this something like `mkAssertion` or `assertFileContent` or `mkFileAssertion` (e.g.)

modules/top-level/output.nix
stasjok: OK, I can change it too, but this PR is about cleaning up generated configs (mostly removing extra empty lines).

MattSturgeon (changes requested):
(This review is for the other (non-mkIf) changes)

Looks good, mostly just minor nits!

stasjok:
I'll check comments tomorrow. Please decide if a second commit removing empty lines from outputs in a various places is wanted or not.

modules/highlights.nix
MattSturgeon: Ah I see, so you do this because all parts of the string are optional, so you're avoiding an empty definition?

That makes more sense. Does that apply to all changes though?

modules/autocmd.nix
MattSturgeon: Got it.

It is unrelated to the "use mkIf instead of optionalString" commit, but not unrelated to this PR if it solves the bigger issue.

lib/neovim-plugin.nix
MattSturgeon: As a general rule, these general use helpers don't know _why_ a feature such as `callSetup` has been disabled, it may be because the caller doesn't want something that feature uses to be evaluated (which mkIf won't solve).

To clarify, `mkIf` still evaluates the content, even when the condition is false.

In this specific case, we'd be calling `toLuaObject cfg.settings` unnecessarily, although maybe that's not a big deal.

But here I think it is better to use `optional`, `optionals`, `optionalAttrs`, `optionalString` or `if then else` as appropriate.

modules/top-level/output.nix
MattSturgeon: When I read "treewide: optionalString -> mkIf" I didn't understand the motivation for the change and just assumed you were doing general cleanup.

MattSturgeon:
> This would be better off in a dedicated PR, IMO.
> I don't see how it relates to this one?

After reading your replies I agree it makes sense in this PR, but it would benefit from a commit message that explains the motivation behind the change, as I was clearly confused 😉

lib/neovim-plugin.nix
stasjok: > To clarify, mkIf still evaluates the content, even when the condition is false.

OK. Got it. Since `mkMerge` is already used here, what do you thin about:

```nix
(optionalAttrs callSetup {
   ${extraConfigNamespace} = ''
    '';
})
```

I think it preserve current behavior and avoid empty string.

MattSturgeon: Agreed, I think that's the right approach 😃

lib/utils.nix
```diff
@@ -4,7 +4,15 @@
...
 }:
 with lib;
-{
+rec {
+  # Whether a string contains something other than whitespace
+  hasContent = str: builtins.match "[[:space:]]*" str == null;
+
+  # Concatenate a list of strings with a newline between each element
+  # skipping strings containing only whitespace characters
+  concatNonEmptyStringsNewline =
```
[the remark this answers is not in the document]
stasjok: That was a hard one. `concatNonEmptyLines` was one of the choices, but it's similar to `concatLines` from nixpkgs library and their behavior is different. concatLines is for concatenating one-line-strings, that's why it appends newlines to every string. This function does a different thing. I named it after concatStringsSep with a Sep suffix changed to the actual separator, since my function doesn't separator as an argument.

But since you suggested `concatNonEmptyLines` as a best choice, so be it. I don't like my choice anyway.

lib/utils.nix
```diff
@@ -118,4 +126,24 @@ with lib;
...
   '';
+
+  # Wrap Vimscript for using in lua,
+  # but only if the string contains something other than whitespaces
+  wrapVimscriptForLua =
+    string:
+    optionalString (hasContent string) ''
+      vim.cmd([[
+      ${string}
+      ]])
```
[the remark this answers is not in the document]
stasjok: That's a valid point. To be honest I didn't even considered a possibility that `]]` can be in the string...

stasjok:
Since other PRs are merged, tests are passing.

lib/utils.nix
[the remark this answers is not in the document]
MattSturgeon: ~~[This looks like](https://noogle.dev/f/lib/concatLines) the same behaviour to me:~~

```nix
concatLines [ "foo" "bar" ]
=> "foo\nbar\n"
```

`concatLines` is _almost_ the same as `concatStringsSep "\n"`.

Other than the trailing newline, which is only added by `concatLines`, incidentally that's probably what we actually _want_ in this case?

Maybe the implementation should actually be `lines: concatLines (filter hasContent lines)`?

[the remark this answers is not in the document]
stasjok: > Maybe the implementation should actually be `lines: concatLines (filter hasContent lines)`?

No, actually I don't want a trailing newline. The reason for this is that most of the time config parts are done like this:

```nix
{
  var = ''
    text
  ''
}
```
This adds a trailing newline by itself. With `concatLines` it'll be two trailing newlines. In my config I just make sure, that all my parts have trailing newlines by itself (parts from files also have trailing newlines, because I don't use `: set nofixendofline`. But the resulting concatenated result wouldn't be right if some strings are assigned with `var = "text";`. To fix this case I usually use a map with first stripping trailing newlines, than adding exactly one.

tests/test-sources/modules/output.nix
```diff
@@ -1,7 +1,92 @@
+{ pkgs, helpers, ... }:
+let
+  inherit (pkgs) lib;
+in
```
MattSturgeon, line 4: I think we can remove this whole section and access these (where needed) via module args in each test case's module

E.g.

```nix
all-configs.module = 
  {
    config,
    pkgs,
    lib,
    helpers,
    ...
  }:
```

stasjok: I wanted to ask something unrelated. Is it intended that I can't just use module as a function without `module` attr? It's because here

https://github.com/nix-community/nixvim/blob/6dc0bda459bcfb2a38cf7b6ed1d6a5d6a8105f00/tests/default.nix#L47

it doesn't handled that it can be a function. To workaround this I have to do `all-configs.module` instead of just `all-configs`

lib/utils.nix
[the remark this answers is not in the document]
MattSturgeon: I'm not sure I agree, since many of the places changed in _this_ PR are going from either using the pattern you describe or using `lib.concatLines` to now using `helpers.concatNonEmptyLines` _without_ a trailing newline.

Since the trailing newline is only added when there is content, I'm not sure I see the issue? Maybe I'm missing your point?

We could provide a `helpers.concatNonEmptyLines'` that takes an additional "trailing" boolean param, but this is a little over-engineered IMO.

tests/test-sources/modules/output.nix
MattSturgeon: > Is it intended that I can't just use module as a function without module attr?

It's not explicitly intended, it's just an edge-case I decided not to handle in order to keep the implementation simpler.

We could gate the `removeAttrs` behind a `isAttrs` or `? tests` check, but doing so felt a little bloated when I initially added full module support to the tests. :shrug: 

Don't really mind either way.

lib/utils.nix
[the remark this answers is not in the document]
stasjok: > I'm not sure I see the issue

Since resulted config is passed through stylua, it's not a big deal. As I mentioned in comments several days ago, I often nit-picking those little things, so that my generated config would be nearly perfect. There is another case I didn't touch here, that I handled in my config: when doing `wrapDo` you actually need to indent every line (except empty lines). Because it's not handled now I intentionally wasn't indenting a first line of the string in `wrapVimscriptForLua`.

> since many of the places changed in _this_ PR are going from either using the pattern you describe or using `lib.concatLines` to now using `helpers.concatNonEmptyLines` _without_ a trailing newline.

What are many places? There is only one place where I changed `concatLines` to `helpers.concatNonEmptyLines`. It's in the opts module. And my version doesn't have problems, because those strings, that are concatenated by `concatNonEmptyLines` all has trailing newlines by itself. But old version has two trailing newlines (one from indented string and one from cancatLines).

For best results all strings concatenated by `concatNonEmptyLines` should have traling newlines. We can handle strings not having trailing newlines by ensuring all parts are have them. `removeSuffix "\n" + "\n"` should do it.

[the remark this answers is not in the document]
MattSturgeon: A more general function that takes `sep` as an argument probably shouldn't be normalising line endings, since such functions are not always used on _lines_.

BTW the mapping function `(s: lib.removeSuffix "\n" s + "\n")` will always insert trailing newlines, like `lib.concatLines` does. It's just less likely to insert additional newlines if it is called with "dirty" input.

I think we should avoid our functions being too magic in unexpected ways. IMO the simplest implementation is to match upstream functionality, in this case just use `lib.concatLines` but filter the list for non-empty lines.

If we have an issue of passing in unintended `\n` in the input lines then we should address that in the calling code (could be another PR).

> It is similar to `concatLines` now, but with separator.

I don't see why we would ever need a different separator to `\n` if we know we're concatenating lines?

IMO this is getting overly complex for the sake of additional newlines that a) aren't that important for _this_ PR and b) should really be addressed in the calling code instead of the helper function.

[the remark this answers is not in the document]
MattSturgeon: > Can you be more specific? What I need to do?

I think we should have:

```nix
concatNonEmptyLines = lines: concatLines (filter hasContent lines);
```

> Because I don't see a problem with my function. I never use `concatLines` myself, because the behavior I want almost always is `concatStringsSep "\n"` with a filter for empty strings.

If you really want, we could also have:

```nix
concatNonEmptyLines' =
  trailing: lines:
  let
    lines' = filter hasContent lines;
  in
  optionalString (lines' != [ ]) (
    concatStringsSep "\n" lines'
    + optionalString trailing "\n"
  );

concatNonEmptyLines = concatNonEmptyLines' true;
```

Depending on how often we really need that.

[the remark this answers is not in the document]
stasjok: > b) should really be addressed in the calling code instead of the helper function.

So if I want it to be addressed _in the calling code_, do you recommend me to use this, righ?

```nix
    content =
      if config.type == "lua" then
        # Lua
        helpers.concatNonEmptyLines [
          config.extraConfigLuaPre
          (helpers.wrapVimscriptForLua config.extraConfigVim)
          config.extraConfigLua
          (lib.removeSuffix "\n" config.extraConfigLuaPost)
        ]
      else
        # Vimscript
        helpers.concatNonEmptyLines [
          (helpers.wrapLuaForVimscript config.extraConfigLuaPre)
          config.extraConfigVim
          (lib.removeSuffix "\n" (
            helpers.wrapLuaForVimscript (
              helpers.concatNonEmptyLines [
                config.extraConfigLua
                (lib.removeSuffix "\n" config.extraConfigLuaPost)
              ]
            )
          ))
        ];
```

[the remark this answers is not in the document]
stasjok: Anyway, you are a maintainer here, so I'll change as you suggested even though your suggested function definitely not want I need. But anyway, let's it have extra newlines. I don't have enough energy to argue, better to finish it somehow, because I guess we won't agree on this one.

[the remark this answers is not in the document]
MattSturgeon: If `extraConfigLuaPost` is likely to have trailing newline(s) then so is `extraConfigLua`, `extraConfigVim` and `extraConfigLuaPre`.

If we're getting that nit-picky, then `removeSuffix "\n"` isn't a robust solution, since there could be any number of trailing newlines hypothetically.

IMHO this isn't in the scope of this PR and we aren't making things any worse than the existing implementation by going with the simple option for now.

If you really think it's important to normalise the ending of those options, maybe we can do that in the options' `apply` functions? But again, I think we should keep that out of _this_ PR.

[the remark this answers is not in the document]
MattSturgeon: > I'll change as you suggested even though your suggested function definitely not what I need. But anyway, let's it have extra newlines. I don't have enough energy to argue

Sorry if that got unnecessarily heated. That wasn't my intention! :sweat_smile: 

I'm not saying we definitely shouldn't deal with redundant newlines, I just don't want to deal with that additional complexity in this PR since it primarily focuses on keeping the output empty by default.

I'm also not saying we definitely should fix them, but we can discuss potential solutions in another issue/discussion/PR/matrix.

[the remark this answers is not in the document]
stasjok: > If extraConfigLuaPost is likely to have trailing newline(s) then so is extraConfigLua, extraConfigVim and extraConfigLuaPre.

But my problem is with with `concatNonEmptyLines` producing unnecessarily newline at the end, so I don't need to strip other variables. They will be concatenated properly. Only the last one shouldn't have newline to count the one added by concatLines. So my complain is that you take my function that get the job done (because my guess is that at least 90% strings already have trailing newlines due to using indented string style) and turn in the function that doesn't.

> If we're getting that nit-picky, then `removeSuffix "\n"` isn't a robust solution, since there could be any number of trailing newlines hypothetically.

That's right, but caller may want it to have some trailing newlines, that's why I usually strip only one. Usecase for this is that both strings from here will be similar:

```nix
{
  a = "txt;
  b = ''
    txt
  ''
}
```

> IMHO this isn't in the scope of this PR and we aren't making things any worse than the existing implementation by going with the simple option for now.

Originally not, but when you suggested improving, I take it to my heart. So with my second commit I wanted to not only produce _whitespace only_ config, but clean up config generation. Anyway, I already gave in to do it your way. I guess comments are refreshed too late, because we often reply to the penultimate comments.

modules/opts.nix
```diff
@@ -55,31 +55,35 @@ in
...
+                for k,v in pairs(${varName}) do
+                  vim.${luaApi}[k] = v
+                end
+              end
+              -- }}}
+            ''
+          ) optionsAttrs
+        );
+      in
+      mkIf (helpers.hasContent content) content;
```
MattSturgeon, line 87: Is there a scenario where `content` could be empty whitespace, but not actually `""`?

I think `concatNonEmptyLines` ensures that its result either has content or is `""`.

(`concatLines` only has a trailing newline if the list is non-empty, and the list will be empty if all items failed the `hasContent` filter).

Not an issue, I'm just asking for my own understanding... In case I'm missing something again :grin:

MattSturgeon (approved):
LGTM, thanks for your efforts and your patience explaining things to me and debating various pros/cons!

I think this change will be appreciated by many!

@nix-community/nixvim any final suggestions before merging?

modules/opts.nix
stasjok: Your are right. My mistake. Mistake was made because because earlier I considered two options:

1. `concatLines` at the top and `helpers.hasContent content` as mkIf condition.
2. `helpers.concatNonEmptyLines` at the top and `content != ""` as a mkIf condition.

So I was rushed changing it before dinner and thought that since I'm adding something with `concatLines` at the top, I have to change also the condition, but I actually don't. With _NonEmpty_ part of the function it would work fine.

stasjok: Changed it back.

MattSturgeon: No problem!

I hope you don't feel too much pressure to produce perfect code; there's really no such thing. :joy:

Everything can always be improved (or done differently) and sometimes it just comes down to preference or giving up and "just making it work".

Feedback in a code review is never personal and isn't always meant as criticism :smiley:

stasjok: > Everything can always be improved (or done differently) and sometimes it just comes down to preference or giving up and "just making it work".

In this case I don't agree that this is just "doing differently". For example I don't have complains in the byte-compilation PR, because I can understand that my code could be not very bad (actually I'm not a programmer, I'm sysops, and I'm junior at best at programming), or suitable, or maintainable. In this case I just think that the new behavior is wrong and counter-interact what I already fixed.

```
nix-repl> c.config.content  
"-- Ignore the user lua configuration\nvim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/n
vim\nvim.opt.runtimepath:remove(vim.fn.stdpath('config') .. \"/after\")  -- ~/.config/nvim/after\nvim.opt.runtimepath:r
emove(vim.fn.stdpath('data') .. \"/site\")     -- ~/.local/share/nvim/site\n\n"
```

See "\n\n" at the end? They will appear in most cases where this function is used. Fortunately stylua will strip all extra newlines, so I'm fine with that at the end. Just need to relax :).

MattSturgeon: > `vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim`

Is this correct output? At a glance it looks like it's missing a linebreak :thinking: 

> In this case I don't agree that this is just "doing differently".
> See "\n\n" at the end? They will appear in most cases where this function is used.

As I said before, we can revisit in the future and figure out the best solution.

We don't need to achieve perfection in a single PR :grinning: 

> Fortunately stylua will strip all extra newlines, so I'm fine with that at the end. Just need to relax :)

:heart:

stasjok: > Is this correct output? At a glance it looks like it's missing a linebreak 🤔

Everything is fine except trailing newlines. It's added to every wrapped config here:

https://github.com/nix-community/nixvim/blob/6dc0bda459bcfb2a38cf7b6ed1d6a5d6a8105f00/modules/top-level/output.nix#L150-L153

MattSturgeon: I think if you had additional "lines" in the `concatNonEmptyLines` call, they would also (usually) end up with `"\n\n"`, one `\n` from the line's definition and one from the concatenation.

While this would likely make the string more readable, you could debate whether or not it is "intended". So it's not just the _final_ trailing line that is an issue, although that is _more_ of an issue.

I think in another PR we can look into normalising this and figuring out exactly what we want. Unless we decide to just rely on stylua for formatting issues like this.

stasjok: > I think if you had additional "lines" in the `concatNonEmptyLines` call, they would also (usually) end up with `"\n\n"`, one `\n` from the line's definition and one from the concatenation.

Are you talking about the middle strings? That's exactly what I want: so that every string (I call it config part) is separated with empty line (empty line means double newline characters). So it is exactly what I want: double newlines between config parts and a single newline at the end. My original function was just relying on provided strings already have trailing newlines (which is the case for almost all config parts, because they are written with indented strings). So my original function worked good enough so that I didn't bother much, or would want to fix original string definition to add trailing newline for consistent behavior. So the only thing that I possibly may want to impoove is to normalize strings. Look again at my example of my «perfect» function:

```
nix-repl> lib.x86_64-linux.helpers.concatNonEmptyLinesSep "" ["a" "b\n" "" "\n" "c"]   
"a\nb\nc\n"

nix-repl> lib.x86_64-linux.helpers.concatNonEmptyLinesSep "\n" ["a" "b\n" "" "\n" "c"] 
"a\n\nb\n\nc\n"
```

Second case is what I wanted in modules/output: two newlines between config parts and only one at the end.

> I think in another PR we can look into normalising this and figuring out exactly what we want. Unless we decide to just rely on stylua for formatting issues like this.

Let's not debate on this anymore. I'm not planning making another PR. The implementation here was good enough. All that can be improved from my point of view is adding my _too magic_ function. I think it's exactly what I want most of the time and it is suitable for both cases: when I want double newlines between strings and when one newline is enough.

Stylua is great, it'll fix all trailing newlines and chomp all triple, quadruple etc newlines to double. It doesn't do anything for Vim-type configs, but who is even using it?

GaetanLepage (approved):
LGTM after a rapid glance.
I trust @MattSturgeon for the details.

Thank you very much @stasjok for this great work and for addressing the detailed review !

GaetanLepage:
@mergifyio queue

PR #1910, opened 2024-07-21
systemModules should be systemPackages in README example

iclanzan:
In the standalone example, I think `environment.systemModules` was intended to be `environment.systemPackages`.


MattSturgeon (approved):
Great, thanks!

PR #1916, opened 2024-07-23
plugins/lsp/jdtls: add Eclipse JDT language server for Java

noar-t:
This adds the Eclipse JDT language server for Java. The existing Java LSP, "Java Language Server" was not suiting my use case. JDT seems to work better for me, so this adds that functionality to the list of LSPs.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -239,6 +239,12 @@ let
       cmd = cfg: [ "${cfg.package}/bin/java-language-server" ];
     }
     {
+      name = "jdt-language-server";
+      description = "Eclipse JDT Language Server for Java";
+      serverName = "jdtls";
+      cmd = cfg: [ (lib.getExe cfg.package) ];
+    }
+    {
       name = "jsonls";
       description = "jsonls for JSON";
       package = pkgs.vscode-langservers-extracted;
--- a/tests/test-sources/plugins/lsp/_lsp.nix
+++ b/tests/test-sources/plugins/lsp/_lsp.nix
@@ -142,6 +142,7 @@
             html.enable = true;
             htmx.enable = true;
             java-language-server.enable = true;
+            jdt-language-server.enable = true;
             jsonls.enable = true;
             jsonnet-ls.enable = true;
             julials.enable = true;


traxys:
```

File: plugins/lsp/language-servers/default.nix [hunk 1]
```diff
@@ -238,6 +238,12 @@ let
       serverName = "java_language_server";
       cmd = cfg: [ "${cfg.package}/bin/java-language-server" ];
     }
+    {
+      name = "jdt-language-server";
+      description = "Eclipse JDT Language Server for Java";
+      serverName = "jdt_language_server";
```
MattSturgeon, line 244: The CI error means this line is wrong, we should reference the nvim-lspconfig docs to see what it should be.

noar-t: Thanks, I will look deeper into it. I was having trouble running the test suite locally as it targets all architectures. Is there an easy way to emulate that or something? I am on NixOS 24.05

noar-t: Looks like they drop the whole name in favor of `jdtls`, https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#jdtls

MattSturgeon: > Thanks, I will look deeper into it.

Brilliant 👍 

> I was having trouble running the test suite locally as it targets all architectures.

It's not important to run all systems locally, it's fine to let CI handle that.

Usually you just need the `tests` command made available in the devshell (`nix develop`).

plugins/lsp/language-servers/default.nix
```diff
@@ -238,6 +238,12 @@ let
       serverName = "java_language_server";
       cmd = cfg: [ "${cfg.package}/bin/java-language-server" ];
     }
+    {
+      name = "jdt-language-server";
+      description = "Eclipse JDT Language Server for Java";
+      serverName = "jdtls";
+      cmd = cfg: [ "${cfg.package}/bin/jdtls" ];
```
MattSturgeon, line 245: Any reason this can't just be:
```suggestion
      cmd = cfg: [ (lib.getExe cfg.package) ];
```
?

I'd assume is the default when `cmd` is not specified, so is this line even needed?

If it isn't the default, maybe it should be :thinking:

noar-t: I do not see `lib.getExe` used anywhere in the existing list of language servers, and looked through `_mk-lsp.nix` and `lsp/default.nix` and it doesn't seem to perform that mapping as a default. If you feel strongly I can take a stab at a new revision.

MattSturgeon: Looks like the default is `cfg: null`, so currently it is needed.

It gets used here: https://github.com/nix-community/nixvim/blob/216d64c158da5523d5b3db0895e1345175c21502/plugins/lsp/language-servers/_mk-lsp.nix#L57

I think it can be improved but not in this PR.

I don't see why `lib.getExe` can't be used though...

I suspect it initially wasn't known about and the precedent of not using it has just been followed without question so far.

traxys: Yeah I think it's exactly the reason. lib.getExe is more maintainable than hardcoding the executable name

MattSturgeon (approved):
Could you rebase so that the merge commits get squashed into your actual change?

Also, I think `"plugins/lsp/jdtls: init"` is fine for the commit summary, since the extra context is available in the commit body.

If not we can handle this ourselves, but you won't get "full" commit attribution.

plugins/lsp/language-servers/default.nix
noar-t: I added this to the change when I rebased.

noar-t:
I believe it should all be rebased properly now, and includes the `lib.getExe` function to get the command properly. Let me know if there is anything else needed.

MattSturgeon:
@mergifyio queue

PR #1918, opened 2024-07-24
plugins/guess-indent: init

GGORG0:
Added https://github.com/NMAC427/guess-indent.nvim

### DIFF
--- a/plugins/utils/guess-indent.nix
+++ b/plugins/utils/guess-indent.nix
@@ -1,0 +1,69 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+helpers.neovim-plugin.mkNeovimPlugin config {
+  name = "guess-indent";
+  originalName = "guess-indent.nvim";
+  defaultPackage = pkgs.vimPlugins.guess-indent-nvim;
+
+  maintainers = [ helpers.maintainers.GGORG ];
+
+  settingsOptions = {
+    auto_cmd = helpers.defaultNullOpts.mkBool true ''
+      Whether to create autocommand to automatically detect indentation
+    '';
+
+    override_editorconfig = helpers.defaultNullOpts.mkBool false ''
+      Whether or not to override indentation set by Editorconfig
+    '';
+
+    filetype_exclude =
+      helpers.defaultNullOpts.mkListOf types.str
+        [
+          "netrw"
+          "tutor"
+        ]
+        ''
+          Filetypes to ignore indentation detection in
+        '';
+
+    buftype_exclude =
+      helpers.defaultNullOpts.mkListOf types.str
+        [
+          "help"
+          "nofile"
+          "terminal"
+          "prompt"
+        ]
+        ''
+          Buffer types to ignore indentation detection in
+        '';
+
+    on_tab_options = helpers.defaultNullOpts.mkAttrsOf types.anything { expandtab = false; } ''
+      A table of vim options when tabs are detected
+    '';
+
+    on_space_options =
+      helpers.defaultNullOpts.mkAttrsOf types.anything
+        {
+          expandtab = true;
+          tabstop = "detected";
+          softtabstop = "detected";
+          shiftwidth = "detected";
+        }
+        ''
+          A table of vim options when spaces are detected
+        '';
+  };
+
+  settingsExample = {
+    auto_cmd = false;
+    override_editorconfig = true;
+    filetype_exclude = [ "markdown" ];
+  };
+}
--- a/tests/test-sources/plugins/utils/guess-indent.nix
+++ b/tests/test-sources/plugins/utils/guess-indent.nix
@@ -1,0 +1,47 @@
+{
+  empty = {
+    plugins.guess-indent.enable = true;
+  };
+
+  defaults = {
+    plugins.guess-indent = {
+      enable = true;
+
+      settings = {
+        auto_cmd = true;
+        override_editorconfig = false;
+        filetype_exclude = [
+          "netrw"
+          "tutor"
+        ];
+        buftype_exclude = [
+          "help"
+          "nofile"
+          "terminal"
+          "prompt"
+        ];
+        on_tab_options = {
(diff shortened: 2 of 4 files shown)


plugins/utils/guess-indent.nix
```diff
@@ -0,0 +1,61 @@
+{
+  lib,
+  helpers,
+  config,
+  pkgs,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin config {
```
GaetanLepage, line 8:
```suggestion
}:
with lib;
helpers.neovim-plugin.mkNeovimPlugin config {
```

Even though this is not recommended in nixpkgs, we adopted this convention across the nixvim codebase.

GaetanLepage (dismissed):
This looks in a very good state already.

GGORG0:
Thanks for the blazingly fast review! Also, I didn't run the testing suite (because it is just too heavy for my machine), so I didn't notice the `maintainers` error

plugins/utils/guess-indent.nix
MattSturgeon: Should we reconsider this? Perhaps we could `inherit (lib) X Y Z` instead?

At least in new files - I'm not proposing we go around changing this in existing code for the sake of it.

tests/test-sources/plugins/utils/guess-indent.nix
```diff
@@ -0,0 +1,17 @@
+{
+  empty = {
+    plugins.guess-indent.enable = true;
+  };
+
```
GaetanLepage, line 5: Could you please add a `defaults` test case, with all the options explicitly set to their default value ?

MattSturgeon:
> I didn't run the testing suite (because it is just too heavy for my machine)

That's reasonable, however note that in the deveshell (`nix develop`) we have a (slightly) lighter `tests` command that only runs the tests in `/tests/test-sources` rather than _all_ checks. It's also possible to select just the specific tests you want to run.

tests/test-sources/plugins/utils/guess-indent.nix
GGORG0: Isn't that just what `empty` does? Or should I explicitly set all options to their default value in another testcase?

MattSturgeon:
Out of curiosity: does this plugin have the same goal as tpope's slueth? How does it differ?

tests/test-sources/plugins/utils/guess-indent.nix
MattSturgeon: No, `empty` will leave all options set to `null` (which we will filter out when calling `setup`).

The "plugin defaults" exist only as documentation, so a `defaults` test is to check they actually work when explicitly set.

(There's no way to test they are actually the default upstream though, afaik)

GGORG0:
> Out of curiosity: does this plugin have the same goal as tpope's slueth? How does it differ?

Yes, it does. However it's written for nvim specifically. Here you can see the comparison: https://github.com/NMAC427/guess-indent.nvim/tree/6cd61f7a600bb756e558627cd2e740302c58e32d?tab=readme-ov-file#alternatives

GGORG0:
Do the CI checks normally take 20+ hours? :disappointed_relieved: That's so dumb that there isn't any caching or selective testing involved. The reviews and bugfixing were done in <30 minutes, and the tests are literally still running...

MattSturgeon:
> Do the CI checks normally take 20+ hours? 😅 That's so dumb that there isn't any caching or selective testing involved. The reviews and bugfixing were done in <30 minutes, and the tests are literally still running...

The tests are [actually finished](https://buildbot.nix-community.org/#/builders/43/builds/366) (all passed) but something has gone wrong reporting some of them to GitHub (because there's over 1000 and we have hit a rate limit)

GGORG0:
> The tests are [actually finished](https://buildbot.nix-community.org/#/builders/43/builds/366) (all passed)

Still, they have finished just **4 hours ago**. That's 16 hours. 

> but something has gone wrong reporting some of them to GitHub (because there's over 1000 and we have hit a rate limit)

Yeah, I'm not surprised - maybe they should just report as 1 huge test, or a small number of categorized tests. 

By the way, is there anything else to do with this PR or is it ready to merge (and GitHub is preventing that because of the unfinished tests)?

MattSturgeon:
> By the way, is there anything else to do with this PR or is it ready to merge (and GitHub is preventing that because of the unfinished tests)?

Could you rebase on main and squash to two commits (one adding you to maintainers, and another adding the plugin)?

Only the `nix-eval` test needs to be reported as passed for mergify to be happy, but we can merge manually without it.

The main thing is having at least one approving review.

I'll try and review properly shortly!

> Yeah, I'm not surprised - maybe they should just report as 1 huge test, or a small number of categorized tests.

We had one huge test until very recently, which buildbot couldn't handle as the memory requirements ended up being extreme.

We'll probably end up with a compromise where we have maybe 20 grouped tests, but we're still experimenting with finding a good balance.

MattSturgeon (approved):
LGTM, once squashed and rebased! Thanks for your efforts!

GGORG0:
> Could you rebase on main and squash to two commits (one adding you to maintainers, and another adding the plugin)?

I think I did what I was supposed to... I had to learn about rebasing, because this is my first serious PR to a big project

MattSturgeon:
@mergifyio queue

MattSturgeon:
> > Could you rebase on main and squash to two commits (one adding you to maintainers, and another adding the plugin)?
> 
> I think I did what I was supposed to... I had to learn about rebasing, because this is my first serious PR to a big project

Looks perfect, thanks!

plugins/utils/guess-indent.nix
JarvisCraft: +1 on *not* using `with`

`nixpkgs` has a tendency on getting rid of it, at least partially (NixOS/nixpkgs#292468) and [best practices] also suggests that `inherit` should be preferred.

[best practices]: https://nix.dev/guides/best-practices#with-scopes

Issue #1854, opened 2024-07-11
[BUG] Otter stopped working after 435ef28


PerchunPak:
| Field        | Description |
|--------------|-------------|
| Plugin       | `otter`  | 
| Nixpkgs      | `unstable, 655a58a7` | 
| Home Manager | `unstable, 3bfaacf4` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

Some time ago I noticed that otter stopped working and today decided to investigate it, after bisecting I found out that 435ef28 is the first commit where otter stopped working

Otter is implemented here: https://github.com/nix-community/nixvim/blob/main/plugins/languages/otter.nix

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim.plugins = {
  otter.enable = true;
  cmp.settings.sources = [ { name = "otter"; } ];
}
```

Tbh I am not sure that example can reproduce the error, it would probably need treesitter and maybe lsp as well, so I just did `sudo nixos-rebuild --flake ... --override-input nixvim ~/dev/nixvim` when bisected.

MattSturgeon:
Thanks for reporting! Are you sure that commit is the issue? I don't see how it could affect your plugin... Does otter interact with treesitter at all?

Any thoughts @khaneliman?

This PR on the other hand did change some behavior by removing the `addCmpSources` option: https://github.com/nix-community/nixvim/pull/1776, though based on your MRE you've noticed that already.

khaneliman:
> Thanks for reporting! Are you sure that commit is the issue? I don't see how it could affect your plugin... Does otter interact with treesitter at all?
> 
> Any thoughts @khaneliman?
> 
> This PR on the other hand did change some behavior by removing the `addCmpSources` option: #1776, though based on your MRE you've noticed that already.

No idea, doesn't touch any related code.

MattSturgeon:
Could you clarify what you mean by "stopped working"?

One thing the treesitter PR did change, is that `plugins.treesitter.settings.highlight.enable` is no longer true by default.

khaneliman:
> Could you clarify what you mean by "stopped working"?
> 
> One thing the treesitter PR did change, is that `plugins.treesitter.settings.highlight.enable` is no longer true by default.

Looking at otter's github, it does look like treesitter is a dependency but it doesn't say what for. This would make sense if it's hooking into it's highlighting for something.

PerchunPak:
[...]

PR #1922, opened 2024-07-25
plugins/otter: add warning if treesitter highlight not enabled

khaneliman:
Follow up to https://github.com/nix-community/nixvim/issues/1854

### DIFF
--- a/plugins/languages/otter.nix
+++ b/plugins/languages/otter.nix
@@ -83,4 +83,14 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       When true, otter handles these cases fully. This is a (minor) performance hit.
     '';
   };
+
+  extraConfig = cfg: {
+    warnings =
+      lib.optional (cfg.enable && config.plugins.treesitter.settings.highlight.enable == null)
+        ''
+          NixVim(plugins.otter): you have enabled otter, but `plugins.treesitter.settings.highlight.enable` is not enabled.
+          Otter functionality might not work as expected without it and `plugins.treesitter.enable` enabled.
+        '';
+  };
+
 }


plugins/languages/otter.nix
```diff
@@ -83,4 +83,12 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       When true, otter handles these cases fully. This is a (minor) performance hit.
     '';
   };
+
+  extraConfig = cfg: {
+    warnings = lib.optional (!(cfg.enable -> config.plugins.treesitter.settings.highlight.enable)) ''
```
MattSturgeon, line 88:
```
error: expected a Boolean but found null: null
at /nix/store/l6ml4gkff7nzqfqwz4j86a696cnfpcql-source/plugins/languages/otter.nix:88:43:
87| extraConfig = cfg: {
88| warnings = lib.optional (!(cfg.enable -> config.plugins.treesitter.settings.highlight.enable)) ''
| ^
```

Since the `highlight.enable` settings option is nullable, we need to handle that.

Note: if you did `enable == true` then `nix fmt` would have "fixed" that for you, leading to this error.

These would be true when enable is null: `enable != null -> enable` or `isBool enable -> enable`

But I think we actually want false when null, so lets go with: `enable != null && enable` or `isBool enable && enable`.

Probably best to have a small helper in a let block: `isTrue = v: isBool v && v`, or maybe we can use [these helpers](https://github.com/nix-community/nixvim/blob/216d64c158da5523d5b3db0895e1345175c21502/lib/utils.nix#L106-L110)?

traxys: Could we only check for `null` and not `false`? Maybe the user will do _something else_ to enable it, and as such we don't want to warn him each time?

plugins/languages/otter.nix
```diff
@@ -83,4 +83,14 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       When true, otter handles these cases fully. This is a (minor) performance hit.
     '';
   };
+
+  extraConfig = cfg: {
+    warnings =
+      lib.optional (cfg.enable -> config.plugins.treesitter.settings.highlight.enable == null)
```
MattSturgeon, line 89:
```suggestion
      lib.optional (cfg.enable && config.plugins.treesitter.settings.highlight.enable == null)
```

Using `->` here would show the warning whenever `cfg.enable == false`

khaneliman: That's not what i'm observing in the local tests. If I set otter to disabled in the tests  I dont get the warning.

MattSturgeon: That's probably because `highlight.enable == null` when you're testing?

`a -> b` is equivalent to `!a || b`. Here I'm pretty sure we want to show the warning only when _both_ `enable` is true and `highlight.enable` is null.

khaneliman: I added ``plugins.treesitter.settings.highlight.enable = true; to the default test and then set otter enable = false.
``

khaneliman: I tried a few different combinations in the otter default test. They all didn't result in a warning. Is the mkNeovimPlugin doing something to prevent it?

- `plugins.treesitter.settings.highlight.enable = true;` and `otter.enable = false`
- `plugins.treesitter.settings.highlight.enable = false;` and `otter.enable = false`
- `plugins.treesitter.settings.highlight.enable = null;` and `otter.enable = false`

khaneliman: That said, I see the same behavior doing it with -> and && so I can change it, just wanted to point out I wasn't seeing the behavior mentioned.

MattSturgeon:
Should we also be checking `plugins.treesitter.enable`? I guess there'd be no way to disable the warning if the user doesn't want to use our treesitter module though.

MattSturgeon (approved):
LGTM, thanks!

#1926 is in the merge queue atm, if I queue this then mergify will rebase it (in Gaétan's name) before merging.

I don't mind waiting for you to rebase after the other PR is merged before queueing.

MattSturgeon:
@mergifyio queue

PR #1927, opened 2024-07-25
plugins/which-key: fix registrations type

### LINKED ISSUE #1901: Update which-key keybind spec
Which-key is moving to a new specification for defining keybindings for version 3. This new spec appears to be similar to lazy.nvim's keybind specification. While this is just a warning right now, this could be a breaking change in the future.

Currently, nixvim documentation suggests to use the following format for definining `programs.nixvim.plugins.which-key.registrations`:
```nix
{ "<leader>p" = "Find git files with telescope";}
```
which generates the following Lua:
```lua
{ ["<leader>p"] = "Find git files with telescope" }
```

This is fine for the old which-key spec, but they are moving to a new spec. That keybinding would now be in Lua:
```lua
{
  {"<leader>p", desc = "Find git files with telescope"}
}
```

I'm relatively new to Nix, so figuring out how to get the `"<leader>p"` by itself in nix was a little tricky. Ended up reading the source code for `toLuaObject` and it looks like `__unkeyed` will do the trick. A simple band-aid fix in Nix could be:
```nix
programs.nixvim.plugins.which-key.registrations = [
  { __unkeyed = "<leader>p"; desc = "Find git files with telescope"; }
];
```
but this doesn't work, as `which-key.registrations` only takes a type "attribute set of anything," not  a list.

Another workaround could be to do the following for multiple keymaps:
```nix
programs.nixvim.plugins.which-key.registrations = {
  __unkeyed1 = { __unkeyed = "<leader>p"; desc = "Find git files with telescope"; };
  __unkeyed2 = { __unkeyed = "<leader>e"; desc = "some other keybind"; };
 __unkeyed3 ... ... ...
};
```
but this is really tedious, and it doesn't seem to work well either. It evaluates, the generated Lua *looks* right to my eyes, but which-key doesn't like it for some reason. Maybe I'm doing something wrong.

(sidenote: in to-lua.nix the comment says:
```
        If this option is true, attrsets like { __unkeyed.1 = "a"; b = "b"; }
        will render as { "a", b = "b" }
```
but this is not correct. Here's what I see in nix-repl.
```
nix-repl> inputs.nixvim.lib.x86_64-linux.helpers.toLuaObject {__unkeyed.1 = "a"; b = "b";}
error: syntax error, unexpected FLOAT, expecting '.' or '='

       at «string»:1:62:

            1| inputs.nixvim.lib.x86_64-linux.helpers.toLuaObject {__unkeyed.1
             |                                                              ^

nix-repl> inputs.nixvim.lib.x86_64-linux.helpers.toLuaObject {__unkeyed1 = "a"; b = "b";}
"{ \"a\", b = \"b\" }"

nix-repl> inputs.nixvim.lib.x86_64-linux.helpers.toLuaObject {__unkeyed1 = "a"; __unkeyed2 = "b";}
"{ \"a\", \"b\" }"

nix-repl> inputs.nixvim.lib.x86_64-linux.helpers.toLuaObject {__unkeyed.a = "a"; b = "b";}
"{ { a = \"a\" }, b = \"b\" }"
```
)

anyway, sorry for the long-winded report. I wanted to at least try to solve the problem before asking about it. Thanks for all the hard work you all do, as a noob to Linux and Nix things have been (mostly) smooth sailing so far.

PerchunPak:
Every time I open nvim, it gives me this message
image

And in `:checkhealth which-key` I see this:

```lua
Checking for issues with your mappings ~
- WARNING You're using an old version of the which-key spec.
  Your mappings will work, but it's recommended to update them to the new version.
  Please check the docs and suggested spec below for more info.
  Mappings: >
  {
    ["<leader>c"] = { name = "[C]ode" },
    ["<leader>d"] = { name = "[D]ocument" },
    ["<leader>r"] = { name = "[R]ename" },
    ["<leader>s"] = { name = "[S]earch" },
    ["<leader>w"] = { name = "[W]orkspace" }
  }
  
  -- Suggested Spec:
  {
    { "<leader>c", group = "[C]ode" },
    { "<leader>d", group = "[D]ocument" },
    { "<leader>r", group = "[R]ename" },
    { "<leader>s", group = "[S]earch" },
    { "<leader>w", group = "[W]orkspace" },
  }
```

Translated to nix syntax it is:

```nix
[
  [ "<leader>c" { group = "[C]ode"; } ]
  [ "<leader>d" { group = "[D]ocument"; } ]
  [ "<leader>r" { group = "[R]ename"; } ]
  [ "<leader>s" { group = "[S]earch"; } ]
  [ "<leader>w" { group = "[W]orkspace"; } ]
]
```

Fixes #1901

### DIFF
--- a/plugins/utils/which-key.nix
+++ b/plugins/utils/which-key.nix
@@ -13,12 +13,15 @@ with lib;
     package = helpers.mkPluginPackageOption "which-key-nvim" pkgs.vimPlugins.which-key-nvim;
 
     registrations = mkOption {
-      type = with types; attrsOf anything;
-      default = { };
+      type = with types; anything;
+      default = [ ];
       description = "Manually register the description of mappings.";
-      example = {
-        "<leader>p" = "Find git files with telescope";
-      };
+      example = [
+        [
+          "<leader>p"
+          { desc = "Find git files with telescope"; }
+        ]
+      ];
     };
 
     plugins = {


JarvisCraft (approved):
Also had this problem in the logs

BTW, should not some migration warning be emitted?

MattSturgeon:
This does switch the option to use the new format, however I'm planning to incorporate this into a refactor that properly deprecates the old format and allows configuring freeform settings

PerchunPak:
> BTW, should not some migration warning be emitted?

I guess it is emitted by `which-key`

> however I'm planning to incorporate this into a refactor

Could you merge this as a temporary workaround? The warning is really annoying.

MattSturgeon:
If we merge a temporary solution, it should be a new option that follows the new v2 spec (confusingly which-key V3 uses a "spec" version 2).

The old option should show a warning when used telling users to migrate.

The problem with doing this, is that users will have to migrate their config twice, unless we can make the temporary option the same as what the new one will be:

- `settings.spec`
- type `listOf (maybeRaw (attrsOf anything))`
- could be done with `helpers.defaultNullOpts.mkListOf (attrsOf anything)`

PerchunPak:
Yeah, I am too lazy for this. For now, I just commented my registrations definition.

PR #1930, opened 2024-07-25
doc: Use correct href for nuscht-search

traxys:
This commit also adds a subdirectory in the docs derivation, to have the same local structure than what is uploaded on github pages.


flake-modules/dev/server.py
```diff
@@ -12,5 +12,5 @@ def end_headers(self):
 
 
 with http.server.HTTPServer(("", PORT), UncachedHTTPHandler) as httpd:
-    print(f"Serving documentation at http://localhost:{PORT}")
+    print(f"Serving documentation at http://localhost:{PORT}/nixvim")
```
MattSturgeon, line 15: Is this new? For parity with gh-pages?

traxys: Yeah it's what I tried to convey in the PR message. The issue is that the href of nuscht-search _must_ be correct. So if we set it to `/nixvim/search` we need to access it at `https://localhost/nixvim/search/`.

MattSturgeon (approved):
LGTM!

traxys:
@Mergifyio queue

PR #1932, opened 2024-07-25
treewide: add global iconsEnabled option

josh:
While I really like Nerd Fonts, it would be great if there was a single global way to disable them for environments where Nerd Fonts won't be used. A number of plugins already provide individual `iconsEnabled` options while others handle icons flags in more ad-hoc way. It would be great to standardize on the `iconsEnabled` name and provide a top level global value they would default to.

Ideally disabling `iconsEnabled` prevents `pkgs.vimPlugins.nvim-web-devicon` from being installed at all. Though there's still some follow-up work that could be done to improve the per plugin handling of their own `iconsEnabled` options.

Open to naming this option whatever is preferred by the maintainers, but `iconsEnabled` seemed most used already.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -15,6 +15,7 @@
     ./files.nix
     ./filetype.nix
     ./highlights.nix
+    ./icons.nix
     ./keymaps.nix
     ./lua-loader.nix
     ./opts.nix
--- a/modules/icons.nix
+++ b/modules/icons.nix
@@ -1,0 +1,8 @@
+{ lib, ... }:
+{
+  options.iconsEnabled = lib.mkOption {
+    type = lib.types.bool;
+    description = "Toggle icon support. Installs nvim-web-devicons.";
+    default = true;
+  };
+}
--- a/plugins/git/diffview.nix
+++ b/plugins/git/diffview.nix
@@ -107,7 +107,7 @@ in
       useIcons = mkOption {
         type = types.bool;
         description = "Requires nvim-web-devicons";
-        default = true;
+        default = config.iconsEnabled;
       };
 
       showHelpHints = mkBool true ''
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -111,7 +111,7 @@ in
       iconsEnabled = mkOption {
         type = types.bool;
         description = "Whether to enable/disable icons for all components.";
-        default = true;
+        default = config.iconsEnabled;
       };
 
       theme = helpers.defaultNullOpts.mkNullable (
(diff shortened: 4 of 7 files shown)


plugins/lsp/lspsaga.nix
```diff
@@ -54,7 +54,7 @@ in
       ui = {
         border = helpers.defaultNullOpts.mkBorder "single" "lspsaga" "";
 
-        devicon = helpers.defaultNullOpts.mkBool true "Whether to use nvim-web-devicons.";
+        devicon = helpers.defaultNullOpts.mkBool config.iconsEnabled "Whether to use nvim-web-devicons.";
```
MattSturgeon, line 57: This doesn't do what you think it does.

This is a "plugin default" not a nix option default, so this will have no effect.

If we actually want to do this, it should be done using a `mkIf` + `mkDefault` config definition, however I feel uneasy about setting freeform settings options "automatically".

@traxys any thoughts?

josh: Isn't the plugin config defined as `cfg = config.plugins.lspsaga` in this scope on L10? `config` seems to be the correct parent scope. It seemed to all work when I tested it.

MattSturgeon: In `defaultNullOpts`, the "default" argument is _not_ the option default.

Instead, it is the "plugin default" and is only used for generating documentation. The _actual_ default is always `null`.

The "plugin default" has no effect on the option's value in nix. Rather it exists as a way to let users know what the upstream plugin will do when the option is undefined (`null`).

josh: Ah, okay, it's specific to `defaultNullOpts`. I can just revert my change to this file. Are other other `mkOption`s okay?

MattSturgeon:
> Are other other `mkOption`s okay?

Functionally, yes. However I think we need to scratch our heads and decide if we actually want to have a global `iconsEnabled` option.

Most of the plugins you've found using `mkOption` are doing so because we haven't gotten around to updating them to the new "freeform settings" format.

Going forward we want to avoid "magic" where we automatically configure stuff we think makes sense, without the user knowing.

This is a _little_ different, because the user should know they enabled the option globally, but it still blurs this line.

There's also a maintenance concern: when adding/updating plugins that have an `icons_enabled` option, we would have to remember to integrate with the global option.

I'm not saying we definitively don't want this, but my initial instinct is that this may be too "magic" for us. I'm open to having a debate here or on matrix.

It'd also be nice if @GaetanLepage and @traxys, along with any other users + contributes, chime in on the issue.

traxys:
Yeah I have a feeling this is a _bit_ too out of scope for nixvim, as it would be more of a feature of a neovim distribution (like LazyVim, LunarVim, ...), something that the current nixvim does not try to be.
I also would note like to have a single option that affects loads of options by default, if we want to go with a global option like this it would need to not do anything by default, and only do something if a user chooses true or false

PR #1933, opened 2024-07-25
github: Fix stable doc path

traxys:
The path has only been moved in main, not on the stable branch


MattSturgeon (approved):

traxys:
@Mergifyio queue

MattSturgeon:
@traxys this (or an equivalent change for `main`) needs backporting, since we also build the docs from `nixos-24.05` when it is pushed to.

PR #1940, opened 2024-07-27
plugins/lsp/nextls: init

tstachl:
Adds `nextls` the language server for Elixir that just works.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -311,6 +311,15 @@ let
       description = "metals for Scala";
     }
     {
+      name = "nextls";
+      description = "The language server for Elixir that just works.";
+      package = pkgs.next-ls;
+      cmd = cfg: [
+        "nextls"
+        "--stdio"
+      ];
+    }
+    {
       name = "nginx-language-server";
       description = "nginx-language-server for `nginx.conf`";
       serverName = "nginx_language_server";
--- a/tests/test-sources/plugins/lsp/_lsp.nix
+++ b/tests/test-sources/plugins/lsp/_lsp.nix
@@ -154,6 +154,7 @@
             lua-ls.enable = true;
             marksman.enable = true;
             metals.enable = true;
+            nextls.enable = true;
             nginx-language-server.enable = true;
             nickel-ls.enable = true;
             nil-ls.enable = true;


plugins/lsp/language-servers/default.nix
```diff
@@ -310,6 +310,16 @@ let
       name = "metals";
       description = "metals for Scala";
     }
+    {
+      name = "nextls";
+      description = "The language server for Elixir that just works.";
+      package = pkgs.next-ls;
+      serverName = "nextls";
+      cmd = cfg: [
+        "${cfg.package}/bin/nextls"
```
MattSturgeon, line 319: Could you use `(lib.getExe cfg.package)` here please?

tstachl: Done

plugins/lsp/language-servers/default.nix
```diff
@@ -310,6 +310,16 @@ let
       name = "metals";
       description = "metals for Scala";
     }
+    {
+      name = "nextls";
+      description = "The language server for Elixir that just works.";
+      package = pkgs.next-ls;
+      serverName = "nextls";
+      cmd = cfg: [
+        (lib.getExe cfg.package)
```
GaetanLepage, line 319:
```suggestion
        "nextls"
```
I think that this solution is more flexible.
Nixvim will ensure that `nextls` is on the `PATH`.

MattSturgeon: We don't do this for other LSPs. They all use an explicit absolute path into the nix store.

OTTOMH, I have a feeling the `cmd` is only used when `package` is non-null anyway?

GaetanLepage: For me, there is no link between the package and the cmd in most cases.
The `package` option (in nixvim) ensures that the LS package is set on the PATH within neovim.
The `cmd` tells neovim how to run the language server.

We could hardcode some `pkgs` path in `cmd` for extreme cases were the `PATH` mechanism doesn't work.
However, for me it should work here, no ?

tstachl: > I have a feeling the cmd is only used when package is non-null anyway?

[nextls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#nextls) does not set a `cmd` by default. I thought it would be a nice to have since it can be overwritten anyway.

GaetanLepage: > [nextls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#nextls) does not set a `cmd` by default. I thought it would be a nice to have since it can be overwritten anyway.

This sure, but I would have set it to `["nextls" "--stdo"]`

GaetanLepage: So that it's completely independent from how the user has installed `nextls`.

MattSturgeon: > For me, there is no link between the package and the cmd in most cases. The `package` option (in nixvim) ensures that the LS package is set on the PATH within neovim. The `cmd` tells neovim how to run the language server.

Ok that makes sense.

Unless I'm mis-remembering, most (all?) cmds currently use an absolute path, so should we be going through and changing that in another PR?

plugins/lsp/language-servers/default.nix
```diff
@@ -310,6 +310,16 @@ let
       name = "metals";
       description = "metals for Scala";
     }
+    {
+      name = "nextls";
+      description = "The language server for Elixir that just works.";
+      package = pkgs.next-ls;
+      serverName = "nextls";
```
GaetanLepage, line 317:
```suggestion
```
You don't need to set this as `serverName` defaults to `name`.

tstachl: Done

GaetanLepage (approved):
All good ! Thanks :)

tstachl:
Sorry, accidentally synced my fork into this branch, have to quickly rebase this.

tstachl:
@GaetanLepage fixed it, can you check again? I rebased with main and pushed the change again.

GaetanLepage:
@Mergifyio queue

PR #1944, opened 2024-07-28
plugins/which-key: deprecate v2 registrations

### LINKED ISSUE #1901: Update which-key keybind spec

MattSturgeon:
These have been replaced in v3 with a new spec.

This is an **interim** solution until we have time to fully migrate which-key to rfc42 freeform settings.

I've done this because the warning shown when starting vim is kinda obnixious and without this the only way to avoid it is to remove all registrations.

As we will (hopefully) migrate which-key to `settings` options soon, I've named the new option `settings.spec` so that we do not need to "rename" it again.

This PR **does not** actually add a freeform settings option.

The old option (`registrations`) is deprecated, when defined it will print:
```
trace: evaluation warning: nixvim (plugins.which-key):
The option definition `plugins.which-key.registrations' in `<unknown-file>' has been deprecated in which-key v3; please remove it.

```

(I haven't added a deprecation notice to the `apply`, so reading the option will not print warnings)

After this warning has been in place for a while, we can upgrade to a `mkRemovedOptionModule` assertion.

~~Marked as a draft since the test cases still need to be updated.~~

Fixes #1901
See https://github.com/nix-community/nixvim/pull/1927#issuecomment-2250725447

### DIFF
--- a/tests/test-sources/plugins/utils/which-key.nix
+++ b/tests/test-sources/plugins/utils/which-key.nix
@@ -8,46 +8,57 @@
       enable = true;
 
       # Testing for registrations
-      registrations."f" = {
-        prefix = "<leader>";
-        mode = [
-          "n"
-          "v"
-          "i"
-          "t"
-          "c"
-          "x"
-          "s"
-          "o"
-        ];
-        name = "Group Test";
-        f = "Label Test";
-        "1" = [
+      settings.spec =
+        let
+          mode = [
+            "n"
+            "v"
+            "i"
+            "t"
+            "c"
+            "x"
+            "s"
+            "o"
+          ];
+        in
+        [
           {
-            __raw = ''
+            __unkeyed-1 = "<leader>f";
+            group = "Group Test";
+            inherit mode;
+          }
+          {
+            __unkeyed-1 = "<leader>ff";
+            desc = "Label Test";
+            inherit mode;
+          }
+          {
+            __unkeyed-1 = "<leader>f1";
+            __unkeyed-2.__raw = ''
               function()
-                print("Raw Lua Code and List KeyMapping Test")
+                print("Raw Lua KeyMapping Test")
               end
             '';
+            desc = "Raw Lua KeyMapping Test";
+            inherit mode;
           }
-          "Raw Lua Code and List KeyMapping Test"
+          {
+            __unkeyed-1 = "<leader>foo";
+            desc = "Label Test 2";
+            inherit mode;
+          }
+          {
+            __unkeyed-1 = "<leader>f<tab>";
+            group = "Group in Group Test";
+            inherit mode;
+          }
+          {
+            __unkeyed-1 = "<leader>f<tab>f";
+            __unkeyed-2 = "<cmd>echo 'Vim cmd KeyMapping Test'<cr>";
+            desc = "Vim cmd KeyMapping Test";
+            inherit mode;
+          }
         ];
-        "oo" = "Label Test 2";
-        "<tab>" = {
-          name = "Group in Group Test";
-          f = [
-            {
-              __raw = ''
-                function()
-                  vim.cmd("echo 'Raw Lua Code and List KeyMapping Test 2'")
-                end
-              '';
-            }
-
-            "Raw Lua Code and List KeyMapping Test 2"
-          ];
-        };
-      };
 
       plugins = {
         marks = true;
--- a/plugins/utils/which-key.nix
+++ b/plugins/utils/which-key.nix
@@ -3,6 +3,7 @@
   helpers,
   pkgs,
   config,
+  options,
   ...
 }:
 with lib;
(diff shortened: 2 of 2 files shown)


tests/test-sources/plugins/utils/which-key.nix
```diff
@@ -8,46 +8,54 @@
...
+            "i"
+            "t"
+            "c"
+            "x"
+            "s"
+            "o"
           ];
-        };
-      };
+          name = "Group Test";
```
khaneliman, line 26: Kinda confused by this one. In reading the documentation, we aren't expected to supply name anymore and the group has a group property.

tests/test-sources/plugins/utils/which-key.nix
```diff
@@ -8,46 +8,54 @@
...
-                  vim.cmd("echo 'Raw Lua Code and List KeyMapping Test 2'")
-                end
-              '';
-            }
+      settings.spec = [
+        {
+          __unkeyed-1 = "f";
 
-            "Raw Lua Code and List KeyMapping Test 2"
+          prefix = "<leader>";
```
khaneliman, line 15: Also, doesn't seem like prefix is supported anymore.. when I create a group of mappings using a prefix with the mappings, they dont appear to group. If I include the prefix in each of the unkeyed attributes then they group properly.

khaneliman: Yeah, looks like its deprecated, too. https://github.com/folke/which-key.nvim/blob/6c1584eb76b55629702716995cca4ae2798a9cca/lua/which-key/mappings.lua#L51-L54

khaneliman: Example of what I'm talking about:
Using prefix: 
image
Using unkeyed with same combination in a single string:
image

tests/test-sources/plugins/utils/which-key.nix
khaneliman: Looks like name is deprecated, so I'd recommend switching it to `group`

tests/test-sources/plugins/utils/which-key.nix
MattSturgeon: Looks like `prefix` used to prefix the `lhs` (key) for every mapping defined in the tree, however I think now it is intended to be done explicitly for each mapping.

This makes sense as the new format is no longer a tree.

tests/test-sources/plugins/utils/which-key.nix
MattSturgeon: Ah, it seems `group` should also be used instead of `desc` when registering a partial-binding (group)

khaneliman (approved):
LGTM, used in my nvim flake and it fixed the mappings warnings. Warning messages from nixvim also pointed out all the locations I had defined registrations across modules.

GaetanLepage (approved):
LGTM !

MattSturgeon:
@mergifyio queue

ForsakenHarmony:
This new config format seems a bit unfortunate, would be nice if there was some remapping before it gets passed to lua, so the format isn't horrible in nix

MattSturgeon:
> This new config format seems a bit unfortunate, would be nice if there was some remapping before it gets passed to lua, so the format isn't horrible in nix

I agree it's not great, but it is an important goal for us to not do bespoke transformations for plugins/options.

While on an individual option you might think it is helpful, there's two major downsides to such magic transformation:
- We end up hiding the underlying lua config from users
- We have to make assumptions, which can end up being broken by upstream changes
- Adding such transformations and (more importantly) maintaining them is no small feat when scaled up to all plugins.

The whole point of "freeform" rfc42-style `settings` options is that we transparently pass the config through to the plugin without doing anything more than converting it to lua.

Even the steps we currently take to filter out `null` and empty attrs are too much transformation in my personal opinion, but that's very difficult to avoid with the way nix modules currently work.

PR #1947, opened 2024-07-28
plugins/treesitter: rename our injection file to nixvim-injections.scm

GaetanLepage:
This removes the following issue when enabling `performance.combinePlugins.standalonePlugins`:
```
> error: collision between `/nix/store/71m5cmbwqbdhyriphml046qq7299s8d7-vimplugin-hmts.nvim-2024-05-07/queries/nix/injections.scm' and `/nix/store/4l7v3q9wlkqjfmrzy6m9k5qpb77b4c74-vimplugin-treesitter-grammar-nix/queries/nix/injections.scm'
```


plugins/languages/treesitter/nixvim-injections.scm
MattSturgeon, line 1: We don't need to rename the file in our repo to affect the name in the /nix/store.

Unless you just think this is a better name?

tests/test-sources/plugins/languages/treesitter/combine-plugins.nix
```diff
@@ -3,12 +3,7 @@
   combine-plugins = {
     performance.combinePlugins.enable = true;
 
-    plugins.treesitter = {
-      enable = true;
-
-      # Exclude nixvim injections for test to pass
-      nixvimInjections = false;
-    };
+    plugins.treesitter.enable = true;
```
MattSturgeon, line 6: Why change the test?

plugins/languages/treesitter/nixvim-injections.scm
GaetanLepage: That's true indeed. Even though, I find this name to be clearer. If it's OK for you we can merge this as is.

GaetanLepage:
I still get a clash between the `injections.scm` file from the `hmts` plugin and the one from the `nix` treesitter grammar.

GaetanLepage:
And I fear that this PR breaks the injection...

khaneliman:
Can we just use `vim.treesitter.query.set("nix", "injections", "(comment) @comment")`

stasjok:
I don't get it. Why do you change nixvim injections, if based on your error you have entirely different problem. You can't have collisions with filesPlugin (what you are trying to change here), because I excluded it. You have collisions with this plugin: https://github.com/calops/hmts.nvim. So you can add "hmts.nvim" to standalonePlugins.

GaetanLepage:
> I don't get it. Why do you change nixvim injections, if based on your error you have entirely different problem. You can't have collisions with files added via `extraFiles` (what you are trying to change here), because I excluded it. You have collisions with this plugin: https://github.com/calops/hmts.nvim. So you can add "hmts.nvim" to standalonePlugins.

You are right, this doesn't address my issue at all.
Closing.

PR #1948, opened 2024-07-28
lib/helpers: build recursively

MattSturgeon:
This ended up being a fairly small change, but the result is much cleaner.

Instead of naming each part of `helpers` and passing those sections to the other parts that need them, we can build `helpers` in a let-block and pass `helpers` itself to parts of `helpers` while building `helpers`.

I also made use of `callPackageWith autoArgs` to clean up some of the repetitive boilerplate.

### DIFF
--- a/lib/autocmd-helpers.nix
+++ b/lib/autocmd-helpers.nix
@@ -1,8 +1,4 @@
-{
-  lib,
-  nixvimOptions,
-  nixvimTypes,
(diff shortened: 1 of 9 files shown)


GaetanLepage (approved):
This is indeed much cleaner !

MattSturgeon:
@mergifyio queue

PR #1949, opened 2024-07-28
otter: ensures no warning is shown when running tests

GaetanLepage:
- **tests/cmp-all-sources: disable the otter source as it triggers a warning when treesitter is missing**
- **tests/otter: add treesitter to avoid warning**

### DIFF
--- a/tests/test-sources/plugins/languages/otter.nix
+++ b/tests/test-sources/plugins/languages/otter.nix
@@ -1,38 +1,47 @@
 {
   empty = {
-    plugins.otter.enable = true;
+    plugins = {
+      otter.enable = true;
+      # Avoid the warning
+      treesitter.settings.highlight.enable = true;
+    };
   };
 
   defaults = {
-    plugins.otter = {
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/languages/otter.nix
```diff
@@ -1,38 +1,51 @@
 {
   empty = {
-    plugins.otter.enable = true;
+    plugins = {
+      otter.enable = true;
+      treesitter = {
+        enable = true;
+        settings.highlight.enable = true;
```
MattSturgeon, line 7: `plugins.treesitter.enable` is fine, but not needed to avoid the warning or to run the test.

```suggestion
        # Avoid the warning
        settings.highlight.enable = true;
```

GaetanLepage: Good point ! Done

MattSturgeon (approved):

GaetanLepage:
@Mergifyio queue

PR #1952, opened 2024-07-30
plugins/spectre: set freeformType for engine options

elohmeier:
This adds the `warn` engine option to be able to silence messages when an executable is missing from the PATH. See https://github.com/nvim-pack/nvim-spectre/pull/242 for the upstream change.

### DIFF
--- a/plugins/utils/spectre.nix
+++ b/plugins/utils/spectre.nix
@@ -30,6 +30,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
           (
             with types;
             attrsOf (submodule {
+              freeformType = with types; attrsOf anything;
               options = {
                 cmd = mkOption {
                   type = types.str;


MattSturgeon:
Thanks for you contribution! However, due to `settings` being a "freeform" option this change shouldn't be needed in order to define the corresponding config.

If you're unable to define the config then that means we have a bigger issue such as a sub-option blocking access to the freeformType.

elohmeier:
I've updated this PR to set the missing freeformType option instead. The above test now works.

plugins/utils/spectre.nix
```diff
@@ -30,6 +30,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
           (
             with types;
             attrsOf (submodule {
+              freeformType = types.attrs;
```
MattSturgeon, line 33:
```suggestion
              freeformType = with types; attrsOf anything;
```

This isn't strictly necessary if we know that "engine" options will always be a **flat** attrset, but the `anything` type allows for **recursive** merging of definitions, while defining nested attrs in multiple locations could clobber each other when using `types.attrs`.

I'll let you decide which type is more appropriate here, `attrs` or `attrsOf anything`.

MattSturgeon (commented):
Much better! Small discussion point that may or may not need changing, then I think we're good to merge.

plugins/utils/spectre.nix
traxys: I'd rather go with attrsOf anything, as `attrs` is soft-deprecated in nixpkgs I believe

elohmeier: Thanks, updated!

MattSturgeon:
There's a few tests failing (even after I tried a rebuild).

Struggling to read the logs on my phone though, so you may have to investigate on your own for now.

MattSturgeon (approved):
Seems tests all good now!

Did you want to rebase and get full commit attribution?

GaetanLepage:
@mergifyio queue

elohmeier:
Great, thanks. I don't want to rebase - let's get it in!

GaetanLepage:
@mergifyio refresh

GaetanLepage:
@mergifyio queue

GaetanLepage:
@mergifyio refresh

GaetanLepage:

PR #1956, opened 2024-07-30
flake/generate-files: move most assertions to `check` derivation

MattSturgeon:
Move the none-ls and efmls assertions out of the generation scripts and into a separate "test" derivation run by `nix flake check` and buildbot.

I didn't move the rust-analyzer assertions since they seem more related to the generation script itself, rather than the output it produces.

I've also made some other improvements to the generation script, such as using nixfmt directly and passing all derivations in via string context instead of calling `nix build` multiple times.

I'll probably go over this again in the daytime since I think the test derivation can be simplified further.


MattSturgeon:
What commit prefix should we be using for the update/gen-files stuff? It seems we've been [flake/generate-files:](https://github.com/nix-community/nixvim/commits/ea33814a1fe5bfe8e99ca89d805d86ac4a0b4bdc/flake-modules/updates/default.nix)

GaetanLepage (approved):
It looks good to me at this point.

For the commit prefix, I would go with flake/generate

MattSturgeon:
@GaetanLepage I've dismissed your approval since I've updated the implementation. Thanks for your early review though!

I've also tested that the update action is able to run on this branch: https://github.com/MattSturgeon/nixvim/actions/runs/10178853612

I've tested the test itself by adding/removing random entries from the none-ls/efmls package declaration files, and everything seems correct:

```
none-ls: The following are not listed upstream, but are declared in plugins/none-ls/packages.nix:
- foo
- bar

efmls: The following are not declared in efmls-configs-pkgs.nix:
- foobar

efmls: The following are not listed upstream, but are declared in efmls-configs-pkgs.nix:
- foobarbaz
- prettierd
- write-good

```

GaetanLepage (approved):
Still good to me !

MattSturgeon:
@mergifyio queue

PR #1959, opened 2024-07-31
flake/generate-files: fix `--commit` line count

MattSturgeon:
`echo` pipes all its output at once, so `wc -l` always counts 1 line.

See https://stackoverflow.com/questions/60954221


MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1962, opened 2024-08-01
lib/helpers: scope all subsections, but keep top-level aliases

MattSturgeon:
This is a small step towards later refactoring our custom `lib`, but doesn't yet make any breaking changes.

All subsections within `helpers` are now available as scoped (nested) attrs, e.g. `helpers.options.defaultNullOpts` or `helpers.utils.mkRaw`.

Everything that was previously accessible from the "top-level" of `helpers` still is, via aliases defined in `helpers.nix`.
In a future PR we should add `lib.warn` to some of those aliases, before later dropping them altogether.
We can discuss in the future which aliases should be deprecated/dropped and if anything else should be moved/renamed.

### DIFF
--- a/lib/helpers.nix
+++ b/lib/helpers.nix
@@ -9,20 +9,76 @@ let
   call = lib.callPackageWith { inherit pkgs lib helpers; };
 
   # Build helpers recursively
-  helpers =
-    {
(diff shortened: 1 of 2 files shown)


traxys (approved):

MattSturgeon:
@mergifyio queue

PR #1963, opened 2024-08-01
lib: provide an "extended lib" to our modules

MattSturgeon:
lib/extend-lib.nix returns a nixpkg's `lib`, extended with our own helpers available as `lib.nixvim`.

The extended lib also merges our maintainer list into `lib.maintainers`.

This is exposed as `helpers.extendedLib`, but when evaluating our modules it should be assigned to `specialArgs.lib`.

Outside of our modules you must still access our helpers via `config.lib.nixvim` or `config.lib.nixvim.extendedLib`.

Within helpers' sub-sections, `lib` is the extended lib.

I added a test, but currently it is only run with a standalone build. For completeness it'd be nice to also run the test with out other module wrappers, but I'm not sure the best way to go about this.

This PR makes **no restructuring changes** to the lib's layout, as discussed in the doc. #1962 is a first step towards doing that, but also makes no breaking changes.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -40,8 +40,6 @@ let
     pkgs = pkgsDoc;
   };
 
-  inherit (helpers.modules) specialArgs;
-
   nixvimPath = toString ./..;
 
   gitHubDeclaration = user: repo: branch: subpath: {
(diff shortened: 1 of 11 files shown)


MattSturgeon:
Currently this is just an alternative way to access `helpers`, however I think new code should prefer using `lib.nixvim`.

Should this PR update documentation and CONTRIBUTING or should we do that after some incubation time?

If we ever get around to migrating all internal usage of `helpers` to `lib.nixvim` then perhaps we can add a deprecation warning to `helpers`. I don't think this should be a priority though.

traxys (approved):

traxys:
I think we should indeed change the mentions from `helpers` to `lib`, as it is more inline with what the other projects are doing, like HM.
Refactoring the entire codebase to get rid of `helpers` should not be _too_ hard, we already did it once (twice maybe?) to refactor how they were imported

GaetanLepage:
Changes look good to me, and I am fine with the change.
Thanks !

GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
Thank you both for your reviews!

> I think we should indeed change the mentions from `helpers` to `lib`, as it is more inline with what the other projects are doing, like HM.

Agreed! As well as updating the docs, there's probably other minor polishing to be done in follow up PRs. E.g. we may not need to explicitly set `specialArgs.lib` if we use `extendedLib.evalModules`.

> Refactoring the entire codebase to get rid of `helpers` should not be _too_ hard, we already did it once (twice maybe?) to refactor how they were imported

If we can do it without much effort that's great! I just didn't want migration to be a blocker for the initial refactoring.

PR #1966, opened 2024-08-02
wrappers: add our lib to the host's `_module.args`

MattSturgeon:
Make our custom "helpers" lib available to host modules as `_module.args.nixvimLib`, as an alternative to accessing via `config.lib`.

We didn't do this before because `helpers` was too generic of a name, but `nixvimLib` should be ok.

e.g:

```nix
{ lib, nixvimLib, ... }:
{
  programs.nixvim.someOption = nixvimLib.mkRaw "print('Hi')";
}
```

Should we consider also making the alias available to nixvim modules? Can always be done later.

## Update

I've made it so that the `nixvimLib` arg in the host modules is equivalent to the `lib` arg in the nixvim modules; i.e. nixvim's "extended" nixpkgs lib.

### DIFF
--- a/docs/user-guide/helpers.md
+++ b/docs/user-guide/helpers.md
@@ -23,6 +23,18 @@ in
 }
 ```
 
+Or you can access the extended `lib` used in standalone builds via the `nixvimLib` module arg:
+
+```nix
+{ nixvimLib, ... }:
+{
+  # You can use nixvimLib.nixvim in your config
+  fooOption = nixvimLib.nixvim.mkRaw "print('hello')";
+}
+```
+
+This "extended" lib, includes everything normally in `lib`, along with some additions from nixvim.
+
 **Note:** the `lib` argument passed to modules is entirely unrelated to the `lib` _option_ accessed as `config.lib`!
 
 A certain number of helpers are defined that can be useful:
--- a/docs/user-guide/helpers.md
+++ b/docs/user-guide/helpers.md
@@ -1,16 +1,17 @@
 # Helpers
 
-If Nixvim is built using the standalone method, you can access our `helpers` as a module arg:
+If Nixvim is built using the standalone method, you can access our "helpers" as part of the `lib` module arg:
 
 ```nix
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Another approach here would be to assign `_module.args.nixvimLib = config.lib.nixvim.extendedLib`, allowing end-users to use `nixvimLib` as a `lib` alternative in their modules:

```nix
{ nixvimLib, ... }:
let
  lib = nixvimLib;
in
{
  programs.nixvim = {
    enable = true;
    foo.bar = lib.nixvim.mkRaw "print('Hello, world!')";
  };
}
```

traxys:
I think we should deprecate helpers in the same PR where we introduce the `nixvimLib`, as I think it's a much better name!

MattSturgeon:
> I think we should deprecate helpers in the same PR where we introduce the `nixvimLib`, as I think it's a much better name!

I don't think we're _ready_ to deprecate helpers, because that requires a treewide migration to ensure we aren't triggering our own deprecation warnings...

Note also, that this `nixvimLib` is only made available to the _host_ module args, not to nixvim modules. The host modules have never had access to the `helpers` arg, only to `config.lib.nixvim` and (historically) `config.nixvim.helpers`.

MattSturgeon:
Also, I forgot to update the PR description, but I ended up implementing my suggestion from https://github.com/nix-community/nixvim/pull/1966#issuecomment-2265542120.

I'll update the docs to reflect the subtle difference when I rebase.

khaneliman (approved):
If the scope of this PR is just granting access to the nixvim lib at a higher level for users, that makes sense. I agree we'd want a larger PR for handling deprecation and migrating internal config if https://github.com/nix-community/nixvim/pull/1989 is going to block PRs when we have warnings.

MattSturgeon:
@khaneliman @traxys I've rebased and updated the changes to the user-guide.

khaneliman (approved):
LGTM

PR #1974, opened 2024-08-03
plugins/gitblame: switch to mkNeovimPlugin

GaetanLepage:


MattSturgeon:
Can't see any glaring issues. Can we try out using `lib.nixvim` instead of `helpers`?

PR #1975, opened 2024-08-03
plugins/gitblame: switch to mkNeovimPlugin

GaetanLepage:


GaetanLepage:
I had to re-create a PR because I had originally used a branch on the main repo (by mistake) and thus couldn't force-push on it...

`helpers` -> `lib.nixvim` seems to work great !

plugins/git/gitblame.nix
```diff
@@ -6,66 +6,133 @@
...
-  config =
-    let
-      setupOptions = {
-        enabled = cfg.enable;
-        message_template = cfg.messageTemplate;
-        date_format = cfg.dateFormat;
-        message_when_not_committed = cfg.messageWhenNotCommitted;
-        highlight_group = cfg.highlightGroup;
-        display_virtual_text = helpers.ifNonNull' cfg.displayVirtualText (
-          if cfg.displayVirtualText then 1 else 0
```
MattSturgeon, line 54: This used to be defined as a bool, but transformed into `enum [ 0 1 ]`. Looks like the new option is still a bool; is that correct?

plugins/git/gitblame.nix
```diff
@@ -6,66 +6,133 @@
...
+    ];
+    delay = 1000;
+    virtual_text_column = 80;
+    use_blame_commit_file_urls = true;
   };
 
-  config =
-    let
-      setupOptions = {
-        enabled = cfg.enable;
```
MattSturgeon, line 48: Looks like we used to set `settings.enabled` by default. Is there a way we can warn that this is no longer the case?

plugins/git/gitblame.nix
GaetanLepage: Yes, the upstream option (in lua) is indeed `bool`: https://github.com/f-person/git-blame.nvim/blob/master/lua/gitblame/config.lua#L11

plugins/git/gitblame.nix
GaetanLepage: Anyway, the default behavior is still `true`. Now the user has the option to explicitly disable it.

plugins/git/gitblame.nix
```diff
@@ -6,66 +6,133 @@
...
-      '';
+  extraOptions = {
+    gitPackage = nixvim.mkPackageOption {
+      name = "git";
+      default = pkgs.git;
     };
+  };
+
+  extraConfig = cfg: { extraPackages = [ cfg.gitPackage ]; };
+
```
MattSturgeon, line 137:
```suggestion
```

traxys:
Could we try removing the `with lib;` too?

MattSturgeon:
> Could we try removing the `with lib;` too?

I agree this may be nice going forward.

Although, in practice `with lib` is unlikely to cause us major issues?

If we only need a handful of things from `lib`, maybe do `inherit (lib) x y z`?

GaetanLepage:
I made an attempt, please, let me hear your feedback.

plugins/git/gitblame.nix
```diff
@@ -5,67 +5,135 @@
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.gitblame;
+  inherit (lib.nixvim) defaultNullOpts;
```
MattSturgeon, line 9: `inherit (lib) types` would also be ok IMO. Or even `types = lib.nixvim.nixvimTypes`?

Off topic: should we consider merging our types into `lib.types` (like we now do for maintainers)? Or is it better to keep our types under the `lib.nixvim` namespace?

MattSturgeon (approved):
No issues from me, but I'm out and about; reading from my phone screen.

Since it'd be easy for me to miss something maybe it's best if @traxys also review?

plugins/git/gitblame.nix
GaetanLepage: > `inherit (lib) types` would also be ok IMO. Or even `types = lib.nixvim.nixvimTypes`?
> 
> Off topic: should we consider merging our types into `lib.types` (like we now do for maintainers)? Or is it better to keep our types under the `lib.nixvim` namespace?

Why not mixing them if the two sets don't clash.

MattSturgeon: The only reason I can think of is to be explicit about what is our custom stuff and what is from nixpkg's lib.

How important that is is up for debate. I think my preference would be to merge our types into `lib.types`.

GaetanLepage:
Should we merge this then ?

MattSturgeon:
> Should we merge this then ?

👍 from me

plugins/git/gitblame.nix
MattSturgeon: See #1978

GaetanLepage:
@mergifyio queue

PR #1978, opened 2024-08-03
lib/types: merge into `extendedLib`

MattSturgeon:
- **lib/types: merge into `extendedLib`**
- **lib: migrate `helpers.nixvimTypes` -> `lib.types`**

As discussed in https://github.com/nix-community/nixvim/pull/1975#discussion_r1702774343

I think this is the cleanest way for us to have custom option types, however we do loose some clarity. You can no longer tell at a glance whether a type is from nixvim or nixpkgs.

@GaetanLepage seems on board, what's your view @traxys?

PR #1982, opened 2024-08-04
plugins/TEMPLATE: drop helpers and adopt new style

GaetanLepage:
Closes #2137

### DIFF
--- a/plugins/TEMPLATE.nix
+++ b/plugins/TEMPLATE.nix
@@ -1,11 +1,13 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-helpers.neovim-plugin.mkNeovimPlugin config {
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
   name = "my-plugin";
   originalName = "my-plugin.nvim"; # TODO replace (or remove entirely if it is the same as `name`)
   defaultPackage = pkgs.vimPlugins.my-plugin-nvim; # TODO replace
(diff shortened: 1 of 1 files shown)


plugins/TEMPLATE.nix
```diff
@@ -14,11 +16,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   # Optionally, explicitly declare some options. You don't have to.
```
MattSturgeon, line 17: Can we update this comment to stress that we actually prefer _not_ including options unless they are **very** important?

MattSturgeon (approved):
Great thanks!

plugins/TEMPLATE.nix
GaetanLepage: We could. Have we actually settled on the philosophy going forward ?

MattSturgeon: > Have we actually settled on the philosophy going forward ?

Not sure. Let's merge this as is for now and try to hash out the other details in another PR/issue/discussion/doc/etc

plugins/TEMPLATE.nix
```diff
@@ -1,11 +1,13 @@
...
-  helpers,
   config,
   pkgs,
   ...
 }:
-helpers.neovim-plugin.mkNeovimPlugin config {
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
```
MattSturgeon, line 10: Do we want to also demonstrate `types = lib.nixvim.nixvimTypes`?

#1978 can be responsible for updating this line to `inherit (lib) types`

GaetanLepage: We should wait for this indeed.

plugins/TEMPLATE.nix
MattSturgeon, line 1: We should probably demonstrate using `description` in the template too, with a comment saying it can be used to summarise the plugin, and also document any quirks (etc)

MattSturgeon:
@mergifyio queue

PR #1983, opened 2024-08-04
plugins/firenvim: init

### LINKED ISSUE #1713: [PLUGIN REQUEST] Firenvim
| Field    | Description |
|----------|-------------|
| Plugin   | [firenvim](https://github.com/glacambre/firenvim)  | 
| Nixpkgs  | [pkg](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=firenvim) | 

## Extra Information

The plugin allows us to run our neovim config in our browser text area.
afaik it fetches the data from nvim's config dir. 

```bash 

Idk this is in the roadmap or not.
I guess this will require some input from firenvim dev too and may require some changes at the plugin end also.
```

GaetanLepage:
Add support for [firenvim](https://github.com/glacambre/firenvim).

Fixes #1713

### DIFF
--- a/plugins/utils/firenvim.nix
+++ b/plugins/utils/firenvim.nix
@@ -1,0 +1,89 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  types = lib.nixvim.nixvimTypes;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "firenvim";
+  defaultPackage = pkgs.vimPlugins.firenvim;
+
+  maintainers = with lib.maintainers; [ GaetanLepage ];
+
+  settingsOptions = {
+    globalSettings = defaultNullOpts.mkAttrsOf' {
+      type = types.anything;
+      pluginDefault = { };
+      example = {
+        alt = "all";
+      };
+      description = "Default settings that apply to all URLs.";
+    };
+
+    localSettings = defaultNullOpts.mkAttrsOf' {
+      type = with types; attrsOf anything;
+      pluginDefault = { };
+      example = {
+        ".*" = {
+          cmdline = "neovim";
+          content = "text";
+          priority = 0;
+          selector = "textarea";
+          takeover = "always";
+        };
+        "https?://[^/]+\\.co\\.uk/" = {
+          takeover = "never";
+          priority = 1;
+        };
+      };
+      description = ''
+        Map regular expressions that are tested against the full URL to settings that are used for
+        all URLs matched by that pattern.
+
+        When multiple patterns match a URL, the pattern with the highest "priority" value is used.
+
+        Note: regular expressions use the [JavaScript flavor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions).
+      '';
+    };
+  };
+
+  settingsExample = {
+    globalSettings.alt = "all";
+    localSettings = {
+      ".*" = {
+        cmdline = "neovim";
+        content = "text";
+        priority = 0;
+        selector = "textarea";
+        takeover = "always";
+      };
+      "https?://[^/]+\\.co\\.uk/" = {
+        takeover = "never";
+        priority = 1;
+      };
+    };
+  };
+
+  settingsDescription = ''
+    Options fed to the `vim.g.firenvim_config` table.
+  '';
+
+  callSetup = false;
+  extraConfig = cfg: {
+    warnings =
+      lib.optional
+        (
+          config.performance.combinePlugins.enable
+          && !(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins)
+        )
+        ''
+          Nixvim (plugins.firenvim): Using `performance.combinePlugins` breaks `firenvim`.
+          Add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
+        '';
+    globals.firenvim_config = lib.modules.mkAliasAndWrapDefsWithPriority lib.id cfg.settings;
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -167,6 +167,7 @@
     ./utils/easyescape.nix
     ./utils/emmet.nix
     ./utils/endwise.nix
+    ./utils/firenvim.nix
     ./utils/flash.nix
     ./utils/floaterm.nix
     ./utils/fzf-lua.nix
--- a/tests/test-sources/plugins/utils/firenvim.nix
+++ b/tests/test-sources/plugins/utils/firenvim.nix
@@ -1,0 +1,28 @@
+{
+  empty = {
+    plugins.firenvim.enable = true;
+  };
+
+  example = {
+    plugins.firenvim = {
+      enable = true;
+
+      settings = {
+        globalSettings.alt = "all";
+        localSettings = {
+          ".*" = {
+            cmdline = "neovim";
+            content = "text";
+            priority = 0;
+            selector = "textarea";
+            takeover = "always";
+          };
+          "https?://[^/]+\\.co\\.uk/" = {
+            takeover = "never";
+            priority = 1;
+          };
+        };
+      };
+    };
+  };
+}


plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+      lib.optional
+        (
+          config.performance.combinePlugins.enable
+          && (!(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins))
+        )
+        ''
+          Nixvim (plugins.firenvim): Using `performance.combinePlugins` breaks `firenvim`.
+          You add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
+        '';
+    globals.firenvim_config = cfg.settings;
```
MattSturgeon, line 84: Could use [`mkAliasAndWrapDefsWithPriority`](https://noogle.dev/f/lib/modules/mkAliasAndWrapDefsWithPriority) here

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+  extraConfig = cfg: {
+    warnings =
+      lib.optional
+        (
+          config.performance.combinePlugins.enable
+          && (!(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins))
+        )
+        ''
+          Nixvim (plugins.firenvim): Using `performance.combinePlugins` breaks `firenvim`.
+          You add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
```
MattSturgeon, line 82:
```suggestion
          Add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
```

Or
```suggestion
          You should add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
```

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+    Options fed to the `vim.g.firenvim_config` table.
+  '';
+
+  callSetup = false;
+  extraConfig = cfg: {
+    warnings =
+      lib.optional
+        (
+          config.performance.combinePlugins.enable
+          && (!(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins))
```
MattSturgeon, line 78:
```suggestion
          && !(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins)
```

This should be fine, right? _Probably_ more readable without `()` spam, though it is subjective.

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+      };
+      "https?://[^/]+\\.co\\.uk/" = {
+        takeover = "never";
+        priority = 1;
+      };
+    };
+  };
+
+  settingsDescription = ''
+    Options fed to the `vim.g.firenvim_config` table.
```
MattSturgeon, line 69: With this and lz-n, this is becoming a pattern. We should consider having a variant of or option in `mkVimPlugin` to handle these style of plugin configs.

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+          priority = 1;
+        };
+      };
+      description = ''
+        Map Javascript patterns that match against the full URL to settings that are used for all
+        URLs matched by that pattern.
+        When multiple patterns match a URL, the pattern with the highest "priority" value is used.
+      '';
+    };
+  };
```
MattSturgeon, line 49: Looks good overall!

plugins/utils/firenvim.nix
GaetanLepage: Yes, we could.
At the same time, adding yet another abstraction for such few occurrences does not seem to urgent to me.
For sure we can consider it a bit later.

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+        };
+        "https?://[^/]+\\.co\\.uk/" = {
+          takeover = "never";
+          priority = 1;
+        };
+      };
+      description = ''
+        Map Javascript patterns that match against the full URL to settings that are used for all
+        URLs matched by that pattern.
+        When multiple patterns match a URL, the pattern with the highest "priority" value is used.
```
MattSturgeon, line 46:
```suggestion
        Map regular expressions that are tested against the full URL to settings that
        are used for all URLs matched by that pattern.

        When multiple patterns match a URL, the pattern with the highest "priority" value is used.

        Note: regular expressions use the [JavaScript flavor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions).
```

plugins/utils/firenvim.nix
GaetanLepage: Sure, let's try !

plugins/utils/firenvim.nix
GaetanLepage: I wasn't aware of that beast !

plugins/utils/firenvim.nix
MattSturgeon: Yes, wasn't meant to be addressed urgently. Was just thinking if we keep coming across this style of settings.

plugins/utils/firenvim.nix
```diff
@@ -0,0 +1,86 @@
...
+      lib.optional
+        (
+          config.performance.combinePlugins.enable
+          && !(lib.elem "firenvim" config.performance.combinePlugins.standalonePlugins)
+        )
+        ''
+          Nixvim (plugins.firenvim): Using `performance.combinePlugins` breaks `firenvim`.
+          Add this plugin to `performance.combinePlugins.standalonePlugins` to prevent any issue.
+        '';
+    globals.firenvim_config = lib.mkAliasAndWrapDefsWithPriority cfg.settings;
```
MattSturgeon, line 84:
```suggestion
    globals.firenvim_config = lib.mkAliasAndWrapDefsWithPriority lib.id cfg.settings;
```

MattSturgeon (approved):
LGTM, thanks!

GaetanLepage:
@mergifyio queue

niksingh710:
I had to manually run the command `:call firenvim#install(0)`.
Is that an intended behavior from nixvim?

traxys:
I think this can't really be done by nixvim itself, as writes to the user home directory. We should doocument it though

niksingh710:
yeah, Add it to the documentation.
I was thinking of an overkill way that would have created a file in `.cache` and based on the existence of that file, It runs the command of installation.
But I don't think that's a clean way to perform installation, as it should be managed by the plugin only.

GaetanLepage:
> I had to manually run the command `:call firenvim#install(0)`.
> Is that an intended behavior from nixvim?

Yes, I tested the plugin this way.
Sure, documenting it would be nice.

niksingh710:
I am trying `:lua print(vim.g.firenvim_config)` and this is giving a `nil` output to me.

my nixvim config is 
```nix
{
  plugins.firenvim = {
    enable = true;
    settings = {
      globalSettings = {
        cmdline = "neovim";
        content = "text";
        priority = 0;
        selector = "textarea";
        takeover = "never";
      };
      localSettings = { ".*" = { takeover = "never"; }; };
    };
  };
}```

**Comment by MattSturgeon** [comment]
> I am trying `:lua print(vim.g.firenvim_config)` and this is giving a `nil` output to me.

Have you re-built your config? Can you verify by running `nixvim-print-init`?

**Comment by niksingh710** [comment]
> Have you re-built your config? Can you verify by running `nixvim-print-init`?

Yep, I have updated my flakes and tested it on a built version of nixvim and also by doing `nix run`. Same results.

`nixvim-print-init` does not contain anything regarding `firenvim`. (have done a regex search)

**Comment by niksingh710** [comment]
@MattSturgeon are you sure about the command `nix run . -- nixvim-print-init` i am running it in my nixvim Dir, and it is opening a blank buffer with the name of `nixvim-print-init`.

In my opinion it should do like this coz after `--` the string is a buffer name for nvim.

Yeah, I can confirm that my file is being loaded coz it errors out if i misspell `firenvim`.

**Comment by MattSturgeon** [comment]
> @MattSturgeon are you sure about the command `nix run . -- nixvim-print-init` i am running it in my nixvim Dir, and it is opening a blank buffer with the name of `nixvim-print-init`.

My mistake, I meant `nix shell . --command "nixvim-print-init"`

> Yeah, I can confirm that my file is being loaded coz it errors out if i misspell `firenvim`.

Ok thanks for checking!

**Comment by niksingh710** [comment]
yeah, running `nix shell . --command "nixvim-print-init"` and then performing a search via `/` results nothing.

**Comment by MattSturgeon** [comment]
Hm, I can reproduce that.

When I try inspecting `options.globals` in the repl, I get an error:

```
nix-repl> :p packages.x86_64-linux.nvim.options.globals.definitions
[ { } { } { } { } { } { enableRemoteUrlRewrites = null; filetypes = null; git = { commits = { format = null; limit = null; sort_by = null; }; }; github = { hosts = null; issues = { fields = null; filter = null; format = null; limit = null; sort_by = null; state = null; }; mentions = { format = null; limit = null; sort_by = null; }; pull_requests = { fields = null; format = null; limit = null; sort_by = null; state = null; }; }; gitlab = { hosts = null; issues = { format = null; limit = null; sort_by = null; state = null; }; mentions = { format = null; limit = null; sort_by = null; }; merge_requests = { format = null; limit = null; sort_by = null; state = null; }; }; remotes = null; trigger_actions = null; } { } { } { } { } { tmux_navigator_disable_when_zoomed = null; tmux_navigator_no_mappings = null; tmux_navigator_no_wrap = null; tmux_navigator_preserve_zoom = null; tmux_navigator_save_on_switch = null; } { sleuth_heuristics = null; sleuth_no_filetype_indent_on = null; } { firenvim_config = { _type = "if"; condition = false; content = { _type = "merge"; contents = error:
       … while calling the 'map' builtin

         at /nix/store/4cpakzyvfw1rmm9v5i3387x6jd2h1v86-source/lib/modules.nix:1081:22:

         1080|       prio = option.highestPrio or defaultOverridePriority;
         1081|       defsWithPrio = map (mkOverride prio) option.definitions;
             |                      ^
         1082|     in mkAliasIfDef option (wrap (mkMerge defsWithPrio));

       error: attribute 'definitions' missing

       at /nix/store/4cpakzyvfw1rmm9v5i3387x6jd2h1v86-source/lib/modules.nix:1081:44:

         1080|       prio = option.highestPrio or defaultOverridePriority;
         1081|       defsWithPrio = map (mkOverride prio) option.definitions;
             |                                            ^
         1082|     in mkAliasIfDef option (wrap (mkMerge defsWithPrio));
```

I wonder if it's caused by `mkAliasAndWrapDefsWithPriority` being unable to handle freeform submodules? I can print `options.globals.definitions` just fine when firenvim isn't enabled.

This prints fine though:

```
nix-repl> :p packages.x86_64-linux.nvim.options.plugins.firenvim.settings.definitions
[ { globalSettings = { alt = "all"; }; localSettings = { ".*" = { cmdline = "neovim"; content = "text"; priority = 0; selector = "textarea"; takeover = "always"; }; }; } ]
```

Perhaps it'd be simpler to just use `lib.derivedConfig` instead of the more complex `mkAliasAndWrapDefsWithPriority`. The former uses the highest priority for the option overall, while the later tries to copy the priority used in each specific definition.

MattSturgeon:
The issue was simply down to a minor oversight that I missed in the review, and for some reason the tests didn't pick up on: to make an alias you need the option from `options` not the final value from `config`.

#1997 resolves the issue.

PR #1985, opened 2024-08-04
lib/modules: add `evalNixvim`

MattSturgeon:
Evaluates a nixvim config, by default also checking warnings and assertions.

This used internally by the standalone wrapper.

I'm planning to use this to refactor the tests implementation:
- The tests don't need a "full" standalone build
- The tests should be able to expect/reject warnings/assertions
- Warnings should fail tests (unless marked as expected)

But I figured it'd be easier to test/review if I push this as a separate PR.

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -16,4 +16,38 @@ rec {
       defaultPkgs = pkgs;
     }
     // extraSpecialArgs;
+
+  # Evaluate nixvim modules, checking warnings and assertions
+  evalNixvim =
+    {
+      modules ? [ ],
+      extraSpecialArgs ? { },
+      # Set to false to disable warnings and assertions
+      # Intended for use with tests, where we may not want to check assertions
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Looking at the implementation in https://github.com/nix-community/nixvim/pull/1986, I'm wondering if the interface here is a little over-engineered?

I think the only place where we'd use it with the default values would be the standalone wrapper :face_with_diagonal_mouth:

MattSturgeon:
> Looking at the implementations in #1986 and #1989, I'm wondering if the interface here is a little over-engineered?

This is resolved in the latest force-push, however I still think we can simplify things further by moving the warning/assertion handling to `finalPackage`'s evaluation; similar to how nixos checks warnings/assertions while evaluating the activation package option.

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #1986, opened 2024-08-04
tests: fail on warnings and minor refactor

MattSturgeon:
- **tests: fail on warnings**
- **tests: make `dontRun` a config option**

This PR refactors things so that warnings and assertions are explicitly checked by the test functions.

The biggest benefit is that it will now be more obvious when tests produce warnings.

Additional benefits include:
- Handle assertions in a derivation, instead of during eval
- `tests.dontRun` is now a module option, simplifying the implementation

In the future we can add additional `tests.*` options, such as "expect" options, assertion/warning whitelists, etc.

Marked as a draft because I'm too tired to test this and ensure it is in a good working state, but code reviews are still welcome!

### DIFF
--- a/tests/test-derivation.nix
+++ b/tests/test-derivation.nix
@@ -1,7 +1,6 @@
 {
   pkgs,
   lib ? pkgs.lib,
-  makeNixvimWithModule,
   ...
 }:
 let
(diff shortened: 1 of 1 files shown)


tests/test-derivation.nix
```diff
@@ -79,20 +78,60 @@ let
...
-            inherit pkgs module extraSpecialArgs;
-            _nixvimTests = true;
+        }@args:
+        let
+          result = lib.evalModules {
+            modules = [
+              module
+              ./_module.nix
+              ../modules/top-level
+            ] ++ lib.optional (args ? dontRun) { tests.dontRun = lib.mkDefault dontRun; };
```
MattSturgeon, line 99: It'd be simpler to drop the `dontTest` attr in moduleList. If we keep the function arg/attr then we need to pass it in to the module config here using `mkDefault`.

@traxys is dropping the attr/fn-arg a breaking change? I assume yes, but probably not an important one?

It'd be nice to have `dontRun` _only_ be a module option (other than in `mkTestDerivationFromNvim`)

tests/test-derivation.nix
```diff
@@ -79,20 +78,60 @@ let
...
+            fi
+            if [ -n "$warnings" ]; then
+              echo "Unexpected warnings:"
+              for it in "$warnings"; do
+                echo "- $it"
+              done
+              echo
+            fi
+            exit 1
+          '';
```
MattSturgeon, line 127: Should this be done on a per-module basis or on a per moduleList basis?

If the latter, how should we handle some modules having errors and others needing to be build/run?

PR #1987, opened 2024-08-05
plugins/luasnip: add filetypeExtend option

### LINKED ISSUE #1967: [BUG] luasnip filetype_extend
| Plugin       | `luasnip`  | 
| Nixpkgs      | `unstable` | 

## Description
Not entirely sure if this is a bug, missing syntax / config, or not important. 

In order to get access to e.g. standardised code comments from `friendly-snippets` it seems like we have to use the `luasnip` function `filetype_extend` (see [here](https://github.com/rafamadriz/friendly-snippets/issues/327), though please correct me if I've misunderstood this). 

## Suggestion
Adding a `filetype_extend` attribute to `plugins.luasnip`, e.g. 
```nix
plugins.luasnip.filetype_extend = {
    "typescript" = ["tsdoc"]:
    "javascript" = ["jsdoc"];
    ...
};
```

## Current workaround
Add the relevant lua lines to `configExtraLua`, e.g. 

```nix
configExtraLua = ''
      require("luasnip").filetype_extend("typescript", { "tsdoc" })
      require("luasnip").filetype_extend("javascript", { "jsdoc" })
      ...
'';
```

GaetanLepage:
Fixes #1967

### DIFF
--- a/plugins/snippets/luasnip/default.nix
+++ b/plugins/snippets/luasnip/default.nix
@@ -133,6 +133,27 @@ in
       '';
       type = types.listOf loaderSubmodule;
     };
+
+    filetypeExtend = mkOption {
+      default = { };
+      type = with types; attrsOf (listOf str);
+      example = {
+        lua = [
+          "c"
+          "cpp"
+        ];
+      };
+      description = ''
+        Wrapper for the `filetype_extend` function.
+        Keys are filetypes (`filetype`) and values are list of filetypes (`["ft1" "ft2" "ft3"]`).
+
+        Tells luasnip that for a buffer with `ft=filetype`, snippets from `extend_filetypes` should
+        be searched as well.
+
+        For example, `filetypeExtend.lua = ["c" "cpp"]` would search and expand c and cpp snippets
+        for lua files.
+      '';
+    };
   };
 
   config =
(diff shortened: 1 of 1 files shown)


plugins/snippets/luasnip/default.nix
```diff
@@ -164,6 +185,11 @@ in
               ''
             ))
           ];
+
+      filetypeExtendConfig = mapAttrsToList (n: v: ''
+        require("luasnip").extend_filetypes("${n}", ${helpers.toLuaObject v})
```
MattSturgeon, line 190: Just a thought, is there a way we could have an attrsOf `<fn_name, settings>` option? That way we support arbitrary functions in a freeform way:

```nix
require('luasnip').${name}(${toLuaObject value})
```

I guess this doesn't work well here because the function takes two arguments.

plugins/snippets/luasnip/default.nix
```diff
@@ -133,6 +133,27 @@ in
       '';
       type = types.listOf loaderSubmodule;
     };
+
+    filetypeExtend = mkOption {
+      default = { };
+      type = with types; attrsOf (listOf str);
```
MattSturgeon, line 139: I expect this is a file-wide issue, but this doesn't support rawLua.

_Ideally_ the whole list or any list item could be rawLua. We can save this for a future work though if the issue applies to other options in this module.

MattSturgeon (approved):
LGTM, thanks!

plugins/snippets/luasnip/default.nix
GaetanLepage: That's a good idea indeed.

GaetanLepage:
@mergifyio queue

GaetanLepage:

PR #1989, opened 2024-08-05
tests: move derivation to module-option & minor cleanup

MattSturgeon:
### Summary

The general idea is that all nixvim config evals will include a "test" package in their output `config` (`config.test.derivation` to be precise).

Taking advantage of `evalNixvim`, added in #1985, the test helpers in `test-derivation.nix` (now `lib/tests.nix`) simply evaluate the modules and return `config.test.derivation`.

### Testing warnigns

This has been moved to #2048 

### Cleanup

Also moved `tests/test-derivation.nix` -> `lib/tests.nix`

Also removed `pkgs.docker-client` from the test-derivation's runtimeInputs. I don't know why it was added, but everything works without it AFAICT.

### TODO

A `evalNixvimTest` (name TBC) may also be helpful, to evaluate without checking assertions and then return the `test` output. Currently, we instead have a `check` argument to `evalNixvim`, when set to false `evalNixvim` will not check assertions/warnings, allowing this to instead be done by the test.

Future tasks that may be worth considering after this PR:
- [ ] check warnings in the test
      #2048 
- [ ] offer `evalNixvimTest` alongside `evalNixvim`?
      Or is the `check` arg ok?
- [ ] Update documentation
- [ ] allow "expecting" and whitelisting assertions/warnings in test modules
- [x] remove usage of `tests.dontRun` and `module =` in test cases
      #2045


modules/top-level/test.nix
```diff
@@ -1,10 +1,90 @@
...
+    };
+
     runNvim = lib.mkOption {
       type = lib.types.bool;
       description = "Whether to run `nvim` in the test.";
       default = true;
     };
+
+    # Output
+    test = lib.mkOption {
```
MattSturgeon, line 25: I'm not a big fan of naming this `config.test.test`...

lib/modules.nix
MattSturgeon, line 1: The changes in _this file_ are from #1985, so they may be easier to review there

MattSturgeon:
Tests failing on warnings seems to be working:
```
withJson> Unexpected warnings:
withJson> - NixVim(plugins.schemastore): you have enabled `json` schemas, but `plugins.lsp.servers.jsonls` is not enabled.
```

modules/top-level/test.nix
traxys: test.derivation?

modules/top-level/test.nix
```diff
@@ -1,10 +1,90 @@
...
+  };
+
+  config = {
+    test.derivation = pkgs.stdenv.mkDerivation {
+      inherit (cfg) name;
+      dontUnpack = true;
+
+      nativeBuildInputs = [
+        config.finalPackage
+        # TODO: Do we really need docker?
```
khaneliman, line 41: Do we know this now?

tests/test-sources/plugins/lsp/schemastore.nix
```diff
@@ -33,7 +36,9 @@
         enable = true;
 
         json = {
-          enable = true;
+          # TODO: re-enable when pkg is fixed and available
+          # https://github.com/NixOS/nixpkgs/pull/335559
+          enable = false;
```
khaneliman, line 41: I, uh, definitely left these warnings for this PR... uh, yeah.

modules/top-level/test.nix
```diff
@@ -1,10 +1,90 @@
...
+          output=$(HOME=$(realpath .) nvim -mn --headless "+q" 2>&1 >/dev/null)
+          if [[ -n $output ]]; then
+            echo "ERROR: $output"
+            exit 1
+          fi
+        '';
+
+      # If we don't do this nix is not happy
+      installPhase = ''
+        touch $out
```
khaneliman, line 86: I was curious about this recently, it seems like we generate the empty out directory for tests. Is there a way we can output the generated config here?

MattSturgeon: The reason we do `mkdir $out` or `touch $out` in tests is that nix considers `$out` not existing to be a build failure.

What we can do instead to make the config (or anything else) available is add extra attrs to the derivation, since a derivation is represented in nix as an attrset.

The idiomatic way to do this would be to define a `passthru` attr

e.g. `passthru.config = result.config`

But I figured this PR was way too big already, without also adding passthru attrs to our tests!

tests/test-sources/plugins/lsp/schemastore.nix
MattSturgeon: Warnings are easily missed in the tests, especially when trusting CI's big green tick.

That's the main motivation for wanting to fail on warnings to begin with :grin:

Maybe we could also have a "ignore all warnings" option too though, which might be better for these kinda temporary disablements?

modules/top-level/test.nix
MattSturgeon: It seems [to build fine](https://buildbot.nix-community.org/#/builders/26/builds/436) without it

khaneliman (approved):
LGTM thanks for testing out removal of docker dependency

tests/test-sources/plugins/lsp/schemastore.nix
khaneliman: I was fine with that being in here, just making a joke about my temporary disables in the flake.lock update leaving warnings. :P

MattSturgeon:
@mergifyio queue

PR #1990, opened 2024-08-05
plugins/which-key: migrate to mkNeovimPlugin

khaneliman:
Working on converting this and there are a few breaking changes for data types of config options https://github.com/folke/which-key.nvim/pull/624/files#diff-a8347a13956b70b044d3e091cd3115e2fcfb44d4ebde516ff03223b2e84c7a7b. 

~- One being the `triggers` option which used to be string or list of strings, but is now a list of attribute sets like the registrations change. What should we do in that case?~
**Very niche setting, continuing with breaking change.**
~- Another issue I was hoping for clarification on is nested properties moving to settings and moving to a different path within settings.~
**Resolved, most can be just parent level rename. Manually specified those that wouldn't be handled properly for nested.**
~- `plugins -> presets -> textObjects` moved and changed from camel to snake case. Does that get handled from the normal `optionsRenamedToSettings`?~
**^**
~- `windows -> winblend` moved to `win -> wo -> winblend`. i'm unable to make a `mkRenamedOption` for that while also using the `optionsRenamedToSettings` for the parent object because of the order of evaluation, it seems. Will I need to explicitly remap all of those or just handle the deprecation like registrations and support forwarding the previous config through, for now?~ 
**Created a deprecation warning for the window setting, dont need to worry about the child setting since it's mentioned in the warning that the options have changed.**

~Lots of TODO notes and some temporary test stuff for testing out renaming and warnings that can be ignored/questions answered~

### DIFF
--- a/plugins/utils/which-key.nix
+++ b/plugins/utils/which-key.nix
@@ -1,29 +1,250 @@
 {
   lib,
-  helpers,
   pkgs,
   config,
   options,
   ...
 }:
 with lib;
-{
-  options.plugins.which-key = {
-    enable = mkEnableOption "which-key.nvim, a plugin that popup with possible key bindings of the command you started typing";
+let
+  inherit (lib.nixvim) defaultNullOpts mkRaw toLuaObject;
+  types = lib.nixvim.nixvimTypes;
 
-    package = helpers.mkPluginPackageOption "which-key-nvim" pkgs.vimPlugins.which-key-nvim;
-
-    registrations = mkOption {
-      type = with types; attrsOf anything;
-      description = ''
-        This option is deprecated, use `settings.spec` instead.
-
-        Note: the keymap format has changed in v3 of which-key.
-      '';
-      visible = false;
+  listOfLen =
+    elemType: len:
+    types.addCheck (types.listOf elemType) (v: length v == len)
+    // {
+      description = "list of ${toString len} ${
+        types.optionDescriptionPhrase (class: class == "noun" || class == "composite") elemType
+      }";
     };
 
-    settings.spec = helpers.defaultNullOpts.mkListOf' {
+  opt = options.plugins.which-key;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "which-key";
+  originalName = "which-key.nvim";
+  defaultPackage = pkgs.vimPlugins.which-key-nvim;
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  # TODO: introduced 2024-08-05: remove after 24.11
+  optionsRenamedToSettings = [
+    "hidden"
+    "icons"
+    "ignoreMissing"
+    "keyLabels"
+    "layout"
+    "motions"
+    "operators"
+    [
+      "plugins"
+      "mark"
+    ]
+    [
+      "plugins"
+      "registers"
+    ]
+    [
+      "plugins"
+      "spelling"
+    ]
+    [
+      "plugins"
+      "presets"
+      "textObjects"
+    ]
+    [
+      "plugins"
+      "presets"
+      "operators"
+    ]
+    [
+      "plugins"
+      "presets"
+      "motions"
+    ]
+    [
+      "plugins"
+      "presets"
+      "windows"
+    ]
+    [
+      "plugins"
+      "presets"
+      "nav"
+    ]
+    [
+      "plugins"
+      "presets"
+      "z"
+    ]
+    [
+      "plugins"
+      "presets"
+      "g"
+    ]
+    "popupMappings"
+    "showHelp"
+    "showKeys"
+    "triggersBlackList"
+    "triggersNoWait"
+  ];
+
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "which-key"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+    in
+    [
+      (lib.mkRenamedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "disable"
+            "buftypes"
+          ]
+        )
+        (
+          settingsPath
+          ++ [
+            "disable"
+            "bt"
+          ]
+        )
+      )
+      (lib.mkRenamedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "disable"
+            "filetypes"
+          ]
+        )
+        (
+          settingsPath
+          ++ [
+            "disable"
+            "ft"
+          ]
+        )
+      )
+      (lib.mkRenamedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "window"
+            "winblend"
+          ]
+        )
+        (
+          settingsPath
+          ++ [
+            "win"
+            "wo"
+            "winblend"
+          ]
+        )
+      )
+      (lib.mkRenamedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "window"
+            "border"
+          ]
+        )
+        (
+          settingsPath
+          ++ [
+            "win"
+            "border"
+          ]
+        )
+      )
+      (lib.mkRemovedOptionModule (basePluginPath ++ [ "triggers" ]) ''
+        Please use `plugins.which-key.settings.triggers` instead.
+
+        See the plugin documentation for more details about the new option signature:
+        https://github.com/folke/which-key.nvim#-triggers
+      '')
+      (lib.mkRemovedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "window"
+            "padding"
+          ]
+        )
+        ''
+          Please use `plugins.which-key.settings.win.padding` instead.
+
+          See the plugin documentation for more details about the new option:
+          https://github.com/folke/which-key.nvim#%EF%B8%8F-configuration
+        ''
+      )
+      (lib.mkRemovedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "window"
+            "margin"
+          ]
+        )
+        ''
+          Please use `plugins.which-key.settings.layout` instead.
+
+          See the plugin documentation for more details about the new options:
+          https://github.com/folke/which-key.nvim#%EF%B8%8F-configuration
+        ''
+      )
+      (lib.mkRemovedOptionModule
+        (
+          basePluginPath
+          ++ [
+            "window"
+            "position"
+          ]
+        )
+        ''
+          Please use `plugins.which-key.settings.layout`, `plugins.which-key.settings.win.title_pos`, or `plugins.which-key.settings.win.footer_pos` instead.
+
+          See the plugin documentation for more details about the new options:
+          https://github.com/folke/which-key.nvim#%EF%B8%8F-configuration
+        ''
+      )
+    ];
+
+  settingsOptions = {
+    preset = defaultNullOpts.mkEnumFirstDefault [
+      "classic"
+      "modern"
+      "helix"
+      false
+    ] "Preset style for WhichKey. Set to false to disable.";
+
+    delay = defaultNullOpts.mkInt' {
+      pluginDefault.__raw = ''
+        function(ctx)
+          return ctx.plugin and 0 or 200
+        end
+      '';
+      description = "Delay before showing the popup. Can be a number or a function that returns a number.";
+    };
+
+    filter = defaultNullOpts.mkLuaFn' {
+      pluginDefault.__raw = ''
+        function(mapping)
+          return true
+        end
+      '';
+      description = "Filter used to exclude mappings";
+    };
+
+    spec = defaultNullOpts.mkListOf' {
       type = with types; attrsOf anything;
       pluginDefault = [ ];
       description = ''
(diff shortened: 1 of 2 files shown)


MattSturgeon:
> Working on converting this and there are a few breaking changes for data types of config options [folke/which-key.nvim#624 (files)](https://github.com/folke/which-key.nvim/pull/624/files#diff-a8347a13956b70b044d3e091cd3115e2fcfb44d4ebde516ff03223b2e84c7a7b).

TBH, 90% of users probably don't have much config in which-key, other than registrations. I don't think we have to gracefully transition all changes to the v3 scheme. Rather `mkRemovedOptionModule` with clear migration instructions would be enough for lesser used options or anything particularly complex.

> * One being the `triggers` option which used to be string or list of strings, but is now a list of attribute sets like the registrations change. What should we do in that case?

IMO this can be a `mkRemovedOptionModule` assertion.

> * Another issue I was hoping for clarification on is nested properties moving to settings and moving to a different path within settings.
> 
> * `plugins -> presets -> textObjects` moved and changed from camel to snake case. Does that get handled from the normal `optionsRenamedToSettings`?

`optionsRenamedToSettings` can handle _either_ string entries or listof-string entries (for nested options). It assumes all renames should use snake_case, so it converts any camelCase option names in the old path to snake_case in the new path.

Here's one example: https://github.com/nix-community/nixvim/blob/5922a48008e5759acb63a12b2de8348ec512760f/plugins/utils/indent-blankline.nix#L19-L112

Note: it's preferred to rename each of the old options, instead of a group of them as otherwise the camelCase/snake_case transformation won't happen for options nested within the group.

If you want anything more bespoke, such as a completely different name, you can manually use `mkRenamedOptionModule`.

> * `windows -> winblend` moved to `win -> wo -> winblend`. I'm unable to make a `mkRenamedOption` for that while also using the `optionsRenamedToSettings` for the parent object because of the order of evaluation, it seems.

This is another reason why it's usually better to rename each option rather than a common path containing a group of options.

> Will I need to explicitly remap all of those or just handle the deprecation like registrations and support forwarding the previous config through, for now?

Without looking at the actual code (yet), I think renaming all the options is probably better.

> Lots of TODO notes and some temporary test stuff for testing out renaming and warnings that can be ignored/questions answered

Thanks for working on this! Don't hesitate to ping as and when you have any questions/issues or would like a code review.

plugins/utils/which-key.nix
```diff
@@ -7,23 +7,94 @@
...
 
-    settings.spec = helpers.defaultNullOpts.mkListOf' {
+  settingsOptions = {
+    preset = helpers.defaultNullOpts.mkNullable (types.either types.bool (
+      types.enum [
+        "classic"
+        "modern"
+        "helix"
+      ]
+    )) "classic" "Preset style for WhichKey. Set to false to disable.";
```
MattSturgeon, line 85: Does the option accept `true` or just `false`? I'm thinking it may be better to include `false` as an enum value instead of having `either bool enum`.

plugins/utils/which-key.nix
```diff
@@ -7,23 +7,94 @@
...
+
+    delay = helpers.defaultNullOpts.mkInt (helpers.mkRaw ''
+      function(ctx)
+        return ctx.plugin and 0 or 200
+      end'') "Delay before showing the popup. Can be a number or a function that returns a number.";
+
+    filter = helpers.defaultNullOpts.mkLuaFn (helpers.mkRaw ''
+      function(mapping)
+        return true
+      end'') "Filter used to exclude mappings";
```
MattSturgeon, line 95: These may be more readable if declared using the `mk'` variants?

Although, we're (slowly) coming around to the idea of declaring fewer `settingsOptions` and relying more on the freeformType.

So it may make more sense _not_ to include some of these lesser used options and instead point users to the upstream docs.

plugins/utils/which-key.nix
```diff
@@ -55,257 +126,265 @@ with lib;
...
 
+    notify = helpers.defaultNullOpts.mkBool true "Show a warning when issues were detected with your mappings.";
+
+    # TODO: clarify breaking change
+    triggers = helpers.defaultNullOpts.mkNullable (with types; listOf (attrsOf anything)) [
+      {
+        __unkeyed = "<auto>";
+        mode = "nxsot";
+      }
+    ] "Manually setup triggers";
```
MattSturgeon, line 137: - Can we use `mkListOf (attrsOf anything)`? This will inject rawLua support.
- I think the breaking change can be tackled in a `mkRemoveOptionModule` assertion message?
- Or a similar warning to `opt.registrations.isDefined` and `opt.window.isDefined` if the old format is still supported upstream?

plugins/utils/which-key.nix
```diff
@@ -55,257 +126,265 @@ with lib;
...
+    window = mkOption {
+      type = with types; attrsOf anything;
+      description = ''
+        This option is deprecated, use `settings.win` instead.
+
+        Note: the window options have changed parameters in v3 of which-key.
+      '';
+      visible = false;
+    };
+  };
```
MattSturgeon, line 360: This would benefit from a `TODO: added/remove after` comment similar to our other deprecations. I guess I neglected to do that when I deprecated `registrations`!

plugins/utils/which-key.nix
```diff
@@ -55,257 +126,265 @@ with lib;
...
-        nixvim (plugins.which-key):
-        The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.
-        You should use `plugins.which-key.settings.spec' instead.
+          Note: the spec format has changed in which-key v3
+          See: https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings
+        ''
+        ++ optional opt.window.isDefined ''
+          nixvim (plugins.which-key):
+          The option definition `plugins.which-key.window' in ${showFiles opt.window.files} has been deprecated in which-key v3; please remove it.
+          You should use `plugins.which-key.settings.win' instead.
```
MattSturgeon, line 381: This warning is great, but I think we still need to implement passing `cfg.window` through to which-key?

If it maps to a setting option, we can do something like:
```
plugins.which-key.settings = optionalAttrs opt.window.isDefined {
  win = lib.modules.mkAliasAndWrapDefsWithPriority lib.id opt.window;
};
```

`lib.id` (aka `x: x`) could be replaced with a transformation function if needed.

plugins/utils/which-key.nix
```diff
@@ -55,257 +126,265 @@ with lib;
...
+
+    window = mkOption {
+      type = with types; attrsOf anything;
+      description = ''
+        This option is deprecated, use `settings.win` instead.
+
+        Note: the window options have changed parameters in v3 of which-key.
+      '';
+      visible = false;
+    };
```
MattSturgeon, line 359: The sheer quantity of options here is worrying to me (from a maintenance perspective). If any one of these changes upstream, we would have to figure out how to update it in our module - assuming we're even aware of the change!

While it's nice at a superficial level to have comprehensive type-checking and documentation in nixvim, it's my view that it isn't worth the drawbacks to do so.

I appreciate it would have taken some effort to check over them all, so thanks for your efforts regardless!

We're still having this debate internally, so take my opinion with a grain of salt.

plugins/utils/which-key.nix
```diff
@@ -55,257 +126,265 @@ with lib;
...
+  };
+
+  extraConfig =
+    cfg:
     let
       cfg = config.plugins.which-key;
       opt = options.plugins.which-key;
```
MattSturgeon, line 366: `cfg` is now defined twice, we can drop the definition in the `let` block.

If we end up with a file-header `let` block, we could move the `opt` definition there too.

Maybe it'd be nice if `mkNeovimPlugin` allowed a variable number of args (similar to e.g. `<pkg>.overrideAttrs`):

```nix
callExtraConfig =
  fnOrAttrs:
  let
    cfg = config.plugins.${name};
    opt = options.plugins.${name};
    fnOrAttrs' = fnOrAttrs cfg;
  in
  if isAttrs fnOrAttrs then
    fnOrAttrs
  else
    if isAttrs fnOrAttrs' then
      fnOrAttrs'
    else
      fnOrAttrs' opt;
```

This'd mean it needs access to `options` as well as `config`, but that could be done by importing a module. Doing that would make passing in `config` redundant too...

plugins/utils/which-key.nix
```diff
@@ -7,23 +7,146 @@
   ...
 }:
 with lib;
-{
-  options.plugins.which-key = {
-    enable = mkEnableOption "which-key.nvim, a plugin that popup with possible key bindings of the command you started typing";
+let
+  opt = options.plugins.which-key;
+in
+helpers.neovim-plugin.mkNeovimPlugin config {
```
MattSturgeon, line 13:
```suggestion
lib.nixvim.neovim-plugin.mkNeovimPlugin config {
```

plugins/utils/which-key.nix
```diff
@@ -7,23 +7,146 @@
   ...
 }:
 with lib;
-{
-  options.plugins.which-key = {
-    enable = mkEnableOption "which-key.nvim, a plugin that popup with possible key bindings of the command you started typing";
+let
+  opt = options.plugins.which-key;
```
MattSturgeon, line 11:
```suggestion
  inherit (lib) types;
  inherit (lib.nixvim) defaultNullOpts;

  opt = options.plugins.which-key;
```

This would allow removing `helpers` from the module args.

plugins/utils/which-key.nix
khaneliman: I guess my approach to covering the options was that when I got started with nix, nixvim helped me discover the options a plugin provided through the documentation and helped implementation a ton through the type definitions.

plugins/utils/which-key.nix
khaneliman: Was thinking maybe like this:
```nix
    extraConfigLua =
      let
        setupOptions = optionalAttrs opt.window.isDefined {
          win = lib.modules.mkAliasAndWrapDefsWithPriority lib.id opt.window;
        };
      in
      ''
        require("which-key").setup({ ${toLuaObject setupOptions}})
      ''
      + optionalString (opt.registrations.isDefined && cfg.registrations != { }) ''
        require("which-key").register(${toLuaObject cfg.registrations})
      '';
```
But, this is what gets generated with that inside the module... which doesn't look right to me.

```lua
require("which-key").setup({
    key_labels = {
        ["<BS>"] = "BACKSPACE",
        ["<CR>"] = "RETURN",
        ["<TAB>"] = "TAB",
        ["<bs>"] = "BACKSPACE",
        ["<cr>"] = "RETURN",
        ["<leader>"] = "SPACE",
        ["<space>"] = "SPACE",
        ["<tab>"] = "TAB",
    },
    spec = {
        { "<leader>c", group = "󰄄 Codesnap", mode = "v" },
        { "<leader>d", desc = "  Debug", mode = "n" },
        { "<leader>gW", group = "󰙅 Worktree" },
        { "<leader>gh", group = "  Hunks" },
        { "<leader>ug", group = "󰊢 Git" },
        { "<leader>p", group = " Preview", mode = "n" },
        { "<leader>h", group = "󱡀 Harpoon" },
        { "<leader>ha", desc = "Add" },
        { "<leader>he", desc = "QuickMenu" },
        { "<leader>hj", desc = "1" },
        { "<leader>hk", desc = "2" },
        { "<leader>hl", desc = "3" },
        { "<leader>hm", desc = "4" },
        { "<leader>l", group = "  LSP" },
        { "<leader>la", desc = "Code Action" },
        { "<leader>ld", desc = "Definition" },
        { "<leader>lD", desc = "References" },
        { "<leader>lf", desc = "Format" },
        { "<leader>lp", desc = "Prev" },
        { "<leader>ln", desc = "Next" },
        { "<leader>lt", desc = "Type Definition" },
        { "<leader>li", desc = "Implementation" },
        { "<leader>lh", desc = "Hover" },
        { "<leader>lr", desc = "Rename" },
        { "<leader>r", group = " Refactor", mode = "x" },
        { "<leader>x", group = " Trouble", mode = "n" },
        { "<leader>b", group = "󰓩 Buffers" },
        { "<leader>bs", group = "󰒺 Sort" },
        { "<leader>g", group = "󰊢 Git" },
        { "<leader>f", group = " Find" },
        { "<leader>r", group = " Refactor" },
        { "<leader>t", group = " Terminal" },
        { "<leader>u", group = " UI/UX" },
    },
})

require("which-key").setup({
    {
        win = {
            _type = "if",
            condition = true,
            content = {
                _type = "merge",
                contents = { { _type = "override", content = { border = "single" }, priority = 100 } },
            },
        },
    },
})
```

MattSturgeon: You're right it shouldn't be set as `settings.win`, since the old format had a different name.

We could take advantage of `settings` being freeform here though and do something like this
https://github.com/nix-community/nixvim/blob/2089eb407d8c5dbd6ca6e93d4988a439ca6446fd/plugins/utils/which-key.nix#L262-L278

But wrapped in an `optionalAttrs opt.window.isDefined {}`:

```nix
config = {
  settings = optionalAttrs opt.window.isDefined {
    window = makeAliasAndWrapWithPriority (
      V:
      let
        mkSpacing =
          x:
          helpers.ifNonNull' x [
            x.top
            x.right
            x.bottom
            x.top
          ];
      in
      {
        inherit (v) border position;
        margin = mkSpacing v.margin;
        padding = mkSpacing v.padding;
        inherit (v) winblend;
      }
    ) opt.window;
  };
}
```

(aliasing and wrapping with priority here is probably overkill, since this a deprecated option and is likely only set in the old location. I included it just for the example)

MattSturgeon: The simple version of that example is:

```nix
config = {
  settings = optionalAttrs opt.window.isDefined {
    window =
    let
      mkSpacing =
        x:
        helpers.ifNonNull' x [
          x.top
          x.right
          x.bottom
          x.top
        ];
    in
    {
      inherit (cfg.window) border position;
      margin = mkSpacing cfg.window.margin;
      padding = mkSpacing cfg.window.padding;
      inherit (cfg.window) winblend;
    };
  };
}
```

plugins/utils/which-key.nix
MattSturgeon: Yes, historically we've taken the same approach of declaring options for pretty much everything.

It is nice to be able to reference nixvim's docs to discover plugin options however this comes with plenty of downsides too:

- no documentation is better than incorrect documentation
  - it's easy for us to get things wrong when manually adding options, or for our options to get out of date due to upstream changes
- type checking is nice, until it prevents defining valid config
  - having typed sub-options overrides the freeformType, somewhat defeating the point of rfc42
- sub-options are the _only_ reason we have to filter out null/empty values from the final `settings` value
  - this is a bigger issue though, as we'd either have to remove _all_ sub-options or find a way to only merge _defined_ options into the final value
  - the latter would require changes to nixpkgs `lib.evalModules` but is probably the correct approach
  - to clarify, all options end up in `config`; undefined options just `throw` when evaluated.
  - This isn't really relevant (the other two points are)

khaneliman: @GaetanLepage @traxys Either of you have an opinion on the scope of options provided in the module? Originally went through and made sure all the plugin's options were defined explicitly, but there are drawbacks (as we've seen previously) to having it done.

GaetanLepage: @MattSturgeon's explanation is quite exhaustive.
Actually, we (maintainers) haven't really settled on a clear decision on this topic.
It is quite hard to do so, as no solution is perfect. I also find a great value in having a lot of options defined. This comes very handy to the users.
However, I am more than aware of the maintenance burden that it implies.

Given the amount of options declared in nixvim today, I see no future where we shift entirely to empty `settings` submodules. In this sense, we will likely keep a hybrid approach, no matter what strategy we encourage/enforce.

In practice, my opinion is that while we agree on a clear way of doing this, plugin maintainers should be granted the freedom of doing what they want.
At the same time, you should be somehow responsible for trying to keep those options up to date.
All in all, if you want to keep those declarations, I am personally fine with it.

MattSturgeon: > In practice, my opinion is that while we agree on a clear way of doing this, plugin maintainers should be granted the freedom of doing what they want.
> At the same time, you should be somehow responsible for trying to keep those options up to date.
> All in all, if you want to keep those declarations, I am personally fine with it.

I'm fine with this in the short term. It seems unfair to ask plugin maintainers to follow a policy we haven't fully decided on yet.

GaetanLepage: > Sure you didn't mean _exhausting_? 😉

No haha ! I sincerely admire your surgically clear prose. It is always a joy to read !

plugins/utils/which-key.nix
```diff
@@ -276,36 +521,28 @@ with lib;
             padding = mkSpacing cfg.window.padding;
             inherit (cfg.window) winblend;
           };
```
MattSturgeon, line 523: I'm not sure, but I think these need to be gated behind checking `isDefined`; so far we've only checked that _some of them_ are defined.

If we accidentally read an undefined option value, the module system will throw "used but not defined".

I think this should be something like:

```nix
window =
let
  mkSpacing = ;# omitted

  setIfDef =
    name: fn: option:
    optionalAttrs option.isDefined { ${name} = fn option.value; };
in
  setIfDef "border" lib.id opt.window.border
  // setIfDef "position" lib.id opt.window.position
  // setIfDef "margin" mkSpacing opt.window.margin
  // setIfDef "padding" mkSpacing opt.window.padding
  // setIfDef "winblend" lib.id opt.window.winblend;
```

plugins/utils/which-key.nix
```diff
@@ -276,36 +521,28 @@ with lib;
...
+          nixvim (plugins.which-key):
+          The option definition `plugins.which-key.window' in ${showFiles opt.window.border.files} has been deprecated in which-key v3; please remove it.
+          You should use `plugins.which-key.settings.win' instead.
+
+          Note: the window options have changed parameters in which-key v3
+        '';
+
+      extraConfigLua = optionalString (opt.registrations.isDefined && cfg.registrations != { }) ''
+        require("which-key").register(${toLuaObject cfg.registrations})
+      '';
```
MattSturgeon, line 546: Now that the only line being set `extraConfigLua` is the optional one, we should probably not use `optionalString`.

When the condition is false, an empty string will be defined which will result in a blank line being needlessly added to init.lua.

Normally, we'd use `mkIf` here, however that's won't do what we want; `cfg.registrations` would still be evaluated even when `opt.registrations.isDefined = false`.

The "correct" solution would be to either use `mkMerge` or an `//` update combined with `optionalAttrs`, e.g:

```nix
extraConfig = cfg: mkMerge [
  {/* main config */}
  (optionalAttrs opt.registrations.isDefined {
    # config that uses cfg.registrations
  })
];
```

plugins/utils/which-key.nix
```diff
@@ -121,153 +475,44 @@ with lib;
...
-        key_labels = cfg.keyLabels;
-        inherit (cfg) motions icons;
-        popup_mappings = {
-          scroll_down = cfg.popupMappings.scrollDown;
-          scroll_up = cfg.popupMappings.scrollUp;
-        };
+      windowIsDefined = any (o: o.isDefined && o.value != null) (attrValues opt.window);
+    in
+    cfg: {
+      plugins.which-key.settings = optionalAttrs windowIsDefined {
```
MattSturgeon, line 506: This is what I originally suggested, but it'd be cleaner IMO to use `mkMerge` here:

```
extraConfig = cfg: lib.mkMerge [
  (optionalAttrs windowIsDefined {
    plugins.which-key.settings.window = #omitted
  })
  {/* main config */}
];
```

Otherwise, we end up with an extra `settings` definition, just setting it to `{}`. Not a big deal though, since that'd be merged fine with more useful definitions.

This is definitely a nit, either approach works and it's personal opinion which is best. On a different day I may prefer the current approach, it really is that subjective!

plugins/utils/which-key.nix
MattSturgeon: Maybe add a (temporary) test to check what happens when `plugins.which-key.window` is partially defined.

MattSturgeon (commented):
I've focused this review on the `window` deprecation. The rest of the changes I've only really skimmed over (so far), but overall they're looking good!

plugins/utils/which-key.nix
MattSturgeon: See #1909

plugins/utils/which-key.nix
khaneliman: I had my personal config (that I sent the init.lua output of during troubleshooting in matrix) and the test for setting just the window.border and it worked fine.

plugins/utils/which-key.nix
```diff
@@ -121,153 +475,44 @@ with lib;
...
-          };
-        };
-        inherit (cfg) operators;
-        key_labels = cfg.keyLabels;
-        inherit (cfg) motions icons;
-        popup_mappings = {
-          scroll_down = cfg.popupMappings.scrollDown;
-          scroll_up = cfg.popupMappings.scrollUp;
-        };
+      windowIsDefined = any (o: o.isDefined && o.value != null) (attrValues opt.window);
```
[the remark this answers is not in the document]
khaneliman: When I remove the `value != null` I get a warning no matter the configuration, for some reason.

plugins/utils/which-key.nix
```diff
@@ -121,153 +475,44 @@ with lib;
...
           top = 1;
           right = 2;
           bottom = 1;
           left = 2;
         } "extra window padding";
-        winblend = helpers.defaultNullOpts.mkNullable (types.ints.between 0
+        winblend = defaultNullOpts.mkNullable (types.ints.between 0
           100
         ) 0 "0 for fully opaque and 100 for fully transparent";
       };
```
MattSturgeon, line 498: These options continuing to be `defaultNullOpts` is what's leading to weirdness in https://github.com/nix-community/nixvim/pull/1990#discussion_r1708188447 and https://github.com/nix-community/nixvim/pull/1990#discussion_r1708170104.

In order for `options.*.isDefined` to be useful, the option must **not** have a default value. If it has a default, `isDefined` is _always_ true.

Additionally, we also want these options to be `visible = false`, since they are deprecated.

We can achieve this by using `mkOption` directly, we just have to manually replicate whatever type the `defaultNullOpts` helper would have produced. Luckily, since the option will not be visible, there's not much point in setting a plugin-default anymore (which would be awkward when using `mkOption` directly).

Here's an example for one of the options:

```nix
position = mkOption {
  type = nullOr (maybeRaw (enumOf [
    "bottom"
    "top"
  ]));
  description = ''
    This option is deprecated, you should use `plugins.which-key.settings.win` instead.
  '';
  visible = false;
  # Note: no default is defined
}
```

plugins/utils/which-key.nix
khaneliman: Not seeing an empty string in my generated config with this and when I tried using the mkMerge on extraConfig I kept getting infinite recursion.

khaneliman:
Last push is removing the graceful deprecation of `window` attribute. During testing, the options weren't being used by the actual plugin even though they are listed as just deprecated.

khaneliman:
Made border and winblend options renames to gracefully transition users with that config.

plugins/utils/which-key.nix
```diff
@@ -1,29 +1,239 @@
...
+      "modern"
+      "helix"
+      false
+    ] "Preset style for WhichKey. Set to false to disable.";
+
+    delay = defaultNullOpts.mkInt' {
+      pluginDefault = ''
+        function(ctx)
+          return ctx.plugin and 0 or 200
+        end'';
```
MattSturgeon, line 224: Needs to be rawLua. Also, we generally prefer to have a newline to make the string syntax more readable, unless you specifically want to avoid the trailing linebreak?


```suggestion
      pluginDefault.__raw = ''
        function(ctx)
          return ctx.plugin and 0 or 200
        end
      '';
```

plugins/utils/which-key.nix
```diff
@@ -1,29 +1,239 @@
...
+      description = "Delay before showing the popup. Can be a number or a function that returns a number.";
     };
 
-    settings.spec = helpers.defaultNullOpts.mkListOf' {
+    filter = defaultNullOpts.mkLuaFn' {
+      pluginDefault = ''
+        function(mapping)
+          return true
+        end'';
```
MattSturgeon, line 232: Same
```suggestion
      pluginDefault.__raw = ''
        function(mapping)
          return true
        end
      '';
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
       ];
     };
 
+    notify = defaultNullOpts.mkBool true "Show a warning when issues were detected with your mappings.";
+
+    triggers = defaultNullOpts.mkListOf (with types; attrsOf anything) [
+      {
+        __unkeyed = "<auto>";
```
MattSturgeon, line 272:
```suggestion
        __unkeyed-1 = "<auto>";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+            end'')
+        ''
+          Start hidden and wait for a key to be pressed before showing the popup.
+
+          Only used by enabled xo mapping modes.
+        '';
+
     plugins = {
-      marks = helpers.defaultNullOpts.mkBool true "shows a list of your marks on ' and `";
+      marks = defaultNullOpts.mkBool true "Shows a list of your marks on ' and `";
```
MattSturgeon, line 290: We need to format descriptions as markdown, also we prefer to have punctation:

```suggestion
      marks = defaultNullOpts.mkBool true "Shows a list of your marks on `'` and <code>\`</code>.";
```

---

Also, I wonder if many of these would be more readable with indented strings?

```suggestion
      marks = defaultNullOpts.mkBool true ''
        Shows a list of your marks on `'` and <code>\`</code>.
      '';
```

MattSturgeon: I'm not sure if we need <code>&lt;code&gt;\\\`&lt;/code&gt;</code> or if <code>&lt;code&gt;\`&lt;/code&gt;</code> would be enough :thinking: 

Based on github's markdown, we'd need the backslash, but commonmark may be different.

This can be double-checked by entering a devshell and running `serve-docs`

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
 
-      registers = helpers.defaultNullOpts.mkBool true ''shows your registers on " in NORMAL or <C-r> in INSERT mode'';
+      registers = defaultNullOpts.mkBool true ''Shows your registers on " in NORMAL or <C-r> in INSERT mode'';
```
MattSturgeon, line 292: Same
```suggestion
      registers = defaultNullOpts.mkBool true ''
        Shows your registers on `"` in NORMAL or `<C-r>` in INSERT mode.
      '';
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
 
       spelling = {
-        enabled = helpers.defaultNullOpts.mkBool true "enabling this will show WhichKey when pressing z= to select spelling suggestions";
-        suggestions = helpers.defaultNullOpts.mkInt 20 "how many suggestions should be shown in the list?";
+        enabled = defaultNullOpts.mkBool true "Enabling this will show WhichKey when pressing z= to select spelling suggestions";
```
MattSturgeon, line 295:
```suggestion
        enabled = defaultNullOpts.mkBool true ''
          Enabling this will show WhichKey when pressing `z=` to select spelling suggestions.
        '';
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+        suggestions = defaultNullOpts.mkInt 20 "How many suggestions should be shown in the list?";
```
MattSturgeon, line 296:
```suggestion
        suggestions = defaultNullOpts.mkInt 20 ''
          How many suggestions should be shown in the list?
        '';
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-        nav = helpers.defaultNullOpts.mkBool true "misc bindings to work with windows";
-        z = helpers.defaultNullOpts.mkBool true "bindings for folds, spelling and others prefixed with z";
-        g = helpers.defaultNullOpts.mkBool true "bindings for prefixed with g";
+        operators = defaultNullOpts.mkBool true "Adds help for operators like d, y, ...";
+        motions = defaultNullOpts.mkBool true "Adds help for motions";
+        text_objects = defaultNullOpts.mkBool true "Help for text objects triggered after entering an operator";
+        windows = defaultNullOpts.mkBool true "Default bindings on <c-w>";
+        nav = defaultNullOpts.mkBool true "Misc bindings to work with windows";
+        z = defaultNullOpts.mkBool true "Bindings for folds, spelling and others prefixed with z";
+        g = defaultNullOpts.mkBool true "Bindings for prefixed with g";
```
MattSturgeon, line 306: These descriptions are probably short enough that they are better as they are instead of changing to indented strings.

Just need to add punctuation and markdown formatting.

```suggestion
        operators = defaultNullOpts.mkBool true "Adds help for operators like `d`, `y`, ...";
        motions = defaultNullOpts.mkBool true "Adds help for motions.";
        text_objects = defaultNullOpts.mkBool true "Help for text objects triggered after entering an operator.";
        windows = defaultNullOpts.mkBool true "Default bindings on `<c-w>`.";
        nav = defaultNullOpts.mkBool true "Misc bindings to work with windows.";
        z = defaultNullOpts.mkBool true "Show WhichKey for folds, spelling and other bindings prefixed with `z`.";
        g = defaultNullOpts.mkBool true "Show WhichKey for bindings prefixed with `g`.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
 
-    keyLabels = helpers.defaultNullOpts.mkAttrsOf types.str { } ''
-      override the label used to display some keys. It doesn't effect WK in any other way.
-    '';
+      border = defaultNullOpts.mkBorder "none" "which-key" "";
 
-    motions = {
-      count = helpers.defaultNullOpts.mkBool true "";
-    };
+      padding = defaultNullOpts.mkNullable (with types; listOf int) [
```
MattSturgeon, line 315: Is it `int` or `unsignedInt`?

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+        1
+        2
+      ] "Extra window padding [top/bottom, right/left]";
```
MattSturgeon, line 318:
```suggestion
      ] "Extra window padding, in the form `[top/bottom, right/left]`.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-      separator = helpers.defaultNullOpts.mkStr "➜" "symbol used between a key and it's label";
-      group = helpers.defaultNullOpts.mkStr "+" "symbol prepended to a group";
-    };
+      title = defaultNullOpts.mkBool true "Whether to show the title.";
 
-    popupMappings = {
-      scrollDown = helpers.defaultNullOpts.mkStr "<c-d>" "binding to scroll down inside the popup";
-      scrollUp = helpers.defaultNullOpts.mkStr "<c-u>" "binding to scroll up inside the popup";
-    };
+      title_pos = defaultNullOpts.mkStr "center" "Position of the title";
```
MattSturgeon, line 322:
```suggestion
      title_pos = defaultNullOpts.mkStr "center" "Position of the title.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-          left = 0;
-        } "extra window margin";
-        padding = helpers.defaultNullOpts.mkNullable spacingOptions {
-          top = 1;
-          right = 2;
-          bottom = 1;
-          left = 2;
-        } "extra window padding";
-        winblend = helpers.defaultNullOpts.mkNullable (types.ints.between 0
+      zindex = defaultNullOpts.mkInt 1000 "Layer depth on the popup window";
```
MattSturgeon, line 324:
```suggestion
      zindex = defaultNullOpts.mkInt 1000 "Layer depth on the popup window.";
```

Are negative numbers allowed, or should this be `mkUnsignedInt`?

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+
+      wo = {
+        winblend = defaultNullOpts.mkNullable (types.ints.between 0
```
MattSturgeon, line 327: Unlike most defaultNullOpts, mkNullable doesn't inject rawLua support into the type.

```suggestion
        winblend = defaultNullOpts.mkNullableWithRaw (types.ints.between 0
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
           100
         ) 0 "0 for fully opaque and 100 for fully transparent";
```
MattSturgeon, line 329:
```suggestion
        ) 0 "`0` for fully opaque and `100` for fully transparent";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-      "^call "
-      "^lua "
-    ] "hide mapping boilerplate";
+      spacing = defaultNullOpts.mkInt 3 "Spacing between columns";
+    };
 
-    showHelp = helpers.defaultNullOpts.mkBool true "show a help message in the command line for using WhichKey";
+    keys = {
+      scroll_up = defaultNullOpts.mkStr "<c-u>" "Binding to scroll up in the popup";
+      scroll_down = defaultNullOpts.mkStr "<c-d>" "Binding to scroll down in the popup";
```
MattSturgeon, line 344:
```suggestion
      scroll_up = defaultNullOpts.mkStr "<c-u>" "Binding to scroll up in the popup.";
      scroll_down = defaultNullOpts.mkStr "<c-d>" "Binding to scroll down in the popup.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+          "case"
+        ])
+        [
+          "local"
+          "order"
+          "group"
+          "alphanum"
+          "mod"
+        ]
+        "Mappings are sorted using configured sorters and natural sort of the keys";
```
MattSturgeon, line 365:
```suggestion
        "Mappings are sorted using configured sorters and natural sort of the keys.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-      helpers.defaultNullOpts.mkListOf types.str
+    replace = {
+      key =
+        defaultNullOpts.mkListOf (types.either nixvimTypes.strLuaFn (with types; listOf str))
+          (mkRaw ''
+            function(key)
+              return require("which-key.view").format(key)
+            end
+          '')
+          "";
```
MattSturgeon, line 377: No description?

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-        }
-        ''
-          list of mode / prefixes that should never be hooked by WhichKey
-          this is mostly relevant for keymaps that start with a native binding
-        '';
+    icons = {
+      breadcrumb = defaultNullOpts.mkStr "»" "Symbol used in the command line area that shows your active key combo";
+      separator = defaultNullOpts.mkStr "➜" "Symbol used between a key and it's label";
+      group = defaultNullOpts.mkStr "+" "Symbol prepended to a group";
+      ellipsis = defaultNullOpts.mkStr "…" "Symbol used for overflow ";
```
MattSturgeon, line 418:
```suggestion
      breadcrumb = defaultNullOpts.mkStr "»" "Symbol used in the command line area that shows your active key combo.";
      separator = defaultNullOpts.mkStr "➜" "Symbol used between a key and its label.";
      group = defaultNullOpts.mkStr "+" "Symbol prepended to a group.";
      ellipsis = defaultNullOpts.mkStr "…" "Symbol used for overflow.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+        F7 = "󱊱";
+        F8 = "󱊲";
+        F9 = "󱊳";
+        F10 = "󱊴";
+        F11 = "󱊵";
+        F12 = "󱊶";
+      } "Icons used by key format.";
+    };
+
+    show_help = defaultNullOpts.mkBool true "show a help message in the command line for using WhichKey";
```
MattSturgeon, line 461:
```suggestion
    show_help = defaultNullOpts.mkBool true "Show a help message in the command line for using WhichKey.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+
+    show_keys = defaultNullOpts.mkBool true "show the currently pressed key and its label as a message in the command line";
```
MattSturgeon, line 463:
```suggestion
    show_keys = defaultNullOpts.mkBool true "Show the currently pressed key and its label as a message in the command line.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
 
     disable = {
-      buftypes = helpers.defaultNullOpts.mkListOf types.str [ ] "Disabled by default for Telescope";
-      filetypes = helpers.defaultNullOpts.mkListOf types.str [ ] "";
+      bt = defaultNullOpts.mkListOf types.str [ ] "Disabled by default for Telescope";
+      ft = defaultNullOpts.mkListOf types.str [ ] "";
```
MattSturgeon, line 467:
```suggestion
      bt = defaultNullOpts.mkListOf types.str [ ] "Disabled by default for Telescope.";
      ft = defaultNullOpts.mkListOf types.str [ ] ""; # TODO: description
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
     };
+
+    debug = defaultNullOpts.mkBool false "Enable wk.log in the current directory.";
```
MattSturgeon, line 470:
```suggestion
    debug = defaultNullOpts.mkBool false "Enable `wk.log` in the current directory.";
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
-        The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.
-        You should use `plugins.which-key.settings.spec' instead.
+    delay = 200;
 
-        Note: the spec format has changed in which-key v3
-        See: https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings
-      '';
+    spec = [
+      {
+        __unkeyed = "<leader>b";
```
MattSturgeon, line 480: While any key starting with `__unkeyed` will work, can we use `__unkeyed-1` in examples since that matches what `lib.nixvim.listToUnkeyedAttrs` will spit out.

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+    notify = false;
+
+    win = {
+      border = "single";
+    };
+
+    expand = 1;
+  };
+
+  # TODO: introduced 2024-08-05: remove after 24.11
```
MattSturgeon, line 543:
```suggestion
  # TODO: introduced 2024-07-29: remove after 24.11
```

plugins/utils/which-key.nix
```diff
@@ -55,257 +265,308 @@ with lib;
...
+      You should use `plugins.which-key.settings.spec' instead.
+
+      Note: the spec format has changed in which-key v3
+      See: https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings
+    '';
+
+    extraConfigLua = optionalString (opt.registrations.isDefined && cfg.registrations != { }) ''
+      require("which-key").register(${toLuaObject cfg.registrations})
+    '';
+  };
```
MattSturgeon, line 571: We can wrap the _whole thing_ in optionalAttrs now, since the entire config block relates to the deprecated `registrations` option:

```nix
  # TODO: introduced 2024-07-29: remove after 24.11
  # NOTE: this may be upgraded to a mkRemoveOptionModule when which-key removes support
  extraConfig =
    cfg:
    let
      # I'm on the fence whether this is better off here or at the top of the file
      # If it's here, then we can remove this whole block in one go when removing the deprecated option, I guess...
      opt = options.plugins.which-key;
    in
      optionalAttrs opt.registrations.isDefined {
        warnings = [
          ''
            nixvim (plugins.which-key):
            The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.

          ''
        ];

        extraConfigLua = mkIf (cfg.registrations != { }) ''
          require("which-key").register(${toLuaObject cfg.registrations})
        '';
      };
```

MattSturgeon (changes requested):
Thanks for your continued contributions!

This is looking pretty good, I've tried to focus more on the detail and pick out any nits with this review. The PR is looking more or less done.

It's up to you, but I still wonder if some of the more "internal" feeling options should be omitted from `settingsOptions`? As discussed before, you can make the judgement call of which options you wish to support until we come up with a policy agreed on by all maintainers.

plugins/utils/which-key.nix
khaneliman: I'll double check. I originally assumed it could be used with negative.

plugins/utils/which-key.nix
khaneliman: Will have to double check. Similar to padding, I was thinking negatives like in web development.

plugins/utils/which-key.nix
khaneliman: Doing this is what causes the infinite recursion.

MattSturgeon: It should be fine in this instance, since we aren't defining anything other than `warnings` and `extraConfigLua`

khaneliman: I tried and it threw the recursion error. I assumed it had to do with referencing `cfg.` within the optionalAttrs

plugins/utils/which-key.nix
khaneliman: image

plugins/utils/which-key.nix
MattSturgeon: It seems relying on `options` to define the _shape_ of the config definition is infinitely recursive. I.e. `{}` vs `{ warnings = []; }` is a different shape, because warnings isn't defined at all in the first example.

This is a constant shape, I've tried it out locally and it works:

```nix
  # TODO: introduced 2024-07-29: remove after 24.11
  # NOTE: this may be upgraded to a mkRemoveOptionModule when which-key removes support
  extraConfig = cfg: {
    warnings = lib.optional opt.registrations.isDefined ''
      nixvim (plugins.which-key):
      The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.

    '';

    extraConfigLua =
      if opt.registrations.isDefined then
        lib.mkIf (cfg.registrations != { }) ''
          require("which-key").register(${toLuaObject cfg.registrations})
        ''
      else
        lib.mkIf false "";
  };
```

This would also work, but feels even messier:

```nix
  extraConfig =
    cfg:
    if opt.registrations.isDefined then
      {
        warnings = [
          # omitted
        ];
        extraConfigLua = lib.mkIf (cfg.registrations != { }) ''
          # omitted
        '';
      }
    else
      {
        warnings = [ ];
        extraConfigLua = lib.mkIf false "";
      };

File: plugins/utils/which-key.nix (hunk 18 above)
Comment (MattSturgeon) [reply plugins/utils/which-key.nix]: Maybe the doc's implementation blocks HTML elements in the markdown? :sob: 

We can try <code>\`\\\`\`</code>, not sure if it'll work :face_with_diagonal_mouth:

Comment (khaneliman) [reply plugins/utils/which-key.nix]: I'm trying out stuff locally, didn't realize we had the serve-docs command. That's really helpful!

Comment (khaneliman) [reply plugins/utils/which-key.nix]: Looks like 
```
`` ` ``
```

File: plugins/utils/which-key.nix (hunk 26 above)
Comment (khaneliman) [reply plugins/utils/which-key.nix]: Which-key throws error on negative. Changing.

File: plugins/utils/which-key.nix (hunk 23 above)
Comment (khaneliman) [reply plugins/utils/which-key.nix:326]: Can be negative to reduce padding. But, requires 2 number.

File: plugins/utils/which-key.nix (hunk 18 above)
Comment (MattSturgeon) [reply plugins/utils/which-key.nix]: This may help: https://meta.stackexchange.com/a/70679

Depending on the markdown flavour we're dealing with, we may be able to do <code>\`\`some code span with a \` in it\`\`</code>

Or alternatively, we may be able to use the html code `&#96;` as <code>\`&amp;#96;\`</code>

Comment (khaneliman) [reply plugins/utils/which-key.nix]: Yeah, that looks like what I ended up doing. Just wrapping it in a double backtick

File: plugins/utils/which-key.nix [hunk 40]
```diff
@@ -55,257 +277,317 @@ with lib;
...
-      add operators that will trigger motion and text object completion
-      to enable all native operators, set the preset / operators plugin above
-    '';
+    win = {
+      no_overlap = defaultNullOpts.mkBool true "Don't allow the popup to overlap with the cursor.";
 
-    keyLabels = helpers.defaultNullOpts.mkAttrsOf types.str { } ''
-      override the label used to display some keys. It doesn't effect WK in any other way.
-    '';
+      border = defaultNullOpts.mkBorder "none" "which-key" "";
```
MattSturgeon, line 333: Any idea what the description should be here? Also, have we double checked `mkBorder` is still the correct type for the new v3 format?

MattSturgeon (changes requested):
Looks like we're down to the final nits now! :rocket:

plugins/utils/which-key.nix
khaneliman: Sorry, forgot this one. It's basically just the border property inherited from neovim's Win type iirc

MattSturgeon (approved):
No more issues from me!

I'm still uncomfortable with having option declarations for _all_ settings-options, but I'm fine with it in the short term if @khaneliman is comfortable maintaining them.

@GaetanLepage @traxys any final feedback before we merge?

@khaneliman thanks again for all your efforts!

GaetanLepage:
All good for me. I had a quick browse of the changes and I haven't seen anything too fishy.
I'm OK for declaring all the options, as it is at the maintainer's initiative.

Kudos to both of you for this chunky PR !

GaetanLepage (approved):

MattSturgeon:
@mergifyio queue

Issue #1784, opened 2024-06-30
modules/misc: nixpkgs module


MattSturgeon:
NixOS[^1], home-manager[^2] and nix-darwin[^3] all have a `nixpkkgs` module where the `pkgs` passed to modules can be configured or overridden.

This can be used to specify a specific lock or channel, and it can be used to define `overlays` and config such as `enableUnfree`.
Configuring `pkgs` is also useful for cross-compiliation (by setting `hostPlatform` and `buildPlatform`).

Currently this is not possible in standalone nixvim builds, instead `pkgs` must be configured _before_ passing it to `makeNixvimWithModule`.

This would also make it possible to have nixvim's `pkgs` be different to that used by the host modules; currently many users are forced to use the standalone build _just_ to use a different nixpkgs channel to their NixOS/home-manager configs.

~~Adding this would also make it easier to add extensions to `pkgs.lib` via an overlay, such as merging our maintainers into `lib.maintainers` or making nixvim's helper lib available as (e.g.) `lib.nixvim`.~~\
EDIT: done in #1963, but it has to be done as a specialArg because `lib` is passed in by `lib.evalModules`.

[^1]: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/nixpkgs.nix
[^2]: https://github.com/nix-community/home-manager/blob/master/modules/misc/nixpkgs.nix
[^3]: https://github.com/LnL7/nix-darwin/blob/master/modules/nix/nixpkgs.nix

MattSturgeon:
Originally commented on https://github.com/nix-community/nixvim/pull/1800#issuecomment-2205696343

Looks like the way home-manager handles the "default" `pkgs` is to set the `pkgsPath` special arg, and use that to construct `pkgs`: `defaultPkgs = import pkgsPath config.nixpkgs`.

Meaning they use a fresh instance, with its own config/overlays, but matching the same locked channel.

Unless these nixos module options are used:

`home-manager.useGlobalPkgs`

> Whether to enable using the system configuration’s `pkgs` argument in Home Manager. This disables the Home Manager options `nixpkgs.*`.

`home-manager.useUserPackages`

> Whether to enable installation of user packages through the `users.users.<name>.packages` option.

I believe when home-manager is used standalone, it sets `pkgsPath` to `<nixpkgs>` when not using flakes or home-manager's `inputs.nixpkgs` otherwise.

[...]

PR #1994, opened 2024-08-08
plugins/luasnip: backport SnipMate loader

cezdro:
I created this commit some time ago (under @nawordar account), but I drive NixOS stable and the `main` branch of NixVim is not compatible anymore (missing cmp-ai package).

### DIFF
--- a/plugins/snippets/luasnip/default.nix
+++ b/plugins/snippets/luasnip/default.nix
@@ -8,6 +8,46 @@
 with lib;
 let
   cfg = config.plugins.luasnip;
+
+  loaderSubmodule = types.submodule {
+    options = {
+      lazyLoad = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether or not to lazy load the snippets
+        '';
+      };
+
+      # TODO: add option to also include the default runtimepath
+      paths =
+        helpers.mkNullOrOption
+          (
+            with helpers.nixvimTypes;
+            oneOf [
+              str
+              path
+              rawLua
+              (listOf (oneOf [
+                str
+                path
+                rawLua
+              ]))
+            ]
+          )
+          ''
+            List of paths to load.
+          '';
+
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Is this purely additions with no changes to existing options? We don't normally backport, but I'm fine with merging the occasional backport if is additive only and has no breaking changes.

I'm not sure what's going on with the CI failure, it may be unrelated.

I'd like to point out though, it is possible to use the `main` nixvim branch by doing a standalone build and then including the package in your system packages. But this is a slightly different workflow to using the nixos/hm/darwin modules.

In the future (#1784), it should be possible to tell nixvim to use a different `pkgs` to your system modules. That would also allow you to have a mixed build where you use 24.05 on the host and unstable in nixvim.

cezdro:
I'm using NixVim as a home-manager module and home-manager as a NixOS module. I tried adding `inputs.nixvim.inputs.nixpkgs.follows = "nixpkgs-unstable"`, but it didn't change anything. Do I understand correctly that home-manager still passes the 24.05 branch of nixpkgs?

Yes, this commit only adds an option and does some clean up as discussed in #1643.

MattSturgeon:
> I'm using NixVim as a home-manager module and home-manager as a NixOS module.

Yes, currently the only way to use a nixvim from unstable on a system running a stable channel is to use nixvim's standalone build method, linked above.

> I tried adding `inputs.nixvim.inputs.nixpkgs.follows = "nixpkgs-unstable"`, but it didn't change anything. Do I understand correctly that home-manager still passes the 24.05 branch of nixpkgs?

Correct, that's what #1784 is about; adding a way to configure `programs.nixvim.nixpkgs.pkgs`.

> Yes, this commit only adds an option and does some clean up as discussed in #1643.

Thanks for clarifying!

> Unfortunately, I don't have a Mac, so I can't test what is going on with the test

Yes, Darwin can be a bit of a pain; some tests just don't run well on it.

MattSturgeon (approved):

MattSturgeon:
@mergifyio queue

PR #1997, opened 2024-08-08
plugins/firenvim: fix aliasing `settings` into `globals`

MattSturgeon:
Follow up to #1983

Needed to use `options.*.settings` instead of `config.*.settings`.

I was concerned that the test-case didn't pick up on this, so I added an new test case with some basic assertions.


traxys (approved):

MattSturgeon:
@mergifyio queue

PR #1999, opened 2024-08-09
plugins/yazi: init

khaneliman:
Adding module for https://github.com/mikavilpas/yazi.nvim

NOTE: It looks like it has a dependency on telescope in the builtin functions for searching. Should that be a warning or added as an extra plugin automatically? To make this more complex, technically a person can setup the integrations property to change the grep behavior to not rely on it..

### DIFF
--- a/plugins/filetrees/yazi.nix
+++ b/plugins/filetrees/yazi.nix
@@ -1,0 +1,160 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  types = lib.nixvim.nixvimTypes;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "yazi";
+  originalName = "yazi.nvim";
+  defaultPackage = pkgs.vimPlugins.yazi-nvim;
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    Yazi is a blazing fast file manager for the terminal.
+    This plugin allows you to open yazi in a floating window in Neovim.
+
+    Some default keybindings have additional dependencies you may need to install or enable.
+    See the [upstream docs](https://github.com/mikavilpas/yazi.nvim?tab=readme-ov-file#%EF%B8%8F-keybindings) for details.
+  '';
+
+  settingsOptions = {
+    log_level = defaultNullOpts.mkLogLevel' {
+      pluginDefault = "off";
+      description = ''
+        The log level to use. Off by default, but can be used to diagnose
+        issues. You can find the location of the log file by running
+        `:checkhealth yazi` in Neovim.
+      '';
+    };
+
+    open_for_directories = defaultNullOpts.mkBool false "";
+
+    use_ya_for_events_reading = defaultNullOpts.mkBool false ''
+      Use the `ya` command to read events.
+
+      Allows more complex behaviors using the `ya` messaging system.
+    '';
+
+    use_yazi_client_id_flag = defaultNullOpts.mkBool false "Allows passing which instance of yazi is being controlled.";
+
+    enable_mouse_support = defaultNullOpts.mkBool false "Enables mouse support.";
+
+    open_file_function = defaultNullOpts.mkLuaFn' {
+      pluginDefault.__raw = ''
+        function(chosen_file)
+          vim.cmd(string.format("edit %s", vim.fn.fnameescape(chosen_file)))
+        end
+      '';
+      description = ''
+        What Neovim should do a when a file was opened (selected) in yazi.
+
+        Defaults to simply opening the file.
+      '';
+    };
+
+    clipboard_register = defaultNullOpts.mkStr "*" ''
+      Some yazi.nvim commands copy text to the clipboard. This is the register
+      yazi.nvim should use for copying. Defaults to "*", the system clipboard.
+    '';
+
+    keymaps =
+      defaultNullOpts.mkNullable (types.either types.attrs (types.enum [ false ]))
+        {
+          show_help = "<f1>";
+          open_file_in_vertical_split = "<c-v>";
+          open_file_in_horizontal_split = "<c-x>";
+          open_file_in_tab = "<c-t>";
+          grep_in_directory = "<c-s>";
+          replace_in_directory = "<c-g>";
+          cycle_open_buffers = "<tab>";
+          copy_relative_path_to_selected_files = "<c-y>";
+          send_to_quickfix_list = "<c-q>";
+        }
+        ''
+          Customize the keymaps that are active when yazi is open and focused.
+
+          Also:
+            - use e.g. `open_file_in_tab = false` to disable a keymap
+            - you can customize only some of the keymaps if you want
+            - Set to `false` to disable all default keymaps.
+        '';
+
+    set_keymappings_function = defaultNullOpts.mkLuaFn null ''
+      Completely override the keymappings for yazi. This function will be
+      called in the context of the yazi terminal buffer.
+    '';
+
+    hooks = {
+      yazi_opened = defaultNullOpts.mkLuaFn' {
+        pluginDefault.__raw = ''
+          function(preselected_path, yazi_buffer_id, config)
+          end
+        '';
+        description = ''
+          If you want to execute a custom action when yazi has been opened,
+          you can define it here.
+        '';
+      };
+
+      yazi_closed_successfully = defaultNullOpts.mkLuaFn' {
+        pluginDefault.__raw = ''
+          function(chosen_file, config, state)
+          end
+        '';
+        description = "When yazi was successfully closed";
+      };
+
+      yazi_opened_multiple_files = defaultNullOpts.mkLuaFn' {
+        pluginDefault.__raw = ''
+          function(chosen_files)
+            vim.cmd("args" .. table.concat(chosen_files, " "))
+          end
+        '';
+        description = ''
+          When yazi opened multiple files. The default is to send them to the
+          quickfix list, but if you want to change that, you can define it here
+        '';
+      };
+    };
+
+    highlight_groups = defaultNullOpts.mkAttributeSet { hovered_buffer = null; } ''
+      Add highlight groups to different yazi events.
+
+      NOTE: this only works if `use_ya_for_events_reading` is enabled, etc.
+    '';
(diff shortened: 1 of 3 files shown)


GaetanLepage:
The developer doesn't list telescope as a dependency. In that sense, I don't think that we should do anything.
Can the plugin be operated normally when simply enabling `plugins.yazi.enable` ?

khaneliman:
> The developer doesn't list telescope as a dependency. In that sense, I don't think that we should do anything. Can the plugin be operated normally when simply enabling `plugins.yazi.enable` ?

Yeah, it's fine to use most of the functionality. [Here's](https://github.com/mikavilpas/yazi.nvim/blob/4e549120e1ff2702b6ea1eff3640a879fffb2141/lua/yazi/config.lua#L46-L66) where I'm talking about.

khaneliman:
Added a description to make note about it.

MattSturgeon:
> The warning would basically have to be checking if the user hasn't overridden the `settings.integrations.grep_in_directory` then check if telescope is enabled or something and even that doesn't feel foolproof.

TBH this seems like something that upstream plugin should be dealing with. If the user configures the plugin with a mapping that needs telescope (even if it's a default mapping) the plugin should really warn that telescope is missing.

Perhaps we should report an issue upstream?

tests/test-sources/plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,68 @@
...
+        };
+
+        highlight_groups = {
+          hovered_buffer = null;
+        };
+
+        floating_window_scaling_factor = 0.9;
+        yazi_floating_window_winblend = 0;
+        yazi_floating_window_border = "rounded";
+
```
MattSturgeon, line 64:
```suggestion
```

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,161 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
+let
+
```
MattSturgeon, line 9:
```suggestion
```

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,161 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+with lib;
```
MattSturgeon, line 7: Do we need `with lib` here? I know we've done this historically, but I'd rather inherit the few things we use repeatedly and access the rest explicitly.

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,161 @@
...
+  maintainers = [ maintainers.khaneliman ];
+
+  description = ''
+    Yazi is a blazing fast file manager for the terminal.
+    This plugin allows you to open yazi in a floating window in Neovim.
+
+    The default grep functionality requires `telescope`.
+    You can override the grep functionality through the `plugins.yazi.settings.integrations.grep_in_directory`
+    and `plugins.yazi.settings.integrations.grep_in_selected_files` if you do not wish to
+    include `telescope` as an extra dependency.
```
MattSturgeon, line 27:
```suggestion
    You can override the grep functionality through the `plugins.yazi.settings.integrations.grep_in_directory`
    and `plugins.yazi.settings.integrations.grep_in_selected_files` options
    if you do not wish to use `telescope`.
    Otherwise you must enable `plugins.telescope` or install it some other way.
```

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,161 @@
...
+      NOTE: this only works if `use_ya_for_events_reading` is enabled, etc.
+    '';
+
+    floating_window_scaling_factor =
+      defaultNullOpts.mkNum 0.9
+        "The floating window scaling factor. 1 means 100%, 0.9 means 90%, etc.";
+
+    yazi_floating_window_winblend = defaultNullOpts.mkNullableWithRaw (types.ints.between 0
+      100
+    ) 0 "`0` for fully opaque and `100` for fully transparent. See :h winblend";
```
MattSturgeon, line 142: Nixfmt has done a _horrible_ job here!

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,161 @@
...
+  settingsExample = {
+    log_level = "debug";
+    open_for_directories = true;
+    enable_mouse_support = true;
+    floating_window_scaling_factor = 0.5;
+    yazi_floating_window_border = "single";
+    yazi_floating_window_winblend = 50;
+  };
+
+  extraPlugins = with pkgs.vimPlugins; [ plenary-nvim ];
```
MattSturgeon, line 160: @GaetanLepage is this a dependency that should be upstreamed to nixpkgs? Or perhaps it's already handled there?

plugins/filetrees/yazi.nix
khaneliman: Yeah... I didn't know if this was another situation where we needed to use `'` and set the options directly.

khaneliman:
```nix
    yazi_floating_window_winblend = defaultNullOpts.mkNullableWithRaw' {
      type = (types.ints.between 0 100);
      pluginDefault = 0;
      description = "`0` for fully opaque and `100` for fully transparent. See :h winblend";
    };
```

plugins/filetrees/yazi.nix
khaneliman: I didn't add it when i added the plugin. Is that just adding a dependency in ovverrides.nix?
```nix
  yazi-nvim = super.yazi-nvim.overrideAttrs {
    dependencies = with self; [ plenary-nvim ];
  };
```

MattSturgeon: > Is that just adding a dependency in overrides.nix?

Yes, I think so. Though @GaetanLepage is more familiar than I am with nixpkgs's vim/nvim contributions.

khaneliman: Created a PR for the dependency upstream and added a TODO.

khaneliman:
> > The warning would basically have to be checking if the user hasn't overridden the `settings.integrations.grep_in_directory` then check if telescope is enabled or something and even that doesn't feel foolproof.
> 
> TBH this seems like something that upstream plugin should be dealing with. If the user configures the plugin with a mapping that needs telescope (even if it's a default mapping) the plugin should really warn that telescope is missing.
> 
> Perhaps we should report an issue upstream?

Looking at the documentation on their README, it is noted in their that it uses telescope when it's available otherwise the binding does nothing.

MattSturgeon:
> Looking at the documentation on their README, it is noted in their that it uses telescope when it's available otherwise the binding does nothing. [mikavilpas/yazi.nvim#%EF%B8%8F-keybindings](https://github.com/mikavilpas/yazi.nvim?tab=readme-ov-file#%EF%B8%8F-keybindings). I think we should be fine with the description text, link to their documentation, and their documentation calling it out, too.

Well spotted. Following that link, it seems `<c-g>` has a similar dependency on [grug-far](https://github.com/MagicDuck/grug-far.nvim).

If both of these fail somewhat gracefully, as the documentation would imply, and on the basis that you've been using the plugin a while without even noticing the optional deps, I think we can reword or remove our note in the description?

khaneliman:
> > Looking at the documentation on their README, it is noted in their that it uses telescope when it's available otherwise the binding does nothing. [mikavilpas/yazi.nvim#%EF%B8%8F-keybindings](https://github.com/mikavilpas/yazi.nvim?tab=readme-ov-file#%EF%B8%8F-keybindings). I think we should be fine with the description text, link to their documentation, and their documentation calling it out, too.
> 
> Well spotted. Following that link, it seems `<c-g>` has a similar dependency on [grug-far](https://github.com/MagicDuck/grug-far.nvim).
> 
> If both of these fail somewhat gracefully, as the documentation would imply, and on the basis that you've been using the plugin a while without even noticing the optional deps, I think we can reword or remove our note in the description?

Updated description to be more clear about the external dependencies. Let me know if it's clear enough.

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,170 @@
...
+    and `plugins.yazi.settings.integrations.grep_in_selected_files` options
+    if you do not wish to use `telescope`.
+    Otherwise you must enable `plugins.telescope` or install it some other way to use this mapping.
+
+
+    The default `<c-g>` search and replace function requires `grug-far`.
+    You can override the replace functionality through the `plugins.yazi.settings.integrations.replace_in_directory`
+    and `plugins.yazi.settings.integrations.replace_in_selected_files` options
+    if you do not wish to use `grug-far`.
+    Otherwise you must install it to use this mapping.
```
MattSturgeon, line 33: Would it be easier long term to just link to the upstream docs?

Something like

```md
Some default keybindings have additional dependencies you may need to install or enable.
See the [upstream docs](https://github.com/mikavilpas/yazi.nvim?tab=readme-ov-file#%EF%B8%8F-keybindings) for details.
```

khaneliman: Yeah, fair enough. Less to maintain and keeps the ownership on upstream.

MattSturgeon (approved):
LGTM. Thanks again!

plugins/filetrees/yazi.nix
```diff
@@ -0,0 +1,160 @@
...
+    log_level = "debug";
+    open_for_directories = true;
+    enable_mouse_support = true;
+    floating_window_scaling_factor = 0.5;
+    yazi_floating_window_border = "single";
+    yazi_floating_window_winblend = 50;
+  };
+
+  # TODO: remove when https://github.com/NixOS/nixpkgs/pull/333506 is available
+  extraPlugins = with pkgs.vimPlugins; [ plenary-nvim ];
```
MattSturgeon, line 159: I have a feeling #333506 will merge soon as it's already got one approval. Shall we just wait until we update to a nixpkgs rev with the change before merging so we don't have to drop this line in another PR?

Don't really mind either way.

khaneliman: Usually a couple days delay before getting to unstable after merge and then whenever we get to flake update. I dont mind keeping track of the nixpkgs PR and when we do another flake.lock update. I usually do these kind of TODO's in my own config with overlays while waiting for fixes from nixos-unstable.

MattSturgeon:
@mergifyio queue

PR #2008, opened 2024-08-10
update-scripts: Adapt to rust-analyzer option changes

traxys:
A newly introduced option added two unhandled cases:
- Sub-module like `properties`
- Descriptions containing markdown headings

In order to avoid shelling out to pandoc to filter _each_ markdown description we only check if there _could_ be an header, and call pandoc in that case. This currently results in pandoc being only called for the newly introduced option

### DIFF
--- a/generated/rust-analyzer.nix
+++ b/generated/rust-analyzer.nix
@@ -172,7 +172,7 @@
       targets and features, with the following base command line:
 
       ```bash
-      cargo check --quiet --workspace --message-format=json --all-targets
+      cargo check --quiet --workspace --message-format=json --all-targets --keep-going
       ```
       .
     '';
@@ -1323,7 +1323,7 @@
     description = ''
       Whether to show const generic parameter name inlay hints.
     '';
-    pluginDefault = false;
+    pluginDefault = true;
     type = {
       kind = "boolean";
     };
@@ -1332,7 +1332,7 @@
     description = ''
       Whether to show generic lifetime parameter name inlay hints.
     '';
-    pluginDefault = true;
+    pluginDefault = false;
     type = {
       kind = "boolean";
     };
@@ -1649,7 +1649,7 @@
     pluginDefault = null;
     type = {
       kind = "integer";
-      maximum = null;
+      maximum = 65535;
       minimum = 0;
     };
   };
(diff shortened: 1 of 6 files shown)


update-scripts/rust-analyzer/default.nix
```diff
@@ -124,6 +134,18 @@ let
       type ? null,
     }:
     let
+      filteredMarkdownDesc =
+        # If there is a risk that the string contains an heading filter it out
```
MattSturgeon, line 138: Can't we just do a line-wize regex replace converting headings to bold lines? Pandoc feels like overkill here just to filter out lines starting with `#`

MattSturgeon: Essentially `sed -e 's/^#\+ \(.*\)/**\1**/`

MattSturgeon: We should be able to do a replacement like that in nix, without resorting to another IFD

traxys: Well yeah, but that would risk changing stuff in code blocks

MattSturgeon: Ah I didn't think of code blocks! This is fine for now then I guess.

I'll look in a bit more detail once I get back

traxys: We could make an option in pure nix, but it would be a more complicated implementation.

I could look into it, maybe it would not be as complicated as I think

GaetanLepage:
Seems good overall. Your change in the lua file is spreading over the 2 commits (minor nit).

MattSturgeon:
> Your change in the lua file is spreading over the 2 commits (minor nit).

To be precise, `heading_filter.lua` is added in the first commit, but is only used in the second.

update-scripts/rust-analyzer/heading_filter.lua
MattSturgeon, line 1: If we're adding a lua file, I think we should be adding stylua to our treefmt config

update-scripts/rust-analyzer/default.nix
```diff
@@ -124,6 +134,18 @@ let
...
+        if lib.hasInfix "# " markdownDescription then
+          builtins.readFile (
+            runCommand "filteredDocumentation" { inherit markdownDescription; } ''
```
MattSturgeon, line 141: Nit: derivation names are usually kebab-case

update-scripts/rust-analyzer/default.nix
MattSturgeon: TBH I don't think it's worth spending time on. This is only really built on CI which doesn't seem to be struggling currently.

update-scripts/rust-analyzer/heading_filter.lua
traxys: I think we already have it

traxys: (https://github.com/nix-community/nixvim/blob/4852f94f8ccae551514df0092a077014bafb95ca/flake-modules/dev/default.nix#L36)

MattSturgeon: Strange, perhaps stylua is a very liberal formatter; the first thing I noticed is that our init.lua files are indented with spaces but the file here is tabs. I figured that'd be something stylua enforces...

generated/rust-analyzer.nix
```diff
@@ -1985,6 +1985,133 @@
...
+                      {arg}
+              ],
+              "progressLabel": "rust-analyzer",
+              "filesToWatch": [
+                      "BUCK",
+              ],
+      }
+      ```
+
+      *On `DiscoverWorkspaceConfig::command`*
```
MattSturgeon, line 2015: I think bold would be better than italic for a header
```suggestion
      **On `DiscoverWorkspaceConfig::command`**
```

plugins/lsp/language-servers/rust-analyzer-config.nix
```diff
@@ -37,6 +37,10 @@ let
         lib.types.int
     else if kind == "object" then
       lib.types.attrsOf lib.types.anything
+    else if kind == "submodule" then
+      lib.types.submodule {
+        options = lib.mapAttrs (_: ty: lib.mkOption { type = mkRustAnalyzerType ty; }) typeInfo.options;
```
MattSturgeon, line 42: I believe @GaetanLepage spent time ensuring all options have a description to avoid some warning being printed somewhere.

IDK if this matters for submodule sub-options, but perhaps we should have something?

generated/rust-analyzer.nix
```diff
@@ -1985,6 +1985,133 @@
       kind = "boolean";
     };
   };
+  "rust-analyzer.workspace.discoverConfig" = {
+    description = ''
+      Enables automatic discovery of projects using
+      \[`DiscoverWorkspaceConfig::command`\].
+
+      \[`DiscoverWorkspaceConfig`\] also requires setting `progress_label` and
```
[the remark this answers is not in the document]
traxys: I think yeah, they are `rustdoc` links, not proper markdown links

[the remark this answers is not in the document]
traxys: To resolve the links you need the rust compiler in essence

plugins/lsp/language-servers/rust-analyzer-config.nix
traxys: I have no idea what description I could provide, an empty description?

generated/rust-analyzer.nix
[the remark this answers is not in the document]
MattSturgeon: In that case are better off removing the markdown link syntax instead of escaping it?

Not really a big deal either way.

plugins/lsp/language-servers/rust-analyzer-config.nix
MattSturgeon: I think empty description is fine, best to check with @GaetanLepage whether we actually need to do this or not

generated/rust-analyzer.nix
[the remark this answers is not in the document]
traxys: I'm not sure pandoc recognises it as links, so I'm not sure we can do much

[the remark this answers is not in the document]
traxys: https://github.com/rust-lang/rust-analyzer/blob/0daeb5c0b05cfdf2101b0f078c27539099bf38e6/crates/rust-analyzer/src/config.rs#L325

Upstream has *exactly* the same markdown.

The generated rustdoc output is [here](https://rust-lang.github.io/rust-analyzer/rust_analyzer/config/index.html) but it seems that the documented item is private, so not parsed by default by rustdoc

generated/rust-analyzer.nix
traxys: fixed (and in the lua filter)

update-scripts/rust-analyzer/heading_filter.lua
traxys: Yeah stylua seems to not transform spaces to tabs (I called explicitly `stylua` on the file, with tabs & spaces)

MattSturgeon (approved):
No major issues from me. Thanks for dealing with this chore @traxys!

traxys:
@Mergifyio queu

Issue #1890, opened 2024-07-18
[BUG] `Bufferline` `diagnostics_update_in_insert` is deprecated


pinarruiz:
| Field        | Description |
|--------------|-------------|
| Plugin       | [`Bufferline.nvim`](https://nix-community.github.io/nixvim/plugins/bufferline/index.html)  | 
| Nixpkgs      | [`unstable`](https://github.com/NixOS/nixpkgs/tree/ad0b5eed1b6031efaed382844806550c3dcb4206) ([`ad0b5eed1b6031efaed382844806550c3dcb4206`](https://github.com/NixOS/nixpkgs/tree/ad0b5eed1b6031efaed382844806550c3dcb4206)) | 

- [X] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

When using the bufferline option [`plugins.bufferline.diagnosticsUpdateInInsert`](https://nix-community.github.io/nixvim/plugins/bufferline/index.html#pluginsbufferlinediagnosticsupdateininsert), i get the following error:

```
diagnostics_update_in_insert is deprecated, use vim.diagnostic.config { update_in_insert = true } instead.
Feature will be removed in bufferline 4.6.3
```

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim.plugins.bufferline = {
  enable = true;
  diagnostics = "nvim_lsp";
  diagnosticsUpdateInInsert = true;
  diagnosticsIndicator = ''
    function(count, level, diagnostics_dict, context)
      local s = " "
      for e, n in pairs(diagnostics_dict) do
        local sym = e == "error" and "󰅚 "
          or (e == "warning" and "󰀪 " or 
            ( e == "hint" and "󰌶 " or "󰋽 ")
          )
        s = s .. n .. sym
      end
      return s
    end
  '';
};
```

Also, out of scope, might be cool to merge all this bufferline diagnostics options into one.

PR #2013, opened 2024-08-12
plugins/bufferline: migrate to mkNeovimPlugin

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1890

Also added support:
-  `separator_style` for the custom separator declaration using an array of 2 strings. 
-  `indicator.style` allow for `none`

Changed:
- `diagnostics` no longer accepts true. Plugin only accepts `false`, `nvim_lsp`, or `coc`.
- `diagnosticsUpdateInInsert` uses `vim.diagnostic.config` instead of passing into options table. 
- Updated some descriptions with notes from upstream docs on usage.

### DIFF
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -46,6 +46,18 @@ rec {
         '';
     };
 
+  mkSettingsRenamedOptionModules =
+    oldPrefix: newPrefix:
+    map (
+      option':
+      let
+        option = toList option';
+        oldPath = oldPrefix ++ option;
+        newPath = newPrefix ++ map nixvim.toSnakeCase option;
+      in
+      mkRenamedOptionModule oldPath newPath
+    );
+
   # A clone of types.coercedTo, but it prints a warning when oldType is used.
   transitionType =
     oldType: coerceFn: newType:
--- a/plugins/bufferlines/bufferline.nix
+++ b/plugins/bufferlines/bufferline.nix
@@ -1,184 +1,306 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.bufferline;
-
-  highlightOptions = {
-    fill = "fill";
-    background = "background";
-
-    tab = "tab";
-    tab_selected = "tabSelected";
-    tab_separator = "tabSeparator";
-    tab_separator_selected = "tabSeparatorSelected";
-    tab_close = "tabClose";
-
-    close_button = "closeButton";
-    close_button_visible = "closeButtonVisible";
-    close_button_selected = "closeButtonSelected";
-
-    buffer_visible = "bufferVisible";
-    buffer_selected = "bufferSelected";
-
-    numbers = "numbers";
-    numbers_visible = "numbersVisible";
-    numbers_selected = "numbersSelected";
-
-    diagnostic = "diagnostic";
-    diagnostic_visible = "diagnosticVisible";
-    diagnostic_selected = "diagnosticSelected";
-
-    hint = "hint";
-    hint_visible = "hintVisible";
-    hint_selected = "hintSelected";
-
-    hint_diagnostic = "hintDiagnostic";
-    hint_diagnostic_visible = "hintDiagnosticVisible";
-    hint_diagnostic_selected = "hintDiagnosticSelected";
-
-    info = "info";
-    info_visible = "infoVisible";
-    info_selected = "infoSelected";
-
-    info_diagnostic = "infoDiagnostic";
-    info_diagnostic_visible = "infoDiagnosticVisible";
-    info_diagnostic_selected = "infoDiagnosticSelected";
-
-    warning = "warning";
-    warning_visible = "warningVisible";
-    warning_selected = "warningSelected";
-
-    warning_diagnostic = "warningDiagnostic";
-    warning_diagnostic_visible = "warningDiagnosticVisible";
-    warning_diagnostic_selected = "warningDiagnosticSelected";
-
-    error = "error";
-    error_visible = "errorVisible";
-    error_selected = "errorSelected";
-
-    error_diagnostic = "errorDiagnostic";
-    error_diagnostic_visible = "errorDiagnosticVisible";
-    error_diagnostic_selected = "errorDiagnosticSelected";
-
-    modified = "modified";
-    modified_visible = "modifiedVisible";
-    modified_selected = "modifiedSelected";
-
-    duplicate = "duplicate";
-    duplicate_visible = "duplicateVisible";
-    duplicate_selected = "duplicateSelected";
-
-    separator = "separator";
-    separator_visible = "separatorVisible";
-    separator_selected = "separatorSelected";
-
-    indicator_visible = "indicatorVisible";
-    indicator_selected = "indicatorSelected";
-
-    pick = "pick";
-    pick_visible = "pickVisible";
-    pick_selected = "pickSelected";
-
-    offset_separator = "offsetSeparator";
-
-    trunc_marker = "trunkMarker";
-  };
+  inherit (lib.nixvim) defaultNullOpts nixvimTypes mkSettingsRenamedOptionModules;
+  types = nixvimTypes;
 in
-{
-  options = {
-    plugins.bufferline = helpers.neovim-plugin.extraOptionsOptions // {
-      enable = mkEnableOption "bufferline";
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "bufferline";
+  originalName = "bufferline.nvim";
+  defaultPackage = pkgs.vimPlugins.bufferline-nvim;
 
-      package = helpers.mkPluginPackageOption "bufferline" pkgs.vimPlugins.bufferline-nvim;
+  maintainers = [ lib.maintainers.khaneliman ];
 
-      mode = helpers.defaultNullOpts.mkEnumFirstDefault [
-        "buffers"
-        "tabs"
-      ] "mode";
+  # TODO: introduced 2024-08-12: remove after 24.11
+  #
+  # NOTE: Old options are equivalent to `settings.options` and
+  # the old highlight options are equivalent to `settings.highlight`,
+  # therefore we can't just use `optionsRenamedToSettings`.
+  imports =
+    let
+      oldOptions = [
+        "mode"
+        "themable"
+        "numbers"
+        "bufferCloseIcon"
+        "modifiedIcon"
+        "closeIcon"
+        "closeCommand"
+        "leftMouseCommand"
+        "rightMouseCommand"
+        "middleMouseCommand"
+        "indicator"
+        "leftTruncMarker"
+        "rightTruncMarker"
+        "separatorStyle"
+        "nameFormatter"
+        "truncateNames"
+        "tabSize"
+        "maxNameLength"
+        "colorIcons"
+        "showBufferIcons"
+        "showBufferCloseIcons"
+        "getElementIcon"
+        "showCloseIcon"
+        "showTabIndicators"
+        "showDuplicatePrefix"
+        "enforceRegularTabs"
+        "alwaysShowBufferline"
+        "persistBufferSort"
+        "maxPrefixLength"
+        "sortBy"
+        "diagnostics"
+        "diagnosticsIndicator"
+        "offsets"
+        [
+          "groups"
+          "items"
+        ]
+        [
+          "groups"
+          "options"
+          "toggleHiddenOnEnter"
+        ]
+        [
+          "hover"
+          "enabled"
+        ]
+        [
+          "hover"
+          "reveal"
+        ]
+        [
+          "hover"
+          "delay"
+        ]
+        [
+          "debug"
+          "logging"
+        ]
+        "customFilter"
+      ];
 
-      themable = helpers.defaultNullOpts.mkBool true "Whether or not bufferline highlights can be overridden externally";
+      oldHighlightOptions = [
+        "fill"
+        "background"
+        "tab"
+        "tabSelected"
+        "tabSeparator"
+        "tabSeparatorSelected"
+        "tabClose"
+        "closeButton"
+        "closeButtonVisible"
+        "closeButtonSelected"
+        "bufferVisible"
+        "bufferSelected"
+        "numbers"
+        "numbersVisible"
+        "numbersSelected"
+        "diagnostic"
+        "diagnosticVisible"
+        "diagnosticSelected"
+        "hint"
+        "hintVisible"
+        "hintSelected"
+        "hintDiagnostic"
+        "hintDiagnosticVisible"
+        "hintDiagnosticSelected"
+        "info"
+        "infoVisible"
+        "infoSelected"
+        "infoDiagnostic"
+        "infoDiagnosticVisible"
+        "infoDiagnosticSelected"
+        "warning"
+        "warningVisible"
+        "warningSelected"
+        "warningDiagnostic"
+        "warningDiagnosticVisible"
+        "warningDiagnosticSelected"
+        "error"
+        "errorVisible"
+        "errorSelected"
+        "errorDiagnostic"
+        "errorDiagnosticVisible"
+        "errorDiagnosticSelected"
+        "modified"
+        "modifiedVisible"
+        "modifiedSelected"
+        "duplicate"
+        "duplicateVisible"
+        "duplicateSelected"
+        "separator"
+        "separatorVisible"
+        "separatorSelected"
+        "indicatorVisible"
+        "indicatorSelected"
+        "pick"
+        "pickVisible"
+        "pickSelected"
+        "offsetSeparator"
+      ];
+
+      basePluginPath = [
+        "plugins"
+        "bufferline"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+      optionsPath = settingsPath ++ [ "options" ];
+      oldHighlightsPath = basePluginPath ++ [ "highlights" ];
+      newHighlightsPath = settingsPath ++ [ "highlights" ];
+    in
+    [
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "extraOptions" ]) optionsPath)
+      (lib.mkRenamedOptionModule (basePluginPath ++ [ "diagnosticsUpdateInInsert" ]) [
+        "diagnostics"
+        "update_in_insert"
+      ])
+      (lib.mkRenamedOptionModule (oldHighlightsPath ++ [ "trunkMarker" ]) (
+        newHighlightsPath ++ [ "trunc_marker" ]
+      ))
+    ]
+    ++ mkSettingsRenamedOptionModules basePluginPath optionsPath oldOptions
+    ++ mkSettingsRenamedOptionModules oldHighlightsPath newHighlightsPath oldHighlightOptions;
+
+  settingsOptions = {
+    options = {
+      mode =
+        defaultNullOpts.mkEnumFirstDefault
+          [
+            "buffers"
+            "tabs"
+          ]
+          ''
+            Mode - set to `tabs` to only show tabpages instead.
+          '';
+
+      themable = defaultNullOpts.mkBool true ''
+        Whether or not bufferline highlights can be overridden externally.
+      '';
 
       numbers =
-        helpers.defaultNullOpts.mkNullable
-          (
-            with types;
-            either (enum [
-              "none"
-              "ordinal"
-              "buffer_id"
-              "both"
-            ]) helpers.nixvimTypes.rawLua
-          )
-          "none"
+        defaultNullOpts.mkEnumFirstDefault
+          [
+            "none"
+            "ordinal"
+            "buffer_id"
+            "both"
+          ]
           ''
             Customize the styling of numbers.
 
-            Either one of "none" "ordinal" "buffer_id" "both" or a lua function:
+            Can also be a lua function:
             ```
             function({ ordinal, id, lower, raise }): string
             ```
           '';
 
-      bufferCloseIcon = helpers.defaultNullOpts.mkStr "" "The close icon for each buffer.";
+      buffer_close_icon = defaultNullOpts.mkStr "" ''
+        The close icon for each buffer.
+      '';
 
-      modifiedIcon = helpers.defaultNullOpts.mkStr "●" "The icon indicating a buffer was modified.";
+      modified_icon = defaultNullOpts.mkStr "●" ''
+        The icon indicating a buffer was modified.
+      '';
 
-      closeIcon = helpers.defaultNullOpts.mkStr "" "The close icon.";
+      close_icon = defaultNullOpts.mkStr "" "The close icon.";
 
-      closeCommand = helpers.defaultNullOpts.mkStr "bdelete! %d" "Command or function run when closing a buffer.";
+      close_command = defaultNullOpts.mkStr "bdelete! %d" ''
+        Command or function run when closing a buffer.
+      '';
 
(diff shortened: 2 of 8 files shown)


plugins/bufferlines/bufferline.nix
```diff
@@ -1,320 +1,435 @@
...
-        require('bufferline').setup${helpers.toLuaObject setupOptions}
+  callSetup = false;
+  extraConfig = cfg: {
+    extraConfigLua =
+      let
+        setupOptions = {
+          options = cfg.settings;
+
+          inherit (cfg) highlights;
+        };
```
MattSturgeon, line 421: We wouldn't normally do this; the `settings` option should literally be "options provided to `setup`".

In other words, we should have `settings.options` and `settings.highlights`.

khaneliman: @MattSturgeon I tried setting up the renames to go under the settings.options and settings.highlights, but am getting `error: invalid regular expression '.*vim.loader.disable().*'` running the bufferline test. Not sure if you can spot something I'm missing here.

khaneliman: Well, I was getting that on my macbook... not getting it on my desktop..

khaneliman:
@GaetanLepage I see you touched this a bit before, so would also like your eyes.

GaetanLepage:
I will have a look later today, thanks !

plugins/bufferlines/bufferline.nix
MattSturgeon: > @MattSturgeon I tried setting up the renames to go under the settings.options and settings.highlights, but am getting `error: invalid regular expression '.*vim.loader.disable().*'` running the bufferline test. Not sure if you can spot something I'm missing here.

Interesting, we'd need the `--show-trace` output to dig into that a bit more. Have you pushed the change that caused the issue?

At first glance, it looks like the string `"vim.loader.disable()"` is being converted into a regex without being properly escaped. Though the actual regex looks to just be re-implementing `lib.hasInfix` anyway.

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+      max_prefix_length = defaultNullOpts.mkInt 15 ''
+        Maximum prefix length used when a buffer is de-duplicated.
+      '';
 
-      diagnosticsIndicator = helpers.defaultNullOpts.mkLuaFn null "Either `null` or a function that returns the diagnostics indicator.";
+      sort_by = defaultNullOpts.mkStr "id" ''
+        How to sort the buffers.
 
-      diagnosticsUpdateInInsert = helpers.defaultNullOpts.mkBool true "Whether diagnostics should update in insert mode";
+        Possible values: `insert_after_current`, `insert_at_end`, `id`, `extension`, `relative_directory`, `directory`, `tabs`.
```
MattSturgeon, line 374: This looks a lot like an enum at first glance? Or is that a non-exhaustive list of examples?

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
-          toggleHiddenOnEnter = helpers.defaultNullOpts.mkBool true "Re-open hidden groups on bufenter.";
+          toggle_hidden_on_enter = defaultNullOpts.mkBool true ''
+            Re-open hidden groups on `BufEnter`.
+          '';
         };
       };
 
       hover = {
-        enabled = mkEnableOption "hover";
+        enabled = lib.mkEnableOption "hover";
```
MattSturgeon, line 412: This should almost certainly be `defaultNullOpts.mkBool false "Whether to enable hover."`

`mkEnableOption` sets the default to `false` instead of `null`.

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
-      highlights = genAttrs (attrValues highlightOptions) (
+    highlights =
+      let
+        newHighlightsOptions = map (
+          option:
+          let
+            optionSnakeCase = lib.nixvim.toSnakeCase option;
+          in
+          optionSnakeCase
+        ) (oldHighlightOptions ++ [ "truncMarker" ]);
```
MattSturgeon, line 432: I don't think we should be using the _old_ option names here? Firstly it seems wrong that we have to convert them to snake_case, additionally we're delaying the inevitable needing to add a new option without affecting the old option deprecations.

We're already _kinda_ running into this issue with `trunkMarker` vs `trunc_marker`.

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+
+  extraOptions = {
+    diagnosticsUpdateInInsert = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to enable diagnostics update in insert mode.
+
+        Only applies when `coc` is selected in option `plugins.settings.options.diagnostic`.
+      '';
```
MattSturgeon, line 450: I think this should be deprecated in favour of our diagnostics option: https://github.com/nix-community/nixvim/blob/dbf6f7bc997dc3a9ab1f014ea075600357226950/modules/diagnostics.nix#L10-L18

E.g:

```nix
(mkRenameOptionModule (basePluginPath ++ [ "diagnosticsUpdateInInsert" ] [
  "diagnostics"
  "update_in_insert"
])
```

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+          fg = commonFgColor;
+        };
+        separator_visible = {
+          fg = commonFgColor;
+        };
+        separator_selected = {
+          bg = commonBgColor;
+          fg = commonFgColor;
+        };
+      };
```
MattSturgeon, line 592: This feels a little overkill for an example?

If users want extensive docs they should go upstream, our example should just give a taste of how to write the config in nix syntax.

I don't mind keeping it, I just wonder if it'd be a little overwhelming for users?

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+    extraConfigLua =
+      lib.mkIf cfg.diagnosticsUpdateInInsert # Lua
+        ''
+          vim.diagnostic.config { update_in_insert = true }
+        '';
+
+    extraPlugins = with pkgs.vimPlugins; [
+      cfg.package
+      nvim-web-devicons
+    ];
```
MattSturgeon, line 605: `cfg.package` is already added by `mkNeovimPlugin`
```suggestion
    extraPlugins = [ pkgs.vimPlugins.nvim-web-devicons ];
```

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+      };
+  };
+
+  extraConfig = cfg: {
+    extraConfigLua =
+      lib.mkIf cfg.diagnosticsUpdateInInsert # Lua
+        ''
+          vim.diagnostic.config { update_in_insert = true }
+        '';
+
```
MattSturgeon, line 601: See above
```suggestion
```

tests/test-sources/plugins/bufferlines/bufferline.nix
```diff
@@ -6,97 +6,114 @@
...
+              local icon, hl = require('nvim-web-devicons').get_icon_by_filetype(opts.filetype, { default = false })
+              return icon, hl
+            end
+          '';
+          separator_style = [
+            "|"
+            "|"
+          ];
+        };
+      };
```
MattSturgeon, line 47: This example seems useful (at first glance), can any of it be integrated into the `settingsExample`?

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+      '';
 
-      closeCommand = helpers.defaultNullOpts.mkStr "bdelete! %d" "Command or function run when closing a buffer.";
+      custom_filter = defaultNullOpts.mkLuaFn null ''
+        ```
+        fun(buf: number, bufnums: number[]): boolean
 
-      leftMouseCommand = helpers.defaultNullOpts.mkStr "buffer %d" "Command or function run when clicking on a buffer.";
+        NOTE: this will be called a lot so don't do any heavy processing here.
+        ```
```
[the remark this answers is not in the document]
khaneliman: This looks like an accident, it has a code block in it from before and I just added a note from their docs about how often it's used.

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +215,395 @@ in
...
+        style =
+          defaultNullOpts.mkEnumFirstDefault
+            [
+              "icon"
+              "underline"
+              "none"
+            ]
+            ''
+              Indicator style.
+            '';
```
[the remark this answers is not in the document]
khaneliman: Yeah, I just started doing it all indented strings because of nits in the previous PR. I think for these blocks I agree it makes sense to be a short string for formatting. Some of the individual options is does seem easier to read with indented strings.

plugins/bufferlines/bufferline.nix
khaneliman: Ah yeah, I meant to come back and make this an enum, I believe. Thanks.

plugins/bufferlines/bufferline.nix
khaneliman: I was worried about that too, but the docs just show the generated attributes.
image

 It's just my personal highlights config to fix transparent catppuccin theme issues. I will pare it down so it's not so long, though.

tests/test-sources/plugins/bufferlines/bufferline.nix
khaneliman: Yeah, makes sense. I'll just copy them over so it's more visible.

plugins/bufferlines/bufferline.nix
khaneliman: Thanks, missed that.

plugins/bufferlines/bufferline.nix
```diff
@@ -1,121 +1,211 @@
...
+      basePluginPath = [
+        "plugins"
+        "bufferline"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+      optionsPath = settingsPath ++ [ "options" ];
+      oldHighlightsPath = basePluginPath ++ [ "highlights" ];
+      newHighlightsPath = settingsPath ++ [ "highlights" ];
+
+      processRenameOptions =
```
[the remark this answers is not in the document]
khaneliman: I thought about that while doing this and didn't know how you all would feel about it. I like the idea of allowing a module more granular control of the renames like this (when needed).

plugins/bufferlines/bufferline.nix
khaneliman: Yeah, fair enough. My intent was to reduce size of the module with more verbose option declarations. Since its in the `settingsOptions` it allows a user to specify anything not called outright so I wasn't too concerned about adding more options, yet.

But it still does feel like a kicking the can down the road solution...

plugins/bufferlines/bufferline.nix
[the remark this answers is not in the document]
khaneliman: Moved to the lib.deprecation module, just left existing name, for now. But, assume we'd want to change it.

plugins/bufferlines/bufferline.nix
khaneliman: Not sure what was going on with it, but after rebooting I can run the tests on my macbook again.

plugins/bufferlines/bufferline.nix
MattSturgeon: > I was worried about that too, but the docs just show the generated attributes.

Yes, we just assign the value to `example` and then the docs print it using `lib.generators.toPretty`.

If you wanted to show a function application you could use `lib.literalExpression`, but I'm assuming you would rather keep the example simple.

> I will pare it down so it's not so long, though.

Yes, I was more concerned about the size since I was already aware the function calls wouldn't show in the docs.

lib/deprecation.nix
```diff
@@ -46,6 +46,21 @@ rec {
         '';
     };
 
+  processRenameOptions =
+    options: oldPath: newPath:
+    (map (
+      option:
+      let
+        result =
```
MattSturgeon, line 54: `let result; in result` is redundant (unless you're re-implementing `lib.fix`), I think we can drop the outer let block here?

lib/deprecation.nix
```diff
@@ -46,6 +46,21 @@ rec {
         '';
     };
 
+  processRenameOptions =
+    options: oldPath: newPath:
```
MattSturgeon, line 50: I'd rather re-arrange the arguments so that `options` is last.

In fact, if we drop the unnecessary parentheses, we could drop the explicit `options` argument entirely and just return `map (x: /* omitted */)`, which will implicitly add the additional arg by returning a curried `map` function.

```nix
oldPath: newPath:
map (
  option:
  let
    # omitted
  in
    # omitted
);
```

lib/deprecation.nix
```diff
@@ -46,6 +46,21 @@ rec {
         '';
     };
 
+  processRenameOptions =
```
MattSturgeon, line 49: You mentioned before that I'm this may not be the best name. Do you have any ideas we can discuss for a better name?

Maybe something that implies it's meant for use renaming to `settings` options? Or that it converts to snake case?

Perhaps `mkSettingsRenameOptionModules`? Or something along those lines?

lib/deprecation.nix
MattSturgeon: On the subject of arguments, it might be better to use `oldPrefix` and `newPrefix`, allowing us to do this within the mapping function:

```nix
let
  optionPath = toList option;
  oldPath = oldPrefix ++ optionPath;
  newPath = newPrefix ++ toSnakeCase optionPath;
in
```

lib/deprecation.nix
khaneliman: I changed it to that right now while doing the other clean up. Didn't know if the name should be more abstract since it's technically not limited to the settings.

MattSturgeon: > Didn't know if the name should be more abstract since it's technically not limited to the settings.

Yeah I agree, my proposal isn't ideal because it can be more general. But in practice where else will we do this and also want to convert to snake_case?

> `mkPrefixedRenamedOptionModulesWithSnakeCase` seems more accurate but kinda wordy..

Yep, lol

khaneliman: > Yeah I agree, my proposal isn't ideal because it can be more general. But in practice where else will we do this and also want to convert to snake_case?
> 
I'm fine with this way, unless someone can come up with something better and we could rename it since it's only got a few usages.

lib/deprecation.nix
MattSturgeon, line 1: I wonder if we can improve the commit message slightly. Probably not worth too much stress over it though.

Perhaps 

```markdown
lib/deprecation: expose `mkSettingsRenamedOptionModules` publicly

Extracted the `optionsRenamedToSettings` implementation from
`mkVimPlugin` and `mkNeovimPlugin` into a new public helper.
```

?

plugins/bufferlines/bufferline.nix
MattSturgeon, line 1: I don't see why `plugins/bufferline: resolve diagnosticsUpdateInInsert deprecation` needs to be a separate commit?

If you do want to keep it separate, we should keep the option in `extraOptions` in the parent commit then deprecate it in the following one.

But my preference would just be to squash it and maybe add a note of any special/interesting deprecations in the commit message body.

plugins/bufferlines/bufferline.nix
```diff
@@ -1,121 +1,192 @@
...
+        "bufferline"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+      optionsPath = settingsPath ++ [ "options" ];
+      oldHighlightsPath = basePluginPath ++ [ "highlights" ];
+      newHighlightsPath = settingsPath ++ [ "highlights" ];
+    in
+    [ (lib.mkRenamedOptionModule (basePluginPath ++ [ "extraOptions" ]) optionsPath) ]
+    ++ (mkSettingsRenamedOptionModules basePluginPath optionsPath oldOptions)
+    ++ (mkSettingsRenamedOptionModules oldHighlightsPath newHighlightsPath oldHighlightOptions)
```
MattSturgeon, line 164: I don't mind if you think they add clarity, but these parenthesis are redundant. (They would be needed if the expressions were list items, since the list syntax uses whitespace as a deliminator)

```suggestion
    ++ mkSettingsRenamedOptionModules basePluginPath optionsPath oldOptions
    ++ mkSettingsRenamedOptionModules oldHighlightsPath newHighlightsPath oldHighlightOptions
```

MattSturgeon: Also, should we consider having the `trunc_marker` and `extraOptions` in the same `[ ]` list, so that we don't have multiple list literals in the concatenation?

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +196,465 @@ in
...
+              ])
+              rawLua
+            ]
+          )
+          "id"
+          ''
+            How to sort the buffers.
+
+            Possible values: `insert_after_current`, `insert_at_end`, `id`, `extension`, `relative_directory`, `directory`, `tabs`.
+            Also accepts a function with a signature `function(buffer_a, buffer_b)` allowing you to compare with custom logic.
```
MattSturgeon, line 375: Seems redundant to have the possible values in the enum type _and_ in the description.

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +196,465 @@ in
...
+                "insert_at_end"
+                "id"
+                "extension"
+                "relative_directory"
+                "directory"
+                "tabs"
+              ])
+              rawLua
+            ]
+          )
```
MattSturgeon, line 369: I'm confused, is this a typo?

We have `mkNullableWithRaw`, which wraps the type in `maybeRaw` (aka `either rawLua`) but then we also have a `oneOf [ rawLua ]`?

plugins/bufferlines/bufferline.nix
MattSturgeon: Maybe we could also define an actual example with a `__raw` function?

plugins/bufferlines/bufferline.nix
khaneliman: Looks like it got left in there when I did mkNullable first. I'll change it over

plugins/bufferlines/bufferline.nix
khaneliman: Yeah, I can try that. I think there was an iteration during development where the order of execution did matter.

plugins/bufferlines/bufferline.nix
MattSturgeon: Was it related to this https://github.com/nix-community/nixvim/pull/2013#discussion_r1715172032?

khaneliman: That probably could have thrown me, too.

plugins/bufferlines/bufferline.nix
MattSturgeon: Just to clarify, I was talking about an option example, not a settings-example. I'm fine with the way you've done it, but just as a reminder you can also do:

```nix
mkFoo' {
  description = "";
  example = {
    foo = "bar";
  };
}
```

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +199,467 @@ in
...
-        "padded_slope"
-        "thick"
-        "thin"
-      ] "thin" "Separator style";
+      right_trunc_marker = defaultNullOpts.mkStr "" ''
+        Right truncation marker.
+      '';
 
-      nameFormatter = helpers.defaultNullOpts.mkLuaFn null ''
+      separator_style = defaultNullOpts.mkNullableWithRaw (
```
MattSturgeon, line 258: We don't need _both_ `mkNullableWithRaw` _and_ `oneOf [ rawLua ]`.

Either use `mkNullable` with `oneOf`, or use `mkNullableWithRaw` with `either`

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +199,467 @@ in
...
+      max_prefix_length = defaultNullOpts.mkInt 15 ''
+        Maximum prefix length used when a buffer is de-duplicated.
+      '';
 
-      diagnosticsUpdateInInsert = helpers.defaultNullOpts.mkBool true "Whether diagnostics should update in insert mode";
+      sort_by =
+        defaultNullOpts.mkNullableWithRaw
+          (
+            with types;
+            enum [
```
MattSturgeon, line 355:
```suggestion
            types.enum [
```

plugins/bufferlines/bufferline.nix
```diff
@@ -125,196 +199,467 @@ in
             ```
```
MattSturgeon, line 199: This option also has the enum values in the description, not a big deal though

MattSturgeon (approved):
I'm happy, thanks for your work on this!

Two more nits below, but they're not too important.

khaneliman:
@mergifyio queue

GaetanLepage (approved):
LGTM, thanks !

GaetanLepage:
@Mergifyio queue

ajaxbits:
Seeing this bug:
```
┏━ 1 Errors:
 ⋮
┃            39|       paths = [
┃              |       ^
┃            40|         config.finalPackage
┃
┃        (stack trace truncated; use '--show-trace' to show the full trace)
┃
┃        error: attribute 'nixvim' missing
┃        at /nix/store/vdwmiyrhb8as39shk5hgqlb5p1g7ra21-source/plugins/bufferlines/bufferline.nix:11:1:
┃            10| in
┃            11| lib.nixvim.neovim-plugin.mkNeovimPlugin config {
┃              | ^
┃            12|   name = "bufferline";
┣━━━
┗━ ∑ ⚠ Exited with 1 errors reported by nix at 13:28:01 after 1s
```

khaneliman:
> Seeing this bug:
> 
> ```
> ┏━ 1 Errors:
>  ⋮
> ┃            39|       paths = [
> ┃              |       ^
> ┃            40|         config.finalPackage
> ┃
> ┃        (stack trace truncated; use '--show-trace' to show the full trace)
> ┃
> ┃        error: attribute 'nixvim' missing
> ┃        at /nix/store/vdwmiyrhb8as39shk5hgqlb5p1g7ra21-source/plugins/bufferlines/bufferline.nix:11:1:
> ┃            10| in
> ┃            11| lib.nixvim.neovim-plugin.mkNeovimPlugin config {
> ┃              | ^
> ┃            12|   name = "bufferline";
> ┣━━━
> ┗━ ∑ ⚠ Exited with 1 errors reported by nix at 13:28:01 after 1s
> ```

Confusing to say the least... it built and deployed fine and I'm using the config myself..

ajaxbits:
Here's the code I'm working with: https://github.com/ajaxbits/nvim

Gonna keep debugging, but feel free to give it a look if you want.

MattSturgeon:
@ajaxbits you override the lib we use here: https://github.com/ajaxbits/nvim/blob/ef8eb6410366e7aed4189cb407585e75b7756434/flake.nix#L34

Which isn't supported because it means our "extended" lib isn't used.

Can you open an issue explaining _why_ you want to do that and I'll try and come up with something 🤔

ajaxbits:
Ah thank you for the explainer. I don't know why that was set. Weirdly, it was working for me all this time until recently. Who knows what that's all about. Removing that SpecialArg fixed my issue. Thank you for your time, and sorry for the false report!

MattSturgeon:
> I don't know why that was set. Weirdly, it was working for me all this time until recently. Who knows what that's all about.

Using an extended lib in our modules is a pretty new change: #1963

PR #2016, opened 2024-08-12
Fix for obsolete nerdfont icons

hansemschnokeloch:
These two icons are removed from nerdfont.
See https://www.nerdfonts.com/cheat-sheet and search for `clock_fast` and `message_reply_text`

### DIFF
--- a/plugins/utils/todo-comments.nix
+++ b/plugins/utils/todo-comments.nix
@@ -76,8 +76,8 @@ in
               TODO = { icon = " "; color = "info"; };
               HACK = { icon = " "; color = "warning"; };
               WARN = { icon = " "; color = "warning"; alt = [ "WARNING" "XXX" ]; };
-              PERF = { icon = " "; alt = [ "OPTIM" "PERFORMANCE" "OPTIMIZE" ]; };
-              NOTE = { icon = " "; color = "hint"; alt = [ "INFO" ]; };
+              PERF = { icon = " ", alt = { "OPTIM", "PERFORMANCE", "OPTIMIZE" } },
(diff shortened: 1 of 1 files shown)


plugins/utils/todo-comments.nix
```diff
@@ -76,8 +76,8 @@ in
               TODO = { icon = " "; color = "info"; };
               HACK = { icon = " "; color = "warning"; };
               WARN = { icon = " "; color = "warning"; alt = [ "WARNING" "XXX" ]; };
-              PERF = { icon = " "; alt = [ "OPTIM" "PERFORMANCE" "OPTIMIZE" ]; };
```
khaneliman, line 79: This looks like the defaults section for the plugin, the defaults I see in their config uses a different icon: 
`PERF = { icon = " ", alt = { "OPTIM", "PERFORMANCE", "OPTIMIZE" } },`

plugins/utils/todo-comments.nix
```diff
@@ -76,8 +76,8 @@ in
...
-              NOTE = { icon = " "; color = "hint"; alt = [ "INFO" ]; };
```
khaneliman, line 80: Same as other, looks like a different default icon:
`NOTE = { icon = " ", color = "hint", alt = { "INFO" } },`

khaneliman: They should still be nix attributes. I was just posting the lua from the plugin.

khaneliman:
Appreciate your PR, but will be superseded by https://github.com/nix-community/nixvim/pull/2030 migration PR.

Issue #1392, opened 2024-04-07
plugins/which-key: support registration with opts


justinrubek:
Currently the `which-key` module only supports calling register once and without opts: https://github.com/nix-community/nixvim/blob/0d85838d204d0f3d1cda22059d3fed0cd757d0ab/plugins/utils/which-key.nix#L211

I am still configuring this plugin manually because I have different keybinds for normal mode and for visual mode. It looks something like this:

```lua
local plugin = require("which-key")
local opts = {
    mode = "n",
    prefix = "<leader>",
    buffer = nil;
    silent = true,
    noremap = true,
    nowait = true,
}
local vopts = {
    mode = "v",
    prefix = "<leader>",
    buffer = nil;
    silent = true,
    noremap = true,
    nowait = true,
}
registrations = { ... }
vregistrations = { ... }
plugin.register(registrations, opts)
plugin.register(vregistrations, vopts)
```

It seems that I could handle normal mode commands having a different prefix by moving the prefix onto the keybinding, but after looking through the which-key docs I don't see any way to specify what mode a registration should belong to. I'd also prefer to keep these in a separate attrset anyway.

I'm not sure what this kind of change should look like. I only have two registrations but feasibly there could be an arbitrary number of them.

traxys:
Indeed that is something we would need to think about. In the mean time you can define the attrset of mappings in Nix, then add to your config: 
```nix
extraLuaConfig = ''
  require("which-key").register(${helpers.toLuaObject yourRegistration}, ${helpers.toLuaObject yourVopts})
'';
```

justinrubek:
With which-key v3 registrations (https://github.com/nix-community/nixvim/pull/1944) the nature of this has changed somewhat, but I don't consider it solved. The current way to specify registrations is a single list under [`plugins.which-key.settings.spec`](https://nix-community.github.io/nixvim/plugins/which-key/settings.html#pluginswhich-keysettingsspec). The only way to have bindings that are grouped (in my above example, I group normal mode vs visual mode) is to avoid using the nixvim module and to handle which-key directly.

I'd say that it has become even less ergonomic to use which-key than it was before. My main issue is that there's no way to specify multiple specs such that you can use inheritance on opts.

Here's what I've done to make my configuration work:

```nix
extraConfigLua = let
  names = builtins.attrNames bindings;
  generateRegistrations = name: let
    inherit (bindings.${name}) spec opts;
[...]

PR #2023, opened 2024-08-14
plugins/which-key: support mappings with corresponding opts

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1392

Follow up to a previous issue about supporting shared options with mappings. Which-key supports passing an options table alongside mappings to share options amongst a list of mappings.

### DIFF
--- a/plugins/utils/which-key.nix
+++ b/plugins/utils/which-key.nix
@@ -559,26 +559,82 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
       '';
       visible = false;
     };
+
+    mappings = mkOption {
+      type =
+        with types;
+        listOf (submodule {
+          options = {
+            spec = mkOption {
+              type = listOf (attrsOf anything);
+              description = ''
+                WhichKey mappings following the same format as `plugins.which-key.settings.spec`.
+
+                Mappings can also be defined in `plugins.which-key.settings.spec`.
+              '';
+            };
+            settings = mkOption {
+              type = nullOr (attrsOf anything);
+              default = null;
+              description = ''
+                WhichKey mapping options to apply to all the mapping in corresponding `plugins.which-key.mappings.spec`.
+              '';
+            };
+          };
+        });
+      default = [ ];
+      description = ''
+        Allows setting mappings with shared opts.
+      '';
+      example = [
+        {
+          spec = [
+            {
+              __unkeyed-1 = "<leader>f";
+              group = "Search";
+            }
+            {
+              __unkeyed-1 = "<leader>fb";
+              group = "Bindings";
+            }
+          ];
+          opts = {
+            mode = "v";
+          };
+        }
+      ];
+    };
   };
 
-  # TODO: introduced 2024-07-29: remove after 24.11
-  # NOTE: this may be upgraded to a mkRemoveOptionModule when which-key removes support
   extraConfig =
     cfg:
-    lib.mkIf opt.registrations.isDefined {
-      warnings = [
-        ''
-          nixvim (plugins.which-key):
-          The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.
-          You should use `plugins.which-key.settings.spec' instead.
+    mkMerge [
+      (lib.mkIf (cfg.mappings != [ ]) {
+        extraConfigLua = ''
+          do
+            local __whichKeyMappings = ${toLuaObject cfg.mappings}
+            for k,v in pairs(__whichKeyMappings) do
+              require('which-key').add(v.spec,v.settings)
+            end
+          end
+        '';
+      })
+      # TODO: introduced 2024-07-29: remove after 24.11
+      # NOTE: this may be upgraded to a mkRemoveOptionModule when which-key removes support
+      (lib.mkIf opt.registrations.isDefined {
+        warnings = [
+          ''
+            nixvim (plugins.which-key):
+            The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.
+            You should use `plugins.which-key.settings.spec' instead.
+            Note: the spec format has changed in which-key v3
+            See: https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings
+          ''
+        ];
 
-          Note: the spec format has changed in which-key v3
-          See: https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings
-        ''
-      ];
-
-      extraConfigLua = lib.optionalString opt.registrations.isDefined ''
-        require("which-key").register(${toLuaObject cfg.registrations})
-      '';
-    };
+        extraConfigLua = lib.optionalString opt.registrations.isDefined ''
+          require("which-key").register(${toLuaObject cfg.registrations})
+        '';
+      })
+    ];
 }
--- a/tests/test-sources/plugins/utils/which-key.nix
+++ b/tests/test-sources/plugins/utils/which-key.nix
@@ -197,7 +197,7 @@
   };
 
   # Testing for registrations
-  mappings = {
+  spec-mappings = {
     plugins.which-key = {
       enable = true;
       settings = {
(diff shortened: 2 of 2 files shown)


plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+
+    mappingsWithOpts = mkOption {
+      type =
+        with types;
+        listOf (submodule {
+          options = {
+            spec = mkOption {
+              type = listOf (attrsOf anything);
+              description = ''
+                WhichKey mappings following the same setup as `plugins.which-key.settings.spec`.
```
MattSturgeon, line 571:
```suggestion
                WhichKey mappings, following the same format as `plugins.which-key.settings.spec`.
```

Not sure on `format` `signature` `structure` `scheme` `shape` etc :thinking:

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+              '';
+            };
+            opts = mkOption {
+              type = attrsOf anything;
+              description = ''
+                WhichKey mapping options to apply to all the mapping in corresponding `plugins.which-key.mappingsWithOpts`.
```
MattSturgeon, line 577: I think the description should explain how this interacts with the `settings` we pass to `setup`; I assume it merges using vim's table extend function, so we could say can be used to **override** global options defined in `settings`?

e.g.
```suggestion
                Used to add or override WhichKey settings, only for these mappings.
                Follows the same format as `plugins.which-key.settings`.
```

I may have misinterpreted what the opts are for though?

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+              '';
+            };
+          };
+        });
```
MattSturgeon, line 581: I think we can set `default = [ ]` here instead of checking `isDefined` below in our usage.

Since it isn't a deprecated option we don't need to avoid having a definition, and having a default will make it easier for end-users to depend on the option value in any of their config.

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+      lib.optionalString opt.registrations.isDefined ''
+        require("which-key").register(${builtins.toJSON cfg.registrations})
+      ''
+      + lib.optionalString opt.mappingsWithOpts.isDefined (
+        lib.concatMapStringsSep "\n" (
+          regWithOpts:
+          "require('which-key').add(${toLuaObject regWithOpts.spec},${toLuaObject regWithOpts.opts})"
+        ) cfg.mappingsWithOpts
+      );
+  };
```
MattSturgeon, line 629: This is problematic because we have two optional strings, not gated behind a `mkIf`. Therefore if both conditions are false we will end up with an extra blank line in `extraConfigLua`.

Instead I think we should leave the original `extraConfig` attr untouched, but add another one using `mkMerge`.

e.g.
```nix
extraConfig = cfg: mkMerge [
  (lib.mkIf (cfg.mappingsWithOpts != [ ]) {
    extraConfigLua = lib.concatMapStringsSep "\n" (
      regWithOpts:
      "require('which-key').add(${toLuaObject regWithOpts.spec},${toLuaObject regWithOpts.opts})"
    ) cfg.mappingsWithOpts;
  })
  # TODO: introduced 2024-07-29: remove after 24.11
  # NOTE: this may be upgraded to a mkRemoveOptionModule when which-key removes support
  (lib.mkIf opt.registrations.isDefined {
    warnings = [
      ''
        nixvim (plugins.which-key):
        The option definition `plugins.which-key.registrations' in ${showFiles opt.registrations.files} has been deprecated in which-key v3; please remove it.
      ''
    ];

    extraConfigLua = lib.optionalString opt.registrations.isDefined ''
      require("which-key").register(${toLuaObject cfg.registrations})
    '';
  })
];
```

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
-    };
+    extraConfigLua =
+      lib.optionalString opt.registrations.isDefined ''
+        require("which-key").register(${builtins.toJSON cfg.registrations})
+      ''
+      + lib.optionalString opt.mappingsWithOpts.isDefined (
+        lib.concatMapStringsSep "\n" (
+          regWithOpts:
+          "require('which-key').add(${toLuaObject regWithOpts.spec},${toLuaObject regWithOpts.opts})"
+        ) cfg.mappingsWithOpts
```
MattSturgeon, line 627: This should work fine, but _usually_ we assign the whole option to a local variable, and then loop over it using `for in pairs`.

E.g. https://github.com/nix-community/nixvim/blob/a96aa9730af8c85dd7ed15e359ac23e9686f0a9a/modules/opts.nix#L74-L82

Or: https://github.com/nix-community/nixvim/blob/a96aa9730af8c85dd7ed15e359ac23e9686f0a9a/plugins/utils/wilder.nix#L257-L263

MattSturgeon (changes requested):

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+        with types;
+        listOf (submodule {
+          options = {
+            spec = mkOption {
+              type = listOf (attrsOf anything);
+              description = ''
+                WhichKey mappings following the same setup as `plugins.which-key.settings.spec`.
+              '';
+            };
+            opts = mkOption {
```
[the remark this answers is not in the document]
khaneliman: I'm keeping it in line with upstreams signature for the wk.add() method of registration. https://github.com/folke/which-key.nvim/blob/6c1584eb76b55629702716995cca4ae2798a9cca/lua/which-key/init.lua#L49

plugins/utils/which-key.nix
khaneliman: Upstream's .add method does a table.insert https://github.com/folke/which-key.nvim/blob/6c1584eb76b55629702716995cca4ae2798a9cca/lua/which-key/init.lua#L49

plugins/utils/which-key.nix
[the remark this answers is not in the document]
MattSturgeon: Sure, but by that logic we should also rename `settings` -> `opts`.

As I said, I'm conflicted on whether we should be consistent internally or match upstream's parameter names.

plugins/utils/which-key.nix
```diff
@@ -559,26 +559,72 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
       '';
       visible = false;
     };
+
+    mappingsWithOpts = mkOption {
```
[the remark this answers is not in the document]
MattSturgeon: Could also simply be named `mappings`, with a note that mappings can also be defined in `settings.spec`?

plugins/utils/which-key.nix
MattSturgeon: I assume the queue is just to allow for `setup` and `add` to be used in any order?

khaneliman: These aren't overriding a global option though, I thought it was basically just applying mapping options to a list instead of having to declare them in each mapping. (That was my understanding)

plugins/utils/which-key.nix
khaneliman: I can follow the pattern. I have an update using this.

plugins/utils/which-key.nix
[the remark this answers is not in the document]
khaneliman: Updated to use the mappings name, since upstream treats opts as a known optional parameter for mappings and recommends calling the `wk.add()` like this.

plugins/utils/which-key.nix
MattSturgeon: My current understanding is that isn't the case. Rather, the "nested" structure is still supported in the spec (`add`s first argument) and a second `opts` argument can also be supplied to set _other_ (non-spec) options at the same time. AFAICT this is undocumented.

Does this make the PR redundant?

khaneliman: > My current understanding is that isn't the case. Rather, the "nested" structure is still supported in the spec (`add`s first argument) and a second `opts` argument can also be supplied to set _other_ (non-spec) options at the same time. AFAICT this is undocumented.
> 
> Does this make the PR redundant?

If that is the case, I would agree it is redundant. I would basically just show different examples and update documentation in this instead...

khaneliman:
Holding off on implementing this, for now. Seems like the existing `settings` config supports the use case of the original issue and the undefined behavior of supplying `opts` to `wk.add()` doesn't provide any benefit in our generated config where we merge all `settings`.

PR #2024, opened 2024-08-15
flake-modules/devshell: add getopt runtimeInput

khaneliman:
Use a consistent behavior of getopt in launchtests.sh. Wasn't able to use tests command on macbook without including this because the macOS getopt didn't support our usage.


MattSturgeon (approved):
Awesome, thanks for figuring this out!

khaneliman:
@mergifyio queue

khaneliman:
@mergifyio dequeue

PR #2025, opened 2024-08-15
plugins/which-key: updated spec examples and tests

khaneliman:
Improve documentation for supported configurations

### DIFF
--- a/plugins/utils/which-key.nix
+++ b/plugins/utils/which-key.nix
@@ -11,6 +11,68 @@ let
   types = lib.nixvim.nixvimTypes;
 
   opt = options.plugins.which-key;
+
+  specExamples = [
+    # Basic group
+    {
+      __unkeyed-1 = "<leader>b";
+      group = "󰓩 Buffers";
+    }
+    # Non-default mode
+    {
+      __unkeyed = "<leader>c";
+      mode = "v";
+      group = "󰄄 Codesnap";
+    }
+    # Group within group
+    {
+      __unkeyed-1 = "<leader>bs";
+      group = "󰒺 Sort";
+    }
+    # Nested mappings for inheritance
+    {
+      mode = [
+        "n"
+        "v"
+      ];
+      __unkeyed-1 = [
+        {
+          __unkeyed-1 = "<leader>f";
+          group = "Normal Visual Group";
+        }
+        {
+          __unkeyed-1 = "<leader>f<tab>";
+          group = "Normal Visual Group in Group";
+        }
+      ];
+    }
+    # Proxy mapping
+    {
+      __unkeyed-1 = "<leader>w";
+      proxy = "<C-w>";
+      group = "windows";
+    }
+    # Create mapping
+    {
+      __unkeyed-1 = "<leader>cS";
+      __unkeyed-2 = "<cmd>CodeSnapSave<CR>";
+      mode = "v";
+      desc = "Save";
+    }
+    # Function mapping
+    {
+      __unkeyed-1 = "<leader>db";
+      __unkeyed-2.__raw = ''
+        function()
+          require("dap").toggle_breakpoint()
+        end
+      '';
+      mode = "n";
+      desc = "Breakpoint toggle";
+      silent = true;
+    }
+  ];
+
 in
 lib.nixvim.neovim-plugin.mkNeovimPlugin config {
   name = "which-key";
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/utils/which-key.nix
```diff
@@ -251,6 +251,31 @@
...
+                {
+                  __unkeyed-1 = "<leader>f<tab>";
+                  group = "Group in Group Test";
+                }
+              ];
+            }
+            # Create mapping
+            {
+              __unkeyed-1 = "<leader>cS";
+              __unkeyed-2 = ":CodeSnapSave<CR>";
```
MattSturgeon, line 274: This _really_ isn't important here, as it's just a test example, but any `:foo<cr>` mapping can be replaced with `<cmd>foo<cr>` with a few caveats:
- it must end with `<cr>`
- it is automatically silent

Using `<cmd>` will work better than `:` when the mapping is used in some modes

tests/test-sources/plugins/utils/which-key.nix
```diff
@@ -251,6 +251,31 @@
               desc = "Vim cmd KeyMapping Test";
               inherit mode;
             }
+            # Nested mapping
```
MattSturgeon, line 254: Is it worth porting over the [test we used to have](https://github.com/nix-community/nixvim/blob/62f32bfc711f0a3c8a52af4c0790345b4b3d2346/tests/test-sources/plugins/utils/which-key.nix)? Or do you think this is enough of a test?

plugins/utils/which-key.nix
```diff
@@ -247,19 +247,25 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
         Please refer to the plugin's [documentation](https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-mappings).
       '';
       example = [
-        {
-          __unkeyed-1 = "<leader>w";
-          proxy = "<C-w>";
-          group = "windows";
-        }
```
MattSturgeon, line 254: Is there an issue with having a `proxy` mapping example?

plugins/utils/which-key.nix
```diff
@@ -484,33 +490,46 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
     delay = 200;
 
     spec = [
```
MattSturgeon, line 492: We could do `spec = options.plugins.which-key.settings.spec.example` so we don't need to come up with two examples?

MattSturgeon: Actually, I think we'd have to use `(options.plugins.which-key.settings.type.getSubOptions [ ]).spec.example`, so maybe it's more readable to define the example in a let block and use it in both instances, or use a `rec` attrset and reference the local definition recursively, e.g. use `rec` to assign `example = settingsExample.spec` in the spec option declaration.

Or just do it how you're doing it already; i.e. have two examples.

MattSturgeon (commented):
No issues. Just some discussion points you're welcome to dismiss :slightly_smiling_face:

plugins/utils/which-key.nix
khaneliman: Sorry, just deleted a group of them and didn't notice the proxy example.

tests/test-sources/plugins/utils/which-key.nix
khaneliman: Good to know, thank!

plugins/utils/which-key.nix
khaneliman: Moved them to a shared let block so we can manage them in one place.

plugins/utils/which-key.nix
```diff
@@ -11,6 +11,68 @@ let
   types = lib.nixvim.nixvimTypes;
 
   opt = options.plugins.which-key;
+
+  specExamples = [
+    # Basic group
```
MattSturgeon, line 16: Is it worth using a `literalExpression` or `literalMD` example so that these comments are shown in the docs? I guess that wouldn't work for `settingsExample` since we can't _nest_ a literal expression _within_ a larger example :thinking:

MattSturgeon (approved):
LGTM, thanks for improving these examples!

plugins/utils/which-key.nix
khaneliman: Yeah, it's a bit confusing in the `settingsExample` with a literalExpression. Can we parse the text attribute from it to get the actual nix ?

MattSturgeon: If `settingsExample` was a literal expression we could include the string part of the "nested" literal, however TBH I don't think having a few comments is worth the effort?

We may also be able to do something with a ["pretty printer"](https://github.com/NixOS/nixpkgs/blob/47f999b551b9e7f515a678e326bd945ac7b7fda8/lib/generators.nix#L471-L474), assuming the docs enable `allowPretty`.

MattSturgeon: But frankly I don't think it's worth the effort here

khaneliman: > If `settingsExample` was a literal expression we could include the string part of the "nested" literal, however TBH I don't think having a few comments is worth the effort?

Yeah, I just saw the generated attribute set from passing a `literalExpression` to the `settingsExample` and got curious if we could grab the `text` attribute and convert it into a nix object easily. 

> But frankly I don't think it's worth the effort here

I agree, just got curious about how it could be done. :P The prettyprint example is interesting.

khaneliman:
@mergifyio queue

plugins/utils/which-key.nix
MattSturgeon: We could have a helper similar to `literalExpression` that returns a pretty-printer that just prints the text? Only issue I see is the pretty-printer doesn't receive the current indentation string...

PR #2026, opened 2024-08-15
Changed default telescope-borders to true instead of false

jaschutte:
I have been struggling to figure out why my telescope didn't have borders, as it usually does when normally enabled. Then I came across this option and saw it was set to false by default. This changes it back to true to be consistent with every other default telescope configuration.

### DIFF
--- a/plugins/colorschemes/base16/default.nix
+++ b/plugins/colorschemes/base16/default.nix
@@ -41,7 +41,7 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     ];
 
   settingsExample = {
-    telescope_borders = true;
+    telescope_borders = false;
     indentblankline = false;
     dapui = false;
   };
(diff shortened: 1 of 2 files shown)


MattSturgeon:
This doesn't do what you think it does. The "default" you've changed is not a real default used in the nix options. It is actually a "plugin default" that we _only_ use in the docs.

`settings` options are not _our_ options per-se, but just a representation of upstream's options. When unset (null) we simply don't include the option in the lua passed to the upstream plugin's `setup` function.

If you think the default should be different, you should probably raise that upstream with the plugin.

You can verify all this for yourself by running `nixvim-print-init` and inspecting the lua that nixvim generated.

jaschutte:
> This doesn't do what you think it does. The "default" you've changed is not a real default used in the nix options. It is actually a "plugin default" that we _only_ use in the docs.
> 
> `settings` options are not _our_ options per-se, but just a representation of upstream's options. When unset (null) we simply don't include the option in the lua passed to the upstream plugin's `setup` function.
> 
> If you think the default should be different, you should probably raise that upstream with the base16 plugin.
> 
> You can verify all this for yourself by running `nixvim-print-init` and inspecting the lua that nixvim generated.

Thank you for the useful information, I'll try contacting upstream.

PR #2028, opened 2024-08-16
tests/lua-loader: builtins.match -> lib.hasInfix 

khaneliman:
Getting invalid regex expression on `aarch64-darwin`, this passes on both `aarch64-darwin` and `x86_64-linux`, for me. I saw that the builtins.match has different behaviors between each platform in https://github.com/NixOS/nix/issues/1537 and https://github.com/NixOS/nix/issues/4758.

### DIFF
--- a/tests/test-sources/modules/lua-loader.nix
+++ b/tests/test-sources/modules/lua-loader.nix
@@ -1,25 +1,25 @@
 {
   # Test that nothing is configured by default
   default.module =
-    { config, ... }:
+    { config, lib, ... }:
     {
       files."files_test.lua" = { };
 
       assertions = [
         {
-          assertion = builtins.match ".*vim\.loader.*" config.content == null;
+          assertion = !lib.hasInfix "vim.loader" config.content;
           message = "No luaLoader configuration is expected in init.lua by default.";
         }
         {
-          assertion = builtins.match ".*vim\.loader.*" config.files."files_test.lua".content == null;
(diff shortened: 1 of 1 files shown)


tests/test-sources/modules/lua-loader.nix
```diff
@@ -7,11 +7,11 @@
 
       assertions = [
         {
-          assertion = builtins.match ".*vim\.loader.*" config.content == null;
+          assertion = builtins.match ".*vim\\.loader.*" config.content == null;
```
MattSturgeon, line 10: We could also do `[.]`, but if this works on all platforms then that's fine.

I'm curious, is `"\\."` the same as `''\.''`? If so maybe using raw strings for regexes would be better?

MattSturgeon: Actually, we can replace all these with `lib.hasInfix`, which just checks if a string contains a given substring.

It's implemented using `match` and a regex escape, but is better than rolling our own regexes.

khaneliman: > Actually, we can replace all these with `lib.hasInfix`, which just checks if a string contains a given substring.
> 
> It's implemented using `match` and a regex escape, but is better than rolling our own regexes.

100%, I’d prefer that to dealing with a regex ourselves. I’ll update in a bit

khaneliman:
@mergifyio queue

Issue #2006, opened 2024-08-10
[BUG] The update CI schedules on forks


MattSturgeon:
If GitHub Actions is enabled on a fork, the update CI is scheduled weekly.

This is a waste of CI minutes, needlessly opens PRs targeting the fork and also sends failure notifications if the 24.05 branch doesn't exist on the fork.

I believe it is possible to have the schedule only run on this repo.

PR #2029, opened 2024-08-16
workflows/update: only run upstream

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2006

Tested running workflow on my repo: 

- Using upstream as repo check (skips running) https://github.com/khaneliman/nixvim/actions/runs/10426474651
- Using my own repo check (runs update) https://github.com/khaneliman/nixvim/actions/runs/10426526382/job/28879614530

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -34,6 +34,7 @@ jobs:
     name: Update the flake inputs and generate options
     runs-on: ubuntu-latest
     timeout-minutes: 40
+    if: github.event_name != 'schedule' || github.repository == 'nix-community/nixvim'
 
     steps:
       - name: Checkout repository


.github/workflows/update.yml
```diff
@@ -34,6 +34,7 @@ jobs:
     name: Update the flake inputs and generate options
     runs-on: ubuntu-latest
     timeout-minutes: 40
+    if: github.repository == 'nix-community/nixvim'
```
MattSturgeon, line 37: This would still allow users to run manually, using workflow despatch. IDK _why_ they'd want to do that, maybe just to test things out?

```suggestion
    if: github.event_name != 'schedule' || github.repository == 'nix-community/nixvim'
```

i.e. `event_name == 'schedule' -> repo == 'nix-community/nixvim'`

MattSturgeon (changes requested):
Thanks for looking into this! One minor suggestion then I think this is good to go :rocket: 

One other issue; since the workflow still runs but the job is _not_ run, would you still get the "Run cancelled" email notifications if you have action failure notifications enabled?

If so, I don't think there's a way to avoid that though, since you can't have conditions on the workflow itself, or on the `schedule` trigger.

I've actually started filtering out these emails anyway, because I get a ton of them from some workflows that have concurrency rules...

.github/workflows/update.yml
khaneliman: It shouldn't allow a non maintainer to workflow dispatch, right? If that was the concern. Otherwise, it'd be nice to have flexibility to run ad-hoc when there is a fix from upstream for stuff broken preventing us from updating on the current nixpkgs revision.

MattSturgeon: Those kinda security concerns are handled by GitHub's permission system. We don't need to deal with them in the workflow; in fact we can't, since anyone is free to fork and change the workflow content.

Rather, we just want to prevent _automatically_ running the workflow on forks.

Any workflow runs on a fork will only affect the fork repo, but it is annoying to have it run automatically.

khaneliman: Yeah, my question was that I wasn't sure of the motivation for the requested change. This prevents us from running it on demand in this repo.

khaneliman: > This would still allow users to run manually, using workflow despatch. IDK _why_ they'd want to do that, maybe just to test things out?
> 
> i.e. `event_name == 'schedule' -> repo == 'nix-community/nixvim'`

Sorry, I think I totally misread this... I see you were saying you wanted them to be able to run on demand on their own repo with the change and it's an OR condition so it will still work for us.

MattSturgeon (approved):
Awesome, thanks again!

khaneliman:
@mergifyio queue

PR #2030, opened 2024-08-17
plugins/todo-comments: migrate to mkNeovimPlugin

khaneliman:


plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,400 @@ let
...
+      after = defaultNullOpts.mkStr "fg" ''
+        "fg" or "bg" or empty.
+      '';
+
+      pattern = defaultNullOpts.mkNullable' {
+        # TODO: Added 2024-08-16, remove after 24.11
+        # Before we modified value automatically turning it into a raw regex.
+        type = transitionType (with types; either str (listOf str)) (
+          text: if isList text then map (elem: mkRaw "[[${elem}]]") text else mkRaw "[[${text}]]"
+        ) (with types; either rawLua (listOf rawLua));
```
MattSturgeon, line 268: I think transition type would be confused by the composite type (listOf). Especially for empty list definitions.

Perhaps we could do `either (transitionType str) (listOf (transitionType str))`?

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,400 @@ let
...
+        pluginDefault.__raw = "[[.*<(KEYWORDS)\\s*:]]";
+        description = ''
```
MattSturgeon, line 270: Just to clarify, `[[ ]]` is just a string syntax in lua (raw strings to be precise), anything that is written using `[[ ]]` can also be written using `" "` or `' '`, so I think this option can actually just be `either str (listOf str)` and the old option can be removed if it's incompatible?

Or if we want to be graceful the old option could be kept and we could copy it's definition to here when it's defined.

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,400 @@ let
...
+          ]
+          ''
+            Arguments to use for the search command in list form.
+          '';
 
-        pattern = helpers.defaultNullOpts.mkStr "\\b(KEYWORDS):" ''
+      pattern = mkNullOrOption' {
+        # TODO: Added 2024-08-16, remove after 24.11
+        # Before we modified value automatically turning it into a raw regex.
+        type = transitionType types.str (text: mkRaw "[[${text}]]") types.rawLua;
```
MattSturgeon, line 342: Same. This doesn't need to be a raw strings.

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,400 @@ let
...
-          max_line_len = cfg.highlight.maxLineLen;
-        };
-        inherit (cfg) colors;
-        search = {
-          inherit (cfg.search) command args;
-          pattern = helpers.ifNonNull' cfg.search.pattern (
-            if isList cfg.search.pattern then
-              (map helpers.mkRaw cfg.search.pattern)
-            else
-              helpers.mkRaw "[[${cfg.search.pattern}]]"
```
MattSturgeon, line 255: This should be effectively equivalent to `toLuaObject cfg.search.pattern`

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,400 @@ let
...
-          }
-        ) commands
-      );
-
-      # Automatically enable plugins if keymaps have been set
-      plugins = mkMerge [
-        (mkIf (cfg.keymaps.todoTrouble != null) { trouble.enable = true; })
-        (mkIf (cfg.keymaps.todoTelescope != null) { telescope.enable = true; })
-      ];
+          mkCompositeOption "Keymap settings for the `:${funcName}` function." {
```
MattSturgeon, line 362: Is it possible to do this is a way that complies with https://github.com/nix-community/nixvim/issues/603?

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,394 @@ let
...
+          cwd = optionalString (keymap.cwd != null) " cwd=${keymap.cwd}";
+          keywords = optionalString (keymap.keywords != null) " keywords=${keymap.keywords}";
+        in
+        optional (keymap != null) {
+          mode = "n";
+          inherit (keymap) key options;
+          action = ":${funcName}${cwd}${keywords}<CR>";
+        }
+      ) commands
+    );
```
MattSturgeon, line 408: I see this is unchanged from the previous implementation, other than being indented further.

However, I think this would be simpler mapping over `cfg.keymaps` instead of mapping over `commands`?

Ideally, I'd like to avoid using `flatten`, perhaps we could do something like:

```nix
keymaps = lib.pipe cfg.keymaps [
  (filterAttrs (n: v: v != null))
  (mapAttrsToList (name: keymap: {
    inherit (keymap) key option;
    mode = "n";
    action =
      let
        cmd = commands.${name};
        cwd = optionalString (keymap.cwd != null) " cwd=${keymap.cwd}";
        keywords = optionalString (keymap.keywords != null) " keywords=${keymap.keywords}";
      in
      "<cmd>${cmd}${cwd}${keywords}<cr>";
  }))
];
```

Looking at the `commands` attrset, we could probably replace it with a `upperFirstChar` style helper.

However, many of the details would be slightly different if we can go down the #603 route.

plugins/utils/todo-comments.nix
```diff
@@ -16,275 +24,394 @@ let
...
+
+    # Automatically enable plugins if keymaps have been set
+    plugins = mkMerge [
+      (mkIf (cfg.keymaps.todoTrouble != null) { trouble.enable = true; })
+      (mkIf (cfg.keymaps.todoTelescope != null) { telescope.enable = true; })
+    ];
```
MattSturgeon, line 414: Is this something we still want to do, or would we rather warn/assert asking the user to enable the plugins themselves?

plugins/utils/todo-comments.nix
MattSturgeon: If you're unfamiliar, the idea is that plugin keymaps should make use of `mkMapOptionSubmodule` instead of implementing their own keymap submodule types.

MattSturgeon: If `mkMapOptionSubmodule` is too limiting, because it doesn't support additional options (`cwd`, `keywords`) I'm open to making it more powerful by adding an `extraOptions` argument or similar.

plugins/utils/todo-comments.nix
khaneliman: This is something I saw and mentioned in https://github.com/nix-community/nixvim/pull/1559. I wasn't sure what everyone's opinions were so I left it untouched. But, based on that PR it would seem we'd want to change it to assertions.

plugins/utils/todo-comments.nix
khaneliman: > I see this is unchanged from the previous implementation, other than being indented further.
> 
Yeah, sorry I didn't touch this bit. I wasn't sure what to do with it, yet.

I think the suggested solution is a lot cleaner.  But, I need to get more familiar with that keymap issue that was mentioned here and above.

MattSturgeon: We can always tackle it in another PR, but the other suggestions above could be better.

khaneliman: If I can tackle that separately it would give me a bit to figure it out and maybe get the extraOptions support for mkMapOptionSubmodule

plugins/utils/todo-comments.nix
MattSturgeon: As discussed in https://github.com/nix-community/nixvim/pull/2030#discussion_r1720865792 this can done in be another PR

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+    mkCompositeOption
+    ;
+  types = lib.nixvim.nixvimTypes;
+
+  keymapsActions = {
     todoQuickFix = "TodoQuickFix";
     todoLocList = "TodoLocList";
     todoTrouble = "TodoTrouble";
     todoTelescope = "TodoTelescope";
   };
```
MattSturgeon, line 22: Up to you, we could drop this, since we have `lib.nixvim.upperFirstChar`: https://github.com/nix-community/nixvim/blob/78fc4be6a830e8dc01f3e66ddbe3243b4bfe8560/lib/utils.nix#L81-L104

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+
+    keymaps = lib.pipe cfg.keymaps [
+      (filterAttrs (n: v: v != null))
+      (mapAttrsToList (
+        name: keymap: {
+          inherit (keymap) key options;
+          mode = "n";
+          action =
+            let
+              cmd = keymapsActions.${name};
```
MattSturgeon, line 420: _If_ you decide to drop the `keymapsActions` attrs:
```suggestion
              cmd = lib.nixvim.upperFirstChar name;
```

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+      name = "ripgrep";
+      default = pkgs.ripgrep;
     };
+  };
+
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion =
+          cfg.keymaps.todoTelescope != null && cfg.keymaps.todoTelescope.key != null
```
MattSturgeon, line 392: Instead of null-checking each step, we can use the fallback operator:

```suggestion
          cfg.keymaps.todoTelescope.key or null != null
```

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+          -> config.plugins.telescope.enable;
+        message = ''
+          Nixvim: You have enabled the `telescope` integration with git-worktree.
+          However, you have not enabled the `telescope` plugin itself (`plugins.telescope.enable = true`).
+        '';
+      }
+      {
+        assertion =
+          cfg.keymaps.todoTrouble != null && cfg.keymaps.todoTrouble.key != null
```
MattSturgeon, line 401: Same:
```suggestion
          cfg.keymaps.todoTrouble.key or null != null
```

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+  };
+
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion =
+          cfg.keymaps.todoTelescope != null && cfg.keymaps.todoTelescope.key != null
+          -> config.plugins.telescope.enable;
+        message = ''
+          Nixvim: You have enabled the `telescope` integration with git-worktree.
```
MattSturgeon, line 395:
```suggestion
          Nixvim(plugins.todo-comments): You have enabled todo-comment's `telescope` integration.
```

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+          However, you have not enabled the `telescope` plugin itself (`plugins.telescope.enable = true`).
+        '';
+      }
+      {
+        assertion =
+          cfg.keymaps.todoTrouble != null && cfg.keymaps.todoTrouble.key != null
+          -> config.plugins.trouble.enable;
+        message = ''
+          Nixvim: You have enabled the `trouble` integration with todo-comments.
```
MattSturgeon, line 404:
```suggestion
          Nixvim(plugins.todo-comments): You have enabled todo-comment's `trouble` integration.
```

MattSturgeon: IDK if the `(plugins.todo-comments)` is really needed since it's obvious from the assertion text, be we _usually_ include something like it in our warnings/assertions?

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+
+            keywords = mkOption {
+              type = types.nullOr types.str;
+              description = ''
+                Comma separated list of keywords to filter results by.
+                Keywords are case-sensitive.
+              '';
+              default = null;
+              example = "TODO,FIX";
+            };
```
MattSturgeon, line 375: This seems kinda backwards, since this is a manually crafted option why don't we just use a `listOf str` type and then `concatStringsSep ","` ourselves later? It's not like this is a representation of a freeform settings option.

We could use `transitionType` here, however maybe it's best to wait until we migrate to `mkMapOptionSubmodule`, in case we need other deprecations.

plugins/utils/todo-comments.nix
MattSturgeon: Well, I guess it's used in the option declaration as well as the keymap transformation. So maybe best to keep it?

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+      basePluginPath = [
+        "plugins"
+        "todo-comments"
+      ];
+    in
+    [
+      (mkRemovedOptionModule (
+        basePluginPath ++ [ "keymapsSilent" ]
+      ) "Use `plugins.todo-comments.keymaps.<COMMAND>.options.silent`.")
+      (mkRenamedOptionModule (basePluginPath ++ [ "extraOptions" ]) (basePluginPath ++ [ "settings" ]))
```
MattSturgeon, line 43: This is the same as `mkNeovimPlugin`'s `deprecateExtraOptions` option:
https://github.com/nix-community/nixvim/blob/78fc4be6a830e8dc01f3e66ddbe3243b4bfe8560/lib/neovim-plugin.nix#L25

I don't mind if you'd rather do it explicitly, but it might be easier to keep the dated `TODO` comments separate if we move this out of imports.

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+    let
+      basePluginPath = [
+        "plugins"
+        "todo-comments"
+      ];
+    in
+    [
+      (mkRemovedOptionModule (
+        basePluginPath ++ [ "keymapsSilent" ]
+      ) "Use `plugins.todo-comments.keymaps.<COMMAND>.options.silent`.")
```
MattSturgeon, line 42: This was added 2023-11-06 in e45be89f1267faaae0e67fd6d3f3a2457314a94c

plugins/utils/todo-comments.nix
MattSturgeon: The imports can be dated 2023-11-06 while `deprecateExtraOptions` and `optionsRenamedToSettings` can be grouped together and dated `2024-08-18`.

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
+    sign_priority = defaultNullOpts.mkInt 8 "Sign priority.";
 
-        bg = helpers.defaultNullOpts.mkStr "BOLD" ''
-          The gui style to use for the bg highlight group.
-        '';
-      };
+    keywords =
+      defaultNullOpts.mkAttrsOf
+        (types.submodule {
+          options = {
```
MattSturgeon, line 146: Note: this will be affected by #1618, however that doesn't affect how you should implement it here. It's just another reason for us to investigate the issue.

MattSturgeon, line 146: Should this have a freeform type? Or should we restrict users to only these sub-options?

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
-              Pattern or list of patterns, used for highlighting (vim regex)
+      multiline_context = defaultNullOpts.mkInt 10 ''
+        Extra lines that will be re-evaluated when changing a line.
+      '';
 
-              Note: the provided pattern will be embedded as such: `[[PATTERN]]`.
-            '';
+      before = defaultNullOpts.mkStr "" ''
+        "fg" or "bg" or empty.
+      '';
```
MattSturgeon, line 248: Looks like maybe an enum?

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
 
-        commentsOnly = helpers.defaultNullOpts.mkBool true ''
-          Uses treesitter to match keywords in comments only.
-        '';
+      keyword = defaultNullOpts.mkStr "wide" ''
+        "fg", "bg", "wide", "wide_bg", "wide_fg" or empty.
+        (wide and wide_bg is the same as bg, but will also highlight
+        surrounding characters, wide_fg acts accordingly but with fg).
+      '';
```
MattSturgeon, line 254: Same (enum?)

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,429 @@
...
 
-        maxLineLen = helpers.defaultNullOpts.mkInt 400 ''
-          Ignore lines longer than this.
-        '';
+      after = defaultNullOpts.mkStr "fg" ''
+        "fg" or "bg" or empty.
+      '';
```
MattSturgeon, line 258: Same (enum?)

MattSturgeon (changes requested):
Think these are the final comments :rocket:

plugins/utils/todo-comments.nix
khaneliman: Thanks, I was hoping this could be simplified with some kind of null conditional operator.

plugins/utils/todo-comments.nix
khaneliman: lol fun copy pasta

plugins/utils/todo-comments.nix
khaneliman: I'm torn on it, these are the kinds of things that are limited to options and it can be nice to have available in documentation. I think it will be easier to remove the maintenance when we figure out the documentation only options.

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,452 @@
...
+              type = types.nullOr types.str;
+              description = "Specify the directory to search for comments";
+              default = null;
+              example = "~/projects/foobar";
+            };
+
+            keywords = mkOption {
+              type =
+                with types;
+                transitionType str (string: if string != null then splitString "," string else null) (
```
MattSturgeon, line 391: I don't _think_ we need the null check in the coercion function?

If the value is null then it'll fail `types.str.check` and will end up being merged using the `nullOr` type anyway.

```suggestion
                transitionType str (splitString ",") (
```

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,452 @@
...
+      (mapAttrsToList (
+        name: keymap: {
+          inherit (keymap) key options;
+          mode = "n";
+          action =
+            let
+              cmd = keymapsActions.${name};
+              cwd = optionalString (keymap.cwd != null) " cwd=${keymap.cwd}";
+              keywords = optionalString (
+                keymap.keywords != null
```
MattSturgeon, line 444:
```suggestion
                keymap.keywords != null && keymap.keywords != [ ]
```

Or is it intentional to be able to pass an empty list?

If not, maybe we can get rid of the `nullOr` (moving it to the left of the `transitionType`)?

plugins/utils/todo-comments.nix
MattSturgeon: I meant having `freeformType = attrsOf anything` in addition to `options = {}`.

We don't need to do this, if you want to restrict the user to only the sub-options declared, but if you don't want that restriction, we need to define a freeformType.

plugins/utils/todo-comments.nix
```diff
@@ -1,290 +1,452 @@
...
+    assertions = [
+      {
+        assertion = cfg.keymaps.todoTelescope.key or null != null -> config.plugins.telescope.enable;
+        message = ''
+          Nixvim(plugins.todo-comments): You have enabled todo-comment's `telescope` integration.
+          However, you have not enabled the `telescope` plugin itself (`plugins.telescope.enable = true`).
+        '';
+      }
+      {
+        assertion = cfg.keymaps.todoTrouble or null != null -> config.plugins.trouble.enable;
```
MattSturgeon, line 423:
```suggestion
        assertion = cfg.keymaps.todoTrouble.key or null != null -> config.plugins.trouble.enable;
```

Typo

plugins/utils/todo-comments.nix
khaneliman: It was required when i was doing transition with null old type, then I realized setting null as an old type caused it to trigger even when it wasn't needed. Now it's only attempting to coerce properly and it can be removed.

plugins/utils/todo-comments.nix
khaneliman: Not sure why someone would choose to pass an empty list, it should be the same effect as null to the plugin so it makes sense to add it to the condition to avoid trying to concat an empty list.

plugins/utils/todo-comments.nix
khaneliman: That makes sense, I was thinking it replaced the existing options for some reason.

MattSturgeon (approved):
Can't see any remaining issues. Thanks for another quality PR! :tada: 

Using `mkMapOptionSubmodule` as per #603 can be postponed to another PR, and may be best done after #2035 is merged anyway.

Up to you if you're happy merging now if if you'd like to ping other maintainers for a second opinion.

No need to wait for the tests to pass though, since mergify will do that for us.

khaneliman:
@mergifyio queue

MattSturgeon:
@mergifyio refresh

MattSturgeon:
@mergifyio queue

PR #2033, opened 2024-08-17
colorschemes/catppuccin: add byteCompile option

stasjok:
This commit adds `colorschemes.catppuccin.byteCompile` options. If enabled it byte compiles the colorscheme. Resulting package has only colorschemes itself in `colors` directory, no any other files. Colorschemes are byte compiled with all provided settings. This is achieved by running neovim instance with configured catppuccin colorscheme and copying byte compiled colorschemes from the cache directory.

Upstream plugin has support for `flavour = "auto"`. In that case setup function loads a flavour based on current background. Currently byteCompile option doesn't support this behavior. If the user didn't set flavour in settings, mocha flavour is used (upstream default for dark background). In theory additional logic can be added for checking current `globalOpts.background` (and its aliases) to emulate upstream 'auto' behavior, but it would compilcate things, so it's not handled by this commit.

Setting `opts.termguicolors` was dropped here, because it's already explicitly set by catppuccin colorscheme.

### DIFF
--- a/plugins/colorschemes/catppuccin.nix
+++ b/plugins/colorschemes/catppuccin.nix
@@ -6,10 +6,20 @@
   ...
 }:
 with lib;
+let
+  flavours = [
+    "latte"
+    "mocha"
+    "frappe"
+    "macchiato"
+  ];
+in
 helpers.neovim-plugin.mkNeovimPlugin config {
   name = "catppuccin";
   isColorscheme = true;
   defaultPackage = pkgs.vimPlugins.catppuccin-nvim;
+  installPackage = false;
+  callSetup = false;
 
   maintainers = [ maintainers.GaetanLepage ];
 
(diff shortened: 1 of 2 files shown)


traxys:
I think this proposition starts to go outside of the scope of nixvim. We try to limit the amount of per-plugin processing we do, as each bespoke implementation adds a maintenance burden, and we have a large ratio of plugins to maintainers.
This seems more like the role of a neovim distribution, which we kind of try to avoid.

There are a few places where it makes sense to go beyond simply allowing to configure plugins, like the plugin management, extra programs, or the performance options that you added, as they have a general effect, and as such are much easier for us to maintain.

If you _need_ to go into the weeds of nixvim for this optimization it may mean that we need a way to allow users to write this themselves in their configuration, or in another project that would build upon nixvim

Issue #1935, opened 2024-07-27
Deprecate `helpers.mkRaw` in favour of `lib.generators.mkLuaInline`


MattSturgeon:
- [ ] `helpers.lua.mkLua` should support nixpkg's "lua-inline" type.
- [ ] Documentation should recommend `mkLuaInline`
- [ ] Transition internal `__raw` to use `mkLuaInline`
- [ ] Add a deprecation warning to `helpers.mkRaw`

Note: there's now also a `lib.generators.toLua` in nixpkgs, however we currently can't use it, because it has no support for mixed list+attrs tables.

khaneliman:
I'm a fan of transitioning away from a custom function to one maintained upstream. That would increase consistency with other repos and lower the tech debt with managing our own library functions.

MattSturgeon:
I agree. I feel a little hesitant because we'll still be maintaining our own "unkeyed attrs" system, however I don't think this needs to be all or nothing straight away.

We can come up with an upstream solution to mixed lua tables separately. Thankfully this is less significant than raw/inline lua, because it is used far less often.

MattSturgeon:
Something else to consider is how we show raw/inline values in the docs.

`lib.generators.toPretty` has support for a custom "pretty print" type (`allowPretty` iirc).

If that's enabled where the docs use `toPretty`, we could replace usage of `mkLuaInline` in defaultText&examples with a "forDocs" function that returns a pretty printer.

I think ideally, our examples should show the function _application_ not the resulting attrs.

Otherwise we'd just have to use `literalExpression`, e.g. `literalExpression "lib.generators.mkLuaInline { foo = false; }"`, but this is only used for _whole_ examples, not small parts of them.

MattSturgeon:
We could have a `mkLuaInlineLiteral`, but since pretty-printers aren't supplied indentation information, we'd have to either work it out in advance or have no indentation.

```nix
mkLuaInlineLiteralIndented = indent: text: {
  __pretty =
    v:
    indent
    + "mkLuaInline "
    + generators.toPretty { inherit indent; } v;
  val = text;
};
```

It'd be nice if toPretty provided a way to call a pretty-printing function with additional information, such as the current indent and toPretty's args. Maybe if the pretty printer has `__prettyWith` instead of `__pretty`?

GaetanLepage:
Thanks @MattSturgeon.
I agree with your migration plan. Doing it step by step looks great.

MattSturgeon:
[...]

PR #2035, opened 2024-08-17
modules/keymap: improve `lua` deprecation

MattSturgeon:
- Replace nullable lua option with a no-default option.
- Made it so the deprecated option is only declared when `lua = true` is passed.
- Replace `normalizeMappings` with a `removeDeprecatedMapAttrs` helper.
- Added warnings for **all options** that historically had `lua` support.

This supersedes and closes #1795

## Removing the attr from the final value

I'd like to do something like this directly in `mkMapOptionSubmodule`, however it doesn't seem to do what you'd expect...

```nix
type // {
  merge = loc: defs: builtins.removeAttrs (type.merge loc defs) [ "lua" ];
};
```

Therefore I've instead added `apply` functions to each keymap option that removes the `lua` attr.

See this [message on matrix](https://matrix.to/#/!wfudwzqQUiJYJnqfSY:nixos.org/$UjMlZGRb5mPCxWdvv6DDGOnuk95ahtuX3TzSHr5dO-w?via=nixos.org&via=matrix.org&via=nixos.dev).

## Showing the warning for all uses

Previously, we only checked the `keymaps` option in our deprecation warning.

We could instead have a `lib.warn` in the `action`'s `apply` function when we call `mkRaw`, however this wouldn't have access to `loc` or `defs` to print a proper warning.

If overriding the `merge` function worked correctly, we could do it there, however I think this is fine for this PR, we can always improve things further in future work.

For now, `modules/keymaps.nix` is responsible for showing warnings for all options that previously had a `lua` sub-option.

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -58,6 +58,12 @@ rec {
 
   mapOptionSubmodule = mkMapOptionSubmodule { };
 
+  # NOTE: options that have the deprecated `lua` sub-option must use `removeDeprecatedMapAttrs`
+  # to ensure `lua` isn't evaluated when (e.g.) generating lua code.
+  # Failure to do so will result in "option used but not defined" errors!
+  deprecatedMapOptionSubmodule = mkMapOptionSubmodule { lua = true; };
+  removeDeprecatedMapAttrs = v: builtins.removeAttrs v [ "lua" ];
+
   mkModeOption =
     default:
     mkOption {
@@ -81,13 +87,12 @@ rec {
       # or an attrset to enable the option and add/override mkOption args.
       key ? true,
       action ? true,
+      lua ? false, # WARNING: for historic use only - do not use in new options!
     }:
-    # TODO remove assert once `lua` option is gone
-    # This is here to ensure no uses of `mkMapOptionSubmodule` set a `lua` default
-    assert !(defaults ? lua);
-    (
-      with types;
-      submodule {
+    with types;
+    submodule (
+      { config, options, ... }:
+      {
         options =
           (optionalAttrs (isAttrs key || key) {
             key = mkOption (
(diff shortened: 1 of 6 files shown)


MattSturgeon:
```
plugins-completion-cmp-all-sources> Running test for all-sources
plugins-completion-cmp-all-sources> ERROR: please log in with :Codeium Auth
```
Seems we have a flaky test?

khaneliman:
> ```
> plugins-completion-cmp-all-sources> Running test for all-sources
> plugins-completion-cmp-all-sources> ERROR: please log in with :Codeium Auth
> ```
> 
> Seems we have a flaky test?
> 
> Also, not sure what this relates to:
> 
> ```
> warning: In a derivation named 'neovim-unwrapped-0.10.1', 'structuredAttrs' disables the effect of the derivation attribute 'disallowedRequisites'; use 'outputChecks.<output>.disallowedRequisites' instead
> ```
> 
> Probably something upstream?

I think these came up recently in the logs and I haven't had a chance to take a look at them. Probably need to create an issue to track em.

khaneliman (approved):
Feels weird that the merge would be completely ignored, resulting in needing to apply in every consumer of a function. But, in lieu of a better alternative it makes sense to me, for now. 

With the conversation around https://github.com/nix-community/nixvim/issues/1935 will this be another place we have a migration away from recommending `__raw`

MattSturgeon:
> Feels weird that the merge would be completely ignored

I don't fully understand it either, we had the same issue implementing `transitionType`, which you'd expect could just override `coercedTo`'s merge function.

My best guess is something in `mkOptionType` is using the `merge` passed to it, rather than the merge in the final attrs... Maybe this is something that could be resolved in nixpkgs?

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio dequeue

MattSturgeon:
@mergifyio queue

Issue #1625, opened 2024-06-03
[BUG] relative links on the index page don't always work


MattSturgeon:
## Description

When accessing the docs' index directly (as [/user-guide/install.html](https://nix-community.github.io/nixvim/user-guide/install.html)), relative links no longer work correctly (i.e. the links to [Home Manager Usage](https://nix-community.github.io/nixvim/modules/hm.html#home-manager-usage) and [Standalone Usage](https://nix-community.github.io/nixvim/modules/standalone.html#standalone-usage)).

## Minimal, Reproducible Example (MRE)

1. Visit https://nix-community.github.io/nixvim
2. Click a relative link
3. Everything works
4. Open `☰` (ToC sidebar)
5. Click ["Installation"](https://nix-community.github.io/nixvim/user-guide/install.html)
6. Repeat step 2 (click relative link)
7. This time the link is broken (points to e.g. [/user-guide/modules/hm.html](https://nix-community.github.io/nixvim/user-guide/modules/hm.html), note the extra `/user-guide` prefix)

MattSturgeon:
I guess we could work around this issue by having a separate index page that isn't `install.html`?

MattSturgeon:
We may be able to leverage mdbook's special handling of README files?

> Links to README.md will be converted to index.html. This is done since some services like GitHub render README files automatically, but web servers typically expect the root file to be called index.html.
https://rust-lang.github.io/mdBook/format/markdown.html#links

PR #2038, opened 2024-08-18
docs: fix wrong modules usage links

HeitorAugustoLN:
Currently, clicking those links leads to a page like this:

image

As they are linked to the wrong files. This pull request fixes it.


MattSturgeon:
This looks like it's attempting to fix https://github.com/nix-community/nixvim/issues/1625, but unfortunately this won't fully solve the issue because those relative links aren't always resolved from the same place.

Essentially, the `install` page can be accessed via either `/` or `/user-guide/install.html`. It is usually accessed at the former location, however relative links are broken at the latter location.

HeitorAugustoLN:
> All this PR does is flip the issue, so now links from the `/` location will be broken while links from `/user-guide/install.html` will work.

Indeed, you are correct. I had not previously noticed that.

PR #2039, opened 2024-08-18
tests: use a link-farm again, but only per-file

MattSturgeon:
- **tests: use a link-farm again, but only per-file**
- **Revert "tests: Allow to test multiple derivations in a single test derivation"**

This is one change picked from #1989, which reduces the scope of that refactoring PR.

This doesn't take any steps to reduce the number of checks we have on the flake, but it does remove some fragile derivation code that was attempting to optimise derivation-count in an area where (AFAICT) this should not affect buildbot.

cc @zowoq (buildbot maintainer) @traxys (reverted commit author)


GaetanLepage (approved):
LGTM !

MattSturgeon:
@mergifyio queue

PR #2040, opened 2024-08-18
lib/keymap: allow extra options/modules in `mkMapOptionSubmodule`

MattSturgeon:
This is needed to allow plugins to add bespoke features to their keymap submodules without having to re-implement the whole thing.

cc @khaneliman as this may be useful in todo-comments


khaneliman (approved):
LGTM this will support the upcoming key map refactor in todo.

MattSturgeon:
@mergifyio queue

PR #2041, opened 2024-08-19
tests: group tests into several link-farms (10 tests per farm)

MattSturgeon:
- **tests/lib-test: correctly print expected/result**
- **lib/util: add `groupListBySize`**
- **tests: group tests by 10**
- **flake-modules/devshell: update tests command with link-farm support**

In an attempt to strike a balance between having one very intensive test and _many_ smaller tests at the `checks.<system>.*` level, this PR groups the tests into bundles of 10, currently resulting in 28 `checks.<system>.test-*` link farms and 46 _total_ checks.

Quoting https://github.com/nix-community/nixvim/pull/2015#issuecomment-2285508636:
> On the old hardware we were using 16 workers for eval, now we are using 32. Only splitting the linkfarm up a bit (e.g. into 4) likely won't help much as they will still end up being evaluated at the same time but by more than that or by path should work.

We can always dial the group size up/down to find the right balance, but 28 groups of 10 feels like a good starting point (112 groups of 10, when multiplied by the 4 systems).

As per https://github.com/nix-community/nixvim/issues/1878#issuecomment-2240793881, we can test memory usage locally using:
```shell
nix run nixpkgs#nix-eval-jobs -- --workers 1 --check-cache-status --force-recurse --flake .#checks
```

cc @zowoq
cc @GaetanLepage

@traxys I've had to re-work some details of your `tests` shell script, supplying it with a pre-generated `tests.json` file, which looks like (full file: [tests.json](https://github.com/user-attachments/files/16655370/tests.json)):

```json
{
  "examples": "test-1.passthru.entries.examples",
  "extended-lib": "test-1.passthru.entries.extended-lib",
  "modules-clipboard": "test-1.passthru.entries.modules-clipboard",
  "modules-commands": "test-1.passthru.entries.modules-commands",
  "modules-editorconfig": "test-1.passthru.entries.modules-editorconfig",
  "modules-files": "test-1.passthru.entries.modules-files",
  "modules-filetypes": "test-2.passthru.entries.modules-filetypes",
  "modules-keymaps": "test-2.passthru.entries.modules-keymaps",
  "modules-lua-loader": "test-2.passthru.entries.modules-lua-loader",
  "modules-options": "test-2.passthru.entries.modules-options",
  "plugins-bufferlines-barbar": "test-3.passthru.entries.plugins-bufferlines-barbar",
  "plugins-bufferlines-bufferline": "test-3.passthru.entries.plugins-bufferlines-bufferline",
  "plugins-colorschemes-base16": "test-3.passthru.entries.plugins-colorschemes-base16",
  "plugins-colorschemes-catppuccin": "test-3.passthru.entries.plugins-colorschemes-catppuccin",
}
```

I _think_ doing this via `passthru` avoids IFD, but I'm unsure how much cost this has added to evaluating/building the devshell.

> [!Note]
> When using the `tests` command, you should no-longer pass a `"test-"` prefix.
> I figured the prefix was redundant in the context of `passthru.entries`.

### DIFF
--- a/flake-modules/dev/devshell.nix
+++ b/flake-modules/dev/devshell.nix
@@ -7,6 +7,7 @@
       lib,
       pkgs,
       config,
+      self',
       system,
       ...
     }:
@@ -44,9 +45,27 @@
 
                     text = builtins.readFile ./launch-test.sh;
                   };
+
+                  tests =
+                    let
+                      checks' = self'.checks;
+                      names = builtins.filter (n: builtins.match "test-.*" n != null) (builtins.attrNames checks');
+                    in
+                    builtins.listToAttrs (
+                      builtins.concatMap (
+                        checkName:
+                        map (testName: {
+                          name = testName;
+                          value = "${checkName}.passthru.entries.${testName}";
+                        }) (builtins.attrNames checks'.${checkName}.passthru.entries)
+                      ) names
+                    );
                 in
                 ''
-                  NIXVIM_SYSTEM=${system} NIXVIM_NIX_COMMAND=${nix} ${lib.getExe launchTest} "$@"
+                  export NIXVIM_SYSTEM=${system}
+                  export NIXVIM_NIX_COMMAND=${nix}
+                  export NIXVIM_TESTS=${pkgs.writers.writeJSON "tests.json" tests}
+                  ${lib.getExe launchTest} "$@"
                 '';
             }
             {
--- a/flake-modules/dev/launch-test.sh
+++ b/flake-modules/dev/launch-test.sh
@@ -36,8 +36,7 @@ nix_args=()
 interactive=false
 
 mk_test_list() {
-  nix eval ".#checks.${system}" --apply builtins.attrNames --json |
-    jq -r 'map(select(startswith("test-")))[]'
+  jq -r 'keys[]' "${NIXVIM_TESTS}"
 }
 
 while true; do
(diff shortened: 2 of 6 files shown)


MattSturgeon:
Initial feedback from buildbot is good: built in 4.5m, compared with the last few builds which took ~1h:

image

Obviously not statistically significant, since this is only one build.

khaneliman:
> Initial feedback from buildbot is good: built in 4.5m, compared with the last few builds which took ~1h:
> Obviously not statistically significant, since this is only one build.

I think it was also faster at this time of day during the last few PR's that I built. Might be good to look at what the runs look like from similar time periods.

MattSturgeon:
Running `nix-eval-jobs` on this branch seems to ramp up memory and then suddenly gc, I'm not sure the best way to measure RAM usage, but just observing my total system RAM usage, it seems to ramp up to a max of around 12GB (sometimes less) before dropping back down to around 5-6GB, then slowly ramping up again.

image

This isn't very scientific though, because I have several browser tabs (etc) open in the background.

zowoq:
I ran the buildbot job again a couple of times and I've ran just the nix-eval-jobs command on the CI host a few times, it all looks good. Thank you for working on this.

khaneliman (approved):
From the bits that I evaluated, I could see marginal memory footprint increase locally but the performance in buildbot looks magnitudes better. Everything seems to function fine on my desktop and laptop (x86 linux and arm darwin). As long as it looks alright for team maintaining buildbot it seems to be a good change.

MattSturgeon:
@mergifyio queue

PR #2045, opened 2024-08-20
tests: move `tests.dontRun` to a module option

MattSturgeon:
## Context
Historically, we've had a special `.tests` attr in test-cases that was removed before evaluating the test-case as a module.

This made defining function-modules awkward, because we can't use `removeAttrs` on a function. We either had to use `isFunction`, `isAttrs` or provide an alternative way to define function-modules; we went with defining function modules as `.module`.

This PR enables us to clean all of this up, by replacing the `tests` attr with a new `test` option set that _is_ evaluated by the module system.

## Overview
Introduces the `test` top-level module, for configuring nixvim test cases.

Replaces the `tests.dontRun` attr used in tests previously with a new `test.runNvim` option.

This further allows us to drop the special handling of attr/function modules and just treat _all_ test cases as modules.

## Deprecation
Deprecates the `dontRun` argument, historically passed to the test-derivation helpers (`mkTestDerivationFromNvim` and `mkTestDerivationFromNixvimModule`).

This is no longer used internally, but may be used by downstream (user) code. I've used `lib.warn` to recommend users use the new module option instead.

## Looking forward
In the future the `test` option-set could be used to define "expectations" such as _"XYZ assertion should be true"_ or _"a warning matching `regex` should be set"_. We may also consider building the test derivation as an output option (`config.test.derivation`?), instead of directing users to use separate helper functions.

## Review notes
Based on work started in #1989, which should simplify that PR once rebased.

NOTE: the vast majority of work in this PR is done in `modules/top-level/test.nix`, `tests/default.nix` and `tests/test-derivation.nix`; all changes in `tests/test-sources` are purely migration to the new system.

### DIFF
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -10,7 +10,7 @@
       { pkgs, config, ... }:
       import ../lib {
         inherit pkgs lib;
-        inherit (config.legacyPackages) makeNixvim makeNixvimWithModule;
+        inherit (config.legacyPackages) makeNixvimWithModule;
       }
     )
   );
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1,13 +1,13 @@
 # Args probably only needs pkgs and lib
 {
-  makeNixvim,
   makeNixvimWithModule,
   pkgs,
+  lib ? pkgs.lib,
   _nixvimTests ? false,
   ...
 }@args:
 {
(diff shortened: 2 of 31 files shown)


khaneliman (approved):
Clean, I like it.

MattSturgeon:
@mergifyio queue

PR #2046, opened 2024-08-20
plugins/todo-comments: migrate keymaps to mkMapOptionSubmodule

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/2030 can now be done since https://github.com/nix-community/nixvim/pull/2040 is implemented.

### DIFF
--- a/plugins/utils/todo-comments.nix
+++ b/plugins/utils/todo-comments.nix
@@ -9,18 +9,12 @@ let
   inherit (lib.nixvim)
     defaultNullOpts
     keymaps
+    mkNullOrOption'
     mkPackageOption
-    mkCompositeOption
     transitionType
     ;
   types = lib.nixvim.nixvimTypes;
 
-  keymapsActions = {
-    todoQuickFix = "TodoQuickFix";
-    todoLocList = "TodoLocList";
-    todoTrouble = "TodoTrouble";
-    todoTelescope = "TodoTelescope";
-  };
 in
 lib.nixvim.neovim-plugin.mkNeovimPlugin config {
   name = "todo-comments";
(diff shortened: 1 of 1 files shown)


plugins/utils/todo-comments.nix
```diff
@@ -398,11 +397,12 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
               default = null;
               example = "TODO,FIX";
             };
-
-            options = keymaps.mapConfigOptions;
           };
-      in
-      mapAttrs mkKeymapOption keymapsActions;
+        };
+        apply = v: if v == null then null else keymaps.removeDeprecatedMapAttrs v;
```
MattSturgeon, line 402: Shouldn't need this apply function, since we don't set `lua = true`

MattSturgeon (changes requested):
Otherwise looks good

plugins/utils/todo-comments.nix
```diff
@@ -371,17 +371,16 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
-              default = null;
-              description = "Key for the `${funcName}` function.";
-            };
+    keymaps = mapAttrs (
+      optionName: funcName:
+      mkNullOrOption' {
+        type = keymaps.mkMapOptionSubmodule {
+          defaults = {
+            mode = "n";
+            action = "<cmd>${funcName}<cr>";
```
MattSturgeon, line 380:
```suggestion
            action = funcName;
```

Maybe rename the argument so we can do `inherit action`?

plugins/utils/todo-comments.nix
```diff
@@ -431,11 +430,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
-      (filterAttrs (n: v: v != null))
+      (filterAttrs (n: keymap: keymap != null))
       (mapAttrsToList (
         name: keymap: {
-          inherit (keymap) key options;
-          mode = "n";
+          inherit (keymap) key mode options;
           action =
             let
               cmd = keymapsActions.${name};
```
MattSturgeon, line 439:
```suggestion
              cmd = keymap.action;
```

plugins/utils/todo-comments.nix
```diff
@@ -398,11 +397,11 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
             };
-
-            options = keymaps.mapConfigOptions;
           };
-      in
-      mapAttrs mkKeymapOption keymapsActions;
+        };
+        description = "Keymap for function ${funcName}";
+      }
+    ) keymapsActions;
```
MattSturgeon, line 404: I think we can inline `keymapsActions` now, since it'll only be used here.

I'd consider using a list together with `lib.nixvim.upperFirstChar` instead of an attrset, though this isn't too important.

MattSturgeon (commented):
The `action` in the plugin's keymap options should just be the attr values from `keymapsActions`

plugins/utils/todo-comments.nix
khaneliman: Actually, this would break the keymap, no? We construct a new action from the combination fo the option name and extraOptions.

khaneliman: Seems like it's getting thrown for a loop because we didn't define action as something the user could choose before. So now, we want to inherit action unless cwd or keywords are defined. Unless we can inject those strings before <cr> conditionally.

plugins/utils/todo-comments.nix
```diff
@@ -372,37 +366,44 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
-      in
-      mapAttrs mkKeymapOption keymapsActions;
+            description = "Keymap for function ${action}";
+          }
+        )
+        {
+          todoQuickFix = "<cmd>TodoQuickFix<cr>";
+          todoLocList = "<cmd>TodoLocList<cr>";
+          todoTrouble = "<cmd>TodoTrouble<cr>";
+          todoTelescope = "<cmd>TodoTelescope<cr>";
```
MattSturgeon, line 405:
```suggestion
          todoQuickFix = "TodoQuickFix";
          todoLocList = "TodoLocList";
          todoTrouble = "TodoTrouble";
          todoTelescope = "TodoTelescope";
```

We will add the `<cmd>` and `<cr>` _later_ when defining the top-level keymap's action.

plugins/utils/todo-comments.nix
```diff
@@ -431,14 +432,13 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
...
+      (filterAttrs (n: keymap: keymap != null))
       (mapAttrsToList (
         name: keymap: {
-          inherit (keymap) key options;
-          mode = "n";
+          inherit (keymap) key mode options;
           action =
             let
-              cmd = keymapsActions.${name};
+              cmd = keymap.action;
```
MattSturgeon, line 441: tbh this is probably more readable if we inline this into the action string below?

```suggestion
```

and then

```nix
"<cmd>${keymap.action}${cwd}${keywords}<cr>";
```

plugins/utils/todo-comments.nix
khaneliman: Sorry, I misinterpreted a different suggested change

plugins/utils/todo-comments.nix
MattSturgeon: We don't need to have an `action` sub-option (we can pass `action=false` to `mkMapOptionSubmodule`), and it is a little pointless IMO, since it's just the same thing as the attrname tbh.

But I think the confusion here is that you're thinking of the plugin-keymap-option's `action` as the same as the _real_ (final) `action` to be used in the top-level `keymaps` option.

MattSturgeon (approved):
Awesome, thanks!

khaneliman:
@mergifyio queue

PR #2048, opened 2024-08-21
modules/test: check warnings/assertions

MattSturgeon:
Extracted from #1989 to reduce its scope. Also based on earlier work from #1986.

This PR moves responsibility for assertions/warnings in tests to the test derivation, instead of printing/throwing them.

They are still printed/thrown when building a non-test nixvim build.

Additionally, two new options are added to control this behaviour. Both default to `true`:
- `test.checkWarnings`: Whether to check `config.warnings` in the test.
- `test.checkAssertions`: Whether to check `config.assertions` in the test.

> [!NOTE]
> Only _module_ warnings/assertions defined using the respective options from `assertions.nix` are checked.
> Anything printed manually with `lib.warn`, `builtins.trace`, etc, or errors thrown with `assert`, `throw`, `abort`, etc _will not_ be handled correctly by the test.
> - Errors would still fail CI; the test would fail before even running
> - Warnings printed manually will just print to the console, bypassing the test

In the future we can expand on this to allow "expecting" assertions, which will make our test system more flexible.

### DIFF
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -6,6 +6,9 @@
 }:
 let
   cfg = config.test;
+
+  inherit (config) warnings;
+  assertions = lib.nixvim.modules.getAssertionMessages config.assertions;
 in
 {
   options.test = {
@@ -18,6 +21,18 @@ in
     runNvim = lib.mkOption {
       type = lib.types.bool;
       description = "Whether to run `nvim` in the test.";
+      default = true;
+    };
+
+    checkWarnings = lib.mkOption {
+      type = lib.types.bool;
+      description = "Whether to check `config.warnings` in the test.";
+      default = true;
+    };
+
+    checkAssertions = lib.mkOption {
+      type = lib.types.bool;
+      description = "Whether to check `config.assertions` in the test.";
       default = true;
     };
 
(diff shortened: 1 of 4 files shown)


MattSturgeon:
I wonder if it'd be worth including a `test.ignoreWarnings` option in this PR?

It may make the schemastore changes simpler, since we wouldn't have to avoid _creating_ the warnings.

khaneliman:
> I wonder if it'd be worth including a `test.ignoreWarnings` option in this PR?
> 
> It may make the schemastore changes simpler, since we wouldn't have to avoid _creating_ the warnings.

My first reaction is that it would be nice to have that functionality. But, I'm trying to think of how often we'd do that. Are there cases that the test might include warnings and we need to silence them so we can keep the test enabled to watch for errors? That could be nice for maintenance.

MattSturgeon:
> I'm trying to think of how often we'd do that. Are there cases that the test might include warnings and we need to silence them so we can keep the test enabled to watch for errors?

The only other example is [nvim-osc52](https://github.com/nix-community/nixvim/blob/38b09c1621b07518b8ba51d68ea4153216a09569/tests/test-sources/plugins/utils/nvim-osc52.nix), a plugin that prints a deprecation warning when enabled.

I've updated this PR to just `mkForce` remove the warnings.

In the future we can extend the `test` module with additional options, but let's not let the scope of this PR creep out of control. :grin:

modules/top-level/test.nix
```diff
@@ -38,19 +38,44 @@ in
...
+            err + showErr name lines
+          ) "" { inherit assertions warnings; };
+        in
+        ''
+          ${lib.toShellVars { inherit errors; }}
+          if [ -n "$errors" ]; then
+            echo -n "$errors"
+            exit 1
+          fi
+        ''
```
MattSturgeon, line 65: We could move some of this logic into the nix eval by doing something like this instead:

```nix
optionalString (errors != "") ''
  echo -n ${lib.escapeShellArg errors}
  exit 1
''
```

Personal preference, it's fine to do it in the bash build script too.

modules/top-level/test.nix
```diff
@@ -38,19 +38,44 @@ in
...
+            lib.optionalString (lines != [ ]) ''
+              Unexpected ${name}:
+              ${lib.concatStringsSep "\n" (lib.map (v: "- ${v}") lines)}
+            '';
+
+          errors = lib.foldlAttrs (
+            err: name: lines:
+            err + showErr name lines
+          ) "" { inherit assertions warnings; };
+        in
```
MattSturgeon, line 58: Would it be better to have all this in the `let` block at the head of the file? (Where we currently only have `cfg = config.test`)

khaneliman: Up to you, I've started to find it easier to have the let blocks be close to their usage unless it's used through a module.

modules/top-level/test.nix
khaneliman: It actually feels kinda weird to take a nix variable and inject it into the bash just to evaluate it again in bash.

MattSturgeon:
> In the future we can extend the `test` module with additional options, but let's not let the scope of this PR creep out of control. 😁

I decided it wasn't too much work to just implement basic `checkWarnigns` and `checkAssertions` options in this PR.

khaneliman (approved):
LGTM

MattSturgeon:
@mergifyio queue

PR #2049, opened 2024-08-21
plugins/rustaceanvim: fix checkhealth error

khaneliman:
Rustaceanvim will throw an error if the configuration is sourced after initialization. Our globals are defined at the top of the init.lua so moving the settings to leverage the globals option scope. If lsp is is enabled, we append the on_attach function to the config as a callback function so that it can be accessed properly on the global config.

An example of running into this can be had while configuring `plugins.neotest.settings.adapters` for usage with rustaceanvim. The setup involves initializing rustaceanvim before our lsp configuration is configured.
```lua
require('neotest').setup {
    -- ...,
    adapters = {
      -- ...,
      require('rustaceanvim.neotest')
    },
}
```

I am able to just set the mkOrder/mkBefore for the lsp `extraConfigLua` but, it would still be nice to not have to worry about ordering and store the rustaceanvim configuration utilizing our globals option. While working on this approach, I did add a couple commits for setting up the init.lua ordering to make sure we have access to the local module table and unset runtime paths immediately, followed by the options so that all other configurations can benefit from those variables.

### DIFF
--- a/modules/opts.nix
+++ b/modules/opts.nix
@@ -84,6 +84,6 @@ in
           ) optionsAttrs
         );
       in
-      mkIf (content != "") content;
+      mkIf (content != "") (mkOrder 600 content); # Move options to top of file below global table
   };
 }
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -285,12 +285,21 @@ in
         '';
       };
 
-      extraConfigLuaPre = lib.mkIf config.wrapRc ''
-        -- Ignore the user lua configuration
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
-        vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
-      '';
+      extraConfigLuaPre = lib.mkOrder 100 (
+        # Add a global table at start of init
+        ''
+          -- Nixvim's internal module table
+          -- Can be used to share code throughout init.lua
+          local _M = {}
+        ''
+        + lib.optionalString config.wrapRc ''
+
+          -- Ignore the user lua configuration
+          vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
+          vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
+          vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
+        ''
+      );
 
       extraPlugins = lib.mkIf config.wrapRc [ config.filesPlugin ];
     };
--- a/plugins/languages/rust/rustaceanvim/default.nix
+++ b/plugins/languages/rust/rustaceanvim/default.nix
@@ -52,14 +52,12 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   callSetup = false;
   extraConfig =
     cfg:
-    let
-      configStr = ''
-        vim.g.rustaceanvim = ${helpers.toLuaObject cfg.settings}
-      '';
-    in
     mkMerge [
       {
         extraPackages = [ cfg.rustAnalyzerPackage ];
+
+        globals.rustaceanvim = cfg.settings;
+
         # TODO: remove after 24.11
         warnings =
           optional
@@ -78,13 +76,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       # If nvim-lspconfig is enabled:
       (mkIf config.plugins.lsp.enable {
         # Use the same `on_attach` callback as for the other LSP servers
-        plugins.rustaceanvim.settings.server.on_attach = mkDefault "__lspOnAttach";
-
-        # Ensure the plugin config is placed **after** the rest of the LSP configuration
-        # (and thus after the declaration of `__lspOnAttach`)
-        plugins.lsp.postConfig = configStr;
+        plugins.rustaceanvim.settings.server.on_attach = mkDefault ''
+          function(client, bufnr)
+            return _M.lspOnAttach(client, bufnr)
+          end
+        '';
       })
-      # Else, just put the plugin config anywhere
-      (mkIf (!config.plugins.lsp.enable) { extraConfigLua = configStr; })
     ];
 }
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -225,7 +225,8 @@ in
           ${cfg.preConfig}
 
           local __lspServers = ${helpers.toLuaObject cfg.enabledServers}
-          local __lspOnAttach = function(client, bufnr)
+          -- Adding lspOnAttach function to nixvim module lua table so other plugins can hook into it.
+          _M.lspOnAttach = function(client, bufnr)
             ${updateCapabilities}
 
             ${cfg.onAttach}
@@ -239,7 +240,7 @@ in
           end
 
           local __setup = ${runWrappers cfg.setupWrappers "{
-            on_attach = __lspOnAttach,
+            on_attach = _M.lspOnAttach,
             capabilities = __lspCapabilities()
           }"}
 


plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -78,18 +78,9 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
-            local rustaceanvim = vim.g.rustaceanvim
-            if rustaceanvim[server] == nil then
-              rustaceanvim.server = {}
-            end
-            rustaceanvim.server.on_attach = ${helpers.toLuaObject cfg.settings.server.on_attach}
-            vim.g.rustaceanvim = rustaceanvim
+        plugins.rustaceanvim.settings.server.on_attach = mkDefault ''
+          function()
+            return __nixvim.lspOnAttach
           end
```
MattSturgeon, line 84: I think we'd need to invoke the function here instead of returning it

```lua
function(client, bufnr)
  __nixvim.lspOnAttach(client, bufnr)
end
```

Depending on how lazy lua is, I guess we _might_ be able to do that without the new `__nixvim` table? I.e. if lua doesn't evaluate the function body until it is called... Not sure.

modules/top-level/output.nix
```diff
@@ -273,25 +273,35 @@ in
...
 
-      extraConfigLuaPre = lib.mkIf config.wrapRc ''
-        -- Ignore the user lua configuration
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
-        vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
-      '';
+        extraPlugins = lib.mkIf config.wrapRc [ config.filesPlugin ];
+      }
+      (lib.mkIf config.wrapRc {
```
khaneliman, line 296: does this need to be elevated to top, as well?

khaneliman: image

modules/opts.nix
```diff
@@ -84,6 +84,6 @@ in
           ) optionsAttrs
         );
       in
-      mkIf (content != "") content;
+      mkIf (content != "") (mkOrder 2 content); # Move options to top of file below global table
```
MattSturgeon, line 87: Rather than using order priorities `1` and `2`, it's probably better to spread them out a little...

The "default" priority is 1000, with 500 being "before" and 1500 being "after".

https://github.com/NixOS/nixpkgs/blob/0141f779e4d05fb114557542dad8e13d5c261c75/lib/modules.nix#L1047-L1049

The first question we should ask is should we be before or after `mkBefore` (500)? Perhaps `M = {}` should be lower than 500, but maybe opts should be higher?

The second question is how much space should we leave around our order priority values for users to position their own definitions. Perhaps sticking to round numbers, such as multiples of 20, 20, 50, etc?

In theory, a user may wish to be able to add some lines _before_ one of our definitions, but after the other. We should leave enough room for them to have some flexibility.

modules/top-level/output.nix
```diff
@@ -285,12 +285,19 @@ in
...
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
-        vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
-      '';
+      extraConfigLuaPre = lib.mkOrder 1 (
+        # Add a global table at start of init
+        ''
+          local M = {}
+
+        ''
+        + lib.optionalString config.wrapRc ''
```
MattSturgeon, line 294: Nit: this could lead to inconsistent line breaks.

If `wrapRc` is false you'll end up with `"\n\n\n"`; the third one coming from `concatLines` used in `types.lines` merge function.

plugins/lsp/default.nix
```diff
@@ -225,7 +225,7 @@ in
           ${cfg.preConfig}
 
           local __lspServers = ${helpers.toLuaObject cfg.enabledServers}
-          local __lspOnAttach = function(client, bufnr)
+          M.lspOnAttach = function(client, bufnr)
```
MattSturgeon, line 228: Because this is a new practice, maybe note in a comment why we are assigning to the `M` global?

modules/top-level/output.nix
```diff
@@ -285,12 +285,19 @@ in
...
-      extraConfigLuaPre = lib.mkIf config.wrapRc ''
-        -- Ignore the user lua configuration
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
-        vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
-      '';
+      extraConfigLuaPre = lib.mkOrder 1 (
+        # Add a global table at start of init
+        ''
+          local M = {}
```
MattSturgeon, line 291: Maybe have a lua comment instead of a nix comment, to add context when reading `nixvim-print-init`?

```lua
-- Nixvim's internal module
local M = {}
```

Not sure what we'd actually want to say though. Normally comments should explain _why_ we do something, but comments in the final init.lua don't need to do that as they are more like documentation and headers... :thinking:

MattSturgeon: Also cc @nix-community/nixvim: are we happy with this being `M` or should it be something like `_M` to indicate internal/private scope?

MattSturgeon (commented):
This looks fine to me generally. A couple nits and some minor points to consider & discuss.

Thanks for figuring out this issue!

Note @traxys: this may simplify part of #1876

modules/top-level/output.nix
traxys: Maybe we shoud do something like `_M`, as it may conflict with user defined variables

khaneliman: Yeah, that's fair. I don't want to break another user's config if they chose to follow the lua standard for a module table, as well.

modules/opts.nix
khaneliman: Were you thinking like 200/300 or 300/400 type of a assignment?

My intention originally was that we should have these global definitions definitely at the top because everything else needs them in scope and defined, as soon as possible. But, I could see a usage of weird `extraConfigLuaPre` injection and that would give them more flexibility with stacking usages.

MattSturgeon: I think anyone using `mkOrder` in their config should hopefully be knowledgeable enough to check what priority the thing they want to be relateve to is. Someone using `mkBefore` may not care as much, but will probably expect their definition to end up before pretty much all others.

I'd suggest that `local _M = {}` should have a fairly low priority. 100 or 200 seems fine. However I'm not convinced the opt definitions should trump `mkBefore`, so maybe something more like 600 or 700 there?

What do you think?

modules/top-level/output.nix
```diff
@@ -285,12 +285,21 @@ in
...
+      extraConfigLuaPre = lib.mkOrder 200 (
+        # Add a global table at start of init
+        ''
+          -- Nixvim's internal module table
+          -- Can be used to share code throughout init.lua
+          local _M = {}
+        ''
+        + lib.optionalString config.wrapRc ''
+
+          -- Ignore the user lua configuration
```
MattSturgeon, line 297: Can you double check this does what you want? Nix's indented strings are weird and strip leading newlines, I can't recall if they only trim the _first_ `"\n"` or all of them...

khaneliman: Sure, i'll double check real quick.

modules/opts.nix
khaneliman: In my head I wanted to ensure these both were at the very top because they seem foundational. I'm not sure what use case someone would have for trying to mkBefore the opts themselves, especially if just from a mkBefore. To me, it felt like it would lead to unexpected results to have a user injected lua config automatically jump above variables they might try and access.

modules/top-level/output.nix
khaneliman: image
LGTM

modules/opts.nix
MattSturgeon: That argument does make sense. To clarify my perspective:
- Users _already_ using `mkBefore` may be tripped up by this change
- Users may wish to define their own variables that are referenced from opts definitions
- I think it is less surprising for definitions wrapped using `mkBefore` to end up before our opts

Not saying I have a strong opinion though; I'm still somewhat on the fence.

modules/top-level/output.nix
MattSturgeon: I'll keep this in mind next time I want a leading newline. Thanks for checking!

modules/opts.nix
MattSturgeon: cc @GaetanLepage @traxys

modules/opts.nix
```diff
@@ -84,6 +84,6 @@ in
           ) optionsAttrs
         );
       in
-      mkIf (content != "") content;
+      mkIf (content != "") (mkOrder 300 content); # Move options to top of file below global table
```
MattSturgeon, line 87: I think a priority after `500` may be better, to avoid surprising users who use `mkBefore`.

modules/top-level/output.nix
```diff
@@ -285,12 +285,21 @@ in
         '';
       };
 
-      extraConfigLuaPre = lib.mkIf config.wrapRc ''
-        -- Ignore the user lua configuration
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
-        vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after
-        vim.opt.runtimepath:remove(vim.fn.stdpath('data') .. "/site")     -- ~/.local/share/nvim/site
-      '';
+      extraConfigLuaPre = lib.mkOrder 100 (
```
MattSturgeon, line 288: I think 100 is fine here, because we probably _do_ want to be before `mkBefore`.

MattSturgeon (approved):
No issues from me, but let's run this by @GaetanLepage and/or @traxys too as it's a somewhat significant change to how we define `extraConfigLuaPre`.

I'm still not convinced that `opts` should be defined before `mkBefore` definitions.

Note @traxys: this may simplify part of #1876

Note @GaetanLepage: this includes some internal refactoring, so pinging as requested.

modules/opts.nix
traxys: (for another PR) This makes me think that we may want to introduce new `late` or `post` variants of the option that are inserted at the end of the file

traxys (approved):

modules/opts.nix
khaneliman: Updated to mkOrder 600 to not have a surprising breaking change for people who might have been using mkBefore with expectations no options are available in that scope.

khaneliman:
@merifyio queue

PR #2052, opened 2024-08-22
plugins/neotest: migrate helpers -> lib.nixvim

khaneliman:
Was just looking at this module while checking something else out and saw it was still using helpers. Migrating to use lib.nixvim reference, instead.

### DIFF
--- a/plugins/neotest/default.nix
+++ b/plugins/neotest/default.nix
@@ -1,12 +1,14 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
 with lib;
-helpers.neovim-plugin.mkNeovimPlugin config {
+let
+  inherit (lib.nixvim) mkRaw;
+in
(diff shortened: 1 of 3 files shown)


plugins/neotest/default.nix
```diff
@@ -1,24 +1,26 @@
...
   imports = [ ./adapters.nix ];
 
-  settingsOptions = (import ./options.nix { inherit lib helpers; }) // {
+  settingsOptions = (import ./options.nix { inherit lib; }) // {
     adapters = mkOption {
-      type = with helpers.nixvimTypes; listOf strLua;
+      type = with types; listOf strLua;
       default = [ ];
-      apply = map helpers.mkRaw;
+      apply = map mkRaw;
```
MattSturgeon, line 23: Not sure this one usage is really worth the `inherit (from)`, but no big deal.

MattSturgeon (approved):
Awesome thanks!

khaneliman:
@mergifyio queue

PR #2061, opened 2024-08-22
docs: document how to install custom tree-sitter grammars

menixator:
Adds a section to tree-sitter's documentation describing how to install your own grammars.

### DIFF
--- a/plugins/languages/treesitter/treesitter.nix
+++ b/plugins/languages/treesitter/treesitter.nix
@@ -12,6 +12,98 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   luaName = "nvim-treesitter.configs";
   defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
+  description = ''
+    Provides an interface to [tree-sitter]
+
+    ### Installing Your Own Grammars with Nixvim
+
+    The grammars you want will usually be included in `nixGrammars` by default.
+    But, in the rare case it isn't, you can build your own and use it with Nixvim like so:
+
+    ```nix
+    { pkgs, ... }:
+    let
+      # Example of building your own grammar
+      treesitter-nu-grammar = pkgs.tree-sitter.buildGrammar {
+        language = "nu";
+        version = "0.0.0+rev=0bb9a60";
+        src = pkgs.fetchFromGitHub {
+          owner = "nushell";
+          repo = "tree-sitter-nu";
+          rev = "0bb9a602d9bc94b66fab96ce51d46a5a227ab76c";
+          hash = "sha256-A5GiOpITOv3H0wytCv6t43buQ8IzxEXrk3gTlOrO0K0=";
+        };
+        meta.homepage = "https://github.com/nushell/tree-sitter-nu";
+      };
+
+      # or you can yoink any grammars in tree-sitter.grammars.''${grammar-name}
+      # treesitter-nu-grammar = pkgs.tree-sitter-grammars.tree-sitter-nu;
+    in
+    {
+
+      programs.nixvim.plugins = {
+        treesitter = {
+          enable = true;
+          settings.indent.enable = true;
+          grammarPackages = pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars ++ [
+            treesitter-nu-grammar
+          ];
+        };
+
+        extraConfigLua =
+          '''
+            do
+              local parser_config = require("nvim-treesitter.parsers").get_parser_configs()
+              -- change the following as needed
+              parser_config.nu = {
+                install_info = {
+                  url = "''${treesitter-nu-grammar}", -- local path or git repo
+                  files = {"src/parser.c"}, -- note that some parsers also require src/scanner.c or src/scanner.cc
+                  -- optional entries:
+                  --  branch = "main", -- default branch in case of git repo if different from master
+                  -- generate_requires_npm = false, -- if stand-alone parser without npm dependencies
+                  -- requires_generate_from_grammar = false, -- if folder contains pre-generated src/parser.c
+                },
+                filetype = "nu", -- if filetype does not match the parser name
+              }
+            end
+          ''';
+
+        # Add as extra plugins so that their `queries/{language}/*.scm` get
+        # installed and can be picked up by `tree-sitter`
+        extraPlugins = [
+          treesitter-nu-grammar
+        ];
+
+      };
+
+    }
+    ```
+
+    The queries for the grammar should be added to one of the runtime directories under `queries/{language}` but sometimes plugins do not conform to this structure.
+
+    In such cases, you can override the source derivation (or the grammar derivation) to move the queries to the appropriate folder: ```nix
+    (
+      (pkgs.fetchFromGitLab {
+        owner = "joncoole";
+        repo = "tree-sitter-nginx";
+        rev = "b4b61db443602b69410ab469c122c01b1e685aa0";
+        hash = "sha256-Sa7audtwH8EgrHJ5XIUKTdveZU2pDPoUq70InQ6qcKA=";
+      }).overrideAttrs
+      (drv: {
+        fixupPhase = '''
+          mkdir -p $out/queries/nginx
+          mv $out/queries/*.scm $out/queries/nginx/
+        ''';
+      })
+    )
+    ```
+
+    Verify if the queries were picked up by running `:TSModuleInfo`.
+
+    [tree-sitter]: https://github.com/tree-sitter/tree-sitter
+  '';
+
   maintainers = [ lib.maintainers.khaneliman ];
 
   # TODO introduced 2024-07-06: remove after 24.11


plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+
+      # or you can yoink any grammars in tree-sitter.grammars.$\{grammar-name}
+      # treesitter-nu-grammar = pkgs.tree-sitter-grammars.tree-sitter-nu;
+    in
+    {
+
+      programs.nixvim.plugins = {
+        treesitter = {
+          enable = true;
+          indent = true;
```
khaneliman, line 48: this is a deprecated option, we should not be using it as an example. This is already a large section and it's not related to the functionality being documented, so it might make sense to just remove, as well. 
```suggestion
          settings.indent.enable = true;
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   luaName = "nvim-treesitter.configs";
   defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
+  description = ''
+    Provides an interface to [tree-sitter]
+
+    ## Installing Your Own Grammars with Nixvim
```
MattSturgeon, line 18: Will need to check using our `serve-docs` devshell command, but I have a feeling mdbook doesn't support markdown headers within the option descriptions. The plugin description may work differently though, not sure.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+      };
+
+      programs.nixvim.extraConfigLua =
+        '''
+          do
+            local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
+            -- change the following as needed
+            parser_config.nu = {
+              install_info = {
+                url = "''${treesitter-nu-grammar}", -- local path or git repo
```
khaneliman, line 62: Did you test this out? From my quick experimenting and following their [readme](https://github.com/nvim-treesitter/nvim-treesitter/?tab=readme-ov-file#adding-parsers) this path needs to find the files below but you link to the derivation above.

menixator: Yes. These two examples are actually from my config.

khaneliman: This section is used for `:TSInstall nu`. But, you install it below with `extraPlugins`. If you ran `:TSInstall nu` you should see a message about a parser not managed by treesitter.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+          grammarPackages = pkgs.vimPlugins.nvim-treesitter.passthru.allGrammars ++ [
+            treesitter-nu-grammar
+          ];
+        };
+      };
+
+      programs.nixvim.extraConfigLua =
+        '''
+          do
+            local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
```
MattSturgeon, line 58: Nit: I find examples are easier to read with explicit parens on function calls.
```suggestion
            local parser_config = require("nvim-treesitter.parsers").get_parser_configs()
```

(I haven't checked if the same change is needed elsewhere in the example)

plugins/languages/treesitter/treesitter.nix
khaneliman: I was able follow the example and set up [fsharp](https://github.com/khaneliman/astronvim/commit/71227a14435ce93997b3e79a04d8a8b761cfe6fa). But, in my own testing of the installation, I didn't find the `extraConfigLua` part to be necessary when using `grammarPackages` with  `nixGrammars`. That section would be relevant for user's not installing through `nixGrammars` and utilizing the `settings.ensure_installed`.

menixator: Hmm. This might be because I'm still on the `24.05` branch. I could not get it to work without adding it into the runtime path like that.

plugins/languages/treesitter/treesitter.nix
menixator: This was the only instance that I could find.

plugins/languages/treesitter/treesitter.nix
khaneliman: image

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        };
+        meta.homepage = "https://github.com/nushell/tree-sitter-nu";
+      };
+
+      # or you can yoink any grammars in tree-sitter.grammars.$\{grammar-name}
+      # treesitter-nu-grammar = pkgs.tree-sitter-grammars.tree-sitter-nu;
+    in
+    {
+
+      programs.nixvim.plugins = {
```
khaneliman, line 45: we should collapse all of these into a nested structure. 
```
programs.nixvim = {
  extraPlugins = [
  ...
  plugins = {
  ...
  extraConfigLua ==
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+    ## Installing Your Own Grammars with Nixvim
+
+    The `nvim-treesitter` derivation generates the grammar packages based on the
+    default configs of nvim-treesitter but you can always install your own
+    grammars.
+
+    ```nix
+    { pkgs, ... }:
+    let
+      #example of adding your own grammar:
```
khaneliman, line 27:
```suggestion
      # example of building your own grammar:
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,101 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+                --  branch = "main", -- default branch in case of git repo if different from master
+                -- generate_requires_npm = false, -- if stand-alone parser without npm dependencies
+                -- requires_generate_from_grammar = false, -- if folder contains pre-generated src/parser.c
+              },
+              filetype = "nu", -- if filetype does not match the parser name
+            }
+          end
+        ''';
+
+      # add them as extra plugins so that the plugin can pick up the for highlighting.
```
khaneliman, line 74:
```suggestion
      # Add as extra plugins so that their `queries/*.scm` get installed and can be picked up by `treesitter`. 
```

plugins/languages/treesitter/treesitter.nix
MattSturgeon: Awesome, thanks for checking! We could debate which heading level to use, but I don't want to get too nit-picky! (I'd probably prefer `###` or `####`, but wouldn't like to say without checking how it renders).

khaneliman: Looks better with `###`. One thing I'm noticing is that it gets rid of the plugin doc link and maintainer link with a description. We might want to fix that.
image

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,103 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   luaName = "nvim-treesitter.configs";
   defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
+  description = ''
+    Provides an interface to [tree-sitter]
+
+    ## Installing Your Own Grammars with Nixvim
```
khaneliman, line 18:
```suggestion
    ### Installing Your Own Grammars with Nixvim
```

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,103 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   luaName = "nvim-treesitter.configs";
   defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
+  description = ''
+    Provides an interface to [tree-sitter]
```
khaneliman, line 16:
```suggestion
```

plugins/languages/treesitter/treesitter.nix
khaneliman: >  One thing I'm noticing is that it gets rid of the plugin doc link and maintainer link with a description. We might want to fix that.

Nvm, I'll create a PR. I see what it's doing.

plugins/languages/treesitter/treesitter.nix
khaneliman: @menixator Sorry to flip on this, I didn't realize it would hotlink to actual tree-sitter and not the plugin again. I think it's safe to add this back.

khaneliman: Afterwards, can you squash all the commits into a single one.

menixator: @khaneliman done.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,103 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+
+    ### Installing Your Own Grammars with Nixvim
+
+    The `nvim-treesitter` derivation generates the grammar packages based on the
+    default configs of nvim-treesitter but you can always install your own
+    grammars.
```
MattSturgeon, line 22: I think this is confusingly worded. `pkgs.vimPlugins.nvim-treesitter` is a derivation (all packages are), but the `plugins.treesitter` module is _not_ a derivation.

I'm not sure it's strictly accurate to say the package derivation generates the grammar derivations, since that is done in this module by providing an overridden package using the `withPlugins` function.

Not sure the best way to word it though.

menixator: How about:
```
If the grammar you want to install isn't provided through `nixpkgs`, you could install it like so:
```

MattSturgeon: I think it's nice to keep an implication that most users don't need to do anything, and that configuring `plugins.treesitter.nixGrammars` and/or `plugins.treesitter.grammarPackages` should cover 80% of usage when users _do_ need to do something.

> If the grammar you want to install isn't available through `nixpkgs`, you could install it like so:

But something like that could definitely fit into the text!

menixator: > I think it's nice to keep an implication that most users don't need to do anything

If that's the case, these docs might be a better fit in some other document since it is sort of a niche case. Having it front and centre of such a ubiquitous plugin might scare people off 👀

MattSturgeon: I think it's fine to have it here. Let's not let perfect be the enemy of good!

menixator: I've reworded it. Let me know if that works for you.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,103 @@ helpers.neovim-plugin.mkNeovimPlugin config {
   luaName = "nvim-treesitter.configs";
   defaultPackage = pkgs.vimPlugins.nvim-treesitter;
 
+  description = ''
+    Provides an interface to [tree-sitter]
+
+    ### Installing Your Own Grammars with Nixvim
+
+    In most cases, the grammars you would will be available with `nixGrammars`.
+    But in the rare case isn't you can build your own and use it with Nixvim
```
khaneliman, line 21:
```suggestion
    But, in the rare case it isn't, you can build your own and use it with Nixvim
```

MattSturgeon (changes requested):
One suggestion I'd like addressed, and a couple _very_ minor nits which you're welcome to ignore.

When you're done are you also able to squash and rebase onto our `main` branch? No big deal if not as we can handle that.

Thanks for working on this, documentation is far too often neglected! :heart: 


@khaneliman what do you think about a follow up PR where we (somehow) include the examples in the test suite? E.g. they could be `readFile`-ed into the description and also added to the test-sources in a similar way to how `examples/` are.

khaneliman:
> @khaneliman what do you think about a follow up PR where we (somehow) include the examples in the test suite? E.g. they could be `readFile`-ed into the description and also added to the test-sources in a similar way to how `examples/` are.

I think it could be pretty useful if we filter out the default config so we don't double up on those examples in documentation. Could probably use a special attribute name designation in tests to be included in the description.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,103 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+
+    }
+    ```
+
+    The queries for the grammar should be added to one of the runtime
+    directories under `queries/{language}` but sometimes plugins do not conform
+    to this structure.
+
+    In such cases, you can override the source derivation (or the grammar
+    derivation) to move the queries to the appropriate folder:
```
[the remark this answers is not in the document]
menixator: Oh that's neovim's `gq` doing its thing haha. Let me just do it manually.

I should really add in something like http://www.nicemice.net/par/ or https://github.com/wincent/wincent/blob/main/aspects/nvim/files/.config/nvim/plugin/par.lua  to neovim to do that. Let me just do this manually.

menixator:
> When you're done are you also able to squash and rebase onto our `main` branch? No big deal if not as we can handle that.

Done. 😄

plugins/languages/treesitter/treesitter.nix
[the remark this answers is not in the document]
menixator: Fixed now.

MattSturgeon (approved):
Awesome, LGTM thanks!

I'll let @khaneliman have the final say since he handled most of the review.

khaneliman (approved):
LGTM, thanks for your help documenting how to use the module!

khaneliman:
@mergifyio queue

plugins/languages/treesitter/treesitter.nix
```diff
@@ -12,6 +12,98 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        ];
+
+      };
+
+    }
+    ```
+
+    The queries for the grammar should be added to one of the runtime directories under `queries/{language}` but sometimes plugins do not conform to this structure.
+
+    In such cases, you can override the source derivation (or the grammar derivation) to move the queries to the appropriate folder: ```nix
```
menixator, line 85: @MattSturgeon / @khaneliman    the ```nix should be moved to a newline here. It borked the documentation completely haha. 
https://nix-community.github.io/nixvim/plugins/treesitter/index.html?highlight=tree-sitter#installing-your-own-grammars-with-nixvim

khaneliman: Doh, if you wanna do a quick PR I can approve and merge

menixator: Done #2073

Issue #2063, opened 2024-08-22
[BUG] lightline is generating a wrong key name for component_function


jaidTw:
| Field        | Description |
|--------------|-------------|
| Plugin       | `lightline`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

The option [plugins.lightline.componentFunction](https://nix-community.github.io/nixvim/plugins/lightline/index.html#pluginslightlinecomponentfunction) has no effect.

I believe this is due to the generated lua code write this option to `vim.g.lightline.componentFunction`, which the key should really be `component_function` according to the plugin description, hence has no effect at all.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  extraConfigLua = ''
function myFunc()
  return "Hi"
end
'';
  plugins = {
      lightline = {
        enable = true;
        componentFunction = {
          myfunc = "v:lua.myFunc";
        };
        active = {
          right = [
            ["myfunc"]
          ];
      };
    };
  };
}
```

This should show Hi on the lightline

Generated `init.lua`:
```lua
-- Set up globals {{{
do
    local nixvim_globals = {
        lightline = {
            active = {
                right = {
                    { "myfunc" },
                },
            },
            -- This should be component_function
            componentFunction = { myfunc = "v:lua.myFunc" },
        },
    }

    for k, v in pairs(nixvim_globals) do
        vim.g[k] = v
    end
end
-- }}}
```

khaneliman:
Thanks for posting this! It should be addressed soon as we do a refactor of it. I might tackle this next if it has a known bug.

jaidTw:
Thanks for your quick fix!

PR #2069, opened 2024-08-23
plugins/lightline: migrate to mkNeovimPlugin

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2063

### DIFF
--- a/plugins/statuslines/lightline.nix
+++ b/plugins/statuslines/lightline.nix
@@ -1,117 +1,215 @@
 {
   lib,
-  helpers,
   config,
+  options,
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.lightline;
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
 in
-{
-  options = {
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "lightline";
+  originalName = "lightline.vim";
+  defaultPackage = pkgs.vimPlugins.lightline-vim;
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    ### Example of defining your own component_function
+
     plugins.lightline = {
-      enable = mkEnableOption "lightline";
+       enable = true;
+       settings.component_function = {
+         readonly = "LightlineReadonly";
+       };
+     };
 
-      package = helpers.mkPluginPackageOption "lightline" pkgs.vimPlugins.lightline-vim;
+     extraConfigLua = '''
+       function LightlineReadonly()
+         local is_readonly = vim.bo.readonly == 1
+         local filetype = vim.bo.filetype
 
-      colorscheme = mkOption {
-        type = with types; nullOr str;
-        default = config.colorscheme;
-        description = "The colorscheme to use for lightline. Defaults to .colorscheme.";
-        example = "gruvbox";
-      };
+         if is_readonly and filetype ~= "help" then
+           return "RO"
+         else
+           return ""
+         end
+       end
+     ''';
+  '';
 
-      componentFunction = mkOption {
-        default = null;
-        type = with types; nullOr (attrsOf str);
-        description = ''
-          A list of function component definitions.
+  # TODO: Added 2024-08-23, remove after 24.11
+  optionsRenamedToSettings = [
+    "colorscheme"
+    "componentFunction"
+    "component"
+    "active"
+    "inactive"
+    "modeMap"
+  ];
 
-          You should define the functions themselves in extraConfig
+  settingsOptions = {
+    colorscheme = defaultNullOpts.mkStr "default" ''
+      The colorscheme to use for lightline.
+      Default theme is equal to `powerline`.
+    '';
+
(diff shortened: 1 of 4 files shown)


plugins/statuslines/lightline.nix
```diff
@@ -1,117 +1,111 @@
...
-in
-{
-  options = {
-    plugins.lightline = {
-      enable = mkEnableOption "lightline";
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts mkNullOrOption;
 
-      package = helpers.mkPluginPackageOption "lightline" pkgs.vimPlugins.lightline-vim;
+  type = types.nullOr (
```
khaneliman, line 11: Not sure of a good name for this

plugins/statuslines/lightline.nix
```diff
@@ -1,117 +1,206 @@
...
+          lineinfo = "%3l=%-2c";
+          line = "%l";
+          column = "%c";
+          close = "%999X X ";
+          winnr = ''%{winnr()}'';
+        }
+        ''
+          Lightline component definitions.
+          Uses 'statusline' syntax.
+          Consult :h 'statusline' for a list of what's available.
```
MattSturgeon, line 86:
```suggestion
          Lightline component definitions. Uses 'statusline' syntax.

          Consult `:h 'statusline'` for a list of what's available.
```

plugins/statuslines/lightline.nix
```diff
@@ -1,117 +1,206 @@
...
-          modeMap
-          ;
-      };
-    in
-    mkIf cfg.enable {
-      extraPlugins = [ cfg.package ];
-      globals.lightline = mkIf (configAttrs != { }) configAttrs;
-    };
+  callSetup = false;
+  extraConfig = cfg: { globals.lightline = cfg.settings; };
```
MattSturgeon, line 205: I think we're moving towards preferring "real" aliases for these kinda things, to preserve wrappers such as `mkDefault` if the user for whatever reason defines config in _both_ `plugins.lightline.settings` _and_ `globals.lightline`...

e.g. https://github.com/nix-community/nixvim/blob/052ee66dbb975960cc151db0b5db6a893503582a/plugins/pluginmanagers/lz-n.nix#L195

and https://github.com/nix-community/nixvim/blob/052ee66dbb975960cc151db0b5db6a893503582a/plugins/utils/firenvim.nix#L92

```suggestion
  extraConfig = cfg: {
    globals.lightline = lib.modules.mkAliasAndWrapDefsWithPriority lib.id options.plugins.lightline.settings;
  };
```

MattSturgeon (changes requested):
Looks good overall

plugins/statuslines/lightline.nix
khaneliman: Thanks, I didn't realize that's what this function was. So it basically creates a 2 way binding with some sort of priority?

MattSturgeon: > So it basically creates a 2 way binding with some sort of priority

That's what `mkRenamedOptionModule` and `mkAliasOptionModule` do. This function only makes a one-way alias, but it does it in a way that preserves any priority wrappers used on definitions.

MattSturgeon: When I say "wrapped with a priority", I mean how for example `mkDefault` _wraps_ the config value as:

```nix
{
  _type = "override";
  priority = 1000;
  content = "foo";
}
```

plugins/statuslines/lightline.nix
```diff
@@ -1,117 +1,209 @@
...
-            ''';
+        You can use the name of a function defined elsewhere.
+      '';
+      example = ''
+        plugins.lightline = {
+          enable = true;
+          settings.component_function = {
+            readonly = "LightlineReadonly";
           };
+        };
```
MattSturgeon, line 49: This isn't an option example; option examples are usually _just_ the value of the option. In this case:
```nix
{
  readonly = "LightlineReadonly";
}
```

plugins/statuslines/lightline.nix
```diff
@@ -1,117 +1,209 @@
...
+
+        extraConfigLua = '''
+          function! LightlineReadonly()
+            return &readonly && &filetype !=# 'help' ? 'RO' : '''
+          endfunction
+        ''';
```
MattSturgeon, line 55: This looks like viml, not lua

plugins/statuslines/lightline.nix
MattSturgeon: I guess if the purpose of the example is to demonstrate referencing a function defined elsewhere, perhaps that should be documented in the option's description?

Or at least make an example using `literalMD` or `literalExpression`.

plugins/statuslines/lightline.nix
khaneliman: Yeah, I was just looking into this. Trying to figure out how to translate their viml to lua.

MattSturgeon (approved):
LGTM, thanks again!

khaneliman:
@mergifyio queue

Issue #2064, opened 2024-08-22
[BUG] Conform plugin config using deprecated syntax for stop after first formatter


347Online:
| Field        | Description |
|--------------|-------------|
| Plugin       | `conform-nvim` |
| Nixpkgs      | `unstable` |
| Home Manager | `unstable` |

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

Conform has deprecated the nested list syntax for stopping after the first available formatter ([#491](https://github.com/stevearc/conform.nvim/pull/491)). New syntax for this uses a flat table, with an additional option `stop_after_first`. Using the previous syntax causes a warning to be shown the first time you the formatter in a given instance of neovim.

I find I end up closing and re-opening Neovim quite often, so I'm seeing this warning extremely frequently and it is very disruptive.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
    enable = true;
    plugins.conform-nvim = {
        # See https://github.com/nix-community/nixvim/blob/1854d591cb0e5be6ad97f5091766cdf28e948265/plugins/lsp/conform-nvim.nix#L30
        # Use a sub-list to run only the first available formatter
        javascript = [ [ "prettierd" "prettier" ] ];
    };
}
```

khaneliman:
Documentation looks like that option supports an attribute set of anything. https://nix-community.github.io/nixvim/plugins/conform-nvim/index.html#pluginsconform-nvimformattersbyft You should be able to do this yourself in its current form.
```nix
        javascript = {
          __unkeyed-1 = "prettierd";
          __unkeyed-2 = "prettier";
          timeout_ms = 2000;
          stop_after_first = true;
        };
```

347Online:
slick!

khaneliman:
Thanks for bringing this up... while looking through documentation for the refactor, it looks like we were multiple versions behind on options and examples.

347Online:
My pleasure! Thanks for the lightning-fast response 😁

PR #2070, opened 2024-08-23
plugins/conform-nvim: migrate to mkNeovimPlugin

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2064

### DIFF
--- a/plugins/lsp/conform-nvim.nix
+++ b/plugins/lsp/conform-nvim.nix
@@ -1,24 +1,100 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.conform-nvim;
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts mkRaw;
 in
-{
-  options.plugins.conform-nvim = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "conform-nvim";
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "conform-nvim";
+  luaName = "conform";
+  originalName = "conform.nvim";
+  defaultPackage = pkgs.vimPlugins.conform-nvim;
 
-    package = helpers.mkPluginPackageOption "conform-nvim" pkgs.vimPlugins.conform-nvim;
+  maintainers = [ lib.maintainers.khaneliman ];
 
-    formattersByFt = helpers.defaultNullOpts.mkAttrsOf' {
-      type = types.anything;
-      # Unknown plugin default
-      description = ''
+  # TODO: added 2024-08-23 remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    "formatters"
+    "formattersByFt"
+    "logLevel"
+    "notifyOnError"
+  ];
+  imports =
+    map
+      (
+        optionName:
+        lib.mkRemovedOptionModule
+          [
+            "plugins"
+            "conform-nvim"
+            optionName
+          ]
+          ''
+            Please use `plugins.conform-nvim.settings.${lib.nixvim.toSnakeCase optionName}` instead.
+
+            Note that nested options will now be snake_case, as well, to match upstream plugin configuration.
+          ''
+      )
+      [
+        "formatAfterSave"
+        "formatOnSave"
+      ];
+
+  settingsOptions =
+    let
+      lsp_format =
+        defaultNullOpts.mkEnumFirstDefault
+          [
+            "never"
+            "fallback"
+            "prefer"
+            "first"
+            "last"
+          ]
+          ''
+            Option for choosing lsp formatting preference.
+
+            - `never`: never use the LSP for formatting.
+            - `fallback`:  LSP formatting is used when no other formatters are available.
+            - `prefer`: Use only LSP formatting when available.
+            - `first`: LSP formatting is used when available and then other formatters.
+            - `last`: Other formatters are used then LSP formatting when available.
+          '';
+
+      timeout_ms = defaultNullOpts.mkUnsignedInt 1000 "Time in milliseconds to block for formatting.";
+
+      quiet = defaultNullOpts.mkBool false "Don't show any notifications for warnings or failures.";
+
+      stop_after_first = defaultNullOpts.mkBool false "Only run the first available formatter in the list.";
+
+      # NOTE: These are the available options for the `default_format_opts` opt.
+      # They are also included in the `format_opts` options available to `format_after_save` and `format_on_save`.
+      defaultFormatSubmodule =
+        with types;
+        (submodule {
+          freeformType = attrsOf anything;
+          options = {
+            inherit
+              lsp_format
+              timeout_ms
+              quiet
+              stop_after_first
+              ;
+          };
+        });
+    in
+    {
+      formatters_by_ft = defaultNullOpts.mkAttrsOf types.anything { } ''
+        Creates a table mapping filetypes to formatters.
+
+        You can run multiple formatters in a row by adding them in a list.
+        If you'd like to stop after the first successful formatter, set `stop_after_first`.
+
         ```nix
           # Map of filetype to formatters
           formattersByFt =
@@ -26,8 +102,12 @@ in
               lua = [ "stylua" ];
               # Conform will run multiple formatters sequentially
               python = [ "isort" "black" ];
-              # Use a sub-list to run only the first available formatter
-              javascript = [ [ "prettierd" "prettier" ] ];
+              # Use stop_after_first to run only the first available formatter
+              javascript = {
+                __unkeyed-1 = "prettierd";
+                __unkeyed-2 = "prettier";
+                stop_after_first = true;
+              };
               # Use the "*" filetype to run formatters on all filetypes.
               "*" = [ "codespell" ];
               # Use the "_" filetype to run formatters on filetypes that don't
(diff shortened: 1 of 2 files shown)


plugins/lsp/conform-nvim.nix
```diff
@@ -1,21 +1,31 @@
...
-    enable = mkEnableOption "conform-nvim";
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+  name = "conform-nvim";
+  originalName = "conform.nvim";
+  defaultPackage = pkgs.vimPlugins.conform-nvim;
 
-    package = helpers.mkPluginPackageOption "conform-nvim" pkgs.vimPlugins.conform-nvim;
+  # TODO: added 2024-08-23 remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
```
MattSturgeon, line 18: `optionsRenamedToSettings` will converts the rename aliases to snake_case, however looking at your `settingsOptions` and test-cases they should still still be using camelCase.

If so, you'll have to use `mkRenameOptionModule` manually.

khaneliman: Sorry, I haven't gotten around to finishing this. I was just posting it to link to the issue and show it was being worked on.

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +86,159 @@ in
...
+              freeformType = types.attrsOf anything;
+              options = {
+                inherit
+                  lsp_format
+                  timeout_ms
+                  quiet
+                  stop_after_first
+                  ;
+              };
+            })
```
MattSturgeon, line 128: It seems this type is common to a few options? Perhaps the submodule type, or even the whole `either` type, should be defined in a let block somewhere? Unless I'm missing some subtle differences between them.

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +86,159 @@ in
...
+          )
+          { }
+          ''
+            If this is set, Conform will run the formatter asynchronously after save.
+            It will pass the table to `conform.format()`.
+            This can also be a function that returns the table.
+            See `:help conform.format` for details.
```
MattSturgeon, line 156: Personally I like to start a new paragraph for footnotes like _"See XYZ for more info."_

It may also be helpful to have either a line break (`\` before eol) or a paragraph break (a blankline) for somewhat separate thoughts such as _"This can also be a function that returns the table."_, though I feel less strongly about that :sweat_smile: 

```suggestion
            If this is set, Conform will run the formatter asynchronously after save.

            It will pass the table to `conform.format()`.
            This can also be a function that returns the table.

            See `:help conform.format` for details.
```

MattSturgeon: I'm also not keen on the over-use of _"it"_; I think it adds more clarity to be a bit more verbose about what _"it"_ actually is.

Ironically, I think my point is proven by the fact I don't know what _"it"_ should be replaced with, therefore I'll forgive you if you also aren't sure :joy:

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +86,159 @@ in
...
-    let
-      setupOptions =
-        with cfg;
-        {
-          formatters_by_ft = formattersByFt;
-          format_on_save =
-            if builtins.isAttrs formatOnSave then
-              {
-                lsp_fallback = formatOnSave.lspFallback;
-                timeout_ms = formatOnSave.timeoutMs;
```
MattSturgeon, line 110: I'm not sure users who had `formatOnSave.*` defined in their configs will get a graceful migration to the new snake_case names in `settings.format_on_save.*`... Seeing as it's a union type, I can't think of a robust way to offer such migration either... :thinking:

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +86,159 @@ in
...
-              }
-            else
-              helpers.mkRaw formatOnSave;
-          format_after_save =
-            if builtins.isAttrs formatAfterSave then
-              { lsp_fallback = formatOnSave.lspFallback; }
```
MattSturgeon, line 116: Same. Also, it seems the old implementation was broken; it is assigning `formatOnSave.*` to `formatAfterSave.*`!

plugins/lsp/conform-nvim.nix
```diff
@@ -1,33 +1,83 @@
...
+      "lspFallback"
+    ]
+    [
+      "formatOnSave"
+      "lspFallback"
+    ]
+    [
+      "formatOnSave"
+      "timeoutMs"
+    ]
```
MattSturgeon, line 32: Re https://github.com/nix-community/nixvim/pull/2070#discussion_r1729802391:

I see you're defining rename modules for these; I think these will fail non-gracefully if users have a string definition for `formatAfterSave` or `formatOnSave` in their configs though, since we're only creating deprecated options for the old submodule options.

I don't think it's (trivially) possible to gracefully handle _both_ the submodule usage and the string usage.

We could either mark the `formatAfterSave` and `formatOnSave` options as removed, so users with either type definition will get an error message. Or we could try and come up with a bespoke implementation of a rename module that covers both types.

MattSturgeon: If we did go down the bespoke rename module route, we could probably have a `types.anything` option\* that _if defined_ checks whether its definition is a string or an attrs, then aliases the resulting config making use of `toSnakeCase`.

_\* Or just keep the existing type I guess..._

If you think this is worth doing I can try and help with the implementation.

plugins/lsp/conform-nvim.nix
khaneliman: They are just a few options that are shared between a few of the properties. There are a lot more supported options but these ones are the most likely to be used.

plugins/lsp/conform-nvim.nix
khaneliman: Not sure what you mean by this? It's similar to other migrations from the renamedToSettings we've been doing.

plugins/lsp/conform-nvim.nix
khaneliman: Sorry, this explains better what you were saying above. I think I had a TODO to come back to this but must have removed it. I started off with a rename for just the parent options and then realized it wouldn't do the rename for the suboptions. Doing both in the same block failed with option doesn't exist errors because the rename happens for the parent option first. 

I'm using strings in my config for these options so I can play around with it a bit. Unless you have an idea in mind for implementation.

MattSturgeon: We could implement _most_ of a rename alias by

1. Keeping the old option declaration (in `extraOptions`)
2. Removing its default 
3. Adding an `optionalAttrs` config, conditional on `options.*.isDefined`
4. Checking for string vs attrs and aliasing as appropriate
5. Also manually implement warnings, preferably complete with `showDefs`

But again, is this option important enough to do this instead of a removal assertion?

plugins/lsp/conform-nvim.nix
khaneliman: I didn't even notice this... that's a bit of a mistake haha

khaneliman:
Updated to mark those options as removed, user's just need to update to nest them under the new `settings` option so migration is quick and painless.

plugins/lsp/conform-nvim.nix
```diff
@@ -1,33 +1,101 @@
...
+  ];
+  imports =
+    map
+      (
+        optionName:
+        lib.mkRemovedOptionModule [
+          "plugins"
+          "conform-nvim"
+          optionName
+        ] "Please use `plugins.conform-nvim.settings.${lib.nixvim.toSnakeCase optionName}` instead."
```
MattSturgeon, line 34: This should ideally also note that sub-option(s) (if any) would also be snake_case.

Unfortunately there's no (simple[^1]) way to know if the user defined a string or an attrset, so the error message would have to be somewhat generic.

[^1]: Accessing `config.*` would throw, as would `option.*.value`. We could merge `option.*definitions` ourselves, but it's not worth it just to provide a better error message!

khaneliman: This look better?

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +110,112 @@ in
             };
         ```
```
MattSturgeon, line 111: Sorry, I missed this in the initial reviews. This description is basically entirely example code with no actual text... I don't _mind_ embedding "contextual" examples like this in a description, but I prefer to also have some _actual_ description along with an _actual_ `example` value.

Don't need to make it perfect though, bare minimum I think we should have a short option description followed by a short explanation of the example usage.

MattSturgeon (changes requested):
Think this is the only remaining issue; sorry I missed it earlier!

plugins/lsp/conform-nvim.nix
khaneliman: Ah, yeah. Forgot to clean that up after I updated the example inside. I'll add a comment to the top describing the option.

MattSturgeon (approved):
Awesome! Thanks again

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +115,112 @@ in
...
+    };
+    format_on_save = # Lua
+      ''
+        function(bufnr)
+          if vim.g.disable_autoformat or vim.b[bufnr].disable_autoformat then
+            return
+          end
 
-      extraConfigLua = ''
-        require("conform").setup(${helpers.toLuaObject setupOptions})
```
khaneliman, line 129: just noticed this while testing...

plugins/lsp/conform-nvim.nix
```diff
@@ -36,97 +116,122 @@ in
...
+      };
+      shfmt = {
+        command = lib.getExe pkgs.shfmt;
+      };
+      shellharden = {
+        command = lib.getExe pkgs.shellharden;
+      };
+      squeeze_blanks = {
+        command = lib.getExe' pkgs.coreutils "cat";
+      };
```
MattSturgeon, line 234: This isn't ideal, remember we're passing the actual value as the example, not the expression text.

This will render in the docs as something like `command = /nix/store/XXXX-shellcheck/bin/shellcheck;`

If you want these examples you'll have to wrap the whole thing in `literalExpression`.

khaneliman: Ah yeah...

khaneliman: Updated to use literalExpression. I do think it provides value to show how you can utilize the property to its fullest if you want to keep these out of global path.

khaneliman:
Last push was just making it include syntax highlighting.

khaneliman:
@mergifyio queue

MattSturgeon (approved):
Using `literalMD` for an example is a little unconventional, but everything seems fine. Thanks for another high value PR!

PR #2071, opened 2024-08-23
flake-modules/list-plugins: add extra filters

khaneliman:
I have been using this more and wanted to create some filters to only output the information I cared about. This adds the ability to filter by kind, state, and whether a plugin has deprecation warnings.

### DIFF
--- a/flake-modules/dev/list-plugins.py
+++ b/flake-modules/dev/list-plugins.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
-from enum import Enum
-from dataclasses import dataclass
-from typing import Optional
 import glob
 import re
+from argparse import ArgumentParser, RawTextHelpFormatter
+from dataclasses import dataclass
+from enum import Enum
+from typing import Optional
 
 QUESTION_MARK = "❔"
 
@@ -65,7 +66,7 @@ class Plugin:
                 kind_icon = "\033[92m" + " "
             case _:
                 assert False
-        deprecation_icon: str = "⚠️" if self.dep_warnings else "  "
+        deprecation_icon: str = "⚠️ " if self.dep_warnings else "  "
 
         return (
             f"| {kind_icon}\033[0m  | {state_icon}  | {deprecation_icon} | {self.path}"
@@ -129,25 +130,52 @@ def _is_excluded(path: str) -> bool:
     return True
 
 
-if __name__ == "__main__":
-    paths: list[str] = glob.glob(
-        pathname="plugins/**/*.nix",
-        recursive=True,
-    )
-
-    filtered_paths: list[str] = list(
-        filter(
-            _is_excluded,
-            paths,
-        )
-    )
+def main(args):
+    paths: list[str] = glob.glob(pathname="plugins/**/*.nix", recursive=True)
+    filtered_paths: list[str] = list(filter(_is_excluded, paths))
     filtered_paths.sort()
 
     print("| Typ | Sty | DW | path")
     print("|-----|-----|----|--------------------------------------------------------")
     for plugin_path in filtered_paths:
-        plugin: Optional[Plugin] = parse_file(
-            path=plugin_path,
-        )
+        plugin: Optional[Plugin] = parse_file(path=plugin_path)
         if plugin is not None:
-            print(plugin)
+            if (
+                (args.kind is None or plugin.kind.name.lower() == args.kind)
+                and (args.state is None or plugin.state.name.lower() == args.state)
+                and (not args.deprecation_warnings or plugin.dep_warnings)
+            ):
+                print(plugin)
+
+
+if __name__ == "__main__":
+    parser: ArgumentParser = ArgumentParser(
+        description="""
+    Analyze Nixvim plugin files
+    Output formats a table showing:
+        If a plugin is written for Neovim or Vim.
+        If the plugin has been updated to latest style standards.
+        If a plugin contains any deprecation warnings.
+    """,
+        formatter_class=RawTextHelpFormatter,
+    )
+    parser.add_argument(
+        "-k",
+        "--kind",
+        choices=[k.name.lower() for k in Kind],
+        help="Filter plugins by kind (neovim, vim, misc, unknown)",
+    )
+    parser.add_argument(
+        "-s",
+        "--state",
+        choices=[s.name.lower() for s in State],
+        help="Filter plugins by state (new, old, unknown)",
+    )
+    parser.add_argument(
+        "-d",
+        "--deprecation-warnings",
+        action="store_true",
+        help="Show only plugins with deprecation warnings",
+    )
+
+    main(parser.parse_args())
--- a/flake-modules/dev/default.nix
+++ b/flake-modules/dev/default.nix
@@ -20,6 +20,7 @@
         projectRootFile = "flake.nix";
 
         programs = {
+          isort.enable = true;
           nixfmt = {
             enable = true;
             package = fmt;
--- a/flake-modules/dev/devshell.nix
+++ b/flake-modules/dev/devshell.nix
@@ -108,7 +108,7 @@
             }
             {
               name = "list-plugins";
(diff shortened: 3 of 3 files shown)


flake-modules/dev/list-plugins.py
```diff
@@ -130,24 +131,36 @@ def _is_excluded(path: str) -> bool:
...
+        "--state",
+        choices=[s.name.lower() for s in State],
+        help="Filter plugins by state (new, old, unknown)",
+    )
+    parser.add_argument(
+        "--deprecation-warnings",
+        action="store_true",
+        help="Show only plugins with deprecation warnings",
     )
+    args = parser.parse_args()
```
MattSturgeon, line 150: To separate the "main logic" from the arg parsing, would it make sense to have a main function that is called with the parsed args?

That way the `if __name__ == "__main__"` block is _only_ responsible for a) parsing args and b) calling `main()`.

flake-modules/dev/list-plugins.py
```diff
@@ -1,10 +1,11 @@
...
 
-from enum import Enum
-from dataclasses import dataclass
-from typing import Optional
+import argparse
 import glob
 import re
+from dataclasses import dataclass
+from enum import Enum
+from typing import Optional
```
MattSturgeon, line 8: Is re-sorting the imports an automated change? Is there any way we can add it to our treefmt configuration?

MattSturgeon (changes requested):
I'm no python expert, but this looks good overall.

Ideally, @GaetanLepage will weigh in here too, as the original author or the script.

flake-modules/dev/list-plugins.py
khaneliman: Yeah, was automated by whatever python formatter I have installed. I never work with python but looks like I have this defined in conform-nvim
```nix
        python = [
          "isort"
          "black"
        ];
```

MattSturgeon: Ok, looks like we can just add [`treefmt.programs.isort.enable`](https://flake.parts/options/treefmt-nix#opt-perSystem.treefmt.programs.isort.enable) then.

Note: our treefmt is configured with `ruff`, not `black`, so you may run into formatting conflicts and may also want to check you've not re-formatted unrelated code.

khaneliman: Sounds good, I'll do that. I'll also switch my formatter to ruff since I don't do python that much either.

flake-modules/dev/list-plugins.py
```diff
@@ -129,25 +130,44 @@ def _is_excluded(path: str) -> bool:
...
+        help="Filter plugins by state (new, old, unknown)",
+    )
+    parser.add_argument(
+        "-d",
+        "--deprecation-warnings",
+        action="store_true",
+        help="Show only plugins with deprecation warnings",
+    )
+
+    args = parser.parse_args()
```
MattSturgeon, line 172: Same. Or alternatively, perhaps inline into the `main()` invocation?

flake-modules/dev/list-plugins.py
```diff
@@ -65,7 +66,7 @@ def __str__(self) -> str:
                 kind_icon = "\033[92m" + " "
             case _:
                 assert False
-        deprecation_icon: str = "⚠️" if self.dep_warnings else "  "
+        deprecation_icon: str = " " if self.dep_warnings else "  "
```
MattSturgeon, line 69: Is this switching from a unicode icon to a nerdfont icon? My browser font can't render the new icon, although it looks fine in the terminal :slightly_smiling_face: 



Seems like an unrelated change, though I'm not against it.

MattSturgeon: Perhaps all icons ought to be defined as CONSTANTS at the top of the file, alongside QUESTION_MARK?

I think cleaning up and/or improving icons is orthogonal to the main goal of this PR, but I don't mind sneaking small related changes in through larger PRs if it reduces overall friction to get trivial changes merged.

Might be better in a dedicated commit? Although it is a small enough change it could be lumped in as one commit, maybe just adding a footnote to the commit message.

MattSturgeon (approved):
No issues, just very minor points worth considering.

Ideally @GaetanLepage will also have time to review before merging, seeing as it was originally his script.

flake-modules/dev/list-plugins.py
khaneliman: On my mac, I could see the icon when testing. On my linux desktop, for some reason the icon wouldn't show. So I grabbed the latest nerd font because it threw me off when testing on my desktop that I couldn't see the filter working. I can break off the icon and output formatting changes into a separate commit.

MattSturgeon: > On my linux desktop, for some reason the icon wouldn't show.

I haven't looked up the codepoint, but I'd guess it's a unicode emoticon, so you need an emoji font to view it.

IIRC I have [Noto Colour Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji/glyphs) installed system wide, but only have a nerdfont available when in my terminal.

>I can break off the icon and output formatting changes into a separate commit.

Up to you. I wouldn't stress too much over it. Adding the change to the commit message would also be fine, for example.

🙂

flake-modules/dev/list-plugins.py
```diff
@@ -129,25 +131,52 @@ def _is_excluded(path: str) -> bool:
...
+            if (
+                (args.kind is None or plugin.kind.name.lower() == args.kind)
+                and (args.state is None or plugin.state.name.lower() == args.state)
+                and (not args.deprecation_warnings or plugin.dep_warnings)
+            ):
+                print(plugin)
+
+
+if __name__ == "__main__":
+    parser: argparse.ArgumentParser = argparse.ArgumentParser(
```
MattSturgeon, line 153: Now that I see it, `parser: argparse.ArgumentParser = argparse.ArgumentParser` does seem a kinda redundant, I feel a little silly for suggesting it... 😕

It is probably better though. A smarter language would make this obvious using formalised constructors and type inference...

I do think `from argparse import ArgumentParser` would be better though; I think that'd get the argparse-related imports down to one line too?

Please tell me to stop commenting on the smallest unimportant details 🫠

flake-modules/dev/list-plugins.py
khaneliman: > I haven't looked up the codepoint, but I'd guess it's a unicode emoticon, so you need an emoji font to view it.
> 
> IIRC I have [Noto Colour Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji/glyphs) installed system wide, but only have a nerdfont available when in my terminal.

Supposedly, I have it installed... https://github.com/khaneliman/khanelinix/blob/c202a44aa74b4a1b2035c232dcc0f86917265f29/modules/shared/system/fonts/default.nix#L32. But, I see it under an /share/x11 folder for some reason in /run/currentsystem and can't find the linked fonts folder...

khaneliman: Yeah, I see that I can view the nerd fonts in alacritty but they aren't showing in kitty. So I just pushed the spacing fix for aligning the table columns with mixed output.

MattSturgeon: > Yeah, I see that I can view the nerd fonts in alacritty but they aren't showing in kitty.

For kitty you might find this interesting:

https://github.com/MattSturgeon/nix-config/blob/8645b1687eda188fcc1c656ce0825315b8ef663d/modules/home-manager/kitty.nix#L58-L64

?

khaneliman: Sweet, that looks like it was similar to what I found in their documentation. I'll try it out real quick.

khaneliman: Worked for me! Thanks!

GaetanLepage (approved):
Looks good, thanks :)

GaetanLepage:
@mergifyio queue

PR #2053, opened 2024-08-22
plugins/telescope: migrate helpers -> lib.nixvim

khaneliman:
Similar to https://github.com/nix-community/nixvim/pull/2052

PR #2072, opened 2024-08-24
plugins/lsp/extensions: remove `with lib` + fix `mkRaw` reference

MattSturgeon:
See https://github.com/nix-community/nixvim/pull/2053#issuecomment-2307961673
cc @andrevmatos

### DIFF
--- a/plugins/telescope/extensions/_helpers.nix
+++ b/plugins/telescope/extensions/_helpers.nix
@@ -1,5 +1,4 @@
 { lib, config, ... }:
-with lib;
 let
   inherit (lib.nixvim) mkPluginPackageOption mkSettingsOption toSnakeCase;
 in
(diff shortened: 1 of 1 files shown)


MattSturgeon:
@andrevmatos could you test this by doing

```shell
nix flake lock --override-input nixvim github:MattSturgeon/nixvim/fix_ts_undo
```

(Once you're done testing you can just revert the changes to you lockfile using git, e.g. `git restore flake.lock`)

andrevmatos (approved):
I can confirm this patch fixes it for me. Thank you for the fastest response, and your contributions to this great community project :pray:

MattSturgeon:
> Edit: my bad, it seems some helpers.mkRaw usage in telescope.extensions.undo.mappings were breaking the config; removing/adjusting them made it evaluate successfully.

No, this really is a real issue you're uncovered. It's possible that changing your config to use/not-use `mkRaw` could _avoid_ the issue on your end, but the underlying issue still exists :smile: 

> I really wish nixvim didn't have this frequent large refactorings, its usage just depends a little too much on the apis and utilities.

We try not to refactor things needlessly; moving from `helpers` -> `lib.nixvim` is a slow, long term and mostly _internal_ change with the end goal being to not have such a generically named module arg that could potentially conflict with other nix tools, or user's _own_ helper tools.

The main (unrelated) refactoring work we're doing is migrating plugins to use the `settings` option pattern, which has the clear user-facing benefit of allowing freeform configs where you can define any option even if we haven't declared the option as existing.\
This is almost always done in such a way that existing config continues to work, but just shows some harmless warnings explaining the change.

MattSturgeon:
> I can confirm this patch fixes it for me. Thank you for the fastest response, and your contributions to this great community project 🙏

Awesome, thanks for testing!

khaneliman (approved):

MattSturgeon:
@mergifyio queue

andrevmatos:
Thank you very much for your kind and detailed explanations. Yes, I see how these changes are for the best. It's just that, neovim frequently requires quite customized options, and these changes require adjusting the configs.

But this may just be my experience due to my nixvim flake config being larger/more complex than my NixOS' by now, with several bits becoming a little bit too tightly coupled with inner nixvim workings after reading its sources (the power and burden of nix) to achieve the desired config output. Specially `settings` migration will make this way easier to maintain, after its done. Thanks!

PR #2073, opened 2024-08-24
docs: fix markdown issue with tree-sitter docs

menixator:
This PR is a fix for a markdown issue caused by #2061

cc: @khaneliman


khaneliman (approved):
Thanks for catching that right away! Already tested it out and it fixes the issue.

khaneliman:
@mergifyio queue

PR #2076, opened 2024-08-24
tests: fix tests by enabling `wrapRc`

MattSturgeon:
Since cbd1003d9d3dcc371f0602a66ff32af9f8c14490 I'm able to add _some_ invalid config definitions to modules the tests are using and get no build error.

For example `extraConfigLua = null;` should produce an invalid type error, but doesn't.

One less visible change in that commit is the move away from using the "standalone" wrapper (`makeNixvimWithModule`), which implicitly sets `wrapRc = true`.

Adding back `wrapRc` to the tests seems to fix the issue, however this makes me wonder if there's an underlying issue with wrapping/not-wrapping?

Perhaps we've simply uncovered a long-standing eval issue that is masked over by using `wrapRc`?

### DIFF
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -60,6 +60,7 @@ let
               "mkTestDerivationFromNixvimModule: the `dontRun` argument is deprecated. You should use the `test.runNvim` module option instead."
               { config.test.runNvim = !dontRun; }
           ))
+          { wrapRc = true; }
         ];
         extraSpecialArgs = {
           defaultPkgs = pkgs;
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -4,6 +4,8 @@
   ...
 }:
 let
+  defaultPkgs = pkgs;
+
   # Create a nix derivation from a nixvim executable.
   # The build phase simply consists in running the provided nvim binary.
   mkTestDerivationFromNvim =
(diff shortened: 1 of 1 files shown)


MattSturgeon:
## Unmasked errors

- ERROR: lightline.vim: Could not load colorscheme gruvbox. (plugins-statuslines-lightline)

(will edit as I read through CI)

lib/tests.nix
```diff
@@ -58,6 +58,7 @@ let
               "mkTestDerivationFromNixvimModule: the `dontRun` argument is deprecated. You should use the `test.runNvim` module option instead."
               { config.test.runNvim = !dontRun; }
           ))
+          { wrapRc = true; }
```
MattSturgeon, line 61: Should this also be included in the `.extend {}` in `mkTestDerivationFromNvim` above?

We don't use that function internally, and anyone downstream _should_ already have a wrapped RC because their `nvim` will have been built by the standalone wrapper, so it's probably not strictly necessary there...

traxys: I don't really see how you would be able to call that function without having a standalone module

MattSturgeon: They could use `evalNixvim`, if if they're really fancy they could manually call `evalModules`.

More reasonably, a home-manager user could decide to use the `config.programs.nixvim.test.derivation` from their home-manager configuration.

MattSturgeon: Ah my mistake, I was talking about how they could get the test _without_ using either of these functions...

You're right, regarding my original comment. It's very unlikely they'd get a standalone nixvim derivation without using the standalone wrapper!

traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2077, opened 2024-08-24
modules/test: use `runCommand` and add `passthru`s

MattSturgeon:
It was bothering me that the test used the fully blown `stdenv.mkDerevation` despite only needing to run a few commands. Therefore I've replaced it with `runCommand`. More specifically, `runCommandNoCCLocal` which should pull in fewer unnecessary build dependencies.

Additionally, I needed to add some pass-through attrs when debugging #2076, so I've created this PR to persist those. Maybe other stuff should be passed through too?

### DIFF
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -46,54 +46,49 @@ in
     };
   };
 
-  config = {
-    test.derivation = pkgs.stdenv.mkDerivation {
-      inherit (cfg) name;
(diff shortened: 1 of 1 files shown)


traxys (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

Issue #1934, opened 2024-07-26
Example in README.md: Lightline.vim could not load colorscheme gruvbox. (While running without `sudo`)


jirafey:
| Field        | Description |
|--------------|-------------|
| Plugin       | `Lightline`  | 
| Nixpkgs      | `24.05` | 
| Home Manager | `24.05` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Minimal, Reproducible Example (MRE)

```nix
  programs.nixvim = {
    enable = true;
    defaultEditor = true;
    colorschemes.gruvbox.enable = true;
    plugins.lightline.enable = true;
  };
}
```

```nix
# home.nix
{ inputs, ... }:

{

  imports = [
    inputs.nixvim.homeManagerModules.nixvim
  ];


  home.username = "user";
  home.homeDirectory = "/home/user";
  # You do not need to change this if you're reading this in the future.
  # Don't ever change this after the first build.  Don't ask questions.
  home.stateVersion = "24.05";

  programs.nixvim = {
    enable = true;
    defaultEditor = true;
    colorschemes.gruvbox.enable = true;
    plugins.lightline.enable = true;
  };
}
```

```nix
# flake.nix
{
  
  inputs = {
    nixpkgs = {
      url = "github:NixOS/nixpkgs?ref=nixos-24.05";
    };
    home-manager = {
      url = "github:nix-community/home-manager?ref=release-24.05";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    nixvim = {
      url = "github:nix-community/nixvim?ref=nixos-24.05";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = inputs @ { self, nixpkgs, home-manager, ... }: {
    nixosConfigurations = {
      hostname = nixpkgs.lib.nixosSystem {
        system = "x86_64-linux";
        modules = [
          ./configuration.nix

	  home-manager.nixosModules.home-manager {
	    home-manager.extraSpecialArgs = { inherit inputs; };
            home-manager.useGlobalPkgs = true;
	    home-manager.useUserPackages = true;
	    home-manager.users.user = {
	      imports = [
	        # DO NOT IMPORT HOME-MANAGER MODULES HERE
	        ./home.nix
	        ./modules/firefox/firefox.nix
	        ./modules/nvim/nvim.nix
	        ./modules/bash/bash.nix
	      ];
	    };
[...]

PR #2079, opened 2024-08-25
readme: fix incorrect demo configuration

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1934 There is an incorrect example, that might have worked previously, but is no longer the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,14 +26,14 @@ lightline plugin:
   programs.nixvim = {
     enable = true;
 
-    colorschemes.gruvbox.enable = true;
-    plugins.lightline.enable = true;
+    colorschemes.catppuccin.enable = true;
+    plugins.lualine.enable = true;
   };
 }
 ```
 
-When we do this, lightline will be set up to a sensible default, and will use
-gruvbox as the colorscheme, no extra configuration required!
+When we do this, lualine will be set up to a sensible default, and will use
+catppuccin as the colorscheme, no extra configuration required!
 
 Check out [this list of real world nixvim configs](https://nix-community.github.io/nixvim/user-guide/config-examples.html)!
 


README.md
```diff
@@ -26,14 +26,14 @@ lightline plugin:
...
+    colorschemes.catppuccin.enable = true;
+    plugins.lualine.enable = true;
   };
 }
 ```
 
-When we do this, lightline will be set up to a sensible default, and will use
-gruvbox as the colorscheme, no extra configuration required!
+When we do this, lualine will be set up to a sensible default, and will use
+catppuccin as the colorscheme, no extra configuration required!
```
MattSturgeon, line 36: Is this still true? If so we should change this behaviour given lightline obviously doesn't work the way we thought it did when we initially implemented that behaviour...

If it isn't true anymore, we are fine to just amend the documentation.

MattSturgeon: Ah sorry, I'm getting mixed up between lightline and lualine. Can you still confirm that we don't auto-set colorschemes for lightline?

khaneliman: The behavior of lightline is to use their default color scheme. Regardless of what my neovim color scheme is, lightline displays the same color scheme.

khaneliman: I think we talked about it elsewhere, but it appears in their documentation that its very much intentional to not automatically utilize the edit colorscheme: https://github.com/itchyny/lightline.vim/blob/58c97bc21c6f657d3babdd4eefce7593e30e75ce/README.md?plain=1#L130

MattSturgeon: > The behavior of lightline is to use their default color scheme. Regardless of what my neovim color scheme is, lightline displays the same color scheme.

I understand that's the upstream plugin behaviour, however we used to set `plugins.lightline.colorscheme` to `config.colorscheme` by default, which lead to #1934 being reported: https://github.com/nix-community/nixvim/blob/e48ce785d9e72c0106319d93e23c5579336ffe33/plugins/statuslines/lightline.nix#L19-L24

It seems this is no longer the case, so my concern is resolved: https://github.com/nix-community/nixvim/blob/7e3ed24e52e2d400b84b22524cba987a495a3f24/plugins/statuslines/lightline.nix#L54-L60

> it appears in their documentation that its very much intentional to not automatically utilize the editor colorscheme

Yes I remember that discussion. Interestingly, digging through older discussions it seems that lightline _does_ support setting `plugins.lightline.settings.colorscheme = "gruvbox"` _if_ a different gruvbox plugin is also installed, see this thread: https://github.com/nix-community/nixvim/discussions/1575#discussioncomment-10254132

khaneliman:
@mergifyio queue

README.md
khaneliman: > > The behavior of lightline is to use their default color scheme. Regardless of what my neovim color scheme is, lightline displays the same color scheme.
> 
> I understand that's the upstream plugin behaviour, however we used to set `plugins.lightline.colorscheme` to `config.colorscheme` by default, which lead to #1934 being reported:
> 

Sorry, that makes more sense... I forgot that we did that. That was a long 2 days ago :P 

> > it appears in their documentation that its very much intentional to not automatically utilize the editor colorscheme
> 
> Yes I remember that discussion. Interestingly, digging through older discussions it seems that lightline _does_ support setting `plugins.lightline.settings.colorscheme = "gruvbox"` _if_ a different gruvbox plugin is also installed, see this thread: [#1575 (reply in thread)](https://github.com/nix-community/nixvim/discussions/1575#discussioncomment-10254132)

That is interesting to me.. I did see some talk about that but, I wasn't sure how it was supported based on what I was reading on their github.

PR #2080, opened 2024-08-25
plugins/lualine: migrate to mkNeovimPlugin

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2191

Migrating to the current standards. Required using a transitionType for the component options that will help keep existing configurations working, but recommend switching to the new way of being explicit in your configuration to match upstream plugin declaration.

### DIFF
--- a/plugins/statuslines/lualine.nix
+++ b/plugins/statuslines/lualine.nix
@@ -1,177 +1,270 @@
 {
   lib,
-  helpers,
-  config,
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.lualine;
+  inherit (lib.nixvim) defaultNullOpts mkSettingsRenamedOptionModules;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "lualine";
+  originalName = "lualine.nvim";
+  package = "lualine-nvim";
 
-  mkSeparatorsOption =
-    {
-      leftDefault ? " ",
-      rightDefault ? " ",
-    }:
-    {
-      left = helpers.defaultNullOpts.mkStr leftDefault "Left separator";
-      right = helpers.defaultNullOpts.mkStr rightDefault "Right separator";
-    };
+  maintainers = [ lib.maintainers.khaneliman ];
 
-  mkComponentOptions =
-    defaultName:
-    helpers.mkNullOrOption (
-      with types;
-      listOf (
-        either str (submodule {
-          options = {
-            name = mkOption {
-              type = types.either types.str helpers.nixvimTypes.rawLua;
-              description = "Component name or function";
-              default = defaultName;
-            };
+  # TODO: Added 2024-09-05, remove after 24.11
+  optionsRenamedToSettings = [
+    "extensions"
+    "sections"
+    "inactiveSections"
+    "tabline"
+    "winbar"
+    "inactiveWinbar"
+  ];
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "lualine"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+      optionsPath = settingsPath ++ [ "options" ];
+      oldOptions = [
+        "theme"
+        "globalstatus"
+        "refresh"
+        "iconsEnabled"
+        "sectionSeparators"
+        "componentSeparators"
+        "disabledFiletypes"
+        "ignoreFocus"
+        "alwaysDivideMiddle"
+      ];
+    in
+    mkSettingsRenamedOptionModules basePluginPath optionsPath oldOptions;
 
-            icons_enabled = helpers.defaultNullOpts.mkBool true ''
-              Enables the display of icons alongside the component.
-            '';
+  settingsOptions =
+    let
+      mkSeparatorsOption =
+        { left, right }:
+        defaultNullOpts.mkNullableWithRaw' {
+          description = ''
+            Filetypes in which to disable lualine.
+            Allows you to specify filetypes that you want to only disable on specific components.
+          '';
+          pluginDefault = {
+            inherit left right;
+          };
+          type =
+            with types;
+            either str (submodule {
+              freeformType = attrsOf anything;
 
-            icon = helpers.mkNullOrOption (
-              with types;
-              either str (submodule {
+              options = {
+                left = defaultNullOpts.mkStr left "Left separator";
+                right = defaultNullOpts.mkStr right "Right separator";
+              };
+            });
+        };
+
+      # NOTE: This option is used for the shared component section definitions.
+      # We used to transform several options for the user to handle unkeying inside an attribute set and
+      # merging in undefined options into the final option. Now that we have freeformType support the user can
+      # manage this configuration exactly as the plugin expects without us transforming the values for them.
+      mkComponentOptions =
+        description:
+        lib.nixvim.mkNullOrOption' {
+          type =
+            with types;
+            let
+              # TODO: added 2024-09-05 remove after 24.11
+              oldAttrs = [
+                [ "name" ]
+                [
+                  "icon"
+                  "icon"
+                ]
+                [ "extraConfig" ]
+              ];
+              isOldType = x: lib.any (loc: lib.hasAttrByPath loc x) oldAttrs;
+              oldType = addCheck (attrsOf anything) isOldType // {
+                description = "attribute set containing ${lib.concatMapStringsSep ", " lib.showOption oldAttrs}";
+              };
+              coerceFn =
+                attrs:
+                lib.pipe attrs [
+                  # Transform old `name` attr to `__unkeyed`
+                  (
+                    x:
+                    if x ? name then
+                      lib.removeAttrs x [ "name" ]
+                      // {
+                        __unkeyed-1 = x.name;
+                      }
+                    else
+                      x
+                  )
+                  # Transform old `icon.icon` attr to `__unkeyed`
+                  (
+                    x:
+                    if x.icon or null ? icon then
+                      x
+                      // {
+                        icon = removeAttrs x.icon [ "icon" ] // {
+                          __unkeyed-1 = x.icon.icon;
+                        };
+                      }
+                    else
+                      x
+                  )
+                  # Merge in old `extraConfig` attr
+                  (x: removeAttrs x [ "extraConfig" ] // x.extraConfig or { })
+                ];
+              newType = submodule {
                 freeformType = attrsOf anything;
 
                 options = {
-                  icon = mkOption {
-                    type = str;
-                    description = "Icon character.";
+                  icons_enabled = defaultNullOpts.mkBool true ''
+                    Whether to display icons alongside the component.
+                  '';
+
+                  icon = lib.nixvim.mkNullOrOption' {
+                    type = either str (attrsOf anything);
+                    description = "The icon to be displayed in the component.";
+                  };
+
+                  separator = mkSeparatorsOption {
+                    left = " ";
+                    right = " ";
+                  };
+
+                  color = defaultNullOpts.mkNullable' {
+                    type = either attrs str;
+                    pluginDefault = lib.literalMD "The color defined by your theme, for the respective section & mode.";
+                    description = "Defines a custom color for the component.";
+                  };
+
+                  padding = defaultNullOpts.mkNullable (oneOf [
+                    int
+                    (submodule {
+                      # In case they add support for top/bottom padding
+                      freeformType = attrsOf (maybeRaw int);
+
+                      options = {
+                        left = defaultNullOpts.mkInt null "Left padding.";
+                        right = defaultNullOpts.mkInt null "Right padding.";
+                      };
+                    })
+                    rawLua
+                  ]) 1 "Amount of padding added to components.";
+
+                  fmt = lib.nixvim.mkNullOrLuaFn' {
+                    description = ''
+                      A lua function to format the component string.
+                    '';
+                    example = ''
+                      function(text)
+                        return text .. "!!!"
+                      end
+                    '';
                   };
                 };
-              })
-            ) "Defines the icon to be displayed in front of the component.";
+              };
+            in
+            maybeRaw (listOf (either str (lib.nixvim.transitionType oldType coerceFn (maybeRaw newType))));
+          inherit description;
+        };
 
-            separator = mkSeparatorsOption { };
-
-            color = helpers.mkNullOrOption (types.attrsOf types.str) ''
-              Defines a custom color for the component.
-            '';
-
-            padding = helpers.defaultNullOpts.mkNullable (types.either types.int (
-              types.submodule {
-                options = {
-                  left = mkOption {
-                    type = types.int;
-                    description = "left padding";
-                  };
-                  right = mkOption {
-                    type = types.int;
-                    description = "left padding";
-                  };
-                };
-              }
-            )) 1 "Adds padding to the left and right of components.";
-
-            fmt = helpers.mkNullOrLuaFn ''
-              A lua function to format the component string.
-
-              Example:
-              ```lua
-                function(text)
-                  return text .. "!!!"
-                end
-              ```
-            '';
-
-            extraConfig = mkOption {
-              type = types.attrs;
-              default = { };
-              description = "extra options for the component";
-            };
-          };
-        })
-      )
-    ) "";
-
-  mkEmptySectionOption = name: {
-    lualine_a = mkComponentOptions "";
-    lualine_b = mkComponentOptions "";
-    lualine_c = mkComponentOptions "";
-    lualine_x = mkComponentOptions "";
-    lualine_y = mkComponentOptions "";
-    lualine_z = mkComponentOptions "";
-  };
-in
-{
-  options = {
-    plugins.lualine = {
-      enable = mkEnableOption "lualine";
-
-      package = lib.mkPackageOption pkgs "lualine" {
-        default = [
-          "vimPlugins"
-          "lualine-nvim"
-        ];
+      mkEmptySectionOption = name: {
+        lualine_a = mkComponentOptions "Left section on left side.";
+        lualine_b = mkComponentOptions "Middle section on left side.";
+        lualine_c = mkComponentOptions "Right section on left side.";
+        lualine_x = mkComponentOptions "Left section on right side.";
+        lualine_y = mkComponentOptions "Middle section on right side.";
+        lualine_z = mkComponentOptions "Right section on right side.";
       };
 
-      gitPackage = lib.mkPackageOption pkgs "git" {
-        nullable = true;
-      };
+    in
+    {
+      options = {
+        icons_enabled = defaultNullOpts.mkBool true ''
+          Whether to enable icons for all components.
 
-      iconsEnabled = mkOption {
-        type = types.bool;
-        description = "Whether to enable/disable icons for all components.";
-        default = true;
-      };
-
-      theme = helpers.defaultNullOpts.mkNullable (
-        with types; either str attrs
-      ) "auto" "The theme to use for lualine-nvim.";
-
-      componentSeparators = mkSeparatorsOption {
-        leftDefault = "";
-        rightDefault = "";
-      };
-
-      sectionSeparators = mkSeparatorsOption {
-        leftDefault = "";
-        rightDefault = "";
-      };
-
-      disabledFiletypes = {
-        statusline = helpers.defaultNullOpts.mkListOf types.str [ ] ''
-          Only ignores the ft for statusline.
+          This option is also available on individual components to control whether they should display icons.
         '';
 
-        winbar = helpers.defaultNullOpts.mkListOf types.str [ ] ''
-          Only ignores the ft for winbar.
+        theme = defaultNullOpts.mkNullable (
+          with types; either str attrs
+        ) "auto" "The theme to use for lualine-nvim.";
+
+        component_separators = mkSeparatorsOption {
+          left = "";
+          right = "";
+        };
+
+        section_separators = mkSeparatorsOption {
+          left = "";
+          right = "";
+        };
+
+        disabled_filetypes = defaultNullOpts.mkNullableWithRaw' {
+          description = ''
+            Filetypes in which to disable lualine.
+            Allows you to specify filetypes that you want to only disable on specific components.
+          '';
+          pluginDefault = { };
+          type =
+            with types;
+            either (listOf (maybeRaw str)) (submodule {
+              freeformType = attrsOf anything;
+
+              options = {
+                statusline = defaultNullOpts.mkListOf str [ ] ''
+                  Hide the statusline component on specified filetypes.
+                '';
+
+                winbar = defaultNullOpts.mkListOf str [ ] ''
+                  Hide the winbar component on specified filetypes.
+                '';
+              };
+            });
+        };
+
+        ignore_focus = defaultNullOpts.mkNullableWithRaw' {
+          type = types.listOf types.str;
+          pluginDefault = [ ];
+          description = ''
+            A list of filetypes that should always show an "unfocused" statusline.
+
+            If the focused window's filetype is in this list, then the most
+            recently focused window will be drawn as the active statusline.
+          '';
+          example = [
+            "neo-tree"
+            "nvim-tree"
+            "mini-files"
+          ];
+        };
+
+        always_divide_middle = defaultNullOpts.mkBool true ''
+          Whether to prevent left sections i.e. 'a','b' and 'c' from taking over the entire statusline
+          even if neither of 'x', 'y' or 'z' are present.
         '';
-      };
 
-      ignoreFocus = helpers.defaultNullOpts.mkListOf types.str [ ] ''
-        If current filetype is in this list it'll always be drawn as inactive statusline and the
-        last window will be drawn as active statusline.
+        globalstatus = defaultNullOpts.mkBool false ''
+          Whether to enable "global" statusline.
+          I.e. having a single statusline at bottom of neovim, instead of one for each individual window.
+        '';
 
-        For example if you don't want statusline of your file tree / sidebar window to have active
-        statusline you can add their filetypes here.
-      '';
+        refresh = {
+          statusline = defaultNullOpts.mkInt 1000 "Refresh time for the status line (ms)";
 
-      alwaysDivideMiddle = helpers.defaultNullOpts.mkBool true ''
-        When set to true, left sections i.e. 'a','b' and 'c' can't take over the entire statusline
-        even if neither of 'x', 'y' or 'z' are present.
-      '';
+          tabline = defaultNullOpts.mkInt 1000 "Refresh time for the tabline (ms)";
 
-      globalstatus = helpers.defaultNullOpts.mkBool false ''
-        Enable global statusline (have a single statusline at bottom of neovim instead of one for
-        every window).
-        This feature is only available in neovim 0.7 and higher.
-      '';
-
-      refresh = {
-        statusline = helpers.defaultNullOpts.mkInt 1000 "Refresh time for the status line (ms)";
-
-        tabline = helpers.defaultNullOpts.mkInt 1000 "Refresh time for the tabline (ms)";
-
-        winbar = helpers.defaultNullOpts.mkInt 1000 "Refresh time for the winbar (ms)";
+          winbar = defaultNullOpts.mkInt 1000 "Refresh time for the winbar (ms)";
+        };
       };
 
       sections = {
@@ -184,7 +277,7 @@ in
         lualine_z = mkComponentOptions "location";
       };
 
-      inactiveSections = {
+      inactive_sections = {
         lualine_a = mkComponentOptions "";
         lualine_b = mkComponentOptions "";
         lualine_c = mkComponentOptions "filename";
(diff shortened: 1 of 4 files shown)


khaneliman:
Created https://github.com/NixOS/nixpkgs/pull/337579 to free up the extraPlugins logic since it's listed as a dependency of the plugin.

MattSturgeon:
> Created [NixOS/nixpkgs#337579](https://github.com/NixOS/nixpkgs/pull/337579) to free up the extraPlugins logic since it's listed as a dependency of the plugin.

Based on https://github.com/NixOS/nixpkgs/pull/338164, I guess we'll need to tackle this the same as #2108?

plugins/statuslines/lualine.nix
```diff
@@ -1,256 +1,401 @@
...
             };
-          };
-        })
-      )
-    ) "";
+          })
+        );
+      # TODO: best way to do this?
+      # Was setting up apply to handle the old way of transforming a config to the setup call
+      apply =
```
khaneliman, line 102: @MattSturgeon Tried handling the deprecation for these nested options and not sure if it's required to be done this way or if you have a simpler way of handling this.. I'm also struggling to find a simple way to handle the deprecation messages unless i can just inline the warn or something here when we process the transformation ?

MattSturgeon: I think it'd be easier to keep the old option and new option separate in this case. I.e. no rename module; instead just remove the `default` and copy over the value _if_ it is defined?

Perhaps the new option could also just be `attrsOf anything`, instead of being a freeform submodule type?

khaneliman: Well, the copying over the defined value is what this apply is for. It'll do the transformation we used to do if they defined it using the old options. Unless, you meant there's an even easier way of copying those values over and transforming them without the apply.

We did a little bit of magic in this submodule with the `name`, `icon`, and `extraConfig` options to be flattened and unkeyed. I wanted to just show a message when I detected any of those usages. The hard part, for me, was that it's basically a submodule spread across a bunch of different options that i have to search for a proper deprecation warning on specific usages. But, if I could do the warning on the generic top level easily with a migration message about the new magic-less way of doing things that would be simpler..

khaneliman:
@MattSturgeon I think I was overcomplicating the warnings... I can just inline the warnings in the apply and then we'll see the warning on the actual usage without having to parse again later.

plugins/statuslines/lualine.nix
MattSturgeon: I meant that we need the apply here because we have a rename alias pointing to this option, right?

So by removing the rename alias, we could do the same thing the apply does but using a config definition instead.

MattSturgeon: I'm thinking we can probably do this with `transitionType`, but note that coercedTo and transitionType don't support submodule-types on the lhs, so we'd have to use  something like:

```nix
let
  oldAttrs = [ "name" "icon" "extraConfig" ];
  isOldType = def: any (name: elem name oldAttrs) (attrNames def);
  oldType = (addCheck (attrsOf anything) isOldType) // {
    description = "attribute set containing ${concatStringsSep ", " oldAttrs}";
  };
  coerceFn = /* Essentially, do what you're currently doing in `apply` */;
in
listOf (
  transitionType
  (addCheck (attrsOf anything) isOldType)
  coerceFn
  (submodule {/* submodule options */})
```

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
+                      optionWithNameTransformed.icon // { "__unkeyed" = optionWithNameTransformed.icon.icon; }
+                    ) [ "icon" ]
+                  else
+                    optionWithNameTransformed;
+
+                # NOTE: Handles the deprecated extraConfig option
+                # Old freeform option was just merged into the existing config
+                flattenedExtraConfig = optionWithIconTransformed.extraConfig or { };
+                optionWithoutExtraConfig = lib.removeAttrs optionWithIconTransformed [ "extraConfig" ];
+                finalTransformedValue = optionWithoutExtraConfig // flattenedExtraConfig;
```
MattSturgeon, line 67: Let's pipe this:

First, replace `optionsWithXYZTransformed` with various coercion functions, `transformName`, `transformIcon`, etc.

Then we can `pipe` the options attr through those transformers:

```nix
coerceFn = attrs: lib.pipe attrs [
  # Transform old `name` attr to `__unkeyed`
  (
    x:
      if x ? name then
        lib.removeAttrs option [ "name" ] // {
          __unkeyed-1 = option.name;
        }
      else
       x
  )
  # Transform old `icon.icon` attr to `__unkeyed`
  # FIXME: this conflicts with `name`, is that intended?
  (
    x:
      if x.icon or null ? icon then
        removeAttrs x [ "icon" ] // {
          __unkeyed-1 = x.icon.icon;
        }
      else
        x
  )
  # Merge in old `extraConfig` attr
  (x: removeAttrs x [ "extraConfig" ] // x.extraConfig or { })
];
```

TBH, I don't think it's less readable with the transformers as inline anonymous lambdas, as above...

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
+
+              color = defaultNullOpts.mkNullable (either attrs str) null ''
+                Defines a custom color for the component.
+
+                Default is the color defined by your theme for that section and mode.
+              '';
+
+              padding = defaultNullOpts.mkNullable (either int (submodule {
+                # In case they add support for top/bottom padding
                 freeformType = attrsOf anything;
```
MattSturgeon, line 95: I wouldn't normally suggest this, but I think here we could be confident of:
```suggestion
                freeformType = attrsOf int;
```

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
+                description = ''
+                  A lua function to format the component string.
+                '';
+                example = ''
+                  ```lua
+                    function(text)
+                      return text .. "!!!"
+                    end
+                  ```
+                '';
```
MattSturgeon, line 119:
```suggestion
                example = ''
                  function(text)
                    return text .. "!!!"
                  end
                '';
```

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
+              };
             };
           };
-        })
-      )
-    ) "";
+        in
+        listOf (lib.nixvim.transitionType oldType coerceFn newType);
```
MattSturgeon, line 124: Note: this won't have rawLua support. IDK if having that is worth the required complexity though...

```nix
  maybeRaw (listOf ((lib.nixvim.transitionType oldType coerceFn (maybeRaw newType))));
```

Technically, we can get away without `maybeRaw newType` because `{ __raw = ""; }` will match both `rawLua.check` and `(attrsOf anything).check`.

Having the list itself be `maybeRaw` could also be achieved using `mkNullableWithRaw` or `mkListOf`, however I suspect we will run into #1618 if we haven't already...

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
+    in
+    mkSettingsRenamedOptionModules basePluginPath optionsPath oldOptions;
 
-      iconsEnabled = mkOption {
-        type = types.bool;
-        description = "Whether to enable/disable icons for all components.";
-        default = true;
-      };
+  settingsOptions = {
+    options = {
```
MattSturgeon, line 182: Typo: 
```suggestion
  settingsOptions = {
```

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.lualine;
+  inherit (lib.nixvim) defaultNullOpts mkSettingsRenamedOptionModules;
+  inherit (lib) types mkOption;
 
+  # TODO: do we really need this
```
MattSturgeon, line 10: It is used in three options. I don't see any harm it in, unless you think the options themselves don't need to be declared and instead we'll use the freeform type for their definitions?

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
   mkSeparatorsOption =
     {
       leftDefault ? " ",
       rightDefault ? " ",
```
MattSturgeon, line 14:
```suggestion
      leftDefault,
      rightDefault,
```

These defaults are only used once, let's just be explicit in all usage?

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
     }:
     {
-      left = helpers.defaultNullOpts.mkStr leftDefault "Left separator";
-      right = helpers.defaultNullOpts.mkStr rightDefault "Right separator";
+      left = defaultNullOpts.mkStr leftDefault "Left separator";
+      right = defaultNullOpts.mkStr rightDefault "Right separator";
     };
 
   mkComponentOptions =
```
MattSturgeon, line 21: Let's add a comment describing the deprecation, when it was added and what it does.

plugins/statuslines/lualine.nix
```diff
@@ -1,260 +1,281 @@
...
-    lualine_y = mkComponentOptions "";
-    lualine_z = mkComponentOptions "";
+    lualine_a = mkComponentOptions "Left section on left side.";
+    lualine_b = mkComponentOptions "Middle section on left side.";
+    lualine_c = mkComponentOptions "Right section on left side.";
+    lualine_x = mkComponentOptions "Left section on right side.";
+    lualine_y = mkComponentOptions "Middle section on right side.";
+    lualine_z = mkComponentOptions "Right section on right side.";
   };
 in
```
MattSturgeon, line 136: Should we move `mkSeparatorsOption`, `mkComponentOptions`, and `mkEmptySectionOption` to a let-block local to `settingsOptions`?

Might make the top section of the file a little more readable?

plugins/statuslines/lualine.nix
MattSturgeon: Ah, my mistake: https://github.com/nix-community/nixvim/blob/7409e80bd2dbbc5d653b27170d2fb33082150df3/plugins/statuslines/lualine.nix#L236-L237

khaneliman: That's not a typo, the plugin has a table of options.

khaneliman: https://github.com/nvim-lualine/lualine.nvim/blob/b431d228b7bbcdaea818bdc3e25b8cdbe861f056/lua/lualine/config.lua#L10

plugins/statuslines/lualine.nix
khaneliman: >   # Transform old `icon.icon` attr to `__unkeyed`
>   # FIXME: this conflicts with `name`, is that intended?
>   (
>     x:
>       if x.icon or null ? icon then
>         removeAttrs x [ "icon" ] // {
>           __unkeyed-1 = x.icon.icon;
>         }
>       else
>         x
>   )

What do you mean about the conflict? There is a transformation we did for nested icon attributes that behaved similar to the name attribute. Basically, we had a icon = string or a icon attrset with an icon = string gets changed to unkeyed.

MattSturgeon:
Current CI failures:

Need to update usage in a test:
(We should really fix the `<unknown-files>` issue with the test-cases)
```
40> Unexpected warnings:
40> - The option `plugins.lualine.componentSeparators' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.component_separators'.
40> - The option `plugins.lualine.sectionSeparators' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.section_separators'.
40> - The option `plugins.lualine.theme' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.theme'.
```

Need to update usage in `pluigins/colorschemes/base16`:
```
> Unexpected warnings:
> - The option `plugins.lualine.theme' defined in `/nix/store/XXX-source/plugins/colorschemes/base16' has been renamed to `plugins.lualine.settings.options.theme'.
```

plugins/statuslines/lualine.nix
MattSturgeon: I'm not too familiar, but just reading the transformers above, both `icon.icon` _and_` `name` are converted into `__unkeyed-1`.

Currently, if both `name` and `icon.icon` are defined, `icon.icon` will win.

khaneliman:
> Current CI failures:
> 
> Need to update usage in a test: (We should really fix the `<unknown-files>` issue with the test-cases)
> 
> ```
> > Unexpected warnings:
> > - The option `plugins.lualine.componentSeparators' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.component_separators'.
> > - The option `plugins.lualine.sectionSeparators' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.section_separators'.
> > - The option `plugins.lualine.theme' defined in `<unknown-file>' has been renamed to `plugins.lualine.settings.options.theme'.
> ```
> 
> Need to update usage in `pluigins/colorschemes/base16`:
> 
> ```
> > Unexpected warnings:
> > - The option `plugins.lualine.theme' defined in `/nix/store/XXX-source/plugins/colorschemes/base16' has been renamed to `plugins.lualine.settings.options.theme'.
> ```

Yeah, I always leave the tests until the implementation is done so I can see that the old configs aren't broken.

plugins/statuslines/lualine.nix
khaneliman: There *shouldn't*  have been a usage for name within icon, before... so I'm not too worried about breaking that config.

MattSturgeon: Ok, it seems the icon transformer should be:

```nix
(
  x:
    if isAttrs (x.icon or null) && x.icon ? icon then
      x // { icon = removeAttrs x.icon [ "icon" ] // {
        __unkeyed-1 = x.icon.icon;
      }
    else
      x
)
```

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,241 @@
...
+          type =
+            with types;
+            let
+              # TODO: added 2024-09-05 remove after 24.11
+              oldAttrs = [
+                "name"
+                "icon"
+                "extraConfig"
+              ];
+              isOldType = def: lib.any (name: lib.elem name oldAttrs) (lib.attrNames def);
```
MattSturgeon, line 82:
```suggestion
              isOldType = x: lib.any (path: lib.hasAttrByPath path x) oldAttrs;
```

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,241 @@
...
+              oldType = addCheck (attrsOf anything) isOldType // {
+                description = "attribute set containing ${lib.concatStringsSep ", " oldAttrs}";
```
MattSturgeon, line 84:
```suggestion
                description = "attribute set containing ${lib.concatMapStringsSep ", " lib.showOption oldAttrs}";
```

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,241 @@
...
+                      // {
+                        __unkeyed-1 = x.name;
+                      }
+                    else
+                      x
+                  )
+                  # Transform old `icon.icon` attr to `__unkeyed`
+                  (
+                    x:
+                    if lib.isAttrs (x.icon or null) && x.icon ? icon then
```
MattSturgeon, line 103:
```suggestion
                    if x.icon or null ? icon then
```

plugins/statuslines/lualine.nix
MattSturgeon: And make `oldAttrs` into a list or lists

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "lualine";
+  originalName = "lualine-nvim";
+
+  package = lib.mkPackageOption pkgs "lualine" {
+    default = [
+      "vimPlugins"
+      "lualine-nvim"
+    ];
+  };
```
MattSturgeon, line 19:
```suggestion
  package = "lualine-nvim";
```

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
+    let
+      mkSeparatorsOption =
+        {
+          leftDefault,
+          rightDefault,
+        }:
+        {
+          left = defaultNullOpts.mkStr leftDefault "Left separator";
+          right = defaultNullOpts.mkStr rightDefault "Right separator";
+        };
```
MattSturgeon, line 64: Nit: I'd probably do:

```nix
mkSeparatorsOption = { left, right }: {
  left = defaultNullOpts.mkStr left "Left separator";
  right = defaultNullOpts.mkStr right "Right separator";
};
```

But it isn't important.

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
+                [ "name" ]
+                [
+                  "icon"
+                  "icon"
+                ]
+                [ "extraConfig" ]
+              ];
+              isOldType = x: lib.any (loc: lib.hasAttrByPath loc x) oldAttrs;
+              oldType = addCheck (attrsOf anything) isOldType // {
+                description = "attribute set containing ${lib.concatMapStringsSep ", " lib.showOption oldAttrs}";
```
MattSturgeon, line 87: Me wishing nixpkgs lib had an `oxfordComma` function...

plugins/statuslines/lualine.nix
```diff
@@ -197,7 +264,7 @@ in
 
       winbar = mkEmptySectionOption "Winbar configuration";
 
-      inactiveWinbar = mkEmptySectionOption "Inactive Winbar configuration";
+      inactive_winbar = mkEmptySectionOption "Inactive Winbar configuration";
```
MattSturgeon, line 267:
```suggestion
      inactive_winbar = mkEmptySectionOption "Winbar configuration used when inactive.";
```

plugins/statuslines/lualine.nix
```diff
@@ -197,7 +264,7 @@ in
 
       winbar = mkEmptySectionOption "Winbar configuration";
```
MattSturgeon, line 265:
```suggestion
      winbar = mkEmptySectionOption "Winbar configuration used when active.";
```

plugins/statuslines/lualine.nix
```diff
@@ -206,55 +273,14 @@ in
         description = "list of enabled extensions";
```
MattSturgeon, line 273:
```suggestion
        description = "List of enabled extensions.";
```

plugins/statuslines/lualine.nix
```diff
@@ -197,7 +264,7 @@ in
 
       winbar = mkEmptySectionOption "Winbar configuration";
 
-      inactiveWinbar = mkEmptySectionOption "Inactive Winbar configuration";
+      inactive_winbar = mkEmptySectionOption "Inactive Winbar configuration";
 
       extensions = mkOption {
         type = with lib.types; nullOr (listOf (either str (attrsOf anything)));
```
MattSturgeon, line 270: Why not use `mkListOf` to get null and rawLua support?

MattSturgeon (changes requested):
I've proof-read the option descriptions:

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
+          rightDefault = "";
+        };
 
-      disabledFiletypes = {
-        statusline = helpers.defaultNullOpts.mkListOf types.str [ ] ''
-          Only ignores the ft for statusline.
-        '';
+        disabled_filetypes = {
+          statusline = defaultNullOpts.mkListOf types.str [ ] ''
+            Only ignores the ft for statusline.
```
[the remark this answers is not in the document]
khaneliman: You can disable based on filetype:
- Entire lualine
- Statusline component
- Winbar component
- tabline ?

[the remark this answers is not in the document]
khaneliman: Looking at the docs we might want to make it more obvious you can use the parent level option? https://github.com/nvim-lualine/lualine.nvim?tab=readme-ov-file#disabling-lualine

plugins/statuslines/lualine.nix
khaneliman: I dont think that's necessarily true. It would be used for active/inactive when inactive isn't defined.

plugins/statuslines/lualine.nix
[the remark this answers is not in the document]
MattSturgeon: I only see `options.disabled_filetypes` passed to `setup()` and other options passed to `hide()`?

plugins/statuslines/lualine.nix
MattSturgeon: Ok. I guess either keep it simple and just add the missing `.`, or maybe attempt to explain how it interacts with `options.inactive_winbar`?

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
-        last window will be drawn as active statusline.
+        ignore_focus = defaultNullOpts.mkListOf types.str [ ] ''
+          If current filetype is in this list it'll always be drawn as inactive statusline and the
+          last window will be drawn as active statusline.
 
-        For example if you don't want statusline of your file tree / sidebar window to have active
-        statusline you can add their filetypes here.
-      '';
+          For example if you don't want statusline of your file tree / sidebar window to have active
+          statusline you can add their filetypes here.
```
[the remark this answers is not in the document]
khaneliman: Realized I haven't done the settingsExample, yet. Going to work on that.

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,241 @@
...
 
-            padding = helpers.defaultNullOpts.mkNullable (types.either types.int (
-              types.submodule {
-                options = {
-                  left = mkOption {
-                    type = types.int;
-                    description = "left padding";
+                  color = defaultNullOpts.mkNullable' {
+                    type = either attrs str;
+                    pluginDefault = literalMD "The color defined by your theme, for the respective section & mode.";
```
MattSturgeon, line 130:
```suggestion
                    pluginDefault = lib.literalMD "The color defined by your theme, for the respective section & mode.";
```

plugins/statuslines/lualine.nix
khaneliman: This is what I had in my config to disable parts of lualine depending on where i'm at. 
```nix
    disabledFiletypes = {
      statusline = [
        # "neo-tree"
        "startify"
      ];
      winbar = [
        "aerial"
        "dap-repl"
        "neo-tree"
        "neotest-summary"
        "startify"
      ];
    };
```

plugins/statuslines/lualine.nix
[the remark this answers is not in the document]
khaneliman: This is what I had in my config to disable parts of lualine depending on where i'm at. 
```nix
    disabledFiletypes = {
      statusline = [
        # "neo-tree"
        "startify"
      ];
      winbar = [
        "aerial"
        "dap-repl"
        "neo-tree"
        "neotest-summary"
        "startify"
      ];
    };
```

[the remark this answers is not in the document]
MattSturgeon: Ok, so under "advanced options", is documented:

```lua
  disabled_filetypes = {     -- Filetypes to disable lualine for.
      statusline = {},       -- only ignores the ft for statusline.
      winbar = {},           -- only ignores the ft for winbar.
  },
```

Can you pass `disabled_filetypes = { "some_ft" }` in addition to the more specific `disabled_filetypes.winbar = { "some_ft" }`?

If so, then we'll have an issue merging list-type definitions at the top `disabled_filetypes` level together with attr-type definitions `winbar =` and `statusline =`... Is this what you meant by parent-level option?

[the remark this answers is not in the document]
khaneliman: I will have to test it out, but it would seem to make sense that you could.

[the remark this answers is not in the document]
khaneliman: I would expect you can do this:
```nix
          disabled_filetypes = {
            __unkeyed = "startify";
            winbar = [
              "aerial"
              "dap-repl"
              "neo-tree"
              "neotest-summary"
            ];
          };
```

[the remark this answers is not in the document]
MattSturgeon: Yes, the freeformType should support that. It may just be a bit unintuitive for users.

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,252 @@
...
-        statusline = helpers.defaultNullOpts.mkListOf types.str [ ] ''
-          Only ignores the ft for statusline.
-        '';
+        disabled_filetypes = defaultNullOpts.mkNullableWithRaw' {
+          description = ''
+            Filetypes in which to disable lualine.
+            Allows you to specify filetypes that you want to only disable on specific components.
+          '';
+          pluginDefault = { };
+          type = types.submodule {
```
MattSturgeon, line 207: Awesome. Should we consider allowing list definitions? Either just a plain `either (listOf str) (submodule)` or a coerceTo using `lib.nixvim.mkUnkeyedAttrs`?

Both approaches have issues, for example using coercion here would not merge multiple list defs well, since they'd each end up with conflicting indexes.

MattSturgeon: A propper `types.luaTable` would help, but that's _way_ out of scope for the PR :grin:

khaneliman: I like the idea of allowing the list. Would allow it to be easier to use when you don't care about granular control.

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,244 @@
...
-      iconsEnabled = mkOption {
-        type = types.bool;
-        description = "Whether to enable/disable icons for all components.";
-        default = true;
-      };
+    in
+    {
+      options = {
+        icons_enabled = defaultNullOpts.mkBool true ''
+          Whether to enable/disable icons for all components.
```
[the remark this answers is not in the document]
khaneliman: Looks like according to their docs you can set it to false to disable all icons for all components. Not sure if there's a priority for child components > parent option, though.

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,278 @@
...
+            If the focused window's filetype is in this list, then the most
+            recently focused window will be drawn as the active statusline.
+          '';
+          example = ''
+            [
+              "neo-tree"
+              "nvim-tree"
+              "mini-files"
+            ]
+          '';
```
MattSturgeon, line 256:
```suggestion
          example =[
            "neo-tree"
            "nvim-tree"
            "mini-files"
          ];
```
:wink:

khaneliman: Lol sorry, was thinking it had to be a string for some reason.

MattSturgeon: It can be, if you use `literalExpression` or `literalMD`

plugins/statuslines/lualine.nix
```diff
@@ -1,177 +1,270 @@
...
-        last window will be drawn as active statusline.
+        disabled_filetypes = defaultNullOpts.mkNullableWithRaw' {
+          description = ''
+            Filetypes in which to disable lualine.
+            Allows you to specify filetypes that you want to only disable on specific components.
+          '';
+          pluginDefault = { };
+          type =
+            with types;
+            either (listOf str) (submodule {
```
MattSturgeon, line 220:
```suggestion
            either (listOf (maybeRaw str)) (submodule {
```

Frustrating having to be this verbose...

MattSturgeon (approved):
Everything looks good, IDK if you want a second opinion, but you've been very thorough in checking upstream options so I trust your changes :rocket:

khaneliman:
> Everything looks good, IDK if you want a second opinion, but you've been very thorough in checking upstream options so I trust your changes 🚀

Thanks, I'm fairly confident in the changes since it works with my config and I'd consider it robust. Was just double checking the diff a bit and everything looks pretty good.

khaneliman:
@mergifyio queue

SuperSandro2000:
I don't think the mkRenamedOptions is working correctly here 

```
       error: A definition for option `programs.nixvim.plugins.lualine.settings.sections.lualine_b' is not of type `null or (list of (string or ((attribute set of anything) or raw lua code))) or raw lua code'. Definition values:
       - In `/nix/store/y93lbhv6j469navzizc944gvgc5hcn4v-source/plugins/by-name/lualine':
```

MattSturgeon:
> I don't think the mkRenamedOptions is working correctly here
> 
> ```
>        error: A definition for option `programs.nixvim.plugins.lualine.settings.sections.lualine_b' is not of type `null or (list of (string or ((attribute set of anything) or raw lua code))) or raw lua code'. Definition values:
>        - In `/nix/store/y93lbhv6j469navzizc944gvgc5hcn4v-source/plugins/by-name/lualine':
> ```

Thanks for reporting. What does the offending definition look like?

MattSturgeon:
This doesn't make much sense, because the new type should be a superset of the old type.

```
# new type:
null or (list of (string or ((attribute set of anything) or raw lua code))) or raw lua code

# old type:
null or (list of (string or (submodule)))
```

SuperSandro2000:
> Thanks for reporting. What does the offending definition look like?

That is the old config https://github.com/NuschtOS/nvim.nix/blob/4aae65ffea0b488803245fd27e6ef5812faa5fa7/modules/default.nix#L79-L92

My attempt at updating is at https://github.com/NuschtOS/nvim.nix/pull/47/commits/d11ad57b7a4b8236150d15f0c83662a21f84bebd

khaneliman:
> > Thanks for reporting. What does the offending definition look like?
> 
> That is the old config https://github.com/NuschtOS/nvim.nix/blob/4aae65ffea0b488803245fd27e6ef5812faa5fa7/modules/default.nix#L79-L92
> 
> My attempt at updating is at [NuschtOS/nvim.nix@`d11ad57` (#47)](https://github.com/NuschtOS/nvim.nix/pull/47/commits/d11ad57b7a4b8236150d15f0c83662a21f84bebd)

I created test cases for both configs and they worked: https://github.com/nix-community/nixvim/compare/main...khaneliman:nixvim:sandro?expand=1

image

Running the old config:
image

Testing and running the new config worked without the warnings.

PR #2081, opened 2024-08-25
plugins/nvim-jdtls: allow lua on nvim-jdtls.data

jcszymansk:
Make it possible to set data directory with lua, so it can be project-dependent.

JDTLS requires separate data directory for every project.

This PR enables users to set e.g.

```nix
data.__raw = "vim.fn.stdpath 'cache' .. '/jdtls/' .. vim.fn.fnamemodify(vim.fn.getcwd(), ':t')";
```

and thus satisfy this requirement.

### DIFF
--- a/plugins/languages/nvim-jdtls.nix
+++ b/plugins/languages/nvim-jdtls.nix
@@ -44,7 +44,7 @@ in
     '';
 
     data = mkOption {
-      type = types.nullOr types.str;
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
Thanks for the addition!

As mentioned, we will refactor this module eventually to current standards. But, this is a safe way to just add additional support to this existing option.

khaneliman:
@mergifyio queue

PR #1904, opened 2024-07-21
plugins/lazy.nvim: Import Statements & Adding user-specified packages to `lazyPath`

my7h3le:
When using the lazy nvim plugin manager I couldn't find any way to add `import` statements to import additional plugin modules in any given spec without having to resort to just using lua in `extraConfigLua` (See: [Importing Specs, `config` & `opts`](https://lazy.folke.io/usage/structuring#%EF%B8%8F-importing-specs-config--opts)).

This PR adds 2 features:
* Import Statements
* Allow user-specified packages to be made available in the `lazyPath` 

> I'm still pretty new to Nix and I'd appreciate any suggestions or comments on this PR

### Importing a single plugin module
A single plugin module can be imported like this
```Nix
{
  plugins = {
    lazy = {
      enable = true;
      plugins = with pkgs.vimPlugins; [
        {
          pkg = <some-package>;
          import = "<some-plugin-module>";
        }
      ];
    };
  };
}
```
The lua equivalent for this would look something like this:
```lua
require("lazy").setup({
  spec = {
    { dir = "<path-to-pkg-in-nix-store>", import = "<some-plugin-module>" },
  }
})
```

### Importing multiple plugin modules
```
Several plugin modules can be imported like this
```Nix
{
  plugins = {
    lazy = {
      enable = true;
      plugins = with pkgs.vimPlugins; [
        {
          pkg = <some-package>;
          import = [
            "<some-plugin-module>"
            "<another-plugin-module>"
          ];
        }
      ];
    };
  };
}
```
The lua equivalent for this would look something like this:
```lua
require("lazy").setup({
  spec = {
    { dir = "<path-to-pkg-in-nix-store>", import = "<some-plugin-module>" },
    { import = "<another-plugin-module>" },
  }
})
```

## Allow user-specified packages to be made available in the `lazyPath` 
Some plugin modules will look for packages/plugins in the `lazyPath`. This was an issue I encountered when trying to set up LazyVim with nixvim. Adding nix packages to the `lazyPath` can be done now from any plugin spec.

>Maybe using the name `packages` isn't good for this, but I couldn't think of anything better at the time and I'm open to suggestions.

Single package to add to `lazyPath` example:
```Nix
{
  plugins = {
    lazy = {
      enable = true;
      plugins = with pkgs.vimPlugins; [
        {
          pkg = <some-package>;
          packages = [
            "<some-package-to-be-added-to-lazy-path>"
            "<another-package-to-be-added-to-lazy-path>"
          ];
        }
[...]

PR #2082, opened 2024-08-25
plugins/lazy.nvim: switch to mkNeovimPlugin

my7h3le:
This PR came about as a result of the comments made in this PR [#1904](https://github.com/nix-community/nixvim/pull/1904). It refactors the existing lazy plugin manager to use `mkNeovimPlugin` and makes the plugin type freeform.

Closes #2174 #1904

Here's an example of how to set up `LazyVim` using the changes to this plugin (Note: this is a standalone flake, saving it to any directory as `flake.nix` and running `nix build` should be enough, to launch neovim run `./result/bin/nvim`)

```Nix
# flake.nix

{
  description = "Setup LazyVim using NixVim";

  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  inputs.nixvim.url = "github:my7h3le/nixvim/lazy-plugin-manager-freeform";
  inputs.nixvim.inputs.flake-parts.follows = "flake-parts";
  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
  inputs.flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";

  outputs =
    {
      self,
      nixpkgs,
      nixvim,
      flake-parts,
    }@inputs:
    flake-parts.lib.mkFlake { inherit inputs; } {
      systems = [
        "aarch64-darwin"
        "aarch64-linux"
        "x86_64-darwin"
        "x86_64-linux"
      ];

      perSystem =
        {
          pkgs,
          lib,
          system,
          ...
        }:
        let
          config = {

            # This path is required to be added at runtime at the beginning of
            # the `init.lua` file for `nvim-treesitter`'s parser_install_dir
            # option to work correctly. Unfortunately `nvim-treesitter`
            # requires the parser installation directory to be read-writeable
            # so it can't be in a path in the nix store.
            extraConfigLuaPre = ''
              vim.opt.runtimepath:prepend(vim.fs.joinpath(vim.fn.stdpath("data"), "site"))
            '';

            plugins = {
              lazy = {
                enable = true;
                settings = {
                  defaults = {
                    # When true `lazy.nvim` will lazy load all plugins by default (`lazy.nvim` sets this
                    # to false by default). 
                    lazy = false;
                  };
                  install = {
                    # Don't install missing plugin dependencies.
                    missing = false;
                  };
                  dev = {
                    # Prevent `lazy.nvim` from fallbacking to fetch plugins
                    # from GitHub. You can set this to true as well.
                    fallback = false;
                  };
                  performance = {
                    rtp = {
                      # Disable some rtp plugins
                      disabled_plugins = [
                        "gzip"
                        "matchit"
                        "matchparen"
                        "netrwPlugin"
                        "tarPlugin"
                        "tohtml"
                        "tutor"
                        "zipPlugin"
                        "gzip"
                        "tarPlugin"
                        "tohtml"
                        "zipPlugin"
                      ];
                    };
                  };
                };

                plugins =
                  with pkgs.vimPlugins;
                  [
                    {
                      pkg = LazyVim;
                      import = "lazyvim.plugins";
                      # LazyVim needs `trouble-nvim` specified as a dependency
                      dependencies = trouble-nvim;
                    }
                    {
                      pkg = nvim-treesitter;
                      opts = {
                        ensure_installed = [
                          "vim"
                          "regex"
                          "lua"
                          "bash"
                          "markdown"
                          "markdown_inline"
                        ];
                        highlight = {
                          enable = true;
                        };
                        parser_install_dir.__raw = ''vim.fs.joinpath(vim.fn.stdpath("data"), "site")'';
                      };
                    }
                    {
                      pkg = telescope-fzf-native-nvim;
                      enabled = true;
                    }
                    {
                      pkg = mason-nvim;
                      enabled = false;
                    }
                    {
                      pkg = mason-lspconfig-nvim;
                      enabled = false;
                    }
                    luvit-meta
                    bufferline-nvim
                    grug-far-nvim
                    tokyonight-nvim
                    cmp-buffer
                    cmp-nvim-lsp
                    cmp-path
                    trouble-nvim
                    conform-nvim
                    dashboard-nvim
                    dressing-nvim
                    flash-nvim
                    friendly-snippets
                    gitsigns-nvim
                    indent-blankline-nvim
                    lazydev-nvim
                    lualine-nvim
                    neo-tree-nvim
                    noice-nvim
                    nui-nvim
                    nvim-autopairs
                    nvim-cmp
                    nvim-lint
                    nvim-lspconfig
                    nvim-notify
                    nvim-ts-autotag
                    nvim-snippets
                    nvim-spectre
                    nvim-treesitter-textobjects
                    persistence-nvim
                    plenary-nvim
                    telescope-nvim
                    todo-comments-nvim
                    ts-comments-nvim
                    which-key-nvim
                  ]
                  ++
                    # nix bundles all `mini-nvim` modules together, but LazyVim
                    # expects individual `mini.nvim` modules so map their
                    # individual names to the bundled `mini-nvim` package.
                    lib.map
                      (mini-module: {
                        name = "${mini-module}";
                        pkg = pkgs.vimPlugins.mini-nvim;
                      })
                      [
                        "mini.ai"
                        "mini.icons"
                        "mini.move"
                        "mini.nvim"
                        "mini.pairs"
                        "mini.splitjoin"
                        "mini.trailspace"
                      ];
              };
            };

          };
          nixvim' = nixvim.legacyPackages."${system}";
          nvim = nixvim'.makeNixvim config;
        in
        {
          packages = {
            inherit nvim;
            default = nvim;
          };
        };
    };
}
```

### DIFF
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -1,42 +1,40 @@
+# TODO: add function (_, opts) docs
 {
-  lib,
-  helpers,
   config,
+  lib,
   pkgs,
   ...
 }:
-with lib;
 let
-  cfg = config.plugins.lazy;
-  lazyPlugins = cfg.plugins;
+  inherit (lib.nixvim)
+    defaultNullOpts
+    keymaps
+    mkNullOrOption
+    mkNullOrLuaFn
+    mkNullOrStrLuaFnOr
+    ;
+  inherit (lib) types;
 
-  processPlugin =
-    plugin:
-    let
-      mkEntryFromDrv =
-        p:
-        if lib.isDerivation p then
-          {
-            name = "${lib.getName p}";
-            path = p;
-          }
-        else
-          {
-            name = "${lib.getName p.pkg}";
-            path = p.pkg;
-          };
-      processDependencies =
-        if plugin ? dependencies && plugin.dependencies != null then
-          builtins.concatMap processPlugin plugin.dependencies
-        else
-          [ ];
-    in
-    [ (mkEntryFromDrv plugin) ] ++ processDependencies;
+  # The regex used in this match function matches a string containing two
+  # segments separated by a single forward slash ('/'). Each segment is
+  # composed of one or more characters that are neither newlines ('\n')
+  # nor forward slashes ('/').
+  #
+  # The pattern ensures that both segments must be non-empty and must not
+  # include line breaks or slashes within them.
+  isShortGitURL = x: lib.isStringLike x && builtins.match "[^\n/]+/[^\n/]+" (toString x) != null;
 
-  processedPlugins = builtins.concatLists (builtins.map processPlugin lazyPlugins);
-  lazyPath = pkgs.linkFarm "lazy-plugins" processedPlugins;
+  # The regex used in this match function matches any string that starts
+  # with either `https://` or `http://`.
+  isGitURL = x: lib.isStringLike x && builtins.match "https?://.*" (toString x) != null;
+
+  inherit (config) isDocs;
 in
-{
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lazy";
+  packPathName = "lazy.nvim";
+  package = "lazy-nvim";
+
   # TODO: added 2025-04-06, remove after 25.05
   imports = [
     (lib.nixvim.mkRemovedPackageOptionModule {
@@ -45,184 +43,486 @@ in
     })
   ];
 
-  options = {
-    plugins.lazy = {
-      enable = mkEnableOption "lazy.nvim";
+  maintainers = with lib.maintainers; [
+    MattSturgeon
+    my7h3le
+  ];
 
-      package = lib.mkPackageOption pkgs [
-        "vimPlugins"
-        "lazy-nvim"
-      ] { };
+  settingsOptions = {
+    # TODO: check and maybe reword this description
+    git.url_format = defaultNullOpts.mkStr "https://github.com/%s.git" ''
+      The default url format that `lazy.nvim` expects short plugin urls to be
+      in. (See: upstream docs for `config.git.url_format` defined here
+      https://lazy.folke.io/configuration);
+    '';
 
-      plugins =
-        with types;
-        let
-          pluginType = either package (submodule {
-            options = {
-              dir = helpers.mkNullOrOption str "A directory pointing to a local plugin";
+    dev.fallback = defaultNullOpts.mkBool false ''
+      When false, `lazy.nvim` won't try to use git to fetch local plugins that
+      don't exist.
+    '';
 
-              pkg = mkOption {
-                type = package;
-                description = "Vim plugin to install";
-              };
-
-              name = helpers.mkNullOrOption str "Name of the plugin to install";
-
-              dev = helpers.defaultNullOpts.mkBool false ''
-                When true, a local plugin directory will be used instead.
-                See config.dev
-              '';
-
-              lazy = helpers.defaultNullOpts.mkBool true ''
-                When true, the plugin will only be loaded when needed.
-                Lazy-loaded plugins are automatically loaded when their Lua modules are required,
-                or when one of the lazy-loading handlers triggers
-              '';
-
-              enabled = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`" ''
-                When false then this plugin will not be included in the spec. (accepts fun():boolean)
-              '';
-
-              cond = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`" ''
-                When false, or if the function returns false,
-                then this plugin will not be loaded. Useful to disable some plugins in vscode,
-                or firenvim for example. (accepts fun(LazyPlugin):boolean)
-              '';
-
-              dependencies = helpers.mkNullOrOption (eitherRecursive str listOfPlugins) "Plugin dependencies";
-
-              init = helpers.mkNullOrLuaFn "init functions are always executed during startup";
-
-              config = helpers.mkNullOrStrLuaFnOr (types.enum [ true ]) ''
-                config is executed when the plugin loads.
-                The default implementation will automatically run require(MAIN).setup(opts).
-                Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name.
-                See also opts. To use the default implementation without opts set config to true.
-              '';
-
-              main = helpers.mkNullOrOption str ''
-                You can specify the main module to use for config() and opts(),
-                in case it can not be determined automatically. See config()
-              '';
-
-              submodules = helpers.defaultNullOpts.mkBool true ''
-                When false, git submodules will not be fetched.
-                Defaults to true
-              '';
-
-              event =
-                with lib.types;
-                helpers.mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
-
-              cmd =
-                with lib.types;
-                helpers.mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on command";
-
-              ft =
-                with lib.types;
-                helpers.mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on filetype";
-
-              keys =
-                with lib.types;
-                helpers.mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on key mapping";
-
-              module = helpers.mkNullOrOption (enum [ false ]) ''
-                Do not automatically load this Lua module when it's required somewhere
-              '';
-
-              priority = helpers.mkNullOrOption number ''
-                Only useful for start plugins (lazy=false) to force loading certain plugins first.
-                Default priority is 50. It's recommended to set this to a high number for colorschemes.
-              '';
-
-              optional = helpers.defaultNullOpts.mkBool false ''
-                When a spec is tagged optional, it will only be included in the final spec,
-                when the same plugin has been specified at least once somewhere else without optional.
-                This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part
-                of the user's plugins
-              '';
-
-              opts =
-                with lib.types;
-                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) ''
-                  opts should be a table (will be merged with parent specs),
-                  return a table (replaces parent specs) or should change a table.
-                  The table will be passed to the Plugin.config() function.
-                  Setting this value will imply Plugin.config()
-                '';
-            };
-          });
-
-          listOfPlugins = types.listOf pluginType;
-        in
-        mkOption {
-          type = listOfPlugins;
-          default = [ ];
-          description = "List of plugins";
-        };
-    };
+    install.missing = defaultNullOpts.mkBool false ''
+      When false, `lazy.nvim` won't try to install missing plugins on startup.
+      Setting this to true won't increase startup time.
+    '';
   };
 
-  config = mkIf cfg.enable {
-    extraPlugins = [ cfg.package ];
+  extraOptions =
+    with types;
+    let
+      shortGitURL = lib.mkOptionType {
+        name = "shorGitURL";
+        description = "a short git url of the form `owner/repo`";
+        descriptionClass = "noun";
+        check = isShortGitURL;
+        merge = lib.mergeEqualOption;
+      };
 
-    dependencies.git.enable = lib.mkDefault true;
+      gitURL = lib.mkOptionType {
+        name = "gitURL";
+        description = "a git url";
+        descriptionClass = "noun";
+        check = isGitURL;
+        merge = lib.mergeEqualOption;
+      };
 
-    extraConfigLua =
-      let
-        pluginToLua =
-          plugin:
-          if isDerivation plugin then
-            { dir = "${lazyPath}/${lib.getName plugin}"; }
-          else
-            {
-              "__unkeyed" = plugin.name;
+      lazyPluginSourceType = oneOf [
+        package
+        path
+        shortGitURL
+        gitURL
+      ];
 
-              inherit (plugin)
-                cmd
-                cond
-                config
-                dev
-                enabled
-                event
-                ft
-                init
-                keys
-                lazy
-                main
-                module
-                name
-                optional
-                opts
-                priority
-                submodules
-                ;
+      lazyPluginType = coercedTo lazyPluginSourceType (source: { inherit source; }) (
+        submodule (
+          { config, ... }:
 
-              dependencies = helpers.ifNonNull' plugin.dependencies (
-                if isList plugin.dependencies then (pluginListToLua plugin.dependencies) else plugin.dependencies
-              );
+          {
+            freeformType = attrsOf anything;
+            options = {
+              source = mkNullOrOption lazyPluginSourceType ''
+                The `source` option i.e. `plugins.lazy.plugins.<plugin>.source` is just
+                a convenience nixvim option that makes it easier to specify a plugin
+                source whether it be a:
 
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
+                  - nix package
+                  - directory path
+                  - git url
+                  - short git url of the form `owner/repo`
+
+                and the `source` option itself is not a part of the upstream
+                `lazy.nvim` plugin spec (See: https://lazy.folke.io/spec).
+
+                As a result the values given to `source` need to be mapped to
+                properties in the upstream `lazy.nvim` plugin spec i.e.
+                ([1]|dir|url). After which the `source` attribute itself will
+                be stripped from the final lua config.
+                ```
+              '';
+
+              # TODO: rework 3rd paragraph of this, make sure you add the fact
+              # that with custom name it has to be defined somewhere else in
+              # the list
+              name = mkNullOrOption str ''
+                A custom name for the plugin used for the local plugin
+                directory and as the display name. By default nixvim will set
+                a default value for this depending on what was specified to
+                `plugins.lazy.plugins.<plugin>.source`, namely if a nix package
+                was given to `plugins.lazy.plugins.<plugin>.source` then name
+                will default to the nix package's name, and if a git url or a
+                short git url was specified, then name will default to the base
+                name of that url.
+
+                It is also worth mentioning that a plugin can have the option
+                `plugins.lazy.plugins.<plugin>.name` defined but not have
+                `plugins.lazy.plugins.<plugin>.source` defined.
+
+                In these cases, a custom `name` can be used instead of `source`
+                and the plugin spec options given will be applied or merged with
+                the original plugin definition. This is particularly useful for
+                plugins that bundle multiple modules (e.g. mini-nvim, which
+                includes mini-ai, mini-trailspace, etc.), where you need to
+                modify options for individual modules without affecting the
+                entire bundle.
+              '';
+
+              dev = defaultNullOpts.mkBool false ''
+                When true, `lazy.nvim` will look for this plugin in the local
+                plugin directory defined at `plugin.lazy.settings.dev.path`.
+              '';
+
+              lazy = defaultNullOpts.mkBool true ''
+                When true, the plugin will only be loaded when needed.
+
+                Lazy-loaded plugins are automatically loaded when their Lua
+                modules are required, or when one of the lazy-loading
+                handlers are triggered.
+
+                You can define which triggers load this plugin using
+                `plugins.lazy.plugins.<plugin>.[event|cmd|ft|keys]`.
+              '';
+
+              enabled = defaultNullOpts.mkStrLuaFnOr bool true ''
+                When false or if a function that returns false is defined
+                then this plugin will not be included in the final spec. The
+                plugin will also be uninstalled when true if the plugin is an
+                out of tree non nix package plugin. (accepts fun():boolean).
+              '';
+
+              cond = defaultNullOpts.mkStrLuaFnOr bool true ''
+                Behaves the same as `plugins.lazy.plugins.<plugin>.enabled`,
+                but won't uninstall the plugin when the condition is false.
+                Useful to disable some plugins in vscode, or firenvim for
+                example. 
+
+                Note: Since out of tree non nix package plugins are only ever
+                uninstalled by `lazy.nvim`, this option as far as nix package
+                plugins are concerned, only serves the same purpose as
+                `plugins.lazy.plugins.<plugin>.enabled`.
+              '';
+
+              # WARNING: Be very careful if changing the type of
+              # `dependencies`. Choosing the wrong type may cause a stack
+              # overflow due to infinite recursion, and it's very possible that
+              # the test cases won't catch this problem. To be safe, perform
+              # thorough manual testing if you do change the type of
+              # `dependencies`.
+              dependencies =
+                let
+                  lazyPluginDependenciesType =
+                    if isDocs then
+                      # Use a stub type for documentation purposes We don't
+                      # need to repeat all the plugin-type sub-options again in
+                      # the docs It'd also be infinitely recursive
+                      lib.mkOptionType {
+                        name = "pluginDependencies";
+                        description = "plugin submodule";
+                        descriptionClass = "noun";
+                        check = throw "should not be used";
+                        merge = throw "should not be used";
+                      }
+                    else
+                      # NOTE: use attrsOf for the LHS, because coercedTo only supports submodules on the RHS
+                      coercedTo (either (attrsOf anything) lazyPluginSourceType) lib.toList (listOf lazyPluginType);
+                in
+                mkNullOrOption lazyPluginDependenciesType ''
+                  In terms of setting values to this option it is effectively
+                  the same as `plugins.lazy.plugins`, with the following
+                  additions in terms of what it can be:
+
+                    - It can also be single string such as for a custom plugin
+                      name e.g. `dependencies = "nvim-treesitter";`
+
+
+                      If it is just a custom name however, then the plugin spec
+                      must be defined somewhere else in the list given to
+                      `plugins.lazy.plugins`. For example:
+
+                      ```Nix
+                        {
+                          plugins.lazy.plugins = with pkgs.vimPlugins; [
+                            {
+
+                              # Only a custom name was given to
+                              # `plugins.lazy.plugins.<plugin>.dependencies`,
+                              # therefore a plugin spec with this name must be
+                              # defined somewhere else in the list given to
+                              # `plugins.lazy.plugins`.
+                              source = nvim-treesitter-textobjects;
+                              dependencies = "nvim-treesitter";
+                            }
+
+                            # We define the plugin spec with name "nvim-treesitter"
+                            # here:
+                            {
+                              source = nvim-treesitter;
+
+                              # Note: By default if a nix package is given to
+                              # `plugins.lazy.plugins.<plugin>.source`, then the
+                              # `name` option will have have a default value of
+                              # that nix package's name. So in this example the
+                              # custom name of "nvim-treesitter" doesn't need to
+                              # be explicitly set here as the nix package
+                              # `pkgs.vimPlugins.nvim-treesitter` name attribute
+                              # is already "nvim-treesitter".
+                              name = "nvim-treesitter";
+
+                              opts = {
+                                highlight = {
+                                  enable = true;
+                                };
+                                ensure_installed = [
+                                  "vim"
+                                  "regex"
+                                  "lua"
+                                  "bash"
+                                  "markdown"
+                                  "angular"
+                                  "markdown_inline"
+                                  "hyprlang"
+                                ];
+                                parser_install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+                              };
+                            }
+                          ];
+                        }
+                      ```
+
+                    - A single nix plugin package (e.g. `dependencies = pkgs.vimPlugins.nvim-treesitter;`)
+
+                  Outside of these additions, what it can be set to is
+                  effectively the same as `plugins.lazy.plugins` i.e. A list of
+                  nix plugin packages and or attribute sets denoting plugin
+                  specs that should be loaded, and loaded before the current
+                  plugin loads. Dependencies are always lazy-loaded unless
+                  specified otherwise. When specifying a name, make sure the
+                  plugin spec has been defined somewhere else. (See:
+                  https://lazy.folke.io/spec). 
+                '';
+
+              init = mkNullOrLuaFn ''
+                Functions that are always executed during startup. Mostly
+                useful for setting vim.g.* configuration used by Vim plugins
+                startup.
+              '';
+
+              config = mkNullOrStrLuaFnOr (enum [ true ]) ''
+                A lua function with signature `fun(LazyPlugin, opts:table)`
+                that gets executed when the plugin loads. It can also just be
+                a `true` boolean.
+
+                The default implementation will automatically run
+                `require(MAIN).setup(opts)` if
+                `plugins.lazy.plugins.<plugin>.opts` is set or if `config`
+                itself is set to `true`. Lazy uses several heuristics to
+                determine the plugin's MAIN module automatically based on the
+                plugin's name. (`plugins.lazy.plugins.<plugin>.opts` is the
+                recommend way to configure plugins).
+              '';
+
+              main = mkNullOrOption str ''
+                You can specify the main module to use for
+                `plugins.lazy.plugins.<plugin>.config` and
+                `plugins.lazy.plugins.<plugin>.opts`, in case it can not be
+                determined automatically.
+
+                See `plugins.lazy.plugins.<plugin>.config`.
+              '';
+
+              submodules = defaultNullOpts.mkBool true ''
+                When `false`, git submodules will not be fetched. Defaults to `true`
+              '';
+
+              event = mkNullOrOption (maybeRaw (either str (listOf str))) ''
+                Lazy-load on event. Events can be specified as `BufEnter`, a
+                pattern like `BufEnter *.lua` or a list of events `event = [
+                "BufRead" "BufEnter" ];`. It can also be a lua function of
+                signature `fun(self:LazyPlugin, event:string[])` that returns a
+                list of strings.
+
+                See https://neovim.io/doc/user/autocmd.html for a more info neovim events.
+                Also see upstream `lazy.nvim` docs at https://lazy.folke.io/spec
+              '';
+
+              cmd = mkNullOrOption (maybeRaw (either str (listOf str))) ''
+                Lazy-load on command. Commands can be specified as a single
+                string like `StartupTime` or a list of strings to lazy load the
+                plugin on any given command. This can also be a lua function that
+                returns a list of strings.
+
+                Example:
+
+                ```Nix
+                  {
+                    plugins.lazy.plugins = with pkgs.vimPlugins; [
+                      {
+                        source = vim-startuptime;
+                        # Lazy load plugin when `StartupTime` command is invoked.
+                        cmd = "StartupTime";
+                      }
+                    ];
+                  }
+                ```
+              '';
+
+              ft = mkNullOrOption (maybeRaw (either str (listOf str))) ''
+                Lazy-load plugin only on a given filetype/filetypes e.g.:
+
+                ```Nix
+                plugins.lazy.plugins = with pkgs.vimPlugins; [
+                  {
+                    source = neorg;
+                    # Only load plugin on "norg" filetyes
+                    # Note: this can also be a list as well (i.e. `ft = [ "norg" ];`)
+                    ft = "norg";
+                    opts = {
+                      load."['core.defaults']" = [ ];
+                    };
+                  }
+                ];
+                ```
+
+                Can also be a lua function that returns a list of strings, that
+                denote filetyes. Example:
+
+                ```Nix
+                plugins.lazy.plugins = with pkgs.vimPlugins; [
+                  {
+                    source = neorg;
+                    # Can also be a function that returns a list of filetypes
+                    ft = "function() return { "norg" } end"
+                    opts = {
+                      load."['core.defaults']" = [ ];
+                    };
+                  }
+                ];
+                ```
+              '';
+
+              # TODO: add better description here
+              keys = mkNullOrOption (listOf (
+                keymaps.mkMapOptionSubmodule {
+                  defaults = {
+                    mode = "n";
+                  };
+                  # HACK: with how `mkMapOptionSubmodule` is currently
+                  # defined, mode will keep being set to default value
+                  # denoted by `defaults.mode`, and the user can not override
+                  # it without this workaround.
+                  extraOptions = {
+                    mode = keymaps.mkModeOption "n";
+                    action = mkNullOrOption (maybeRaw str) ''
+                      The action to execute.
+                    '';
+                    ft = mkNullOrOption (either (maybeRaw str) (listOf (maybeRaw str))) ''
+                      Make the keybind trigger for only certain file type/s.
+                    '';
+                  };
+                }
+              )) "Lazy-load on key mapping";
+
+              module = mkNullOrOption (enum [ false ]) ''
+                Do not automatically load this Lua module when it's required somewhere.
+              '';
+
+              priority = defaultNullOpts.mkInt 50 ''
+                Only useful for start plugins i.e.
+                `plugins.lazy.plugins.<plugin>.lazy = false` to force loading
+                certain plugins first. Default priority is 50. It's
+                recommended to set this to a high number for colorschemes.
+              '';
+
+              optional = defaultNullOpts.mkBool false ''
+                Optional specs are only included in the final configuration
+                if the corresponding plugin is also specified as a required
+                (non-optional) plugin elsewhere. This feature is
+                particularly helpful for Neovim distributions, allowing
+                them to pre-configure settings for plugins that users may
+                or may not have installed.
+              '';
+
+              opts = defaultNullOpts.mkAttrsOf anything { } ''
+                The opts value can be one of the following:
+
+                - An attribute set: The equivalent of this in lua would be a
+                table. This options defined in this attribute set will be
+                merged with any existing configuration settings from the parent
+                specification of the plugin.
+
+                - A function that returns a table: The returned table will
+                completely replace any existing configuration settings from
+                the plugin's parent specifications.
+
+                - A function that modifies a table: This function will
+                receive the existing configuration table as an argument and
+                can modify it directly.
+
+                In all cases, the resulting configuration table will be passed
+                to the `plugins.lazy.plugins.<plugin>.config` function. Setting
+                the opts value automatically implies that the
+                `plugins.lazy.plugins.<plugin>.config` function will be called.
+                (See: https://lazy.folke.io/spec#spec-setup)
+              '';
             };
 
-        pluginListToLua = map pluginToLua;
-
-        plugins = pluginListToLua cfg.plugins;
-
-        packedPlugins = if length plugins == 1 then head plugins else plugins;
-      in
-      mkIf (cfg.plugins != [ ]) ''
-        require('lazy').setup(
-          {
-            dev = {
-              path = "${lazyPath}",
-              patterns = {"."},
-              fallback = false
-            },
-            spec = ${lib.nixvim.toLuaObject packedPlugins}
+            # Set the name of a plugin by default. This is mostly for
+            # convenience as it avoids long nix store paths be shown as the
+            # display names when running `:Lazy`.
+            config.name = lib.mkIf (config.source != null) (
+              if (lib.isDerivation config.source) then
+                lib.mkDefault "${lib.getName config.source}"
+              else
+                lib.mkDefault "${builtins.baseNameOf config.source}"
+            );
           }
         )
-      '';
+      );
+
+      lazyPluginsListType = listOf lazyPluginType;
+    in
+    {
+      luarocksPackage = lib.mkPackageOption pkgs "luarocks" { nullable = true; };
+
+      plugins = lib.mkOption {
+        type = lazyPluginsListType;
+        default = [ ];
+        description = "List of plugins";
+      };
+    };
+
+  extraConfig = cfg: {
+    dependencies.git.enable = lib.mkDefault true;
+
+    extraPackages = [
+      cfg.luarocksPackage
+    ];
+    plugins.lazy.settings.spec =
+      let
+        convertToLazyKeyMappingType =
+          km:
+          {
+            __unkeyed-1 = km.key;
+          }
+          // lib.optionalAttrs (lib.hasAttr "action" km) { __unkeyed-2 = km.action; }
+          // lib.optionalAttrs (lib.hasAttr "mode" km) { inherit (km) mode; }
+          // lib.optionalAttrs (lib.hasAttr "ft" km) { inherit (km) ft; }
+          // lib.removeAttrs km [
+            "action"
+            "mode"
+            "ft"
+          ];
+
+        pluginToSpec =
+          plugin:
+          lib.concatMapAttrs (
+            key: value:
+            if key == "source" && value != null then
+              {
+                dir = if (lib.isDerivation value || lib.isPath value) then "${value}" else null;
+                __unkeyed = if (isShortGitURL value) then value else null;
+                url = if (isGitURL value) then value else null;
+              }
+            else if (key == "name" && value != null) && ((plugin.source or null) == null) then
+              { __unkeyed = value; }
+            else if key == "dependencies" && value != null then
+              { dependencies = map pluginToSpec value; }
+            # If the user explicitly sets the 'dir' option, use the provided
+            # value instead of an extrapolated default.
+            else if (key == "dir" && value != null) then
+              { dir = value; }
+            # If the user explicitly sets the 'url' option, use the provided
+            # value instead of an extrapolated default.
+            else if (key == "url" && value != null) then
+              { url = value; }
+            # If the 'keys' option is given, we need to explicitly convert that into a form that `lazy.nvim` expects.
+            else if (key == "keys" && value != null) then
+              { keys = map convertToLazyKeyMappingType value; }
+            # Preserve all other key value pair mappings, this also handles
+            # freeform options.
+            else
+              { "${key}" = value; }
+          ) plugin;
+      in
+      map pluginToSpec cfg.plugins;
   };
 }
--- a/tests/test-sources/plugins/pluginmanagers/lazy.nix
+++ b/tests/test-sources/plugins/pluginmanagers/lazy.nix
@@ -1,20 +1,82 @@
 { pkgs, ... }:
+
+# Note: do not use `plenary-nvim` or any plugin that is a `rockspec` for tests,
+# this is because `lazy.nvim` by default uses the luarocks package manager to
+# process rockspecs. It might be possible to use a rockspec in a test if the
+# rockspec itself does not depend on any other rockspecs but this has not been
+# tested (See:
+# https://github.com/nix-community/nixvim/pull/2082#discussion_r1746585453).
+#
+# Also the plugins and dependency combinations used in the tests are
+# arbitrary.
+
 {
   # Empty configuration
   empty = {
     plugins.lazy.enable = true;
   };
 
-  test = {
+  no-packages = {
+    dependencies.git.enable = false;
+    plugins.lazy.enable = true;
+  };
+
+  nix-package-plugins = {
+    plugins.lazy = {
+      enable = true;
+
+      plugins = with pkgs.vimPlugins; [
+        # A plugin can be just a nix package
+        vim-closer
+
+        # A plugin can also be an attribute set with `source` set to a nix
+        # package.
+        { source = trouble-nvim; }
+      ];
+    };
+  };
+
+  out-of-tree-plugins = {
+    # Don't run neovim for this test, as it's purely to test module evaluation.
+    test.runNvim = false;
+    plugins.lazy = {
+      enable = true;
+
+      plugins = [
+        # `source` can also be a short git url of the form `owner/repo`
+        { source = "echasnovski/mini.align"; }
+
+        # `source` can also be a full git url with `http://` or `https://`
+        {
+          source = "https://github.com/nvim-telescope/telescope.nvim";
+          enabled = true;
+          version = false;
+        }
+        {
+          source = "http://github.com/norcalli/nvim-colorizer.lua";
+          enabled = true;
+          version = false;
+        }
+      ];
+    };
+  };
+
+  general-tests = {
     plugins.lazy = with pkgs.vimPlugins; {
       enable = true;
 
       plugins = [
-        vim-closer
+        # Test freeform
+        {
+          source = trouble-nvim;
+          # The below is not actually a property in the `lazy.nvim` plugin spec
+          # but is purely to test freeform capabilities of the `lazyPluginType`.
+          blah = "test";
+        }
 
         # Load on specific commands
         {
-          pkg = vim-dispatch;
+          source = vim-dispatch;
           optional = true;
           cmd = [
             "Dispatch"
@@ -26,13 +88,46 @@
 
         # Load on an autocommand event
         {
-          pkg = vim-matchup;
+          source = vim-matchup;
           event = "VimEnter";
+        }
+
+        # Load on key mapping
+        {
+          source = neo-tree-nvim;
+          keys = [
+            {
+              action = "<cmd>make<CR>";
+              key = "<C-m>";
+              mode = [
+                "v"
+                "n"
+              ];
+              ft = [
+                "*.txt"
+                "*.cpp"
+              ];
+              options = {
+                silent = true;
+              };
+            }
+          ];
+        }
+
+        # Load on key mapping without specifying action
+        {
+          source = markdown-preview-nvim;
+          cmd = "MarkdownPreview";
+          keys = [
+            {
+              key = "<C-g>";
+            }
+          ];
         }
 
         # Load on a combination of conditions: specific filetypes or commands
         {
-          pkg = ale;
+          source = ale;
           name = "w0rp/ale";
           ft = [
             "sh"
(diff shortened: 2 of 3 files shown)


plugins/pluginmanagers/lazy.nix
```diff
@@ -41,10 +41,13 @@ in
     plugins.lazy = {
       enable = mkEnableOption "lazy.nvim";
 
+      setup = mkOption { type = types.submodule { freeformType = types.attrsOf types.anything; }; };
```
MattSturgeon, line 44: You probably want `mkSettingsOption`: https://github.com/nix-community/nixvim/blob/665680a5caeb57ff73e0a112ac805f5456f2a91d/lib/options.nix#L337-L342

Although, if there's no sub-options declared, we could also just make this a `type = attrsOf anything` option without using a submodule. (I'd prefer this)

Either way, we should use the standard description _"Options provided to the `require('lazy').setup` function."_ https://github.com/nix-community/nixvim/blob/665680a5caeb57ff73e0a112ac805f5456f2a91d/lib/neovim-plugin.nix#L35

We'd also want to name the option `settings`. While I agree naming it `setup` does make sense, we've standardized on using `settings` for structured config that'll be passed to a plugins configuration method (i.e. usually the `setup` function).

On reflection, I think it makes sense to just use `mkNeovimPlugin` here.

plugins/pluginmanagers/lazy.nix
```diff
@@ -203,13 +213,18 @@ in
...
+              {
+                dev = {
+                  path = "${lazyPath}";
+                  patterns = [ "." ];
+                  fallback = false;
+                };
+                spec = packedPlugins;
+              }
+              // cfg.setup
+            )
```
MattSturgeon, line 227: Instead of updating the final value here, we would usually assign these values to `plugins.lazy.settings` using the module system. That way we can take advantage of things like `mkDefault`, `mkIf`, etc.

plugins/pluginmanagers/lazy.nix
```diff
@@ -160,40 +163,47 @@ in
           plugin:
           let
             keyExists = keyToCheck: attrSet: lib.elem keyToCheck (lib.attrNames attrSet);
+            converted =
```
MattSturgeon, line 166: This whole `let` block (`pluginToLua` etc) is making me think there's bigger issues with this plugin that need to be cleaned up. I think that is best tackled as a follow up PR to this one.

It seems the `plugins` option is essentially syntactic sugar around the `settings.sepc` option? While we wouldn't normally do that, maybe it does have some merit here, if implemented in a sane way...

If we do decide it makes sense here, maybe the `plugins.lz-n` would benefit from a similar wrapper, although IIRC we already have a `plugins` option there, which is a structured-config representation of a _separate_ function call to `setup`.

plugins/pluginmanagers/lazy.nix
MattSturgeon: `settingsOptions` should only contain sub-options that correspond exactly to what the upstream plugin expects in its `setup({})` table.

If we need to define additional options (either internal or user-facing), they can be added to `extraOptions` instead (and will sit _alongside_ `settings`).

MattSturgeon: Essentially, the settigns option iteslf should usually serve the role being served here by `setupOptions`.

plugins/pluginmanagers/lazy.nix
MattSturgeon: I think it's fine to have a `plugins` option, but the `cfg.plugins` value should be assigned to `settings.spec` _without_ using a `setupOptions` let-variable.

e.g.
```nix
extraConfig = cfg: {
  plugins.lazy.settings.spec =
    let
      # omitted
    in
      packedPlugins;
}
```

MattSturgeon:
> So it seems like nixvim is broken (at least when using the lazy plugin manager) past commit: [78abafe280b1ea102fda879799b590da5d84725f plugins/lz-n: init](https://github.com/nix-community/nixvim/commit/78abafe280b1ea102fda879799b590da5d84725f) in the upstream main branch even without my changes in this PR. I haven't figured out the exact cause yet, and won't be able to rebase and resolve the merge conflicts until then

Strange. That commit doesn't make any changes to the lazy plugin at all, so I don't understand how that could be the case.

MattSturgeon:
> I get `neotest-golang` attribute missing errors. I get similar such errors for different attributes in the subsequent commits on the upstream main branch as well.

Strange that it'd happen on that commit, but that looks like mismatched nixpkgs channels. Essentially we're accessing `pkgs.neotest-golang` which exists on the nixpkgs we use in our flake, but it doesn't exist in the `pkgs` you're using locally to evaluate your flake.

plugins/pluginmanagers/lazy.nix
MattSturgeon: > Or because dev itself is a sub table it should still be in setupOptions?

We shouldn't have a `setupOptions` at all.

> dev is actually in the setup({}) table

Gotcha, wasn't sure based on your earlier example :slightly_smiling_face:

MattSturgeon:
Seems there's two errors being picked up by CI currently, both part of the [`test-17` group](https://buildbot.nix-community.org/#/builders/2798/builds/391):

```
error: builder for '/nix/store/XXX-empty.drv' failed with exit code 1;
       last 8 log lines:
       > ERROR: Error detected while processing /nix/store/6kqddn2g51j1lvbqsv6v17f6565ihpbk-init.lua:
       > Invalid plugin spec {
       >   dev = {
       >     fallback = false,
       >     path = "/nix/store/k71lsglx88q8ycaqj8vrj56hh80r4ccw-lazy-plugins",
       >     patterns = { "." }
       >   }
       > }
```
And
```
error: builder for '/nix/store/XXX-test.drv' failed with exit code 1;
       last 1 log lines:
       > ERROR: gitsigns: git not in path. Aborting setup
```

MattSturgeon:
> Do you know how to run just the test-17 group?

You can do `nix build .#checks.<system>.test-17` (replace `<system>` as appropriate).

If you enter a devshell, you can also use our `tests` command to run a specific test from within that group, e.g. interactively with `tests -i`.

Further, if you don't want to use a devshell, you can still get at individual tests within a group by using the `entries` passthru (tab-completion can help here).

E.g. `nix build .#checks.x86_64-linux.test-17.entries.plugins-pluginmanagers-lazy --show-trace`

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,206 @@
...
+                if lib.isDerivation p then
+                  {
+                    name = "${lib.getName p}";
+                    path = p;
+                  }
+                else
+                  {
+                    name = "${lib.getName p.pkg}";
+                    path = p.pkg;
+                  };
```
MattSturgeon, line 153: If you use [`pkgs.linkFarmFromDrvs`](https://noogle.dev/f/pkgs/linkFarmFromDrvs) below, this can be simplified to:

```nix
if lib.isDerivation p then p else p.pkg
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,206 @@
...
-                ;
-
-              dependencies = helpers.ifNonNull' plugin.dependencies (
-                if isList plugin.dependencies then (pluginListToLua plugin.dependencies) else plugin.dependencies
-              );
-
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
-            };
+        pkgs.linkFarm "lazy-plugins" processedPlugins;
```
MattSturgeon, line 164:
```suggestion
        pkgs.linkFarmFromDrvs "lazy-plugins" (builtins.concatMap processPlugin lazyPlugins);
```

1. Use `linkFarmFromDrvs` to simplify `mkEntryFromDrv`
2. Inlining `processedPlugins` here is more readable IMO
3. Use `concatMap` instead of `concatLists` + `map`

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,206 @@
...
+                    name = "${lib.getName p.pkg}";
+                    path = p.pkg;
+                  };
+              processDependencies =
+                if plugin ? dependencies && plugin.dependencies != null then
+                  builtins.concatMap processPlugin plugin.dependencies
+                else
+                  [ ];
+            in
+            [ (mkEntryFromDrv plugin) ] ++ processDependencies;
```
MattSturgeon, line 160: I think `mkEntryFromDrv` and `processDependencies` can probably be inlined, now that they're simplified.

```suggestion
            [ (if lib.isDerivation plugin then plugin else plugin.pkg) ]
            ++ lib.optionals (plugin.dependencies or null != null) (builtins.concatMap processPlugin plugin.dependencies);
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,206 @@
...
+                else
+                  {
+                    name = "${lib.getName p.pkg}";
+                    path = p.pkg;
+                  };
+              processDependencies =
+                if plugin ? dependencies && plugin.dependencies != null then
+                  builtins.concatMap processPlugin plugin.dependencies
+                else
+                  [ ];
```
MattSturgeon, line 158:
```suggestion
                lib.optionals (plugin.dependencies or null != null) (builtins.concatMap processPlugin plugin.dependencies);
```

1. Use `or null` instead of doing a `?` check _and_ a `!= null` check
2. Use [`optionals`](https://noogle.dev/f/lib/lists/optionals) instead of `if then else []`

MattSturgeon (commented):
> So I was able to simplify the `pluginToLua` function, I'll add that as a temporary commit to this PR, and if it's deemed as appropriate for this PR itself I can squash it but if not I'll move it into it's own separate PR

Looks much better. I think it makes sense as part of this PR.

I only really went over this part of the PR in this review, but I'll try and find time to fully review soon as everything seems to be coming together :grinning:

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
-            };
+        pkgs.linkFarmFromDrvs "lazy-plugins" (builtins.concatMap processPlugin lazyPlugins);
 
-        pluginListToLua = map pluginToLua;
+      setupOptions =
+        let
+          pluginToSpec =
+            plugin:
+            if isDerivation plugin then
+              { dir = "${lazyPath}/${lib.getName plugin}"; }
```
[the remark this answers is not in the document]
MattSturgeon: Any reason you can't just do `dir = "${plugin}"` to stringify the derivation to a path? 

What is `lazyPath` here, the linkFarm's path?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+            else
+              plugin
+              // {
+                "__unkeyed" = plugin.name;
+                dir = if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${plugin.pkg.name}";
```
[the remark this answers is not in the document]
MattSturgeon: > it doesn't work when using `plugin.pkg.name` in the if block above instead of `${lib.getName plugin}`

Interesting. Are you saying `pkgsLinkFarmFromDrvs` works or doesn't work?

`lib.getName` does do a bit of extra magic, for instance it can take either a derivation _or a string_. IDK if it does anything else different though...

MattSturgeon, line 158:
```suggestion
                dir = if plugin.dir or null != null then plugin.dir else "${plugin.pkg}";
```

plugins/pluginmanagers/lazy.nix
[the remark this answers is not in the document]
MattSturgeon: May be useful to check the content of `/nix/store/0pbb5n8ymxqb6xfvnm0vcjpf6rg5pjlq-init.lua`, or try and figure out what is happening with the repl.

> install: skipping file '/dev/fd/63', as it was replaced while being copied

This is an error from the derivation's builder, but kinda confusing.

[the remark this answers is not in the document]
MattSturgeon: Just to clarify, IDK if this is related to your issue, but:

The content of the linkFarm will just be symlinks to the _liked_ derivations, so `${thelazyPath}/${pkg.name}` is functionally identical to `${pkg}`.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
-                if isList plugin.dependencies then (pluginListToLua plugin.dependencies) else plugin.dependencies
-              );
-
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
-            };
+        pkgs.linkFarmFromDrvs "lazy-plugins" (builtins.concatMap processPlugin lazyPlugins);
 
-        pluginListToLua = map pluginToLua;
+      setupOptions =
```
MattSturgeon, line 148: We shouldn't have a `setupOptions`

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
 
-          listOfPlugins = types.listOf pluginType;
+      lazyPath =
+        let
+          processPlugin =
+            plugin:
+            [ (if lib.isDerivation plugin then plugin else plugin.pkg) ]
+            ++ lib.optionals ((plugin.dependencies or null) != null) (
+              builtins.concatMap processPlugin plugin.dependencies
+            );
```
MattSturgeon, line 144: `lazyPath` isn't needed; we never actually use it directly and we already have references to all the derivations that should be installed.

plugins/pluginmanagers/lazy.nix
MattSturgeon, line 153:
```suggestion
              { dir = "${plugin}"; }
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
 
-        pluginListToLua = map pluginToLua;
+      setupOptions =
+        let
+          pluginToSpec =
+            plugin:
+            if isDerivation plugin then
+              { dir = "${lazyPath}/${lib.getName plugin}"; }
+            else
+              plugin
```
MattSturgeon, line 155:
```suggestion
              builtins.removeAttrs plugin [ "name" "pkgs" ]
```

plugins/pluginmanagers/lazy.nix
MattSturgeon, line 158:
```suggestion
                dir = if plugin.dir or null != null then plugin.dir else "${plugin.pkg}";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+              // {
+                "__unkeyed" = plugin.name;
+                dir = if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${plugin.pkg.name}";
+                dependencies = helpers.ifNonNull' plugin.dependencies (
```
MattSturgeon, line 159:
```suggestion
                dependencies = lib.mkIf (plugin.dependencies or null != null) (
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+                  if isList plugin.dependencies then (pluginListToSpecs plugin.dependencies) else plugin.dependencies
```
MattSturgeon, line 160: What is `dependencies` when it isn't a list?

Can we do:

```nix
builtins.map pluginToSpec (toList plugin.dependencies)
```

?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+                );
+                # Set pkg property to null as it's been used and no longer needed.
+                pkg = null;
```
MattSturgeon, line 163: Probably better to just include `"pkgs"` in the `removeAttrs` above instead?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+          pluginListToSpecs = map pluginToSpec;
 
-        packedPlugins = if length plugins == 1 then head plugins else plugins;
-      in
-      mkIf (cfg.plugins != [ ]) ''
-        require('lazy').setup(
+          pluginSpecsList = pluginListToSpecs lazyPlugins;
+          packedPluginSpecs = if length pluginSpecsList == 1 then head pluginSpecsList else pluginSpecsList;
+        in
+        if length pluginSpecsList > 0 then
```
MattSturgeon, line 171: This needlessly calculates the list length

```suggestion
        if pluginSpecsList != [ ] then
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
 
-        plugins = pluginListToLua cfg.plugins;
+          pluginListToSpecs = map pluginToSpec;
 
-        packedPlugins = if length plugins == 1 then head plugins else plugins;
-      in
-      mkIf (cfg.plugins != [ ]) ''
-        require('lazy').setup(
+          pluginSpecsList = pluginListToSpecs lazyPlugins;
+          packedPluginSpecs = if length pluginSpecsList == 1 then head pluginSpecsList else pluginSpecsList;
```
MattSturgeon, line 169: Is this a necessary step? I'd rather just always define a list for simplicity.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
+              };
 
-        plugins = pluginListToLua cfg.plugins;
+          pluginListToSpecs = map pluginToSpec;
 
-        packedPlugins = if length plugins == 1 then head plugins else plugins;
-      in
-      mkIf (cfg.plugins != [ ]) ''
-        require('lazy').setup(
+          pluginSpecsList = pluginListToSpecs lazyPlugins;
```
MattSturgeon, line 168: Sometimes it's better to have the logic together inline :slightly_smiling_face: 

```suggestion
          pluginSpecsList = map pluginToSpec lazyPlugins;
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
           }
-        )
-      '';
-  };
+        else
+          { };
+    in
+    {
+      extraPackages = [ cfg.gitPackage ];
+      plugins.lazy.settings = setupOptions;
```
MattSturgeon, line 185: `setupOptions` can be inlined here

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
-        packedPlugins = if length plugins == 1 then head plugins else plugins;
-      in
-      mkIf (cfg.plugins != [ ]) ''
-        require('lazy').setup(
+          pluginSpecsList = pluginListToSpecs lazyPlugins;
+          packedPluginSpecs = if length pluginSpecsList == 1 then head pluginSpecsList else pluginSpecsList;
+        in
+        if length pluginSpecsList > 0 then
           {
             dev = {
```
MattSturgeon, line 173: Do we need to use `dev`? It feels a bit hacky.

Based on the [docs](https://lazy.folke.io/spec), I think we should just set `dir` and `dev` would be user-configured only if they want to _override_ our dir...

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,223 +1,187 @@
...
-    [ (mkEntryFromDrv plugin) ] ++ processDependencies;
+      lazyPluginType =
+        with types;
+        either package (submodule {
+          freeformType = types.attrsOf types.anything;
+          options = {
+            dir = helpers.mkNullOrOption str "A directory pointing to a local plugin";
+
+            pkg = mkOption {
+              type = package;
```
MattSturgeon, line 26: Should this be nullable, to allow users to install out-of-tree local plugins?

MattSturgeon: Actually, I think we can make this submodule a bit smarter:

```nix
types.coercedTo
  types.package
  (pkg: { inherit pkg; })
  (types.submodule (
    { config, ... }:
    {
      options = { dir; pkg; };

      config.dir = lib.mkIf (config.pkg != null) (
        lib.mkDefault config.pkg
      );
    }
  ))
```

tests/test-sources/plugins/pluginmanagers/lazy.nix
```diff
@@ -90,10 +98,73 @@
...
+    };
+  };
+
+  dir-only-plugin = {
+    plugins.lazy = with pkgs.vimPlugins; {
+      enable = true;
+      plugins = [ { dir = "${LazyVim}"; } ];
+    };
+  };
+
```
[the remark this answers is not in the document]
MattSturgeon: Such a test-case would be inherently impure, and therefore potentially flakey, so we could have one to test the module eval, but we should set `test.runNvim = false` to avoid actually running nvim and downloading the URL.

Since (in tests) nvim runs in a derivation builder, I don't think we'd be able to download anyway because I'm pretty sure buildscripts are sandboxed from the internet to enforce purity.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,166 @@
...
 
-              pkg = mkOption {
-                type = package;
-                description = "Vim plugin to install";
+      specs =
+        let
+          pluginToSpec =
+            plugin:
+            if isDerivation plugin then
+              { dir = "${lib.getName plugin}"; }
```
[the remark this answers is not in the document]
MattSturgeon: I wonder if we're experiencing something like https://github.com/NixOS/nixpkgs/issues/335016

I need to look into this when I have time to delve into it. Maybe ask around on some matrix rooms or discourse forums.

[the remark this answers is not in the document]
MattSturgeon: > I guess all that's left to get this PR in, would be to update the nixpkgs in the flake.lock file

Glad you got to the bottom of it, I was really scratching my head!

If so, we do that (fairly) regularly so feel free to do any other cleanup in the meantime. Let me know when you'd like another review.

You're welcome to bump the flake.lock temporarily in this PR (but we won't merge it like that, instead this can be revised once we've done a separate update PR).

Note that bumping nixpkgs currently has other issues elsewhere in nixvim.

plugins/pluginmanagers/lazy.nix
```diff
@@ -17,105 +17,113 @@ nixvim.neovim-plugin.mkNeovimPlugin {
...
+                with helpers.nixvimTypes;
+                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) ''
+                  opts should be a table (will be merged with parent specs),
+                  return a table (replaces parent specs) or should change a table.
+                  The table will be passed to the Plugin.config() function.
+                  Setting this value will imply Plugin.config()
+                '';
+            };
+
+            config.name = lib.mkIf (config.pkg != null) (lib.mkDefault "${lib.getName config.pkg}");
```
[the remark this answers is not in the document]
MattSturgeon: This is because of the type coercion.

`types.coercedTo` normalizes the final/merged option value to the "final" type, in this case that means any `types.package` definitions will be converted to `{ pkg = definition; }` in the final config value.

> Personally I prefer the non type coercion approach

Gotcha. Is that because of [`9d9dc58` (#2082)](https://github.com/nix-community/nixvim/pull/2082/commits/9d9dc581c00c6670686a41866d3fa3adaa57ca85)? Or because you don't like the idea of merging definitions in this way?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
 {
   lib,
   helpers,
-  config,
   pkgs,
   ...
 }:
 with lib;
```
khaneliman, line 7:
```suggestion

```

If we're in here migrating the plugin, we can clean up the `with lib;` usage, too.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
 {
   lib,
   helpers,
```
khaneliman, line 3:
```suggestion

```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
...
-  config,
   pkgs,
   ...
 }:
 with lib;
-let
-  cfg = config.plugins.lazy;
-  lazyPlugins = cfg.plugins;
+nixvim.neovim-plugin.mkNeovimPlugin {
```
khaneliman, line 8:
```suggestion
let
  inherit (lib.nixvim) defaultNullOpts mkNullOrOption;
  inherit (lib) types;
in
lib.nixvim.neovim-plugin.mkNeovimPlugin {
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
...
-            path = p.pkg;
-          };
-      processDependencies =
-        if plugin ? dependencies && plugin.dependencies != null then
-          builtins.concatMap processPlugin plugin.dependencies
-        else
-          [ ];
-    in
-    [ (mkEntryFromDrv plugin) ] ++ processDependencies;
+  maintainers = with maintainers; [
```
khaneliman, line 13:
```suggestion
  maintainers = with lib.maintainers; [
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
...
 
-      plugins =
-        with types;
-        let
-          pluginType = either package (submodule {
+          with types;
+          {
+            freeformType = attrsOf anything;
             options = {
               dir = helpers.mkNullOrOption str "A directory pointing to a local plugin";
```
khaneliman, line 28:
```suggestion
              dir = mkNullOrOption str "A directory pointing to a local plugin";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,61 +1,33 @@
...
 
-              pkg = mkOption {
-                type = package;
-                description = "Vim plugin to install";
-              };
+              pkg = helpers.mkNullOrOption package "Vim plugin to install";
```
khaneliman, line 30:
```suggestion
              pkg = mkNullOrOption package "Vim plugin to install";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -70,21 +42,25 @@ in
                 or when one of the lazy-loading handlers triggers
               '';
 
-              enabled = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`" ''
+              enabled = helpers.defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```
khaneliman, line 45:
```suggestion
              enabled = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -70,21 +42,25 @@ in
...
                 When false then this plugin will not be included in the spec. (accepts fun():boolean)
               '';
 
-              cond = helpers.defaultNullOpts.mkStrLuaFnOr types.bool "`true`" ''
+              cond = helpers.defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```
khaneliman, line 49:
```suggestion
              cond = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -70,21 +42,25 @@ in
...
                 then this plugin will not be loaded. Useful to disable some plugins in vscode,
                 or firenvim for example. (accepts fun(LazyPlugin):boolean)
               '';
 
-              dependencies = helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str listOfPlugins) "Plugin dependencies";
+              # Note: do not change the type of `dependencies` to
+              # `lazyPluginsListType` it must be `helpers.nixvimTypes.eitherRecursive
+              # str lazyPluginsListType`. While nixvim tests won't fail it can cause
+              # stack overflow errors when using nixvim in home-manager.
+              dependencies = helpers.mkNullOrOption (helpers.nixvimTypes.eitherRecursive str lazyPluginsListType) "Plugin dependencies";
```
khaneliman, line 59:
```suggestion
              dependencies = mkNullOrOption (types.eitherRecursive str lazyPluginsListType) "Plugin dependencies";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -70,21 +42,25 @@ in
...
 
               init = helpers.mkNullOrLuaFn "init functions are always executed during startup";
 
-              config = helpers.mkNullOrStrLuaFnOr (types.enum [ true ]) ''
+              config = helpers.mkNullOrStrLuaFnOr (enum [ true ]) ''
```
khaneliman, line 63:
```suggestion
              config = mkNullOrStrLuaFnOr (enum [ true ]) ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -142,81 +118,31 @@ in
...
-      '';
+      );
+
+      lazyPluginsListType = types.listOf lazyPluginType;
+    in
+    {
+      gitPackage = lib.mkPackageOption pkgs "git" { nullable = true; };
+      luarocksPackage = lib.mkPackageOption pkgs "luarocks" { nullable = true; };
+
+      plugins = mkOption {
```
khaneliman, line 134:
```suggestion
      plugins = lib.mkOption {
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                  See config.dev
+                '';
+
+                lazy = defaultNullOpts.mkBool true ''
+                  When true, the plugin will only be loaded when needed.
+                  Lazy-loaded plugins are automatically loaded when their Lua modules are required,
+                  or when one of the lazy-loading handlers triggers
+                '';
+
+                enabled = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```
khaneliman, line 78:
```suggestion
                enabled = defaultNullOpts.mkStrLuaFnOr bool true ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                  When false then this plugin will not be included in the spec. (accepts fun():boolean)
+                '';
+
+                cond = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
```
khaneliman, line 82:
```suggestion
                cond = defaultNullOpts.mkStrLuaFnOr bool true ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                '';
+
+                lazy = defaultNullOpts.mkBool true ''
+                  When true, the plugin will only be loaded when needed.
+                  Lazy-loaded plugins are automatically loaded when their Lua modules are required,
+                  or when one of the lazy-loading handlers triggers
+                '';
+
+                enabled = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
+                  When false then this plugin will not be included in the spec. (accepts fun():boolean)
```
khaneliman, line 79:
```suggestion
                  When false then this plugin will not be included in the spec.
```
The type will explain this

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                url = mkNullOrOption str "A custom git url where the plugin is hosted";
+
+                pkg = mkNullOrOption package "Vim plugin to install";
+
+                name = mkNullOrOption str "Name of the plugin to install";
+
+                dev = defaultNullOpts.mkBool false ''
+                  When true, a local plugin directory will be used instead.
+                  See config.dev
```
khaneliman, line 69: > See config.dev

What does this mean? Should we explain it more here or link to docs?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                '';
+
+                lazy = defaultNullOpts.mkBool true ''
+                  When true, the plugin will only be loaded when needed.
+                  Lazy-loaded plugins are automatically loaded when their Lua modules are required,
+                  or when one of the lazy-loading handlers triggers
+                '';
```
khaneliman, line 76:
```suggestion
                  When `true`, the plugin will only be loaded when needed.
                  
                  Lazy-loaded plugins are automatically loaded when their Lua modules are required,
                  or when one of the lazy-loading handlers triggers.
                '';
```

Do we want to explain which triggers these are ? I see there are multiple options so maybe a description at the top of the module would be good to explain how to use this?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                enabled = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
+                  When false then this plugin will not be included in the spec. (accepts fun():boolean)
+                '';
+
+                cond = defaultNullOpts.mkStrLuaFnOr bool "`true`" ''
+                  When false, or if the function returns false,
+                  then this plugin will not be loaded. Useful to disable some plugins in vscode,
+                  or firenvim for example. (accepts fun(LazyPlugin):boolean)
```
khaneliman, line 85:
```suggestion
                  When `false`, or if the function returns false,
                  then this plugin will not be loaded. 
                  
                  This can be used to disable some plugins in vscode or firenvim, for example.
```

Wonder if the last part should just be an option example.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                # `types.either` here, as using just `types.either` also leads
+                # to stack overflow. 
+                dependencies = mkNullOrOption (eitherRecursive lazyPluginType lazyPluginsListType) ''
+                  A list of plugin names or plugin specs that should be
+                  loaded when the plugin loads. Dependencies are always
+                  lazy-loaded unless specified otherwise. When specifying a
+                  name, make sure the plugin spec has been defined somewhere
+                  else. This can also be a single string such as for a short
+                    plugin url (See: https://lazy.folke.io/spec).
+                '';
```
khaneliman, line 103:
```suggestion
                dependencies = mkNullOrOption (eitherRecursive lazyPluginType lazyPluginsListType) ''
                  A list of plugin names or plugin specs that should be
                  loaded when the plugin loads.
                  
                  Dependencies are always lazy-loaded unless specified otherwise.
                  
                  When specifying a name, make sure the plugin spec has been defined somewhere else.
                  This can also be a single string such as for a short plugin URL. (See: https://lazy.folke.io/spec).
                '';
```

Should this text be broken up more and maybe give examples?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                init = mkNullOrLuaFn "init functions are always executed during startup";
```
khaneliman, line 105:
```suggestion
                init = mkNullOrLuaFn "Functions that are always executed during startup.";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                config = mkNullOrStrLuaFnOr (enum [ true ]) ''
+                  config is executed when the plugin loads.
+                  The default implementation will automatically run require(MAIN).setup(opts).
+                  Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name.
+                  See also opts. To use the default implementation without opts set config to true.
+                '';
```
khaneliman, line 112:
```suggestion
                  Config that is executed when the plugin loads.
                  
                  The default implementation will automatically run `require(MAIN).setup(opts)`.
                  Lazy uses several heuristics to determine the plugin's MAIN module automatically based on the plugin's name.
                  See also `opts`. To use the default implementation without `opts` set `config` to `true`.
                '';
```

What are main and opts here referring to? If it's a module option we can clarify with the path to the option. 
ie: `plugins.lazy.plugin.opts` or `plugins.lazy.plugin.config` ?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
-              opts =
-                with helpers.nixvimTypes;
-                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) ''
-                  opts should be a table (will be merged with parent specs),
-                  return a table (replaces parent specs) or should change a table.
-                  The table will be passed to the Plugin.config() function.
-                  Setting this value will imply Plugin.config()
+                main = mkNullOrOption str ''
+                  You can specify the main module to use for config() and opts(),
+                  in case it can not be determined automatically. See config()
```
khaneliman, line 116:
```suggestion
                main = mkNullOrOption str ''
                  You can specify the main module to use for `config()` and `opts()`,
                  in case it can not be determined automatically. See `config()`
```
Similar to above, is this a module option we can show the path to  ?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
-        in
-        mkOption {
-          type = listOfPlugins;
-          default = [ ];
-          description = "List of plugins";
+
+                submodules = defaultNullOpts.mkBool true ''
+                  When false, git submodules will not be fetched.
+                  Defaults to true
+                '';
```
khaneliman, line 122:
```suggestion
                submodules = defaultNullOpts.mkBool true ''
                  When `false`, git submodules will not be fetched.
                '';
```

Defaults will be shown in docs for us based on the option definition above.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                event =
+
```
khaneliman, line 125:
```suggestion
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                  mkNullOrOption (maybeRaw (
+                    either str (listOf str)
+                  )) "Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua";
+
+                cmd = mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on command";
+
+                ft = mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on filetype";
+
+                keys = mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on key mapping";
```
khaneliman, line 134: These could all do with examples, probably..

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+
+                module = mkNullOrOption (enum [ false ]) ''
+                  Do not automatically load this Lua module when it's required somewhere
```
khaneliman, line 137:
```suggestion
                  Do not automatically load this Lua module when it's required somewhere.
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                '';
+
+                priority = mkNullOrOption number ''
+                  Only useful for start plugins (lazy=false) to force loading certain plugins first.
+                  Default priority is 50. It's recommended to set this to a high number for colorschemes.
```
khaneliman, line 142:
```suggestion
                  Only useful for start plugins (lazy=false) to force loading certain plugins first.
                  
                  Default priority is 50. It's recommended to set this to a high number for colorschemes.
```
We should associate the lazy=false with the actual module option and maybe set the pluginDefault

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                '';
+
+                optional = defaultNullOpts.mkBool false ''
+                  When a spec is tagged optional, it will only be included in the final spec,
+                  when the same plugin has been specified at least once somewhere else without optional.
+                  This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part
+                  of the user's plugins
```
khaneliman, line 149:
```suggestion
                  Optional specs are only included in the final configuration if the corresponding plugin 
                  is also specified as a required (non-optional) plugin elsewhere.

                  This feature is particularly helpful for Neovim distributions, 
                  allowing them to pre-configure settings for plugins that users may or may not have installed.
```

Maybe something like this?

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+                '';
+
+                opts =
+
+                  mkNullOrOption (maybeRaw (attrsOf anything)) ''
+                    opts should be a table (will be merged with parent specs),
+                    return a table (replaces parent specs) or should change a table.
+                    The table will be passed to the Plugin.config() function.
+                    Setting this value will imply Plugin.config()
+                  '';
```
khaneliman, line 159:
```suggestion
                  mkNullOrOption (maybeRaw (attrsOf anything)) ''
                    The opts value can be one of the following:

                    - A table: This table will be merged with any existing configuration settings from parent specifications.
                    - A function that returns a table: The returned table will completely replace any existing configuration settings from parent specifications.
                    - A function that modifies a table: This function will receive the existing configuration table as an argument and can modify it directly.

                     In all cases, the resulting configuration table will be passed to the Plugin.config() function.
                     Setting the opts value automatically implies that Plugin.config() will be called.
                  '';
```

I think this might make it more clear what is occurring? Maybe we should link to their docs, too?

khaneliman (commented):
I'm sure @MattSturgeon will have better recommendations for wording of descriptions

plugins/pluginmanagers/lazy.nix
khaneliman: Yeah, figured it was just a copypasta from their source/docs when I was comparing default values from there :P

khaneliman:
> > > So I was able to simplify the `pluginToLua` function, I'll add that as a temporary commit to this PR, and if it's deemed as appropriate for this PR itself I can squash it but if not I'll move it into it's own separate PR
> > 
> > 
> > Looks much better. I think it makes sense as part of this PR.
> > I only really went over this part of the PR in this review, but I'll try and find time to fully review soon as everything seems to be coming together 😀
> 
> Thanks for taking the time to look over it!

Thanks for taking this module on, it looks like a beast of one!

plugins/pluginmanagers/lazy.nix
khaneliman: Alright, looking at those docs makes it more clear what it's doing. 

```suggestion
                  This is the directory configured at `plugins.lazy.settings.dev.path`.
```

plugins/pluginmanagers/lazy.nix
khaneliman: Sorry, I wasn't very clear. I see that this is basically a core concept of the plugin and how you can control when things get lazy loaded through the various handlers we have options for. You can define a plugin level description with useful documentation like we do in [treesitter](https://github.com/nix-community/nixvim/blob/facf6b2d0c9e22d858956d1d458eac6baf155a08/plugins/by-name/treesitter/default.nix#L15) and some others. 

We might also benefit by saying a short form of this description which settings those are: 
```suggestion
                  When true, the plugin will only be loaded when needed.
                  Lazy-loaded plugins are automatically loaded when their Lua modules are required,
                  or when one of the lazy-loading handlers are triggered.
                  
                  You can define which triggers load this plugin using `plugins.lazy.plugins.<plugin>.[event|cmd|ft|keys]`.
                '';
```

plugins/pluginmanagers/lazy.nix
khaneliman: I was trying to say I like it when our documentation references our actual module options verbiage that a user will be configuring the plugin with, like we do [here](https://github.com/nix-community/nixvim/blob/facf6b2d0c9e22d858956d1d458eac6baf155a08/plugins/cmp/default.nix#L20-L24). Instead of just saying `lazy=false` it's nice to point out what the actual config setting is the user should use, ie: `plugins.lazy.plugins.<plugin>.lazy = false`. Maybe it's too explicit, some usages we have just use the name of the setting itself. 

The default would be nice to be a part of the option definition, though. Instead of hiding it in the description, we can use `defaultNullOpts.mkInt 50 <description>` to set a pluginDefault that gets shown in the documentation.

plugins/pluginmanagers/lazy.nix
MattSturgeon: That sounds like lazy.nvim is telling us this option can be used to configure plugins that **may or may not** be loaded.

It is saying that a different part of lazy's configuration is responsible for saying what/when plugins are loaded.

MattSturgeon:
How's this going functionally? Are you still running into issues getting the correct store paths assigned to plugin `dir`? Do you need help with anything?

plugins/pluginmanagers/lazy.nix
MattSturgeon: Re: `gw` and `gq`, I generally wrap manually. I usually try to wrap semantically; preferring line breaks at punctuation or the end of a phrase if possible. Unless it looks too silly and unbalanced.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,204 @@
...
+    };
+
+  extraConfig =
+    cfg:
+    let
+      # The `pkg` property isn't a part of the `lazy.nvim` plugin spec, while
+      # it shouldn't do any harm but it does take up unecessary space in the
+      # init.lua file. Since we're done using it we will strip it from the
+      # final list of specs.
+      removePkgAttrFromPlugin =
```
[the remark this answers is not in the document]
MattSturgeon: Have you double checked `nixvim-print-init`? By default `toLuaObject` filters out any null or empty items.

Yes, this is a horrible hack, but it's currently the only reasonable way we have to filter out null sub-options from freeform submodules...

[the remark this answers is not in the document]
MattSturgeon: I'll have a look at the actual code a bit later, but it sounds like we need a `builtins.removeAttrs plugin [ "pkg" ]` somewhere when we convert `cfg.plugins` -> `settings.spec`.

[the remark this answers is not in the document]
MattSturgeon: My view is (unlike `settings`) `plugins` isn't aiming to be a 1-to-1 mapping of the upstream config format.

Instead it is an alternative (nicer) way to define `settings.spec`, whatever tranformations need to be done to it should be done when defining `plugins.lazy.settings.spec`.

This means users will have a predictable value at `config.plugins.lazy.plugins` if they want to read that in their own modules for any reason.

### Dumb exercise in what's possible

To be clear, I'm **not** recommend you use `apply` here.

But for the sake argument, if you did want to transform the value at `config.plugins.plugins` to be different in some way from what users define, this could be done with either a custom option-type's `merge` function (this is how `coercedTo` works) or defining an `apply` function when calling `mkOption`.

`apply` acts as a final transformation on the fully merged option value, so if I wanted to I could do something really stupid like:

```nix
{ config, lib, ... }:
{
  # Declare foo as a boolean option, that converts its final value to a stupid string...
  options.foo = lib.mkOption {
    type = types.bool;
    apply = v: "was defined as ${lib.boolToString v}!";
  };

  config = {
    # Define foo to be true
    foo = true;

    assertions = [
      {
        # NOTE: while `foo` is defined as a bool, its value is a string!
        # DO NOT DO THIS lol
        assertion = config.foo == "was defined as true!";
        message = "apply failed";
      }
    ];
  };
}
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,288 @@
...
+                    # Custom name for vim plugin
+                    "oil"
+
+                    # The plugin gets defined by using the custom name that was
+                    # previously given.
+                    {
+                      name = "oil";
+                      pkg = oil-nvim;
+                    }
+                  ];
```
MattSturgeon, line 95: > Exception: Failed to render option `plugins.lazy.plugins.*.__unkeyed`

I suspect that's because this isn't in a markdown codeblock

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,288 @@
...
+                '';
+
+                keys = mkNullOrOption (maybeRaw (either str (listOf str))) "Lazy-load on key mapping";
+
+                module = mkNullOrOption (enum [ false ]) ''
+                  Do not automatically load this Lua module when it's required somewhere.
+                '';
+
+                priority = mkNullOrOption number ''
```
MattSturgeon: We should usually prefer the defaultNullHelpers for options that'll (eventually) be going through `toLuaObject`.

This is also how we usually document "plugin defaults"

```suggestion
                priority = defaultNullOpts.mkNum 50 ''
```

If you need more control, the following is equivalent:

```nix
priority = defaultNullOpts.mkNum' {
  pluginDefault = 50;
  description = ''
    Only useful for start plugins i.e.
    `plugins.lazy.plugins.<plugin>.lazy = false` to force loading
    certain plugins first.

    It is recommended to set this to a high number for colorschemes.
  '';
}
```

MattSturgeon: defaultNullOpts will usually wrap `maybeRaw` support around the type, allowing users to configure either the declared type _or_ `__raw` lua code.

`pluginDefault` is actually optional, if we don't know one. But it is required when using the "short form" version of the helper functions (the one without the `'`).

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,288 @@
...
+                optional = defaultNullOpts.mkBool false ''
+                  Optional specs are only included in the final configuration
+                  if the corresponding plugin is also specified as a required
+                  (non-optional) plugin elsewhere. This feature is
+                  particularly helpful for Neovim distributions, allowing
+                  them to pre-configure settings for plugins that users may
+                  or may not have installed.
+                '';
+
+                opts = mkNullOrOption (maybeRaw (attrsOf anything)) ''
```
MattSturgeon, line 217:
```suggestion
                opts = defaultNullOpts.mkAttrsOf anything { } ''
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,288 @@
...
+
+                  In all cases, the resulting configuration table will be
+                  passed to the Plugin.config() function. Setting the opts
+                  value automatically implies that Plugin.config() will be
+                  called. (See: https://lazy.folke.io/spec#spec-setup)
+                '';
+              };
+
+              config.name = lib.mkIf (cfg.pkg != null) (lib.mkDefault "${lib.getName cfg.pkg}");
+              config.dir = lib.mkIf (cfg.pkg != null) (lib.mkDefault "${cfg.pkg}");
```
MattSturgeon, line 237:
```suggestion
              config = lib.mkIf (cfg.pkg != null) {
                name = lib.mkDefault "${lib.getName cfg.pkg}";
                dir = lib.mkDefault "${cfg.pkg}";
              };
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,284 @@
...
+
+                # WARNING: Be very careful if changing the type of
+                # `dependencies`. Choosing the wrong type may cause a stack
+                # overflow due to infinite recursion, and it's very possible
+                # that the test cases won't catch this problem. To be safe,
+                # perform thorough manual testing if you do change the type of
+                # `dependencies`. Also use `types.eitherRecursive` instead of
+                # `types.either` here, as using just `types.either` also leads
+                # to stack overflow. 
+                dependencies = mkNullOrOption (eitherRecursive lazyPluginType lazyPluginsListType) ''
```
MattSturgeon, line 147: This is concerning... In particular:

> Also use `types.eitherRecursive` instead of `types.either` here, as using just `types.either` also leads to stack overflow.

`eitherRecursive` and `either` are usually the same thing; the former only exists for documentation purposes (and should probably be replaced by checking `config.isDocs`).

Essentially, `either` is normally the same as `eitherRecursive`. When evaluating options for documentation we recurse into most types that have sub-options. If eitherRecursive is used, we _don't_ recurse in the docs.

If `eitherRecursive` has a different behaviour to `either` outside of our docs, then something is very broken.

MattSturgeon: I would personally prefer to implement this something like:

```nix
dependenciesType =
  if config.isDocs then
    # Use a stub type for documentation purposes
    # We don't need to repeat all the plugin-type sub-options again in the docs
    # It'd also be infinitely recursive
    lib.mkOptionType {
      description = "plugin submodule";
      descriptionClass = "noun";
      check = throw "should not be used";
      merge = throw "should not be used";
    }
  else
    lazyPluginType;
in
  dependencies = mkOption {
    type = types.coercedTo dependenciesType lib.toList (listOf dependenciesType);
  };
```

plugins/pluginmanagers/lazy.nix
MattSturgeon: (When it's a custom option that doesn't map directly to upstream, then we maybe don't want `defaultNullOpts`)

plugins/pluginmanagers/lazy.nix
MattSturgeon: Just a reminder: `settings` is a freeform option. This means we don't need to declare sub-options for _everything_ that can be configured.

Anything config definitions that don't match an option we've declared will use the freeformType `attrsOf anything`.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,278 @@
...
+  name = "lazy";
+  originalName = "lazy.nvim";
+  package = "lazy-nvim";
+
+  maintainers = with lib.maintainers; [
+    MattSturgeon
+    my7h3le
+  ];
+
+  settingsOptions = with types; {
```
MattSturgeon, line 21: Just a reminder: `settings` is a freeform option. This means we don't need to declare sub-options for _everything_ that can be configured.

Anything config definitions that don't match an option we've declared will use the freeformType `attrsOf anything`.

plugins/pluginmanagers/lazy.nix
MattSturgeon: TBH I don't think we even need separate `pkg` and `dir` options...

Couldn't we have a `source` option that is `either path package`?

We could still have a default for name, e.g:
```nix
config.name = mkOptionDefault (
  if isDerivation config.source then
    getName config.source
  else
    baseNameOf config.source
);
```

I'm also debating `mkDefault` vs `mkOptionDefault` here. The latter is lower priority (higher priority number); its the same as a default passed directly to `mkOption`.

plugins/pluginmanagers/lazy.nix
MattSturgeon: > I can totally remove them though if it's deemed better to avoid having them 🙂

It's always a difficult judgement call. Let's revisit this towards the end of the review process, unless anyone already has a strong opinion.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
+                      name = "oil";
+                      pkg = oil-nvim;
+                    }
+                  ];
+                  ```
+                '';
+
+                dir = mkNullOrOption str ''
+                  A directory pointing to a local plugin path e.g. "~/plugins/trouble.nvim".
+                '';
```
MattSturgeon, line 84: I'm confused. Both `dir` and `__unkeyed` say they can be the path to a directory where the plugin is found... Are they the same thing?

MattSturgeon: Ah I see, you had `url` `pkg` and `dir` all mapping internally to `__unkeyed` if defined?

In this case I think we just remove the `__unkeyed` option from this submodule and do that kinda conversion later, when defining `settings.spec`.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
-              '';
-
-              opts =
-                with helpers.nixvimTypes;
-                helpers.mkNullOrOption (maybeRaw (attrsOf anything)) ''
-                  opts should be a table (will be merged with parent specs),
-                  return a table (replaces parent specs) or should change a table.
-                  The table will be passed to the Plugin.config() function.
-                  Setting this value will imply Plugin.config()
+                pkg = mkNullOrOption package "Vim plugin to install";
```
MattSturgeon, line 88: I think we can combine the various "source" options into a single one. Or at the very least fewer.

First let's double-check if `types.path` is in fact a superset of `types.package`, if so that simplifies things. Especially if the un-stringified derivation attrs survive the type merging process.

We can also use a "url" type if needed, `types.strMatching "https?://.*"` for example.

I'm imaging something like:

```nix
source = mkOption {
  type = with types; oneOf [
    path
    package
    (strMatching "https?://.*")
  ];
}
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
+                '';
+              };
+
+              config.name = lib.mkIf (cfg.pkg != null) (lib.mkDefault "${lib.getName cfg.pkg}");
+              config.dir = lib.mkIf (cfg.pkg != null) (lib.mkDefault "${cfg.pkg}");
+            }
+          )
+        );
+
+      lazyPluginsListType = types.listOf lazyPluginType;
```
MattSturgeon, line 228:
```suggestion
```

This can just be defined inline as e.g. `type = listOf lazyPluginType`

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
+    {
+      gitPackage = lib.mkPackageOption pkgs "git" { nullable = true; };
+      luarocksPackage = lib.mkPackageOption pkgs "luarocks" { nullable = true; };
+
+      plugins = lib.mkOption {
+        type = lazyPluginsListType;
+        default = [ ];
+        description = "List of plugins";
+      };
+    };
```
MattSturgeon, line 239: This option would benefit greatly from an `example` and also a fleshed out `description`.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
         };
+
+      removePkgAttrFromPlugins = plugins: map removePkgAttrFromPlugin plugins;
+    in
+    {
+      extraPackages = [
+        cfg.gitPackage
+        cfg.luarocksPackage
+      ];
+      plugins.lazy.settings.spec = removePkgAttrFromPlugins cfg.plugins;
```
MattSturgeon, line 262:
```suggestion

      plugins.lazy.settings.spec = map toPluginSpec cfg.plugins;
```

Let's inline `removePkgAttrFromPlugins` here and rename `removePkgAttrFromPlugin` to `toPluginSpec`.

MattSturgeon (commented):
Apologies if I seem a bit confused and/or inconsistent in this review. This is quite a complex plugin (and refactor) to wrap my head around!

plugins/pluginmanagers/lazy.nix
MattSturgeon: After going over everything again, I don't think there were as many settings options as I thought. With this dropped, there's currently zero.

I'm leaning towards including at least some of them :saluting_face:

plugins/pluginmanagers/lazy.nix
MattSturgeon: I'm open to discussion on this. Is there benefit to having multiple? Or do they map to several options in the upstream plugin spec?

plugins/pluginmanagers/lazy.nix
MattSturgeon: Gotcha. So these options all map to upstream options, but there's some redundancy?

I think we need to better understand how upstream handles these before finalizing a design here.

As for `__unkeyed` I feel like it can be named something better if this is an option we're intentionally doing transformations to, since our usual rules of "pass exactly what the user configured in `settings` to `setup`" don't really apply here. :thinking:

plugins/pluginmanagers/lazy.nix
MattSturgeon: > package is a top-level store path (/nix/store/hash-name)

i.e. its a subset of pathInStore:
> path in the Nix store

which itself is a subset of path:
> isStringLike && firstChar == "/"

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
-{
-  options = {
-    plugins.lazy = {
-      enable = mkEnableOption "lazy.nvim";
+          plugin;
 
-      gitPackage = lib.mkPackageOption pkgs "git" {
-        nullable = true;
-      };
+      lazyPluginCoercibleType = with types; either str package;
```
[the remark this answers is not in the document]
MattSturgeon: > Removing it being a name does make sense though as having a name isn't entirely useful as other properties for the plugin would have to be defined later in the list with an attribute set

Good observation. I agree it makes no sense to coerce a name string to an attrset if a name definition without any other config is effectively useless. If a user _really_ wanted to, they could still do it manually with `settings.spec` or `plugins.*.name` (if we have/keep that).

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,222 +1,264 @@
...
-              };
+        types.coercedTo lazyPluginCoercibleType coerceToLazyPluginSpec (
+          submodule (
+            { config, ... }:
+
+            let
+              cfg = config;
+            in
+            {
+              freeformType = attrsOf anything;
```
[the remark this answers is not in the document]
MattSturgeon: What feels weird to me is that this is simultaneously a freeform type, where users expect their definitions to be used verbatim, and also a wrapper abstraction on top of `settings.spec`.

The two seem contradictory to me.

We could go "old-school" and have an `extraOptions` option, but that feels like a bad solution...

It'll _probably_ be fine if we just document the transformation clearly, but it still feels off.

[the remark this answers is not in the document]
MattSturgeon: Yes, if we declare `options.extraOptions = mkOption { type = attrsOf anything; }` instead of a freeformType.

As I said, it's a bit old school...

plugins/pluginmanagers/lazy.nix
MattSturgeon: Looks like you're using the _submodule's_ `config` which won't have `isDocs`.

The `isDocs` option is declared in the `programs.nixvim` submodule, so it won't exist in your submodule.

plugins/pluginmanagers/lazy.nix
khaneliman: I always prefer having a decent amount of settings included to make it obvious what functionality can be customized without going to upstream docs. But, some plugins have too many that it becomes just pure tech debt to support all the settingsOptions.

plugins/pluginmanagers/lazy.nix
MattSturgeon: That sounds like a valid reason to have it accept any string. Makes me wonder if "source" is an appropriate name though, if it can also be a "plugin name"?

If we like the idea of a type-checked option named source, we could also have a "name" option. Bearing in mind it doesn't necessarily have to correspond 1-to-1 with upstream's plugin spec, since we can do conversion when assigning to `settings.spec`.

No strong preferences here, just exploring possible options :slightly_smiling_face:

Mikilio:
I noticed that the `build` spec is missing. Is there a reason for that? Should I instead use nix override for that? Couldn't this module do it instead?

MattSturgeon:
> When running tests using the dev shell I can't apply the --show-trace flag

You should be able to run e.g.:

```sh
tests plugins-by-name-lazy -- --show-trace
```

The `tests` command uses the fairly common pattern of `--` separating parsed args on the left and literal args passed through to some underlying program on the right.

Some programs also use `--` to explicitly separate args from filenames, e.g. `git log -- some-file txt`

plugins/pluginmanagers/lazy.nix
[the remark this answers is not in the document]
MattSturgeon: On reflection, I think the freeformType is fine, so long as we don't end up shadowing any attr names with custom options that prevent the user from defining plugin-options.

We also need to take care to `removeAttrs` all our custom options before serialising to lua.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,42 +1,41 @@
...
+  inherit (config) isDocs;
 in
-{
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lazy";
+  packPathName = "lazy.nvim";
+  package = "lazy-nvim";
+
+  # TODO: remove this
+  deprecateExtraOptions = true;
```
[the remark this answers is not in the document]
MattSturgeon: This implements a rename warning for a legacy option called `extraOptions` which is now replaced by the freeform `settings` option.

Setting this to true causes `mkNeovimPlugin` to add an alias: `mkRenamedOptionModule ["plugins" "lazy" "extraOptions"] ["plugins" "lazy" "settings"]`

We've probably had the alias long enough that users should've migrated though. See https://github.com/nix-community/nixvim/issues/3181

plugins/pluginmanagers/lazy.nix
```diff
@@ -45,184 +44,358 @@ in
...
+            # value instead of an extrapolated default.
+            else if (key == "dir" && value != null) then
+              { dir = value; }
+            # If the user explicitly sets the 'url' option, use the provided
+            # value instead of an extrapolated default.
+            else if (key == "url" && value != null) then
+              { url = value; }
+            # If the 'keys' option is given, we need to explicitly convert that into a form that `lazy.nvim` expects.
+            else if (key == "keys" && value != null) then
+              { keys = map convertToLazyKeyMappingType value; }
```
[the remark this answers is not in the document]
MattSturgeon: I think inlining it would result in a lot of logic nested within a nested if-else branch, so it's probably better as you have it now.

plugins/pluginmanagers/lazy.nix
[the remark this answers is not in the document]
MattSturgeon: If it was added at the same time as the other rename warnings, it should stay for a while.

They can all be moved to a separate file though if preferred.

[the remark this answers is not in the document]
MattSturgeon: Looks like the plugin doesn't currently have an `extraOptions` option, and never did, so yeah this isn't needed 👍

GaetanLepage:
I ended up opening a simpler PR: #3835.
I will thus close this one. Thanks a lot for taking the time to give this non-trivial task a try :)

my7h3le:
@GaetanLepage thanks again and my bad! I just haven't been able to work on it since

PR #2097, opened 2024-08-27
tests: test the tests fail when expected

MattSturgeon:
Adds a regression test for #2076. This test ensures that `extraConfigLua` is used in `finalPackage` and that the test will fail correctly when running `nvim` results in unexpected output.

I was hoping to also test more explicitly that an eval failure would be triggered when `extraConfigLua` is set to something invalid, but verifying the `"Hello, world!"` string makes it into the build log should cover most of those cases too.

At the very least this regression test ensures:
- `extraConfigLua` is used in the `init.lua` used to run `nvim` in the test
- `test.runNvim` correctly fails when neovim prints unexpected stdout

I used [`pkgs.testers.testBuildFailure`](https://nixos.org/manual/nixpkgs/unstable/#tester-testBuildFailure) as a helper, to aid in expecting a derivation build to fail.

### DIFF
--- a/flake-modules/tests.nix
+++ b/flake-modules/tests.nix
@@ -25,6 +25,11 @@
           inherit (self.lib.${system}.check) mkTestDerivationFromNixvimModule;
         };
 
+        failing-tests = import ../tests/failing-tests.nix {
+          inherit pkgs;
+          inherit (self.lib.${system}.check) mkTestDerivationFromNixvimModule;
+        };
+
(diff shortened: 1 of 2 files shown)


khaneliman (approved):
LGTM. I see the failure being reported in a nix flake check when I undo the previous fix. I see it succeed nix flake check with the current state.

traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2098, opened 2024-08-27
plugins/which-key: fix icon examples

khaneliman:
Which-key implemented proper icon support, utilize the icon attribute to properly override/set icons so that alignment doesn't get thrown off and it's inherited with child mappings did something similar in my own config to fix the weird rendering https://github.com/khaneliman/khanelivim/commit/13a38e596f092c2afbfa34d6e4f1a7b05137733d


traxys (approved):

khaneliman:
@mergifyio queue

PR #2100, opened 2024-08-27
tests: add regression test for warnings + assertions

MattSturgeon:
Follow up to #2097, #2076, & #2048.

Expand "failing-tests" to also ensure warnings & assertions are correctly handled by `mkTestDerivationFromNixvimModule`.

Also did some minor cleanup:
- Call `failing-tests.nix` using `pkgs.callPackage`
- Replace repetive use of `testBuildFailure` with a wrapper
  `mkFailingNixvimTest`

You can run just the "failing" tests by using:
```shell
nix build .#checks.x86_64-linux.failing-tests

# Or for more detail:
nix build --show-trace --print-build-logs --verbose \
  .#checks.x86_64-linux.failing-tests
```

You can also use the `entries` passthu to access a specific sub-test:
```shell
nix build .#checks.x86_64-linux.failing-tests.entries.fail-on-assertions
```
_(tab-completion can help here)_

See also, [`linkFarmFromDrvs`](https://github.com/NixOS/nixpkgs/blob/0fefe244b14d6f140d98dff6ddeb648174fb29f5/pkgs/build-support/trivial-builders/default.nix#L545-L569)

PR #2101, opened 2024-08-28
tests: simplify fetch-tests slightly

MattSturgeon:
Simplify by reducing the number of transformations done to the test-files' test-case modules attr.

Since `pkgs.linkFarm` can accept _either_ a list or an attrset, we don't need to transform the attrset into a list.

This has no functional change, and only slightly simplifies some internals. It may end up being redundant if we re-structure the tests again, but I spotted the slight over-complication while working on #2100 so figured I'd make a PR.


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2104, opened 2024-08-28
flake: defragment lib+wrappers, also cleanup tests

MattSturgeon:
- **lib: include standalone wrapper in top-level `lib`**
- **tests: use callPackage pattern for flake checks**


lib/default.nix
```diff
@@ -1,12 +1,16 @@
...
-{
+lib.fix (self: {
   # Add all exported modules here
   check = import ./tests.nix { inherit lib pkgs; };
   helpers = import ./helpers.nix (args // { inherit _nixvimTests; });
-}
+
+  # TODO: Consider renaming these?
+  makeNixvimWithModule = import ../wrappers/standalone.nix pkgs flake;
+  makeNixvim = module: self.makeNixvimWithModule { inherit module; };
```
khaneliman, line 15: what were you thinking for these names?

khaneliman (commented):
Overall, it feels cleaner and easier to follow the dependency chain of setting up checks. I like not having all the boilerplate for one liners in a few files and searching for the definition. 

Looks like something isn't setup right though according to buildbot.

lib/default.nix
MattSturgeon: `mkNixvim` and `mkNixvimWith`

- it seems we usually prefer `mk` over `make`, although it's not consistent 
- the `With` suffix is conventionally used in nixpkgs for a more powerful function variant that takes structured attr args
- `WithModule` is misleading, because both functions take a module

khaneliman: >     * `WithModule` is misleading, because both functions take a module

This is what was throwing me when looking at their definitions, as well...

MattSturgeon: While I'd like to rename these, I'm hesitant to do so yet as there's other changes I'd like to make and I think we should minimize user-facing refactoring where possible.

In particular, I'd like to get to a point where our `lib` (including the standalone wrapper) doesn't depend on `pkgs` or `system`. This matches the design of other similar flakes like nixos and home-manager, whose equivalent function (e.g. `nixpkgs.lib.nixosSystem`) is made available without being in the "perSystem" part of the flake outputs.

Achieving this goal may be fairly involved though and require a few breaking changes and ugly compromises, including re-imagining the standalone wrapper and it's function signature. That'd be the optimal time to rename it, I think.

PR #2105, opened 2024-08-28
plugins/nvim-highlight-colors: init

Tom-Hubrecht:


Tom-Hubrecht:
This adds a plugin for nvim-highlight-colors, it is a draft as the corresponding vimPlugin does not exist in nixpkgs and I do not intend to contribute it upstream.

refaelsh:
> the corresponding vimPlugin does not exist in nixpkgs

In that case, is that even possible to add such a plugin to `nixvim`?

refaelsh:
Also, I see that your plugin is indeed in nixpkgs: https://search.nixos.org/packages?channel=24.05&show=vimPlugins.nvim-highlight-colors&from=0&size=50&sort=relevance&type=packages&query=nvim-highlight-colors.

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
+{
+  lib,
+  helpers,
```
MattSturgeon, line 3: `helpers` is deprecated, you can use `lib.nixvim` instead:
```suggestion
```

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+}:
+
+let
+  inherit (helpers.defaultNullOpts)
+    mkBool
+    mkEnum
+    mkListOf
+    mkStr
+    mkStr'
+    ;
```
MattSturgeon, line 16: To match the style used elsewhere in nixvim:
```suggestion
  inherit (lib.nixvim) defaultNullOpts;
```

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+in
+helpers.neovim-plugin.mkNeovimPlugin config {
```
MattSturgeon, line 18: We (very) recently dropped the `config` arg:
```suggestion
lib.nixvim.neovim-plugin.mkNeovimPlugin {
```

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+  name = "nvim-highlight-colors";
+  defaultPackage = pkgs.vimPlugins.nvim-highlight-colors;
```
MattSturgeon, line 20: Watch out for #2139, if it is merged first you'll have to do:

```suggestion
  package = "nvim-highlight-colors";
```

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+
+  maintainers = [ helpers.maintainers.thubrecht ];
```
MattSturgeon, line 22:
```suggestion
  maintainers = [ lib.maintainers.thubrecht ];
```

plugins/ui/nvim-highlight-colors.nix
MattSturgeon, line 1: Could you introduce a test file please, you can reference other tests in `tests/test-sources/plugins` and/or recently merged PRs.

Generally, we like to have an `empty` test case, a `defaults` test case and (ideally) an `example` test case.

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+
+  settingsOptions = {
```
MattSturgeon, line 24: Note: because `settings` is a freeform option, there is no need to declare sub-options for _every_ upstream plugin option. Users can define any config they like, so having too many options can just end up being a maintenance burden.

The judgement call is yours to make though, I won't block a PR for having "too many" settings options :grin:

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+    render = mkEnum [
+      "background"
+      "foreground"
+      "virtual"
+    ] "background" "The render style used.";
```
MattSturgeon, line 29:
```suggestion
    render = defaultNullOpts.mkEnumFirstDefault [
      "background"
      "foreground"
      "virtual"
    ] "The render style used.";
```

plugins/ui/nvim-highlight-colors.nix
```diff
@@ -0,0 +1,96 @@
...
+            };
+          };
+        })
+        [ ]
+        ''
+          A list of custom colors.
+        '';
+
+    exclude_filetypes = mkListOf lib.stypes.str [ ] "A list of filetypes to exclude from highlighting.";
+    exclude_buftypes = mkListOf lib.stypes.str [ ] "A list of buftypes to exclude from highlighting.";
```
MattSturgeon, line 80: Typo:

```suggestion
    exclude_filetypes = mkListOf lib.types.str [ ] "A list of filetypes to exclude from highlighting.";
    exclude_buftypes = mkListOf lib.types.str [ ] "A list of buftypes to exclude from highlighting.";
```

MattSturgeon (commented):
The CI failure seems to just be a typo. Otherwise this is more or less there, suggestions below are fairly minor. :grin:

Issue #2107, opened 2024-08-28
A flag to enable icon support in telescope


florentc:
The [`telescope.nvim`](https://github.com/nvim-telescope/telescope.nvim) plugin has support for icons through an optional dependeny: [`nvim-web-devicons`](https://github.com/nvim-tree/nvim-web-devicons).

In nixvim, there is no dedicated flag to enable this optional feature.

Depending on whether another plugin (e.g. `lualine`) that relies on `nvim-web-devicons` is enabled or not, the icon feature will be provided or not in telescope. If not, `pkgs.vimPlugins.nvim-web-devicons` may be added to `extraPlugins` manually.

I think it would be nice if there was a dedicated `plugin.telescope.enableIconSupport` flag to enable `nvim-web-devicons`.

khaneliman:
This is something that I've thought of recently and wasn't sure if it was the best practice to keep adding options to every plugin that might use icons. Some plugins are just automatically adding it as a dependency, some offer it as an option like 

```nix
iconsEnabled = mkOption {
        type = types.bool;
        description = "Toggle icon support. Installs nvim-web-devicons.";
        default = true;
      };
```
I think for now, I can add it as a quick option, though. But, it would be nice to have an easier to maintain solution.

MattSturgeon:
Nixvim doesn't have a well defined way to manage optional dependencies currently.

Most of the time (in newer code, anyway), we provide nullable `package` options, allowing users to a) set to null to forego installing and b) select a different version or overridden variant of the package to install instead.

Adding a `iconsPackage` option that defaults to `pkgs.vimPlugins.nvim-wev-devicons` but can be set to `null` if unwanted is probably the right approach here?

Issue #2004, opened 2024-08-10
Better `package` option helpers 


MattSturgeon:
I'd like to achieve two things here:

- [x] #1950 
- [x] lazy evaluation of the default

The first point should be achieved by providing the helper with a "pkgs path" that can be used with `getAttrByPath` and also `concatStringsSep "."` to provide `default` and `defaultText` respectively.

I'm not sure about the second point. Maybe we'd need a different type that is lazier, although this may break things like `mkIf`.

Or maybe we can do something like `either package (functionTo package)` and provide a lazy function that returns the package as the default?

MattSturgeon:
I wonder if the option default would be lazy if we didn't evaluate the default package's `meta.homepage`? Although, maybe that's only evaluated when rendering the docs?

MattSturgeon:
As I understand it, `meta.nixvimInfo.url` usually depends on evaluating the package, however this will only be evaluated when building the docs.

Therefore, the main blocker to making package-type option defaults "lazy" is _either_ not including the docs in the final build _or_ not having `meta.nixvimInfo.url` default to the `<pkg>.meta.homepage` and instead always set `url` to a string literal...

The only thing I'm confused about, is when including the man-docs, we include the package from our flake outputs. It's probably better practice to build the docs using the `pkgs` available in the module eval (see #2186), however using the flake output _should_ mean that mismatched pkgs channel errors are not possible, unless the flake inputs are overridden using `follows`...

MattSturgeon:
I wonder if its enough to _not_ avoid evaluating the package defaults, but instead allow them to fail gracefully.

We could probably just wrap the `package.default.meta.homepage` in `tryEval`.

We'd still want to check that plugins have a valid url on _our_ nixpkgs lock, but this could be done in a new test instead of during module eval.

PR #2108, opened 2024-08-29
plugins/treesitter: add iconsPackage

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2107


plugins/languages/treesitter/treesitter.nix
```diff
@@ -324,6 +324,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
       description = "Grammar packages to install";
     };
 
+    iconsPackage = helpers.mkPackageOption {
+      name = "nvim-web-devicons";
+      default = pkgs.vimPlugins.nvim-web-devicons;
```
MattSturgeon, line 329: Really must get around to implementing #2004...

plugins/languages/treesitter/treesitter.nix
```diff
@@ -411,9 +416,10 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
     extraFiles = mkIf cfg.nixvimInjections { "queries/nix/injections.scm".source = ./injections.scm; };
 
-    extraPlugins = mkIf (cfg.package != null) [
+    extraPlugins = mkIf (cfg.package != null || cfg.iconsPackage != null) [
```
MattSturgeon, line 419: This feels a bit messy. Could we use mkMerge here, or perhaps just concat optional lists?

Thinking about it perhaps we don't need to check `!= null` anyway, because our `extraPlugins` option already takes care of that for us... 🤔

MattSturgeon (commented):
LGTM, but the `extraPlugins` definition is a little messy with all the null-checking. Feels like reading Java 😅

plugins/languages/treesitter/treesitter.nix
khaneliman: Yeah, I think I started writing lib.optionals and couldn't remember if there was a reason for the null check.. i think it was before you added the null check into the mkNeovimPlugin...

khaneliman: So, iconsPackage accepts a null if someone doesn't wish to install the package. But, extraPlugins doesn't so we need to ignore setting it when it's null.

plugins/languages/treesitter/treesitter.nix
```diff
@@ -411,10 +416,12 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
-    extraPlugins = mkIf (cfg.package != null) [
-      (mkIf cfg.nixGrammars (cfg.package.withPlugins (_: cfg.grammarPackages)))
-      (mkIf (!cfg.nixGrammars) cfg.package)
-    ];
+    extraPlugins =
+      lib.optionals (cfg.iconsPackage != null) [
+        cfg.iconsPackage
+      ]
+      ++ lib.optionals cfg.nixGrammars [ (cfg.package.withPlugins (_: cfg.grammarPackages)) ]
+      ++ lib.optionals (!cfg.nixGrammars) [ cfg.package ];
```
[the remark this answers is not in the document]
MattSturgeon: It'd be nice if we still used mkIf/mkMerge/etc, but I think using `optional` and `if then else` ends up so much more readable that it's worth it?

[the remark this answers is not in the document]
khaneliman: > I think we still need a `cfg.package != null` check somewhere.

I didn't think we had to since we don't accept null pluginPackage, still.

MattSturgeon (approved):
Awesome, no issues from me. Thanks again!

plugins/languages/treesitter/treesitter.nix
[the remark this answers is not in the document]
khaneliman: > > > I think we still need a `cfg.package != null` check somewhere.
> > 
> > 
> > I didn't think we had to since we don't accept null pluginPackage, still.
> 
> That's precisely _why_ we need to check it here, to avoid accidentally adding a `[null]` definition.

It can't even evaluate that far if someone tried to provide a null cfg.package.

```bash
error: A definition for option `plugins.treesitter.package' is not of type `package'. Definition values:
       - In `<unknown-file>': null
```

khaneliman:
@mergifyio queue

stasjok:
Why is it added to treesitter? Treesitter doesn't use icons. I think it was meant to be added for telescope...

PR #2112, opened 2024-08-29
treewide: standardize iconsPackage

khaneliman:
Follow up to discussion https://github.com/nix-community/nixvim/issues/2107#issuecomment-2316603375 to standardize how we handle the icons dependencies. Setting a default `iconsPackage` for each module that has been adding it as an `extraPlugins` and deprecating options that used to be a boolean specifically for this behavior. 

I see a few occurrences for `gitPackage` outside of the scope of my `iconsPackage` change, I will address those in a follow up PR.

### DIFF
--- a/plugins/filetrees/neo-tree.nix
+++ b/plugins/filetrees/neo-tree.nix
@@ -57,6 +57,11 @@ in
 
       package = helpers.mkPluginPackageOption "neo-tree" pkgs.vimPlugins.neo-tree-nvim;
 
+      iconsPackage = helpers.mkPackageOption {
+        name = "nvim-web-devicons";
+        default = pkgs.vimPlugins.nvim-web-devicons;
+      };
+
       sources =
         helpers.defaultNullOpts.mkListOf types.str
           [
@@ -1116,10 +1121,9 @@ in
         // cfg.extraOptions;
     in
     mkIf cfg.enable {
-      extraPlugins = with pkgs.vimPlugins; [
+      extraPlugins = [
         cfg.package
-        nvim-web-devicons
-      ];
+      ] ++ lib.optional (cfg.iconsPackage != null) cfg.iconsPackage;
 
       extraConfigLua = ''
         require('neo-tree').setup(${helpers.toLuaObject setupOptions})
--- a/plugins/filetrees/nvim-tree.nix
+++ b/plugins/filetrees/nvim-tree.nix
@@ -41,6 +41,11 @@ in
 
     package = helpers.mkPluginPackageOption "nvim-tree" pkgs.vimPlugins.nvim-tree-lua;
 
+    iconsPackage = helpers.mkPackageOption {
+      name = "nvim-web-devicons";
+      default = pkgs.vimPlugins.nvim-web-devicons;
+    };
+
     disableNetrw = helpers.defaultNullOpts.mkBool false "Disable netrw";
 
     hijackNetrw = helpers.defaultNullOpts.mkBool true "Hijack netrw";
@@ -1154,10 +1159,9 @@ in
       '';
     in
     mkIf cfg.enable {
-      extraPlugins = with pkgs.vimPlugins; [
+      extraPlugins = [
         cfg.package
-        nvim-web-devicons
-      ];
+      ] ++ lib.optional (cfg.iconsPackage != null) cfg.iconsPackage;
 
       autoCmd =
         (optional autoOpenEnabled {
--- a/plugins/utils/alpha.nix
+++ b/plugins/utils/alpha.nix
@@ -2,6 +2,7 @@
   lib,
   helpers,
   config,
+  options,
   pkgs,
   ...
 }:
@@ -55,10 +56,16 @@ in
 
       package = helpers.mkPluginPackageOption "alpha-nvim" pkgs.vimPlugins.alpha-nvim;
 
+      # TODO: deprecated 2024-08-29 remove after 24.11
       iconsEnabled = mkOption {
         type = types.bool;
         description = "Toggle icon support. Installs nvim-web-devicons.";
-        default = true;
+        visible = false;
+      };
+
+      iconsPackage = helpers.mkPackageOption {
+        name = "nvim-web-devicons";
+        default = pkgs.vimPlugins.nvim-web-devicons;
       };
 
       theme = mkOption {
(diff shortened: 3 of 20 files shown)


plugins/filetrees/neo-tree.nix
```diff
@@ -57,6 +57,16 @@ in
 
       package = helpers.mkPluginPackageOption "neo-tree" pkgs.vimPlugins.neo-tree-nvim;
 
+      gitPackage = helpers.mkPackageOption {
+        name = "git";
+        default = pkgs.git;
+      };
```
traxys, line 63: This is from #2113 right ?

plugins/filetrees/nvim-tree.nix
```diff
@@ -41,6 +41,16 @@ in
 
     package = helpers.mkPluginPackageOption "nvim-tree" pkgs.vimPlugins.nvim-tree-lua;
 
+    gitPackage = helpers.mkPackageOption {
+      name = "git";
+      default = pkgs.git;
+    };
```
traxys, line 47: Same?

plugins/utils/alpha.nix
```diff
@@ -139,9 +146,24 @@ in
...
     mkIf cfg.enable {
-      extraPlugins = [ cfg.package ] ++ (optional cfg.iconsEnabled pkgs.vimPlugins.nvim-web-devicons);
+      # TODO: deprecated 2024-08-29 remove after 24.11
+      warnings = lib.mkIf opt.iconsEnabled.isDefined [
+        ''
+          nixvim (plugins.alpha):
+          The option definition `plugins.alpha.iconsEnabled' in ${showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+          You should use `plugins.alpha.iconsPackage' instead.
+        ''
+      ];
```
traxys, line 160: This does not trigger due to the `default` value? It's freeform types that are strange in that way then?

traxys (commented):
We could add this option definition somwhere in

plugins/filetrees/neo-tree.nix
khaneliman: I can move it into that one, but I didn't touch that plugin in that PR.

traxys:
lualine is not handled in this because it will be in  #2080 ?

khaneliman:
> lualine is not handled in this because it will be in #2080 ?

Yeah, I didn't do it because of that PR... but I theoretically could if we wanted to actually get all the changes in right away.

plugins/utils/alpha.nix
khaneliman: I removed the default for it so that it only triggers when someone has explicitly declared it in their configuration.

plugins/filetrees/nvim-tree.nix
khaneliman: Same as above, just didn't want to deal with git conflicts tbh. I can move them into the corresponding named PR, if you want.

plugins/filetrees/neo-tree.nix
traxys: I'd prefer to only have iconPackages in this PR, it's better for the git history

khaneliman: Alright, I'll suffle the changes around a bit.

traxys (approved):

plugins/utils/alpha.nix
traxys: Sorry I missed that in the diff --'

khaneliman:
@mergifyio queue

plugins/utils/alpha.nix
MattSturgeon, line 160: Is there a reason this can't just be a `mkRenamedOptionModule`?

khaneliman: Would that support different data types? We went from a boolean to a package

plugins/utils/alpha.nix
```diff
@@ -139,9 +146,24 @@ in
...
+
+      extraPlugins =
+        [ cfg.package ]
+        ++ lib.optional (
+          cfg.iconsPackage != null && (opt.iconsEnabled.isDefined && cfg.iconsEnabled)
```
MattSturgeon, line 165: This logic seems wrong; if `iconsEnabled` is not defined then the package will never be added (`cfg.iconsPackage != null && false` is false).

I think you want:

```nix
cfg.iconsPackage != null && (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
```

plugins/utils/fzf-lua.nix
```diff
@@ -103,28 +110,44 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
+        ''
+          nixvim (plugins.fzf-lua):
+          The option definition `plugins.fzf-lua.iconsEnabled' in ${showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+          You should use `plugins.fzf-lua.iconsPackage' instead.
+        ''
+      ];
 
-    extraPackages = optional (cfg.fzfPackage != null) cfg.fzfPackage;
+      extraPlugins = lib.mkIf (
+        cfg.iconsPackage != null && (opt.iconsEnabled.isDefined && cfg.iconsEnabled)
```
MattSturgeon, line 129: Same

plugins/utils/alpha.nix
MattSturgeon: Ah, one is a nullable package, the other is a Boolean enable option.

stasjok:
It seems too many plugins now include `iconsPackage` option. If a user wants to disable it, they need to set it to null for each plugin individually. All options are doing the same thing. There is also the `mini.icons` (from `mini-nvim`) plugin, which can be used as an alternative to `nvim-web-devicons`. It includes a special function that can mimic the nvim-web-devicons plugin, though it's not perfect, it can serve as a substitute. Given this, changing the icons package is not easy. Maybe this can be a global option instead?

khaneliman:
> It seems too many plugins now include `iconsPackage` option. If a user wants to disable it, they need to set it to null for each plugin individually. All options are doing the same thing. There is also the `mini.icons` (from `mini-nvim`) plugin, which can be used as an alternative to `nvim-web-devicons`. It includes a special function that can mimic the nvim-web-devicons plugin, though it's not perfect, it can serve as a substitute. Given this, changing the icons package is not easy. Maybe this can be a global option instead?

It replaced hard coded includes so it isn’t a new thing that we install it. I agree we might want to do something global, though..

PR #2113, opened 2024-08-29
treewide: standardize gitPackage

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/2112

PR #2114, opened 2024-08-29
treewide: use plugin package option

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/2113

### DIFF
--- a/plugins/by-name/fzf-lua/default.nix
+++ b/plugins/by-name/fzf-lua/default.nix
@@ -33,8 +33,6 @@ in
 helpers.neovim-plugin.mkNeovimPlugin {
   name = "fzf-lua";
 
-  extraPackages = [ pkgs.fzf ];
-
   maintainers = [ maintainers.GaetanLepage ];
 
   inherit settingsOptions settingsExample;
--- a/plugins/by-name/neoclip/default.nix
+++ b/plugins/by-name/neoclip/default.nix
@@ -1,7 +1,8 @@
 {
   lib,
+  config,
   helpers,
-  pkgs,
+  options,
   ...
 }:
 with lib;
@@ -235,8 +236,30 @@ helpers.neovim-plugin.mkNeovimPlugin {
   };
 
   extraConfig = cfg: {
-    extraPlugins = mkIf (
-      isBool cfg.settings.enable_persistent_history && cfg.settings.enable_persistent_history
-    ) [ pkgs.vimPlugins.sqlite-lua ];
+    # TODO: added 2024-09-14 remove after 24.11
+    plugins.sqlite-lua.enable = mkOverride 1490 true;
+    warnings =
+      optional
+        (
+          isBool cfg.settings.enable_persistent_history
+          && cfg.settings.enable_persistent_history
+          && options.plugins.sqlite-lua.enable.highestPrio == 1490
+        )
+        ''
+          Nixvim (plugins.neoclip) `sqlite-lua` automatic installation is deprecated.
+          Please use `plugins.sqlite-lua.enable`.
+        '';
+
+    assertions = [
+      {
+        assertion =
+          isBool cfg.settings.enable_persistent_history && cfg.settings.enable_persistent_history
+          -> config.plugins.sqlite-lua.enable;
+        message = ''
+          Nixvim (plugins.neoclip): The persistent history sqlite storage backend needs `sqlite-lua` to function as intended.
+          You can enable it by setting `plugins.sqlite-lua.enable` to `true`.
+        '';
+      }
+    ];
   };
 }
--- a/plugins/by-name/neorg/default.nix
+++ b/plugins/by-name/neorg/default.nix
@@ -20,6 +20,11 @@ with lib;
       ];
     };
 
+    neorgTelescopePackage = lib.mkPackageOption pkgs [
+      "vimPlugins"
+      "neorg-telescope"
+    ] { nullable = true; };
+
     lazyLoading = helpers.defaultNullOpts.mkBool false '''';
 
     logger =
@@ -148,12 +153,18 @@ with lib;
           Telescope support for neorg (`core.integrations.telescope`) is enabled but the
           telescope plugin is not.
         '')
+        ++ (optional (telescopeSupport && (cfg.neorgTelescopePackage == null)) ''
+          Telescope support for neorg (`core.integrations.telescope`) is enabled but the
+          `neorgTelescopePackage` package is not set.
+        '')
         ++ (optional ((hasAttr "core.defaults" cfg.modules) && (!config.plugins.treesitter.enable)) ''
           Neorg's `core.defaults` module is enabled but `plugins.treesitter` is not.
           Treesitter is required when using the `core.defaults`.
         '');
 
-      extraPlugins = [ cfg.package ] ++ (optional telescopeSupport pkgs.vimPlugins.neorg-telescope);
+      extraPlugins = [
+        cfg.package
+      ] ++ (optional (telescopeSupport && cfg.neorgTelescopePackage != null) cfg.neorgTelescopePackage);
 
       extraConfigLua = ''
         require('neorg').setup(${helpers.toLuaObject setupOptions})
--- a/plugins/by-name/nvim-jdtls/default.nix
+++ b/plugins/by-name/nvim-jdtls/default.nix
@@ -20,6 +20,10 @@ in
       ];
     };
 
+    jdtLanguageServerPackage = lib.mkPackageOption pkgs "jdt-language-server" {
+      nullable = true;
+    };
+
     cmd = helpers.mkNullOrOption (types.listOf types.str) ''
       The command that starts the language server.
 
@@ -94,21 +98,11 @@ in
   config =
     let
       cmd =
-        if (cfg.cmd == null) then
-          let
-            data =
-              if (cfg.data == null) then
-                throw ''
-                  You have to either set the 'plugins.nvim-jdtls.data' or the 'plugins.nvim-jdtls.cmd'
-                  option.
-                ''
-              else
-                cfg.data;
-          in
-          [ (lib.getExe pkgs.jdt-language-server) ]
+        if (cfg.cmd == null && cfg.jdtLanguageServerPackage != null) then
+          [ (lib.getExe cfg.jdtLanguageServerPackage) ]
           ++ [
             "-data"
-            data
+            cfg.data
           ]
           ++ (optionals (cfg.configuration != null) [
             "-configuration"
@@ -125,9 +119,23 @@ in
       } // cfg.extraOptions;
     in
     mkIf cfg.enable {
+      assertions = [
+        {
+          assertion = cfg.cmd != null || cfg.data != null;
+          message = "You have to either set the `plugins.nvim-jdtls.data` or the `plugins.nvim-jdtls.cmd` option.";
+        }
+        {
+          assertion = cfg.cmd == null -> cfg.jdtLanguageServerPackage != null;
+          message = ''
+            Nixvim (plugins.nvim-jdtls) You haven't defined a `cmd` or `jdtLanguageServerPackage`.
+
+            The default `cmd` requires `plugins.nvim-jdtls.jdtLanguageServerPackage` to be set.
+          '';
+        }
+      ];
       extraPlugins = [ cfg.package ];
 
-      extraPackages = [ pkgs.jdt-language-server ];
+      extraPackages = [ cfg.jdtLanguageServerPackage ];
 
       autoCmd = [
         {
--- a/plugins/by-name/octo/default.nix
+++ b/plugins/by-name/octo/default.nix
@@ -12,8 +12,6 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   maintainers = [ helpers.maintainers.svl ];
 
-  extraPackages = [ pkgs.gh ];
-
   settingsOptions = {
     use_local_fs = helpers.defaultNullOpts.mkBool false ''
       Use local files on right side of reviews.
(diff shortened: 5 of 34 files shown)


plugins/git/neogit/default.nix
```diff
@@ -102,6 +102,13 @@ helpers.neovim-plugin.mkNeovimPlugin config {
     };
   };
 
+  extraOptions = {
+    whichPackage = helpers.mkPackageOption {
```
khaneliman, line 106: > I wonder if for neogit we need a more freeform approach, e.g. `attrsOf (nullOr package)`.
> 
> Or at least a `whichPackage` option.
> 
> Saying that, is it even possible to do anything without the `which` command? Installing the dependency seems a little overkill. Maybe it's for cross-platform reasons 🤷

@MattSturgeon  I'm not familiar with why the dependency was added... i might have to look into it

MattSturgeon: I'm also unfamiliar, it was not present [before migrating](https://github.com/nix-community/nixvim/blob/82a24d0d422bcc24eb33ebe7aa2571a78f4eafdf/plugins/git/neogit.nix) to mkNeovimPlugin, so it was added in #971

AFACT, it does no harm. And could lead to better consistency on platforms that use a shell builtin or a different implementation of `which`?

cc @GaetanLepage as original implementor

MattSturgeon:
This will need to be re-done in line with #2150 (sorry)

plugins/by-name/openscad/default.nix
```diff
@@ -53,10 +69,9 @@ in
     in
     mkIf cfg.enable {
       extraPlugins =
-        with pkgs.vimPlugins;
         [ cfg.package ]
-        ++ (optional (fuzzyFinder == "skim") skim-vim)
-        ++ (optional (fuzzyFinder == "fzf") fzf-vim);
+        ++ (optional (fuzzyFinder == "skim" && cfg.skimPackage != null) cfg.skimPackage)
+        ++ (optional (fuzzyFinder == "fzf" && cfg.fzfPackage != null) cfg.fzfPackage);
```
MattSturgeon, line 74: I wonder if gating installation behind _both_ a nullable package option _and_ another option makes sense, especially elsewhere where we use `hasInfix` on a settings option... :thinking: 

Maybe we should just provide a `fyzzyFinder` package option and if it isn't null, install it?

I'm not sure what the current enum is used for elsewhere, but perhaps it could default to the package's `meta.mainProgram`?

plugins/by-name/neorg/default.nix
```diff
@@ -20,6 +20,14 @@ with lib;
       ];
     };
 
+    neorgTelescopePackage = lib.mkPackageOption pkgs "neorg-telescope" {
```
khaneliman, line 23: Moved to a plugin module?

plugins/by-name/yanky/default.nix
```diff
@@ -343,6 +343,14 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   extraOptions = {
     enableTelescope = mkEnableOption "the `yank_history` telescope picker.";
+
+    sqliteLuaPackage = lib.mkPackageOption pkgs "sqlite-lua" {
```
khaneliman, line 347: moved to a plugin module?

plugins/by-name/neoclip/default.nix
```diff
@@ -234,9 +234,21 @@ helpers.neovim-plugin.mkNeovimPlugin {
     };
   };
 
+  extraOptions = {
+    sqliteLuaPackage = lib.mkPackageOption pkgs "sqlite-lua" {
```
khaneliman, line 238: moved to a plugin module?

plugins/by-name/openscad/default.nix
khaneliman: I liked the idea and went through and implemented it... but, it looks like these are vimPlugins which don't get a meta.mainProgram. 
```bash
 error: attribute 'mainProgram' missing
       at /nix/store/2bmb1607gm9rfsm39lx6wdz45a9qdbzs-source/plugins/by-name/openscad/default.nix:68:35:
           67|         (mkIf (cfg.fuzzyFinderPackage != null) {
           68|           openscad_fuzzy_finder = cfg.fuzzyFinderPackage.meta.mainProgram;
           69|         })
```

plugins/by-name/neorg/default.nix
khaneliman: Eh, looks like it wouldn't really makes sense on its own. https://github.com/nvim-neorg/neorg-telescope?tab=readme-ov-file#-installation

plugins/by-name/openscad/default.nix
MattSturgeon: Thanks for trying. I'm guessing we also can't reliably use `lib.getName` either, since the package name may not match the exe name.

Maybe we could have something like:

```nix
fuzzyFinder = lib.mkOption {
  type = nullOr str; # or enum still
};
fuzzyFinderPackage =
  let
    defaults = {
      skim = "skim-vim";
      fzf = "fzf-vim";
    };
  in
    lib.mkOption {
      type = nullOr package;
      default =
        if defaults ? cfg.fuzzyFinder then
          getAttrByPath (toList defaults.${cfg.fuzzyFinder}) null pkgs
        else
          null;
      defaultText = literalMD ''
        # TODO
      '';
    };
```

IDK if there's a good way to generate the default text from the defaults attrs. I don't think it'd be easy to use `mkPackageOption` if we need such custom logic?

khaneliman: Yeah, that's what was sounding like too much of a hassle. It's either something we need to maintain the available options explicitly through multiple package options or through that lookup function, basically. Do you think we just rip the migration out of this package ?

plugins/by-name/yanky/default.nix
khaneliman: Created https://github.com/nix-community/nixvim/pull/2241 so we can just link to that module's package.

plugins/by-name/openscad/default.nix
khaneliman: I'm gonna pull it out of this PR, for now. This module is in need of a refactor and can be handled then.

plugins/by-name/fzf-lua/default.nix
```diff
@@ -33,8 +33,6 @@ in
 helpers.neovim-plugin.mkNeovimPlugin {
   name = "fzf-lua";
 
-  extraPackages = [ pkgs.fzf ];
```
MattSturgeon, line 36: Could you mention in the commit message why it isn't needed?

plugins/utils/rest.nix
```diff
@@ -14,8 +14,6 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   luaName = "rest-nvim";
   package = "rest-nvim";
 
-  extraPackages = [ pkgs.curl ];
```
MattSturgeon, line 17: Same

plugins/by-name/neoclip/default.nix
```diff
@@ -235,8 +235,16 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+      {
+        assertion =
+          (isBool cfg.settings.enable_persistent_history && cfg.settings.enable_persistent_history)
+          -> config.plugins.sqlite-lua.enable;
+        message = ''
+          Nixvim (plugins.neoclip): The persistent history sqlite storage backend needs `sqlite-lua` to function as intended.
+          You can enable it by setting `plugins.sqlite-lua.enable` to `true`.
+        '';
+      }
+    ];
```
MattSturgeon, line 248: This is technically a breaking change, since we're asking users to enable another plugin to pass the assertion.

I wonder about a transition period where we set `plugins.sqlite-lua.enable` with a low-priority[^1] definition, then show a warning if the option's `highestPrio` matches our definition.

[^1]: an option's default or `mkOptionDefault` would use prio 1500, so maybe 1400 would be suitable for our purposes?

plugins/by-name/nvim-jdtls/default.nix
```diff
@@ -125,9 +119,23 @@ in
...
+        }
+        {
+          assertion = cfg.cmd == null -> cfg.jdtLanguageServerPackage != null;
+          message = ''
+            Nixvim (plugins.nvim-jdtls) You haven't defined a `cmd` or `jdtLanguageServerPackage`.
+
+            The default `cmd` requires `plugins.nvim-jdtls.jdtLanguageServerPackage` to be set.
+          '';
+        }
+      ];
```
MattSturgeon, line 135: Same

plugins/by-name/yanky/default.nix
```diff
@@ -354,6 +354,13 @@ helpers.neovim-plugin.mkNeovimPlugin {
           Nixvim (plugins.yanky): The telescope integration needs telescope to function as intended
         '';
       }
+      {
+        assertion = cfg.settings.ring.storage == "sqlite" -> config.plugins.sqlite-lua.enable;
+        message = ''
+          Nixvim (plugins.yanky): The sqlite storage backend needs `sqlite-lua` to function as intended.
+          You can enable it by setting `plugins.sqlite-lua.enable` to `true`.
+        '';
+      }
```
MattSturgeon, line 363: Same

plugins/by-name/neoclip/default.nix
khaneliman: So you want it to be in between mkOptionDefault and mkDefault, in case someone sets their own mkDefault config? Were you thinking of setting this in each of the modules that are transitioning or in the sqlite-lua module?

MattSturgeon: The sqlite-lua module will declare the `enable` option with `default = false` (implicitly priority 1500).

Plugin modules that _require_ sqlite-lua will set `enable = mkOverride 1490 true`, which is equivalent to the current behavior of installing the package.

A per-plugin deprecation warning can then check `options.plugins.sqlite-lua.enable.highestPrio` (and the value too I guess).

plugins/by-name/nvim-jdtls/default.nix
khaneliman: This isn't a breaking change, we default it with a package still. We're just creating an assertion for people that disable that package.

khaneliman: Unless I misunderstood lib.mkPackageOption, that is. I thought it defaulted to the pkg name you specify.

MattSturgeon: My mistake. I spotted the assertion but didn't look close enough at what it was doing!

khaneliman: All good, I was real confused for a minute :P

plugins/by-name/neorg/default.nix
```diff
@@ -20,6 +20,14 @@ with lib;
...
+      nullable = true;
+      default = [
+        "vimPlugins"
+        "neorg-telescope"
+      ];
+    };
```
MattSturgeon, line 29: The alternate syntax can be used if you prefer:

```nix
neorgTelescopePackage = lib.mkPackageOption pkgs [
  "vimPlugins"
  "neorg-telescope"
] { nullable = true; };
```

The tail of the list will be used as the "name".

plugins/by-name/octo/default.nix
```diff
@@ -167,9 +165,16 @@ helpers.neovim-plugin.mkNeovimPlugin {
     };
   };
 
+  extraOptions = {
+    ghPackage = lib.mkPackageOption pkgs "gh" {
+      nullable = true;
+    };
```
MattSturgeon, line 171: I'd be tempted to pass a more verbose name for use in the docs?

```suggestion
    ghPackage = lib.mkPackageOption pkgs "GitHub CLI" {
      default = "gh";
      nullable = true;
    };
```

Can also pass `extraDescription` but I don't think that's useful in this case?

plugins/lsp/default.nix
```diff
@@ -16,6 +16,13 @@ in
     plugins.lsp = {
       enable = mkEnableOption "neovim's built-in LSP";
 
+      package = lib.mkPackageOption pkgs "nvim-lspconfig" {
+        default = [
+          "vimPlugins"
+          "nvim-lspconfig"
+        ];
+      };
```
MattSturgeon, line 24: Could use the alternate syntax again.

plugins/pluginmanagers/lazy.nix
```diff
@@ -41,6 +41,13 @@ in
     plugins.lazy = {
       enable = mkEnableOption "lazy.nvim";
 
+      package = lib.mkPackageOption pkgs "lazy-nvim" {
+        default = [
+          "vimPlugins"
+          "lazy-nvim"
+        ];
+      };
```
MattSturgeon, line 49: Same

plugins/pluginmanagers/packer.nix
```diff
@@ -14,6 +14,13 @@ in
     plugins.packer = {
       enable = mkEnableOption "packer.nvim";
 
+      package = lib.mkPackageOption pkgs "packer-nvim" {
+        default = [
+          "vimPlugins"
+          "packer-nvim"
+        ];
+      };
```
MattSturgeon, line 22: Same

MattSturgeon (approved):
I don't see any issues. Thanks for going through everything!

plugins/lsp/default.nix
```diff
@@ -16,6 +16,11 @@ in
     plugins.lsp = {
       enable = mkEnableOption "neovim's built-in LSP";
 
+      package = lib.mkPackageOption pkgs [
+        "vimPlugins"
+        "nvim-lspconfig"
+      ];
```
MattSturgeon, line 22:
```suggestion
      ] { };
```

This will error as "partially applied function assigned to option" or however its worded

khaneliman:
@mergifyio queue

PR #2118, opened 2024-08-30
plugins/telescope: add iconsPackage

khaneliman:
Revert [treesitter](https://github.com/nix-community/nixvim/pull/2108) add (I dont know what was going on in my head)

### DIFF
--- a/plugins/telescope/default.nix
+++ b/plugins/telescope/default.nix
@@ -6,7 +6,12 @@
 }:
 with lib;
 let
-  inherit (lib.nixvim) keymaps mkNullOrOption toLuaObject;
+  inherit (lib.nixvim)
+    keymaps
+    mkNullOrOption
+    mkPackageOption
+    toLuaObject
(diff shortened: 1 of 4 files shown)


plugins/telescope/default.nix
```diff
@@ -85,6 +90,11 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
         added to `extraPackages`.
       '';
     };
+
+    iconsPackage = mkPackageOption {
+      name = "nvim-web-devicons";
+      default = pkgs.vimPlugins.nvim-web-devicons;
+    };
```
MattSturgeon, line 97: Is the package actually installed anywhere?

khaneliman: No, that would only make sense.

MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2120, opened 2024-08-30
internals: cleanup with lib;

khaneliman:
Went through the internal modules and libraries cleaning up `with lib;` usage. Have another branch going through plugins but wanted to get opinions and extra eyes on this first.  

Tested with my own config and it seems to work fine with no regressions, but would like some extra eyes on this.

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -7,7 +7,6 @@
   hmOptions,
   search,
 }:
-with lib;
 let
   inherit (evaledModules.config.meta) nixvimInfo;
 
@@ -31,17 +30,17 @@ let
   removeWhitespace = builtins.replaceStrings [ " " ] [ "" ];
 
   getSubOptions =
-    opts: path: optionalAttrs (isVisible opts) (removeUnwanted (opts.type.getSubOptions path));
+    opts: path: lib.optionalAttrs (isVisible opts) (removeUnwanted (opts.type.getSubOptions path));
 
   isVisible =
     opts:
-    if isOption opts then
+    if lib.isOption opts then
       opts.visible or true
     else if opts.isOption then
       opts.index.options.visible or true
     else
       let
-        filterFunc = filterAttrs (_: v: if isAttrs v then isVisible v else true);
+        filterFunc = lib.filterAttrs (_: v: if lib.isAttrs v then isVisible v else true);
 
         hasEmptyIndex = (filterFunc opts.index.options) == { };
         hasEmptyComponents = (filterFunc opts.components) == { };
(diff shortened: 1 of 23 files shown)


docs/mdbook/default.nix
```diff
@@ -54,11 +53,11 @@ let
...
           opts
         else
-          filterAttrs (_: component: component.isOption && (isVisible component)) opts;
-      path = removeWhitespace (concatStringsSep "/" path);
+          lib.filterAttrs (_: component: component.isOption && (isVisible component)) opts;
+      path = removeWhitespace (lib.concatStringsSep "/" path);
       moduleDoc =
         let
-          info = optionalAttrs (hasAttrByPath path nixvimInfo) (getAttrFromPath path nixvimInfo);
+          info = lib.optionalAttrs (lib.hasAttrByPath path nixvimInfo) (lib.getAttrFromPath path nixvimInfo);
```
MattSturgeon, line 60: If we're touching this line anyway, may as well improve it to use [`attrByPath`](https://noogle.dev/f/lib/attrByPath):
```suggestion
          info = lib.attrByPath path { } nixvimInfo;
```

MattSturgeon (approved):
My preference would be to lean towards inheriting things explicitly from `builtins` and `lib.<scope>`, at least in files where we use particular functions many times.

The advantage of inheriting from `lib.<scope>` instead of the top-level `lib` is we can be more explicit about where a function comes from (and how it is categorized), while also being less verbose in the actual implementation code because we've inherited the function.

khaneliman:
@MattSturgeon Were you thinking more like removing the `lib.` and doing 
```nix
let
  inherit (lib.options)
    isOption
    mkEnableOption
    mkPackageOption
    mkOption
    ;
  inherit (lib.attrsets)
    optionalAttrs
    ;
  inherit (lib.strings)
    optionalString
    ;
in
```

MattSturgeon:
> Were you thinking more like removing the `lib.` and doing

That'd be my usual preference, but only when you judge its worth it to have the inherits instead of being verbose throughout the file...

E.g. If we only use `mkOption` twice, and on relatively simple lines, maybe we don't need to inherit it. On the other hand if `isOption` is used on a busy line where the `lib.` would be otherwise distracting maybe it is worth it?

Regardless, I'm happy with the PR as-is. We could go round in circles deciding the best approach for each function in each file :sweat_smile:

khaneliman:
> > Were you thinking more like removing the `lib.` and doing
> 
> That'd be my usual preference, but only when you judge its worth it to have the inherits instead of being verbose throughout the file...
> 
> E.g. If we only use `mkOption` twice, and on relatively simple lines, maybe we don't need to inherit it. On the other hand if `isOption` is used on a busy line where the `lib.` would be otherwise distracting maybe it is worth it?
> 

This is what my preference was, as well. I didn't think it was worth it with just a couple usages, but once it starts to feel repetitive or make it hard to read, it makes sense to inherit. 

> Regardless, I'm happy with the PR as-is. We could go round in circles deciding the best approach for each function in each file 😅

Sounds good. I'll see if I get anymore input and possibly merge tomorrow.

traxys:
I skimmed through the PR, and nothing jumped at me

khaneliman:
@mergifyio queue

PR #2121, opened 2024-08-31
[poc] tests/plugins: test package options

khaneliman:
Just a little POC from a discussion about testing options functionality working, as expected for when we add or don't add packages.

### DIFF
--- a/tests/test-sources/plugins/bufferlines/bufferline.nix
+++ b/tests/test-sources/plugins/bufferlines/bufferline.nix
@@ -1,7 +1,29 @@
+{ lib, pkgs, ... }:
 {
-  empty = {
-    plugins.bufferline.enable = true;
-  };
+  empty =
+    { config, ... }:
+    {
+      plugins.bufferline.enable = true;
+
+      assertions = [
+        {
+          assertion =
+            config.extraPlugins != [ ]
+            && lib.any (
+              x: lib.trace "${x.pname or ""}" x.pname or null == "bufferline.nvim"
+            ) config.extraPlugins;
+          message = "bufferline package wasn't found when it was expected";
+        }
+        {
+          assertion =
+            config.extraPlugins != [ ]
+            && lib.any (
+              x: lib.trace "${x.pname or ""}" x.pname or null == "nvim-web-devicons"
+            ) config.extraPlugins;
+          message = "nvim-web-devicons package wasn't found when it was expected";
+        }
+      ];
+    };
 
   example = {
     plugins.bufferline = {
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/bufferlines/bufferline.nix
```diff
@@ -124,10 +125,85 @@
...
+    {
+      plugins.bufferline = {
+        enable = true;
+      };
+
+      assertions = [
+        {
+          assertion =
+            config.extraPlugins != [ ]
+            -> lib.any (
```
MattSturgeon, line 139: I think this should be `&&`. Using `->` here will be true when `extraPlugins` is empty, and will only check the `lib.any` when it is not empty.

You can read `->` as "implies". A logical implication is true when the left-hand-side is false, but when the lhs is true, it also checks the right-hand-side.

`a -> b` is equivalent to `!a || b`.

MattSturgeon:
```suggestion
            config.extraPlugins != [ ]
            && lib.any (
```

tests/test-sources/plugins/bufferlines/bufferline.nix
```diff
@@ -124,10 +125,85 @@
...
+              x:
+              lib.trace "${x.pname or x.name}" x.pname or x.name != null && x.pname or x.name == "bufferline.nvim"
```
MattSturgeon, line 141: No need for a null check here. `null == "foo"` is valid nix and will evaluate to `false`.

I think what you want instead is `x.pname or x.name or null == "bufferline.nvim"`, however in final code you should know if you're checking `name` or `pname`, so just check one not both.

tests/test-sources/plugins/pluginmanagers/packer.nix
```diff
@@ -136,10 +137,19 @@
...
     plugins.packer = {
       enable = true;
       gitPackage = null;
     };
+
+    assertions = [
+      {
+        assertion =
+          cfg.config.extraPackages != [ ]
+          -> lib.any (x: x.pname != null && x.pname != "git") cfg.config.extraPackages;
```
MattSturgeon, line 150: If `x` is not an attrset that contains `pname`, then this will throw an eval error. You probably want `x.pname or null != "git"`

tests/test-sources/plugins/bufferlines/bufferline.nix
```diff
@@ -1,7 +1,29 @@
...
 {
-  empty = {
-    plugins.bufferline.enable = true;
-  };
+  empty =
+    { config, ... }:
+    {
+      plugins.bufferline.enable = true;
+
+      assertions = [
```
khaneliman, line 8: would it make more sense to test the package existence in our existing empty test cases?

tests/test-sources/plugins/pluginmanagers/packer.nix
```diff
@@ -136,10 +137,19 @@
...
     plugins.packer = {
       enable = true;
       gitPackage = null;
     };
+
+    assertions = [
+      {
+        assertion =
+          cfg.config.extraPackages == [ ]
+          || lib.any (x: lib.trace "${x.pname or ""}" x.pname or null != "git") cfg.config.extraPackages;
```
khaneliman, line 150: looks like we need to check using or when a derivation might not contain any extraPackages with the disabling of a package option

MattSturgeon: Is this supposed to be using `lib.all`?

I.e. _all packages are not named "git"_

If so we shouldn't need to check for an empty list.

khaneliman: Yeah, that'd be cleaner.

tests/test-sources/plugins/bufferlines/bufferline.nix
MattSturgeon: I think it's nice to keep the "empty" case simple.

Maybe we should consider adding `test.runNvim = false` to these assertion-based tests though, since we don't actually need to run neovim if all we care about is the assertion.

tests/test-sources/plugins/bufferlines/bufferline.nix
```diff
@@ -1,7 +1,29 @@
...
+        {
+          assertion =
+            config.extraPlugins != [ ]
+            && lib.any (
+              x: lib.trace "${x.pname or ""}" x.pname or null == "bufferline.nvim"
```
MattSturgeon, line 13: Looking on noogle, there's also [`lib.getName`](https://noogle.dev/f/lib/getName), although I think that could fail when encountering submodule-type plugin definitions...

tests/test-sources/plugins/bufferlines/bufferline.nix
khaneliman: Good idea, don't need to waste processing time that way if it's just evaluation time testing.

MattSturgeon: Makes me think we could do with a `test.build` toggle to completely disable _building_ `finalPackage`, although we'd probably still want to ensure it is evaluated somehow (just not built)

MattSturgeon: > Makes me think we could do with a `test.build` toggle to completely disable _building_ `finalPackage`

This was added as `test.buildNixvim` which also implicitly disables `test.runNvim`.

Issue #2125, opened 2024-08-31
[BUG] `none-ls` broken


gador:
| Field        | Description |
|--------------|-------------|
| Plugin       | `none-ls`  | 
| Nixpkgs      | `5629520edecb69630a3f4d17d3d33fc96c13f6fe`  (current master)| 
| Home Manager | `unstable` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

After an update of my flake, the following error appears:
```
       … while calling anonymous lambda

         at /nix/store/g50vm3mi7j7ysaimxryw0hv7f95x0lrf-source/lib/modules.nix:648:14:

          647|           optionDecls = filter
          648|             (m: m.options?_type
             |              ^
          649|                 && (m.options._type == "option"

       … from call site

         at /nix/store/bbww7acvbh82z5air318r5iqn2cy1rcb-source/plugins/none-ls/_mk-source-plugin.nix:58:12:

           57|         }
           58|         // lib.optionalAttrs (pkg != null) { default = pkg; }
             |            ^
           59|       );

       … while calling 'optionalAttrs'

         at /nix/store/g50vm3mi7j7ysaimxryw0hv7f95x0lrf-source/lib/attrsets.nix:1352:5:

         1351|     cond:
         1352|     as:
             |     ^
         1353|     if cond then as else {};

       error: attribute 'prisma' missing

       at /nix/store/bbww7acvbh82z5air318r5iqn2cy1rcb-source/plugins/none-ls/packages.nix:158:23:

          157|       phpcs = pkgs.phpPackages.php-codesniffer;
          158|       prisma_format = pkgs.nodePackages.prisma;
             |                       ^
          159|       ptop = pkgs.fpc;
```

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
    enable = true;
};
```

msanft:
Experiencing the same

MattSturgeon:
Should've been fixed by #2128

PR #2128, opened 2024-08-31
plugins/none-ls: fix due to upstream renaming prisma

gador:
See #2125

due to https://github.com/NixOS/nixpkgs/pull/337521

This has already been merged in `unstable` and thus `nixvim` will fail on up-to-date `home-manager` or `NixOS` deployments, but not on older configs. Unfortunately, `nixpkgs`  missed to apply an alias for the package.

Due to this change `nix flake check --all-systems` fails with 
```
       … while checking flake output 'checks'

         at «none»:0: (source not available)

       … while checking the derivation 'checks.x86_64-linux.extend'

         at «none»:0: (source not available)

       error: attribute 'prisma' missing

       at /nix/store/cvgwzs7pal633x4q03lp6wzcrk7g66q9-source/plugins/none-ls/packages.nix:158:23:

          157|       phpcs = pkgs.phpPackages.php-codesniffer;
          158|       prisma_format = pkgs.prisma;
             |                       ^
          159|       ptop = pkgs.fpc;
       Did you mean prism?
```

EDIT:
This PR also fixes the move of `eslint` to `pkgs` which will fix:

`error: Alias eslint is still in node-packages.nix` which is due to https://github.com/NixOS/nixpkgs/pull/337888/

### DIFF
--- a/plugins/lsp/language-servers/efmls-configs-pkgs.nix
+++ b/plugins/lsp/language-servers/efmls-configs-pkgs.nix
@@ -94,7 +94,6 @@ pkgs: {
       vulture
       ;
     inherit (nodePackages)
-      eslint
       eslint_d
       prettier
       alex
@@ -113,6 +112,10 @@ pkgs: {
     cmake_lint = cmake-format;
     dartfmt = dart;
     dotnet_format = dotnet-runtime;
+    # TODO: Added 2024-08-31; remove 2024-11-31
+    # eslint was moved out of nodePackages set without alias
+    # Using fallback as a transition period
+    eslint = pkgs.eslint or pkgs.nodePackages.eslint;
     fish_indent = fish;
     gofmt = go;
     goimports = go-tools;
--- a/plugins/none-ls/packages.nix
+++ b/plugins/none-ls/packages.nix
@@ -155,7 +155,10 @@ pkgs: {
       phpcbf = pkgs.phpPackages.php-codesniffer;
       phpcsfixer = pkgs.phpPackages.php-cs-fixer;
       phpcs = pkgs.phpPackages.php-codesniffer;
-      prisma_format = pkgs.nodePackages.prisma;
+      # TODO: Added 2024-08-31; remove 2024-11-31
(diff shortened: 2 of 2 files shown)


khaneliman:
We'll have to handle this in our https://github.com/nix-community/nixvim/pull/2127 if you want to target that as the base

gador:
Thanks. By basing on `update/main` this should do the trick, right?

khaneliman:
That branch will be the flake.lock updates. But, it doesn't look like a nixpkgs update was caught by it, yet.

plugins/none-ls/packages.nix
```diff
@@ -155,7 +155,7 @@ pkgs: {
       phpcbf = pkgs.phpPackages.php-codesniffer;
       phpcsfixer = pkgs.phpPackages.php-cs-fixer;
       phpcs = pkgs.phpPackages.php-codesniffer;
-      prisma_format = pkgs.nodePackages.prisma;
+      prisma_format = pkgs.prisma;
```
khaneliman, line 158:
```suggestion
      prisma_format = pkgs.prisma or pkgs.nodePackages.prisma;
```

plugins/lsp/language-servers/efmls-configs-pkgs.nix
```diff
@@ -94,7 +95,6 @@ pkgs: {
       vulture
       ;
     inherit (nodePackages)
-      eslint
```
khaneliman, line 97: this needs a fallback, too. Should be something like `eslint = pkgs.eslint or pkgs.nodePackages.eslint;`

MattSturgeon (approved):
LGTM, thanks! Have we reported the lack of an alias as an issue upstream?

gador:
> LGTM, thanks! Have we reported the lack of an alias as an issue upstream?

Done: https://github.com/NixOS/nixpkgs/pull/337521#issuecomment-2323022673

MattSturgeon:
@mergifyio queue

Issue #2124, opened 2024-08-31
[BUG] gitgutter broken


gador:
| Field        | Description |
|--------------|-------------|
| Plugin       | `gitgutter`  | 
| Nixpkgs      | `5629520edecb69630a3f4d17d3d33fc96c13f6fe` (current master)| 
| Home Manager | `unstable` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

(Probably) due to https://github.com/nix-community/nixvim/commit/efb004a61fb0173a2e5625c401ef6738ff2bea55 (from @khaneliman specifically  https://github.com/nix-community/nixvim/blob/ee6ee48bbe1ffa88fd4b2af7d68ab0315bc817f0/plugins/git/gitgutter.nix#L196) the following error is observed:

```nix
       … while evaluating the option `home-manager.users.gador.programs.nixvim.extraPackages':

       … while calling anonymous lambda

         at /nix/store/bbww7acvbh82z5air318r5iqn2cy1rcb-source/modules/output.nix:40:32:

           39|       description = "Extra packages to be made available to neovim";
           40|       apply = builtins.filter (p: p != null);
             |                                ^
           41|     };

       … while calling anonymous lambda

         at /nix/store/9yqbs6sdhxhcjf7ynmwn3sav65y5v805-source/lib/types.nix:552:14:

          551|       merge = loc: defs:
          552|         map (x: x.value) (filter (x: x ? value) (concatLists (imap1 (n: def:
             |              ^
          553|           imap1 (m: def':

       error: A definition for option `home-manager.users.gador.programs.nixvim.extraPackages."[definition 9-entry 2]"' is not of type `null or package'. Definition values:
       - In `/nix/store/bbww7acvbh82z5air318r5iqn2cy1rcb-source/plugins/git/gitgutter.nix': [ ]
 ```

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
    enable = true;
    plugins.gitgutter = {
      enable = true;
    };
};
```

MattSturgeon:
Yeah, looks like it should be
```nix
extraPackages = [
  cfg.gitPackage
  cfg.grepPackage
]; 
```

khaneliman:
> Yeah, looks like it should be
> 
> ```nix
> extraPackages = [
>   cfg.gitPackage
>   cfg.grepPackage
> ]; 
> ```

[...]

PR #2129, opened 2024-08-31
plugins/gitgutter: fix grepPackage

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2124 and adds some of the package assertions @MattSturgeon and I were playing around with in https://github.com/nix-community/nixvim/pull/2121

Looks like it needs to be refactored to newest standards, as well. But, that will be a different PR.


MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2131, opened 2024-08-31
Add Scope Plugin

insipx:
Adds [tiagovla/scope.nvim](https://github.com/tiagovla/scope.nvim) to utils


lib/maintainers.nix
```diff
@@ -18,7 +18,7 @@
     github = "Bodleum";
     githubId = 60107449;
     name = "Daniel Laing";
-    keys = [ { fingerprint = "0821 8B96 DC73 85E5 BB7C  A535 D264 3BD2 13BC 0FA8"; } ];
+    keys = [{ fingerprint = "0821 8B96 DC73 85E5 BB7C  A535 D264 3BD2 13BC 0FA8"; }];
```
MattSturgeon, line 21: It seems you're using a different formatter to us. Could you instead try using `nix fmt` to format using the formatter we've configured in our flake?

lib/maintainers.nix
```diff
@@ -95,4 +95,10 @@
     githubId = 30784060;
     name = "Pedro Sánchez";
   };
+  insipx = {
+    email = "github.tech@liquidthink.net";
+    github = "insipx";
+    githubId = 6452260;
+    name = "Andrew Plaza";
+  };
```
MattSturgeon, line 103: Can you do this in a separate commit, `"maintainers: add insipx"`?

This commit should come before the main one, which will be titled `"plugins/scope: init"`.

plugins/utils/scope.nix
```diff
@@ -0,0 +1,50 @@
...
+  originalName = "scope.nvim";
+  defaultPackage = pkgs.vimPlugins.scope-nvim;
+
+  maintainers = [ lib.maintainers.insipx ];
+
+  settingsOptions = {
+    hooks = {
+      pre_tab_enter = helpers.defaultNullOpts.mkLuaFn "" ''
+        Run custom logic before entering a tab.
+      '';
```
MattSturgeon, line 18: If we don't know the plugin default, can you use the `'` variant and _not_ supply a `pluginDefault`?

```nix
pre_tab_enter = helpers.defaultNullOpts.mkLuaFn' {
  description = ''
    Run custom logic before entering a tab.
  '';
}
```

Or alternatively, if the plugin defaults to the option being nil/undefined, perhaps we should set our `pluginDefault` to `null`?

tests/test-sources/plugins/utils/scope.nix
```diff
@@ -0,0 +1,21 @@
...
+    plugins.scope = {
+      enable = true;
+
+      settings = {
+        hooks = {
+          pre_tab_enter.__raw = ''
+            function()
+              print("About to enter tab")
+            end
+          '';
```
MattSturgeon, line 16: If this really is the plugin-default, it should also be reflected in the option declaration.

Ideally, the `defaults` test should explicitly set the plugin-default for _all_ declared settings options.

If this is more of an "example" test-case, that's fine too. But it should be reflected in the test-case name.

plugins/utils/scope.nix
```diff
@@ -0,0 +1,50 @@
+{ lib
+, helpers
```
MattSturgeon, line 2: We've soft-deprecated using `helpers`. Instead you should use `lib.nixvim`, which is the same thing.

MattSturgeon (changes requested):
A few minor points to address, but overall looks good!

lib/maintainers.nix
MattSturgeon: Also, if you enter a devshell (e.g. using direnv, or by running `nix develop`), a git "pre-commit" hook will be installed that checks the formatting is correct each time you make a commit. :grin:

insipx:
Thanks for your review @MattSturgeon ! A bit new to contributing to a nix project so thanks for the patience. I'll make sure to fix everything to the standard!

tests/test-sources/plugins/utils/scope.nix
```diff
@@ -0,0 +1,21 @@
+{
+  empty = {
+    plugins.scope.enable = true;
+  };
+
+  example = {
```
MattSturgeon, line 6: I'd like to see a `defaults` test case too, ideally.

```nix
defaults = {
  plugins.scope = {
    enable = true;

    settings.hooks = {
      pre_tab_enter = null;
      post_tab_enter = null;
      pre_tab_leave = null;
      post_tab_leave = null;
      pre_tab_close = null;
      post_tab_close = null;
    };
  };
};
```

MattSturgeon (approved):
I've double checked [here](https://github.com/tiagovla/scope.nvim/blob/main/lua/scope/config.lua), and it seems `nil` (or `null`) _is_ the correct plugin-default.

All my other feedback seems to be addressed, thanks for contributing this excellent first PR.

If you're able, it'd be great to have a "defaults" test-case, too. But for this plugin that's a fairly minor point, so I won't block on it :grin: 

If you like, you can rebase on our `main` branch, which will give you "full" committer credit. Otherwise we can rebase for you, and you'll still get "author" credit.

MattSturgeon:
@mergifyio queue

Issue #1945, opened 2024-07-28
[BUG] format: no formatter for path: .editorconfig


MattSturgeon:
When running nix fmt, the warning will print (assuming it isn't cached):

```
WARN format: no formatter for path: .editorconfig
```

PR #2132, opened 2024-08-31
flake-modules/dev: treefmt ignore .editorconfig

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1945

I didn't see an ini formatter that we could use for this file, so we can just ignore it like we do a few other files.


MattSturgeon (approved):
❤️

khaneliman:
@mergifyio queue

Issue #2094, opened 2024-08-26
cmp `autoEnableSources` behaviour should be documented more prominently


MattSturgeon:
As mentioned here: https://github.com/nix-community/nixvim/discussions/928#discussioncomment-10455697, users often miss the nuances of how [`plugins.cmp.settings.sources`](https://nix-community.github.io/nixvim/plugins/cmp/settings/index.html#pluginscmpsettingssources) and [`plugins.cmp.autoEnableSources`](https://nix-community.github.io/nixvim/plugins/cmp/index.html#pluginscmpautoenablesources) work.

Currently we have the following notice under `settings.sources`:

> WARNING: If `plugins.cmp.autoEnableSources` Nixivm will automatically enable the corresponding source plugins. This will work only when this option is set to a list. If you use a raw lua string, you will need to explicitly enable the relevant source plugins in your nixvim configuration.

PR #2136, opened 2024-09-01
plugins/cmp: add description to elevate autoEnableSource info

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2094


plugins/completion/cmp/default.nix
```diff
@@ -16,6 +16,43 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
 
+  description = ''
+    ### Completion Source Installation
+
+    If `plugins.cmp.autoEnableSources` Nixivm will automatically enable the corresponding source
+    plugins. This will work only when this option is set to a list.
+    If you use a raw lua string, you will need to explicitly enable the relevant source plugins in
+    your nixvim configuration.
+
+    ```nix
```
traxys, line 27: Maybe add a small caption to tell the difference in the two blocks, something like "With auto-enabled sources", "Without auto-enabled sources" ?

plugins/completion/cmp/default.nix
```diff
@@ -16,6 +16,43 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   maintainers = [ maintainers.GaetanLepage ];
 
+  description = ''
+    ### Completion Source Installation
+
+    If `plugins.cmp.autoEnableSources` Nixivm will automatically enable the corresponding source
```
MattSturgeon, line 22: I think we should be more explicit than "if \<option\>", "if \<option\> is \<value\>" feels better.

I also think we should somehow indicate that this is the default state, however my wording below is kinda clunky, and I'm not happy with it :sweat: 
```suggestion
    If `plugins.cmp.autoEnableSources` is true, which it is by default, Nixivm will automatically enable the corresponding source
```

plugins/completion/cmp/default.nix
```diff
@@ -16,6 +16,45 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   maintainers = [ maintainers.GaetanLepage ];
 
+  description = ''
+    ### Completion Source Installation
+
+    If `plugins.cmp.autoEnableSources` is `true` Nixivm will automatically enable the corresponding source
+    plugins. This is the default behavior, but will work only when this option is set to a list.
+    If you use a raw lua string or set `plugins.cmp.autoEnableSources` to `false`, you will need to explicitly enable the relevant source plugins in
+    your nixvim configuration.
```
MattSturgeon, line 25: I'd be tempted to add a paragraph break here, and maybe use semantic line-wrapping at the comma, but neither are big issues.

```suggestion
    If `plugins.cmp.autoEnableSources` is `true` Nixivm will automatically enable the corresponding source
    plugins. This is the default behavior, but will work only when this option is set to a list.

    If you use a raw lua string or set `plugins.cmp.autoEnableSources` to `false`,
    you will need to explicitly enable the relevant source plugins in your nixvim configuration.
```

MattSturgeon (approved):
Awesome. Let's give @traxys chance to review if he is able, but all our feedback seems to be addressed :tada:

traxys (approved):

khaneliman:
@mergifyio queue

PR #2115, opened 2024-08-29
plugins/nvim-orgmode: init

refaelsh:
I want to add this plugin to Nixvim: https://github.com/nvim-orgmode/orgmode.
It is already in Nixpkgs here: https://search.nixos.org/packages?channel=24.05&show=vimPlugins.orgmode&from=0&size=50&sort=relevance&type=packages&query=orgmode.
I followed the following docs: https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md.

I hope I did everything right.
Furthermore, I especially have doubts about the `tests` parts of the docs, I am not that good at Nix to understand what is said there.

Thanks in advance for reviewing this PR.

PR #2137, opened 2024-09-01
Update the template with helpers -> lib.nixvim

refaelsh:
This PR is in cointiunuce of this comment: https://github.com/nix-community/nixvim/pull/2115#discussion_r1739915637.

Closes #1982

### DIFF
--- a/plugins/TEMPLATE.nix
+++ b/plugins/TEMPLATE.nix
@@ -1,11 +1,10 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-helpers.neovim-plugin.mkNeovimPlugin config {
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
   name = "my-plugin";
   originalName = "my-plugin.nvim"; # TODO replace (or remove entirely if it is the same as `name`)
   defaultPackage = pkgs.vimPlugins.my-plugin-nvim; # TODO replace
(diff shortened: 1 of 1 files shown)


plugins/TEMPLATE.nix
```diff
@@ -1,11 +1,10 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
```
MattSturgeon, line 6:
```suggestion
}:
let
  inherit (lib.nixvim) defaultNullOpts;
in
```

plugins/TEMPLATE.nix
```diff
@@ -14,11 +13,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
 
   # Optionally, explicitly declare some options. You don't have to.
   settingsOptions = {
-    foo = helpers.defaultNullOpts.mkUnsignedInt 97 ''
+    foo = lib.nixvim.defaultNullOpts.mkUnsignedInt 97 ''
```
MattSturgeon, line 16:
```suggestion
    foo = defaultNullOpts.mkUnsignedInt 97 ''
```

plugins/TEMPLATE.nix
```diff
@@ -14,11 +13,11 @@ helpers.neovim-plugin.mkNeovimPlugin config {
...
       The best birth year.
     '';
 
-    great_feature = helpers.defaultNullOpts.mkBool false ''
+    great_feature = lib.nixvim.defaultNullOpts.mkBool false ''
```
MattSturgeon, line 20:
```suggestion
    great_feature = defaultNullOpts.mkBool false ''
```

MattSturgeon (changes requested):
Awesome, we've been working on this in #1982, but as that's stalled we can merge here and make any further changes in the future.

If you could apply the suggestions below and then rebase/squash so that your commit is titled ``"plugins/template: `helpers` -> `lib.nixvim`"`` or similar, that'd be great! :grin:

MattSturgeon:
Actually, looking at #1982 it already includes all these changes and is only stalled on further changes that'd be nice to have.

I'll go ahead and merge the original PR, but thanks for addressing this regardless!

refaelsh:
> I'll go ahead and merge the original PR, but thanks for addressing this regardless!

Sure. No problem.

Issue #1970, opened 2024-08-02
`plugins.*.package` should be nullable, to use package from system.


MattSturgeon:
### Discussed in https://github.com/nix-community/nixvim/discussions/1969

<div type='discussions-op-text'>

<sup>Originally posted by **niksingh710** August  2, 2024</sup>
```nix
  plugins = {
    ltex-extra = {
      enable = true;
      # This has been made null so that nvix uses system latex package and nix run becomes faster (no downloads for test or build)
      # vimplugin-ltex_extra.nvim
      package = null;
    };
    vimtex.enable = true;
  };
```
  
  I am trying this to make ltex-extra use the ltex-extra package that is available in my system.

but this throws an error  

```
error: A definition for option `plugins.ltex-extra.package' is not of type `package'. Definition values:
       - In `/nix/store/fdn1lw1j5wlmpqxixr92hy8nvx7igshf-lang/tex.nix': null
```</div>

MattSturgeon:
My understanding is that we don't allow null in these package options because we need to be able to add _something_ to neovim's rtp (via `extraPlugins`).

I don't know if there's a way to do this by getting system installed packages from the PATH, so this may be a WONTFIX.
@traxys @GaetanLepage Do you have any insight into this?

traxys:
I don't know what such a feature would look like indeed, maybe we could allow strings, and this would add the corrsponding path to the rtp?

MattSturgeon:
Perhaps just making the option nullable is enough, but we'd have to document clearly that users need to manually edit their rtp for the plugin to work.

I expect many users would wrongly expect to be able to just add `vimPlugins.*` to their `home.packages` or equivalent.

Instead they could either add to their rtp via `extraConfigLuaPre` or by adding their own package to `extraPlugins`... Although doing that would be no different from setting `package`.

niksingh710:
Enabling null should be an option to set up a conditional configuration for `nixd, statix, and deadnix` such that the Language Server Protocol (LSP) only activates them if they are present on the system.
and if the user is explicitly making a package `null` then they should know the workaround.
Not only for lsp packages but others too.

MattSturgeon:
Most package options in nixvim _are_ nullable already; anywhere that the program can be loaded from the PATH.

It should only be plugin-related package options that aren't, because they need to be added to the rtp to be loaded by neovim.

niksingh710:
[...]

Issue #1950, opened 2024-07-30
`mkPackageOption` helper with `defaultText`


MattSturgeon:
Nixpkgs has `lib.mkPackageOption` ([documentation](https://nixos.org/manual/nixos/unstable/#sec-option-declarations-util-mkPackageOption), [noogle](https://noogle.dev/f/lib/options/mkPackageOption), [source](https://github.com/NixOS/nixpkgs/blob/e1d5bdcc6c01d7a764b0239cbdf20dcceaac393d/lib/options.nix#L107-L221)) which takes a "pkgs path" and produces an option with an appropriate default and literalExpression defaultText.

```nix
# e.g.
lib.mkPackageOption pkgs "hello" { }
# is like
lib.mkOption {
  type = lib.types.package;
  default = pkgs.hello;
  defaultText = lib.literalExpression "pkgs.hello";
  description = "The hello package to use.";
}
```

We should use this where possible, and copy the implementation in our nullable version.

We could do this in a non-breaking way by having our nullable helper check the `pkg` arg type; lists and strings can be pkgs-paths while derivations should be actual packages.

Eventually, we could warn when a derivation is passed in to hunt down any remaining call sites, but I see this more as something that can improve future code rather than something we should be in a hurry to migrate existing code to.

# TODO:
- [x] Add support to `mkVimPlugin` and `mkNeovimPlugin`
      #2139
- [x] Implement fancier support (i.e. adding the option _within_ `mkNeovimPlugin`, using an imported module)
- [x] Migrate all plugins `defaultPackage`
- [x] Remove the deprecated helpers (`mkPackageOption` & `mkPluginPackageOption`)

MattSturgeon:
We have two helpers currently.

- `mkPluginPackageOption`, is also just `types.package`.
  - Used 93 times accross the codebase
  - Two of those are in `mkNeovimPlugin` and `mkVimPlugin`...
  - This function could be replaced with nixpkg's `mkPackageOption`, or could be a wrapper that sets a custom description.
- `mkPackageOption` is similar, but has a `nullOr` type.
  - It is currently used 51 times across the codebase.

MattSturgeon:
Fixed by #2150

PR #2139, opened 2024-09-01
lib/*-plugins: initial support for `lib.mkPackageOption`

MattSturgeon:
## Summary

Adds initial support for receiving a package _option_ instead of a package _derivation_ in `mkVimPlugin` and `mkNeovimPlugin`. This is done to enable the use of [`lib.mkPackageOption`](https://nixos.org/manual/nixos/unstable/#sec-option-declarations-util-mkPackageOption) as discussed in #1950.

This soft-deprecates our current helpers, `lib.nixvim.mkPluginPackageOption` and `lib.nixvim.mkPackageOption`.

Also implemented this for the plugin [TEMPLATE](https://github.com/nix-community/nixvim/pull/2139/files#diff-93af850d88bc329f8ac9a11854624cfd9cf99d723bac17464f4dac5993425f3d), [cmp sources](https://github.com/nix-community/nixvim/pull/2139/files#diff-961bb62e8020d29c7e320ed7a45345f1fa86e338342bb44df02bd161e7ff96c2), telescope extensions, and [lsp language-servers](https://github.com/nix-community/nixvim/pull/2139/files#diff-84e25e99a6dafddf43e738af5502df469f8deb4f4eed194544c2de10133abe00).

## Why

An option with a package default, will be rendered using `lib.generators.toPretty` as (e.g.) `<derivation vimplugin-cmp-ai-2024-08-01>`.

We could specify a `defaultText` for all our package options, e.g. `defaultText = lib.literalExpression "pkgs.cmp-ai"`, however `lib.mkPackageOption` automates this by setting both `default` and `defaultText` based on an attr-path.

#### Screenshots

<details><summary>Before+after screenshots</summary>
<p>

### Before

image

### After

image

</p>
</details>

## Alternative/future approach

It would be **possible** to have `mk*Plugin` handle calling `lib.mkPackageOption` internally, however it'd have to do so within an imported module in order to gain access to the (correct) `pkgs` arg[^1].

We may need to do this _anyway_ to also gain access to `options` and `config` without having to pass them into the helpers, but that is out of scope for this PR...

See this POC:

<details><summary>Proof of concept</summary>
<p>

```nix
mkNeovimPlugin =
  {
    name,
    defaultPackage,
    # omitted for clarity
  }:
  {
    imports = [
      # Import a module, so we have access to the "correct" pkgs
      ({ config, options, pkgs, ... }: {
        options.plugins.${name} = {
          enable = lib.mkEnableOption name;
          package =
            if lib.isOption defaultPackage then
              defaultPackage
            else
              lib.mkPackageOption pkgs name {
                default = defaultPackage;
              };
        };
      })
    ];
    # Rest omitted for clarity
  }
```

</p>
</details> 

[^1]: As part of `helpers`, we do have access to _a_ `pkgs`, but it may not be the correct one configured for use within the module eval. Incidentally, this is making me think we should separate out all helper functions that rely on `pkgs` and put them somewhere else so that `helpers` doesn't need a `pkgs`.

### DIFF
--- a/docs/user-guide/faq.md
+++ b/docs/user-guide/faq.md
@@ -29,6 +29,8 @@ The [nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#managing-plugins-w
 
 When using NixVim it is possible to encounter an error of the type `attribute 'name' missing`, for example it could look like:
 
+<!-- TODO: Update example now that we use `mkPackageOption` -->
+
 ```
        (stack trace truncated; use '--show-trace' to show the full trace)
 
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -28,13 +28,12 @@ with lib;
       colorscheme ? name,
       # options
       originalName ? name,
-      # WARNING: `defaultPackage` is deprecated by `package`,
-      defaultPackage ? throw "mkVimPlugin called without either `package` or `defaultPackage`.",
       # Can be a string, a list of strings, or a module option:
       # - A string will be intrpreted as `pkgs.vimPlugins.${package}`
       # - A list will be interpreted as a "pkgs path", e.g. `pkgs.${elem1}.${elem2}.${etc...}`
       # - An option will be used as-is, but should be built using `lib.mkPackageOption`
-      package ? helpers.mkPluginPackageOption originalName defaultPackage,
+      # Defaults to `name`, i.e. `pkgs.vimPlugins.${name}`
+      package ? name,
       settingsOptions ? { },
       settingsExample ? null,
       settingsDescription ? "Options provided to the `require('${luaName}')${setup}` function.",
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -17,7 +17,7 @@ with lib;
     {
       name,
       maintainers,
-      url ? defaultPackage.meta.homepage,
+      url ? throw "default",
       imports ? [ ],
       description ? null,
       # deprecations
@@ -28,7 +28,13 @@ with lib;
       colorscheme ? name,
       # options
       originalName ? name,
-      defaultPackage,
+      # WARNING: `defaultPackage` is deprecated by `package`,
+      defaultPackage ? throw "mkVimPlugin called without either `package` or `defaultPackage`.",
+      # Can be a string, a list of strings, or a module option:
+      # - A string will be intrpreted as `pkgs.vimPlugins.${package}`
+      # - A list will be interpreted as a "pkgs path", e.g. `pkgs.${elem1}.${elem2}.${etc...}`
+      # - An option will be used as-is, but should be built using `lib.mkPackageOption`
+      package ? helpers.mkPluginPackageOption originalName defaultPackage,
       settingsOptions ? { },
       settingsExample ? null,
       settingsDescription ? "Options provided to the `require('${luaName}')${setup}` function.",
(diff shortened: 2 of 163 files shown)


lib/vim-plugin.nix
```diff
@@ -5,7 +5,7 @@ with lib;
     config:
     {
       name,
-      url ? if defaultPackage != null then defaultPackage.meta.homepage else null,
+      url ? if packageOption == null then null else packageOption.default.meta.homepage,
```
traxys, line 8: nit: can't this be done with a `or` ?

traxys (commented):
It would have been nice to remove completly the `defaultPackage` in this PR, it would simplify a bit the helpers, though I think I saw there are around 150 occurences, it may be some work...

Maybe a regex could replace them all? Not sure

lib/vim-plugin.nix
MattSturgeon: Actually this isn't needed; when I first started work on this `defaultPackage` and `packageOption` were nullable and the `package` option would be omitted from `options`.

That behaviour is never used in nixvim, so I dropped it; `packageOption` is no longer nullable.

To answer your original question, using `or` would cause `packageOption.default.meta.homepage` to silently be null when the package has no homepage, however we'd want that to error so the maintainer knows he has to set `url` explicitly.

MattSturgeon:
> It would have been nice to remove completely the `defaultPackage` in this PR, though I think I saw there are around 150 occurences, it may be some work...

I agree it'd be nice, however I don't want to rush to replace all `defaultPackage`s with `lib.mkPackageOption` if we're still considering the imported-module approach.

A treewide substitution of `defaultPackage = pkgs.foo` -> `defaultPackage = "foo"` would be much easier as well...

traxys:
Indeed, it would be nice to leave the `defaultPackage` argument when no customization is required

khaneliman:
So this gives us clearer documentation and nullable packages for plugins, which are big wins to resolve https://github.com/nix-community/nixvim/issues/1970 and https://github.com/nix-community/nixvim/issues/1950. Does it also have the benefit of the lazy package evaluation to prevent pkg errors being thrown from unused options or is that still an issue that requires a a different solution?

traxys:
I think because we include the man page by default it will never be lazy, whatever we do. We may need to reconsider the man page inclusion by default

MattSturgeon:
I have some gripes with how we build the man docs too; they will always use our flake's pkgs rather than the pkgs used in module eval.

To achieve laziness, I think we need to either ditch the man pages being included by default, and/or remove automatically getting the `meta.homepage` from the default package.

khaneliman:
Did you wanna add a test to verify how it responds to null packages with the new functionality?

MattSturgeon:
We already have some `null` packages in lsp/language-servers, and upstream document how it should behave when `default = null`, so I don't see it as a major issue. If you have a suggestion for a test case I'm happy to consider adding something though.

khaneliman:
Sorry, was in the mindset of top level package option. These were already nullable.

lib/neovim-plugin.nix
```diff
@@ -29,7 +29,12 @@ with lib;
       colorscheme ? name,
       # options
       originalName ? name,
-      defaultPackage,
+      # WARNING: `defaultPackage` is being replaced with `packageOption`,
+      # which should be defined using: lib.mkPackageOption pkgs originalName { default = [ "path" "to" "package" ]; }
+      # NOTE: We can't do that here (without importing a module) because we don't have access to the `pkgs` module arg.
```
MattSturgeon, line 34: Should this be tackled in this PR? I.e. implementing the package option within an imported-module

traxys: It would not need to much additional complexity with #2140 right?

MattSturgeon:
> So this gives us clearer documentation and nullable packages for plugins

@khaneliman just to clarify, this does **not** introduce nullable package options, unless a maintainer chooses to explicitly pass `nullable = true` to `mkPackageOption`, like I did in mkLsp.

MattSturgeon:
@traxys this now migrates all call-sites for `mkVimPlugin` and `mkNeovimPlugin` to use the new pattern, so I've removed the old `defaultPackage` argument from those functions.

I don't want to remove `helpers.mkPackageOption` and `helpers.mkPluginPackageOption` in _this_ PR, because the treewide parts are already getting large, and will be difficult to review if we also have to check various ad-hoc package options are migrated correctly (e.g. options like `plugins.*.gitPackage`).

There's currently 89 matches for `mkPluginPackageOption` and 60 matches for `mkPackageOption` in the repo, although some of those are for `lib.mkPackageOption` (there's 40 matches for `/helpers\.mkPackageOption/`).

docs/user-guide/faq.md
```diff
@@ -29,6 +29,8 @@ The [nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#managing-plugins-w
 
 When using NixVim it is possible to encounter an error of the type `attribute 'name' missing`, for example it could look like:
 
+<!-- TODO: Update example now that we use `mkPackageOption` -->
```
khaneliman, line 32: Did we want to handle this in this PR or follow up?

khaneliman (approved):
LGTM, I very much appreciate the direction of this PR.

docs/user-guide/faq.md
MattSturgeon: I think we'll leave it for later. Next time a user reports a mismatched nixpkgs error we can use the error message they report in the example here.

MattSturgeon:
@mergifyio queue

PR #2140, opened 2024-09-01
lib/*-plugin: drop `config` arg

MattSturgeon:
- **lib/vim-plugin: drop `config` arg**
- **lib/neovim-plugin: drop `config` arg**

Instead of passing in `config` explicitly, we can access it via an imported module.

This will work better if we need to access additional module args, such as `options` or `pkgs`.

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -14,7 +14,6 @@ with lib;
   };
 
   mkNeovimPlugin =
-    config:
     {
       name,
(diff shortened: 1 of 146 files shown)


traxys (approved):
Did not check that none was forgotten, but that should be handled by tests

khaneliman (approved):
LGTM

MattSturgeon:
@mergifyio queue

PR #2144, opened 2024-09-02
Add indent option to nvim-orgmode

refaelsh:
This PR adds the `org_startup_indented` option to nvim-orgmode.

### DIFF
--- a/plugins/languages/nvim-orgmode.nix
+++ b/plugins/languages/nvim-orgmode.nix
@@ -21,10 +21,16 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin config {
     org_default_notes_file = defaultNullOpts.mkStr "" ''
       A path to the default notes file.
     '';
+    org_startup_indented = defaultNullOpts.mkBool false ''
+      When set to true, uses Virtual indents to align content visually.
+      The indents are only visual, they are not saved to the file.
+      When set to false, does not add any Virtual indentation.
+    '';
   };
 
   settingsExample = {
     org_agenda_files = "~/orgfiles/**/*";
     org_default_notes_file = "~/orgfiles/refile.org";
+    org_startup_indented = true;
   };
 }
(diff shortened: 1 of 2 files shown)


MattSturgeon:
This is unnecessary because `settings` is a freeform option. This means users can configure _anything_, even if we don't declare a corresponding sub-option.

This will add the sub-option to our docs, and will enforce type-checking (i.e. checking it is a boolean), but these are both double-edged swords because it is more surface area for us to keep in sync with any changes in the upstream plugin.

refaelsh:
> it is more surface area for us to keep in sync with any changes in the upstream plugin.

I understand. Then I will close this PR.

But I have another question please: maybe I should then open a PR that rempove this https://github.com/nix-community/nixvim/blob/18b7597e6ca4b98a6c3f20ddc9783165d5998018/plugins/languages/nvim-orgmode.nix#L17 and this https://github.com/nix-community/nixvim/blob/18b7597e6ca4b98a6c3f20ddc9783165d5998018/plugins/languages/nvim-orgmode.nix#L20?

MattSturgeon:
There's no hard and fast rules currently for whether or not we should have any particular settings-options.

You're more than welcome to re-open this PR if you think it's useful to have the sub-option declared in nixvim. Alternatively, we could add it to the `settingsExample` _without_ declaring an option.

As for the options you've already added, I think they're fine. There's plenty of other plugins that have _far more_ settings-options declared than this one.

I don't mind if/how you want to continue with this PR, I just wanted to ensure you were aware that `settings` was freeform before proceeding 😁

refaelsh:
> Alternatively, we could add it to the settingsExample without declaring an option.

I prefer to remove the 2 options but leave them in the `settingsExample`.

> I don't mind if/how you want to continue with this PR

I would prefer to leave it closed if thats ok by you.

> I just wanted to ensure you were aware that settings was freeform before proceeding 😁

Now that I am aware, I wanna remove the 2 options I added. If thats ok by you, I will open a PR that removes them.

PR #2153, opened 2024-09-03
plugins/dracula.nvim: init

refaelsh:
This PR adds the dracula.nvim plugin.
Here the link to nixpkgs: https://search.nixos.org/packages?channel=24.05&show=vimPlugins.dracula-nvim&from=0&size=50&sort=relevance&type=packages&query=dracula-nvim.

Thanks in advance for reviewing :-)

### DIFF
--- a/plugins/colorschemes/dracula-nvim.nix
+++ b/plugins/colorschemes/dracula-nvim.nix
@@ -1,10 +1,9 @@
 {
   lib,
-  config,
   pkgs,
   ...
 }:
-lib.nixvim.neovim-plugin.mkNeovimPlugin config {
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
   name = "dracula";
   originalName = "dracula.nvim ";
   defaultPackage = pkgs.vimPlugins.dracula-nvim;
--- a/plugins/colorschemes/dracula-nvim.nix
+++ b/plugins/colorschemes/dracula-nvim.nix
@@ -1,0 +1,20 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+lib.nixvim.neovim-plugin.mkNeovimPlugin config {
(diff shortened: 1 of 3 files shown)


MattSturgeon (approved):
Awesome, lgtm

MattSturgeon:
@mergifyio queue

plugins/colorschemes/dracula-nvim.nix
```diff
@@ -0,0 +1,19 @@
...
+  defaultPackage = pkgs.vimPlugins.dracula-nvim;
+
+  maintainers = [ lib.nixvim.maintainers.refaelsh ];
+
+  settingsExample = {
+    settings = {
+      italic_comment = true;
+      colors.green = "#00FF00";
+    };
+  };
```
MattSturgeon, line 18: Ah, I'm an idiot. This should've not had a nested `settings` within the example.

Also it seems I missed that there were unsquashed fixup commits.

refaelsh:
> Ah, I'm an idiot. This should've not had a nested settings within the example.

Should I push the fix to `add-dracula-nvim3` or open a new PR?

> Also it seems I missed that there were unsquashed fixup commits.

Does not seem critical to me but:
1. I will take a not of it next time.
2. Its too late to do something about it, right?

khaneliman:
> > Ah, I'm an idiot. This should've not had a nested settings within the example.
> 
> Should I push the fix to `add-dracula-nvim3` or open a new PR?
> 
Was handled in https://github.com/nix-community/nixvim/pull/2156
> > Also it seems I missed that there were unsquashed fixup commits.
> 
> Does not seem critical to me but:
> 
>     1. I will take a not of it next time.
> 
>     2. Its too late to do something about it, right?

Not without rewriting git history, which is highly discouraged on the trunk branch. So we just have to leave it be.

PR #2155, opened 2024-09-03
Revert "Remove config"

GaetanLepage:
This reverts commit 3f9cf9f961ba734d85021248c01b38cd8f2aa173.

The original change (https://github.com/nix-community/nixvim/pull/2153) was wrong in several aspects:
- colorscheme is not declared as such and is then under `plugins.dracula` instead of being `colorschemes.dracula`
- the latter already exists and is based on

### DIFF
--- a/plugins/colorschemes/dracula-nvim.nix
+++ b/plugins/colorschemes/dracula-nvim.nix
@@ -1,9 +1,10 @@
 {
   lib,
+  config,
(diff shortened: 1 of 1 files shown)


traxys:
This did not revert the correct commit.
I don't think we can say the plugin has no value, it brings tree sitter highlight groups

PR #2158, opened 2024-09-04
plugins/navic: migrate to mkNeovimPlugin

khaneliman:


plugins/bufferlines/navic.nix
```diff
@@ -45,12 +66,12 @@ in
     };
 
     lsp = {
-      autoAttach = helpers.defaultNullOpts.mkBool false ''
+      autoAttach = defaultNullOpts.mkBool false ''
```
MattSturgeon, line 69: It seems option names haven't been converted to snake_case. [Checking upstream](https://github.com/SmiteshP/nvim-navic?tab=readme-ov-file#-customise), it seems they should be.

MattSturgeon (changes requested):

plugins/bufferlines/navic.nix
khaneliman: i think I accidentally left it while testing the deprecations..

MattSturgeon (approved):
:heart:

khaneliman:
@mergifyio queue

PR #2159, opened 2024-09-04
plugins/barbecue: migrate to mkNeovimPlugin

khaneliman:


plugins/bufferlines/barbecue.nix
```diff
@@ -1,34 +1,72 @@
...
-in
-{
-  options.plugins.barbecue = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "barbecue-nvim";
+  inherit (lib.nixvim) defaultNullOpts;
 
-    package = helpers.mkPluginPackageOption "barbecue-nvim" pkgs.vimPlugins.barbecue-nvim;
-
-    attachNavic = helpers.defaultNullOpts.mkBool true ''
+  mkListStr = defaultNullOpts.mkNullable (lib.types.listOf lib.types.str);
```
MattSturgeon, line 9: 1. Is this abstraction really needed?
2. This doesn't support rawLua

I think we can just use `mkListOf types.str` throughout?

MattSturgeon (commented):
No issues with the changes, but I don't think we should keep `mkListStr`. At least not as-is.

MattSturgeon (approved):
Can't see any issues :grinning:

khaneliman:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #2150, opened 2024-09-03
treewide: migrate to `lib.mkPackageOption`

MattSturgeon:
[Adds a test](https://github.com/nix-community/nixvim/pull/2150/files#diff-4824391e6ed74ff5dccc7c7e96859742e3054a7a6531351378a9f576d864630c) that ensures "package" options use a "literalExpression" in their defaultText; i.e. it validates `lib.mkPackageOption` was used to build the package options.

All options whose `default` is a derivation are covered by the test, other than submodule sub-options.

This PR completely removes `lib.nixvim.mkPackageOption` and `lib.nixvim.mkPluginPackageOption` in favour of using `lib.mkPackageOption` treewide.

<details><summary>Example build log produced by the test _without_ the other fixes</summary>
<p>

```
structuredAttrs is enabled
Found 433 issues:
- colorschemes.dracula.package (<derivation vimplugin-dracula-vim-2024-07-26>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/plugins/colorschemes/dracula.nix

- package (<derivation neovim-unwrapped-0.10.1>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/modules/top-level/output.nix

- plugins.alpha.iconsPackage (<derivation vimplugin-nvim-web-devicons-2024-08-04>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/plugins/utils/alpha.nix

- plugins.alpha.package (<derivation vimplugin-alpha-nvim-2024-02-25>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/plugins/utils/alpha.nix

- plugins.auto-session.package (<derivation vimplugin-auto-session-2024-08-04>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/plugins/utils/auto-session.nix

- plugins.autoclose.package (<derivation vimplugin-autoclose.nvim-2024-07-21>), declared in:
  - /nix/store/56ri0qrf22klkh6nnpqhmzv7z8jdvcwd-source/plugins/utils/autoclose.nix

- plugins.barbar.iconsPackage (<derivation vimplugin-nvim-web-devicons-2024-08-04>), declared in:
[...]

PR #2160, opened 2024-09-04
plugins/{efmls,none-ls}: use `mkPackageOption` and/or `defaultText`

MattSturgeon:
- **plugins/efmls: use `mkPackageOption`**
- **plugins/none-ls: use `defaultText` in package options**

Extracted from #2150 to reduce its scope.

This PR refactors the efmls and none-ls package list files to use package names/paths instead of _actual_ package derivations. This then allows us to show a useful `defaultText` in the resulting package options.

### DIFF
--- a/plugins/lsp/language-servers/efmls-configs-pkgs.nix
+++ b/plugins/lsp/language-servers/efmls-configs-pkgs.nix
@@ -1,4 +1,8 @@
-lib: {
+lib:
+let
+  inherit (import ../../../lib/pkg-lists.nix lib) topLevel scoped;
+in
+{
   # efmls-configs tools that have no corresponding nixpkgs package
   unpackaged = [
     "blade_formatter"
@@ -28,20 +32,6 @@ lib: {
 
   # Mapping from a efmls-configs tool name to the corresponding nixpkgs package
   packaged =
-    let
-      # TODO: move these helpers to a shared location so that none-ls can also use them
-      topLevel = names: lib.genAttrs names lib.id;
-      scoped = lib.concatMapAttrs (
-        scope: v:
-        if builtins.isAttrs v then
-          lib.mapAttrs (_: loc: [ scope ] ++ loc) (scoped v)
-        else
-          lib.genAttrs (lib.toList v) (name: [
-            scope
-            name
-          ])
-      );
-    in
     # Top-level packages
     topLevel [
       "actionlint"
--- a/plugins/lsp/language-servers/efmls-configs-pkgs.nix
+++ b/plugins/lsp/language-servers/efmls-configs-pkgs.nix
@@ -1,4 +1,4 @@
-pkgs: {
+lib: {
   # efmls-configs tools that have no corresponding nixpkgs package
   unpackaged = [
     "blade_formatter"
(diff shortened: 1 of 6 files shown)


tests/generated.nix
MattSturgeon, line 1: We may want to add a test here to ensure all listed packages actually exist in `pkgs`.

plugins/lsp/language-servers/efmls-configs-pkgs.nix
```diff
@@ -27,115 +27,145 @@ pkgs: {
...
+    ] lib.id
+    # Scoped packages
+    //
+      lib.concatMapAttrs
+        (
+          scope: names:
+          lib.genAttrs names (name: [
+            scope
+            name
+          ])
```
traxys, line 99: This does not seem right, it generates a list of two names, but that results in two packages named `nodePackges` and `eslint_d` trying to get installed for example.

I'm not sure what is the right syntax hele, `nodePackages.eslint_t` ?

tests/generated.nix
traxys: Don't we have a test that enables all the sources, and this will result in an install of all the packages?

MattSturgeon: Correct, but because we can't use `or` fallbacks in the pkgs list anymore, I added a fallback here: https://github.com/nix-community/nixvim/pull/2160/files#diff-688c596fe956b7b2a491279b4b53c180e111d19a1bb5d20fc73060cead254374R61-R62

So I'm not sure the tests would actually reliably fail for missing packages anymore.

Maybe I should tackle that fallback in another PR though and just let it fail to eval in this one?

plugins/lsp/language-servers/efmls-configs-pkgs.nix
MattSturgeon: I'll have to look into it in the repl. The intended behaviour is to produce `[ "nodePackagse" "eslint_t" ]`, e.g:

```nix
{
  autopep8 = [ "python3" "autopep8" ];
  eslint_t = [ "nodePackagse" "eslint_t" ];
  prettier =  [ "nodePackagse" "prettier" ];
}
```

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -7,11 +7,15 @@
...
   tools = import ../../../generated/efmls-configs.nix;
-  packagingStatus = import ./efmls-configs-pkgs.nix pkgs;
+  packagingStatus = import ./efmls-configs-pkgs.nix lib;
 
   # Filter packages that are not compatible with the current platform
   toolPkgs = lib.filterAttrs (
-    a: pkg: lib.meta.availableOn pkgs.stdenv.hostPlatform pkg
+    name: loc:
+    let
+      pkg = lib.attrByPath (lib.toList loc) null pkgs;
```
MattSturgeon, line 16: If I add a warning to the `null` fallback here, I get to the bottom of the issue in https://github.com/nix-community/nixvim/pull/2160#discussion_r1743722663:
```suggestion
      loc' = lib.toList loc;
      pkg = lib.attrByPath loc' (lib.warn "${lib.concatStringsSep "." loc'} cannot be found in pkgs!" null) pkgs;
```

```
trace: evaluation warning: python3.autopep8 cannot be found in pkgs!
trace: evaluation warning: eslint cannot be found in pkgs!
trace: evaluation warning: python3.flake8 cannot be found in pkgs!
trace: evaluation warning: python3.mdformat cannot be found in pkgs!
trace: evaluation warning: python3.vulture cannot be found in pkgs!
trace: evaluation warning: python3.autopep8 cannot be found in pkgs!
trace: evaluation warning: eslint cannot be found in pkgs!
trace: evaluation warning: python3.flake8 cannot be found in pkgs!
trace: evaluation warning: python3.mdformat cannot be found in pkgs!
trace: evaluation warning: python3.vulture cannot be found in pkgs!
trace: evaluation warning: python3.autopep8 cannot be found in pkgs!
trace: evaluation warning: eslint cannot be found in pkgs!
trace: evaluation warning: python3.flake8 cannot be found in pkgs!
trace: evaluation warning: python3.mdformat cannot be found in pkgs!
trace: evaluation warning: python3.vulture cannot be found in pkgs!
```

plugins/lsp/language-servers/efmls-configs.nix
```diff
@@ -28,13 +23,19 @@ in
...
+      (lib.attrsets.mapAttrs (
+        name: loc:
+        lib.mkPackageOption pkgs name {
+          nullable = true;
+          default = loc;
+        }
+      ))
+      # Filter packages that are not compatible with the current platform
+      # FIXME: shouldn't we still have options even when the default isn't available??
+      (lib.filterAttrs (_: opt: lib.meta.availableOn pkgs.stdenv.hostPlatform opt.default))
```
MattSturgeon, line 37: I think we should still have options, but maybe remove the default?

```suggestion
      lib.mapAttrs (_: opt:
        if lib.meta.availableOn pkgs.stdenv.hostPlatform opt.default then
          opt
        else
          removeAttrs opt [ "defaultText" ] // { default = null; }
      )
```

plugins/lsp/language-servers/efmls-configs.nix
MattSturgeon: Fixed by allowing "scope" to be a nested attr name and moving `python3` packages to `python3.pks`.

tests/generated.nix
traxys: Yeah, we should catch that in the update PR?

plugins/lsp/language-servers/efmls-configs.nix
traxys: Yeah, that would indicate to the user that the package should be managed externally

plugins/lsp/language-servers/efmls-configs-pkgs.nix
```diff
@@ -27,115 +31,135 @@ pkgs: {
...
+    # Scoped packages
+    // scoped {
+      python3.pkgs = [
+        "autopep8"
+        "flake8"
+        "mdformat"
+        "vulture"
+      ];
+      nodePackages = [
+        "eslint" # FIXME: No way to have a transition fallback...
```
traxys, line 104: I don't understand the comment, was it renamed?

plugins/lsp/language-servers/efmls-configs.nix
traxys: (I feel like it's pretty rude to  have differing options per host platform)

traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2161, opened 2024-09-04
plugins/codeium-nvim: migrate to mkNeovimPlugin

khaneliman:


plugins/completion/cmp/sources/codeium-nvim.nix
```diff
@@ -1,85 +1,96 @@
...
-  cfg = config.plugins.codeium-nvim;
+  inherit (lib.nixvim) defaultNullOpts;
 in
-{
-  meta.maintainers = [ maintainers.GaetanLepage ];
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "codeium-nvim";
+  originalName = "codeium.nvim";
+  luaName = "codeium";
+  defaultPackage = pkgs.vimPlugins.codeium-nvim;
```
MattSturgeon, line 13:
```suggestion
```

MattSturgeon: If the package attrname was different to `name`, we could specify it as a string to `package`.

khaneliman:
Any idea on a workaround for the enable option ?

```bash
error: The option `plugins.codeium-nvim.enable' in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources/codeium-nvim.nix' is already declared in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources'.
Test failure
```

MattSturgeon:
> Any idea on a workaround for the enable option ?
> 
> ```shell
> error: The option `plugins.codeium-nvim.enable' in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources/codeium-nvim.nix' is already declared in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources'.
> Test failure
> ```

We should remove `codeium-nvim` from `default.nix`. We can achieve the same thing by adding `imports = [
 { cmpSourcePlugins.codeium-nvim = "codeium"; }
]`

khaneliman:
> > Any idea on a workaround for the enable option ?
> > ```shell
> > error: The option `plugins.codeium-nvim.enable' in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources/codeium-nvim.nix' is already declared in `/nix/store/mn24kg30ai6bd043w542n157qwk41h4n-source/plugins/completion/cmp/sources'.
> > Test failure
> > ```
> 
> We should remove `codeium-nvim` from `default.nix`. We can achieve the same thing by adding
> 
> ```nix
> imports = [
>  # Register nvim-cmp association
>  { cmpSourcePlugins.codeium-nvim = "codeium"; }
> ]
> ```

If I do that it fails to find the plugin, if i swap it to 
```nix
    { cmpSourcePlugins.codeium = "codeium-nvim"; }
```
I get errors on renamed properties because the renamedOptionsToSettings doesn't get loaded

plugins/completion/cmp/sources/default.nix
```diff
@@ -197,7 +197,7 @@ in
 {
   # For extra cmp plugins
   imports = [
-    ./codeium-nvim.nix
```
MattSturgeon, line 200: We still need to import `codeium-nvim.nix` here

plugins/completion/cmp/sources/default.nix
```diff
@@ -197,7 +197,7 @@ in
...
+    { cmpSourcePlugins.codeium = "codeium-nvim"; }
```
MattSturgeon, line 200: This could be done here, but I think it makes more sense to do this within the `codeium-nvim.nix` module.

plugins/completion/cmp/sources/default.nix
```diff
@@ -197,7 +197,7 @@ in
 {
```
MattSturgeon, line 197: You need to remove the `{ pluginName = "codeium-nvim"; sourceName = "codium"; }` attrs from the list above.

Otherwise we're creating the enable/package option twice, once using `mkCmpSourcePlugin` and once using `mkNeovimPlugin`.

khaneliman: Doh! That was the thing that threw me off the wrong path...

khaneliman: Thank you very much!

MattSturgeon:
Should we consider moving the plugin out of `plugins/completion/cmp/sources/` and up to `plugins/completion/`? Since cmp/sources is no longer responsible for building the plugin module...

This would be the second cmp-source plugin to be defined outside of `plugins/completion/cmp/sources/`; the other one being `otter`.

plugins/completion/codeium-nvim.nix
```diff
@@ -0,0 +1,116 @@
...
+    } "The path to the directory where the Codeium server will be downloaded to.";
+
+    api = {
+      host = defaultNullOpts.mkStr "server.codeium.com" ''
+        The hostname of the API server to use.
+      '';
+
+      port = defaultNullOpts.mkPositiveInt 443 ''
+        The port of the API server to use.
+      '';
```
MattSturgeon, line 74: The old option was (effectively) `coercedTo ints.positive toString str`.

https://github.com/nix-community/nixvim/blob/17f4aa555615df37ec68ed75a04b8a8b1d36fd98/plugins/completion/cmp/sources/codeium-nvim.nix#L66

1. Does the plugin support lua numbers as ports, or does it expect numeric strings?
2. If end-users had string definitions in their old config, these will now fail the type-check.

MattSturgeon (commented):
Looks good overall :rocket:

plugins/completion/codeium-nvim.nix
khaneliman: By looking at their config example and the usage, I would think just a string? 

https://github.com/Exafunction/codeium.nvim/blob/f6a2ef32a9e923cb0104a19d3e426b0e40e49505/lua/codeium/api.lua#L167-L173

Should we limit it to just strs or keep handling the number?

MattSturgeon: Ideally we should match the upstream type, in this case it'd be `types.strMatching "[0-9]+"`, I think?

Because we have the rename, we need to still _support_ ints, but we can do this using `transitionType`, right?

plugins/completion/codeium-nvim.nix
```diff
@@ -0,0 +1,121 @@
...
+      __raw = "vim.fn.stdpath('cache') .. '/codeium/bin'";
+    } "The path to the directory where the Codeium server will be downloaded to.";
+
+    api = {
+      host = defaultNullOpts.mkStr "server.codeium.com" ''
+        The hostname of the API server to use.
+      '';
+
+      port = defaultNullOpts.mkNullable' {
+        type = with lib.types; coercedTo ints.positive toString (strMatching "[0-9]+");
```
MattSturgeon, line 74: `coercedTo` vs `transitionType`; i.e. should we by deprecating the type coercion?

MattSturgeon: Also, this doesn't have `rawLua` support. We could just use `mkNullableWithRaw'` I think? I don't _think_ that'd break the coercion...

khaneliman: Ah, yeah... forgot that's the benefit of transitionType. I'll change that.

plugins/completion/codeium-nvim.nix
```diff
@@ -0,0 +1,121 @@
...
+      __raw = "vim.fn.stdpath('cache') .. '/codeium/bin'";
+    } "The path to the directory where the Codeium server will be downloaded to.";
+
+    api = {
+      host = defaultNullOpts.mkStr "server.codeium.com" ''
+        The hostname of the API server to use.
+      '';
+
+      port = defaultNullOpts.mkNullableWithRaw' {
+        type = with lib.types; lib.nixvim.transitionType ints.positive toString (strMatching "[0-9]+");
```
MattSturgeon, line 74:
```suggestion
        # TODO: Added 2024-09-05; remove after 24.11
        type = with lib.types; lib.nixvim.transitionType ints.positive toString (strMatching "[0-9]+");
```

MattSturgeon: Also, is it `lib.nixvim.transitionType` or `lib.types.transitionType`? I guess if tests pass, this _is_ correct.

khaneliman:
@mergifyio queue

khaneliman:
@mergifyio refresh

khaneliman:
@mergifyio queue

PR #2164, opened 2024-09-04
plugins/by-name: init

MattSturgeon:
Add support for automatically importing any directories under `plugins/by-name`.

## Documentation

- [x] update CONTRIBUTING.md
- [ ] add a `plugins/by-name/README.md` (maybe?)

## Validation

A test is included that does some validation on the by-name directory and plugin-options\* enabled within it:

- [x] run the validation tests in the git pre-commit hook
- [x] by-name should contain only directories
- [x] plugin name should match directory name where its options are declared
- [x] option path starts with either "colorschemes" or "plugins"
- [x] Ensure a matching testfile exists in the correct location
- [x] by-name plugins should have a `default.nix`
  - This should be detected anyway as the import will fail if there's no `default.nix` file
- [ ] one plugin per directory
  - This seems an unnecessary check. It's probably not worth dealing with the complexity...

We could also have a CI workflow running on new PRs only, that validates whether new plugins are added to `by-name` instead of a category.

_\* "Plugin options" are collected by looking for options matching `*.*.enable`_

## Migration

See https://github.com/nix-community/nixvim/pull/2164#issuecomment-2329482663

**I'd like an overall consensus on this before working on such migrations.** See "Design decisions" below.

## Future tooling

I think it'd be useful to have a devshell command that interactively builds a new plugin module based on the template.

You would be prompted for things like "type" (vim, lua, etc) "name", "pkg name" "setup fn", "globals prefix", and it'd automatically produce a minimal module file.

```
$ create-plugin
What's the plugin's name? 
> hello

What's the plugin type? (1. neovim (i.e. `require().setup()`) 2. vim with globals 3. vim without settings)
> 1

Current configuration:
1. name = "hello"
2. type = lua/neovim with setup function
3. luaName = "hello"
4. pkg = pkgs.vimPlugins.hello
5. setup = ".setup"

Would you like to change anything? (type a line number to modify it, or <return> to finish)
>
```

## Design decisions

- We don't need a prefix directory (like in nixpkgs) because we are dealing with a smaller scale
  - By prefix, I mean the `"na"` directory in `pkgs/by-name/na/name/package.nix`.
  - We can just have `plugins/by-name/<name>/default.nix` to keep things simple.
  - Nixpkgs has around 500 prefixes alone in `pkgs/by-name`, while we have around 200 files total matching `plugins/*/*.nix`.
  - Having a prefix level would complicate both the implementation and usage, e.g. actually searching for a plugin's module.
  - Tab completion for `plugins/by-name/examp<TAB>` should be easier without a prefix.
- We shouldn't need to add by-name plugins to an imports list
  - It is trivial to read the directory and fold that into a list of paths to import.
  - This is one less thing for new contributors to worry about when adding a plugin module.
- We should exclusively use directories with `default.nix` files.
  - This adds consistency.
  - It is one less judgement call to make on new plugins (i.e. whether to use `foo/default.nix` or `foo.nix`)
  - Having the plugin already in its own directory makes future expansion easier: https://github.com/nix-community/nixvim/pull/2164#issuecomment-2329862507
  - This simplifies the implementation; we only need to handle `type == "directory"` files.
- Some categories should remain
  - e.g. where we build many plugin modules with a specialized helper, such as mkLsp or mkCmpPlugin
  - I have no opinion on whether these should be in or out of by-name, but I have no immediate plans to move them.
  - Plugins outside of `by-name` should be imported manually in `plugins/default.nix`.


khaneliman:
My opinion was that we'd just do a migration over to the new structure right away.

MattSturgeon:
> My opinion was that we'd just do a migration over to the new structure right away.

Sure, but I'd like the rest of the design decisions discussed & agreed on before investing time into that.

khaneliman:
> ## Design decisions
> 
>     * We don't need a prefix directory (like in nixpkgs) because we are dealing with a smaller scale
>       
>       * By prefix, I mean the `"na"` directory in `pkgs/by-name/na/name/package.nix`.
>       * We can just have `plugins/by-name/<name>/default.nix` to keep things simple.
>       * Nixpkgs has around 500 prefixes alone in `pkgs/by-name`, while we have around 200 files total matching `plugins/*/*.nix`.
>       * Having a prefix level would complicate both the implementation and usage, e.g. actually searching for a plugin's module.
>       * Tab completion for `plugins/by-name/examp<TAB>` should be easier without a prefix.
> 

I think this makes sense, we don't have a large enough scale project to warrant much (if any) nesting. The only thing I could see is maybe doing alphabetical nesting on first character if we were really worried about too large of a directory, for whatever reason. 

>     * We shouldn't need to add by-name plugins to an imports list
>       
>       * It is trivial to read the directory and fold that into a list of paths to import.
>       * This is one less thing for new contributors to worry about when adding a plugin module.
> 
>     * We should exclusively use directories with `default.nix` files.
>       
>       * This adds consistency.
>       * It is one less judgement call to make on new plugins (i.e. whether to use `foo/default.nix` or `foo.nix`)
>       * This simplifies the implementation; we only need to handle `type == "directory"` files.

Completely agree, I think we talked about this previously, but it will make it consistent in finding a plugin and not have to worry about sorting when simple plugins are still using directories instead of a single file.

MattSturgeon:
Ok, I've included an initial round of migration.

<p>

```
# First, move the loose files
for f in plugins/utils/*.nix
    set name (path change-extension '' $f | path basename)
    mkdir -p plugins/by-name/$name
    git mv $f plugins/by-name/$name/default.nix
end

# next, move the remaining dirs over
for f in plugins/utils/*
    git mv $f plugins/by-name/
end

# Finally, remove the old imports in `default.nix`
vim plugins/default.nix

# We also need to move the test files, for consistency:
for f in tests/test-sources/plugins/utils/*.nix
    set name (path change-extension '' $f | path basename)
    mkdir -p tests/test-sources/plugins/by-name/$name
    git mv $f tests/test-sources/plugins/by-name/$name/default.nix
end
for f in tests/test-sources/plugins/utils/*
    git mv $f tests/test-sources/plugins/by-name/
end
```

</p>
</details>

modules/plugins.nix
```diff
@@ -1,4 +1,17 @@
...
+  inherit (builtins) readDir pathExists;
+  inherit (lib.attrsets) foldlAttrs;
+  inherit (lib.lists) optional optionals;
+  by-name = ../plugins/by-name;
+in
 {
-  imports = [ ../plugins/default.nix ];
+  imports =
+    [ ../plugins ]
+    ++ optionals (pathExists by-name) (
```
MattSturgeon, line 11: Now that `by-name` is in use, shall we remove the `pathExists` check?

khaneliman:
Yeah, initial impression is that directory layout should be fine. I'm curious what @traxys and @GaetanLepage think. I think utils was our largest directory of plugins and that being flattened into by-name isn't that large.

traxys:
> We should exclusively use directories with default.nix files. 
Another advantage is that it allows to easily split the default.nix file in sub-nix files if need arises, wheras if you had `<name>.nix` you would need to create  `<name>/default.nix`

traxys:
I agree with all the design choices

MattSturgeon:
How far should this PR go with migrating existing plugins to `by-name`?

We already have enough migrated to demonstrate it working, and I don't want to spend too long deciding which plugins should/should not be moved over.

khaneliman:
> How far should this PR go with migrating existing plugins to `by-name`?
> 
> We already have enough migrated to demonstrate it working, and I don't want to spend too long deciding which plugins should/should not be moved over.

If we're going to be leaving some behind out of by-name, I'd say wherever you feel comfortable stopping. The ones that get left behind we can decide if they should be migrated or not in separate discussions. My original thought was that it was gonna be an all out migration, but I do see the notes about some things not easily migrated, due to the shared code and helpers. So, I think it's fair to just knock out the bulk ones and niche ones can be done later, if it makes sense.

khaneliman:
> ## Future tooling
> 
> I think it'd be useful to have a devshell command that interactively builds a new plugin module based on the template.
> 
> You would be prompted for things like "type" (vim, lua, etc) "name", "pkg name" "setup fn", "globals prefix", and it'd automatically produce a minimal module file.
> 
> ```
> $ create-plugin
> What's the plugin's name? 
> > hello
> 
> What's the plugin type? (1. neovim (i.e. `require().setup()`) 2. vim with globals 3. vim without settings)
> > 1
> 
> Current configuration:
> 1. name = "hello"
> 2. type = lua/neovim with setup function
> 3. luaName = "hello"
> 4. pkg = pkgs.vimPlugins.hello
> 5. setup = ".setup"
> 
> Would you like to change anything? (type a line number to modify it, or <return> to finish)
> >
> ```
> 

I think this is a great idea! The internal tooling and helpers have gotten to the point it should be fairly simple to generate a skeleton module.

MattSturgeon:
```
> Found plugin modules with mismatched option & directory names (3)
> - plugins.coq-nvim is declared in 'coq'
> - plugins.cmp-ai is declared in 'cmp/sources'
> - plugins.nvim-cmp is declared in 'cmp/deprecations.nix'
```

MattSturgeon (commented):
This should be ready for review.

Note that @GaetanLepage has specifically asked to have a say before merging.

Ideally, I'd like approval from all maintainers on this one.

I appreciate it is a large PR, however the _vast_ majority of changes are simple renames with no changes to the files being moved.
The _actual_ changes are mostly in:
- CONTRIBUTING.md (updated docs)
- plugins/default.nix (removing imports)
- modules/plugins.nix (auto-import implementation)
- tests/plugin-by-name.nix (various tests)

Everything within `plugins/*` and `tests/test-sources/` should exclusively be moving files without actual modification.

I've limited the migration to the low-hanging fruit. Partly for my own sanity, partly to avoid controversy/distraction, partly to keep the _bulk_ part of the review simple.
(Reviewers would have to take more care if I migrated more complex plugins to by-name)

Hopefully the focus of the reviews can be on the design decisions, the extent of the tests (should they go further? additional tests?), any nit-picking with the implementation, and (of course) proof-reading the documentation.

cc @traxys @khaneliman

MattSturgeon:
I've added _very basic_ empty tests for the few plugins that didn't have any by running:

```fish
# fish
for name in autoclose commentary easyescape floaterm fugitive gitmessenger intellitab nvim-bqf nvim-colorizer surround treesitter-playground treesitter-refactor vim-matchup
    echo "adding test for $name"
    set dir tests/test-sources/plugins/by-name/$name
    mkdir -p $dir
    echo "{
      empty = {
        plugins.$name.enable = true;
      };
    }" > $dir/default.nix
    git add $dir/default.nix
end
```

With this, I felt comfortable adding a test to `tests/plugins-by-name.nix` that validates a `default.nix` file exists for all directories under `plugins/by-name` **and** a corresponding `tests/test-sources` file also exists.

khaneliman:
> I've added _very basic_ empty tests for the few plugins that didn't have any by running:
> 
> ```fish
> # fish
> for name in autoclose commentary easyescape floaterm fugitive gitmessenger intellitab nvim-bqf nvim-colorizer surround treesitter-playground treesitter-refactor vim-matchup
>     echo "adding test for $name"
>     set dir tests/test-sources/plugins/by-name/$name
>     mkdir -p $dir
>     echo "{
>       empty = {
>         plugins.$name.enable = true;
>       };
>     }" > $dir/default.nix
>     git add $dir/default.nix
> end
> ```
> 
> With this, I felt comfortable adding a test to `tests/plugins-by-name.nix` that validates a `default.nix` file exists for all directories under `plugins/by-name` **and** a corresponding `tests/test-sources` file also exists.

I love this concept. I was wondering about a good strategy recently for detecting a missing test file for a plugin when I had found a couple recently that were missing during some recent treewide changes. This will make it very easy to remind and enforce that we are validating the basic instantiation of a plugin.

GaetanLepage (approved):
Implem looks good to me.

Thank you very much, this will definitely increase efficiency.

GaetanLepage:
You have conflicts though.

MattSturgeon:
> You have conflicts though.

Yeah, the problem is the PR renames the vast majority of `plugins/`, so any new additions or changes will conflict. `git rebase` can usually resolve the conflict automatically though.

khaneliman (approved):
LGTM, I look forward to the convenience of by-name

tests/plugins-by-name.nix
```diff
@@ -0,0 +1,190 @@
...
+          echo
+        fi
+      }
+
+      showErrs 'symlinks' "''${symlink[@]}"
+      showErrs 'regular files' "''${regular[@]}"
+      showErrs 'unknown-type files' "''${unknown[@]}"
+
+      if (( $errs > 0 )); then
+        exit $errs
```
traxys, line 84: We should always `exit 1`, else if by bad luck we have 256 errors it would return a 0 status code

(Though this is _not_ a big problem, as we would not create an $out file, failing the build anyway)

tests/plugins-by-name.nix
```diff
@@ -0,0 +1,190 @@
...
+        ((++errs))
+        echo "The following (''${#missingTests[@]}) test files do not exist:"
+        for name in "''${missingTests[@]}"; do
+          echo "  - tests/test-sources/plugins/by-name/$name/default.nix"
+        done
+        echo
+      fi
+
+      if (( $errs > 0 )); then
+        exit $errs
```
traxys, line 123: Same

tests/plugins-by-name.nix
MattSturgeon: > On POSIX systems the standard exit code is 0 for success and any number from 1 to 255 for anything else

Ah, good catch!

traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2170, opened 2024-09-04
plugins/efmls: move `eslint` pkg ref to top-level

### LINKED ISSUE #2168: [BUG] Alias eslint is still in node-packages.nix
I just updated my flake lock, and found I can no longer build nixvim.

Using standalone and even without any `follows`, I get:

```
error: Alias eslint is still in node-packages.nix
```

I suspect this was caused by #2160, specifically [this line](https://github.com/nix-community/nixvim/pull/2160/files#diff-d701989168ac134a24e9192c153e76d80ef5d40f16431647914fe2711da98bddR104).

What I can't figure out is why CI is happy here, yet my config fails to build...

MattSturgeon:
Fixes #2168

I'm somewhat confused, when testing locally with `tests plugins-lsp-efmls-configs` I get `error: eslint cannot be found in pkgs`. However, when overriding my nixvim config's flake input to use this branch, this indeed fixes #2168...

See also https://github.com/NixOS/nixpkgs/pull/339641 ([tracker](https://nixpk.gs/pr-tracker.html?pr=339641)) and #2128

### DIFF
--- a/plugins/lsp/language-servers/efmls-configs-pkgs.nix
+++ b/plugins/lsp/language-servers/efmls-configs-pkgs.nix
@@ -53,6 +53,7 @@ in
       "djlint"
       "dmd"
       "dprint"
+      "eslint"
       "fish"
       "flawfinder"
       "fnlfmt"
@@ -101,7 +102,6 @@ in
         "vulture"
       ];
       nodePackages = [
-        "eslint" # FIXME: No way to have a transition fallback...
         "eslint_d"
         "prettier"
         "alex"


khaneliman:
I only see this error when I do a flake lock update for nixpkgs in nixvim or I have a follows from my flakes nixpkgs input on the latest unstable. This should be moved to https://github.com/nix-community/nixvim/pull/2127 so that it's handled in the correct flake lock revision.

MattSturgeon:
Fixed in #2127

PR #2174, opened 2024-09-05
plugins/lazy: add `performance` option

wrvsrx:


MattSturgeon:
Thanks for your contribution! This is better handled by moving to a _freeform `settings` option_, which is being worked on in #2082.

We could merge this, but then we'd have to add additional "rename" aliases in #2082, which I'm keen to avoid if possible...

cc @my7h3le

wrvsrx:
Thanks for response. I'll wait for #2082 to be merged.

wrvsrx:
Given that PR #2082 has been inactive for some time, I'd like to propose that my pull request be prioritized for merging.

wrvsrx:
ping @MattSturgeon.

MattSturgeon (approved):
This increases the number of options we'll need to alias & deprecate, so it feels like a step in the wrong direction.

But given #2082 is stalled and freeform settings do not seem to be happening imminently, I won't block this.

PR #2177, opened 2024-09-05
plugins/mini-icons: add mockDevIcons

khaneliman:
Supports mini.icons mocking the behavior of nvim-web-devicons.


plugins/utils/mini.nix
```diff
@@ -82,9 +85,13 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
-    extraConfigLua = lib.foldlAttrs (lines: name: config: ''
-      ${lines}
-      require(${lib.nixvim.toLuaObject "mini.${name}"}).setup(${lib.nixvim.toLuaObject config})
-    '') "" cfg.modules;
+    extraConfigLua =
+      lib.foldlAttrs (lines: name: config: ''
+        ${lines}
+        require(${lib.nixvim.toLuaObject "mini.${name}"}).setup(${lib.nixvim.toLuaObject config})
+      '') "" cfg.modules
+      + lib.optionalString ((lib.hasAttr "icons" cfg.modules) && cfg.mockDevIcons) ''
```
MattSturgeon, line 93: Nit: I think it is better to use language syntax when we don't need a (e.g. partially applied) function?

`(cfg.modules ? icons)`

plugins/utils/mini.nix
```diff
@@ -82,9 +85,13 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+        MiniIcons.mock_nvim_web_devicons()
```
MattSturgeon, line 94: Q: where does the `MiniIcons` name come from? I didn't notice us importing it, i.e. using `require()`.

plugins/utils/mini.nix
```diff
@@ -11,6 +11,9 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   package = "mini-nvim";
 
   extraOptions = {
+    mockDevIcons = lib.mkEnableOption "mockDevIcons" // {
+      description = "Whether to tell `mini.icons` to emulate `nvim-web-devicons` for plugins that don't natively support it.";
```
MattSturgeon, line 15: Should we explicitly mention that this can be used as an alternative to `plugins.web-devicons.enable`? E.g. as a footnote in another paragraph.

Also, I'm wondering if we should have:

```nix
defaultText = literalMD "`false` **NOTE**: This option is experimental and the default value may change without notice.";
```

MattSturgeon (approved):
Everything seems good. Minor comments/questions below, but no objections.

plugins/utils/mini.nix
khaneliman: Right above, we will do a require for `mini.icons` when the user supplies it in the attribute set. That's why I need to make sure that they added it before I call that.

plugins/utils/mini.nix
```diff
@@ -82,9 +90,13 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
-    extraConfigLua = lib.foldlAttrs (lines: name: config: ''
-      ${lines}
-      require(${lib.nixvim.toLuaObject "mini.${name}"}).setup(${lib.nixvim.toLuaObject config})
-    '') "" cfg.modules;
+    extraConfigLua =
+      lib.foldlAttrs (lines: name: config: ''
+        ${lines}
+        require(${lib.nixvim.toLuaObject "mini.${name}"}).setup(${lib.nixvim.toLuaObject config})
+      '') "" cfg.modules
+      + lib.optionalString ((cfg.modules ? icons) && cfg.mockDevIcons) ''
```
MattSturgeon, line 98:
```suggestion
      # `MiniIcons` is only in scope if we've called `require('mini.icons')` above
      + lib.optionalString (cfg.modules ? icons && cfg.mockDevIcons) ''
```

plugins/utils/mini.nix
```diff
@@ -82,9 +90,13 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   callSetup = false;
   hasSettings = false;
   extraConfig = cfg: {
```
MattSturgeon, line 92: I think we need an assertion, to ensure the option is used correctly:

`assertion = cfg.mockDevIcons -> cfg.mocules ? icon`

MattSturgeon (changes requested):

MattSturgeon (approved):
:heart:

khaneliman:
@mergifyio queue

PR #2184, opened 2024-09-06
plugins/treesitter-playground: move removal notice to treesitter plugin

MattSturgeon:
I noticed in #2164 that treesitter-playground exists as a plugin module, but is just a removal notice.

That's fine and great, however it doesn't make sense in the context of `plugins/by-name` IMO.

I think it makes logical sense to keep it in the treesitter plugin? However, it may be useful to have a "top-level" plugins deprecations file where removed plugins can be listed in a common location. I could revise this PR to do that, if desired.

I'm thinking something like `plugins/deprecation.nix`:
```nix
{ lib, ... }:
let
  removed = {
    treesitter-playground = ''
      The `treesitter-playground` plugin has been deprecated since the functionality is included in Neovim.
      Use:
      - `:Inspect` to show the highlight groups under the cursor
      - `:InspectTree` to show the parsed syntax tree ("TSPlayground")
      - `:PreviewQuery` to open the Query Editor (Nvim 0.10+)
    ''
  };
in
{
  imports = mapAttrsToList (name: lib.mkRemovedOptionModule [ "plugins" name ]) removed;
}
```


khaneliman (dismissed):

MattSturgeon:
As discussed in PMs, we'll end up needing something like this for nvim-osc52 soon. So I've amended the PR to use the alternate approach proposed above.

khaneliman (approved):
LGTM, as noted, we can utilize this for future top-level deprecations for plugins that become unmaintainable or get removed.

MattSturgeon:
@mergifyio queue

PR #2190, opened 2024-09-07
tests: set `_file` to avoid `<unknown-file>` messages

MattSturgeon:
Set `_file` in test-case modules, to avoid `<unknown-file>` messages.

Since the module system is importing an anonymous fnOrAttr module, instead of a `path` type module, it doesn't have enough context to figure out the module's file automatically.

Usually such non-path modules would just inherit their importer's `_file`, however this doesn't apply when passing modules directly to `lib.evalModules` since there is no importer to inherit from.

## Example test failurues

Adding `foo = ""` to a test case in `tests/test-sources/examples.nix` produces the following errors:

Without the change:
```
error: The option `foo' does not exist. Definition values:
- In `<unknown-file>': ""
```

With the change:
```
error: The option `foo' does not exist. Definition values:
- In `/nix/store/ss06gh6ig3nk0pvh1a7n5x1a3g47x91j-test-sources/examples.nix': ""
```

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -5,7 +5,7 @@
   pkgsUnfree,
 }:
 let
-  fetchTests = import ./fetch-tests.nix;
+  fetchTests = import ./fetch-tests.nix { inherit lib pkgs helpers; };
   test-derivation = import ../lib/tests.nix { inherit pkgs lib; };
   inherit (test-derivation) mkTestDerivationFromNixvimModule;
 
@@ -17,10 +17,7 @@ let
     };
 
   # List of files containing configurations
-  testFiles = fetchTests {
-    inherit lib pkgs helpers;
-    root = ./test-sources;
-  };
+  testFiles = fetchTests ./test-sources;
 
   exampleFiles = {
     name = "examples";
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -10,9 +10,13 @@ let
   inherit (test-derivation) mkTestDerivationFromNixvimModule;
 
   moduleToTest =
-    name: module:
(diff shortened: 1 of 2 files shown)


tests/default.nix
```diff
@@ -5,22 +5,23 @@
...
   inherit (test-derivation) mkTestDerivationFromNixvimModule;
 
   moduleToTest =
-    name: module:
+    file: name: module:
     mkTestDerivationFromNixvimModule {
-      inherit name module;
+      inherit name;
+      module = {
+        _file = file;
```
MattSturgeon, line 17:
```suggestion
        _file = "${file}:${name}";
```

This (hack) would cause the test-case name to also be shown in the filename, e.g.
```
error: The option `foo' does not exist. Definition values:
- In `/nix/store/ss06gh6ig3nk0pvh1a7n5x1a3g47x91j-test-sources/examples.nix:simple-plugin': ""
```

GaetanLepage (approved):
I am not sure to 100% understand this section of the code (I am not the author).
However, the implementation seems right to me and the fact that you have tested it is enough for me.

khaneliman (approved):
LGTM, tested locally and is awesome being able to pinpoint the test file that has the failing test case.

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio dequeue

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergify refresh

MattSturgeon:
@mergify queue

Issue #2197, opened 2024-09-08
none-ls: source packages missing 


MattSturgeon:
It seems some of the none-ls source plugins are not being found in `pkgs`:

```
evaluation warning: python3.pkgs.cfn-lint cannot be found in pkgs!
evaluation warning: phpPackages.php-codesnifferpkgs cannot be found in pkgs!
evaluation warning: qt6.qtdeclarativepkgs cannot be found in pkgs!
evaluation warning: luaPackages.tl cannot be found in pkgs!
evaluation warning: rubyPackages.erb-formatterpkgs cannot be found in pkgs!
evaluation warning: phpPackages.php-codesnifferpkgs cannot be found in pkgs!
evaluation warning: phpPackages.php-cs-fixerpkgs cannot be found in pkgs!
evaluation warning: prisma cannot be found in pkgs!
evaluation warning: qt6.qtdeclarativepkgs cannot be found in pkgs!
```

https://github.com/nix-community/nixvim/blob/e1a5e39eb9d863f2ba8ec7d00eaa7f4ea29f688c/plugins/none-ls/packages.nix#L180

I think we need to add a check for `hasAtteByPath` in our `generated` test to avoid this regression again.

khaneliman:
Prisma one doesn't make sense.. that's in nixos-unstable. Unless our lockfile is just behind that change, still?

MattSturgeon:
After #2198 we're still left with:

```
evaluation warning: python3.pkgs.cfn-lint cannot be found in pkgs!
evaluation warning: luaPackages.tl cannot be found in pkgs!
evaluation warning: prisma cannot be found in pkgs!
```

khaneliman:
So the first 2 I can't explain, yet. The third is just because our current nixpkgs revision is from before it was moved from nodePackages to top-level.

MattSturgeon:
This one isn't using a list-type path:

https://github.com/nix-community/nixvim/blob/e1a5e39eb9d863f2ba8ec7d00eaa7f4ea29f688c/plugins/none-ls/packages.nix#L127

khaneliman:
> This one isn't using a list-type path:
> 
> https://github.com/nix-community/nixvim/blob/e1a5e39eb9d863f2ba8ec7d00eaa7f4ea29f688c/plugins/none-ls/packages.nix#L127
> 
> Same:
> 
[...]

PR #2198, opened 2024-09-08
plugins/none-ls: fix pkg names

khaneliman:
Partially Resolves:
 https://github.com/nix-community/nixvim/issues/2197

i think some of these just got messed up with a regex change or something

Prisma is due to the nixpkgs revision, but the other 2 I'm not sure... I can checkout that branch and run a nix build of them.


MattSturgeon (approved):
Oops!

khaneliman:
@mergifyio queue

PR #2199, opened 2024-09-09
plugins/none-ls: fix pkg names

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/2198 and finishes fixing the naming issues in https://github.com/nix-community/nixvim/issues/2197


MattSturgeon (approved):
Thanks for fixing my bugs... Let's keep the issue open until we have a regression test though.

khaneliman:
@mergifyio queue

Issue #1258, opened 2024-03-13
[BUG] Enabling plugins.lsp.rust-analyzer and rustaceanvim spawns 2 LSPs


alisonjenkins:
| Field        | Description |
|--------------|-------------|
| Plugin       | `rustaceanvim`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `-` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

If you have both lsp.rust-analyzer.enable set to true and are using rustaceanvim and work on a rust project 2 instances of rust-analyzer will spawn. Additionally the one used by rustaceanvim does not appear to honour the LSP config provided to the plugin.

I discovered this as I am trying to ignore the `.direnv` folder in my Rust projects as rust-analyzer gets stuck on scanning for roots on it indefinitely using this config.

To reproduce after setting up this config, edit a Rust project and do ```:LspInfo``` you will see 2 rust-analyzers. One which is just running ```rust-analyzer``` and another that is using the absolute path to an instance of ```rust-analyzer``` in the Nix store. The one using the absolute path is the lsp.servers one, the one using just rust-analyzer is the rustaceanvim instance.

I have not yet tested disabling rustaceanvim's instance of rust-analyzer to see if I can get the lsp.servers instance of rust-analyzer to take note of the config provided but the rustaceanvim one does not seem to honour the ignoreDirs option.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  plugins = {
    rustaceanvim = {
      enable = true;

      server.settings = {
        files = {
          excludeDirs = [ ".direnv" ];
        };
      };
    };

    lsp = {
      enable = true;

      servers = {
        rust-analyzer = {
          enable = true;
          installRustc = true;
          installCargo = true;
        };
      };
    };
  };
}
```

traxys:
Well the upstream plugin mentions:

> Warning
>
> Do not call the [nvim-lspconfig.rust_analyzer](https://github.com/neovim/nvim-
> lspconfig) setup or set up the lsp client for rust-analyzer manually, as doing so may cause conflicts.

alisonjenkins:
> Well the upstream plugin mentions:
> 
> > Warning
[...]

PR #2202, opened 2024-09-09
plugins/rustaceanvim: add lsp assertion

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1258

### DIFF
--- a/plugins/languages/rust/rustaceanvim/default.nix
+++ b/plugins/languages/rust/rustaceanvim/default.nix
@@ -56,6 +56,16 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
         globals.rustaceanvim = cfg.settings;
 
+        assertions = [
+          {
+            assertion = cfg.enable -> !config.plugins.lsp.servers.rust-analyzer.enable;
+            message = ''
+              Nixvim (plugins.rustaceanvim): Both `plugins.rustaceanvim.enable` and `plugins.lsp.servers.rust-analyzer.enable` are true.
+              Disable one of them otherwise you will have multiple clients attached to each buffer.
+            '';
+          }
+        ];
+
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Thanks

khaneliman:
@mergifyio queue

plugins/languages/rust/rustaceanvim/default.nix
```diff
@@ -56,6 +56,16 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
         globals.rustaceanvim = cfg.settings;
 
+        assertions = [
+          {
+            assertion = cfg.enable -> !config.plugins.lsp.servers.rust-analyzer.enable;
```
MattSturgeon, line 61: I don't think this does what you intended.

First of all, `cfg.enable` is always true within `extraConfig`. This means we're effectively only checking the rhs of the expression.

If we moved it out of `extraConfig` and into an imported module, then the assertion would be:

_"Assert that `cfg.enable` is false **or** `rust-analyzer.enable` is false"_

Which does sound correct based on your messages text.

MattSturgeon (changes requested):

plugins/languages/rust/rustaceanvim/default.nix
khaneliman: I mean, the intended behavior is to not allow both enabled at the same time. If this gets enabled, we are checking the other isn’t also enabled. The lhs being redundant doesn’t change that, it was just to make it obvious what was going on

MattSturgeon: Yeah you're right.

> it was just to make it obvious what was going on

Gotcha. Ironically, I think that's what threw me off :grinning:

PR #2211, opened 2024-09-09
plugins/neotest: move to by-name

khaneliman:
Doesn't resolve issue i'm looking into, but no harm in moving it into by-name since it's a self contained plugin folder.


MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2212, opened 2024-09-09
plugins/languages: move to by-name

khaneliman:
Also just individual plugins that don't share code actually.


khaneliman:
Oops looks like I need to add stuff to the new ltex-extra test, real quick.

traxys (approved):

khaneliman:
> I was hesitant to move `languages/*` given how "generic" some of the names will be... But I suppose since the name always matches `plugin.<name>` from the option path it shouldn't be an issue?
> 
> Is there any value in "language" plugins being grouped together in a category vs being in by-name?

I didn't think so, like you said, they are unique by-name already and it doesn't help me to find several of the plugins when they are nested that way.

tests/test-sources/plugins/by-name/ltex-extra/default.nix
MattSturgeon, line 1: Just double checking, this is intentionally a new file, not a half-done rename, right?

tests/test-sources/plugins/by-name/nix/default.nix
MattSturgeon, line 1: Same

tests/test-sources/plugins/by-name/plantuml-syntax/default.nix
MattSturgeon, line 1: Same

tests/test-sources/plugins/by-name/rust-tools/default.nix
```diff
@@ -71,4 +71,13 @@
...
   };
+
+  rust-analyzer-options = {
+    plugins.rust-tools = {
+      enable = true;
+      server = {
+        assist.expressionFillDefault = "default";
+      };
+    };
+  };
```
MattSturgeon, line 82: This feels unrelated to the PR? Is it testing a regression you encountered while moving files around?

khaneliman:
@mergifyio queue

MattSturgeon (approved):
Everything looks good if you're happy with the files I flagged below :grin:

tests/test-sources/plugins/by-name/rust-tools/default.nix
khaneliman: This was just to make sure the imports were being tested since I saw a similar test in the rustaceanvim plugin.

tests/test-sources/plugins/by-name/ltex-extra/default.nix
khaneliman: There were a few plugins that didn't have tests

tests/test-sources/plugins/by-name/rust-tools/default.nix
khaneliman: Just added it since we were moving files around and wanted to double check the plugin could still access the out of directory options file.

tests/test-sources/plugins/by-name/ltex-extra/default.nix
MattSturgeon: Thought so, just wanted to double check. :grin:

PR #2213, opened 2024-09-09
docs: Render alerts using mdbook-alerts

### LINKED ISSUE #1791: [docs] support + use markdown alerts
GitHub has support for various [markdown alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts), using a similar syntax to block quotes.

> [!TIP]
> This is a tip!

MDbook [has a few CSS classes](https://rust-lang.github.io/mdBook/format/mdbook.html) to do something similar, but having to insert html tags is not ideal.

This crate looks interesting: https://crates.io/crates/mdbook-alerts

traxys:
Closes #1791

### DIFF
--- a/docs/mdbook/book.toml
+++ b/docs/mdbook/book.toml
@@ -8,3 +8,5 @@ title = "nixvim docs"
 [output.html.fold]
 enable = true
 level = 0
+
+[preprocessor.alerts]
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -252,7 +252,10 @@ pkgs.stdenv.mkDerivation {
 
   phases = [ "buildPhase" ];
 
-  buildInputs = [ pkgs.mdbook ];
+  buildInputs = [
+    pkgs.mdbook
+    pkgs.mdbook-alerts
+  ];
   inputs = lib.sourceFilesBySuffices ./. [
     ".md"
     ".toml"


MattSturgeon:
Wow this proved simpler than expected! Awesome job @traxys 

Can I ask, how will alerts render in the man-docs? Seeing as that's built using different tooling.

khaneliman:
This is actually gonna be awesome for some of the warnings and stuff throughout the docs to be more apparent.

traxys:
I'm not sure, I'll need to look at it, but it's a separate problem. I think be are going to need to write a pandoc filter or something like that

MattSturgeon:
Hm, I'm not sure this is working how I expected...

image

Perhaps we're missing the preprocessor config? See https://github.com/lambdalisue/rs-mdbook-alerts?tab=readme-ov-file#usage


> ```
> [book]
> authors = ["Alisue"]
> language = "en"
> multilingual = false
> src = "src"
> title = "mdBook Alerts preprocessor"
> 
> # ADD THIS
> [preprocessor.alerts]
> ```

docs/mdbook/book.toml
```diff
@@ -8,3 +8,5 @@ title = "nixvim docs"
 [output.html.fold]
 enable = true
 level = 0
+
+[preprocessor.alerts]
```
khaneliman, line 12: Thought that was what this was for

MattSturgeon: Yes, looks correct. But in my testing it is rendering a blockquote not an alert

MattSturgeon: I thought it might be a cache issue with the CSS, but hard-reloading with devtools open doesn't help.

Even copying the example verbatim from upstream.

The DOM looks like:
image

MattSturgeon: Ah, I think I figured it out. It is the same reason why we can't use headings in option descriptions; mdbook isn't rendering the markdown for the options, the nixpkgs tooling is.

The preprocessor is only running the first part of the page, up until the first option. This first part is effectively a "user-guide" mdbook page, while the rest of the page with options is just concatenated in by us.

Moving the example alerts into a plugin description or a user-guide file will render as:
image

MattSturgeon: Hm, actually. I'm not quite sure why option descriptions are handled differently.

@traxys did you write these bits?
https://github.com/nix-community/nixvim/blob/e864f2c9e0cda66b0f419b2c3e36374c177d3712/docs/mdbook/default.nix#L230-L248

https://github.com/nix-community/nixvim/blob/e864f2c9e0cda66b0f419b2c3e36374c177d3712/docs/mdbook/default.nix#L265-L273

At first glance, we're just concating raw markdown, and _then_ building. So mdbook shouldn't know what is an option and what is normal md... :confused:

MattSturgeon (approved):
This is correct and will allow us to use alerts throughout much of the docs.

Unfortunately there is another issue (https://github.com/nix-community/nixvim/pull/2213#discussion_r1751006609) meaning some markdown features (incl. this) are not working in the option descriptions.

I think we should keep that out of scope here, merge and then investigate that separately.

Thanks again for doing this, it will be a very useful documentation tool! 🎉

docs/mdbook/book.toml
MattSturgeon: Looking at the implementation, this bit is where we write a plugin's markdown file:
https://github.com/nix-community/nixvim/blob/e864f2c9e0cda66b0f419b2c3e36374c177d3712/docs%2Fmdbook%2Fdefault.nix#L199-L202

`pkgs.writeText "module-doc" opts.index.moduleDoc` is working as intended, allowing us to use all markdown features.

`mkMDDoc opts.index.options` is not.

MattSturgeon: `mkMDDoc` is using nixpkgs tooling to get the (commonmark) markdown from the options-json: https://github.com/nix-community/nixvim/blob/e864f2c9e0cda66b0f419b2c3e36374c177d3712/docs%2Fmdbook%2Fdefault.nix#L13-L18

So it's likely that some preprocessing is happening in the nixpkgs function.

traxys: The processing _may_ involve adding escapes to the \[ and \]

traxys:
@mergifyio queue

docs/mdbook/book.toml
traxys: (I did not write those, I modified them a bit a while ago)

Issue #1466, opened 2024-04-26
[BUG] Treesitter error when running rest plugin


mrityunjaygr8:
| Field        | Description |
|--------------|-------------|
| Plugin       | `rest`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

I have enabled the rest.nvim plugin, and when i try to send a request using `:Rest run` I am getting an error saying that `Error executing Lua callback: ...0.9.5/share/nvim/runtime/lua/vim/treesitter/language.lua:93: no parser for 'conf' language, see :help treesitter-parsers` 

Please see the attached screenshot
![image](https://github.com/nix-community/nixvim/assets/14573967/c54e699c-62e6-4b2a-ad7a-87eaf6de9ddd)

Let me know what other information is required from my side to help fix this issue

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  plugins.rest.enable = true;
  plugins.treesitter.enable = true;
  plugins.treesitter.ensureInstalled = ["lua" "xml" "http" "json" "graphql"];
}
```

traxys:
Maybe you need "conf" in ensureInstalled ?

khaneliman:
It looks like there's no `conf` treesitter parser and your neovim configuration has associated the `.http` file with that fieltype. From what I can see, you need to setup a filetype associate for the .http files to http and use the tree-sitter-http parser. 

```nix
  filetype = {
    extension = {
      "http" = "http";
    };
  };
```

PR #2214, opened 2024-09-09
plugins/rest: handle dependencies, ease configuration, and cleanup

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1466

### DIFF
--- a/plugins/by-name/rest/default.nix
+++ b/plugins/by-name/rest/default.nix
@@ -1,4 +1,5 @@
 {
+  config,
   lib,
   helpers,
   pkgs,
@@ -401,21 +402,33 @@ helpers.neovim-plugin.mkNeovimPlugin {
   };
 
   extraOptions = {
+    curlPackage = lib.mkPackageOption pkgs "curl" {
+      nullable = true;
+    };
+
     enableHttpFiletypeAssociation = lib.mkOption {
       type = types.bool;
       default = true;
       description = ''
-        Sets up the filetype association of `.http` files to trigger treesitter support.
+        Sets up the filetype association of `.http` files to trigger treesitter support to enable `rest` functionality.
       '';
     };
   };
 
   extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = config.plugins.treesitter.enable;
+        message = ''
+          Nixvim (plugins.rest): Requires the `http` parser from `plugins.treesitter`, please set `plugins.treesitter.enable`.
+        '';
+      }
+    ];
+
+    extraPackages = [ cfg.curlPackage ];
 
     filetype = lib.mkIf cfg.enableHttpFiletypeAssociation {
-      extension = {
-        "http" = "http";
-      };
+      extension.http = "http";
     };
   };
 }
--- a/plugins/by-name/rest/default.nix
+++ b/plugins/by-name/rest/default.nix
@@ -399,4 +399,23 @@ helpers.neovim-plugin.mkNeovimPlugin {
       ]
     ];
   };
+
+  extraOptions = {
+    enableHttpFiletypeAssociation = lib.mkOption {
+      type = types.bool;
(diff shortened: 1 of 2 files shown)


plugins/by-name/rest/default.nix
```diff
@@ -230,63 +233,63 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
         default: `{}`
       '';
-      apply = v: if isAttrs v then mapAttrs (_: helpers.mkRaw) v else v;
+      apply = v: if lib.isAttrs v then lib.mapAttrs (_: lib.nixvim.mkRaw) v else v;
```
MattSturgeon, line 236: We really must get around to moving this to the type itself. Having to do this in every option that uses strLua is too awkward.

`coercedTo str mkRaw rawLua // { description = ""; }` is all we need :sob:

plugins/by-name/rest/default.nix
```diff
@@ -399,4 +402,31 @@ helpers.neovim-plugin.mkNeovimPlugin {
       ]
     ];
   };
+
+  extraOptions = {
+    curlPackage = lib.mkPackageOption pkgs "curl" {
+      nullable = true;
+    };
+
+    enableHttpFiletypeAssociation = lib.mkEnableOption "filetype association of http files";
```
MattSturgeon, line 411: Is the plugin useful without this filetype? If yes, this may be correct. If 90% of users will want it, maybe we make an exception to our usual "everything must by explicit" policy and set a sane-default here?

khaneliman: Oh yeah, this defaults to false doesn't it...

plugins/by-name/rest/default.nix
```diff
@@ -399,4 +402,31 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+  };
+
+  extraConfig = cfg: {
+    assertions = [
+      {
+        assertion = cfg.enable -> config.plugins.treesitter.enable;
```
MattSturgeon, line 417:
```suggestion
        assertion = config.plugins.treesitter.enable;
```

As we discussed earlier. I'd rather not have `cfg.enable` checks in `extraConfig` as it'll confuse people into thinking the config is always defined, even when not enabled.

plugins/by-name/rest/default.nix
```diff
@@ -399,4 +402,31 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+        '';
+      }
+    ];
+
+    extraPackages = [ cfg.curlPackage ];
+
+    filetype = lib.mkIf cfg.enableHttpFiletypeAssociation {
+      extension = {
+        "http" = "http";
+      };
```
MattSturgeon, line 429:
```suggestion
      extension.http = "http";
```

MattSturgeon: Or `filetype.extension = mkIf () { http = "http"; }`

plugins/by-name/rest/default.nix
MattSturgeon: We can either `// {default = true;}` or use `mkOption`

MattSturgeon (approved):

plugins/by-name/rest/default.nix
khaneliman: Just changing the definition from 
```nix
  strLikeType =
    description:
    lib.mkOptionType {
      name = "str";
      inherit description;
      descriptionClass = "noun";
      check = v: lib.isString v || isRawType v;
    };
```

khaneliman:
@mergifyio queue

plugins/by-name/rest/default.nix
MattSturgeon: Not in this PR of course, but this will be the first step before #2003

Issue #2217, opened 2024-09-10
docs: use consistent admonitions syntax in options and user-guides


MattSturgeon:
- [x] #2213
- [x] #2218
- [ ] #2221

#2213 added support for rendering github-flavoured "markdown alerts":

```md
> [!TIP]
> This is a tip
```

Using the mdbook-alerts preprocessor, we now renders this as:

```html
<div class="mdbook-alerts mdbook-alerts-tip">
  <p class="mdbook-alerts-title">
    <span class="mdbook-alerts-icon"></span>
    tip
  </p>
  <p>This is a tip</p>
</div>
```

> [!TIP]
> This is a tip

---

However, we render our option descriptions using [`nixos-render-docs`](https://github.com/NixOS/nixpkgs/blob/1376d89f1394fd5111a7306a3e72c8f9359e36d8/pkgs/tools/nix/nixos-render-docs/default.nix#L21), via the [`optionsCommonMark` function](https://github.com/NixOS/nixpkgs/blob/1376d89f1394fd5111a7306a3e72c8f9359e36d8/nixos/lib/make-options-doc/default.nix#L182-L190).
https://github.com/nix-community/nixvim/blob/e864f2c9e0cda66b0f419b2c3e36374c177d3712/docs%2Fmdbook%2Fdefault.nix#L13-L18

This takes an `options.json` file parses its markdown to an AST and spits out commonmark, asciidoc, or manpages (as requested).

Because it uses an abstract syntax tree, recognized-but-invalid symbols (such as `[foo]` where there's no corresponding `[foo]: URL`) end up being escaped (`\[foo\]`) in the resulting commonmark output.

It's also probably not a good idea to use syntax not supported by nixos-render-docs in option descriptions, since it will render poorly in the manpage docs.

---

As documented [here](https://github.com/NixOS/nixpkgs/blob/master/doc/README.md#admonitions), nixpkgs have adopted an alternative syntax for admonitions:

```md
::: {.warning}
This is a warning
:::
```

This is based on pandoc’s [fenced `div`s syntax](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/fenced_divs.md) which is _intended_ to be used to add HTML classes or IDs to the rendered `div`.

`nixos-render-docs options commonmark` will render the above as `**Warning:** This is a warning`, which mdbook then renders as:

```html
<p><strong>Warning:</strong> This is a warning</p>
```
**Warning:** This is a warning

[...]

PR #2218, opened 2024-09-10
docs: patch nixos-render-docs to output GFM alerts

MattSturgeon:
Follow up to https://github.com/nix-community/nixvim/pull/2213

Initial work towards #2217

This patches `nixos-render-docs` to use GFM-style alerts when rendering admonitions to markdown.

By default, it renders them as `**<kind>:** <text>`.

This **does not** add support for _parsing_ GFM-style alerts as admonitions in option descriptions, so the nixpkgs-style syntax should be used there for now.

For example, the following renders as:

```md
Install the man pages for NixVim options.

::: {.warning}
This is a warning!
:::
```

image


traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2219, opened 2024-09-10
colorschemes/modus: add new configuration option

nwjsmith:
The `hide_inactive_statusline` option is available in newer versions of the theme.

### DIFF
--- a/plugins/colorschemes/modus.nix
+++ b/plugins/colorschemes/modus.nix
@@ -46,6 +46,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
       Disable setting the background color.
     '';
 
+    hide_inactive_statusline = defaultNullOpts.mkBool false ''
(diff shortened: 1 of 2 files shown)


MattSturgeon:
`settings` is a freeform option, so you can already configure this sub-option without this PR.

If we merge this, it will be for documentation & type-checking purposes only.

MattSturgeon (approved):
I don't see any issues. The option matches upstream's docs:

> `hide_inactive_statusline = false, -- Hide statuslines on inactive windows. Works with the standard **StatusLine**`

While it's not required to have options declared for all upstream plugin options due to the freeform nature of `settings` I'm happy to merge if others agree.

cc @nix-community/nixvim

khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2221, opened 2024-09-10
docs: Support GFM style admonitions in option descriptions

### LINKED ISSUE #2217: docs: use consistent admonitions syntax in options and user-guides

traxys:
Follow up to https://github.com/nix-community/nixvim/pull/2213 & https://github.com/nix-community/nixvim/pull/2218. Fixes #2217.

This is based on https://github.com/antfu/markdown-it-github-alerts/blob/23d2d84c738e206b74762e388ca35c076d170028/src/index.ts#L61-L88.


docs/default.nix
```diff
@@ -40,8 +40,8 @@ let
         patches = old.patches or [ ] ++ [
           # Adds support for GFM-style admonitions in rendered commonmark
           ./0001-Output-GFM-admonition.patch
-          # TODO:add support for _parsing_ GFM admonitions too
-          # https://github.com/nix-community/nixvim/issues/2217
+          # Adds support for parsing GFM-style admonitions
+          ./0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```
MattSturgeon, line 44: Nit: ideally the patch should be generated on top of the other patch, and then this would be `0002`

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
+--- a/nixos_render_docs/md.py
++++ b/nixos_render_docs/md.py
+@@ -341,6 +341,40 @@ def _parse_blockattrs(info: str) -> Optional[tuple[AttrBlockKind, Optional[str],
+         return ('figure', id, classes)
+     return None
+ 
++_ALERT_PATTERN = re.compile(r"^\[\!(TIP|NOTE|IMPORTANT|WARNING|CAUTION)\]", re.IGNORECASE)
++
++def _gfm_alert(md: markdown_it.MarkdownIt):
++    def gfm_alert_to_adm(state: markdown_it.rules_core.StateCore):
```
MattSturgeon, line 26: This function is large enough that it makes me uncomfortable having it directly in a patch file.

I think it'll end up being a maintenance nightmare if we ever want to work on it.

I'm tempted to suggest extracting it into either a loose file or a nix string variable.

If we did that, we could add a `@PLACEHOLDER@` to the patch, then substitute in the actual function in `postPatch`.

MattSturgeon (commented):
I'm not familiar enough with `nixos-render-docs` to spot any footguns, but if the patched tests are passing then we should be good. Thanks for tackling this so quickly, I wasn't looking forward to it!

Have you tested out these changes?

What do you think of the suggestions below? I think `_gfm_alert` defined in a loose file that is injected using `substituteInPlace` may be more maintainable; I always struggle working directly with patch files 😅 

It'd be nice to have some comments in the implementation too, and I think a loose file would make adding/editing/reading them easier.

What do you think?

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon: Or maybe I just need to re-read the patch on a larger screen and with fresh eyes tomorrow.

Just using a patch is certainly simpler than having a patch, a loose file, _and_ a substitution...

MattSturgeon:
> This is based on [antfu/markdown-it-github-alerts@`23d2d84`/src/index.ts#L61-L88](https://github.com/antfu/markdown-it-github-alerts/blob/23d2d84c738e206b74762e388ca35c076d170028/src/index.ts#L61-L88)

I'm also looking at https://github.com/ByPikod/markdown-it-github-alert/blob/f17613451396d7aa43cf6e8a3a81a4663ff1f27b/src/plugin.ts#L72-L122 to see compare the two

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: (in term of simplicity)

traxys:
I tested on a plugin that it worked, though I only tested with a  `TIP`

MattSturgeon:
> > This is based on [antfu/markdown-it-github-alerts@`23d2d84`/src/index.ts#L61-L88](https://github.com/antfu/markdown-it-github-alerts/blob/23d2d84c738e206b74762e388ca35c076d170028/src/index.ts#L61-L88)
> 
> I'm also looking at [ByPikod/markdown-it-github-alert@`f176134`/src/plugin.ts#L72-L122](https://github.com/ByPikod/markdown-it-github-alert/blob/f17613451396d7aa43cf6e8a3a81a4663ff1f27b/src/plugin.ts#L72-L122) to see compare the two

The other implementation leverages the "actual" blockquote rule, extracted from MarkdownIt. I wonder if that makes it more robust? [L65-L69](https://github.com/ByPikod/markdown-it-github-alert/blob/f17613451396d7aa43cf6e8a3a81a4663ff1f27b/src/plugin.ts#L65-L69)

I also like having `matchAlertType` in a separate function (returns `null` or "type"), see: [parser.ts](https://github.com/ByPikod/markdown-it-github-alert/blob/f17613451396d7aa43cf6e8a3a81a4663ff1f27b/src/parser.ts#L13). I think this makes things more readable in the end and allows for slightly more robust parsing (see some of the comments in that file).

Using named regex groups may also help to that end. Python uses `(?P<name>.*)` syntax. You then use either `m.groupdict()` or `m.group('name')` instead of `m.group(0)`.

traxys:
I agree that the implementation you linked seemed a bit more robust, but we should be _ok_ as nested GFM alerts are not permitted.

I skimmed through the nixos-render-docs implementation and they _seem_ to use the same level of rules for their handling of `::: {.warning}`.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++
++                first_content = next(
++                    (t for t in tokens[start_index:end_index + 1] if t.type == "inline"),
++                    None,
++                )
++                if first_content is None:
++                    continue
++
++                if (m := _ALERT_PATTERN.match(first_content.content)) is None:
++                    continue
```
MattSturgeon, line 45: I like the idea of splitting this into another function, but it isn't that important.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++
++                first_content.content = first_content.content[len(m[0]):].lstrip()
```
MattSturgeon, line 47:
```suggestion
+                # Remove "> [!TIP]\n" from the token
+                first_content.content = first_content.content[m.endpos:].lstrip()
```

- Let's use `Match.endpos` to get the substring index
- Let's add a comment so this makes sense in 12 months time :upside_down_face: 
- Do we really need `lstrip()`? I think it would be better to match whitespace in the regex.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++    def gfm_alert_to_adm(state: markdown_it.rules_core.StateCore):
++        tokens = state.tokens
++        for (i, token) in enumerate(tokens):
++            if token.type == "blockquote_open":
++                open = tokens[i]
++                start_index = i
++                while tokens[i].type != "blockquote_close" and i < len(tokens):
++                    i += 1
++                close = tokens[i]
++                end_index = i
```
MattSturgeon, line 35: Lets add a comment, maybe `Look for the start/end of the blockquote`? Maybe also clarify it relates to _tokens_ not chars? I.e. we are dealing with a list of `Token` objects.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++
++                first_content = next(
++                    (t for t in tokens[start_index:end_index + 1] if t.type == "inline"),
++                    None,
++                )
++                if first_content is None:
++                    continue
```
MattSturgeon, line 42: I don't fully understand this, so I think it needs some comments.

It looks for the first "inline" type token in the pre-tokenized blockquote? Using the index-bounds we searched for above.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++
++                if (m := _ALERT_PATTERN.match(first_content.content)) is None:
++                    continue
++
++                first_content.content = first_content.content[len(m[0]):].lstrip()
++
++                open.type = "admonition_open"
++                open.meta['kind'] = m[1].lower()
```
MattSturgeon, line 50: Nit: let's be explicit about this being a regex group:
```suggestion
+                open.meta['kind'] = m.group(1).lower()
```

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++
++                close.type = "admonition_close"
```
MattSturgeon, line 52: I don't see us declaring or reading `open` and `close` anywhere, so I assume they are coming from MarkdownIt. Are we inheriting these from a super class?

(forgive my lack of python-foo)

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,145 @@
...
++        pass
++
++    md.core.ruler.after('block', 'github-alerts', gfm_alert_to_adm)
```
MattSturgeon, line 55: I wouldn't be against inlining `gfm_alert_to_adm` to an anymous lambda.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon: Let's also add a comment here, stating what we're doing here at the end of the function.

I guess something like `set the open & close class properties in the same way as nixpkgs admonitions`

MattSturgeon (approved):
I have some suggestions for making this more readable (below) but overall I'm happy, so here's my green tick :grinning: 

Awesome work again @traxys :tada:

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: Yeah exactly, it mimics a JS `slice(....).find(....)`

MattSturgeon: I guess it makes me wonder what non-inline token types it may run into. And whether any of them should trigger errors or an early return.

traxys: I copied this behaviour from the project I linked, I don't know a lot about markdown-it nor python

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: What part would you want in another function?

docs/0002-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,112 @@
+From 50c0e6b6075b8011e7643bf4fcce710030a72dd8 Mon Sep 17 00:00:00 2001
+From: Quentin Boyer <quentin+dev@familleboyer.net>
+Date: Wed, 11 Sep 2024 19:21:12 +0200
+Subject: [PATCH 2/2] nixos-render-docs: Support gfm-style admonitions
```
MattSturgeon, line 4: I know this is unimportant, but:

Since this is now 2/2 can you update my patch too to be part of the patchset?

If you do that, maybe reword one of the commits so either none or both use the `nixos-render-docs: ` prefix?

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: `open = <.....>` declares the variable at the start of the loop

docs/gfm-alerts-to-admonitions/default.nix
```diff
@@ -0,0 +1,30 @@
...
+  lib,
+  setuptools,
+  pytest-regressions,
+}:
+buildPythonPackage {
+  pname = "gfm-alerts-to-admonitions";
+  version = "0.0";
+  format = "pyproject";
+
+  src = ./.;
```
MattSturgeon, line 14: This probably doesn't really matter in nixvim, but this is frowned upon because changes to _anything_ in `./.` will trigger a rebuild, even stuff in _this file_ that'd normally just be metadata or passthru.

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: I don't like those very much :/ I find the current approach readable enough.
I also took other plugins as examples, they were all structured like this (in nixos-render-docs or in markdown it plugins)

docs/default.nix
```diff
@@ -37,11 +37,19 @@ let
       inherit lib;
 
       nixos-render-docs = prev.nixos-render-docs.overrideAttrs (old: {
+        propagatedBuildInputs = old.propagatedBuildInputs ++ [
```
MattSturgeon, line 40: I know this is kinda off-topic, but I can never wrap my head around when to use native or not and propagated or not for buildInputs...

My understanding is:
- buildInputs generally only relate to the drv builder, not the package's runtime
- `native` buildInputs are for exes that should be run from the PATH during the build
- `propagated` buildInputs are for deps that should also be available to builders of _other packages_ that include this package in _their_ buildInputs (i.e. transitives get propagated)
- propagatedNativeBuildInputs is also a thing, see above two

Here we shouldn't need propagated right?

docs/default.nix
```diff
@@ -37,11 +37,19 @@ let
...
+          (final.python3.pkgs.callPackage ./gfm-alerts-to-admonitions {
+            # We need to do this because nixos-render-docs does it internally, and it results in two
+            # different markdown-it-py derivations
+            markdown-it-py = final.python3.pkgs.markdown-it-py.overridePythonAttrs { doCheck = false; };
```
MattSturgeon, line 44: I don't fully understand why we need to override `doCheck` here? Especially since we aren't overriding/overlaying `markdown-it-py` anywhere :face_with_diagonal_mouth:

docs/gfm-alerts-to-admonitions/gfm_alerts_to_admonitions/main.py
MattSturgeon, line 1: Much more maintainable :heart_eyes:

docs/gfm-alerts-to-admonitions/tests/test_plugin.py
MattSturgeon, line 1: We even have tests :astonished:

docs/gfm-alerts-to-admonitions/tests/test_plugin.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from gfm_alerts_to_admonitions import gfm_alert_to_admonition
+from markdown_it import MarkdownIt
+
+
+@pytest.mark.parametrize("kind", ["tip", "note", "important", "warning", "caution"])
+def test_parse(data_regression, kind):
+    input = f"> [!{kind.upper()}]\n> This is an *alert*"
```
MattSturgeon, line 8: `upper` shouldn't be needed, GFM-alerts are case-insensitive.

Maybe we should test mixed cases too?

flake-modules/dev/default.nix
```diff
@@ -52,6 +52,7 @@
             "**.scm"
             "**.svg"
             "**/man/*.5"
+            "/docs/gfm-alerts-to-admonitions/tests/**/*.yml"
```
MattSturgeon, line 55: Any reason why? (we have a yaml formatter configured IIRC)

MattSturgeon (commented):
Awesome :tada:

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: `endpos` is not the same as `len(m[0])`:

> The value of endpos which was passed to the [search()](https://docs.python.org/3/library/re.html#re.Pattern.search) or > [match()](https://docs.python.org/3/library/re.html#re.Pattern.match) method of a [regex object](https://docs.python.org/3/library/re.html#re-objects). This is the index into the string beyond which the RE engine will not go.

In pratice here it is the length of the string

docs/0002-nixos-render-docs-Support-gfm-style-admonitions.patch
traxys: I don't know how to apply your patch in git, I use `patch -p1`, so it would reset all the metadata :/

docs/gfm-alerts-to-admonitions/default.nix
traxys: Yeah, I saw how nixos-render-docs did it, but I did not want to bother with the complexity of filtering. I'm betting on us not touching this part very often

docs/default.nix
```diff
@@ -97,6 +105,9 @@ let
 in
 {
   inherit options-json;
+  inherit (pkgsDoc) nixos-render-docs;
+
+  gfm-alerts-to-admonitions = pkgsDoc.python3.pkgs.callPackage ./gfm-alerts-to-admonitions { };
```
MattSturgeon, line 110: I wonder if `gfm-admonitions-to-nixpkgs` is a better name?

Admonitions is a generic term, but the tokens we're targeting are nixpkgs-sepecific.

docs/default.nix
traxys: We don't, but nixos-render-docs [does](https://github.com/NixOS/nixpkgs/blob/b94f2597143d383aeed61a8f036b596d34a26dfc/pkgs/tools/nix/nixos-render-docs/default.nix#L7-L17).

I fiddled with a number of ways to pass our dependency, this one was the only one which did not raise a double-dependency error

docs/gfm-alerts-to-admonitions/tests/test_plugin.py
traxys: The regex _should_ match different cases, we could indeed add a test

flake-modules/dev/default.nix
traxys: `pytest-regression` performs a `diff`, so the formatter messes with it

MattSturgeon: Makes sense. Could you note that in a comment?

docs/default.nix
traxys: I feel like `nixpkgs` is too vague, it would have to be `nixos-render-docs` to be descriptive enough, but it's a bit long.

The plugin itself does not really rely on nixos-render-docs though, it creates `admonition_open` and `admoninition_close` tokens.
How to parse them is orthogonal to this plugin, so the name still makes sense for me

docs/gfm-alerts-to-admonitions/tests/test_plugin.py
MattSturgeon: The implementation looks good, we probably don't need a test but it can't hurt.

docs/gfm-alerts-to-admonitions/gfm_alerts_to_admonitions/main.py
```diff
@@ -0,0 +1,51 @@
...
+                    None,
+                )
+                if first_content is None:
+                    continue
+
+                if (m := _ALERT_PATTERN.match(first_content.content)) is None:
+                    continue
+
+                # Remove ' [!TIP]' from the token
+                first_content.content = first_content.content[len(m[0]) :]
```
MattSturgeon, line 43: Looks like you tried `m.endpos` but reverted. Did it not do what I thought it would?

traxys: Yeah, I responded in the original suggestion with what it really does

traxys: (I was glad to have the tests, as they caught the issue immediately)

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon: Ah, I guess we want [`m.end(0)`](https://docs.python.org/3/library/re.html#re.Match.end) then. Sorry, just doing my best to read the docs and usually reading the wrong part :sweat_smile:

docs/0001-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon: I was originally thinking we could be inspired by [parser.ts](https://github.com/ByPikod/markdown-it-github-alert/blob/f17613451396d7aa43cf6e8a3a81a4663ff1f27b/src/parser.ts#L13), which exports a "matcher" function returning either `null` or the matched type.

However, I think it is fine to keep this inline as it makes removing the matched substring easier.

docs/0002-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon:
```
git am --directory pkgs/**/nixos-render-docs/src ~/nixvim/main/docs/0001-Output-GFM-admonition.patch
```

docs/default.nix
MattSturgeon: Is this because both our package and nixos-render-docs have a propagated dep on markdown-it-py? (sorry I'm still a bit confused)

traxys: Yeah exactly, and they end up being _not_ the same package due to their override

MattSturgeon: Maybe an alternative wording for the comments could be:

> Use the same override as `nixos-render-docs` does, to avoid "duplicate dependency" errors

docs/gfm-alerts-to-admonitions/default.nix
MattSturgeon: Normally `./src` would be best practice, but as I said it shouldn't be too important for an internal package only used when we build our docs.

docs/gfm-alerts-to-admonitions/gfm_alerts_to_admonitions/__init__.py
```diff
@@ -0,0 +1 @@
+from .main import gfm_alert_to_admonition
```
MattSturgeon, line 1: We're getting a linting error: https://buildbot.nix-community.org/#/builders/448/builds/1104/steps/1/logs/stdio

traxys: I have no idea what that means xD

traxys: I'll look it up

traxys: See https://docs.astral.sh/ruff/rules/unused-import/ for more details

flake-modules/dev/default.nix
```diff
@@ -52,6 +52,9 @@
             "**.scm"
             "**.svg"
             "**/man/*.5"
+            # Those files are generated by pytest-regression, which then `diff`s them.
+            # Formatting them will make the tests fail.
+            "/docs/gfm-alerts-to-admonitions/tests/**/*.yml"
```
MattSturgeon, line 57: Seems to not be working. Maybe the leading `/` is incorrect?

https://buildbot.nix-community.org/#/builders/720/builds/1095/steps/1/logs/stdio

docs/gfm-alerts-to-admonitions/gfm_alerts_to_admonitions/main.py
```diff
@@ -0,0 +1,51 @@
...
+
+import markdown_it
+
+_ALERT_PATTERN = re.compile(
+    r"^\[\!(TIP|NOTE|IMPORTANT|WARNING|CAUTION)\]\s*", re.IGNORECASE
+)
+
+
+def gfm_alert_to_admonition(md: markdown_it.MarkdownIt):
+    def gfm_alert_to_adm(state: markdown_it.rules_core.StateCore):
```
MattSturgeon, line 11: I think we can make this a little less verbose with some imports;

```py
from markdown_it import MarkdownIt
from markdown_it.rules_core import StateCore
```

Maybe we could also add type annotations for the `[Token]` list below? [I think](https://markdown-it-py.readthedocs.io/en/latest/api/markdown_it.token.html) that'd be `from markdown_it.token import Token`

docs/gfm-alerts-to-admonitions/default.nix
MattSturgeon: To clarify, I wasn't talking about the `filter` function used with `lib.cleanSourceWith`.

I was referring to https://github.com/NixOS/nixpkgs/issues/301014:

> Derivations that use `src = ./.` will be rebuilt on any change to the `.nix` file that includes them, including formatting changes caused by `nixfmt-rfc-style` for [RFC 166](https://github.com/NixOS/rfcs/blob/master/rfcs/0166-nix-formatting.md?rgh-link-date=2024-04-02T18%3A12%3A28Z).

docs/default.nix
khaneliman: > I know this is kinda off-topic, but I can never wrap my head around when to use native or not and propagated or not for buildInputs...
> 
> My understanding is:
> 
>     * buildInputs generally only relate to the drv builder, not the package's runtime
> 
>     * `native` buildInputs are for exes that should be run from the PATH during the build
> 
>     * `propagated` buildInputs are for deps that should also be available to builders of _other packages_ that include this package in _their_ buildInputs (i.e. transitives get propagated)
> 
>     * propagatedNativeBuildInputs is also a thing, see above two
> 
> 
> Here we shouldn't need propagated right?

buildInputs are runtime available packages, actually.
native are build only
propogated do the same but to derivations that inherit it.

docs/0001-nixos-render-docs-Output-GFM-admonition.patch
```diff
@@ -44,5 +44,5 @@ index 4ff0bc3095c3..e151325338ca 100644
  [
  multiline
 -- 
-2.46.0
+2.45.2
```
khaneliman, line 47: Not sure I understand what this version change does

docs/0002-nixos-render-docs-Support-gfm-style-admonitions.patch
```diff
@@ -0,0 +1,112 @@
...
++.br
++This is a \\fInote\\fR
++.RE
++.sp
++.RS 4
+ \\fBWarning\\fP
+ .br
+ foo
+-- 
+2.45.2
```
khaneliman, line 111: is the version change just to match this? or could we just change it at this location?

docs/0001-nixos-render-docs-Output-GFM-admonition.patch
traxys: @MattSturgeon and I woul have had different git versions, as I re-generated the patch so it would be 1/2

docs/0002-nixos-render-docs-Support-gfm-style-admonitions.patch
MattSturgeon: As said in https://github.com/nix-community/nixvim/pull/2221#discussion_r1758341074 it is just the git version used to run `git format-patch`.

I asked traxys to amend my prior patch so that a) it'd be `[1/2]` and b) it'd be worded consistently in the commit message.

I can push up some fixup commits with patch files generated by my git version to avoid the noise in the diff, if you like.

MattSturgeon (approved):
Apologies; I had to force-push as b61fae44991ea6c322985b6fdf45fa052485b15d would've been a no-op if committed at the HEAD of the PR.
EDIT: If you don't plan to merge immediately, I'll also take the liberty of rebasing after I merge #2229, since that'll conflict.

Once rebased & autosquashed this all looks good to me.

TBH, I wasn't convinced we'd ever get around to doing this and was making piece with using github-syntax in user-guides and nixpkgs-syntax in options... Thank you so much for not only tackling it, but also re-working it as a separately maintainable package derivation!

I'd love to have @GaetanLepage's eyes on this too, as he is far more experienced in python than me. But I don't want to put any pressure on him due to his other commitments, so it's up to you how long you will wait for additional feedback before merging.

khaneliman (approved):
LGTM, would be nice to have some docs about writing documentation in our contributing... but I understand that's always hard to have the energy/motivation to do, especially from scratch.

docs/0001-nixos-render-docs-Output-GFM-admonition.patch
khaneliman: Ah, alright, I rarely look that closely at git patches. That makes sense.

MattSturgeon:
> LGTM, would be nice to have some docs about writing documentation in our contributing...

I don't think we'd need much more than this on the topic:

```md
> [!TIP]
> You can use GitHub-style alerts within option descriptions if you need to emphasise something. e.g.
>
> ```md
> > [!WARNING]
> > This is a warning!
> ```
```

> [!TIP]
> You can use GitHub-style alerts within option descriptions if you need to emphasise something. e.g.
>
> ```md
> > [!WARNING]
> > This is a warning!
> ```

...But we'd presumably want it to be part of a wider section on style and syntax? So probably best in a follow up PR.

khaneliman:
> > Would be nice to have some docs about writing documentation in our contributing...
> 
> I don't think we'd need much more than this on the topic:
> 
> ```
> > [!TIP]
> > You can use GitHub-style alerts within option descriptions if you need to emphasise something. e.g.
> >
> > ```md
> > > [!WARNING]
> > > This is a warning!
> > ```
> ```
> 
> Tip
> 
> You can use GitHub-style alerts within option descriptions if you need to emphasise something. e.g.
> 
> ```
> > [!WARNING]
> > This is a warning!
> ```
> 
> ...But we'd presumably want it to be part of a wider section on style and syntax? So probably best in a follow up PR.

Yeah, I wasn't very clear... I meant it'd be nice to have a quick bit about this functionality... but it might feel weird by itself without corresponding documentation on the documentation strategy, in general.

traxys:
@mergifyio queue

PR #2222, opened 2024-09-11
plugins/rest: add enableTelescope and revert by-name

khaneliman:
No idea why, yet. But, moving the plugin to by-name breaks treesitter and neotest plugins ( that we know of ). Just moving it back to manual import, for now. While I was in there, I figured I'd add this telescope integration. 

Resolves similar issue I had while testing https://github.com/nix-community/nixvim/discussions/2220#discussion-7166232 and an unreported issue I've had with neotest. 

Also noticed our nixpkgs version of the vimPlugins.rest-nvim is a little out of date. Created https://github.com/NixOS/nixpkgs/pull/341146 to try and catch it up.


MattSturgeon (approved):
LGTM

khaneliman:
@mergifyio queue

PR #2228, opened 2024-09-12
nixvimInfo: access `package.default` using `tryEval`

MattSturgeon:
Allow getting `meta.homepage` from the default package to fail gracefully when the nixpkgs channel lock is out-of-sync with ours.

This is a draft because we still need to:
- [ ] see if this fixes (or at least helps with) end-user mismatched/out-of-sync nixpkgs lock errors
- [ ] add a test to ensure plugin maintainers set a valid URL (or use a package that has `meta.homepage`)

### DIFF
--- a/lib/neovim-plugin.nix
+++ b/lib/neovim-plugin.nix
@@ -49,6 +49,7 @@
     }@args:
     let
       namespace = if isColorscheme then "colorschemes" else "plugins";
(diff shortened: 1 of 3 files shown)


MattSturgeon:
Anyone have a reproducible example that would cause a "not found in pkgs" error without this PR?

PR #2229, opened 2024-09-12
docs: move pkgs overlays to their own file

MattSturgeon:
Will conflict with #2221 and #2230, so let's hold off on merging.

### DIFF
--- a/docs/pkgs.nix
+++ b/docs/pkgs.nix
@@ -1,0 +1,44 @@
+{
+  system,
+  nixpkgs,
+}:
+let
+  # FIXME:
+  # Building the docs evaluates many package-option defaults, some of which are unfree.
+  # This usually happens when we include the package option value in another option's default without
+  # using a literalExpression defaultText.
+  config = {
+    allowUnfree = true;
+  };
+
+  # Extend nixpkg's lib, so that we can handle recursive leaf types such as `either`
+  libOverlay = final: prev: {
+    types = prev.types // {
+      either =
+        t1: t2:
+        prev.types.either t1 t2
+        // {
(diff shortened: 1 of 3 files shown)


docs/pkgs.nix
```diff
@@ -0,0 +1,56 @@
+{
+  system,
+  nixpkgs,
+}:
+let
+  # Building the docs evaluates each plugin's default package, some of which are unfree
+  # FIXME: is this still true? Most packages should only be evaluated for their meta.homepage URL now...
+  config = {
+    allowUnfree = true;
+  };
```
MattSturgeon, line 10: I guess it _is_ still needed:

```
… while evaluating the default value of option `plugins.codeium-vim.settings.bin`

error: Package ‘codeium-1.10.0’ in /nix/store/XXX-source/pkgs/by-name/co/codeium/package.nix:51 has an unfree license (‘unfree’), refusing to evaluate.
```

We sometimes use the package-option value in the default for another option; in this case we're probably doing `lib.getExe cfg.package`.

khaneliman (approved):
LGTM, makes it more clear what code is docs specific and what is just overlaying tweaks to support our docs.

docs/pkgs.nix
MattSturgeon: We can probably remove the "allow unfree" if we do a treewide adding `defaultText` for any options whose default evaluates a package-option's value.

But that's beyond the scope of this PR.

It would allow us to go back to using `pkgs.extend` instead of importing a fresh nixpkgs instance here, though.

MattSturgeon:
@mergifyio queue

PR #2230, opened 2024-09-12
lib/types: simplify `eitherRecursive` by defining it only once

MattSturgeon:
- **docs: remove unused `oneOfRecursive`**

This function is not used anywhere in nixvim, and was never even included in our `lib/types.nix`.

- **lib/types: simplify `eitherRecursive` by defining it only once**

Overriding the implementation in the docs is redundant. We can do the same thing directly in `lib/types.nix`.


traxys (approved):

khaneliman (approved):

MattSturgeon:
@mergifyio queue

Issue #2210, opened 2024-09-09
Re-imagined standalone mode (`nixvimConfig`)


MattSturgeon:
Currently, standalone mode is a function that takes a nixvim module and optionally a nixpkgs instance, then returns a wrapped neovim package.

I won't go over all the pitfalls of this approach here, but some are covered in https://github.com/nix-community/nixvim/discussions/2186.

In contrast, both nixos and home-manager have the concept of a "configuration". A configuration is simply the result of calling `lib.evalModules` with nixvim's modules and a user's nixvim modules; specifically an attrset with the form:
```nix
{
  _type = "configuration";
  options; # options, as passed to modules
  config; # config, as passed to modules
  _module; # internal module config, e.g. _module.args
  extendModules; # a function to re-evaluate with additional modules
  type; # a submodule option-type, can be used instead of `types.submodule`
  class; # usually null, a nominal module type. When non-null, it is used to filter imported modules
}
```

Incidentally, `extendModules` is a function similar to `evalModules` itself, taking `modules`, `sepcialArgs`, and `prefix` as optional args. It returns a new configuration, similar conceptually to our current `<nixvim>.extend` function.

Tools like `nixos-rebuild` and `home-manager` simply build [`<configuration>.config.system.build.toplevel`](https://search.nixos.org/options?channel=unstable&show=system.build.toplevel&from=0&size=50&sort=relevance&type=packages&query=system.build); similar to our `finalPackage` this is just a `package`-type module option.

I propose that (at some point after 24.11) we begin promoting the use of our `evalNixvim` wrapper function as an alternative to the current standalone wrapper (`makeNixvim`).

Users would consume this by having a `nixvimConfigurations` flake output, and/or a `nixvimModules` flake output for un-evaluated modules.

They can then re-use `nixvimConfigurations.<system>.<name>.config.finalPackage` a in their flake `packages` output and/or in their other configuration modules (e.g. nixos `environment.systemPackages`).

One thing that feels a little clunky here is `nixvimConfigurations.<system>`; historically a "configuration" would be system-specific. Although in theory there's nothing stopping you from having perSystem nixos/home-manager configs somewhere in your flake outputs...

We may also want to adopt a similar option structure to nixos; putting our build outputs under the `config.build.*` namespace instead of at the top `config.*` level (done in #2306).

[...]

PR #2231, opened 2024-09-12
lib: segregate and deprecate functions that need `pkgs`

MattSturgeon:
Splits everything that depends on a `pkgs` instance into an `optionalAttrs`, allowing `helpers.nix` to be bootstrapped without `pkgs`. The separated functions may be deprecated later, depending on how other refactoring pans out.

This required some refactoring:
- `modules.specialArgs` is only available when `pkgs` is used
- `modules.specialArgsWith` now requires `defaultPkgs` be provided
- `builders.*` now have `*With` variants that take `pkgs` as an argument

This is the first small step towards https://github.com/nix-community/nixvim/discussions/2186, #2210, and related refactoring.

### DIFF
--- a/lib/builders.nix
+++ b/lib/builders.nix
@@ -1,21 +1,39 @@
-{ lib, pkgs }:
-rec {
+{ lib }:
+# NOTE: use local recursion instead of accessing `lib.nixvim.builders`.
+# The latter isn't a fixed shape since it may get the deprecated functions meregd in,
+# which would lead to infinite recursion.
+lib.fix (builders: {
+  # Curry a nixpkgs instance into the *With functions below, dropping the `With` suffix
+  withPkgs =
+    pkgs:
+    lib.concatMapAttrs (
+      name: fn:
+      let
+        match' = builtins.match "(.*)With" name;
+        name' = builtins.head match';
+      in
+      lib.optionalAttrs (match' != null) {
+        ${name'} = fn pkgs;
+      }
+    ) builders;
+
   /*
     Write a lua file to the nix store, formatted using stylua.
 
     # Type
 
     ```
-    writeLua :: String -> String -> Derivation
+    writeLuaWith :: pkgs -> String -> String -> Derivation
     ```
 
     # Arguments
 
+    - [pkgs] A nixpkgs instance
     - [name] The name of the derivation
     - [text] The content of the lua file
   */
-  writeLua =
-    name: text:
+  writeLuaWith =
+    pkgs: name: text:
     pkgs.runCommand name { inherit text; } ''
       echo -n "$text" > "$out"
 
@@ -33,16 +51,17 @@ rec {
     # Type
 
     ```
-    writeByteCompiledLua :: String -> String -> Derivation
+    writeByteCompiledLuaWith :: pkgs -> String -> String -> Derivation
     ```
 
     # Arguments
 
+    - [pkgs] A nixpkgs instance
     - [name] The name of the derivation
     - [text] The content of the lua file
   */
-  writeByteCompiledLua =
-    name: text:
+  writeByteCompiledLuaWith =
+    pkgs: name: text:
     pkgs.runCommandLocal name { inherit text; } ''
       echo -n "$text" > "$out"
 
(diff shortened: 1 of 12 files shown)


lib/helpers.nix
```diff
@@ -97,5 +89,32 @@ let
     toLuaObject = helpers.lua.toLua;
     mkLuaInline = helpers.lua.mkInline;
   };
+
+  # Deprecated aliased for functions that rely on `pkgs`
```
traxys, line 93: I think this needs a comment to say that all the attrs sets inside this nested attr set _need_ to be added in the manual "recursiveUpdate", in case we forget in 12 monts

MattSturgeon: Hopefully it'll be gone in 12 months... But I agree let's add a comment.

MattSturgeon:
I'll try and update our _usage_ of most of these helpers to the new "With" versions tomorrow.

I'm thinking within the relevant modules we can do something like:

```nix
builders =
  genAttrs [
    "writeLua"
  ] (name: lib.nixvim.builders.${name + "With"} pkgs);
```

Or maybe there should be a helper function that applies `pkgs` to all the other functions in a set? 🤔

lib/builders-with-pkgs.nix
```diff
@@ -0,0 +1,71 @@
+{ lib, pkgs }:
+with lib.nixvim.builders;
```
khaneliman, line 2: Think we could be remove the `with ...` and be specific with inherits or inline?

lib/builders-with-pkgs.nix
```diff
@@ -0,0 +1,71 @@
...
+rec {
```
khaneliman, line 3: what makes this recursive? i don't see it at first glance

lib/default.nix
```diff
@@ -1,4 +1,5 @@
 # Args probably only needs pkgs and lib
+# TODO: build `lib` without needing `pkgs`
```
khaneliman, line 2: Were you planning on addressing this in this PR or a follow up?

khaneliman (commented):
Sorry, haven't been too active because I've been spending all my time on work stuff and fixing packages in the nixpkgs release cycle... LGTM overall, just a few questions while I was skimming it

lib/builders-with-pkgs.nix
MattSturgeon: In this case we know what we're importing with the `with` statement, and we happen to be using every single attr from `lib.nixvim.builders`...

I'm fine keeping it, but I appreciate `with` can give off some funky smells 😅

lib/builders-with-pkgs.nix
MattSturgeon: I don't think it is anymore (the old file was). I'll see if I can remove it.

lib/default.nix
MattSturgeon: That'll need some more thought. The "whole" lib includes the mkNixvimTest functions that all also expect `pkgs`, and this is all publicly documented and in the standalone flake template...

I don't have a firm idea, but my general thought is that we'll come up with new functions that don't require `pkgs`, then transition away from the old API after 24.11?

lib/builders-with-pkgs.nix
khaneliman: Ah, yeah, you're right. I didn't get to the builder.nix until after this part and didn't realize it was literally overriding everything from there. So, it is probably safe in this usage. I've just noticed it basically breaks LSPs ability to detect missing scopes when a `with` is present so I hate seeing them, now. :P

lib/builders-with-pkgs.nix
MattSturgeon: Refactored the implementation to be more dynamic. This file is now removed.

lib/builders-with-pkgs.nix
MattSturgeon: I replaced the `with` with an `inherit (lib.nixvim) builders` to be more explicit. But you'll never see it because I then removed the file entirely in some refactoring! :sweat_smile:

lib/helpers.nix
MattSturgeon: I added `` # WARNING: recursive merging must be handled manually; below in `self` ``

lib/builders.nix
```diff
@@ -1,135 +1,160 @@
...
+        name' = builtins.head match';
+      in
+      lib.optionalAttrs (match' != null) {
+        ${name'} = fn pkgs;
       }
-    );
-}
+    ) builders;
+in
+builders // { inherit withPkgs; }
```
khaneliman, line 160: could this be simplified into just having the withPkgs in the let block ?

khaneliman (approved):
LGTM overall, just curious about the builders definition structure

lib/builders.nix
MattSturgeon: We could do it recursively, true. The `builtins.match` should already filter out `withPkgs`

MattSturgeon: I've updated it to be recursive. The diff is smaller now because the indentation is no longer changed.

Are you still happy?

khaneliman: Looks better, at least to me.

MattSturgeon:
@mergifyio queue

PR #2234, opened 2024-09-13
lib: rename `helpers.nix` -> `default.nix`

MattSturgeon:
The old `default.nix` is inlined into the "lib" flake-module, which is the only place it was used.

This allows "helpers" to take its rightful place at the root of `./lib` 👑

I've also tweaked the implementation to use `lib.fix`, as I find this more readable than re-implementing it using `let self; in self`.


khaneliman (approved):
LGTM

MattSturgeon:
@mergifyio queue

Issue #2181, opened 2024-09-06
plugins/todo-comments: assertions read `cfg.keymaps.todoTelescope.key`, which may be an undefined option


trueNAHO:
Why are the `programs.nixvim.plugins.todo-comments.keymaps.todoTelescope` and `programs.nixvim.plugins.todo-comments.keymaps.todoTrouble` options not allowed to be `null`:

> ```nix
>     assertions = [
>       {
>         assertion = cfg.keymaps.todoTelescope.key or null != null -> config.plugins.telescope.enable;
>         [...]
>       }
>       {
>         assertion = cfg.keymaps.todoTrouble.key or null != null -> config.plugins.trouble.enable;
>         [...]
>       }
>     ];
> ```

These assertions were introduced in commit https://github.com/nix-community/nixvim/commit/379ae77a76ba2818b0c7876e97636f7064b84db6 ("plugins/todo-comments: migrate to mkNeovimPlugin") and require me to use the

```nix
programs.nixvim.plugins.todo-comments.keymaps.todoTrouble =
  lib.mkIf <CONDITION> { key = <VALUE>; };
```

workaround instead of the simpler

```nix
programs.nixvim.plugins.todo-comments.keymaps.todoTrouble.key =
  lib.mkIf <CONDITION> <VALUE>;
```

pattern in my [standalone Home Manager setup](https://github.com/trueNAHO/dotfiles/commit/f5726edfeb321269bc033622ccd3d959fe22026e).

Would it be possible to lift this non-`null` restriction?

---

Note: I have been inactive in NixVim discussions for more than half a year.

MattSturgeon:
These assertions don't say "cfg.keymaps.todoTelescope.key` can't be null, they say "if `cfg.keymaps.todoTelescope.key` isn't null, then `config.plugins.telescope.enable` should be true.

khaneliman:
Yeah, those assertions won't prevent that. I tested with adding todoTrouble that way and it works for me? https://github.com/khaneliman/astronvim/commit/43a06d25c196c492f1a5191b36b227e5b39e6f27

trueNAHO:
> These assertions don't say "`cfg.keymaps.todoTelescope.key` can't be null", they say "if `cfg.keymaps.todoTelescope.key` isn't null, then `config.plugins.telescope.enable` should be true".

Thanks for the clarification.

> EDIT: Oh, I see... when I disable trouble it throws the error because we're accessing the key option....
>
> ```shell
[...]

PR #2235, opened 2024-09-13
plugins/todo-comments: support conditional key mappings

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2181

### DIFF
--- a/plugins/by-name/todo-comments/default.nix
+++ b/plugins/by-name/todo-comments/default.nix
@@ -373,7 +373,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
               defaults = {
                 inherit action;
                 mode = "n";
+                key = null;
               };
+
+              key.type = with types; nullOr str;
 
               extraOptions = {
                 cwd = mkOption {
@@ -430,7 +433,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
     extraPackages = [ cfg.ripgrepPackage ];
 
     keymaps = lib.pipe cfg.keymaps [
-      (filterAttrs (n: keymap: keymap != null))
+      (filterAttrs (n: keymap: keymap != null && keymap.key != null))
       (mapAttrsToList (
         name: keymap: {
           inherit (keymap) key mode options;
--- a/tests/test-sources/plugins/by-name/todo-comments/default.nix
+++ b/tests/test-sources/plugins/by-name/todo-comments/default.nix
@@ -1,3 +1,4 @@
+{ lib, ... }:
 {
   empty = {
     plugins.todo-comments.enable = true;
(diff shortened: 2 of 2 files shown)


GaetanLepage (approved):
LGTM

lib/keymap-helpers.nix
```diff
@@ -107,9 +107,10 @@ rec {
           (lib.optionalAttrs (isAttrs key || key) {
             key = lib.mkOption (
               {
-                type = types.str;
+                type = with types; nullOr str;
```
MattSturgeon, line 110: Hm, I'm not convinced we _always_ want this to be nullable? E.g. in the top-level `keymaps` option, it wouldn't make sense to have a mapping with a null lhs.

lib/keymap-helpers.nix
```diff
@@ -107,9 +107,10 @@ rec {
...
                 description = "The key to map.";
                 example = "<C-m>";
+                default = null;
```
MattSturgeon, line 113: If we do this, then this could be `defaults.key or null` instead of the optionalAttrs below.

I'm wondering if it makes more sense to check the submodule's `_module.args.name` and use that as the default?

That way when the submodule is used in an `attrsOf` type, the attr name would be the default for the key option.

MattSturgeon (commented):
Could you explain the use-case for having `key` set to null? Currently I don't understand what that gains us.

IMO the default should be whatever the plugin would do when set to null. For example, if a plugin would do `mapAttrs (name: mapping: { key = if mapping.key == null then name else mapping.key; })`, then we should just have the option default to the `name` module arg.

Remember that these mapping submodules aren't used for `settings` options, so having a useful default is also valid if we can think of one.

In the todo plugin that has issues, what is `key` used for, and what is supposed to happen when it is or is not defined?

lib/keymap-helpers.nix
khaneliman: Well, it needs to be something to access it for conditional mappings like in the test case where we don't want to set it sometimes. If it's a string we have to determine if it's valid to be a keymap, whereas if it's null we know to skip it.

MattSturgeon: Would that make more sense if the keymap as a whole was nullable?

I just struggle to understand having a partial keymap, where we have a action, mode, etc, but no key.

khaneliman:
Alright, @MattSturgeon helped me find a method to localize it to this plugin since the other plugins didn't benefit from the change.

MattSturgeon (approved):
LGTM! Thanks!

We can revisit wider improvements to the submodule itself later, but having this local for now keeps our options open.

khaneliman:
@mergifyio queue

PR #2238, opened 2024-09-13
plugins/wtf: add history and grep_history keymaps

khaneliman:
While working on https://github.com/nix-community/nixvim/pull/2235 I found these 2 in the docs for the plugin and they weren't supported.

### DIFF
--- a/plugins/by-name/wtf/default.nix
+++ b/plugins/by-name/wtf/default.nix
@@ -24,6 +24,18 @@ let
       mode = "n";
       action.__raw = "require('wtf').search";
     };
+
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):
Could probably be done freeform, and maybe set `action` using a submodule (`action = "require('wtf').${name}`, with `name` coming from an attrsOf composite type, like we do in `extraFiles`). But those changes would be larger and would need some consideration.

This is great as is, though. So thanks!

khaneliman:
@mergifyio queue

Issue #2103, opened 2024-08-28
[PLUGIN REQUEST] ts-comments.nvim


SymphonySimper:
| Field    | Description |
|----------|-------------|
| Plugin   | `ts-comments.nvim`  |
| Homepage | `https://github.com/folke/ts-comments.nvim/`     |
| Nixpkgs  | `vimPlugins.ts-comments-nvim` |

## Extra Information

Requires Neovim >= 0.10.0

PR #2247, opened 2024-09-15
plugins/ts-comments: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2103

### DIFF
--- a/plugins/by-name/ts-comments/default.nix
+++ b/plugins/by-name/ts-comments/default.nix
@@ -1,0 +1,89 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "ts-comments";
+  originalName = "ts-comments.nvim";
+  package = "ts-comments-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsOptions = {
+    lang =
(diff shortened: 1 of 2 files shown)


plugins/by-name/ts-comments-nvim/default.nix
```diff
@@ -0,0 +1,90 @@
...
+          graphql = "# %s";
+          handlebars = "{{! %s }}";
+          hcl = "# %s";
+          html = "";
+          hyprlang = "# %s";
+          ini = "; %s";
+          ipynb = "# %s";
+          javascript = {
+            __default = "// %s";
+            "/* %s */" = null;
```
traxys, line 42: Shouldn't this be unkeyed entries?

traxys: Or is it an alternate syntax accepted by the plugin?

khaneliman: Yeah, just late night stupidity

MattSturgeon (approved):
I don't see any issues. The pluginDefault _could_ be misleading, if users assume defining one attr will override the others, but that's a minor issue and the solution isn't worth the verbosity.

khaneliman:
@mergifyio queue

Issue #1527, opened 2024-05-16
[PLUGIN REQUEST] Compiler.nvim


PowerUser64:
| Field    | Description |
|----------|-------------|
| Plugin   | Compiler.nvim  | 
| Homepage | https://github.com/Zeioth/compiler.nvim | 
| Nixpkgs  | false | 

## Extra Information

This plugin depends on [overseer.nvim](https://github.com/stevearc/overseer.nvim), which is in nixpkgs but not in nixvim (#1528).

sachinchaudhary1808:
this is a good one i think i wanted this but it was not in nixpkgs 
so someone has to add compiler.nvim as nix pkgs ?

PowerUser64:
I asked @GaetanLepage how to add vim plugins to nixpkgs and apparently it's not that hard, but when I followed these steps (linked later) to try submitting a PR, I didn't get the same result as what was in some of the PR's I was seeing. Maybe if you try you'll get different results? The comment with how to do it is here: https://github.com/nix-community/nixvim/issues/1338#issuecomment-2025118208

GaetanLepage:
> I asked @GaetanLepage how to add vim plugins to nixpkgs and apparently it's not that hard, but when I followed these steps (linked later) to try submitting a PR, I didn't get the same result as what was in some of the PR's I was seeing. Maybe if you try you'll get different results? The comment with how to do it is here: [#1338 (comment)](https://github.com/nix-community/nixvim/issues/1338#issuecomment-2025118208)

What method have you tried ? Running the script or manually appending the relevant snippet to `generated.nix` ?

PowerUser64:
https://github.com/NixOS/nixpkgs/pull/312410 a PR has been made :white_check_mark:

PR #2248, opened 2024-09-15
plugins/compiler: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1527

From what I could tell, there's not much to setup with this?

### DIFF
--- a/plugins/by-name/compiler/default.nix
+++ b/plugins/by-name/compiler/default.nix
@@ -1,0 +1,19 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "compiler";
+  originalName = "compiler.nvim";
+  package = "compiler-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    > [!Note]
+    > Some languages require you manually install their compilers in your machine, so `compiler.nvim` is able to call them.
+    > Please check [here], as the packages will be different depending your operative system.
+
+    [here]: https://github.com/Zeioth/Compiler.nvim/wiki/how-to-install-the-required-dependencies
+  '';
+}
--- a/tests/test-sources/plugins/by-name/compiler/default.nix
+++ b/tests/test-sources/plugins/by-name/compiler/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.compiler.enable = true;
+  };
+}


traxys:
Maybe we should add [this](https://github.com/Zeioth/compiler.nvim?tab=readme-ov-file#required-system-dependencies) warning to our documentation?
> Some languages require you manually install their compilers in your machine, so compiler.nvim is able to call them. [Please check here](https://github.com/Zeioth/Compiler.nvim/wiki/how-to-install-the-required-dependencies), as the packages will be different depending your operative system.

khaneliman:
> Maybe we should add [this](https://github.com/Zeioth/compiler.nvim?tab=readme-ov-file#required-system-dependencies) warning to our documentation?
> 
> > Some languages require you manually install their compilers in your machine, so compiler.nvim is able to call them. [Please check here](https://github.com/Zeioth/Compiler.nvim/wiki/how-to-install-the-required-dependencies), as the packages will be different depending your operative system.

Tried adding the note, is that how you guys will be using the new admonitions?

khaneliman:
Alright, fixed the admonition. 
image

plugins/by-name/compiler/default.nix
```diff
@@ -0,0 +1,19 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "compiler";
+  originalName = "compiler.nvim";
+  package = "compiler-nvim";
```
MattSturgeon, line 8: Should we be passing `callSetup = false` and/or `hasSettings = false`?

khaneliman: It looked like upstream had a setup function and passes an opts table, but they don't currently have documented configuration.

MattSturgeon: The lazy.nvim usage example includes `opts = {}`, making me think they _do_ have a setup function and maybe a settings table?

khaneliman: So, I didn't want to restrict future configuration if they do start using the opts that can be provided. https://github.com/Zeioth/compiler.nvim/blob/e889774c7f2fefad60f8e6638d324223b07d2624/lua/compiler/init.lua#L9

MattSturgeon: I came to the same conclusion when I did my own research :+1: 

The setup function does take an opts parameter, but it is currently unused internally.

I agree we should keep things as-is.

khaneliman:
@mergifyio queue

Issue #1528, opened 2024-05-16
[PLUGIN REQUEST] overseer.nvim


PowerUser64:
| Field    | Description |
|----------|-------------|
| Plugin   | overseer.nvim  | 
| Homepage | https://github.com/stevearc/overseer.nvim | 
| Nixpkgs  | vimPlugins.overseer-nvim | 

## Extra Information

Required for Compiler.nvim.

GaetanLepage:
@PowerUser64 wouldn't you want to give it a shot ?
Honestly, it is not that hard to do and I can help you getting a PR merged :)
Feel free to ask any question on matrix also.

(If you do not want to try or don't have time, no worry, I'll do it)

PowerUser64:
Yeah, I'd be interested to learn how. I'll send you a message later!

PR #2249, opened 2024-09-15
plugins/overseer: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1528

### DIFF
--- a/plugins/by-name/overseer/default.nix
+++ b/plugins/by-name/overseer/default.nix
@@ -1,0 +1,132 @@
+{
+  lib,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "overseer";
+  originalName = "overseer.nvim";
+  package = "overseer-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsOptions = {
+    strategy = defaultNullOpts.mkStr "terminal" ''
+      Default task strategy.
+    '';
+
+    templates = defaultNullOpts.mkListOf types.str [ "builtin" ] ''
+      Template modules to load.
+    '';
+
+    auto_detect_success_color = defaultNullOpts.mkBool true ''
+      When true, tries to detect a green color from your colorscheme to use for success highlight.
+    '';
+
+    dap = defaultNullOpts.mkBool true ''
+      Whether to patch nvim-dap to support preLaunchTask and postDebugTask
+    '';
+
+    task_list =
+      defaultNullOpts.mkAttrsOf types.anything
+        {
+          default_detail = 1;
+          max_width = [
+            100
+            0.2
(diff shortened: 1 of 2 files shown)


plugins/by-name/overseer-nvim/default.nix
```diff
@@ -0,0 +1,86 @@
+{
+  lib,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "overseer-nvim";
```
GaetanLepage, line 10: Genuine question, why not `overseer` in this case ?

khaneliman: Didn't know what the preference was for following repo name or lua setup name

GaetanLepage: The convention we have been using until now mostly correspond to the lua name, i.e. plugin name without the `.nvim` suffix.

khaneliman: Sounds good I'll change these init prs to do that

GaetanLepage:
Could you also take a quick look at the previous draft #1759 just in case something interesting lies there ?

GaetanLepage (approved):
LGTM !

khaneliman:
> Could you also take a quick look at the previous draft #1759 just in case something interesting lies there ?

I didn't notice that PR but, it looks like they were going full verbose, I was just just aiming for the functionality specific options here in the declared options and leaving out the UI options. So, there were a couple extra keymap options I'll include since they did the heav lifting of defining them already.

GaetanLepage:
> I didn't notice that PR but, it looks like they were going full verbose, I was just just aiming for the functionality specific options here in the declared options and leaving out the UI options. So, there were a couple extra keymap options I'll include since they did the heavy lifting of defining them already.

Ok good for me !

plugins/by-name/overseer/default.nix
```diff
@@ -0,0 +1,132 @@
...
+            n = {
+              "<CR>" = "Submit";
+              "<C-s>" = "Submit";
+              "q" = "Cancel";
+              "?" = "ShowHelp";
+            };
+          }
+          ''
+            Set keymap to false to remove default behavior
+            You can add custom keymaps here as well (anything vim.keymap.set accepts)
```
GaetanLepage, line 123: Ensure to end all sentences with a period.
Like so, we can provide a consistent documentation style.

khaneliman:
Also looks like the old keymap option might not be fully satisfactory.. Might need to update to follow https://github.com/nix-community/nixvim/issues/603

GaetanLepage (approved):

khaneliman:
Just pushed a subtle change to untie the keymaps from our submodule and allow any input so we dont accidentally restrict a user from upstream keymap usage.

khaneliman:
@mergifyio queue

PR #2250, opened 2024-09-15
plugins/idris2: init

mitchmindtree:
Adds the `idris2` neovim plugin.

Also changes `idris2-lsp` to enable this new `idris2` plugin when enabled, as it's required for `idris2-lsp` to work at all. I'm not 100% sure this is the best way to achieve this - let me know if you'd prefer this was done another way!

### DIFF
--- a/plugins/by-name/idris2/default.nix
+++ b/plugins/by-name/idris2/default.nix
@@ -1,0 +1,11 @@
+{
+  helpers,
+  lib,
+  ...
+}:
+helpers.neovim-plugin.mkNeovimPlugin {
+  name = "idris2";
+  originalName = "idris2";
+  package = "idris2-nvim";
+  maintainers = [ lib.maintainers.mitchmindtree ];
+}
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -256,12 +256,20 @@ let
     }
     {
       name = "idris2-lsp";
-      description = "Idris 2 Language Server";
+      description = ''
+        Idris 2 Language Server.
+        Enabling this also enables the required `idris2` plugin.
+      '';
       serverName = "idris2_lsp";
       package = [
         "idris2Packages"
         "idris2Lsp"
       ];
+      extraConfig =
+        cfg:
+        mkIf cfg.enable {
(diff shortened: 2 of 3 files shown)


traxys:
Why is the plugin required? Just for the filetype? In that case I would be more comfortable adding either a note in the documentation, or setting the filetype association in the LSP extra configuration

mitchmindtree:
> Why is the plugin required? Just for the filetype?

No, the `idris2` plugin provides an interface to the `idris2` compiler, and `idris2-lsp` appears to use this and doesn't function whatsoever without the `idris2` plugin present. I figured this is a convenient way to make it enabled by default?

traxys:
I personally prefer to add warnings over enabling behind a user's back, but maybe someone has as different suggestion? (@MattSturgeon, @khaneliman, @GaetanLepage ?)

MattSturgeon:
> I personally prefer to add warnings over enabling behind a user's back, but maybe someone has as different opinion?

I don't have a strong opinion.

Most module-based config in the nix ecosystem takes a sane defaults behind a single enable option approach. In nixvim we try to be a bit more explicit about what ends up in init.lua, but I don't have a strong objection to enabling things by default when it's simple & obvious, the only sane choice 90% of the time, well documented, and easy to override.

If we do decide to go down the enabled-by-default route here, I'd recommend using `lib.mkDefault` to make it easier to override though.

mitchmindtree:
>  I'd recommend using `lib.mkDefault`

Good call, I've updated the `idris2-lsp` to enable the required `idris2` plugin with `lib.mkDefault true`.

khaneliman:
> I personally prefer to add warnings over enabling behind a user's back, but maybe someone has as different opinion? (@MattSturgeon, @khaneliman, @GaetanLepage ?)

Yeah, I think for a lot of things where it's an extra configuration option to ancillary tools/behavior it makes sense to not auto enable and warn about requiring another configuration. When it's basically required I think it makes sense to default enable but allow disabling, in case they want to handle the dependency a different way.

traxys (approved):

traxys:
@Mergifyio queue

GaetanLepage:
I agree with what has been said. `mkDefault true` is fine to me.

PR #2253, opened 2024-09-15
plugins/lsp: add ts-ls

tsssni:
https://github.com/neovim/nvim-lspconfig/pull/3232

Set `plugins.lsp.tsserver.enable = true` results in the warning `tsserver is deprecated, use ts_ls instead` when launching neovim. `lsp.ts-ls` is added to address the issue, and `lsp.tsserver` is retained since it still exists in current version of nvim-lspconfig.


plugins/lsp/language-servers/default.nix
```diff
@@ -620,6 +620,12 @@ let
...
     }
+    {
+      name = "ts-ls";
+      serverName = "ts_ls";
+      description = "tsserver for TypeScript";
+      package = "typescript-language-server";
+    }
     {
       name = "tsserver";
       description = "tsserver for TypeScript";
```
MattSturgeon, line 631: Should nixvim print a deprecation warning when this used?

plugins/lsp/language-servers/default.nix
```diff
@@ -620,6 +620,12 @@ let
       description = "tinymist for Typst";
       settingsOptions = import ./tinymist-settings.nix { inherit lib helpers; };
     }
+    {
+      name = "ts-ls";
+      serverName = "ts_ls";
+      description = "tsserver for TypeScript";
+      package = "typescript-language-server";
```
khaneliman, line 627: this is still the same package just a different executable?

plugins/lsp/language-servers/default.nix
MattSturgeon: Alternatively, seeing as both use the same package under the hood perhaps we could simply make this a `mkRenameOptioModule` alias pointing to the new name?

tests/test-sources/plugins/lsp/_lsp.nix
```diff
@@ -210,6 +210,7 @@
             texlab.enable = true;
             tflint.enable = true;
             tinymist.enable = true;
+            ts-ls.enale = true;
```
khaneliman, line 213:
```suggestion
            ts-ls.enable = true;
```

plugins/lsp/language-servers/default.nix
```diff
@@ -620,6 +620,12 @@ let
       description = "tinymist for Typst";
       settingsOptions = import ./tinymist-settings.nix { inherit lib helpers; };
     }
+    {
+      name = "ts-ls";
+      serverName = "ts_ls";
+      description = "tsserver for TypeScript";
```
khaneliman, line 626:
```suggestion
      description = "ts_ls for TypeScript";
```
> Note that `ts_ls` is not `tsserver`.

khaneliman:
Reading the error this might be dependent on https://github.com/nix-community/nixvim/pull/2252

plugins/lsp/language-servers/default.nix
```diff
@@ -708,5 +713,6 @@ in
       ./pylsp.nix
       ./rust-analyzer.nix
       ./svelte.nix
+      tsserverRenameModule
```
MattSturgeon, line 716: I would either just do this inline:

```suggestion
      (lib.mkRenamedOptionModule (baseLspPath ++ [ "tsserver" ]) (baseLspPath ++ [ "ts-ls" ]))
```

Or I would have a set of renames we could expand on in the future:

```nix
  renamedServers = {
    tsserver = "ts-ls";
  };
  renameModules = mapAttrsToList (old: new: lib.mkRenamedOptionModule (baseLspPath ++ [ old ]) (baseLspPath ++ [ new ]) renamedServers
in
  lspModules
  ++ renameModules
  ++ [
    ./ccls.nix
    # ...
```

plugins/lsp/language-servers/default.nix
```diff
@@ -621,8 +621,9 @@ let
       settingsOptions = import ./tinymist-settings.nix { inherit lib helpers; };
     }
     {
-      name = "tsserver";
-      description = "tsserver for TypeScript";
+      name = "ts-ls";
+      serverName = "ts_ls";
+      description = "ts-ls for TypeScript";
```
khaneliman, line 626:
```suggestion
      description = "ts_ls for TypeScript";
```

The actual server is ts_ls, even though our module uses snake case.

plugins/lsp/language-servers/default.nix
```diff
@@ -700,8 +701,14 @@ in
     let
       mkLsp = import ./_mk-lsp.nix;
       lspModules = map mkLsp servers;
+      baseLspPath = [ "plugins" "lsp" "servers" ];
+      renamedServers = {
+        tsserver = "ts-ls";
+      };
```
MattSturgeon, line 707: It's kinda confusing having two `let` blocks in this file, but I think this would make more sense in the first `let` block? As in directly after the `servers` list.

That way the first let block is purely declarative while this one is purely implementation.

khaneliman:
LGTM for the most part, just fixing up our flake update real quick so this is supported.

khaneliman:
Looks like we're getting infinite recursion message from the rename?

MattSturgeon:
> Looks like we're getting infinite recursion message from the rename?

Trying to debug locally, but it's not making much sense...

I also stumbled on some places where we reference `tsserver` internally, e.g. `plugins.none-ls.sources.formatting.prettier`.

MattSturgeon:
In fact it is the prettier definition that seems to be causing inf.rec:

while evaluating definitions from `/nix/store/42mc0najpyf3izl3f62plm6hj0g7pfzn-source/plugins/none-ls/prettier.nix'

tsssni:
> In fact it is the prettier definition that seems to be causing inf.rec:
> 
> while evaluating definitions from `/nix/store/42mc0najpyf3izl3f62plm6hj0g7pfzn-source/plugins/none-ls/prettier.nix'

Fixed by replacing `tsserver` with `ts-ls` in prettier.nix

MattSturgeon:
> Fixed by replacing `tsserver` with `ts-ls` in prettier.nix

Well done, but that does make me worry that any renames under LSP/language-servers will have the same issue when someone tries to use the old name...

I think we'll have to investigate this separately.

For now, the outstanding ci failures are formatting (you need to run `nix fmt`) and:

> ERROR: [lspconfig] Cannot access configuration for ts_ls. Ensure this server is listed in `server_configurations.md` or added as a custom server.

Perhaps we are still waiting on a nixpkgs bump?

khaneliman:
> > Fixed by replacing `tsserver` with `ts-ls` in prettier.nix
> 
> Well done, but that does make me worry that any renames under LSP/language-servers will have the same issue when someone tries to use the old name...
> 
> I think we'll have to investigate this separately.
> 
> For now, the outstanding ci failures are formatting (you need to run `nix fmt`) and:
> 
> > ERROR: [lspconfig] Cannot access configuration for ts_ls. Ensure this server is listed in `server_configurations.md` or added as a custom server.
> 
> Perhaps we are still waiting on a nixpkgs bump? Or this may be resolved by rebasing.

I can try using this branch with my config, since i have tsserver set.

khaneliman:
Rename seems to work, fine.

khaneliman (approved):
LGTM thanks

MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2254, opened 2024-09-15
flake: add an empty `nixvimConfiguration` to the `legacyPackages` output

MattSturgeon:
This was already used in a couple test cases, however having it as a flake output allows for 

1. easier debugging using `nix repl` or `nix eval` 
2. doing things like `nix eval .#nixvimConfiguration.config.plugins --apply builtins.attrNames --json` to very quickly get a JSON list of all plugins.

This would be useful in a CI task that looks for newly added plugins, for example.

NOTE: it has to go in `legacyPackages` and not `packages` because the latter can only contain derivations. Non-drv attrsets will throw an error when included in `packages`.

### DIFF
--- a/flake-modules/legacy-packages.nix
+++ b/flake-modules/legacy-packages.nix
@@ -2,7 +2,7 @@
   perSystem =
     {
       pkgs,
-      config,
+      helpers,
       makeNixvimWithModule,
       ...
     }:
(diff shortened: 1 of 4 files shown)


khaneliman (approved):
LGTM

MattSturgeon:
@mergifyio queue

PR #2257, opened 2024-09-15
plugins/smartcolumn: init

khaneliman:
Create module for https://github.com/m4xshen/smartcolumn.nvim

### DIFF
--- a/plugins/by-name/smartcolumn/default.nix
+++ b/plugins/by-name/smartcolumn/default.nix
@@ -1,0 +1,80 @@
+{
+  lib,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "smartcolumn";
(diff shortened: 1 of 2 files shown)


plugins/by-name/smartcolumn/default.nix
```diff
@@ -0,0 +1,80 @@
...
+  originalName = "smartcolumn.nvim";
+  package = "smartcolumn-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsOptions = {
+    colorcolumn = defaultNullOpts.mkNullable (with types; either str (listOf str)) "80" ''
+      Column with to highlight.
+      Supports multiple values for more column highlights.
+    '';
```
MattSturgeon, line 20: Seems strange upstream would expect a string :thinking:

MattSturgeon: But it is correct https://github.com/m4xshen/smartcolumn.nvim?tab=readme-ov-file#-configuration

MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2258, opened 2024-09-15
plugins/glow: remove helpers; support custom styles

khaneliman:
Forgot we created a module for this while trying to create modules for plugins in my config. I pass in a style path in my glow config to load a custom style and the module didn't support it.

### DIFF
--- a/plugins/by-name/glow/default.nix
+++ b/plugins/by-name/glow/default.nix
@@ -1,10 +1,12 @@
 {
   lib,
-  helpers,
   pkgs,
   ...
 }:
-helpers.neovim-plugin.mkNeovimPlugin {
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
   name = "glow";
   originalName = "glow.nvim";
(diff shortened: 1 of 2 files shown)


plugins/by-name/glow/default.nix
```diff
@@ -35,28 +37,31 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
-      "dark"
-      "light"
-    ] (helpers.mkRaw "vim.opt.background") "Glow style.";
+    style = defaultNullOpts.mkNullable (
+      with lib.types;
+      either (maybeRaw str) (enum [
+        "dark"
+        "light"
+      ])
+    ) (lib.nixvim.mkRaw "vim.opt.background") "Glow style.";
```
MattSturgeon, line 46: Using str and an enum of strings together feels odd. Should the option just be using `mkStr`?

khaneliman: I thought so too, but I didn't know if this helped show the default 2 options from the plugin better.

MattSturgeon (approved):

khaneliman:
@mergifyio queue

PR #2267, opened 2024-09-18
plugins/nvim-surround: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1346
Closes https://github.com/nix-community/nixvim/pull/1371
Rename surround to vim-surround and init nvim-surround.

### DIFF
--- a/plugins/by-name/nvim-surround/default.nix
+++ b/plugins/by-name/nvim-surround/default.nix
@@ -1,0 +1,129 @@
+{
+  lib,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nvim-surround";
+  originalName = "nvim-surround";
+  package = "nvim-surround";
+
+  maintainers = with lib.maintainers; [
+    khaneliman
+    AndresBermeoMarinelli
+  ];
+
+  settingsOptions = {
+    keymaps = defaultNullOpts.mkAttrsOf types.str ''
+      {
+        insert = "<C-g>s";
+        insert_line = "<C-g>S";
+        normal = "ys";
+        normal_cur = "yss";
+        normal_line = "yS";
+        normal_cur_line = "ySS";
+        visual = "S";
+        visual_line = "gS";
+        delete = "ds";
+        change = "cs";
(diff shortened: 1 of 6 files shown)


plugins/deprecation.nix
```diff
@@ -10,13 +10,31 @@ let
...
+  imports =
+    (lib.mapAttrsToList (
+      name:
+      lib.mkRemovedOptionModule [
+        "plugins"
+        name
+      ]
+    ) removed)
+    ++ (lib.mapAttrsToList (
+      oldName: newName:
```
MattSturgeon, line 29: I think `old` and `new` is fine, the added verbosity of `*Name` shouldn't be needed given the surrounding context.

plugins/by-name/nvim-surround/default.nix
```diff
@@ -0,0 +1,127 @@
...
+              };
+            }) "An attribute set with two keys: `target` and `replacement`.";
+          };
+        };
+      in
+      lib.nixvim.mkNullOrOption (types.attrsOf surroundType) ''
+        Associates each key with a "surround". The attribute set contains
+        the 'add', 'find', 'delete', and 'change' keys.
+
+        Default: see upstream documentation.
```
MattSturgeon, line 124: Couldn't wee make use of the `pluginDefault` field here?

E.g. `pluginDefault = literalMD "See [upstream documentation](http://......)"`

plugins/by-name/nvim-surround/default.nix
```diff
@@ -0,0 +1,127 @@
...
+                        end
+                    end
+                  '';
+                };
+              };
+            }) "An attribute set with two keys: `target` and `replacement`.";
+          };
+        };
+      in
+      lib.nixvim.mkNullOrOption (types.attrsOf surroundType) ''
```
MattSturgeon, line 120: I assume no `rawLua` support is a deliberate choice to avoid https://github.com/nix-community/nixvim/issues/1618?

MattSturgeon (approved):
Looks good. I like how the rename migration is handled.

Feels like `plugins/deprecation.nix` may make https://github.com/nix-community/nixvim/pull/1657 redundant, if we're happy having a one-way rename with no plan for another plugin to "take over" the name at a later date.

khaneliman:
@mergifyio queue

Issue #1380, opened 2024-04-05
[PLUGIN REQUEST] `package-info.nvim`


pinarruiz:
| Field    | Description |
|----------|-------------|
| Plugin   | `package-info.nvim`  | 
| Homepage | https://github.com/vuki656/package-info.nvim     | 
| Nixpkgs  | `vimPlugins.package-info-nvim` | 

## Extra Information

From plugin's [README](https://github.com/vuki656/package-info.nvim?tab=readme-ov-file#-features):

> - Display latest dependency versions as virtual text
> - Upgrade dependency on current line to latest version
> - Delete dependency on current line
> - Install a different version of a dependency on current line
> - Install new dependency
> - Automatic package manager detection
> - Loading animation hook (to be placed in status bar or anywhere else)

PR #2268, opened 2024-09-18
plugins/package-info: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1380

### DIFF
--- a/plugins/by-name/package-info/default.nix
+++ b/plugins/by-name/package-info/default.nix
@@ -1,6 +1,7 @@
 {
   lib,
   pkgs,
+  config,
   ...
 }:
 let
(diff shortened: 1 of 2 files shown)


plugins/by-name/package-info/default.nix
```diff
@@ -0,0 +1,122 @@
...
+
+    packageManagerPackage =
+      lib.mkPackageOption pkgs
+        [
+          "nodePackages"
+          "npm"
+        ]
+        {
+          nullable = true;
+        };
```
MattSturgeon, line 105: How likely is it that users will want us to install this by default? I get the impression that the plugin tries to glean a package manager heuristically, in which case it seems likely it is installed externally?

If you don't think its needed, we could have this default to null and include an example instead?

```nix
lib.mkPackageOption pkgs "package manager" {
  nullable = true;
  default = null;
  example = "pkgs.nodePackages.npm";
}
```

khaneliman: Yeah, I think it makes sense, in this case, to default to null. Most people installing this plugin would have their own package manager and/or a dev shell.

khaneliman:
@mergifyio queue

PR #2272, opened 2024-09-18
lib: fix escaping bugs in wrapVimscriptForLua and wrapLuaForVimscript

jfly:
These functions had very similar bugs: they didn't check if their chosen "close token" was already present in the string they're escaping.

I went ahead and did the work implied by the TODOs: search for a "close token" that is *not* in the original string. Pretty simple concept, but it turned into an annoying amount of code. I couldn't find anything in upstream nixpkgs lib, or some clever insight about lua/vimscript that makes this work unecessary, but I'll be thrilled (and a little bummed about a wasted afternoon) to learn about something.

### DIFF
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -125,27 +125,98 @@ rec {
     end
   '';
 
+  /**
+    Convert the given String to a Lua [long literal].
+    For example, you could use this to safely pass a Vimscript string to the
+    `vim.cmd` function.
+
+    [long literal]: https://www.lua.org/manual/5.4/manual.html#3.1
+
+    # Examples
+
+    ```nix
+    nix-repl> toLuaLongLiteral "simple"
+    "[[simple]]"
+    ```
+
+    ```nix
+    nix-repl> toLuaLongLiteral "]]"
+    "[=[]]]=]"
+    ```
+
+    # Type
+
+    ```
+    toLuaLongLiteral :: String -> String
+    ```
+  */
+  toLuaLongLiteral =
+    string:
+    let
+      findTokens =
+        depth:
+        let
+          infix = lib.strings.replicate depth "=";
+          tokens.open = "[${infix}[";
+          tokens.close = "]${infix}]";
+        in
+        if lib.hasInfix tokens.close string then findTokens (depth + 1) else tokens;
+
+      tokens = findTokens 0;
+    in
+    tokens.open + string + tokens.close;
+
+  /**
+    Convert the given String into a Vimscript [:let-heredoc].
+    For example, you could use this to invoke [:lua].
+
+    [:let-heredoc]: https://neovim.io/doc/user/eval.html#%3Alet-heredoc
+    [:lua]: https://neovim.io/doc/user/lua.html#%3Alua-heredoc
+
+    # Examples
+
+    ```nix
+    toVimscriptHeredoc "simple"
+    => "<< EOF\nsimple\nEOF"
+    ```
+
+    ```nix
+    toVimscriptHeredoc "EOF"
+    => "<< EOFF\nEOF\nEOFF"
+    ```
+
+    # Type
+
+    ```
+    toVimscriptHeredoc :: String -> String
+    ```
+  */
+  toVimscriptHeredoc =
+    string:
+    let
+      findToken =
+        depth:
+        let
+          token = "EOF" + lib.strings.replicate depth "F";
+        in
+        if lib.hasInfix token string then findToken (depth + 1) else token;
+
+      token = findToken 0;
+    in
+    ''
+      << ${token}
+      ${string}
+      ${token}'';
+
   # Wrap Vimscript for using in lua,
   # but only if the string contains something other than whitespaces
-  # TODO: account for a possible ']]' in the string
   wrapVimscriptForLua =
-    string:
-    lib.optionalString (hasContent string) ''
-      vim.cmd([[
-      ${string}
-      ]])
-    '';
+    string: lib.optionalString (hasContent string) "vim.cmd(${toLuaLongLiteral string})";
 
   # Wrap lua script for using in Vimscript,
   # but only if the string contains something other than whitespaces
-  # TODO: account for a possible 'EOF' if the string
   wrapLuaForVimscript =
-    string:
-    lib.optionalString (hasContent string) ''
-      lua << EOF
-      ${string}
-      EOF
-    '';
+    string: lib.optionalString (hasContent string) "lua ${toVimscriptHeredoc string}";
 
   # Split a list into a several sub-list, each with a max-size of `size`
   groupListBySize =
--- a/tests/lib-tests.nix
+++ b/tests/lib-tests.nix
@@ -374,6 +374,41 @@ let
         ];
       };
     };
+
+    testEscapeStringForLua = {
+      expr = lib.mapAttrs (_: helpers.utils.toLuaLongLiteral) {
+        simple = "simple";
+        depth-one = " ]] ";
+        depth-two = " ]] ]=] ";
+      };
+      expected = {
+        simple = "[[simple]]";
+        depth-one = "[=[ ]] ]=]";
+        depth-two = "[==[ ]] ]=] ]==]";
+      };
+    };
+
+    testEscapeStringForVimscript = {
+      expr = lib.mapAttrs (_: helpers.utils.toVimscriptHeredoc) {
+        simple = "simple";
+        depth-one = "EOF";
+        depth-two = "EOF EOFF";
+      };
+      expected = {
+        simple = ''
+          << EOF
+          simple
+          EOF'';
+        depth-one = ''
+          << EOFF
+          EOF
+          EOFF'';
+        depth-two = ''
+          << EOFFF
+          EOF EOFF
+          EOFFF'';
+      };
+    };
   };
 in
 if results == [ ] then


MattSturgeon:
Thanks a lot for implementing this! It will make nesting vim in lua (and vice versa) much more robust!

My gut feeling is it can be simplified a little, but I'll take a proper look soon and see if I have any suggestions or ideas.

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+    We invoke `buildSurround` builder starting a depth 0, and increasing the
+    depth until we have a Surround where `surround.close` is not in the
+    string we're escaping.
+
+    We return the original string, first (possibly, depending on
+    `surroundWithNewlines`) surrounded by newlines , and secondly surrounded by
+    the Surround we found above.
+    ```
+  */
+  escapeString =
```
jfly, line 170: I'm just realizing this is a pretty generic name for a very specific thing. A better name might be something like `heredocStyleEscapeString`. An even better thing to do might be to move this out of the `lib` attrset and into a let binding on top to emphasize that this is an implementation detail, and not meant to be used outside of nixvim.

Happy to make whatever changes here folks suggest.

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+      surround = findSurround {
+        inherit string buildSurround;
+        depth = 0;
+      };
+      newlines = if surroundWithNewlines then "\n" else "";
+    in
+    "${surround.open}${newlines}${string}${newlines}${surround.close}";
+
+  /**
+    Given a string, escape it as a Lua [long literal].
```
MattSturgeon, line 203: I don't think this really relates to escaping, so I'm interested in alternate naming...

To me, escaping might be converting `\n` into `\\n`, for example.

I think what we're doing here is safely _embedding_ a string into some language using enclosing syntax that we are confident wraps the string without an early termination.

I guess, conceptually, it is similar to escaping (otherwise) terminating quotes from a string literal. However functionally we implement it differently; we aren't escaping the string, we are adapting the opening/closing syntax to better fit the string.

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+        let
+          surround = buildSurround depth;
+        in
+        if lib.hasInfix surround.close string then
+          findSurround {
+            inherit string buildSurround;
+            depth = depth + 1;
+          }
+        else
+          surround;
```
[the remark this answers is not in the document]
MattSturgeon: I am open to keeping the current structure, but I thing `surroundWithNewlines` should go regardless.

The same thing could be achieved by including newlines in the built `open` and `close` sequences, for example.

Do you think there is a general use for the function currently named `escapeString`?

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+      buildSurround,
+      surroundWithNewlines,
+    }:
+    let
+      findSurround =
+        {
+          string,
+          buildSurround,
+          depth,
+        }:
```
MattSturgeon, line 182: If we keep things as they are, there is no need to pass `string` and `buildSurrond` into `findSurround`, because they are already lexically in scope (i.e. in scope via closure):

```suggestion
      findSurround =
        depth:
```

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+        else
+          surround;
+
+      surround = findSurround {
+        inherit string buildSurround;
+        depth = 0;
+      };
+      newlines = if surroundWithNewlines then "\n" else "";
+    in
+    "${surround.open}${newlines}${string}${newlines}${surround.close}";
```
MattSturgeon, line 200: I wonder if this is more readable as string concatenation instead of interpolation? Unsure.

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+
+          # EOF
+          # EOFF
+          # EOFFF
+          # ...
+          buildSurround =
+            depth:
+            let
+              fs = lib.strings.replicate (depth + 1) "F";
+              token = "EO${fs}";
```
MattSturgeon, line 286:
```suggestion
              token = "EOF" + lib.strings.replicate depth "F";
```

Or

```suggestion
              token = "EOF${lib.strings.replicate depth "F"}";
```

(I prefer the first option)

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+            in
+            {
+              open = token;
```
MattSturgeon, line 289: Would including the `<< ` here make the usage below simpler? There'd no longer be a need to do `"<< ${escapedString}"`...


```suggestion
              open = "<< ${token}";
```

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+              close = token;
+              inherit depth;
+            };
+        };
+      in
+      "<< ${escapedString}"
```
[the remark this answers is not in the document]
MattSturgeon: Or handle it by having `buildSurround` inject the newlines into its open/close tokens

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
+          buildSurround =
+            depth:
+            let
+              fs = lib.strings.replicate (depth + 1) "F";
+              token = "EO${fs}";
+            in
+            {
+              open = token;
+              close = token;
+              inherit depth;
```
MattSturgeon, line 291:
```suggestion
```

I don't think this is used?

tests/lib-tests.nix
```diff
@@ -374,6 +374,28 @@ let
...
+    };
+
+    testEscapeStringForVimscript = {
+      expr = {
+        simple = helpers.utils.escapeStringForVimscript "simple";
+        depth-one = helpers.utils.escapeStringForVimscript "EOF";
+      };
+      expected = {
+        simple = "<< EOF\nsimple\nEOF";
+        depth-one = "<< EOFF\nEOF\nEOFF";
```
MattSturgeon, line 396: We can use "indented strings" to make this more legible:

```suggestion
        simple = ''
          << EOF
          simple
          EOF'';
        depth-one = ''
          << EOFF
          EOF
          EOFF'';
```

Note: while it's nice to have the closing `''` on its own line, this will add a trailing `\n` to the string.

tests/lib-tests.nix
```diff
@@ -374,6 +374,28 @@ let
...
+        simple = "[[simple]]";
+        depth-one = "[=[]]]=]";
+      };
+    };
+
+    testEscapeStringForVimscript = {
+      expr = {
+        simple = helpers.utils.escapeStringForVimscript "simple";
+        depth-one = helpers.utils.escapeStringForVimscript "EOF";
+      };
```
MattSturgeon, line 393:
```suggestion
      expr = lib.mapAttrs (_: helpers.utils.escapeStringForVimscript) {
        simple = "simple";
        depth-one = "EOF";
        depth-two = "EOF EOFF";
      };
```

We can use `mapAttrs` to generalize the test-case.

tests/lib-tests.nix
```diff
@@ -374,6 +374,28 @@ let
         ];
       };
     };
+
+    testEscapeStringForLua = {
+      expr = {
+        simple = helpers.utils.escapeStringForLua "simple";
+        depth-one = helpers.utils.escapeStringForLua "]]";
+      };
```
MattSturgeon, line 382:
```suggestion
      expr = lib.mapAttrs (_: helpers.utils.escapeStringForLua) {
        simple = "simple";
        depth-one = "]]";
        depth-two = "]] ]=]";
      };
```

Same

MattSturgeon (commented):
Looks good overall.

I think most of the simplification I originally had in mind would make this less "general purpose".

At this point my main gripe is the naming, since I don't think "escaping" accurately describes what we're doing nor does it imply the intended use case.

I've left a few questions, suggestions and discussion points below, but nothing I feel too strongly about.

lib/utils.nix
jfly: Good catch. I think that was leftover from a WIP implementation that you never saw.

lib/utils.nix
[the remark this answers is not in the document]
jfly: `surroundWithNewlines` is now gone (thanks to your other comments).

Please take another look.

lib/utils.nix
```diff
@@ -125,27 +125,185 @@ rec {
...
   # but only if the string contains something other than whitespaces
-  # TODO: account for a possible ']]' in the string
   wrapVimscriptForLua =
-    string:
-    lib.optionalString (hasContent string) ''
-      vim.cmd([[
-      ${string}
-      ]])
-    '';
+    string: lib.optionalString (hasContent string) "vim.cmd(${escapeStringForLua string})";
```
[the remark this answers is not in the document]
jfly: Maybe! Your call. Hopefully none of your users are using these functions directly (I am not).

Since it affects the public API of nixvim, I'd vote for doing that in a separate PR at least, since this has turned into quite a bit of code.

jfly (commented):
Thanks for the excellent review @MattSturgeon!

I've added another commit with the changes, so it's easier for your to see what's changed since your review. Happy to squash that commit now or later. Just tell me what's best.

I'll now go and resolve the threads I consider resolved. Feel free to reopen them if you disagree.

lib/utils.nix
jfly: I'm just realizing I invented the noun "Surround" and the verb "surround" for this PR when there was already the word "wrap" in this repo (and I could have invented the noun "Wrap"). I'm not sure there's much value in having both terms floating around.

Let me know what you think is best, @MattSturgeon

lib/utils.nix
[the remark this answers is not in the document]
MattSturgeon: Sure, we can tackle that another time. Historically we haven't been _too_ concerned about deprecating or removing things from the public API in cases where it seems unlikely that anyone is using them. But you are correct we ought to be more careful really.

As for `surroundFor` I'm happy either way. Keeping it out of the public API is probably better. I'm also curious about what the implementation would look like if we inlined it into both `buildSurround` implementations. I.e. would the potential simplification outweigh the code duplication. :thinking:

MattSturgeon:
Regarding https://github.com/nix-community/nixvim/pull/2272#discussion_r1767578063:

I was thinking we could avoid needing a `surroundFor` by doing something like this in each wrapper:

```nix
buildSurround =
  depth:
  let
    token = "EOF" + lib.strings.replicate depth "F";
  in
    if lib.hasInfix token string then
      buildSurround (depth + 1)
    else
      {
        open = "<< ${token}";
        close = token;
      };

surround = buildSurround 0;
```

lib/utils.nix
MattSturgeon: Hm. Naming is always a challenge!

Personally I like the names that are clear about the end result, so `toLuaLongLiteral` and `toVimscriptHeredoc` would be my current preference. Good suggestions!

We could be even more explicit and replace `to` with `wrapAs`, but I think that's unnecessary.

MattSturgeon:
> Regarding [#2272 (comment)](https://github.com/nix-community/nixvim/pull/2272#discussion_r1767578063):
> 
> I was thinking we could avoid needing a `surroundFor` by doing something like this in each wrapper:
> 
> Do you think having this small degree of redundancy (using `hasInfix` in two places) would be worth it to remove `surroundFor`?

Thinking about it, doing it this way would allow slightly more bespoke `buildSurround` implementations, which may allow further simplification.

E.g.
```nix
string:
let
  buildToken =
    depth:
    let
      token = "EOF" + lib.strings.replicate depth "F";
    in
      if lib.hasInfix token string then
        buildSurround (depth + 1)
      else
        token;

  token = buildToken 0;
in
''
  << ${token}
  ${string}
  {$token}'';
```

or
```nix
string:
let
  buildToken =
    depth:
    let
      infix = lib.strings.replicate depth "=";
    in
      if lib.hasInfix "]${infix}]" string then
        buildSurround (depth + 1)
      else
        infix;

  token = buildToken 0;
in
"[${infix}[${string}]${infix}]";
```

lib/utils.nix
```diff
@@ -125,27 +125,160 @@ rec {
...
+        inherit string;
+
+        # EOF
+        # EOFF
+        # EOFFF
+        # ...
+        buildSurround =
+          depth:
+          let
+            token = "EO" + lib.strings.replicate (depth + 1) "F";
```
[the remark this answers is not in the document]
jfly: Very much agreed!

lib/utils.nix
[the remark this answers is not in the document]
jfly: > I'm also curious about what the implementation would look like if we inlined it into both buildSurround implementations. I.e. would the potential simplification outweigh the code duplication. 🤔

Good question. The fact that it has such a weird name is perhaps a smell that it shouldn't really exist. That said, I can't really bring myself to duplicate it (perhaps because I sunk all that time into writing a docstring for it :rofl:).

I've floated `surroundFor` up to the top, but I'm leaving this conversation unresolved. I defer to you as maintainer on whatever you think is best.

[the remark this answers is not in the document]
jfly: Oh, nevermind! I'm just seeing your latest two comments (https://github.com/nix-community/nixvim/pull/2272#issuecomment-2362184330 and https://github.com/nix-community/nixvim/pull/2272#issuecomment-2362218500), and I'm inclined to agree with you.

Updating the PR. Standby...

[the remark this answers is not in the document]
jfly: Updated. I agree that inlining the implementation is better. Love the result, thanks!

Marking this thread resolved.

lib/utils.nix
```diff
@@ -125,27 +125,93 @@ rec {
...
+      buildInfix =
+        depth:
+        let
+          infix = lib.strings.replicate depth "=";
+        in
+        if lib.hasInfix "]${infix}]" string then buildInfix (depth + 1) else infix;
+
+      infix = buildInfix 0;
+    in
+    "[${infix}[${string}]${infix}]";
```
jfly, line 163: @MattSturgeon, .ooking at this one more time: one unfortunate fact of this implementation is that we've repeated the "how to build a close token" logic in 2 places, and it's really important that they stay in sink.

I'm going to rework this a little bit. Standby...

MattSturgeon: It is a little ugly, but I don't think it is a major issue.

If you don't like it, you could use the previous approach of returning a `{open,close}` attrset. Or you could do something like `builtins.replaceStrings ["]"] ["["] closingToken`, which is also ugly, but less redundant.

jfly: Updated. @MattSturgeon, please take another look.

lib/utils.nix
```diff
@@ -125,27 +125,95 @@ rec {
...
+          tokens.close = "]${infix}]";
+        in
+        if lib.hasInfix tokens.close string then buildTokens (depth + 1) else tokens;
+
+      tokens = buildTokens 0;
+    in
+    tokens.open + string + tokens.close;
+
+  /**
+    Conver the given String into a Vimscript [:let-heredoc], for use by [:lua].
```
MattSturgeon, line 168:
```suggestion
    Convert the given string into a Vimscript [:let-heredoc].

    For example, this could be used to pass a lua string to the [:lua] command.
```

1. Typo `conver` -> `convert`
2. I don't _think_ "String" needs to be capitalized?
3. `:lua` isn't the only use case, so we should re-phrase it as a "for instance".

lib/utils.nix
```diff
@@ -125,27 +125,95 @@ rec {
     end
   '';
 
+  /**
+    Convert the given String to a Lua [long literal].
```
MattSturgeon, line 129: By all means this is not a requirement, but if you wanted you could mention a potential use-case being to pass a vimscript script to the `vim.cmd()` function? Similar to the note in `toVimscriptHeredoc`'s docstring.

You're welcome to dismiss this feedback without any action.

> For example, you could use this to safely pass a vimscript string to the `vim.cmd` function.

MattSturgeon (commented):
> That said, I can't really bring myself to duplicate it (perhaps because I sunk all that time into writing a docstring for it 🤣).

Apologies for that! Hopefully the end-result being simpler is worth it for your wasted effort?

Everything looks good to me now. Just a typo and a minor tweak in the doccomments, squash and then merge.

Thanks for working on this and collaborating enthusiastically in the review process!

lib/utils.nix
```diff
@@ -151,16 +151,18 @@ rec {
   toLuaLongLiteral =
     string:
     let
-      buildInfix =
+      buildTokens =
```
MattSturgeon, line 154: We could debate the naming here, I wonder if `findTokens` and `findInfix` make more sense, since we are essentially _searching for_ a closing token that doesn't conflict with the string?

Not a big deal and doesn't blosk merging.

lib/utils.nix
jfly: 1. Typo fixed
2. Ah, thanks for pointing that out. This was intentional, to try evoke [this type hint I saw higher up in the file](https://github.com/jfly/nixvim/blob/40df98e4691f36cbc0a178aed20b10895183d22a/lib/utils.nix#L55). But perhaps it just looks weird, and isn't The Nix Way. This PR is probably the first time I've ever written a nix docstring.
3. Done.

lib/utils.nix
jfly: I like it!

lib/utils.nix
jfly: I agree, I prefer "find" over "build".

I don't think we have a function that should be called `findInfix`, though. We now have a `findToken` (for vimscript) and `findTokens` (for lua).

jfly:
> Everything looks good to me now. Just a typo and a minor tweak in the doccomments, squash and then merge.

Done and done.

> Thanks for working on this and collaborating enthusiastically in the review process!

Thank you! Contributing to this project ranks among the smoothest experiences I've had:

- Excellent `CONTRIBUTING.md`.
- A very useful shortcut documented at the top of `tests/lib-tests.nix`.
- A quick ACK of my PR to let me know it's on your radar.
- Excellent PR feedback with good ideas in anon-judgmental tone.

lib/utils.nix
MattSturgeon: Yeah it seems types used in the type-signature are usually capitalised. Outside that, in the prose of the docstring, I think we can be less formal.

But I'm fine either way if it was intended. :+1:

MattSturgeon:
@mergifyio queue

Issue #2273, opened 2024-09-19
alpha: allow for custom Lua code for sections.val


Mikilio:
https://github.com/nix-community/nixvim/blob/9307b201a3dc57d5b71ded4f897ea9d096544877/plugins/by-name/alpha/default.nix#L31

It makes only sense to allow for raw lua code to provide strings here, as it is a pain to define nix functions for it.

khaneliman:
This whole plugin module needs to be rewritten but, if you are in a need for that rawlua support immediately, I can PR a fix for it.

Mikilio:
It's not important since the entire plugin is just eye candy, but I appreciate the timely response. I actually think it's incredible.

Issue #1382, opened 2024-04-06
[BUG] Test config for Alpha using terminal is throwing error


elythh:
| Field        | Description |
|--------------|-------------|
| Plugin       | `alpha.nvim`  |

- [X] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

Nixvim is throwing error when using the test Alpha config using terminal type

## Minimal, Reproducible Example (MRE)

```nix
    plugins.alpha = {
      enable = true;
      layout = [
        {
          type = "terminal";
          command = "thisisfine";
          width = 46;
          height = 25;
          opts = {
            redraw = true;
          };
        }
      ];
    };
```

result
```
Error detected while processing VimEnter Autocommands for "*":
Error executing lua callback: ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:389: at
tempt to call a nil value
stack traceback:
        ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:389: in function 'layout'
        ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:629: in function 'draw'
        ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:723: in function 'start'
        ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:786: in function <...dir/pack
/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:784>
Press ENTER or type command to continue
```

elythh:
Thank you !
![image](https://github.com/user-attachments/assets/cf8a70c7-45ca-4a9f-bf73-8f3cba986d2b)

PR #2276, opened 2024-09-19
plugins/alpha: rawLua support; remove helpers and with lib; fix terminal type

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2273
Resolves https://github.com/nix-community/nixvim/issues/1382

Documentation states that we need to call require('alpha.term') to support a terminal type. Tested without using a terminal type and didn't see any negative side effects to just including it without having to parse through a user's layout configuration. But, in the rewrite it might be nice to properly check if it's even needed.

### DIFF
--- a/plugins/by-name/alpha/default.nix
+++ b/plugins/by-name/alpha/default.nix
@@ -1,13 +1,13 @@
 {
   lib,
-  helpers,
   config,
   options,
   pkgs,
   ...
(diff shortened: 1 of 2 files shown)


traxys (approved):

khaneliman:
@mergifyio queue

PR #2281, opened 2024-09-20
treewide: avoid passing `pkgs` to our lib

MattSturgeon:
In the two places where our "helpers" lib is exclusively internal (flake module args and building the docs), we no longer supply `pkgs`.

In the other 4 locations, we now note why we still provide a `pkgs` in a comment.

Additionally, this refactors the work done in #2231 by _always_ including the deprecated functions.

When `pkgs` is provided they are wrapped in `lib.warn`; when `pkgs` is missing, they are stubs that `throw` a deprecation message instead.

## Future work

For the `nixpkgs` module, I'm thinking we'll need to supply either a `pkgsPath` (`inputs.nixpkgs`) or our flake `inputs`. This could be optional, but `evalNixvim` may not work as intended when it isn't provided.

If we ever get to the point where the "standalone" wrapper is fully part of our lib, then we may also need to provide the flake outputs (`self`), or at least provide access to the `man-docs` package somehow. Alternatively, the modules could build their own copy of the man-docs.

After 24.11 is stabilised, I plan to drop `pkgs` from our lib, so any uses that currently have a note will also be able to get helpers without first needing a pkgs instance.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -33,12 +33,11 @@ let
       ) opt.declarations;
     };
 
-  evaledModules = lib.evalModules {
-    specialArgs = helpers.modules.specialArgsWith {
+  evaledModules = helpers.modules.evalNixvim {
+    extraSpecialArgs = {
       defaultPkgs = pkgs;
     };
     modules = [
-      ../modules/top-level
       { isDocs = true; }
     ];
   };
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -7,7 +7,7 @@ let
   # We overlay a few tweaks into pkgs, for use in the docs
   pkgs = import ./pkgs.nix { inherit system nixpkgs; };
   inherit (pkgs) lib;
-  helpers = import ../lib { inherit lib pkgs; };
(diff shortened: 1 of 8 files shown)


traxys:
I don't remember exactly, do we have warnings when the functions depending on pkgs are used?

MattSturgeon:
Do you think they should have removal stubs when `pkgs` isn't provided? E.g. `mapAttrs (name: _: throw "${name} has been removed. Use ${name}With instead.")`

MattSturgeon:
I've pushed a commit up that makes it so the deprecated functions are always present in nixvim's lib. Now instead of being missing, they throw a deprecation error when a `pkgs` instance is not provided.

If a `pkgs` instance _is_ provided, they print a deprecation warning, as before.

This means the deprecation message implementation is a little messier, however the overall nixvim-lib builder is actually simpler. We're now able to drop `lib.recursiveUpdateUntil` from the implementation for instance.

khaneliman (approved):
LGTM, don't see anything crazy

MattSturgeon:
@mergifyio queue

Issue #2119, opened 2024-08-30
Standardize Shared Packages/Plugins


khaneliman:
In response to https://github.com/nix-community/nixvim/pull/2112#issuecomment-2320384864 it would be good to consolidate our usages for certain `extraPlugins` and `extraPackages` that will affect other plugins. Currently, we do a lot of stuff with `git` and `nvim-web-devicons`, but there are other packages/plugins that are added as optional dependencies that have side effects on all / other plugins. 

I'm not sure where this option should live and what it should look like, yet.

MattSturgeon:
Perhaps we could do `cfg.package.override (prev: { dependencies = prev.dependencies ++ [ cfg.fooPackage ]; })`?

This would keep things a bit more isolated...

khaneliman:
I don't think that will actually isolate the plugin for an individual plugin. The inclusion of the dependency for lualine upstream caused side effects with other plugins because it just gets installed along with it.

stasjok:
Actually in my comment I meant something like a global option `iconsProvider`. And the user can change it to nvim-web-devicons or mini.icons with the ability to change settings for those providers. Both have it's own `setup()` function. To emulate nvim-web-devicons api MiniIcons provides `MiniIcons.mock_nvim_web_devicons()` function.

But maybe they both should be just plugins with it's own enable options. Mini.nvim already have a plugin definition, but there is no option to mock nvim-web-devicons (`MiniIcons.mock_nvim_web_devicons()` should probably be called in extraConfigLuaPre before other plugins to emulate nvim-web-devicons (if enabled)). For now if I want to use mini.icons, I need to set iconsPackage to null for every plugin explicitly. And I need to call `MiniIcons.mock_nvim_web_devicons()` somewhere (ideally there should be an option for this).

Also maybe a known icon provider plugins can change some internal value when enabled. In other plugins instead of iconsPackage the value of this option can be checked. If some icon plugin is already enabled, then do nothing. If nothing is enabled, then enable some default implementation (nvim-web-devicons).

MattSturgeon:
Looking at them as plugins in their own right, I think we should handle them as `plugins.*.enable`.

Calling `MiniIcons.mock_nvim_web_devicons` could be an extra-option under `plugins.mini`; e.g. `plugins.mini.mockDevIconsAPI` or something. IDK if it should be on or off by default.

[...]

PR #2283, opened 2024-09-20
treewide: iconsPackage -> plugins enabled

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2119

Just wanted to double check if we wanted to do something like this that we did for sqlite-lua transition before I implement this in all the packages that I added iconsPackage to in https://github.com/nix-community/nixvim/pull/2112

With this approach, we will make it easier for a user to toggle between `nvim-web-devicons` and `mini.icons` since they can explicitly enable/disable either plugin. The default is still to use `nvim-web-devicons` to be a transition period of the old behavior. A user can explicitly disable the plugin and enable mini.icons and it will switch to using that instead.

### DIFF
--- a/plugins/by-name/alpha/default.nix
+++ b/plugins/by-name/alpha/default.nix
@@ -51,20 +51,6 @@ let
   };
 in
 {
-  # TODO: added 2024-09-20 remove after 24.11
-  imports = [
-    (lib.mkRemovedOptionModule
-      [
-        "plugins"
-        "alpha"
-        "iconsPackage"
-      ]
-      ''
-        Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
-      ''
-    )
-  ];
-
   options = {
     plugins.alpha = {
       enable = lib.mkEnableOption "alpha-nvim";
@@ -167,23 +153,11 @@ in
     lib.mkIf cfg.enable {
 
       # TODO: added 2024-09-20 remove after 24.11
-      warnings = lib.optionals opt.iconsEnabled.isDefined (
-        [
-          ''
-            The option definition `plugins.alpha.iconsEnabled' in ${lib.showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
-          ''
-        ]
-        ++
-          lib.optional
-            (
-              (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
-              && options.plugins.web-devicons.enable.highestPrio == 1490
-            )
-            ''
-              Nixvim (plugins.alpha) `web-devicons` automatic installation is deprecated.
-              Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
-            ''
-      );
+      warnings = lib.optionals opt.iconsEnabled.isDefined [
+        ''
+          The option definition `plugins.alpha.iconsEnabled' in ${lib.showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+        ''
+      ];
       plugins.web-devicons =
         lib.mkIf
           (
--- a/plugins/by-name/alpha/default.nix
+++ b/plugins/by-name/alpha/default.nix
@@ -51,6 +51,20 @@ let
   };
 in
 {
+  # TODO: added 2024-09-20 remove after 24.11
+  imports = [
+    (lib.mkRemovedOptionModule
+      [
+        "plugins"
+        "alpha"
+        "iconsPackage"
+      ]
+      ''
+        Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
+      ''
+    )
+  ];
+
   options = {
     plugins.alpha = {
       enable = lib.mkEnableOption "alpha-nvim";
@@ -68,11 +82,6 @@ in
         description = "Toggle icon support. Installs nvim-web-devicons.";
         visible = false;
       };
-
-      iconsPackage = lib.mkPackageOption pkgs [
-        "vimPlugins"
-        "nvim-web-devicons"
-      ] { nullable = true; };
 
       theme = mkOption {
         type = with types; nullOr (maybeRaw str);
@@ -156,20 +165,41 @@ in
       opt = options.plugins.alpha;
     in
     lib.mkIf cfg.enable {
-      # TODO: deprecated 2024-08-29 remove after 24.11
-      warnings = lib.mkIf opt.iconsEnabled.isDefined [
-        ''
-          nixvim (plugins.alpha):
-          The option definition `plugins.alpha.iconsEnabled' in ${lib.showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
-          You should use `plugins.alpha.iconsPackage' instead.
-        ''
-      ];
 
-      extraPlugins =
-        [ cfg.package ]
-        ++ lib.optional (
-          cfg.iconsPackage != null && (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
-        ) cfg.iconsPackage;
+      # TODO: added 2024-09-20 remove after 24.11
+      warnings = lib.optionals opt.iconsEnabled.isDefined (
+        [
+          ''
+            The option definition `plugins.alpha.iconsEnabled' in ${lib.showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+          ''
+        ]
+        ++
+          lib.optional
+            (
+              (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
+              && options.plugins.web-devicons.enable.highestPrio == 1490
+            )
+            ''
+              Nixvim (plugins.alpha) `web-devicons` automatic installation is deprecated.
+              Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
+            ''
+      );
+      plugins.web-devicons =
+        lib.mkIf
+          (
+            opt.iconsEnabled.isDefined
+            && cfg.iconsEnabled
+            && !(
+              config.plugins.mini.enable
+              && config.plugins.mini.modules ? icons
+              && config.plugins.mini.mockDevIcons
+            )
+          )
+          {
+            enable = lib.mkOverride 1490 true;
+          };
+
+      extraPlugins = [ cfg.package ];
 
       assertions = [
         {
--- a/plugins/by-name/barbar/default.nix
+++ b/plugins/by-name/barbar/default.nix
@@ -1,6 +1,7 @@
 {
+  config,
   lib,
-  pkgs,
+  options,
   ...
 }:
 with lib;
@@ -172,6 +173,17 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
           ]
         )
       )
+      # TODO: added 2024-09-20 remove after 24.11
+      (lib.mkRemovedOptionModule
+        [
+          "plugins"
+          "barbar"
+          "iconsPackage"
+        ]
+        ''
+          Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
+        ''
+      )
     ]
     ++ (map
       (
@@ -195,11 +207,6 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
     );
 
   extraOptions = {
-    iconsPackage = lib.mkPackageOption pkgs [
-      "vimPlugins"
-      "nvim-web-devicons"
-    ] { nullable = true; };
-
     keymaps = mapAttrs (
       optionName: funcName:
       mkNullOrOption' {
@@ -217,7 +224,18 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   };
 
   extraConfig = cfg: {
-    extraPlugins = mkIf (cfg.iconsPackage != null) [ cfg.iconsPackage ];
+    # TODO: added 2024-09-20 remove after 24.11
+    plugins.web-devicons = mkIf (
+      !(
+        config.plugins.mini.enable
+        && config.plugins.mini.modules ? icons
+        && config.plugins.mini.mockDevIcons
+      )
+    ) { enable = mkOverride 1490 true; };
+    warnings = optional (options.plugins.web-devicons.enable.highestPrio == 1490) ''
+      Nixvim (plugins.barbar) `web-devicons` automatic installation is deprecated.
+      Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
+    '';
 
     keymaps = filter (keymap: keymap != null) (
       # TODO: switch to `attrValues cfg.keymaps` when removing the deprecation warnings above:
--- a/plugins/by-name/barbar/default.nix
+++ b/plugins/by-name/barbar/default.nix
@@ -173,17 +173,6 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
           ]
         )
       )
-      # TODO: added 2024-09-20 remove after 24.11
-      (lib.mkRemovedOptionModule
-        [
-          "plugins"
-          "barbar"
-          "iconsPackage"
-        ]
-        ''
-          Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
-        ''
-      )
     ]
     ++ (map
       (
@@ -232,10 +221,6 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
         && config.plugins.mini.mockDevIcons
       )
     ) { enable = mkOverride 1490 true; };
-    warnings = optional (options.plugins.web-devicons.enable.highestPrio == 1490) ''
-      Nixvim (plugins.barbar) `web-devicons` automatic installation is deprecated.
-      Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
-    '';
 
     keymaps = filter (keymap: keymap != null) (
       # TODO: switch to `attrValues cfg.keymaps` when removing the deprecation warnings above:
--- a/plugins/by-name/chadtree/default.nix
+++ b/plugins/by-name/chadtree/default.nix
@@ -12,20 +12,6 @@ let
   mkListStr = helpers.defaultNullOpts.mkNullable (types.listOf types.str);
 in
 {
-  # TODO: added 2024-09-20 remove after 24.11
-  imports = [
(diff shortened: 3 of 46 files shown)


MattSturgeon:
cc @stasjok: do you have an opinion?

plugins/by-name/chadtree/default.nix
```diff
@@ -502,11 +498,21 @@ in
...
+        optional
+          (
+            (cfg.theme == null || cfg.theme.iconGlyphSet == "devicons")
+
+            && options.plugins.web-devicons.enable.highestPrio == 1490
+          )
+          ''
+            Nixvim (plugins.chadtree) `web-devicons` automatic installation is deprecated.
+            Please use `plugins.web-devicons.enable`.
+          '';
```
MattSturgeon, line 513: Should this warning give any hints for how to use `mini.icons` instead of `web-devicons`? Or if we had an assertion, perhaps those kinda notes could go there?

Not that I'm implying I do or don't want an assertion...

plugins/by-name/chadtree/default.nix
```diff
@@ -502,11 +498,21 @@ in
...
-        [ cfg.package ]
-        ++ (optional (
-          cfg.iconsPackage != null && (cfg.theme == null || cfg.theme.iconGlyphSet == "devicons")
-        ) cfg.iconsPackage);
+      # TODO: added 2024-09-20 remove after 24.11
+      plugins.web-devicons.enable = mkOverride 1490 true;
+      warnings =
+        optional
+          (
+            (cfg.theme == null || cfg.theme.iconGlyphSet == "devicons")
```
MattSturgeon, line 506: If this is true when `cfg.theme == null`, perhaps it should also be true when `cfg.theme.iconGlyphSet == null`?

plugins/by-name/barbar/default.nix
```diff
@@ -195,11 +195,6 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
     );
 
   extraOptions = {
-    iconsPackage = lib.mkPackageOption pkgs [
-      "vimPlugins"
-      "nvim-web-devicons"
-    ] { nullable = true; };
```
MattSturgeon, line 201: Should we have removed-option-modules for these? We've only had the options for 3 weeks, but they are pretty user-facing :thinking: 

I'm not yet voicing an opinion, just raising the issue for discussion.

khaneliman: That's what I was wondering, too...

plugins/by-name/chadtree/default.nix
MattSturgeon: E.g. `elem cfg.theme.iconGlyphSet or null [ null "devisons" ]`

plugins/by-name/barbar/default.nix
khaneliman: I might just do it to be safe... I know we said we didn't need to in the linked issue, but it took me longer to get to this than I expected.

plugins/by-name/chadtree/default.nix
khaneliman: If I do the mkRemovedOptionModule then it will contain those instructions.

khaneliman: > Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.

plugins/by-name/chadtree/default.nix
khaneliman: I was just wanting to show the warning for the condition that used to auto install the package.

plugins/by-name/alpha/default.nix
```diff
@@ -51,6 +51,21 @@ let
...
+  # TODO: added 2024-09-20 remove after 24.11
+  imports = [
+    (lib.mkRemovedOptionModule
+      [
+        "plugins"
+        "alpha"
+        "iconsPackage"
+      ]
+      ''
+        The `iconsPackage` option has been removed.
```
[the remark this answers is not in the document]
khaneliman: Ah, yeah. I'll clean it up after I get the build errors figured out.

plugins/by-name/fzf-lua/default.nix
```diff
@@ -112,17 +121,26 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
-          The option definition `plugins.fzf-lua.iconsEnabled' in ${showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
-          You should use `plugins.fzf-lua.iconsPackage' instead.
-        ''
-      ];
+      warnings =
+        lib.optional opt.iconsEnabled.isDefined [
+          ''
+            nixvim (plugins.fzf-lua):
+            The option definition `plugins.fzf-lua.iconsEnabled' in ${showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+            You should use `plugins.fzf-lua.iconsPackage' instead.
```
stasjok, line 129: Probably should update this message.

stasjok:
> cc @stasjok: do you have an opinion?

Don't have any objections to the concept of this PR.

khaneliman:
image

Deprecation messages working with local config and with disabling it, nothing.

plugins/by-name/alpha/default.nix
```diff
@@ -156,20 +165,31 @@ in
...
+            The option definition `plugins.alpha.iconsEnabled' in ${lib.showFiles opt.iconsEnabled.files} has been deprecated; please remove it.
+            Please use `plugins.web-devicons` or `plugins.mini.modules.icons` with `plugins.mini.mockDevIcons` instead.
+          ''
+        ]
+        # TODO: added 2024-09-20 remove after 24.11
+        ++
+          lib.optional
+            (
+              (opt.iconsEnabled.isDefined -> cfg.consEnabled)
+              && options.plugins.web-devicons.enable.highestPrio == 1490
```
MattSturgeon, line 183: This doesn't look right.

1. `cfg.consEnabled` looks like a typo
2. The `&&` would mean `plugins.web-devicons.enable`'s priority is only checked when the lhs is true, but we are enabling it below unconditionally

Perhaps this should be two separate warnings? Or an `||` condition?

plugins/by-name/chadtree/default.nix
```diff
@@ -502,11 +512,21 @@ in
...
+
+            && options.plugins.web-devicons.enable.highestPrio == 1490
```
MattSturgeon, line 522: Again, this warning has an `&&` check, but we define the enable option unconditionally above. Am I missing something?

plugins/by-name/fzf-lua/default.nix
```diff
@@ -112,17 +121,26 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+          ''
+        ]
+        ++ lib.optional
+          (
+            (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
 
-      extraPlugins = lib.mkIf (
-        cfg.iconsPackage != null && (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
-      ) [ cfg.iconsPackage ];
+            && options.plugins.web-devicons.enable.highestPrio == 1490
```
MattSturgeon, line 136: Same

plugins/by-name/lspsaga/default.nix
```diff
@@ -461,18 +470,25 @@ in
...
-      (isBool cfg.implement.enable && cfg.implement.enable)
-      && (isBool cfg.symbolInWinbar.enable && !cfg.symbolInWinbar.enable)
-    ) [ "You have enabled the `implement` module but it requires `symbolInWinbar` to be enabled." ];
+    # TODO: added 2024-09-20 remove after 24.11
+    plugins.web-devicons.enable = mkOverride 1490 true;
+    warnings =
+      lib.optional
+        (
+          (cfg.ui.devicon == null || cfg.ui.devicon)
+          && options.plugins.web-devicons.enable.highestPrio == 1490
```
MattSturgeon, line 479: Same

plugins/by-name/alpha/default.nix
khaneliman: Nah, it should be this condition, but I need to fix the enable below.

MattSturgeon: Perhaps we can nest both the warning and the enable under a single `mkIf` then?

We could always use `mkMerge` if there's other `warnings` defined separately?

khaneliman: I didn't realize I had basically duplicate warnings... I was fixing the triggering of the warnings after your comment about the unconditional enabling.

khaneliman: I pushed an update that is fixing when it enables the plugin and consolidates the warning.

MattSturgeon: I almost pointed out the other warning checking `isDefined` but figured we'd focus on one thing at a time 😁

plugins/by-name/alpha/default.nix
```diff
@@ -156,20 +165,24 @@ in
       opt = options.plugins.alpha;
     in
     lib.mkIf cfg.enable {
-      # TODO: deprecated 2024-08-29 remove after 24.11
-      warnings = lib.mkIf opt.iconsEnabled.isDefined [
```
MattSturgeon, line 160: By dropping this we no longer warn when someone defines `iconsEnabled = false`.

If the option truly is deprecated, then any definition should have a warning. It's fine to have different warnings based on additional things though, and/or to build the warning text dynamically based on additional factors.

khaneliman: Doh, you're right...

khaneliman: I was trying to consolidate the message down so there weren't so many warnings for what I thought was that same thing.

MattSturgeon: Yeah, ideally we'd only print one warning message. I don't mind if it's one "dumb" all encompassing messages, or if it takes additional conditions into account, but we should definitely always have _some_ warning when the option is defined.

plugins/by-name/alpha/default.nix
```diff
@@ -156,20 +165,27 @@ in
...
-        ) cfg.iconsPackage;
+      # TODO: added 2024-09-20 remove after 24.11
+      warnings =
+        lib.optional
+          (
+            opt.iconsEnabled.isDefined
+            || (
+              (opt.iconsEnabled.isDefined -> cfg.iconsEnabled)
+              && options.plugins.web-devicons.enable.highestPrio == 1490
+            )
```
MattSturgeon, line 177: This is logically redundant. Both the LHS and RHS of the `||` require `isDefined`. Therefore the RHS has no effect.

I guess what we want is _either_:
- show one warning when the RHS is _not_ true and a different warning when it _is_ true
- have a single warning, but dynamically adapt the warning text depending on whether the RHS is true.

MattSturgeon: E.g.

```nix
lib.optional opt.iconsEnabled.isDefined 
   (
     if cfg.iconsEnabled && options.plugins.web-devicons.enable.highestPrio == 1490 then
      "warning 1"
    else
      "warning 2"
   )
```

stasjok:
Is it truly necessary to include verbose warnings in every plugin? Perhaps it would be easier to have a single warning in the `plugins.web-devicons` that is triggered by the condition `cfg.enable.highestPrio == 1490`. In that case, it means that some plugin has enabled it, and the user didn't override it with explicit `enable = true` or `enable = false`. Just ensure that `plugins.web-devicons.enable = mkOverride 1490 true;` in all plugins remains backward compatible with the previous behavior, and I believe that would enough.

MattSturgeon:
Maybe having this deprecation in one central place, and doing everything based on a list of plugins would make sense.

Could be something like `plugins/icons-deprecation.nix` or `plugins/by-name/web-devicons/deprecation.nix` for instance 🤔

khaneliman:
> Is it truly necessary to include verbose warnings in every plugin? Perhaps it would be easier to have a single warning in the `plugins.web-devicons` that is triggered by the condition `cfg.enable.highestPrio == 1490`. In that case, it means that some plugin has enabled it, and the user didn't override it explicitly with `enable = true` or `enable = false`. Just ensure that conditions for `plugins.web-devicons.enable = mkOverride 1490 true;` in all plugins remains backward compatible with the previous behavior, and I believe that would be enough.

That’s what I would have preferred, but in my original implementation for SQLite-lua I was told to do it in each package.

plugins/by-name/alpha/default.nix
khaneliman: I understand that they are separate, this approach was in response to :

>Yeah, ideally we'd only print one warning message. I don't mind if it's one "dumb" all encompassing messages, or if it takes additional conditions into account, but we should definitely always have some warning when the option is defined.

I was just making sure we print the same warning when the option is used, at all. I didn't care if the user had used it differently, I'm just letting them know we don't have the option available anymore and we will print the warning telling them to use a new option.

khaneliman: If we're going back to wanting two separate warnings I would just revert it back to how I had it originally.

khaneliman:
I don't want to just across the board enable the plugin for everything in a list. Most of them were conditionally enabled. I could create a separate list in the deprecations module, if we really want that to house it for some reason, for the ones that enabled it unconditionally. Otherwise, I will keep the enabling of the plugin central to the modules to make that decision. The deprecation module would be specific for how to handle detecting a deprecated usage.

khaneliman:
@MattSturgeon anything else you think needs to be addressed?

plugins/by-name/trouble/default.nix
```diff
@@ -300,14 +300,8 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
-    iconsPackage = lib.mkPackageOption pkgs [
-      "vimPlugins"
-      "nvim-web-devicons"
-    ] { nullable = true; };
-  };
-
   extraConfig = cfg: {
-    extraPlugins = mkIf (cfg.iconsPackage != null) [ cfg.iconsPackage ];
+    # TODO: added 2024-09-20 remove after 24.11
+    plugins.web-devicons.enable = mkOverride 1490 true;
```
MattSturgeon, line 305: I think we should be doing this only if `mini.icons` isn't enabled?

```suggestion
    plugins.web-devicons.enable = mkIf (!(config.plugins.mini.modules.icons.enable && config.plugins.mini.mockDevIcons)) (mkOverride 1490 true);
```

Having this littered throughout the codebase rather than defined in a central place isn't ideal, but I'm fine with it as this is probably a deprecation that we _actually_ plan to remove after 24.11.

Given some plugins have additional requirements for `plugins.web-devicons` to be enabled, I'm also fine with doing it separately in each plugin.

plugins/deprecation.nix
```diff
@@ -36,5 +51,36 @@ in
...
+    ) iconsPackagePlugins
+    # Show a warning when web-devicons is auto-enabled
+    ++ [
+      (
+        { config, options, ... }:
+        {
+          config = lib.mkIf (options.plugins.web-devicons.enable.highestPrio == 1490) {
+            warnings = [
+              ''
+                Nixvim: `plugins.web-devicons` was enabled automatically because one of the following plugins was enabled.
```
MattSturgeon, line 75:
```suggestion
                Nixvim: `plugins.web-devicons` was enabled automatically because the following plugins are enabled.
```

(My mistake) because we're filtering the list below. Alternatively we could keep the "one of" wording and remove the filter.

MattSturgeon (commented):
Sorry it took me a while to get back to this.

Are you happy with the current mostly-centralised warnings approach? I got the impression from some prior comments you would rather take a different approach.

Other than some minor issues (below) everything looks good to me.

plugins/by-name/trouble/default.nix
khaneliman: Hmm, originally I was just doing it to replicate the original behavior of our auto install in each of the plugins. But, the intention of the change is truly to support alternative icon providers so I agree it does make sense to change the auto install to only be done when a user isn't expecting the new icon provider to take effect.

khaneliman:
> Are you happy with the current mostly-centralised warnings approach? I got the impression from some prior comments you would rather take a different approach.

Sorry if I came across wrong, I like the idea of centrally handling the deprecation stuff, but thought the central file would be polluted with checking a ton of configs if we also included the conditional enabling logic there for each plugin where it was only installed under specific circumstances.

khaneliman:
Alright, updated them all to only automatically enable when a user doesn't have mini.icons with mockDevIcons enabled.

plugins/by-name/barbar/default.nix
```diff
@@ -217,7 +212,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   };
 
   extraConfig = cfg: {
-    extraPlugins = mkIf (cfg.iconsPackage != null) [ cfg.iconsPackage ];
+    # TODO: added 2024-09-20 remove after 24.11
+    plugins.web-devicons.enable = mkIf (
+      !(config.plugins.mini.modules ? icons && config.plugins.mini.mockDevIcons)
```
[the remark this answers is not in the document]
khaneliman: Oh yeah... lol I forgot about enable too

MattSturgeon:
```
The option `plugins.nvim-tree.iconsPackage' in `/nix/store/XXX-source/plugins/by-name/nvim-tree' is already declared in `/nix/store/XXX-source/plugins/deprecation.nix'.
```

khaneliman:
> `The option `plugins.nvim-tree.iconsPackage' in `/nix/store/XXX-source/plugins/by-name/nvim-tree' is already declared in `/nix/store/XXX-source/plugins/deprecation.nix'.`
> 
> Looks like some plugins are still declaring/removing the `iconsPackage` option themselves even though `plugins/deprecation.nix` is handling that centrally.

Probably just a merge conflict resolution issue

plugins/by-name/alpha/default.nix
```diff
@@ -156,20 +151,29 @@ in
...
+            && cfg.iconsEnabled
+            && !(
+              config.plugins.mini.enable
+              && config.plugins.mini.modules ? icons
+              && config.plugins.mini.mockDevIcons
+            )
+          )
+          {
+            enable = lib.mkOverride 1490 true;
+          };
```
MattSturgeon, line 174: Is there any difference between doing

```nix
plugins.web-devicons = mkIf () { enable = mkOverride; };
```

vs
```nix
plugins.web-devicons.enable = mkIf () (mkOverride);
```

?

I have a feeling the former is better than the latter? No big deal, I just noticed that both techniques have been used in the PR.

khaneliman: Shouldn't be, if it was done it wasn't intentional though.

MattSturgeon (approved):
LGTM. Up to you if/how you want to deal with https://github.com/nix-community/nixvim/pull/2283#discussion_r1770611195

Apologies the review process hasn't been as smooth as we would like, but thanks for pushing through and getting the PR done!

khaneliman:
@mergifyio queue

Issue #1937, opened 2024-07-27
[BUG] LSP `volar` requires additional configuration


yzx9:
| Field        | Description |
|--------------|-------------|
| Plugin       | `lsp`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

The `@vue/language-server` (referred to as `volar` in the nixvim LSP plugin) introduces a new "Hybrid" mode starting from version ^2.0.0. This mode relies on `tsserver` with additional configuration.

According to the [official documentation](https://github.com/vuejs/language-tools?tab=readme-ov-file#hybrid-mode-configuration-requires-vuelanguage-server-version-200), the following configuration works within the `unstable` channel:

```nix
programs.nixvim = {
  plugins.lsp = {
    enable = true;
    servers = {
      # additional configuration for volar
      # https://github.com/vuejs/language-tools?tab=readme-ov-file#hybrid-mode-configuration-requires-vuelanguage-server-version-200
      tsserver = {
        enable = true;
        filetypes = [
          "typescript"
          "javascript"
          "javascriptreact"
          "typescriptreact"
          "vue"
        ];
        extraOptions = {
          init_options = {
            plugins = [
              {
                name = "@vue/typescript-plugin";
                location = "${lib.getBin pkgs.vue-language-server}/lib/node_modules/@vue/language-server";
                languages = [ "vue" ];
              }
            ];
          };
        };
      };
      volar = {
        enable = true;
        package = pkgs.vue-language-server;
      };
    };
  };
}
```

However, I am unsure how to add dependencies and configuration by default, which prevents me from contributing.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  plugins.lsp = {
    enable = true;
    servers = {
      volar.enable = true;
    };
  };
}
```

PR #2287, opened 2024-09-21
plugins/lsp/volar: add ts-ls integration

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/1937

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -691,6 +691,32 @@ let
         "nodePackages"
         "@volar/vue-language-server"
       ];
+      extraOptions = {
+        tslsIntegration = mkOption {
+          type = types.bool;
+          description = ''
+            Enable integration with TypeScript language server.
+          '';
+          default = true;
+          example = false;
+        };
+      };
+      extraConfig = cfg: {
+        plugins.lsp.servers.ts-ls = lib.mkIf (cfg.enable && cfg.tslsIntegration) {
+          filetypes = [ "vue" ];
+          extraOptions = {
+            init_options = {
+              plugins = [
+                {
+                  name = "@vue/typescript-plugin";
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -691,6 +691,32 @@ let
...
+          description = ''
+            Enable integration with TypeScript language server.
+          '';
+          default = true;
+          example = false;
+        };
+      };
+      extraConfig = cfg: {
+        plugins.lsp.servers.ts-ls = {
+          filetypes = [ "vue" ];
```
traxys, line 706: If we don't have a matching default for ts files this will disable them won't it?

khaneliman: This block was intended to just append to the user's existing ts-ls config when the integration is enabled.

khaneliman: Added a couple tests to show what I meant.

traxys: The issue is that `ts-ls.filetypes` is `null` by default. This means that enabling the `tslsIntegration` but leaving the `ts-ls.filetypes` as default will result on only  `{ "vue" }` being passed to nvim-lspconfig. I think we need to set the propagate the default value of  `{ "javascript", "javascriptreact", "javascript.jsx", "typescript", "typescriptreact", "typescript.tsx" }`

khaneliman: Oh, you mean the plugin's default has those values and we override the default plugin behavior by enabling it.

traxys: Exactly, I think the fix is to hardcode the plugin's default as the nix default

khaneliman: Sounds good, I'll update that when I get home later.

traxys (approved):

khaneliman:
@mergifyio queue

PR #2293, opened 2024-09-22
plugins/otter: Remove cmp

refaelsh:
Cmp integration is no longer needed for this plugin.
Here is a quote from their README:

> If you previously used the otter nvim-cmp source, you can remove it, as the completion results now come directly via the cmp-nvim-lsp source together with other language servers.

Fixes #2243

### DIFF
--- a/plugins/by-name/otter/default.nix
+++ b/plugins/by-name/otter/default.nix
@@ -19,9 +19,17 @@ helpers.neovim-plugin.mkNeovimPlugin {
         "otter"
         "addCmpSources"
       ]
-      "Otter is now supported by `plugins.cmp.autoEnableSources`, adding `otter` to `cmp` sources will enable this plugin."
+      ''
+        You should use the "cmp-nvim-lsp" source instead.
+        To quote upstream's README:
+        > If you previously used the otter nvim-cmp source, you can remove it, as the completion results now come directly via the cmp-nvim-lsp source together with other language servers.
+      ''
     )
+
     # Register nvim-cmp association
+    # TODO: Otter is no longer a cmp-source
+    # Deprecated 2024-09-22; remove after 24.11
+    # Note: a warning is implemented in plugins/cmp/auto-enable.nix
     { cmpSourcePlugins.otter = "otter"; }
   ];
 
--- a/plugins/cmp/auto-enable.nix
+++ b/plugins/cmp/auto-enable.nix
@@ -69,6 +69,11 @@ in
           If you want to keep using raw lua for defining your sources:
           - Ensure you enable the relevant plugins manually in your configuration;
           - Dismiss this warning by explicitly setting `autoEnableSources` to `false`;
+        ''
+        # TODO: Added 2024-09-22; remove after 24.11
+        ++ optional (elem "otter" enabledSources) ''
+          Nixvim (plugins.cmp): "otter" is listed in `settings.sources`, however it is no longer a cmp source.
+          Instead, you should enable `plugins.otter` and use the "cmp-nvim-lsp" completion source.
         '';
 
       # If the user has enabled the `foo` and `bar` sources, then `plugins` will look like:


traxys:
We should either mark the option as removed or deprecated, instead of outright deleting it

plugins/by-name/otter/default.nix
```diff
@@ -17,12 +17,8 @@ helpers.neovim-plugin.mkNeovimPlugin {
       [
         "plugins"
         "otter"
-        "addCmpSources"
```
khaneliman, line 20: You can't just remove the removeOptionModule setting. This would be saying that the plugin itself was removed. The `plugins.otter.addCmpSources` option WAS removed, so we can't remove this

refaelsh:
> We should either mark the option as removed or deprecated, instead of outright deleting it

Good point. I have no idea how to do it :-)

plugins/by-name/otter/default.nix
refaelsh: I do not understand.

traxys:
You can look at `mkRemovedOptionModule`, or maybe `mkDeprecatedSubOptionModule`

MattSturgeon:
> You can look at `mkRemovedOptionModule`, or maybe `mkDeprecatedSubOptionModule`

I think `mkDeprecatedSubOptionModule` may have been broken from the start. I've been meaning to look into it again.

plugins/by-name/otter/default.nix
MattSturgeon: The line you removed is part of a "removal module" for an option we previously removed. It's unrelated to the change you're trying to make.

Instead you just need to update the _text_ of the removal notice below.

plugins/by-name/otter/default.nix
```diff
@@ -17,12 +17,8 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
       ]
-      "Otter is now supported by `plugins.cmp.autoEnableSources`, adding `otter` to `cmp` sources will enable this plugin."
     )
-    # Register nvim-cmp association
-    { cmpSourcePlugins.otter = "otter"; }
```
MattSturgeon, line 25: I'm not sure the correct way to deprecate this. If a user is currently relying on `plugins.cmp.settings.sources` and `autoEnableSources`, then removing this association will mean otter is no longer enabled.

refaelsh: IMHO, if a user has `plugins.cmp.settings.sources` and `autoEnableSources` in his config, he should get a warning.

plugins/by-name/otter/default.nix
```diff
@@ -19,7 +19,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
         "otter"
         "addCmpSources"
       ]
-      "Otter is now supported by `plugins.cmp.autoEnableSources`, adding `otter` to `cmp` sources will enable this plugin."
+      "The `plugins.otter.addCmpSources` option is deprecated and will be rmoved. A quote from the README: \"If you previously used the otter nvim-cmp source, you can remove it, as the completion results now come directly via the cmp-nvim-lsp source together with other language servers.\""
```
MattSturgeon, line 22: It's inaccurate to say "will be removed". If we have a removal-module then the option _is_ already removed.

Also, note that the removal module itself already includes text to this effect:

https://github.com/NixOS/nixpkgs/blob/c976169e5c334ae42bdf239a868e6a218df63e67/lib/modules.nix#L1112-L1115

```
The option definition `${showOption optionName}' in ${showFiles opt.files} no longer has any effect; please remove it.
${replacementInstructions}
```

plugins/by-name/otter/default.nix
MattSturgeon: Yeah. I think this will require some manual handling in `plugins/cmp/auto-enable.nix`

plugins/by-name/otter/default.nix
refaelsh: I removed the `will be removed` words.

refaelsh: I am not really sure how I can improve it more.

plugins/by-name/otter/default.nix
```diff
@@ -19,7 +19,7 @@ helpers.neovim-plugin.mkNeovimPlugin {
         "otter"
         "addCmpSources"
       ]
-      "Otter is now supported by `plugins.cmp.autoEnableSources`, adding `otter` to `cmp` sources will enable this plugin."
+      "The `plugins.otter.addCmpSources` option is deprecated. A quote from the README: \"If you previously used the otter nvim-cmp source, you can remove it, as the completion results now come directly via the cmp-nvim-lsp source together with other language servers.\""
```
MattSturgeon, line 22:
```suggestion
      ''
        A quote from the README:
        > If you previously used the otter nvim-cmp source, you can remove it, as the completion results now come directly via the cmp-nvim-lsp source together with other language servers.
      ''
```

plugins/by-name/otter/default.nix
MattSturgeon: I've pushed up how I would suggest handling this. If you're happy you can squash this into your commit.

plugins/cmp/auto-enable.nix
```diff
@@ -69,6 +69,11 @@ in
           If you want to keep using raw lua for defining your sources:
           - Ensure you enable the relevant plugins manually in your configuration;
           - Dismiss this warning by explicitly setting `autoEnableSources` to `false`;
+        ''
+        # TODO: Added 2024-09-22; remove after 24.11
+        ++ optional (elem "otter" enabledSources) ''
+          Nixvim (plugins.cmp): "otter" is listed in `settings.sources`, however it is no longer a cmp source.
+          Instead, you should enable `plugins.otter` and use the "cmp-nvim-lsp" completion source.
```
MattSturgeon, line 76: Currently, this warning will show if `autoEnableSources` is enabled _and_ `otter` is listed in `settings.sources`.

Perhaps it would make sense for this to also be shown even when `autoEnableSources` isn't true? But that'd be a more complex change.

plugins/by-name/otter/default.nix
refaelsh: > I've pushed up how I would suggest handling this. If you're happy you can squash this into your commit.

Yes, I am very happy with your suggestions! Thank you :-)

plugins/cmp/auto-enable.nix
refaelsh: > Perhaps it would make sense for this to also be shown even when autoEnableSources isn't true?

IMHO, yes. But:
1. It requires more work - per your words.
2. IMHO, very few users have `autoEnableSources = false`.

Therefor, I think what you did is enough.

MattSturgeon (approved):
no issues from me. Can you rebase and squash as something like "plugins/otter: remove cmp source registration"?

khaneliman (approved):
LGTM thanks for catching this

MattSturgeon:
@mergifyio queue

refaelsh:
Thank you all :-)

PR #2294, opened 2024-09-22
lib/types/pluginLuaConfig: only merge `pre` and `post` when defined

MattSturgeon:
Follow up to https://github.com/nix-community/nixvim/pull/1876#discussion_r1770586071

- Have `pre` and `post` default to `null`
- Only merge them into `content` when non-null
- Mention this in `content`'s description

cc @traxys

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -116,19 +116,23 @@ rec {
 
   pluginLuaConfig = types.submodule (
     { config, ... }:
+    let
+      inherit (builtins) toString;
+      inherit (lib.nixvim.utils) mkBeforeSection mkAfterSection;
+    in
     {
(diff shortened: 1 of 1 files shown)


traxys (approved):

MattSturgeon:
@mergifyio queue

stasjok:
Since they are in `mkMerge`, maybe it's faster to use `lib.optional`?

MattSturgeon:
I generally default to using `mkIf` to allow the module system to "do its thing". I guess in this case there's nothing really for the module system _to_ do, so using optional(s) should result in less work done.

Not sure if there's a good way to measure such performance differences though.

PR #2300, opened 2024-09-23
tests/package-options: allow option defaults to throw

MattSturgeon:
This is often an intended behaviour, so just assume any option whose default throws is not a package-type option

Extracted from #2022


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2301, opened 2024-09-24
modules/nixpkgs: initial `pkgs` option; drop `defaultPkgs` specialArg

MattSturgeon:
Subset of #2022, working towards #1784 

- **modules: move `nixpkgs` module to top-level modules**
- **modules/nixpkgs: initial `pkgs` option**, _drop `defaultPkgs` specialArg_
- **lib/modules: remove `specialArgsWith`**

## Note for end-users

This PR allows you to configure a `pkgs` for nixvim's submodule that is different to the one used in your "host" configuration. In this context, a "host" configuration could be home-manager, nixos, or nix-darwin.

For example, this means you don't neccessarily have to use nixvim's standalone build in order to use nixvim's main branch together with a "stable" nixpkgs.

Assuming you have `inputs.nixpkgs-unstable` in your flake and you pass `inputs` into your module config somehow (e.g. via `specialArgs`) you can do:
```nix
{ inputs, pkgs, ... }:
{
  # Use nixpkgs-unstable for nixvim:
  programs.nixvim.nixpkgs.pkgs = inputs.nixpkgs-unstable.legacyPackages.${pkgs.stdenv.hostPlatform.system};
}
```

You could use a let-in block to make the above more readable:
```nix
{ inputs, pkgs, ... }:
let
  inherit (pkgs.stdenv.hostPlatform) system;
  pkgs-unstable = inputs.nixpkgs-unstable.legacyPackages.${system};
in
{
  programs.nixvim.nixpkgs.pkgs = pkgs-unstable;
}
```

> [!NOTE]
> In the future we plan to simplify this further, by using the nixpkgs from _our_ flake inputs by default.
>
> This would mean you wouldn't have to do anything in order to use nixvim's main branch with a stable "host" configuration.
>
> However, this would be a breaking change for anyone relying on a specific pkgs instance (configured in their host modules) also being used by their nixvim config.

## Overview

I've implemented _only_ the `nixpkgs.pkgs` option, omitting other `nixpkgs.*` options that would allow instantiating a nixpkgs instance internally for now for the sake of simplicity.

This is enough to allow end-users to set `programs.nixvim.nixpkgs.pkgs` to something _other_ than their host module's `pkgs`, e.g. allowing nixvim unstable to be used on nixos stable without resorting to a _standalone_ nixvim.

- The `nixpkgs.pkgs` option is used to allow an "externally constructed" nixpkgs instance to be used as the pkgs arg.
  (implemented)
- The rest of `nixpkgs.*` is used to "internally" construct a nixpkgs instance, by evaluating a nixpkgs source path with the configured config, system, overlays, etc.
   (unimplemented)

Internally, this also allows the initial migration of `defaultPkgs` -> `pkgsPath`, however I've ran into a couple roadblocks:
- We still need `pkgsPath` to be a specialArg because we import using it. Maybe this is a bad idea and we should just copy the upstream modules?
- If `pkgsPath` is a "string with context", then the `meta` module runs into errors. This means we can't just set `pkgsPath = inputs.nixvim` because that has context.


MattSturgeon:
I'm starting to think attempting to pass a `pkgsPath` in as a specialArg is an anti-pattern. TBH specialArgs are always a last resort for when `_module.args` or another module option can't be used.

Looking at home-manager's implementation, they add the modules they import from nixos directly into the modules list passed to `lib.evalModules`. This avoids the need for a specialArg, but is (technically) no less portable.

Nix-darwin simply copy the relevant modules into their repo, avoiding the need to import anything from nixos's modules. They also define a custom option `nixpkgs.source` which is used as the nixpkgs path to import. This is done using normal option definitions, no specialArgs.

MattSturgeon:
Made a draft for now; I'll refactor `pkgsPath` to no longer be a specialArg after #2302 is merged.

MattSturgeon:
I've dropped the `pkgsPath` stuff from this branch. For now it is enough to simply remove `defaultPkgs`.

In combination with #2307 and #2302, we will now be able to remove `lib.nixvim.modules.specialArgs` and `specialArgsWith` from our public API, as they can be entirely internal to `evalNixvim`.

lib/modules.nix
```diff
@@ -48,4 +40,9 @@ rec {
       (lib.filter (x: !x.assertion))
       (lib.map (x: x.message))
     ];
+
+  # TODO: Removed 2024-09-25
+  # NOTE: integrated into `evalNixvim`
+  specialArgs = "`modules.specialArgs` has been removed.";
```
GaetanLepage, line 46:
```suggestion
  specialArgs = throw "`modules.specialArgs` has been removed.";
```

wrappers/darwin.nix
```diff
@@ -18,11 +18,11 @@ let
   cfg = config.programs.nixvim;
   nixvimConfig = config.lib.nixvim.modules.evalNixvim {
     extraSpecialArgs = {
-      defaultPkgs = pkgs;
       darwinConfig = config;
     };
     modules = [
       ./modules/darwin.nix
+      { nixpkgs.pkgs = lib.mkDefault pkgs; }
```
MattSturgeon, line 25: We could instead do this in a common `wrappers/modules/*.nix` module:

```nix
{ pkgs, lib, ... }:
{
  programs.nixvim.nixpkgs.pkgs = lib.mkDefault pkgs;
}
```

If imported with a relative path, then the module would benefit from automatic `_file` metadata.

Such a module may also be a starting point for a future hm-stlye "useGlobalPackages" option?

MattSturgeon: Moved to `_shared.nix` for now.

lib/tests.nix
```diff
@@ -61,11 +61,14 @@ let
               "mkTestDerivationFromNixvimModule: the `dontRun` argument is deprecated. You should use the `test.runNvim` module option instead."
               { config.test.runNvim = !dontRun; }
           ))
-          { wrapRc = true; }
+          {
+            wrapRc = true;
+            nixpkgs = {
+              inherit pkgs;
+            };
```
MattSturgeon, line 68: Style:

```suggestion
            nixpkgs.pkgs = pkgs;
```

??

modules/top-level/files/default.nix
```diff
@@ -15,6 +15,16 @@ let
     modules = lib.optionals (!config.isDocs) [
       ../../.
       ./submodule.nix
+      # Pass module args through to the submodule (except `name`)
+      # Wrap each arg with the correct priority
+      {
+        _file = ./.;
```
MattSturgeon, line 21: I think `types.submodule` will apply a "default file" gleaned from the submodule options's declaration location.

Testing this out by removing `_file` and adding an error (`foo = "bar"`):

```
error: The option `files."after/ftplugin/python.lua".foo' does not exist. Definition values:
- In `/nix/store/XXX-source/modules/top-level/files': "bar"
Test failure
```

```suggestion
```

lib/tests.nix
khaneliman: I generally think of it as cleaner to flatten where you can and only break it up when it gets too long. (or obviously when the parent attribute is used again)

modules/top-level/nixpkgs.nix
```diff
@@ -0,0 +1,78 @@
...
+  lib,
+  ...
+}:
+let
+  cfg = config.nixpkgs;
+  opt = options.nixpkgs;
+in
+{
+  options.nixpkgs = {
+    pkgs = lib.mkOption {
```
khaneliman, line 13: What other possible options were you planning for with this? I was just curious in the example about the ux of wanting to configure nixvim's `pkgs` and it felt off to be nested under a nixpkgs attribute, for some reason. I don't know if it was just because it makes me think it implies we might allow setting a different `pkgs` for another input or something..

khaneliman: As I wrote it out, though, I think the nesting makes sense.

khaneliman (approved):

khaneliman:
@MattSturgeon I feel like this has come up numerous times, did we have an issue associated with this or was it always just discussions and matrix?

modules/top-level/nixpkgs.nix
MattSturgeon: I intend to add most options available in nixos/hm/nix-darwin's `nixpkgs.*` modules.

In particular:
- `source`: the nixpkgs source to be imported
  (hopefully we can default to our `inputs.nixpkgs`)
- `config`: the so-called [global configuration](https://nixos.org/manual/nixpkgs/unstable/#chap-packageconfig) ([opts](https://nixos.org/manual/nixpkgs/unstable/#sec-config-options-reference))
- `overlays`: overlays to extend pkgs with
- platform options, `hostPlatform`, `buildPlatform` etc

khaneliman: > I intend to add most options available in nixos/hm/nix-darwin's `nixpkgs.*` modules.
> 
> In particular:
> 
>     * `source`: the nixpkgs source to be imported
>       (hopefully we can default to our `inputs.nixpkgs`)
> 
>     * `config`: the so-called [global configuration](https://nixos.org/manual/nixpkgs/unstable/#chap-packageconfig) ([opts](https://nixos.org/manual/nixpkgs/unstable/#sec-config-options-reference))
> 
>     * `overlays`: overlays to extend pkgs with
> 
>     * platform options, `hostPlatform`, `buildPlatform` etc
> 
> 
> In contrast to the `pkgs` option added in this PR, most of these options relate to constructing a pkgs instance internally.

Awesome, that's helpful context. Thanks!

MattSturgeon:
> @MattSturgeon I feel like this has come up numerous times, did we have an issue associated with this or was it always just discussions and matrix? Or was it just the one you created to track it? Sorry, I'm tired and scatter brained posting... lol

The tracking issue is https://github.com/nix-community/nixvim/issues/1784

modules/top-level/nixpkgs.nix
```diff
@@ -0,0 +1,78 @@
...
+      #   }
+      # '';
+      type = lib.types.pkgs // {
+        description = "An evaluation of Nixpkgs; the top level attribute set of packages";
+      };
+      example = lib.literalExpression "import <nixpkgs> { }";
+      description = ''
+        If set, the `pkgs` argument to all Nixvim modules is the value of this option.
+
+        <!-- TODO: remove -->
```
khaneliman, line 27: Forgot to ask what was going on with the multiple TODO's here.

MattSturgeon: Each TODO relates to a section of documentation that should be added/removed once the other `nixpkgs.*` options are implemented.

MattSturgeon: It currently renders as:

image

modules/top-level/nixpkgs.nix
```diff
@@ -0,0 +1,78 @@
...
+in
+{
+  options.nixpkgs = {
+    pkgs = lib.mkOption {
+      # TODO:
+      # defaultText = lib.literalExpression ''
+      #   import "''${nixos}/.." {
+      #     inherit (cfg) config overlays localSystem crossSystem;
+      #   }
+      # '';
```
MattSturgeon, line 19: It probably makes sense to include a defaultText explaining the current behaviour:

```nix
defaultText = literalMD ''
  The `pkgs` inherited from your host config (i.e. NixOS, home-manager, or nix-darwin),
  or the `pkgs` supplied to `makeNixvimWithModule` when building a standalone nixvim.

  > [!CAUTION]
  > This default will be removed in a future version of nixvim
''
```

MattSturgeon: I've added this, it now renders the default as:
image

MattSturgeon:
@mergifyio queue

PR #2302, opened 2024-09-24
modules/misc: include `assertions` and `meta` modules in the repo

MattSturgeon:
Based on the modules we previously imported from NixOS. I've re-formatted the files, removed `with lib` and re-wrote a couple things to better fit the code-style in nixvim. You can view the upstream files for comparison:
- https://github.com/NixOS/nixpkgs/blob/814a4e48/nixos/modules/misc/assertions.nix
- https://github.com/NixOS/nixpkgs/blob/814a4e48/nixos/modules/misc/meta.nix

This will allow us to drop the `defaultPkgs` specialArg and avoids needing the proposed `pkgsPath` specialArg.

It could also lead to future divergence from upstream (intended or otherwise), for example as suggested in #2309.

### DIFF
--- a/modules/misc/meta.nix
+++ b/modules/misc/meta.nix
@@ -1,0 +1,56 @@
+{ lib, ... }:
+# Based on https://github.com/NixOS/nixpkgs/blob/814a4e48/nixos/modules/misc/meta.nix
+let
+  maintainer = lib.mkOptionType {
+    name = "maintainer";
+    check = m: lib.elem m (lib.attrValues lib.maintainers);
+    merge =
+      loc: defs:
+      let
+        def = lib.last defs;
+      in
+      {
+        ${def.file} = def.value;
+      };
+  };
+
+  listOfMaintainers = lib.types.listOf maintainer // {
+    # Returns attrset of
+    #   { "module-file" = [
+    #        "maintainer1 <first@nixos.org>"
+    #        "maintainer2 <second@nixos.org>" ];
+    #   }
+    merge =
+      loc: defs:
+      lib.pipe defs [
+        (lib.imap1 (
+          n: def:
+          lib.imap1 (
+            m: value:
+            maintainer.merge (loc ++ [ "[${toString n}-${toString m}]" ]) [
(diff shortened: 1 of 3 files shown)


modules/misc/meta.nix
```diff
@@ -0,0 +1,73 @@
...
+    #   internal = true;
+    #   example = "./meta.chapter.md";
+    #   description = ''
+    #     Documentation prologue for the set of options of each module.  This
+    #     option should be defined at most once per module.
+    #   '';
+    # };
+  };
+
+  meta.maintainers = [ lib.maintainers.pierron ];
```
traxys, line 72: I don't think we want to leave this maintainer x)

modules/misc/meta.nix
```diff
@@ -0,0 +1,73 @@
...
+    #     # { file = "module location"; value = <path/to/doc.xml>; }
+    #     merge = loc: defs: defs;
+    #   };
+    #   internal = true;
+    #   example = "./meta.chapter.md";
+    #   description = ''
+    #     Documentation prologue for the set of options of each module.  This
+    #     option should be defined at most once per module.
+    #   '';
+    # };
```
traxys, line 69: Yeah I agree that we should maybe at some point merge our nixvimInfo with this module, as they basically serve the same purpose. Though let's do that later

MattSturgeon: Shall we keep the commented code as a reminder, or remove it and open an issue?

modules/misc/meta.nix
MattSturgeon: I guess not :+1:

modules/misc/meta.nix
traxys: I think an issue is a better reminder

MattSturgeon: Opened https://github.com/nix-community/nixvim/issues/2309

traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2306, opened 2024-09-24
modules/output: move outputs to `build` scope

MattSturgeon:
- **modules/output: move outputs to `build` scope**
- **modules/output: add `build.initSource` option**

This PR refactors our output options by moving anything we're "building" to `config.build.*`. This is also used as an opportunity to rename some of the options. Feedback on the names chosen is appreciated!

The intention is adding some structure to our output options, allowing for growth without things getting too messy..

#2313 is also based on this, since otherwise I'd have to re-write it to resolve merge conflicts.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -68,24 +68,31 @@ in
       '';
     };
 
-    finalPackage = mkOption {
-      type = types.package;
-      description = "Wrapped Neovim.";
-      readOnly = true;
-    };
+    build = {
+      # TODO: `standalonePackage`; i.e. package + printInitPackage + man-docs bundled together
 
-    initPath = mkOption {
-      type = types.str;
-      description = "The path to the `init.lua` file.";
-      readOnly = true;
-      visible = false;
-    };
+      package = mkOption {
+        type = types.package;
+        description = "Wrapped Neovim.";
+        readOnly = true;
+      };
 
-    printInitPackage = mkOption {
-      type = types.package;
-      description = "A tool to show the content of the generated `init.lua` file.";
-      readOnly = true;
-      visible = false;
+      initFile = mkOption {
+        type = types.path;
+        description = "The generated `init.lua` file.";
+        readOnly = true;
+        visible = false;
+      };
+
+      printInitPackage = mkOption {
+        type = types.package;
+        description = ''
+          A tool to show the content of the generated `init.lua` file.
+          Run using `${config.build.printInitPackage.meta.mainProgram}`.
+        '';
+        readOnly = true;
+        visible = false;
+      };
     };
   };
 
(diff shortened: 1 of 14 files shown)


lib/deprecation.nix
```diff
@@ -59,6 +59,52 @@ rec {
       lib.mkRenamedOptionModule oldPath newPath
     );
 
+  # Based on `lib.modules.doRename` and `mkRenamedOptionModule`,
+  # but the `to` config alias is only defined when `from.isDefined`.
+  # This is intended for aliasing `readOnly` options, where defining `to` could trip assertions.
+  mkReadOnlyRenamedOptionModule =
```
MattSturgeon, line 65: I'd like to simplify or remove this before merging

MattSturgeon:
I've dropped the assertion/warning change from this PR and moved it into a followup: #2313 

Hopefully the reduced scope makes it easier to review!

modules/top-level/output.nix
```diff
@@ -50,24 +50,49 @@ in
...
-      readOnly = true;
-      visible = false;
+          This is usually identical to `build.initFile`, however if `performance.byteCompileLua` is enabled,
+          this option will refer to the un-compiled lua source file.
+        '';
+        readOnly = true;
+        visible = false;
+      };
+
+      printInit = mkOption {
```
khaneliman, line 87: I don't know if I like this rename when reading through its usage in the tests and modules. Seems like its a change from our <name>Package usage everywhere for package options.

MattSturgeon: Makes sense. If no one has any other suggestions I'll go with `build.printInitPackage` then.

MattSturgeon: (Done)

khaneliman (approved):

GaetanLepage (commented):
Overall LGTM and I trust you on the necessity of this change.
You are surely thinking of more long-term refactoring.

One question: you went from `initPath` to `initSource` "implicitly". What is the motivation for this change ?

MattSturgeon:
> One question: you went from `initPath` to `initSource` "implicitly". What is the motivation for this change ?

Actually, `initPath` was renamed to `initFile` as I felt this more accurately conveys that it is a built file in the nix-store.

`initSource` is a new option that allows users to access the init file _without_ byte compilation (as seen in `nixvim-print-init`). It is only different to `initFile` when `performance.byteCompileLua` is used.

Maybe `initSourceFile` would actually be a better name? Otherwise "source" may wrongly imply it is an input option rather than an output?

GaetanLepage:
> Actually, `initPath` was renamed to `initFile` as I felt "file" more accurately conveys that it is a built file in the nix-store. Open to feedback though.

Oh right. I agree, this is a better name.



> `initSource` is a new option that allows users to access the init file _without_ byte compilation (as seen in `nixvim-print-init`). It is only different to `initFile` when `performance.byteCompileLua` is used.

Yes, this was perfectly clear to me thanks to the comments.



> Maybe `initSourceFile` would actually be a better name? Otherwise "source" may wrongly imply it is an input option rather than an output?

I think it's fine, no need to change it.

GaetanLepage (approved):

traxys (approved):

PR #2307, opened 2024-09-24
wrappers: use `(evalModules {}).type` for nixvim submodule

### LINKED ISSUE #2305: Set `shorthandOnlyDefinesConfig = false;` in hm-config
## Description
The home manager module defines the nixvim module with `shorthandOnlyDefinesConfig = true;`. This means when trying to import a config module, one gets an error unless they make the nixvim config module into a function
https://github.com/nix-community/nixvim/blob/main/wrappers/hm.nix#L24

```nix
# as a home-manager module
programs.nixvim = {
  imports = [
    ./nixvim-config.nix 
  ]
}
```
Causes the error `error: The option `home-manager.users.username.programs.nixvim.imports' does not exist.`

To fix it, one has to [declare the home manager to not be "shorthand"](https://nlewo.github.io/nixos-manual-sphinx/development/option-types.xml.html#:~:text=With%20this%20option%20enabled%2C%20defining%20a%20non%2Dconfig%20section%20requires%20using%20a%20function%3A%20the%2Dsubmodule%20%3D%20%7B%20...%20%7D%3A%20%7B%20options%20%3D%20%7B%20...%20%7D%3B%20%7D.), which gets rid of the error:

```nix
# as a home-manager module
programs.nixvim = {...}:{
  imports = [
    ./nixvim-config.nix 
  ]
}
```

# The Fix
Set `shorthandOnlyDefinesConfig = false;`. I dont see nixvim using a config option, so this should not break anything. 

## My Usecase
I want to expose nixvim as a package of my config flake as well as be part of the home manager config. Therefore I have one single config that defines the nixvim options which is then imported in my home-manager config and exposed as a package via `inputs'.nixvim.legacyPackages.makeNixvim`

MattSturgeon:
This simplifies the wrapper modules, adding a little consistency.

Rather than passing all the same stuff to `types.submoduleWith` that we do internally in `evalNixvim`, we can just use `evalNixvim` itself and then inherit its [`type`](https://github.com/NixOS/nixpkgs/blob/b4cde4d7f1fc84f6ff41ad165dab793dedaedfa0/lib/modules.nix#L327-L329) output.

This also means `modules.specialArgsWith` is now entirely internal and could be removed from out public API before we branch off 24.11.

This also fixes #2305 because we no longer enable `shorthandOnlyDefinesConfig`. We added that a while ago because we had an `options` option, however we renamed that to `opts` in #1342 so we no longer benefit from it.

### DIFF
--- a/modules/top-level/files/default.nix
+++ b/modules/top-level/files/default.nix
@@ -10,7 +10,6 @@ let
   inherit (lib) types;
 
   fileModuleType = types.submoduleWith {
-    shorthandOnlyDefinesConfig = true;
     inherit specialArgs;
     # Don't include the modules in the docs, as that'd be redundant
     modules = lib.optionals (!config.isDocs) [
--- a/wrappers/darwin.nix
+++ b/wrappers/darwin.nix
@@ -16,22 +16,22 @@ let
     types
     ;
   cfg = config.programs.nixvim;
+  nixvimConfig = config.lib.nixvim.modules.evalNixvim {
+    extraSpecialArgs = {
+      defaultPkgs = pkgs;
+      darwinConfig = config;
+    };
+    modules = [
+      ./modules/darwin.nix
+    ];
+    check = false;
+  };
 in
 {
   options = {
     programs.nixvim = mkOption {
+      inherit (nixvimConfig) type;
       default = { };
-      type = types.submoduleWith {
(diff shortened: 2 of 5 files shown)


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2312, opened 2024-09-25
wrappers: explicitly set `_file` for wrapper modules

MattSturgeon:
This won't be detected automatically, since the modules are all returned from function calls instead of being path references.


traxys (approved):

MattSturgeon:
@mergifyio queue

PR #2313, opened 2024-09-25
modules/output: check warnings+assertions on `build.package` option

MattSturgeon:
Instead of `evalNixvim` checking `config.assertions` and `config.warnings`, this can be done when reading the `config.build.package` option.

This is intended to simplify `evalNixvim`, the wrappers, and the test derivation

This is similar to the approach used by NixOS, where assertions are [checked](https://github.com/NixOS/nixpkgs/blob/2af19cfb6aa40768c4bbefd801a136270e099191/nixos/modules/system/activation/top-level.nix#L62-L68) while evaluating its `config.system.build.toplevel` [option value](https://github.com/NixOS/nixpkgs/blob/2af19cfb6aa40768c4bbefd801a136270e099191/nixos/modules/system/activation/top-level.nix#L357).

The previously incubating `check = false` argument is now replaced with an "unchecked" package option. This is used by `config.build.test` and by wrapper modules that prefer to pass nixvim's assertions up to the host config.

<details><summary>Notes regarding 2306</summary>
<p>

This was originally part of #2306 ~~and is blocked by that PR~~.

I've pushed up the original version (based on #2306) to [`build_opt_checks_warnings_on_2306`](https://github.com/mattsturgeon/nixvim/tree/build_opt_checks_warnings_on_2306), for reference when resolving the inevitable merge conflicts :cold_sweat: 

</p>
</details>

### DIFF
--- a/flake-modules/legacy-packages.nix
+++ b/flake-modules/legacy-packages.nix
@@ -15,7 +15,6 @@
           extraSpecialArgs = {
             defaultPkgs = pkgs;
           };
-          check = false;
         };
       };
     };
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -20,32 +20,17 @@ rec {
     {
       modules ? [ ],
       extraSpecialArgs ? { },
-      # Set to false to disable warnings and assertions
-      # Intended to aid accessing config.test.derivation
-      # WARNING: This argument may be removed without notice:
-      check ? true,
-    }:
-    let
-      result = lib.evalModules {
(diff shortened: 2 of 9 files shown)


MattSturgeon:
Instead of waiting for #2306, I've rebased this PR onto main and pushed a snapshot of what it looked like based on #2306 to `build_opt_checks_warnings_on_2306`.

MattSturgeon:
@mergifyio queue

PR #2323, opened 2024-09-27
docs/home-manager: eval options without checking config definitions

MattSturgeon:
By default `lib.evalModules` will check all config definitions match a declared option, leading to errors like:

```
error: The option `wrapRc' does not exist.
```

Defining a [`freeformType`](https://github.com/NixOS/nixpkgs/blob/de38446996523da09e299aee94ceb73bd83a8bb4/lib/modules.nix#L205-L219) would instead check such definitions match the given type, producing mismatched type errors instead of "does not exist".

[`_module.check`](https://github.com/NixOS/nixpkgs/blob/de38446996523da09e299aee94ceb73bd83a8bb4/lib/modules.nix#L198-L203) can disable definition checking entirely, allowing us to evaluate the option declaration without checking the config definitions.

If `_module.check` is disabled, then definitions for missing options will simply not exist in the final merged `config` attrs.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -42,8 +42,12 @@ let
     ];
   };
 
-  hmOptions = builtins.removeAttrs (lib.evalModules { modules = [ ../wrappers/modules/hm.nix ]; })
-    .options [ "_module" ];
+  hmOptions = builtins.removeAttrs (lib.evalModules {
+    modules = [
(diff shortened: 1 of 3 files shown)


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2332, opened 2024-09-28
lib: standardize use of `lib` and `self`

MattSturgeon:
Things done:
- Drop `helpers` in internal useage
- Access nixvim's lib via `self`
- Access the (un-extended) lib via `lib`

While I like the idea of having `lib.nixvim` access internally within our lib, keeping our lib (`self`) separate from the extended lib will be better in the long-term.

This is intended to simplify accessing a lib overlay without needing to first evaluate parts of our custom lib (#2328).

~~`lib/types.nix` is a bit of a special case,~~ it isn't _really_ part of our (`lib.nixvim` scoped) custom lib. Instead it is part of the lib extension overlay. Therefore we can use either strategy of passing it `{ lib = final; }` _or_ `{ lib = prev; inherit self; }`. Feedback is welcome on which approach is preferred.

EDIT: the above paragraph isn't true anymore. To avoid awkward bootstrapping we now (once again) build our types internally, as `lib.nixvim.types`. This is still merged into `lib.types` in the overlay though.

### DIFF
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -15,7 +15,7 @@
         # NOTE: this is the publicly documented flake output we've had for a while
         check = import ../lib/tests.nix { inherit lib pkgs; };
         # NOTE: user-facing so we must include the legacy `pkgs` argument
-        helpers = import ../lib { inherit lib pkgs; };
+        helpers = (import ../lib { inherit lib pkgs; }).public;
       }
     )
   );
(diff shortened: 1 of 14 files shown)


MattSturgeon:
CI failures have highlighted some cases where we use our custom types within our nixvim scoped lib.

I wonder if types should once again be built within `lib.nixvim`, and then this final value can be merged later in the overlay... :thinking:

PR #2336, opened 2024-09-29
lib: remove `helpers` from internal usage

MattSturgeon:
Simply removes all usage of `helpers` from the internals of `lib.nixvim`. It is also dropped from the available attrs in `call`.


GaetanLepage (approved):
Diff LGTM

MattSturgeon:
@mergifyio queue

PR #2337, opened 2024-09-29
lib/types: make `strLua` coerce strings to rawLua

MattSturgeon:
- **lib/types: make `strLua` coerce strings to rawLua**
- **lib/options: remove redundant `mkRaw` `apply` functions**
- **plugins: remove redundant `mkRaw` `apply` functions**

Related to #2003, but does not yet take any steps to deprecate strLua.

This PR simply shifts responsibility for coercing strings to rawLua from the option to the type.

In normal usage, this shouldn't break anything. If an option is also applying `mkRaw` then that is redundant, but idempotent and will do no harm. Also, `apply` happens _after_ the type is merged anyway.

In abnormal usage, this could break things. E.g. if a user is incorrectly using `strLua` in their own option, but **not** expecting the value to be coerced to rawLua... While worth noting, I don't think we need to be concerned about such incorrect usage.

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -32,7 +32,22 @@ rec {
     check = v: (isRawType v) || (v ? __empty);
   };
 
-  maybeRaw = type: types.either type rawLua;
+  maybeRaw =
+    elemType:
+    let
+      luaFirst = types.either rawLua elemType;
+      elemFirst = types.either elemType rawLua;
+    in
+    luaFirst
+    // {
+      name = "maybeRaw";
+      inherit (elemFirst) description;
+      nestedTypes = {
+        left = lib.warn "maybeRaw.nestedTypes: `left` is a deprecated alias for `elemType`." elemType;
+        right = lib.warn "maybeRaw.nestedTypes: `right` is a deprecated alias for `rawLua`." rawLua;
+        inherit rawLua elemType;
+      };
+    };
 
   # Describes an boolean-like integer flag that is either 0 or 1
   # Has legacy support for boolean definitions, added 2024-09-08
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -8,14 +8,21 @@ let
     mkNullOrOption
     ;
 
-  strLikeType =
+  mkStrLuaType =
     description:
     lib.mkOptionType {
-      name = "str";
+      name = "strLua";
       inherit description;
       descriptionClass = "noun";
       check = v: lib.isString v || isRawType v;
-      merge = lib.options.mergeEqualOption;
+      merge =
+        loc: defs:
+        lib.pipe defs [
+          # Coerce strings to rawLua
+          # TODO: consider deprecating this behaviour
+          (lib.map (def: def // { value = lib.nixvim.mkRaw def.value; }))
+          (lib.options.mergeEqualOption loc)
+        ];
     };
   isRawType = v: v ? __raw && lib.isString v.__raw;
 in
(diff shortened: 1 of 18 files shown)


traxys (approved):
I'd love to deprecate strLua, but I think that it's a massive endeavour, due to all the legacy plugins that _may_ be affected

MattSturgeon:
This is failing for `maybeRaw (attrsOf strLua)` types.

Because `maybeRaw` checks the given type _before_ checking for `rawLua`, the following definition is treated as `attrsOf strLua`:

```nix
{ __raw = "foo"; }
```

And is therefore merged as:
```nix
{
  __raw.__raw = "foo";
}
```

This is correct for an `attrsOf strLua`, but not for a `maybeRaw (attrsOf strLua)`.

To fix this, we need to have `maybeRaw` check for rawLua _before_ checking the other type.

lib/types.nix
```diff
@@ -32,7 +39,17 @@ rec {
...
+    type:
+    let
+      luaFirst = types.either rawLua type;
+      typeFirst = types.either type rawLua;
+    in
+    luaFirst
+    // {
+      name = "maybeRaw";
+      inherit (typeFirst) description;
+    };
```
MattSturgeon, line 52: @traxys are you still happy after this addition to the PR?
This fixes https://github.com/nix-community/nixvim/pull/2337#issuecomment-2381400788.

MattSturgeon:
```
… while evaluating definitions from `/nix/store/XXX-source/plugins/lsp/language-servers/efmls-configs.nix':
… while evaluating the option `plugins.efmls-configs.setup.all.linter':
error: expected a list but found null: null
```

lib/types.nix
MattSturgeon: https://github.com/nix-community/nixvim/pull/2337#issuecomment-2381405395 was resolved by also overriding `nestedTypes` to match the prior attrs.

traxys (commented):
Ah I didn't think about that. The changes are still fine

lib/types.nix
```diff
@@ -32,7 +39,22 @@ rec {
...
+  maybeRaw =
+    elemType:
+    let
+      luaFirst = types.either rawLua elemType;
+      elemFirst = types.either elemType rawLua;
+    in
+    luaFirst
+    // {
+      name = "maybeRaw";
+      inherit (elemFirst) description;
```
MattSturgeon, line 51: Alternatively, we could construct a description ourselves. But due to nix's lazyness, calling `either` just to build a description is still fairly cheap.

If we wanted to do a custom description for maybeRaw, perhaps another PR can tackle that?

MattSturgeon: For reference, here's `either`'s description [impl](https://github.com/NixOS/nixpkgs/blob/21be6da8c784dfd6284f39f9782960387a44ac62/lib/types.nix#L975-L981):
```nix
description =
  if t1.descriptionClass or null == "nonRestrictiveClause"
  then
    # Plain, but add comma
    "${t1.description}, or ${optionDescriptionPhrase (class: class == "noun" || class == "conjunction") t2}"
  else
    "${optionDescriptionPhrase (class: class == "noun" || class == "conjunction") t1} or ${optionDescriptionPhrase (class: class == "noun" || class == "conjunction" || class == "composite") t2}";
```

MattSturgeon:
@mergifyio queue

khaneliman:
I was getting so confused when I was rebasing my strlua branch... didn't realize this got done already lmao

MattSturgeon:
My bad didn't spot you had a branch! Did your branch cover anything I missed?

PR #2341, opened 2024-09-29
lib/types: `maybeRaw` should propagate `elemType`'s sub-options

MattSturgeon:
This is _technically_ redundant, because we patch `either` to be recursive. But I think it is best if `maybeRaw` does this explicitly, without relying on that behaviour.

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -54,6 +54,11 @@ rec {
         right = lib.warn "maybeRaw.nestedTypes: `right` is a deprecated alias for `rawLua`." rawLua;
         inherit rawLua elemType;
       };
+      inherit (elemType)
+        getSubModules
+        getSubOptions
+        substSubModules
+        ;
     };
 
   # Describes an boolean-like integer flag that is either 0 or 1


MattSturgeon:
```
error: A definition for option `plugins.cmp.settings.sources' is not of type `list of (attribute set of anything)'. Definition values:
- In `/nix/store/a9v1l0scrw2nmsp7k40kywijf78mns6s-test-sources/plugins/cmp/default.nix':
    {
      __raw = ''
        cmp.config.sources({
          {
            {name = "path"},
    ...
```

I have no idea how _this change_ could trigger that though...

lib/types.nix
```diff
@@ -54,6 +54,11 @@ rec {
         right = lib.warn "maybeRaw.nestedTypes: `right` is a deprecated alias for `rawLua`." rawLua;
         inherit rawLua elemType;
       };
+      inherit (elemType)
+        getSubModules
+        getSubOptions
+        substSubModules
```
MattSturgeon, line 60: I think a proper `substSubModules` is a little more complex than just re-using the elem-type's

GaetanLepage:
Is this PR still relevant?

MattSturgeon:
> Is this PR still relevant?

Forgot it existed tbh, but I think it may still be relevant for potentially allowing us to drop our patches for making `types.either` return sub-options.

It needs some cleanup or reduction in scope, as I didn't understand things like `getSubsetModules` when I wrote this.

PR #2342, opened 2024-09-29
lib/keymaps: remove `lua` attr in type's `merge` fn

MattSturgeon:
Rather than having each option remove `lua` in their `apply` function, we can do this in the submodule type's `merge` function.

Yes, it's a little annoying that we have to indent the submodule definition in order to override it's `merge` attr.

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -61,11 +61,8 @@ rec {
 
   mapOptionSubmodule = mkMapOptionSubmodule { };
 
-  # NOTE: options that have the deprecated `lua` sub-option must use `removeDeprecatedMapAttrs`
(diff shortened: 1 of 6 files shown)


MattSturgeon:
Testing in the repl:

With `lib.modules.mergeDefinitions` and/or the type's `merge` function, using `lib.nixvim.keymaps.deprecatedMapOptionSubmodule` seems to work, but using `options..keymaps.type.nestedTypes.elemType` does not.

I don't understand, because these two should be the same thing...

This feels like the same weirdness that we ran into when trying to override `types.coercedTo` for use in `transitionType`.

MattSturgeon:
Could this be related to the option-types default functor?

Issue #2335, opened 2024-09-29
[PLUGIN REQUEST] lsp.mdx_analyzer


Eveeifyeve:
Add support for https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#mdx_analyzer
As this will allow nix users to use the mdx lsp for mdx projects.

Eveeifyeve:
I am looking to make a pr, but I am waiting on a package.

Issue #2163, opened 2024-09-04
[PLUGIN REQUEST] Add somesass_ls LSP server


Yasso9:
Hi, thanks a lot for this amazing project !

Would it be possible to add somesass_ls, a LSP server for SCSS and SASS (alternative to CSS) ?

Here is the server configurations part on the LSP documentation : https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#somesass_ls

khaneliman:
I couldn't find it in nixpkgs upstream. We need to get that added upstream before we can add a module for it.

PR #2365, opened 2024-10-05
plugins/lsp: Auto generate the list of lsp servers

traxys:
This PR allows to auto-generate the list of all lsp servers from the plugin.
The approach taken mirrors the way upstream generates their documentation.

Closes https://github.com/nix-community/nixvim/issues/2163
Closes https://github.com/nix-community/nixvim/issues/2335


Eveeifyeve:
This solves: #2335

update-scripts/nvim-lspconfig/desc-filter.lua
```diff
@@ -0,0 +1,3 @@
+function Header(elem)
+	return pandoc.Strong(elem.content)
+end
```
khaneliman, line 3: Wish inline lua in python didn't require an additional dependency...

update-scripts/nvim-lspconfig/default.nix
```diff
@@ -0,0 +1,36 @@
...
+    lspconfig = "${vimPlugins.nvim-lspconfig}";
+    nativeBuildInputs = [
+      pandoc
+      python3
+    ];
+  }
+  ''
+    export HOME=$(realpath .)
+    # Generates `lsp.json`
+    ${nvim}/bin/nvim -u NONE -E -R --headless +'luafile ${./lspconfig-servers.lua}' +q
```
khaneliman, line 34:
```suggestion
    ${lib.getExe nvim} -u NONE -E -R --headless +'luafile ${./lspconfig-servers.lua}' +q
```
?

khaneliman (commented):
This is an awesome job, this will reduce a lot of maintenance burden!

update-scripts/nvim-lspconfig/lspconfig-servers.lua
```diff
@@ -0,0 +1,52 @@
...
+		return func(k, t[k])
+	end)
+end
+
+require_all_configs()
+
+info = sorted_map_table(configs, function(server_name, server_info)
+	local description = nil
+	if server_info.document_config.docs ~= nil then
+		description = server_info.document_config.docs.description
```
khaneliman, line 37: Do they provide any other metadata? Or does everything get thrown in here?

generated/lspconfig-servers.json
```diff
@@ -0,0 +1,1705 @@
+[
+  {
+    "cmd": ["als"],
+    "desc": "https://github.com/agda/agda-language-server\n\nLanguage Server for Agda.\n",
```
khaneliman, line 4: Was just curious because I see they basically always have the website and then a description in here. Made me think about how we provide a link to upstream docs for the plugins and was curious if we could grab the url and strip it out to be used for the proper link instead of crowding the description with it.

plugins/lsp/language-servers/default.nix
```diff
@@ -6,197 +6,52 @@
...
-    {
-      name = "dagger";
-      description = "dagger for Cuelang";
-      package = "cuelsp";
-    }
-    {
-      name = "dartls";
-      description = "dart language-server";
-      package = "dart";
+  renamedServers = {
```
khaneliman, line 9: Should we split the renames into a deprecations.nix and the override / extended configs into a separate file that can be imported below?

update-scripts/nvim-lspconfig/lspconfig-servers.lua
traxys: No there is quite a lot of metadata, like the default values for root patterns, filetypes, or other configuration items.
I went with the minimal amount to replicate what we had

generated/lspconfig-servers.json
traxys: I'd rather to do that in another PR, but it would be better I agree

khaneliman (approved):
Thanks for tackling this!

update-scripts/nvim-lspconfig/lspconfig-servers.lua
MattSturgeon: Agreed, we can expand on this in future PRs if there's a need.

traxys:
@mergify queue

plugins/lsp/language-servers/_renamed.nix
SuperSandro2000, line 1: Why aren't the names normalized? In nixpkgs/NixOS we usually normalize packages to not contain underscores and options unless they are freeform also usually don't contain underscores.

plugins/lsp/language-servers/default.nix
```diff
@@ -739,27 +171,48 @@ let
...
-    }
-  ];
-  renamedServers = {
-    tsserver = "ts-ls";
+    };
   };
+
+  lspPackages = import ../lsp-packages.nix;
+
+  getLspPackage =
```
SuperSandro2000, line 182: I find it a bit confusing that you can activate an lsp that doesn't exist eg angularls and don't get any warning that there is no package for it. Don't get me wrong, I love it, that it can now be easily extended I just want to know when package is null and I need to manually set it :)

SuperSandro2000: see https://github.com/NuschtOS/nvim.nix/pull/61 , specifically https://github.com/NuschtOS/nvim.nix/commit/938a5c576c67ae7651094974c35f8620f3611916

traxys: We have that behaviour for none-ls, so I thought that the lsp servers were the same, it is indeed much better to require an explicit null to acknowlege that the server must be installed by other means

MattSturgeon: OTTOMH, the other places where this is possible (none-ls & efmls) have no-default package options in cases where we don't have a known package.

This forces the user to set the option explicitly or they will get "option accessed but not defined" assertion errors.

plugins/lsp/lsp-packages.nix
MattSturgeon, line 1: I can't see any changes to `tests/generated.nix`, but usually we'd want to update the test there to validate the generated options and declared packages are in sync.

tests/default.nix
```diff
@@ -42,24 +42,37 @@ let
...
+    (helpers.groupListBySize 10)
+    (lib.imap1 (
+      i: group: rec {
+        name = "test-${toString i}";
+        value = pkgs.linkFarm name group;
+      }
+    ))
+    builtins.listToAttrs
+  ]
+)
```
MattSturgeon, line 73: Hard to tell on mobile, but is this just a formatting change brought on by adding (unnecessary) parentheses?

traxys: Yeah I think. I added them at some point when trying something else

Issue #2362, opened 2024-10-04
[BUG] Deprecated option nvim-ufo


aspauldingcode:
| Field        | Description |
|--------------|-------------|
| Plugin       | `nvim-ufo`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `24.05` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description
Option `close_fold_kinds` in `nvim-ufo` is deprecated, 
use `close_fold_kinds_for_ft` instead.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
    plugins.nvim-ufo = {
        enable = true;
        enableGetFoldVirtText = true;
        closeFoldKinds = {
            imports = true;
            comment = true;
        };
    };
}
```

PR #2369, opened 2024-10-06
plugins/nvim-ufo: migrate to mkNeovimPlugin

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2362

### DIFF
--- a/plugins/by-name/nvim-ufo/default.nix
+++ b/plugins/by-name/nvim-ufo/default.nix
@@ -1,87 +1,133 @@
 {
   lib,
-  config,
-  pkgs,
   ...
 }:
 let
   inherit (lib.nixvim) defaultNullOpts;
-
-  cfg = config.plugins.nvim-ufo;
 in
-{
-  options.plugins.nvim-ufo = lib.nixvim.neovim-plugin.extraOptionsOptions // {
-    enable = lib.mkEnableOption "nvim-ufo";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nvim-ufo";
+  luaName = "ufo";
+  package = "nvim-ufo";
 
-    package = lib.mkPackageOption pkgs "nvim-ufo" {
-      default = [
-        "vimPlugins"
-        "nvim-ufo"
-      ];
-    };
+  maintainers = [ lib.maintainers.khaneliman ];
 
-    openFoldHlTimeout = defaultNullOpts.mkInt 400 ''
+  # TODO: added 2024-10-05 remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    "openFoldHlTimeout"
+    "providerSelector"
+    "closeFoldKinds"
+    "foldVirtTextHandler"
+    "enableGetFoldVirtText"
+    [
+      "preview"
+      "winConfig"
+      "border"
+    ]
+    [
+      "preview"
+      "winConfig"
+      "winblend"
+    ]
+    [
+      "preview"
+      "winConfig"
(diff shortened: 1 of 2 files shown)


plugins/by-name/nvim-ufo/default.nix
```diff
@@ -1,87 +1,134 @@
...
 let
-  cfg = config.plugins.nvim-ufo;
+  inherit (lib.nixvim) defaultNullOpts;
 in
-{
-  options.plugins.nvim-ufo = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "nvim-ufo";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nvim-ufo";
+  originalName = "nvim-ufo";
```
GaetanLepage, line 10:
```suggestion
```
I think that this defaults to `name`.

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -1,87 +1,133 @@
...
+      "maxheight"
+    ]
+    [
+      "preview"
+      "mappings"
+    ]
+  ];
+
+  settingsOptions = {
+    open_fold_hl_timeout = defaultNullOpts.mkInt 400 ''
```
GaetanLepage, line 50:
```suggestion
    open_fold_hl_timeout = defaultNullOpts.mkUnsignedInt 400 ''
```

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -1,87 +1,133 @@
...
-    };
+  maintainers = [ lib.maintainers.khaneliman ];
 
-    openFoldHlTimeout = helpers.defaultNullOpts.mkInt 400 ''
+  # TODO: added 2024-10-05 remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings = [
+    "openFoldHlTimeout"
+    "providerSelector"
+    "closeFoldKinds"
```
GaetanLepage, line 20: Isn't the new name `close_fold_kinds_for_ft` ? If so, you need to manually add an `mkRenamedOptionSubmodule` to `imports`.

GaetanLepage (changes requested):

plugins/by-name/nvim-ufo/default.nix
khaneliman: It is deprecated and not removed. They are different types, as well. This is a rename for compatibility but the documented option is the new name.

khaneliman: https://github.com/kevinhwang91/nvim-ufo/commit/7b14dd650b57291a49f037f1ede325d46e08b258

plugins/by-name/nvim-ufo/default.nix
khaneliman: Wondering if the winblend and maxheight should be, as well..

khaneliman: Updated them both, winblend is 0-100 and maxheight doesn't sound like something that could be negative.

plugins/by-name/nvim-ufo/default.nix
GaetanLepage: Ok I see. So we do want the values given to `closeFoldKinds` to be forwarded to `settings.close_fold_kinds` then, and let the plugin warn the user about the deprecation. I have correctly understood ?

If so, your implem looks good.

GaetanLepage (approved):
LGTM !

khaneliman:
@mergify queue

PR #2370, opened 2024-10-06
plugins/auto-session: migrate to mkNeovimPlugin 

khaneliman:


plugins/by-name/auto-session/default.nix
```diff
@@ -1,121 +1,192 @@
...
+        ]
+      ) (settingsPath ++ [ "auto_restore" ]))
+
+      (lib.mkRenamedOptionModule (
+        basePluginPath
+        ++ [
+          "bypassSessionSaveFileTypes"
+        ]
+      ) (settingsPath ++ [ "bypass_save_filetypes" ]))
+      (lib.mkRenamedOptionModule (nestedAutoSessionPluginPath ++ [ "enableLastSession" ]) (
```
GaetanLepage, line 73: Those can't work simply with `optionsRenamedToSettings` ?

khaneliman: all of these are changed from a weird nested path to a non nested path for different names. when i was reviewing the documentation it looked like these names matched the https://github.com/rmagatti/auto-session/blob/9d02776ed42874d37869dc683396234e3724b52d/lua/auto-session/config.lua#L120-L151 global variable config names so I'm renaming then to match the https://github.com/rmagatti/auto-session/blob/9d02776ed42874d37869dc683396234e3724b52d/lua/auto-session/config.lua#L63-L107 setup table names

plugins/by-name/auto-session/default.nix
```diff
@@ -1,121 +1,192 @@
...
-{
-  options.plugins.auto-session = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "auto-session";
-
-    package = lib.mkPackageOption pkgs "auto-session" {
-      default = [
-        "vimPlugins"
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "auto-session";
+  originalName = "auto-session";
```
GaetanLepage, line 11:
```suggestion
```
No need as this will be set as is by default.

plugins/by-name/auto-session/default.nix
GaetanLepage: Oh right, I read to fast.
Maybe factorizing this with a function could have been a but cleaner, but this is definitely not a blocker.

GaetanLepage (approved):
LGTM overall

khaneliman:
@mergify queue

PR #2393, opened 2024-10-08
update-scripts: Add a script to search for nvim-lspconfig server packages

traxys:
This script generates a list on stdout of possible matches for nvim-lspconfig servers.
It can't easily generate something more advanced because a lot of lsp servers have false positives, for example groovy's default command is `java -jar .....`, this would match the `java` package, not the groovy language server.

### DIFF
--- a/flake-modules/dev/devshell.nix
+++ b/flake-modules/dev/devshell.nix
@@ -111,6 +111,11 @@
               command = ''${pkgs.python3.interpreter} ${./list-plugins.py} "$@"'';
               help = "List plugins and get implementation infos";
             }
+            {
+              name = "locate-lsp-packages";
+              command = ''${pkgs.python3.interpreter} ${./locate-lsp-packages.py}'';
+              help = "Locate (with nix-index) LSP servers in nixpkgs";
+            }
           ];
       };
     };
--- a/flake-modules/dev/locate-lsp-packages.py
+++ b/flake-modules/dev/locate-lsp-packages.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+# This script requires nix-locate
+
+import json
+import os
+import subprocess
+from dataclasses import dataclass
+
+
+def find_project_root(root_identifier: str) -> None | str:
+    current_path = os.getcwd()
+    while True:
+        if root_identifier in os.listdir(current_path):
+            return current_path
+        parent_path = os.path.dirname(current_path)
+        if parent_path == current_path:
+            return None
+        os.chdir("..")
+        current_path = os.getcwd()
+
+
+@dataclass
+class CustomCommand:
+    package: str
+    cmd: list[str]
+
+
+@dataclass
+class PackageList:
+    unpackaged: list[str]
+    packages: dict[str, str | list[str]]
+    custom_cmd: dict[str, CustomCommand]
+
+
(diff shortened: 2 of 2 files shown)


update-scripts/nvim-lspconfig/locate-lsp-packages.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+# This script requires nix-locate
```
khaneliman, line 2: should this be a nix shebang with nix-locate included?

flake-modules/dev/devshell.nix
```diff
@@ -13,6 +13,7 @@
     }:
     lib.optionalAttrs (inputs.devshell ? flakeModule) {
       devshells.default = {
+        packages = [ pkgs.python3 ];
```
khaneliman, line 16: or included in here?

update-scripts/nvim-lspconfig/locate-lsp-packages.py
traxys: I'm not sure, I think nix-locate needs a nix-index database to be installzd

khaneliman: Oh, that sounds right... I think that's why I ended up adding nix-index-database to my flake.

traxys: I could add a dependency on https://github.com/nix-community/nix-index-database for this, but it's maybe a bit heavy?

khaneliman: Yeah, I don't know if it's worth an extra input.

khaneliman:
@traxys haven't tested yet, but will this be used for the update that generates the json https://github.com/nix-community/nixvim/pull/2399 we use ?

traxys:
Well this is the use of this script, to help search for the corresponding nixkpkgs package

flake-modules/dev/devshell.nix
MattSturgeon, line 16: Any reason why you're adding python to the devshell instead of packaging your script as a derivation (like we normally do)?

update-scripts/nvim-lspconfig/locate-lsp-packages.py
```diff
@@ -0,0 +1,66 @@
...
+
+import json
+import os
+import subprocess
+
+repo = os.path.realpath(os.path.dirname(__file__) + "/../..")
```
MattSturgeon, line 8: Maybe this is why? I think there'd still be work-arounds, such as providing the relevant files at the CLI or via env vars. These could then be made available to the script via a wrapper or a patch/fixup phase in the drv builder.

MattSturgeon (approved):
Don't see any issues, though I don't fully understand all design decisions.

MattSturgeon:
@traxys friendly bump :slightly_smiling_face:

khaneliman:
@traxys Were you still interested in getting this added? Would be helpful to have it available, in source, the next time we have lsp changes.

khaneliman (approved):
LGTM as a good starting point. We can tweak anything we find while using it

traxys:
@mergify queue

PR #2394, opened 2024-10-08
 options-search: configure title 

MarcelCoding:
I've just added a parameter to the search that allows configuring the title. I hope the suggested one is what you like.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -72,6 +72,7 @@ in
       nuschtosSearch.packages.mkSearch {
         optionsJSON = options-json + "/share/doc/nixos/options.json";
         urlPrefix = "https://github.com/nix-community/nixvim/tree/main";
+        title = "Nixvim options search";
         inherit baseHref;
       };
   in


docs/default.nix
```diff
@@ -72,6 +72,7 @@ in
       nuschtosSearch.packages.mkSearch {
         optionsJSON = options-json + "/share/doc/nixos/options.json";
         urlPrefix = "https://github.com/nix-community/nixvim/tree/main";
+        title = "NixVim Options Search";
```
GaetanLepage, line 75:
```suggestion
        title = "Nixvim options search";
```

Maybe this is more natural. I have no strong feeling though.

GaetanLepage (changes requested):
Thanks for the change, this is a good idea.

Could you please get rid of the `flake.lock` update ? Is it necessary for this change ?

MarcelCoding:
I've just committed the change into the search repo, as I need to update the search input in order that the `title` parameter exists.

PS: I've only updated the search input and left all others untouched.

MarcelCoding:
(I've copied the *old* title from the current navigation sidebar.)
image

GaetanLepage:
> I've just committed the change into the search repo, as I need to update the search input in order that the `title` parameter exists.

Ok I see. I have updated the lockfile in the meantime, so you should be able to rebase safely now.

MarcelCoding:
done

GaetanLepage (approved):
LGTM ! Thanks for downstreaming this change to nixvim :)

GaetanLepage:
@mergify queue

Issue #2396, opened 2024-10-09
[Request] Treesitter Lua queries for Nix files


blackxored:
Given the amount of code people write in Lua, I think it would be great to include treesitter queries that automatically highlight lua code in `extraConfigLua` and friends, just started with nixvim today and is already hard to read the Lua code without syntax highlighting.

khaneliman:
This is available already through our `plugins.treesitter.nixvimInjections`.

MattSturgeon:
In addition to the [above](https://github.com/nix-community/nixvim/issues/2396#issuecomment-2401177310), there's also [hmts.nvim](https://github.com/calops/hmts.nvim?tab=readme-ov-file) which can be enabled via `plugins.hmts.enable`. This adds ts queries for several home-manager options and can be used alongside `plugins.treesitter.nixvimInjections`.

PR #2397, opened 2024-10-09
plugins/treesitter: update description

khaneliman:
Added some more upfront details about the plugin's default settings. Lots of people assume just enabling the plugin will give them the behaviors of highlighting they expect, but the plugin does not default this to be the case.

image

Resolves https://github.com/nix-community/nixvim/issues/2396

### DIFF
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -15,6 +15,12 @@ helpers.neovim-plugin.mkNeovimPlugin {
   description = ''
     Provides an interface to [tree-sitter]
 
+    > [!NOTE]
+    > This plugin defaults to all functionality disabled.
+    >
+    > Please explicitly enable the features you would like to use in `plugins.treesitter.settings`.
+    > For example, to enable syntax highlighting use the `plugins.treesitter.settings.highlight.enable` option.
+
     ### Installing tree-sitter grammars from Nixpkgs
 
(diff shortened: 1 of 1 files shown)


plugins/by-name/treesitter/default.nix
```diff
@@ -15,6 +15,11 @@ helpers.neovim-plugin.mkNeovimPlugin {
   description = ''
     Provides an interface to [tree-sitter]
 
+    > [!NOTE]
+    > This plugin defaults to all functionality disabled.
+    >
+    > Please explicitly enable the features you would like to use in `plugins.treesitter.settings`.
```
MattSturgeon, line 21: Maybe we should give a concrete example in this note?

```md
> For example, to enable syntax highlighting use the `plugins.treesitter.settings.highlight.enable` option.
```

MattSturgeon (approved):
LGTM! Not sure how it _resolves_ #2396 specifically, but it's definitely related and that issue can certainly be closed as more of a question than an issue.

khaneliman:
> LGTM! Not sure how it _resolves_ #2396 specifically, but it's definitely related and that issue can certainly be closed as more of a question than an issue.

Yeah, this issue has been addressed multiple times and I just keep trying to make the default module behavior more obvious each time.. lol

khaneliman:
@mergify queue

PR #2402, opened 2024-10-10
tests/lsp-servers: de-couple from grouped tests

MattSturgeon:
Follow up to cab6b0c9febfecd96e4cda3dfe4ec8675f6d9ddb from #2365, which intended to ungroup the "all servers" lsp test for performance reasons. This does that more explicitly, by handling the test in the same way as other "loose" tests outside of `tests/test-sources`.


khaneliman (approved):

MattSturgeon:
@mergify queue

PR #2401, opened 2024-10-10
tests/generated: validate declared lsp packages

MattSturgeon:
Follow up to #2365

PR #2403, opened 2024-10-10
plugins/lsp: use a no-default option when there is no default provided

MattSturgeon:
As per https://github.com/nix-community/nixvim/pull/2365#discussion_r1793846911

### DIFF
--- a/plugins/lsp/language-servers/_mk-lsp.nix
+++ b/plugins/lsp/language-servers/_mk-lsp.nix
@@ -8,7 +8,7 @@
   name,
   description ? "Enable ${name}.",
   serverName ? name,
-  package ? name,
+  package ? null,
   url ? null,
   cmd ? (cfg: null),
   settings ? (cfg: cfg),
@@ -50,16 +50,32 @@ in
       package =
         if lib.isOption package then
           package
-        else
+        else if args ? package then
           lib.mkPackageOption pkgs name {
             nullable = true;
             default = package;
+          }
+        else
+          # If we're not provided a package, we should provide a no-default option
+          lib.mkOption {
+            type = types.nullOr types.package;
+            description = ''
+              The package to use for ${name}. Has no default, but can be set to null.
+            '';
           };
 
       cmd = mkOption {
         type = with types; nullOr (listOf str);
         default =
-          if (cfg.package or null) != null then if builtins.isFunction cmd then cmd cfg else cmd else null;
+          # TODO: do we really only want the default `cmd` when `package` is non-null?
+          if !(opt.package.isDefined or false) then
+            null
+          else if cfg.package == null then
+            null
+          else if builtins.isFunction cmd then
+            cmd cfg
+          else
+            cmd;
         description = ''
           A list where each entry corresponds to the blankspace delimited part of the command that
           launches the server.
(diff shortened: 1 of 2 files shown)


khaneliman (approved):

plugins/lsp/language-servers/default.nix
```diff
@@ -179,18 +179,6 @@ let
...
   lspPackages = import ../lsp-packages.nix;
 
-  getLspPackage =
-    name:
-    if lib.hasAttr name lspPackages.packages then
-      { package = lspPackages.packages.${name}; }
-    else if lib.hasAttr name lspPackages.customCmd then
-      { inherit (lspPackages.customCmd.${name}) package cmd; }
-    else
-      assert lib.assertMsg (lib.elem name lspPackages.unpackaged) "LSP server ${name} is unknown";
```
MattSturgeon, line 189: We should probably wait for #2401 before dropping this assertion.

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -50,10 +50,18 @@ in
       package =
         if lib.isOption package then
           package
-        else
+        else if args ? package then
```
MattSturgeon, line 53:
```suggestion
        else if args ? ${package} then
```

MattSturgeon: (nevermind, I'm dumb... It was correct before :joy:)

traxys (approved):

khaneliman:
Hmm eval failing with those accessed but not defined errors

traxys:
Ah this requires adaptations in the tests to pass null to the package it seems like

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -50,16 +50,32 @@ in
...
-          if (cfg.package or null) != null then if builtins.isFunction cmd then cmd cfg else cmd else null;
+          # TODO: do we really only want the default `cmd` when `package` is non-null?
+          if !(opt.package.isDefined or false) then
+            null
+          else if cfg.package == null then
+            null
+          else if builtins.isFunction cmd then
+            cmd cfg
+          else
+            cmd;
```
MattSturgeon, line 78: I don't want to address this in this PR, but I've added a TODO comment regardless.

Essentially, `plugins.lsp.servers.<name>.cmd` is eventually passed through to `plugins.lsp.enabledServers.*.extraOptions`, so in a way it is a freeform sub-option.

It also seems that whether `cfg.package` is null or not is unrelated to whether we should use the default cmd.

Perhaps this comes from our history of defining a command based on things like `lib.getExe cfg.package`? Hopefully we've moved away from doing that and can enforce a policy where default `cmd` values do not reference the nix-store and are always defined.

If we keep the current behavior, we should also have a `defaultText` that explains how the default is conditional.

MattSturgeon:
> Hmm eval failing with those accessed but not defined errors

> Ah this requires adaptations in the tests to pass null to the package it seems like

Should all be resolved now @khaneliman @traxys

khaneliman (approved):

MattSturgeon:
@mergify queue

Issue #2398, opened 2024-10-09
[BUG] <BUG> Trouble's auto_open throws error and does not auto open


whoisthis0192837465:
| Field        | Description |
|--------------|-------------|
| Plugin       | `Trouble`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

I tried to enable auto open in trouble, but then an error was thrown at every neovim startup

You specified `auto_open = true` in your global config.
This is probably not what you want.
Add it to the mode you want to auto open instead.
```lua
opts = {
  modes = {
    diagnostics = { auto_open = true },
  }
}
```
Disabling global `auto_open`.

## Minimal, Reproducible Example (MRE)

```nix
{ ... }: {
  programs.nixvim = {
    enable = true;
    plugins = {
      trouble = {
        enable = true;
        settings = {
          auto_close = true;
          auto_open = true;
        };
      };
    };
  };
}
```

khaneliman:
Ah, I do see the plugin throws that message when it detects that... https://github.com/folke/trouble.nvim/blob/254145ffd528b98eb20be894338e2d5c93fa02c2/lua/trouble/config/init.lua#L223-L234 We will need to update the settings to redirect... I wonder if that was the previous behavior and we just never updated with new trouble version.

khaneliman:
Yeah, looks like in 2.X they had everything global level. Thanks for reporting, will update when I get a chance.

PR #2407, opened 2024-10-11
plugins/trouble: v3 cleanup

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2398

Looks like there was a rewrite of the plugin recently that we didn't update the module to properly account for in documentation. Cleaning up the documented options and updating test cases.

### DIFF
--- a/plugins/by-name/trouble/default.nix
+++ b/plugins/by-name/trouble/default.nix
@@ -115,186 +115,267 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
   ];
 
   settingsOptions = {
-    position =
-      defaultNullOpts.mkEnum
-        [
-          "top"
-          "left"
-          "right"
-          "bottom"
-        ]
-        "bottom"
-        ''
-          Position of the list.
-        '';
-
-    height = defaultNullOpts.mkInt 10 ''
-      Height of the trouble list when position is top or bottom.
-    '';
-
-    width = defaultNullOpts.mkInt 50 ''
-      Width of the list when position is left or right.
-    '';
-
-    icons = defaultNullOpts.mkBool true "Use devicons for filenames";
-
-    mode = defaultNullOpts.mkEnum [
-      "workspace_diagnostics"
-      "document_diagnostics"
-      "quickfix"
-      "lsp_references"
-      "loclist"
-    ] "workspace_diagnostics" "Mode for default list";
-
-    fold_open = defaultNullOpts.mkStr "" "Icon used for open folds";
-
-    fold_closed = defaultNullOpts.mkStr "" "Icon used for closed folds";
-
-    group = defaultNullOpts.mkBool true "Group results by file";
-
-    padding = defaultNullOpts.mkBool true "Add an extra new line on top of the list";
-
-    cycle_results = defaultNullOpts.mkBool true "Whether to cycle item list when reaching beginning or end of list";
-
-    action_keys =
-      lib.mapAttrs
-        (
-          action: config:
-          defaultNullOpts.mkNullable (
-            with lib.types; either str (listOf str)
-          ) config.default config.description
-        )
-        {
-          close = {
-            default = "q";
-            description = "Close the list";
-          };
-          cancel = {
-            default = "<esc>";
-            description = "Cancel the preview and get back to your last window / buffer / cursor";
-          };
-          refresh = {
-            default = "r";
-            description = "Manually refresh";
-          };
-          jump = {
-            default = [
-              "<cr>"
-              "<tab>"
-            ];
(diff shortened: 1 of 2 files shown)


plugins/by-name/trouble/default.nix
```diff
@@ -115,200 +115,248 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+        Property = " ";
+        String = " ";
+        Struct = "󰆼 ";
+        TypeParameter = " ";
+        Variable = "󰀫 ";
+      };
+    } "Define custom icons.";
   };
 
   extraConfig = cfg: {
```
MattSturgeon, line 352: Do you want to warn when the old settings attrs are present in `cfg.settings`?

plugins/by-name/trouble/default.nix
```diff
@@ -115,200 +115,248 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
-      Height of the trouble list when position is top or bottom.
+    auto_close = defaultNullOpts.mkBool false ''
+      Automatically close the list when you have no diagnostics.
     '';
 
-    width = helpers.defaultNullOpts.mkInt 50 ''
-      Width of the list when position is left or right.
+    auto_preview = defaultNullOpts.mkBool true ''
+      Automatically preview the location of the diagnostic.
+      <esc> to close preview and go back to last window.
```
MattSturgeon, line 124: I think the `<esc>` may be stripped by nixos-render-docs as a potential html tag.

We probably want `` `<esc>` `` anyway though.

MattSturgeon (commented):
Don't see any major issues 😁

plugins/by-name/trouble/default.nix
khaneliman: Thinking just a warning about deprecation/removal or mapping them to a mkRemovedOption for an actual assertion?

MattSturgeon: I don't think renames are appropriate for freeform settings options.

I'm fine with doing nothing, but if you want to do something I'd suggest just a warning if x, y, or z attrs are present under `cfg.settings`.

I think you did something similar in another PR recently?

khaneliman: > I don't think renames are appropriate for freeform settings options.
> 
> I'm fine with doing nothing, but if you want to do something I'd suggest just a warning if x, y, or z attrs are present under `cfg.settings`.
> 
> I think you did something similar in another PR recently?

Yeah, looks like it was the rest.nvim v3 migration, I can do that again probably

khaneliman: Heading out the door for a little family time for a bit, I pushed a quick implementation but I'll double check the list of fields when I get back later before merge.

plugins/by-name/trouble/default.nix
```diff
@@ -115,200 +115,274 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+        Variable = "󰀫 ";
+      };
+    } "Define custom icons.";
   };
 
   extraConfig = cfg: {
+    # TODO: introduced 2024-10-12: remove after 24.11
+    warnings =
+      let
+        definedOpts = lib.filter (opt: lib.hasAttrByPath (lib.toList opt) cfg.settings) [
```
MattSturgeon, line 356: This is more general/future-proof, and maybe we'll end up factoring it out into `lib.deprecation` if it becomes a common pattern.

But if you wanted to, we can simplify here because none of _these_ are nested attr paths:

```nix
definedOpts = lib.filter (opt: cfg.settings ? ${opt}) [
```

or
```nix
definedOpts = lib.filter (lib.flip lib.hasAttr cfg.settings) [
```

plugins/by-name/trouble/default.nix
```diff
@@ -115,200 +115,274 @@ helpers.neovim-plugin.mkNeovimPlugin {
...
+          "indent_lines"
+          "win_config"
+          "include_declaration"
+          "signs"
+          "use_diagnostic_signs"
+        ];
+      in
+      lib.optional (definedOpts != [ ]) ''
+        Nixvim(plugins.trouble): The following v2 settings options are no longer supported in v3:
+        ${lib.concatMapStringsSep "\n" (opt: "  - ${lib.showOption (lib.toList opt)}") definedOpts}
```
MattSturgeon, line 376: Which allows simplifying this too:
```suggestion
        ${lib.concatMapStringsSep "\n" (opt: "  - ${opt}") definedOpts}
```

MattSturgeon (approved):
Looking over the diff, it seems you've covered everything. :rocket:

khaneliman:
@mergify queue

PR #2414, opened 2024-10-13
modules/output: obtain plugin configs from wrapped neovim

stasjok:
This:
* restores modules/output generated init.lua tests removed in 3d1224a039d70098c78d263b0867352d95f2194b, but without IFD.
* changes where generated plugin configs are obtained. Instead of using the result of `makeNeovimConfig`, generated config is obtained from `wrappedNeovim.initRc`. This should avoid breaking after NixOS/nixpkgs#344541 is merged. See also conversation in #2343.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -285,8 +285,10 @@ in
         }
       );
 
+      wrappedNeovim' = pkgs.wrapNeovimUnstable package neovimConfig;
+
       customRC = helpers.concatNonEmptyLines [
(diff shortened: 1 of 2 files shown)


GaetanLepage (approved):
Thanks for taking care of this !

Diff LGTM but I'm not that familiar with this part of the codebase.
@nix-community/nixvim could you guys have a quick look ?

GaetanLepage:
https://github.com/NixOS/nixpkgs/pull/344541 merged !

traxys (approved):

traxys:
@mergify queue

PR #2416, opened 2024-10-13
plugins/lsp/hls: handle automatic installation of required GHC

### LINKED ISSUE #2413: [BUG] Haskell Language Server needs GHC
| Plugin       | `lsp`  | 
| Nixpkgs      | `unstable` | 

## Description

Just like `rust-analyzer` depends on `cargo` and `rustc`. `haskell-language-server` needs to be installed so it can work.
A solution could be installing it with `haskell-language-server`, or just like `rust-analyzer`, it could be an option (which I think should be better)

image

```nix
programs.nixvim = {
  lsp.enable = true;
  lsp.servers.hls = {
    enable = true;
    filetypes = [ "cabal" "haskell" "lhaskell" ];
  };
}
```

GaetanLepage:
Ensures `lsp.servers.hls` (haskell language server) users make an explicit choice regarding whether the required GHC dependency is installed or not.
This is the same approach as in `lsp.servers.rust_analyzer`.

Fixed #2413

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -220,6 +220,7 @@ in
     ++ [
       ./ccls.nix
       ./efmls-configs.nix
+      ./hls.nix
       ./pylsp.nix
       ./rust-analyzer.nix
       ./svelte.nix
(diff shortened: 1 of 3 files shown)


traxys (approved):

GaetanLepage:
@mergify queue

HeitorAugustoLN:
Thanks!

PR #2430, opened 2024-10-17
modules/nixpkgs: add `overlays` option

MattSturgeon:
- **modules/nixpkgs: restructure `nixpkgs.pkgs.isDefined` assertion**
- **modules/nixpkgs: add `overlays` option**

Follow up to #2301 and working towards #1784; this PR adds the `nixpkgs.overlays` option.

This change is supported by some related refactoring to the existing assertion to a) reduce future indentation changes to the module and b) ensure the assertion message is actually shown to users. Without this change, `_module.args.pkgs` being absent will cause other errors that get thrown before we are able to check the module assertions.

Additionally, I've added a test to ensure the new option works correctly.

I would like a better option example for `nixpkgs.overlays`. Currently I'm using the same example as NixOS. Perhaps we could do something using the `vimPlugins = prev.vimPlugins.extend` pattern? IDK, struggling to find "useful" real-world examples that are small and obvious enough to be a good example.

### DIFF
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -55,6 +55,76 @@ in
         > Use this option with care.
       '';
     };
+
+    overlays = lib.mkOption {
+      type =
+        let
+          overlayType = lib.mkOptionType {
+            name = "nixpkgs-overlay";
+            description = "nixpkgs overlay";
+            check = lib.isFunction;
+            merge = lib.mergeOneOption;
+          };
+        in
+        lib.types.listOf overlayType;
+      default = [ ];
+      # First example from https://nixos.org/manual/nixpkgs/unstable/#what-if-your-favourite-vim-plugin-isnt-already-packaged
+      # Second example from https://github.com/nix-community/nixvim/pull/2430#discussion_r1805700738
+      # Third example from https://github.com/atimofeev/nixos-config/blob/0b1c1c47c4359d6a2aa9a5eeecb32fa89ad08c88/overlays/neovim-unwrapped.nix
+      example = lib.literalExpression ''
+        [
+
+          # Add a vim plugin that isn't packaged in nixpkgs
+          (final: prev: {
+            easygrep = final.vimUtils.buildVimPlugin {
+              name = "vim-easygrep";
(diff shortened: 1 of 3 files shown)


modules/top-level/nixpkgs.nix
```diff
@@ -55,31 +55,71 @@ in
...
+      # FIXME: use an example that is topical for vim
+      example = lib.literalExpression ''
+        [
+          (self: super: {
+            openssh = super.openssh.override {
+              hpnSupport = true;
+              kerberos = self.libkrb5;
+            };
+          })
+        ]
```
MattSturgeon, line 80: Any ideas for a better vim-related example?

flake-modules/tests.nix
```diff
@@ -52,6 +72,6 @@
         package-options = pkgs.callPackage ../tests/package-options.nix { inherit nixvimConfiguration; };
 
         lsp-all-servers = pkgs.callPackage ../tests/lsp-servers.nix { inherit nixvimConfiguration; };
-      } // import ../tests { inherit pkgs pkgsUnfree helpers; };
+      } // callTests ../tests { };
```
MattSturgeon, line 75: Should this commit go further and incorporate the full changes from [`01ef05e` (#2104)](https://github.com/nix-community/nixvim/pull/2104/commits/01ef05e140731acd9a81b2a70a67259aa140b411)?

Maybe I'll pull this out into another PR, but if I do it'll block this PR.

modules/top-level/nixpkgs.nix
khaneliman: Was wondering if we could do one that adds a tree-sitter-X grammar to the treesitter package ?

khaneliman: overlays = [ inputs.neovim-nightly-overlay.overlays.default ];
?

flake-modules/tests.nix
MattSturgeon: I've opened #2433 to tackle this refactoring separately.

khaneliman (approved):
LGTM

MattSturgeon:
@mergify queue

PR #2433, opened 2024-10-18
tests: refactor to use `callTest(s)` pattern + move calls to `tests/default.nix`

MattSturgeon:
- **tests: move test derivations to `tests/default.nix`**
- **tests: use `callTest` pattern**
- **tests: cleanup tests to better use `callTest` pattern**

This takes ideas from #2104 (reducing the scope of that PR) and is required for #2430, as otherwise the auto-formatting leads to an unnecessarily large diff.

This PR contains two key ideas in its refactoring:
- use the `callPackage` pattern with a nixvim-specific `callTest` variant
- rename `tests/default.nix` -> `tests/main.nix` since there are other tests in `tests/` not "owned" by `default.nix`
- move the builk of the implementation in `flake-modules/test.nix` to a new `tests/default.nix`, so that we minimize the implementation details contained within `flake-modules` and allow `tests/` to take responsibility for all its test derivations.

### DIFF
--- a/tests/main.nix
+++ b/tests/main.nix
@@ -1,0 +1,66 @@
+# Collects the various test modules in tests/test-sources/ and groups them into a number of test derivations
+{
+  lib ? pkgs.lib,
+  helpers,
+  pkgs,
+  pkgsUnfree,
+}:
+let
+  fetchTests = import ./fetch-tests.nix { inherit lib pkgs helpers; };
+  test-derivation = import ../lib/tests.nix { inherit pkgs lib; };
+  inherit (test-derivation) mkTestDerivationFromNixvimModule;
+
+  moduleToTest =
+    file: name: module:
+    mkTestDerivationFromNixvimModule {
+      inherit name;
+      module = {
+        _file = file;
+        imports = [ module ];
+      };
+      pkgs = pkgsUnfree;
+    };
+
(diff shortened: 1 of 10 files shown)


tests/main.nix
```diff
@@ -0,0 +1,69 @@
+# Collects the various test modules in tests/test-sources/ and groups them into a number of test derivations
```
MattSturgeon, line 1: Is there a better name than `main.nix`?

- `main.nix` could be confusing, as `default.nix` is the "main" entrypoint to a directory
- `modules.nix` could make sense (the file is responsible for making tests from the modules in `test-sources`), but we already have a `modules.nix` that tests the wrapper modules...
- `generated.nix` could make sence too, but we already have a `generated.nix` test that tests the _ci-generated files_ are valid

Maybe `modules.nix` is the way to go, if we renamed the current `modules.nix` to `wrappers.nix`? But this is making the diff much larger than it really should be for a minor refactoring PR...

GaetanLepage (approved):

tests/main.nix
GaetanLepage: I agree `modules.nix` would be better. We can do it in another PR though.

MattSturgeon:
@mergify queue

Issue #1621, opened 2024-06-02
[FEATURE REQUEST] `vimdiffAlias` support for all wrappers


MattSturgeon:
https://github.com/nix-community/nixvim/pull/1353 added a `vimdiffAlias` support to the home-manager wrapper for nixvim, but it'd be nice to have support in all wrappers.

Better yet, we could have support upstream in nixpkgs, since that is where `vimAlias` and `viAlias` is implemented.

## Extra Information

https://github.com/nix-community/nixvim/pull/1353 uses a shell alias, but the nixpkgs implementations use symlinks on the PATH.

We would need to use `pkgs.wrapProgram`, `pkgs.makeWrapper`, or similar on platforms where we can't make shell aliases (or all platforms).

PR #2441, opened 2024-10-21
docs: refactor wrapper-options docs

MattSturgeon:
Refactor the previous `hm.md` docs to cover platform-specific options for _all_ wrappers.

#### Future work

In the future, we could factor out "common" options, but I think that is best done if we end up with lots of them. For now it is probably simpler to keep things as-is; i.e. common options are displayed multiple times.

We probably also want the hm-specific options to be available on all platforms. This may be best achieved by upstreaming `vimdiff` alias to the nixpkgs vim wrapper, as discussed in https://github.com/nix-community/nixvim/pull/1353 (tracked in https://github.com/nix-community/nixvim/issues/1621).

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -4,7 +4,6 @@
   evaledModules,
   nixosOptionsDoc,
   transformOptions,
-  hmOptions,
   search,
 }:
 let
@@ -227,6 +226,29 @@ let
     ) docs.modules;
   };
 
+  wrapperOptions =
+    lib.genAttrs
+      [
+        "hm"
+        "nixos"
+        "darwin"
+      ]
+      (
+        name:
+        let
+          configuration = lib.evalModules {
+            modules = [
+              ../../wrappers/modules/${name}.nix
+              {
+                # Ignore definitions for missing options
+                _module.check = false;
+              }
+            ];
+          };
+        in
+        removeUnwanted configuration.options
+      );
+
   prepareMD = ''
     # Copy inputs into the build directory
     cp -r --no-preserve=all $inputs/* ./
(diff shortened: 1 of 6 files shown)


docs/mdbook/default.nix
```diff
@@ -243,8 +265,10 @@ let
     substituteInPlace ./SUMMARY.md \
       --replace-fail "@NIXVIM_OPTIONS@" "$(cat ${pkgs.writeText "nixvim-options-summary.md" mdbook.nixvimOptions})"
 
-    substituteInPlace ./modules/hm.md \
-      --replace-fail "@HM_OPTIONS@" "$(cat ${mkMDDoc hmOptions})"
+    substituteInPlace ./modules/wrapper-options.md \
+      --replace-fail "@NIXOS_OPTIONS@" "$(cat ${mkMDDoc wrapperOptions.nixos})" \
+      --replace-fail "@HM_OPTIONS@" "$(cat ${mkMDDoc wrapperOptions.hm})" \
+      --replace-fail "@DARWIN_OPTIONS@" "$(cat ${mkMDDoc wrapperOptions.darwin})"
```
MattSturgeon, line 271: Maybe we should be creating a separate file for each wrapper here, rather than substituting them into a single template file...

Something like
```nix
foldAttrs (
  str: name: options:
  str + "\n install ${mkMDDoc options} ./modules/${name}.md"
) "" wrapperOptions
```

MattSturgeon: Up to you if this blocks the PR

khaneliman: Do you want to just split them out to be separate documentation pages instead of being shown on a single page? It might be cleaner than having them smushed together on the same page right now with the formatting.

MattSturgeon: Yeah. I think we should do it this way, and perhaps also set a `meta.docs` in each module too if we want a pre-ample header.

Perhaps that should follow in another PR to keep this one simple.

khaneliman: It's kinda hard at a glance to see where one module beings and another ends with the current page. I agree we could do it cleaner but this at least shows all the different available.

khaneliman (approved):

MattSturgeon:
@mergify queue

MattSturgeon:
@mergify refresh

MattSturgeon:
@mergify queue

PR #2445, opened 2024-10-22
{docs,wrappers}: refactor to prepare for separated wrapper-option doc pages

MattSturgeon:
- ~~**docs: move `user-guide` and `modules` into `mdbood` dir**~~
- **docs: simplify `mdbook.nixvimOptions` impl**
- **wrappers/modules: move `enable` to `shared.nix`**
- **wrappers: add `meta.wrappers` options to wrappers**

I've moved `modules/` to `wrappers/` in docs URLs. `modules/` felt confusing, but I'm not convinced `wrappers/` is right either. Maybe `platforms/` would be better? If we're gonna change the URLs let's try and get it right first time, or postpone the change for later.

I've introduced a `meta.wrappers.name` option, that is only present on wrapper modules (i.e. not currently in standalone). If standalone ever wanted to be part of the "wrapper" docs maybe it should also get the option, but that's easily refactored.
This option is used to set the "human readable" name of the wrapper module, as it should appear in documentation (headings, links, etc).

I'm unsure currently if it will be best to a) show "common" options shared by all non-standalone wrappers together and then unique options separately, or b) continue showing all options per wrapper, but split the docs into several pages.

Perhaps structured as a nested tree:
- Platforms (index page with summary): `wrappers/index.md`
  - NixOS: `wrappers/nixos.md`
  - home-manager: `wrappers/hm.md`
  - nix-darwin: `wrappers/darwin.md`

While feedback on next steps is welcome, and _may_ influence this PR, currently I think it best to actually implement it in a follow up PR. (We'll see)...

Additionally, for context: the new `shared.nix` wrapper module will likely house the option declaration for the proposed `nixpkgs.useGlobalPackages` option. This option is where we'll handle gradually deprecating the current behaviour of inheriting the host module's `pkgs` into our module eval.

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -224,6 +224,35 @@ let
       in
       "${padding}- [${name}](${path})"
     ) docs.modules;
+
+    wrapperOptionFiles = lib.mapAttrs' (name: options: {
+      name = options.meta.wrapper.name.value;
+      # TODO:
+      # value.path = "./modules/${name}.md";
+      value.file = mkMDDoc options;
+    }) wrapperOptions;
+
+    wrapperOptionDocs =
+      pkgs.runCommandNoCCLocal "wrapper-option-doc"
+        {
+          __structuredAttrs = true;
+          files = lib.mapAttrs (name: value: value.file) mdbook.wrapperOptionFiles;
+          # TODO:
+          # paths = lib.mapAttrs (name: value: value.path) mdbook.wrapperOptionFiles;
+        }
+        ''
+          for name in "''${!files[@]}"
+          do
+            # $file contains the docs built by mkMDDoc
+            file="''${files[$name]}"
+            # TODO: consider putting wrapper docs in separate files:
+            # path="''${paths[$name]}"
+            echo >> "$out"
+            echo "## $name" >> "$out"
+            echo >> "$out"
+            cat "$file" >> $out
+          done
+        '';
   };
 
   wrapperOptions =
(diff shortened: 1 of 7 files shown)


docs/mdbook/default.nix
```diff
@@ -218,12 +218,41 @@ let
...
+      value.file = mkMDDoc options;
+    }) wrapperOptions;
+
+    wrapperOptionDocs =
+      pkgs.runCommandNoCCLocal "wrapper-option-doc"
+        {
+          __structuredAttrs = true;
+          files = lib.mapAttrs (name: value: value.file) mdbook.wrapperOptionFiles;
+          # TODO:
+          # paths = lib.mapAttrs (name: value: value.path) mdbook.wrapperOptionFiles;
```
GaetanLepage, line 241:
```suggestion
          files = lib.mapAttrs (_: value: value.file) mdbook.wrapperOptionFiles;
          # TODO:
          # paths = lib.mapAttrs (_: value: value.path) mdbook.wrapperOptionFiles;
```

GaetanLepage (approved):
Apart from my minor comment (that you are free to ignore), this PR looks fine to me.
It feels less hacky at least.

MattSturgeon:
FYI: I spotted locally that the man docs also use the `user-guide` files, so I'll revert the change moving them into `docs/mdbook`

https://github.com/nix-community/nixvim/blob/0562e519ec0e69125c5edc917d41bccb54a534fd/docs/man/default.nix#L38-L44

For simplicity I'll probably also drop the `modules`->`wrappers` rename for now.

There's also some errors I need to resolve.

docs/mdbook/default.nix
GaetanLepage: No you are right, this is actually better.

PR #2447, opened 2024-10-23
plugins/typescript-tools: migrate to mkNeovimPlugin

khaneliman:
Plugin already had a settings option that can't take advantage of rename module option functionality, so we're using the deprecated settings options warning behavior introduced in previous plugin version upgrades to alert the user when they are using options that should be moved.

### DIFF
--- a/plugins/by-name/typescript-tools/default.nix
+++ b/plugins/by-name/typescript-tools/default.nix
@@ -1,24 +1,26 @@
 {
   lib,
-  pkgs,
-  config,
   ...
 }:
 let
   inherit (lib.nixvim) defaultNullOpts;
-  cfg = config.plugins.typescript-tools;
 in
-{
-  options.plugins.typescript-tools = {
-    enable = lib.mkEnableOption "typescript-tools";
-    package = lib.mkPackageOption pkgs "typescript-tools" {
-      default = [
-        "vimPlugins"
-        "typescript-tools-nvim"
-      ];
-    };
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "typescript-tools";
+  originalName = "typescript-tools.nvim";
+  package = "typescript-tools-nvim";
 
-    onAttach = defaultNullOpts.mkLuaFn "__lspOnAttach" "Lua code to run when tsserver attaches to a buffer.";
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  # TODO:: introduced 10-22-2024: remove after 24.11
+  optionsRenamedToSettings = [
+    "onAttach"
+    "handlers"
+  ];
+
+  settingsOptions = {
+    on_attach = defaultNullOpts.mkLuaFn "__lspOnAttach" "Lua code to run when tsserver attaches to a buffer.";
+
     handlers = lib.mkOption {
       type = with lib.types; nullOr (attrsOf strLuaFn);
       default = null;
@@ -34,9 +36,9 @@ in
     };
 
     settings = {
-      separateDiagnosticServer = defaultNullOpts.mkBool true "Spawns an additional tsserver instance to calculate diagnostics";
+      separate_diagnostic_server = defaultNullOpts.mkBool true "Spawns an additional tsserver instance to calculate diagnostics";
 
-      publishDiagnosticOn =
+      publish_diagnostic_on =
         defaultNullOpts.mkEnum
           [
             "change"
@@ -47,7 +49,7 @@ in
             Either "change" or "insert_leave". Determines when the client asks the server about diagnostics
           '';
 
-      exposeAsCodeAction = lib.mkOption {
+      expose_as_code_action = lib.mkOption {
         type =
           with lib.types;
           either (enum [ "all" ]) (
@@ -64,26 +66,26 @@ in
         description = "Specify what to expose as code actions.";
       };
 
-      tsserverPath = lib.nixvim.mkNullOrStr ''
+      tsserver_path = lib.nixvim.mkNullOrStr ''
         Specify a custom path to `tsserver.js` file, if this is nil or file under path
         doesn't exist then standard path resolution strategy is applied
       '';
 
-      tsserverPlugins =
+      tsserver_plugins =
         with lib.types;
         lib.nixvim.mkNullOrOption (listOf (maybeRaw str)) ''
           List of plugins for tsserver to load. See this plugins's README
           at https://github.com/pmizio/typescript-tools.nvim/#-styled-components-support
         '';
 
-      tsserverMaxMemory =
+      tsserver_max_memory =
         lib.nixvim.mkNullOrOption (with lib.types; maybeRaw (either ints.unsigned (enum [ "auto" ])))
           ''
             This value is passed to: https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes
             Memory limit in megabytes or "auto"(basically no limit)
           '';
 
-      tsserverFormatOptions = lib.mkOption {
+      tsserver_format_options = lib.mkOption {
         type = with lib.types; nullOr (attrsOf anything);
         default = null;
         description = "Configuration options that well be passed to the tsserver instance. Find available options [here](https://github.com/microsoft/TypeScript/blob/v5.0.4/src/server/protocol.ts#L3418)";
@@ -99,7 +101,7 @@ in
         };
       };
 
-      tsserverFilePreferences = lib.mkOption {
+      tsserver_file_references = lib.mkOption {
         type = with lib.types; nullOr (attrsOf anything);
         default = null;
         description = "Configuration options that well be passed to the tsserver instance. Find available options [here](https://github.com/microsoft/TypeScript/blob/v5.0.4/src/server/protocol.ts#L3439)";
@@ -113,15 +115,15 @@ in
         };
       };
 
-      tsserverLocale = defaultNullOpts.mkStr "en" ''
+      tsserver_locale = defaultNullOpts.mkStr "en" ''
         Locale of all tsserver messages. Supported locales here: https://github.com/microsoft/TypeScript/blob/3c221fc086be52b19801f6e8d82596d04607ede6/src/compiler/utilitiesPublic.ts#L620
       '';
 
-      completeFunctionCalls = defaultNullOpts.mkBool false ''
+      complete_function_calls = defaultNullOpts.mkBool false ''
         Mirror of VSCode's `typescript.suggest.completeFunctionCalls`
       '';
 
-      includeCompletionsWithInsertText = defaultNullOpts.mkBool true ''
+      include_completions_with_insert_text = defaultNullOpts.mkBool true ''
         Mirror of VSCode's `typescript.suggest.completeFunctionCalls`
       '';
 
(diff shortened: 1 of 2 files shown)


plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,43 @@ in
...
+          "disableMemberCodeLens"
+          [
+            "jsxCloseTag"
+            "enable"
+          ]
+          [
+            "jsxCloseTag"
+            "filetypes"
+          ]
+        ];
```
MattSturgeon, line 192: This would benefit from a) a "TODO: added; remove after" comment and b) some explanatory comments regarding why we can't just do a normal optionRenamedToSettings

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,43 @@ in
...
       in
-      ''
-        require('typescript-tools').setup(${helpers.toLuaObject options})
+      lib.optional (definedOpts != [ ]) ''
+        Nixvim(plugins.typescript-tools): The following settings have moved under `plugins.typescript-tools.settings.settings` with snake_case:
+        ${lib.concatMapStringsSep "\n" (opt: "  - ${lib.showOption (lib.toList opt)}") definedOpts}
       '';
```
MattSturgeon, line 197: In addition to this warning, it'd be nice to _try_ and have some migration aliases...

Defining `settings.settings` conditionally on whether or not one of the old attrs is present would be infinitely recursive, _however_ I think we can work around that by always defining `settings.settings`, and having the _content_ be dependent on whether the old attrs are in `cfg.settings`.

Something like:
```nix
# Psudonix
settings.settings = lib.optionalAttrs (definedOpts != [ ]) (
  fold (
    acc: oldName:
    recursiveUpdate acc (
      setAttrByPath (toSnakeCasePath oldName) (getAttrFromPath oldName cfg.settings)
    )
  ) { } definedOpts
)
```

Also: we should check that the plugin itself doesn't care that we're passing additional unknown table entries through to its `setup` function.

khaneliman: That's what i'm running into, trying to see if I can get around infinite recursion.

plugins/by-name/typescript-tools/default.nix
```diff
@@ -160,43 +160,69 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+        let
+          definedOpts = builtins.trace "Defined options in cfg.settings" (
+            lib.filter (opt: lib.hasAttrByPath (lib.toList opt) cfg.settings) oldSettings
+          );
+
+          migratedSettings = builtins.trace "Migrated settings" (
+            lib.fold (
+              acc: oldName:
+              let
+                newPath = lib.nixvim.toSnakeCase oldName;
```
MattSturgeon, line 208: I don't think this handles list-paths correctly? `toSnakeCase` is type String -> String

https://github.com/nix-community/nixvim/blob/029eafd70d6e28919a9ec01a94a46b51c4ccff40/lib/utils.nix#L64-L77

MattSturgeon: I think you want:
```nix
let
  oldPath = toList oldName;
  newPath = map toSnakeCase oldPath;
in
```

khaneliman: I thought I had that in here at one point... must have reverted it trying to figure out the infinite recursion.

plugins/by-name/typescript-tools/default.nix
MattSturgeon: Thanks for trying! If it doesn't work, don't sweat too hard over it. I knew `attrsOf` had some inf-recursion limitations here, I just wasn't sure how it'd be affected if the conditional stuff was "a level above" the attr being defined...

My curiosity is itching to figure out a solution, but it isn't worth laboring over when the warnings alone are _good enough_.

khaneliman: Yeah... I really wanted to get it to work because it'd be one less thing for people to have to do. But, it's feeling a bit above my skillset, here.

MattSturgeon: I just realised, because we can't rename these to the correct names, we are actually _breaking_ users configs.

I also had a thought about how we can make the renames _kinda_ work:

Maybe instead of doing the recursive `settings.settings = mapToNewSettings cfg.settings`, we could fallback to our "old school" _set `settingsOptions` in a let-in block before calling `toLuaObject`_ technique.

This would bypass the infinite recursion, because we wouldn't be assigning to the `settings` option.

khaneliman:
Guess I need to check if they are defined before merging them in

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,68 @@ in
...
+      ];
+    in
+    {
+      plugins.lsp.postConfig =
+        let
+          # TODO:: introduced 10-22-2024: remove after 24.11
+          renamedSettings = lib.foldl' (
+            acc: optPath:
+            let
+              snakeCaseName = lib.concatStringsSep "_" (lib.map lib.nixvim.toSnakeCase (lib.toList optPath));
```
khaneliman, line 197: I don't like this at all, but I keep breaking it when I change it...

MattSturgeon: This doesn't preserve the nested attrs. E.g. `jsxCloseTag.enable` will be renamed to `jsx_close_tag_enable`

Instead we'll need to replace `snakeCaseName` with a `snakeCasePath`, and combine [`recursiveUpdate`](https://noogle.dev/f/lib/recursiveUpdate) with [`setAttrByPath`](https://noogle.dev/f/lib/setAttrByPath).

We also don't really need to handle the null-checking here, since `toLuaObject` will do that for us. Even though I wish it didn't...

We can use [`getAttrFromPath`](https://noogle.dev/f/lib/getAttrFromPath) with no fallback value, because `definedOpts` has already been filtered using [`hasAttrByPath`](https://noogle.dev/f/lib/hasAttrByPath).

We don't need `toList` either, because you already normalized `definedOpts` to all be attrpath-lists.

```nix
renamedSettings = lib.foldl' (
  acc: optPath:
  let
    snakeCasePath = builtins.map lib.nixvim.toSnakeCase optPath;
    optValue = lib.getAttrFromPath optPath cfg.settings;
  in
  lib.recursiveUpdate acc (lib.setAttrByPath snakeCasePath optValue)
) { } definedOpts;
```

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,68 @@ in
...
+        [ "codeLens" ]
+        [ "disableMemberCodeLens" ]
+        [
+          "jsxCloseTag"
+          "enable"
+        ]
+        [
+          "jsxCloseTag"
+          "filetypes"
+        ]
```
MattSturgeon, line 187: Given these are the only nested options moved to settings.settings, and they only need snakeCase\_conversion at the top-level, we could dramatically simplify everything below by only handling the top-level attr rename?

So instead of a list of attr-paths, we'd have a list of (top-level) attr names.

Although, that may result in slightly less nice warning messages.

MattSturgeon: Maybe we could keep the nested attrs for the warnings and "has attr" filtering, but then use `map head` to get the top-level attrs to simplify the renames?

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,68 @@ in
...
+        ''
+          require('typescript-tools').setup(${lib.nixvim.toLuaObject setupOptions})
+        '';
+
+      # TODO:: introduced 10-22-2024: remove after 24.11
+      # Nested settings can't have normal mkRenamedOptionModule functionality so we can only
+      # alert the user that they are using the old values
+      warnings = lib.optional (definedOpts != [ ]) ''
+        Nixvim(plugins.typescript-tools): The following settings have moved under `plugins.typescript-tools.settings.settings` with snake_case:
+        ${lib.concatMapStringsSep "\n" (opt: "  - ${lib.showOption (lib.toList opt)}") definedOpts}
```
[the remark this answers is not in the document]
khaneliman:
```suggestion
         ${lib.concatMapStringsSep "\n" (opt: "  - plugins.typescript-tools.settings.${lib.showOption (lib.toList opt)}") definedOpts}
```
Would be accurate but very verbose and different from the other implementations of this kind of warning.

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +157,68 @@ in
...
+            // lib.optionalAttrs (optValue != null) {
+              ${snakeCaseName} = optValue;
+            }
+          ) { } definedOpts;
+
+          setupOptions = {
+            inherit (cfg.settings) handlers on_attach;
+
+            settings = cfg.settings.settings // renamedSettings;
           };
```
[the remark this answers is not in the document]
khaneliman: > This prevents settings being truely freeform.

Completely overlooked this when I was just testing the old behavior.. lol

khaneliman:
image
Looks much better, thanks @MattSturgeon !

plugins/by-name/typescript-tools/default.nix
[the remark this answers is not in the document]
khaneliman: image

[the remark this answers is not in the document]
MattSturgeon: I agree, it's better with the original less verbose version

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +158,117 @@ in
...
+                  value = if lib.isAttrs v then filterAttrsRecursivePath predicate v path' else v;
+                }
+              ) (lib.attrNames set)
+            );
+
+          setupOptions =
+            filterAttrsRecursivePath (
+              path: value: !builtins.elem path definedOpts
+              # TODO: Maybe also remove if value is an attrset whose only children are in definedOpts?
+              # Although, that isn't important because toLuaObject will filter out empty attrsets anyway...
```
MattSturgeon, line 256:
```suggestion
```

Keep the comments if you want, but when I first wrote them I only really intended them for the review :slightly_smiling_face:

plugins/by-name/typescript-tools/default.nix
```diff
@@ -157,38 +158,117 @@ in
...
+            ) cfg.settings [ ]
+            // {
+              settings = lib.recursiveUpdate cfg.settings.settings renamedSettings;
```
MattSturgeon, line 259: This won't properly merge list types (well, any non-attrs), but I don't think doing so is worth the complexity...

plugins/by-name/typescript-tools/default.nix
khaneliman: lol sorry i'm on auto pilot this morning

plugins/by-name/typescript-tools/default.nix
MattSturgeon: One way to solve this may be to use the module system to do the merging for us, instead of doing it ourselves here.

While we've already established we can't recursively assign to `settings`, maybe we could have an internal `finalSettings` option?

```nix
finalSettings = lib.mkOption {
  type = with lib.types; attrsOf anything;
  internal = true;
}
```

```nix
finalSettings = lib.mkMerge [
  (filterAttrsRecursivePath (path: value: !builtins.elem path definedOpts) cfg.settings [ ])
  { settings = renamedSettings; }
]
```

khaneliman: > This won't properly merge list types (well, any non-attrs), but I don't think doing so is worth the complexity...

Hopefully the data contract upstream doesn't change too much by the time we remove the deprecation transformation, it seemed to work fine for the current situation but if we run into issues I can revisit this. Thanks for the help!

khaneliman:
@mergify queue

PR #2450, opened 2024-10-23
plugins/lsp-format: migrate to mkNeovimPlugin

khaneliman:
Not sure if anyone is familiar with a good way to represent the submodule options for the top level language options. I decided to just do a rename from setup -> settings and mention the configuration through a description and examples.

### DIFF
--- a/plugins/by-name/lsp-format/default.nix
+++ b/plugins/by-name/lsp-format/default.nix
@@ -1,66 +1,68 @@
 {
   lib,
   config,
-  pkgs,
   ...
 }:
-let
-  cfg = config.plugins.lsp-format;
-in
-{
-  options.plugins.lsp-format = lib.nixvim.neovim-plugin.extraOptionsOptions // {
-    enable = lib.mkEnableOption "lsp-format.nvim";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "lsp-format";
+  originalName = "lsp-format.nvim";
+  package = "lsp-format-nvim";
 
-    package = lib.mkPackageOption pkgs "lsp-format.nvim" {
-      default = [
-        "vimPlugins"
-        "lsp-format-nvim"
+  maintainers = [ lib.maintainers.khaneliman ];
+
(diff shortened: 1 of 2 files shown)


plugins/by-name/lsp-format/default.nix
```diff
@@ -89,35 +89,28 @@ in
...
+          ''
+            require("lsp-format").on_attach(client)
+          '';
 
-        servers =
-          if (isList cfg.lspServersToEnable) then
-            genAttrs cfg.lspServersToEnable (serverName: {
-              onAttach.function = ''
+      servers =
+        if (lib.isList cfg.lspServersToEnable) then
```
GaetanLepage, line 105: Shouldn't we use `lib.optionalAttrs` here ?

MattSturgeon: This is just a formatting change (the `if then else` is pre-existing code), but yes it can be simplified to `optionalAttrs`.

Maybe it should even be a `mkIf`, I'm not sure... (they do subtly different things)

GaetanLepage: Yes, this was pre-existing code.



> Maybe it should even be a `mkIf`, I'm not sure... (they do subtly different things)

`optionalAttrs` and `if ... then ... else {}` are equivalent. However `mkIf` operates at a later stage (when the option values are resolved).
`mkIf` should be better then.

khaneliman: optionalAttrs works but for some reason mkIf fails because it still tries to evaluate the inner expression with the default string value.

MattSturgeon: Yeah, one of the advertised advantages(?) of `mkIf` is that the module system _usually_ still evaluates the definition value, even when the condition is false.

Thanks for testing it out!

khaneliman: Ah yeah, I remember us talking about that before. Makes sense.

MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2454, opened 2024-10-24
plugins/lsp: no call fn in keymaps.extra example

mightyiam:
Because by calling we do not get the function reference but its return
value, instead.


traxys (approved):

traxys:
@mergify queue

PR #2461, opened 2024-10-26
plugins/nui: add nui and its test

DataHearth:
I added the <https://github.com/MunifTanjim/nui.nvim/> plugin. It was already packed at nixpkgs and up-to-date on unstable

### DIFF
--- a/plugins/by-name/nui/default.nix
+++ b/plugins/by-name/nui/default.nix
@@ -1,0 +1,11 @@
+{ lib, ... }:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nui";
+  originalName = "nui.nvim";
+  package = "nui-nvim";
+  description = "UI Component Library for Neovim";
+  maintainers = [ lib.maintainers.DataHearth ];
+
+  callSetup = false;
+  hasSettings = false;
+}
--- a/tests/test-sources/plugins/by-name/nui/default.nix
+++ b/tests/test-sources/plugins/by-name/nui/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.nui.enable = true;
+  };
+}


plugins/by-name/nui/default.nix
```diff
@@ -0,0 +1,9 @@
+{ lib, ... }:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nui";
+  originalName = "nui.nvim";
+  package = "nuinvim";
```
traxys, line 5: I think the package `nui-nvim`.
I also think that the plugin has no setup function, you'll need to disable it

DataHearth: Sry, wasn't meant to be ready for review ^^. I haven't tested it yet. But yeah, you're right, syntax error and correct the setup

plugins/by-name/nui/default.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, ... }:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nui";
+  originalName = "nui.nvim";
+  package = "nui-nvim";
+  url = "https://github.com/MunifTanjim/nui.nvim/";
```
MattSturgeon, line 6: Most packages have a `meta.homepage`. If this one does (I haven't checked), we don't need to set `url` ourselves.

plugins/by-name/nui/default.nix
```diff
@@ -0,0 +1,11 @@
...
+  description = "UI Component Library for Neovim";
+  maintainers = [ lib.maintainers.DataHearth ];
+
+  callSetup = false;
```
MattSturgeon, line 10: If this is a plugin with no settings, we probably also want `hasSettings = false`.

Otherwise, we need to "use" `cfg.settings` somewhere.

DataHearth: I'm not quite sure. I doesn't have any "main" settings like some other plugins, but work with submodules instead for its configuration. So I would say no for this use-case. I'll update the hasSettings params.

khaneliman: Looks like it's got components to be referenced to construct a UI and doesn't have its own settings to configure. Not sure what a user is going to do with this module in the context of nixvim, unless it's just for an easy way to install the plugin through a plugins.*.enable syntax.

traxys: Well with the addition of plugins.*.luaConfig it can also help segment the configuration

MattSturgeon: Just to play _devil's advocate_ a little:

If the plugin is only really ever used as a dependency of some other plugin, and never really needs to be installed/used by end-users, then we shouldn't really _need_ a `plugins.*.enable` and instead can rely on:

1. Declaring plugin packages depend on this plugin upstream in nixpkgs
2. Adding the plugin-package to nixvim's `extraPlugins` if for whatever reason that isn't done

traxys: Note that nixvim (or more precisly nixpgks) automatically installs depedent plugins

traxys (approved):

traxys:
@mergifyio queue

PR #2462, opened 2024-10-26
docs/contributing: improve mk(Neovim/Vim)Plugin function documentation

DataHearth:
I tried to add a little bit of documentation regarding the `mkPlugin` functions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,13 +31,15 @@ Most of nixvim is dedicated to wrapping neovim plugins such that we can configur
 To add a new plugin you need to do the following.
 
 1. Add a file in the correct sub-directory of [`plugins`](plugins).
-  - Most plugins should be added to [`plugins/by-name/<name>`](plugins/by-name).
-    Plugins in `by-name` are automatically imported 🚀
-  - Occasionally, you may wish to add a plugin to a directory outside of `by-name`, such as [`plugins/colorschemes`](plugins/colorschemes).
-    If so, you will also need to add your plugin to [`plugins/default.nix`](plugins/default.nix) to ensure it gets imported.
-    Note: the imports list is sorted and grouped. In vim, you can usually use `V` (visual-line mode) with the `:sort` command to achieve the desired result.
+
+- Most plugins should be added to [`plugins/by-name/<name>`](plugins/by-name).
+  Plugins in `by-name` are automatically imported 🚀
+- Occasionally, you may wish to add a plugin to a directory outside of `by-name`, such as [`plugins/colorschemes`](plugins/colorschemes).
+  If so, you will also need to add your plugin to [`plugins/default.nix`](plugins/default.nix) to ensure it gets imported.
+  Note: the imports list is sorted and grouped. In vim, you can usually use `V` (visual-line mode) with the `:sort` command to achieve the desired result.
 
 2. The vast majority of plugins fall into one of those two categories:
+
 - _vim plugins_: They are configured through **global variables** (`g:plugin_foo_option` in vimscript and `vim.g.plugin_foo_option` in lua).\
   For those, you should use the `lib.nixvim.vim-plugin.mkVimPlugin`.\
   -> See [this plugin](plugins/utils/direnv.nix) for an example.
@@ -45,22 +47,135 @@ To add a new plugin you need to do the following.
   For those, you should use the `lib.nixvim.neovim-plugin.mkNeovimPlugin`.\
   -> See the [template](plugins/TEMPLATE.nix).
 
-3. Add the necessary parameters for the `mkNeovimPlugin`/`mkVimPlugin`:
-  - `name`: The name of the plugin. The resulting nixvim module will have `plugins.<name>` as a path.\
-    For a plugin named `foo-bar.nvim`, set this to `foo-bar` (subject to exceptions).
-  - `originalName`: The "real" name of the plugin (i.e. `foo-bar.nvim`). This is used mostly in documentation.
-  - `package`: The nixpkgs package attr for this plugin
-     e.g. `"foo-bar-nvim` for `pkgs.vimPlugins.foo-bar-nvim`, or `[ "hello" "world" ]` for `pkgs.hello.world`.
-  - `maintainers`: Register yourself as a maintainer for this plugin:
-    - e.g. `maintainers = [ lib.maintainers.JosephFourier ]`
-    - If you are already registered as a [`nixpkgs` maintainer][nixpkgs-maintainers], there is nothing more to do.
-    - Otherwise, you should add yourself to our [`maintainers.nix`](lib/maintainers.nix) list.
-    - See the documentation at the top of the [`nixpkgs` maintainers list][nixpkgs-maintainers] for more detail.
-  - `settingsOptions`: All or some (only the most important ones) option declarations for this plugin settings.\
-    See below for more information
-  - `settingsExample`: An example of what could the `settings` attrs look like.
+3. Add the necessary arguments when calling either [`mkNeovimPlugin`](#mkneovimplugin) or [`mkVimPlugin`](#mkvimplugin)
 
-[nixpkgs-maintainers]: https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix
+#### `mkNeovimPlugin`
+
+The `mkNeovimPlugin` function provides a standardize way to create a `Neovim` plugin.
+This is intended to be used with lua plugins that have a `setup` function,
+although it is flexible enough to be used with similar variants of plugins.
+A template plugin can be found in (plugins/TEMPLATE.nix)[https://github.com/nix-community/nixvim/blob/main/plugins/TEMPLATE.nix].
+
+| Parameter                    | Description                                                                                                                                                                                                                                          | Required | Default Value                                                                                   |
+| ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ----------------------------------------------------------------------------------------------- |
+| **name**                     | The name of the plugin.                                                                                                                                                                                                                              | Yes      | N/A                                                                                             |
+| **url**                      | The URL of the plugin's repository.                                                                                                                                                                                                                  | Yes      | `package` parameter's `meta.homepage`                                                           |
+| **callSetup**                | Indicating whether to call the setup function. Useful when `setup` function needs customizations.                                                                                                                                                    | No       | `true`                                                                                          |
+| **colorscheme**              | The name of the colorscheme.                                                                                                                                                                                                                         | No       | `name` parameter                                                                                |
+| **configLocation**           | The location for the Lua configuration.                                                                                                                                                                                                              | No       | `"extraConfigLuaPre"` if `isColorscheme` then `extraConfigLuaPre`, otherwise `"extraConfigLua"` |
+| **deprecateExtraOptions**    | Indicating whether to deprecate the `extraOptions` attribute. Mainly used for old plugins.                                                                                                                                                           | No       | `false`                                                                                         |
+| **description**              | A brief description of the plugin. Can also be used for non-normative documentation, warnings, tips and tricks.                                                                                                                                      | No       | `null`                                                                                          |
+| **extraConfig**              | Additional configuration for the plugin.                                                                                                                                                                                                             | No       | `{}`                                                                                            |
+| **extraOptions**             | Module options for the plugin, to be added _outside_ of the `settings` option. These should be Nixvim-specific options.                                                                                                                              | No       | `{}`                                                                                            |
+| **extraPackages**            | Extra packages to include.                                                                                                                                                                                                                           | No       | `[]`                                                                                            |
+| **extraPlugins**             | Extra plugins to include.                                                                                                                                                                                                                            | No       | `[]`                                                                                            |
+| **hasConfigAttrs**           | Indicating whether the plugin has configuration attributes.                                                                                                                                                                                          | No       | `true`                                                                                          |
+| **hasSettings**              | Indicating whether the plugin has settings. A `settings` option will be created if true.                                                                                                                                                             | No       | `true`                                                                                          |
+| **imports**                  | Additional modules to import.                                                                                                                                                                                                                        | No       | `[]`                                                                                            |
+| **isColorscheme**            | Indicating whether the plugin is a colorscheme.                                                                                                                                                                                                      | No       | `false`                                                                                         |
+| **luaName**                  | The Lua name for the plugin.                                                                                                                                                                                                                         | No       | `name` parameter                                                                                |
+| **maintainers**              | Maintainers for the plugin.                                                                                                                                                                                                                          | No       | `[]`                                                                                            |
+| **optionsRenamedToSettings** | Options that have been renamed and move to the `settings` attribute.                                                                                                                                                                                 | No       | `[]`                                                                                            |
+| **originalName**             | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.nvim"`.                                                                                                                                                       | No       | `name` parameter                                                                                |
+| **package**                  | The nixpkgs package attr for this plugin. Can be a string, a list of strings, a module option, or any derivation. For example, "foo-bar-nvim" for `pkgs.vimPlugins.foo-bar-nvim`, or `[ "hello" "world" ]` will be referenced as `pkgs.hello.world`. | No       | `name` parameter                                                                                |
+| **settingsDescription**      | A description of the settings provided to the `setup` function.                                                                                                                                                                                      | No       | `"Options provided to the require('${luaName}')${setup} function."`                             |
+| **settingsExample**          | An example configuration for the plugin's settings.                                                                                                                                                                                                  | No       | `null`                                                                                          |
+| **settingsOptions**          | Options representing the plugin's settings. This is optional because `settings` is a "freeform" option. See [Declaring plugin options](#declaring-plugin-options).                                                                                   | No       | `{}`                                                                                            |
+| **setup**                    | The setup function for the plugin.                                                                                                                                                                                                                   | No       | `".setup"`                                                                                      |
+
+##### Functionality
+
+The `mkNeovimPlugin` function generates a Nix module that:
+
+1. Defines the plugin's metadata, including maintainers, description, and URL.
+2. Sets up options for enabling the plugin, specifying the package, and configuring settings and Lua configuration.
+
+##### Example Usage
+
+```nix
+mkNeovimPlugin {
+  name = "example-plugin";
+  maintainers = [ lib.maintainers.user ];
+  url = "https://github.com/example/example-plugin";
+  description = "An example Neovim plugin";
+  settingsOptions = {
+    option1 = lib.mkOption {
+      type = lib.types.str;
+      default = "default-value";
+      description = "An example option";
+    };
+  };
+}
+```
+
+Here's an example plugin: [lsp_lines.nvim](https://github.com/nix-community/nixvim/commit/8815180c625e9766b2cb3126756b97e994998228)
+
+This example defines a Neovim plugin named `example-plugin` with specified maintainers, URL, description, settings options, and additional configuration. `package` will be 'example-plugin'
+thanks to package referring to the `name` attribute.
+
+#### `mkVimPlugin`
+
+The `mkVimPlugin` function provides a standardized way to create a `Vim` plugin.
+This is intended to be used with traditional vim plugins, usually written in viml.
+Such plugins are usually configured via vim globals, but often have no configurable options at all.
+
+| Parameter                    | Description                                                                                                                                                                                                                                        | Required | Default Value                         |
+| ---------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ------------------------------------- |
+| **name**                     | The name of the Vim plugin.                                                                                                                                                                                                                        | Yes      | N/A                                   |
+| **url**                      | The URL of the plugin repository.                                                                                                                                                                                                                  | Yes      | `package` parameter's `meta.homepage` |
+| **colorscheme**              | The name of the colorscheme.                                                                                                                                                                                                                       | No       | `name` parameter                      |
+| **deprecateExtraConfig**     | Flag to deprecate extra configuration.                                                                                                                                                                                                             | No       | `false`                               |
+| **description**              | A description of the plugin. Can also be used for non-normative documentation, warnings, tips and tricks.                                                                                                                                          | No       | `null`                                |
+| **extraConfig**              | Extra configuration for the plugin.                                                                                                                                                                                                                | No       | `cfg: {}`                             |
+| **extraOptions**             | Extra options for the plugin.                                                                                                                                                                                                                      | No       | `{}`                                  |
+| **extraPackages**            | Extra packages to include.                                                                                                                                                                                                                         | No       | `[]`                                  |
+| **extraPlugins**             | Extra plugins to include.                                                                                                                                                                                                                          | No       | `[]`                                  |
+| **globalPrefix**             | Global prefix for the settings.                                                                                                                                                                                                                    | No       | `""`                                  |
+| **imports**                  | A list of imports for the plugin.                                                                                                                                                                                                                  | No       | `[]`                                  |
+| **isColorscheme**            | Flag to indicate if the plugin is a colorscheme.                                                                                                                                                                                                   | No       | `false`                               |
+| **maintainers**              | The maintainers of the plugin.                                                                                                                                                                                                                     | No       | `[]`                                  |
+| **optionsRenamedToSettings** | List of options renamed to settings.                                                                                                                                                                                                               | No       | `[]`                                  |
+| **originalName**             | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.vim"`.                                                                                                                                                      | No       | `name` parameter                      |
+| **package**                  | The nixpkgs package attr for this plugin. Can be a string, a list of strings, a module option, or any derivation. For example, "foo-bar-vim" for `pkgs.vimPlugins.foo-bar-vim`, or `[ "hello" "world" ]` will be referenced as `pkgs.hello.world`. | No       | `name` parameter                      |
+| **settingsExample**          | Example settings for the plugin.                                                                                                                                                                                                                   | No       | `null`                                |
+| **settingsOptions**          | Options representing the plugin's settings. This is optional because `settings` is a "freeform" option. See [Declaring plugin options](#declaring-plugin-options).                                                                                 | No       | `{}`                                  |
+
+##### Functionality
+
+The `mkVimPlugin` function generates a Nix module that:
+
+1. Defines the plugin's metadata, including maintainers, description, and URL.
+2. Sets up options for enabling the plugin, specifying the package, and configuring settings and extra configuration.
+
+##### Example Usage
+
+```nix
+mkVimPlugin {
+  name = "example-plugin";
+  url = "https://github.com/example/plugin";
+  maintainers = [ lib.maintainers.user ];
+  description = "An example Vim plugin.";
+  globalPrefix = "example_";
+}
+```
+
+Simple vim plugins already implemented:
+
+- [earthly.vim](https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/earthly/default.nix)
+- [vim-nix](https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/nix/default.nix)
+
+All the plugins are located under the `plugins` folder. If you want which plugins are defined as vim plugins, follow these steps:
+
+```bash
+# Ensure you are in the nixvim directory
+cd nixvim
+
+# Either setup nix-direnv, or manually enter a devshell using:
+nix develop
+
+# List all created plugins with `mkVimPlugin`
+list-plugins -k vim
+```
 
 #### Declaring plugin options
 
@@ -87,6 +202,7 @@ There are a number of helpers added into `lib` that can help you correctly imple
 - `lib.types.rawLua`: A type to represent raw lua code. The values are of the form `{ __raw = "<code>";}`.
 
 The resulting `settings` attrs will be directly translated to `lua` and will be forwarded the plugin:
+
 - Using globals (`vim.g.<globalPrefix><option-name>`) for plugins using `mkVimPlugin`
 - Using the `require('<plugin>').setup(<options>)` function for the plugins using `mkNeovimPlugin`
 


GaetanLepage:
This looks really good for now !

DataHearth:
Thanks! If you want me to add anything, add details about steps, the function description or whatsoever, feel free. After thinking about it, it seems to me to be a good start for these 2 functions.

GaetanLepage:
> @GaetanLepage Thanks! If you want me to add anything, add details about steps, the function description or whatsoever, feel free. After thinking about it, it seems to me to be a good start for these 2 functions.

Yes, this definitely looks like a good start.
For me, this is already enough for one PR.
Of course, subsequent improvements of the doc can be done in following PRs.

GaetanLepage (approved):
Thanks once again for this very useful work :)

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
-[nixpkgs-maintainers]: https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix
+3. Add the necessary parameters for the [`mkNeovimPlugin`](#mkneovimplugin)/[`mkVimPlugin`](#mkvimplugin)
+
+#### `mkNeovimPlugin`
+
+The `mkNeovimPlugin` function provide a standardize way to create a `Neovim` plugin.
+
+| Parameter                  | Description                                                                                                                                                                                                 | Required | Default Value |
+|----------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|---------------|
+| **name**                   | The name of the plugin.                                                                                                                                                                                      | Yes      | N/A           |
```
khaneliman, line 56: Wonder if it might be nice to sort these tables alphabetically or by required status

DataHearth: It should already be sorted by required attr. Would you like to sort them only by name or name and required?

khaneliman: Name and required would be nice.

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **maintainers**            | Maintainers for the plugin.                                                                                                                                                                                  | Yes      | N/A           |
+| **url**                    | The URL of the plugin's repository.                                                                                                                                                                          | Yes      | If not set, `package`'s' `homepage` attribute |
```
khaneliman, line 58:
```suggestion
| **url**                    | The URL of the plugin's repository.                                                                                                                                                                          | No     | If not set, `package`'s' `homepage` attribute |
```
I've never passed url to a `mkNeovimPlugin`, so I wouldn't say that's required. The default value even explains why it's not required.

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **package**                | The nixpkgs package attr for this plugin. Can be a string, a list of strings, a module option, or any derivation. For example, "foo-bar-nvim" for `pkgs.vimPlugins.foo-bar-nvim`, or `[ "hello" "world" ]` will be referenced as `pkgs.hello.world`. | No       | `name` parameter |
+| **imports**                | Additional Nix modules to import.                                                                                                                                                                            | No       | `[]`            |
```
MattSturgeon, line 60:
```suggestion
| **imports**                | Additional modules to import.                                                                                                                                                                            | No       | `[]`            |
```

I never know how to word it when referencing modules.

"Nix modules" is wrong, because the module system comes from nixpkgs/nixos _not_ from nix itself.

"Nixpkgs modules" or "NixOS modules" could be used, but could also be confusing.

"Nixvim modules" is also legitimate, because the modules are intended for use with nixvim.

Or just "modules" on its own is probably enough, and removes the need to think about the additional word :sweat_smile:

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **description**            | A brief description of the plugin.                                                                                                                                                                           | No       | `null`          |
+| **deprecateExtraOptions**  | Indicating whether to deprecate the `extraOptions` attribute. Mainly used for old plugins.                                                                                                                   | No       | `false`         |
+| **optionsRenamedToSettings** | Options that have been renamed and move to the `settings` attribute.                                                                                                                                       | No       | `[]`            |
```
MattSturgeon, line 63:
```suggestion
| **deprecateExtraOptions**  | Indicating whether to deprecate the `extraOptions` attribute. Only used when migrating old plugins.                                                                                                                   | No       | `false`         |
| **optionsRenamedToSettings** | Options that have been renamed and move to the `settings` attribute. Only used when migrating old plugins.                                                                                                                                       | No       | `[]`            |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **extraConfig**            | Additional configuration for the plugin.                                                                                                                                                                     | No       | `{}`            |
+| **extraPlugins**           | Extra plugins to include.                                                                                                                                                                                    | No       | `[]`            |
+| **extraPackages**          | Extra packages to include.                                                                                                                                                                                   | No       | `[]`            |
+| **callSetup**              | Indicating whether to call the setup function.                                                                                                                                                               | No       | `true`          |
```
MattSturgeon, line 79:
```suggestion
| **callSetup**              | Indicating whether to call the setup function. Useful when the `setup` function needs to be called in a special way.                                                                                                                                                              | No       | `true`          |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **installPackage**         | Indicating whether to install the package.                                                                                                                                                                   | No       | `true`          |
```
MattSturgeon, line 80: Is this deprecated? What scenarios do we use this in?

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **settingsOptions**        | Options for the plugin's settings.                                                                                                                                                                           | No       | `{}`            |
+| **settingsExample**        | An example configuration for the plugin's settings.                                                                                                                                                          | No       | `null`          |
+| **settingsDescription**    | A description of the settings provided to the `require('${luaName}')${setup}` function.                                                                                                                      | No       | `"Options provided to the require('${luaName}')${setup} function."` |
+| **hasSettings**            | Indicating whether the plugin has settings.                                                                                                                                                                  | No       | `true`          |
```
MattSturgeon, line 72:
```suggestion
| **hasSettings**            | Indicating whether the plugin has settings. A `settings` option will be created if true.                                                                                                                                                                  | No       | `true`          |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **originalName**           | The original name of the plugin.                                                                                                                                                                             | No       | `name` parameter |
+| **settingsOptions**        | Options for the plugin's settings.                                                                                                                                                                           | No       | `{}`            |
+| **settingsExample**        | An example configuration for the plugin's settings.                                                                                                                                                          | No       | `null`          |
+| **settingsDescription**    | A description of the settings provided to the `require('${luaName}')${setup}` function.                                                                                                                      | No       | `"Options provided to the require('${luaName}')${setup} function."` |
```
MattSturgeon, line 71:
```suggestion
| **settingsDescription**    | A description of the settings provided to the `setup` function.                                                                                                                      | No       | `"Options provided to the require('${luaName}')${setup} function."` |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **optionsRenamedToSettings** | Options that have been renamed and move to the `settings` attribute.                                                                                                                                       | No       | `[]`            |
+| **isColorscheme**          | Indicating whether the plugin is a colorscheme.                                                                                                                                                              | No       | `false`         |
+| **colorscheme**            | The name of the colorscheme.                                                                                                                                                                                 | No       | `name` parameter |
+| **configLocation**         | The location for the Lua configuration.                                                                                                                                                                      | No       | `"extraConfigLuaPre"` if `isColorscheme` is `true`, otherwise `"extraConfigLua"` |
```
MattSturgeon, line 66:
```suggestion
| **configLocation**         | The location for the Lua configuration.                                                                                                                                                                      | No       | if `isColorscheme` then `"extraConfigLuaPre"`, otherwise `"extraConfigLua"` |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **url**                    | The URL of the plugin's repository.                                                                                                                                                                          | Yes      | If not set, `package`'s' `homepage` attribute |
+| **package**                | The nixpkgs package attr for this plugin. Can be a string, a list of strings, a module option, or any derivation. For example, "foo-bar-nvim" for `pkgs.vimPlugins.foo-bar-nvim`, or `[ "hello" "world" ]` will be referenced as `pkgs.hello.world`. | No       | `name` parameter |
+| **imports**                | Additional Nix modules to import.                                                                                                                                                                            | No       | `[]`            |
+| **description**            | A brief description of the plugin.                                                                                                                                                                           | No       | `null`          |
```
MattSturgeon, line 61:
```suggestion
| **description**            | A brief description of the plugin. Can also be used for non-normative documentation, warnings, tips and tricks.                                                                                                                                                                           | No       | `null`          |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **settingsExample**        | An example configuration for the plugin's settings.                                                                                                                                                          | No       | `null`          |
+| **settingsDescription**    | A description of the settings provided to the `require('${luaName}')${setup}` function.                                                                                                                      | No       | `"Options provided to the require('${luaName}')${setup} function."` |
+| **hasSettings**            | Indicating whether the plugin has settings.                                                                                                                                                                  | No       | `true`          |
+| **extraOptions**           | Additional options for the plugin.                                                                                                                                                                           | No       | `{}`            |
```
MattSturgeon, line 73:
```suggestion
| **extraOptions**           | Module options for the plugin, to be added _outside_ of the `settings` option. These should be Nixvim-specific options.                                                                                                                                                                          | No       | `{}`            |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **configLocation**         | The location for the Lua configuration.                                                                                                                                                                      | No       | `"extraConfigLuaPre"` if `isColorscheme` is `true`, otherwise `"extraConfigLua"` |
+| **hasConfigAttrs**         | Indicating whether the plugin has configuration attributes.                                                                                                                                                  | No       | `true`          |
+| **originalName**           | The original name of the plugin.                                                                                                                                                                             | No       | `name` parameter |
+| **settingsOptions**        | Options for the plugin's settings.                                                                                                                                                                           | No       | `{}`            |
```
MattSturgeon, line 69:
```suggestion
| **settingsOptions**        | Options representing the plugin's settings. This is optional because `settings` is a "freeform" option. See <insert link here>.                                                                                                                                                                           | No       | `{}`            |
```

We want to keep reminding contributors that it is not necessary to represent plugin settings as nix options in order for end users to configure them.

We should link to the relevant sub-heading where this is discussed in more detail.

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **colorscheme**            | The name of the colorscheme.                                                                                                                                                                                 | No       | `name` parameter |
+| **configLocation**         | The location for the Lua configuration.                                                                                                                                                                      | No       | `"extraConfigLuaPre"` if `isColorscheme` is `true`, otherwise `"extraConfigLua"` |
+| **hasConfigAttrs**         | Indicating whether the plugin has configuration attributes.                                                                                                                                                  | No       | `true`          |
+| **originalName**           | The original name of the plugin.                                                                                                                                                                             | No       | `name` parameter |
```
MattSturgeon, line 68:
```suggestion
| **originalName**           | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.nvim"`.                                                                                                                                                                             | No       | `name` parameter |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+| **isColorscheme**          | Indicating whether the plugin is a colorscheme.                                                                                                                                                              | No       | `false`         |
+| **colorscheme**            | The name of the colorscheme.                                                                                                                                                                                 | No       | `name` parameter |
+| **configLocation**         | The location for the Lua configuration.                                                                                                                                                                      | No       | `"extraConfigLuaPre"` if `isColorscheme` is `true`, otherwise `"extraConfigLua"` |
+| **hasConfigAttrs**         | Indicating whether the plugin has configuration attributes.                                                                                                                                                  | No       | `true`          |
```
MattSturgeon, line 67: I can't even recall what this does or what it's for :eyes:

MattSturgeon (commented):
Some thoughts, ideas, rewording suggestions, etc.

Happy to discuss alternatives or whether particular ideas should be part of a followup PR.

DataHearth:
@MattSturgeon @khaneliman I'll check your comments tonight and see what I can come up with 👍.

I'll definitely sort the table and try to check the formatting applied by GitHub in case of multiline string. It's difficult to read it when not rendered.

CONTRIBUTING.md
MattSturgeon: Looking around online, it seems GFM may not support wrapping/unwrapping table lines and if we wanted this we'd have to resort to HTML tables...

E.g. no mention in the [GFM spec](https://github.github.com/gfm/#tables-extension-)
And no usuful answers on this [Stack Overflow thread](https://stackoverflow.com/questions/15955160/markdown-table-with-long-lines)

DataHearth:
I started working on it. I Haven't finished to review comments and update links.

CONTRIBUTING.md
DataHearth: @MattSturgeon Hey! It's when the plugin is configured at runtime I suppose. That's how I interpret these lines: https://github.com/nix-community/nixvim/blob/main/lib/neovim-plugin.nix#L30-L31

CONTRIBUTING.md
DataHearth: I can't a relevant link to set here for a simple freeform plugin already added to nixvim. Do you have one?

DataHearth:
Hey! I've updated, sorted and added examples. Tell me what you think about it :smile:

CONTRIBUTING.md
MattSturgeon: Ah this was from @traxys recent PR. Any input on how this should be documented? Especially as it's a rather niche argument...

MattSturgeon:
Thanks! I noticed there's still a few unaddressed feedback threads, maybe they were hidden by github and you didn't spot them?

image


On the whole everything looks good.

The CI failure seems unrelated, it has been cached from a CI run that got cancelled...

Are you ready to undraft this or is there anything in particular you plan to improve first?

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+
+#### `mkVimPlugin`
+
+The `mkVimPlugin` function provides a standardized way to create a `Vim` plugin.
+
+| Parameter                | Description                                                                 | Required | Default Value |
+|--------------------------|-----------------------------------------------------------------------------|----------|---------------|
+| **name**                 | The name of the Vim plugin.                                                 | Yes      | N/A           |
+| **url**                  | The URL of the plugin repository.                                           | Yes      | N/A           |
+| **maintainers**          | The maintainers of the plugin.                                              | Yes      | Throw if not set and `package` doesn't contain a `homepage` attribute |
```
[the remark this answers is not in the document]
DataHearth: As you said, and I verified, it'll take the meta.homepage attr from the original nixpkgs pkg https://github.com/nix-community/nixvim/blob/main/lib/neovim-plugin.nix#L82 . I guess it'll throw an error if both unset. The original destructed attr has a default value to throw, but I'll never be called. We use the `args.url`, which, if I understood our conversations and the wiki and the magic happening behind, doesn't use the default value.

CONTRIBUTING.md
DataHearth: I haven't checked TBH. I only know the  `callSetup` which is used to tell when a lib is packaged (generally) and doesn't really on a setup function.

DataHearth:
> Thanks! I noticed there's still a few unaddressed feedback threads, maybe they were hidden by github and you didn't spot them?
> 
> image
> 
> On the whole everything looks good.
> 
> The CI failure seems unrelated, it has been cached from a CI run that got cancelled...
> 
> Are you ready to undraft this or is there anything in particular you plan to improve first?

Hey! I was missing some comments left by you. The GitHub app is kind of shitty for following PR reviews... I've made an update with the last comments unresolved. I'll open it for review. But I kind of want a clean review if you (all) have time, please 🙏. I want it to be clear enough to avoid looking at different files/plugins to check how things are done for newcomers ^^

CONTRIBUTING.md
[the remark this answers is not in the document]
MattSturgeon: This is the `maintainers` attribute, not the `url` one

CONTRIBUTING.md
MattSturgeon: Seems I added it back in https://github.com/nix-community/nixvim/commit/843fb302ebaa2559b7c50ad69ba9a00ae1a5836d, although it is a very niche option...

A quick search shows it is only used by `treesitter`: https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/treesitter/default.nix#L431-L434

Maybe add a caveat:
```suggestion
| **installPackage**         | Whether to install the package. This should almost never be disabled. It is occasionally disabled by plugin modules that need to apply additional overrides to the package before it is added to `extraPackages`. | No       | `true`          |
```

CONTRIBUTING.md
MattSturgeon: I think [`#declaring-plugin-options`](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md#declaring-plugin-options) is probably what I had in mind.

CONTRIBUTING.md
khaneliman: Sorry, didn't see this PR. Thanks for catching that Matt. I remember you adding that specifically for treesitter way back when lol

CONTRIBUTING.md
MattSturgeon: This was moved up to line 62, but I think it'd still be nice to add the `Useful when the `setup` function needs to be called in a special way.` suggestion?

CONTRIBUTING.md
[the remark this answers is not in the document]
DataHearth: Ohh, yeah! Mb... I'm updating it.

CONTRIBUTING.md
DataHearth: > Seems I added it back in https://github.com/nix-community/nixvim/commit/843fb302ebaa2559b7c50ad69ba9a00ae1a5836d, although it is a very niche option...
> 
> 
> 
> A quick search shows it is only used by `treesitter`: https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/treesitter/default.nix#L431-L434
> 
> 
> 
> Maybe add a caveat along the lines of?
> 
> ```suggestion
> 
> | **installPackage**         | Whether to install the package. This should almost never be disabled. It is occasionally disabled by plugin modules that need to apply additional overrides to the package before it is added to `extraPackages`. | No       | `true`          |
> 
> ```

@MattSturgeon Do you have something in mind? Without really knowing the full purpose of it, I find the line enough to either not mess with it, or start working on it.

MattSturgeon: > Do you have something in mind?

What I had in mind was something along the lines of my earlier suggestion:

``Whether to install the package. This should almost never be disabled. It is occasionally disabled by plugin modules that need to apply additional overrides to the package before it is added to `extraPackages`.``

This is a little wordy though, I'll admit!

> Without really knowing the full purpose of it, I find the line enough to either not mess with it, or start working on it.

Yeah, I agree this is difficult. And we don't need to have everything perfect since things can always be improved down the line.

CONTRIBUTING.md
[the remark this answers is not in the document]
DataHearth: It's defined as `[]` liked you said :+1: . Good catch!

CONTRIBUTING.md
DataHearth: I've made a small change to your sentence. It should be a little bit less verbose. But also a little bit less comprehensible. Lets see what you think about it :+1:

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+| **imports**              | A list of imports for the plugin.                                           | No       | `[]`          |
+| **isColorscheme**        | Flag to indicate if the plugin is a colorscheme.                            | No       | `false`       |
+| **maintainers**          | The maintainers of the plugin.                                              | No      | `[]` |
+| **optionsRenamedToSettings** | List of options renamed to settings.                                    | No       | `[]`          |
+| **originalName**         | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.nvim"`.                                            | No       | `name` parameter |
+| **package**              | The package for the plugin.                                                 | No       | `name` parameter |
+| **settingsExample**      | Example settings for the plugin.                                            | No       | `null`        |
+| **settingsOptions**      | Settings options for the plugin.                                            | No       | `{}`          |
```
MattSturgeon, line 139:
```suggestion
| **settingsOptions**      | Options representing the plugin's settings. This is optional because `settings` is a "freeform" option. See [Declaring plugin options](#declaring-plugin-options).    | No       | `{}`          |
```

Should probably match the equivalent in `mkNeovimPlugin`

CONTRIBUTING.md
```diff
@@ -45,22 +45,119 @@ To add a new plugin you need to do the following.
...
+
+
+#### `mkVimPlugin`
+
+The `mkVimPlugin` function provides a standardized way to create a `Vim` plugin.
+
+| Parameter                | Description                                                                 | Required | Default Value |
+|--------------------------|-----------------------------------------------------------------------------|----------|---------------|
+| **name**                 | The name of the Vim plugin.                                                 | Yes      | N/A           |
+| **url**                  | The URL of the plugin repository.                                           | Yes      | N/A           |
```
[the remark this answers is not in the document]
MattSturgeon: This isn't resolved? Unless you disagree?

Ideally, this should match the `mkNeovimPlugin` equivalent https://github.com/nix-community/nixvim/pull/2462/files#r1818746848

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+  url = "https://github.com/example/plugin";
+  maintainers = [ lib.maintainers.user ];
+  description = "An example Vim plugin.";
+  globalPrefix = "example_";
+}
+```
+
+Simple vim plugins already implemented:
+- [earthly.vim](https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/earthly/default.nix)
+- [vim-nix](https://github.com/nix-community/nixvim/blob/6f210158b03b01a1fd44bf3968165e6da80635ce/plugins/by-name/nix/default.nix)
```
MattSturgeon, line 162: :heart:

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+| **extraPlugins**         | Extra plugins to include.                                                   | No       | `[]`          |
+| **globalPrefix**         | Global prefix for the settings.                                             | No       | `""`          |
+| **imports**              | A list of imports for the plugin.                                           | No       | `[]`          |
+| **isColorscheme**        | Flag to indicate if the plugin is a colorscheme.                            | No       | `false`       |
+| **maintainers**          | The maintainers of the plugin.                                              | No      | `[]` |
+| **optionsRenamedToSettings** | List of options renamed to settings.                                    | No       | `[]`          |
+| **originalName**         | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.nvim"`.                                            | No       | `name` parameter |
+| **package**              | The package for the plugin.                                                 | No       | `name` parameter |
```
MattSturgeon, line 137: Minor nit: this is an intrinsic issue with markdown tables, but the column deliminators have started creeping out of alignment... Same for `mkNeovimPlugin`.

Is your goal to set each column divider at a particular char-column and then just have a few exceptions for rows with particularly wordy fields?

Maybe just have a final check when all else is done that the table is formatted/aligned as you intended :grin:

MattSturgeon (commented):
Looks pretty much done to my eye, thanks again for all your efforts!

Just some minor consistency issues below:

DataHearth:
> Looks pretty much done to my eye, thanks again for all your efforts!
> 
> 
> 
> Just some minor consistency issues below:

Hey! I'll fix the last comments this weekend ^^.

DataHearth:
Hey! I didn't forget the PR 😉. I'm planning to work on it tomorrow between launch.

MattSturgeon:
> I'm planning to work on it tomorrow between launch.

No pressure!

CONTRIBUTING.md
DataHearth: Hey! TBH, I really hate MD tables due to the lack of readability on large columns... So I tried my best in my latest change to fit a *readable* design. Tell me what you think. Another shi**y solution would be to create an over-engineered method with a json/yaml input, a simple CLI command to template this and `sed` the content. So, you don't interact with raw MD tables and just the read the rendered MD.

CONTRIBUTING.md
MattSturgeon: ♥️ was supposed to be positive feedback 😂

CONTRIBUTING.md
MattSturgeon: I don't think pre-processing the tables is a good solution, especially as we want the contributing file to be rendered correctly on both GitHub and our mdbook docs.

Maybe down the line we could experiment with another format; for instance, I believe GitHub also supports AsciiDocs, I imagine mdbook can too. However that's way out of scope here.

I'll take a look at the formatting next time I'm on a big enough screen, but I'm sure it's fine 🫠

MattSturgeon (commented):
I've not had a chance to take a look from my PC yet, but just scrolling through on my phone I've spotted a couple more minor points worth discussing

CONTRIBUTING.md
MattSturgeon: Just had a look on a wider screen. Looks fine other than two lines:

<p>

image

</p>
</details> 

[Line 67](https://github.com/DataHearth/nixvim/blob/2650c1852840bf3df4fca271c9750b25711a06a5/CONTRIBUTING.md?plain=1#L67) (`extraOption`) and [Line 73](https://github.com/DataHearth/nixvim/blob/2650c1852840bf3df4fca271c9750b25711a06a5/CONTRIBUTING.md?plain=1#L73) (`installPackage`) seem to have an extra space somewhere

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+| **maintainers**                  | The maintainers of the plugin.                                                                                                                                                                               | No       | `[]`             |
+| **optionsRenamedToSettings**     | List of options renamed to settings.                                                                                                                                                                         | No       | `[]`             |
+| **originalName**                 | The original name of the plugin, usually the plugin's github repo name. E.g. `"foo-bar.nvim"`.                                                                                                               | No       | `name` parameter |
+| **package**                      | The package for the plugin.                                                                                                                                                                                  | No       | `name` parameter |
```
MattSturgeon, line 137: Minor inconsitency: the neovem version has a long description with an example. Should this one be the same, seeing as the implementation is the same?

> ``The nixpkgs package attr for this plugin. Can be a string, a list of strings, a module option, or any derivation. For example, "foo-bar-nvim" for `pkgs.vimPlugins.foo-bar-nvim`, or `[ "hello" "world" ]` will be referenced as `pkgs.hello.world`.``

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+| **extraOptions**                 | Extra options for the plugin.                                                                                                                                                                                | No       | `{}`             |
+| **extraPackages**                | Extra packages to include.                                                                                                                                                                                   | No       | `[]`             |
+| **extraPlugins**                 | Extra plugins to include.                                                                                                                                                                                    | No       | `[]`             |
+| **globalPrefix**                 | Global prefix for the settings.                                                                                                                                                                              | No       | `""`             |
```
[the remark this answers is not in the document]
DataHearth: I'll just copy past what you've written. I don't even know what a global prefix in `vim` is :joy: . 

Regarding the demonstration, this can be very useful, but I feel like it'll quickly overload the reader's ability to quickly do a search & find in this doc. He/She will be redirected to another nix example. Which is exactly what I don't want to do when coming for this section (for me at least).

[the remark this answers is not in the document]
DataHearth: Maybe in a different doc or the global documentation in mdbook ?

CONTRIBUTING.md
```diff
@@ -45,22 +45,132 @@ To add a new plugin you need to do the following.
...
+
+All the plugins are located under the `plugins` folder. If you want which plugins are defined as vim plugins, follow these steps:
+```bash
+git clone https://github.com/nix-community/nixvim.git
+
+cd nixvim
+
+# Make the to have nix-direnv enabled and running ;) => https://github.com/nix-community/nix-direnv
+# List all created plugins with `mkVimPlugin`
+list-plugins -k vim
```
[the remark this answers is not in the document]
DataHearth: Really like that. Much better :+1:

CONTRIBUTING.md
DataHearth: Good catch, should definitely be the same.

CONTRIBUTING.md
```diff
@@ -45,22 +45,134 @@ To add a new plugin you need to do the following.
...
+| **deprecateExtraOptions**  | Indicating whether to deprecate the `extraOptions` attribute. Mainly used for old plugins.                                                                                                                   | No       | `false`       |
+| **description**            | A brief description of the plugin. Can also be used for non-normative documentation, warnings, tips and tricks.                                                                                              | No       | `null`        |
+| **extraConfig**            | Additional configuration for the plugin.                                                                                                                                                                     | No       | `{}`          |
+| **extraOptions**           | Module options for the plugin, to be added _outside_ of the `settings` option. These should be Nixvim-specific options.                                                                                       | No       | `{}`          |
```
MattSturgeon, line 67: Slightly out of alignment:

```suggestion
| **extraOptions**           | Module options for the plugin, to be added _outside_ of the `settings` option. These should be Nixvim-specific options.                                                                                      | No       | `{}`          |
```

CONTRIBUTING.md
```diff
@@ -45,22 +45,134 @@ To add a new plugin you need to do the following.
...
+| **hasConfigAttrs**         | Indicating whether the plugin has configuration attributes.                                                                                                                                                  | No       | `true`        |
+| **hasSettings**            | Indicating whether the plugin has settings. A `settings` option will be created if true.                                                                                                                     | No       | `true`        |
+| **imports**                | Additional modules to import.                                                                                                                                                                                | No       | `[]`          |
+| **installPackage**         | Whether to install the package. This is a very niche option that can be used when the final package overrides the plugin's configuration. In most cases, it shouldn't disabled.                              | No       | `true`        |
```
MattSturgeon, line 73: I don't think it's accurate to say the _package_ overrides the _plugin_ config...

This wording should be a bit better, but still hard to git it exactly right:

```suggestion
| **installPackage**         | Whether to install the package. This is a very niche option, but one use case is adding additional overrides to the final package. In most cases, it shouldn't disabled.                                        | No       | `true`        |
```

---

The fact this is so clunky to describe makes me wonder if we should have a transformer argument instead of a toggle :thinking: 

Something like `packageTransformer ? (pkg: pkg)` "Used to apply final transformations to the configured package, such as necessary overrides"

Maybe I'll open a PR once this is merged

MattSturgeon (changes requested):
Final two nits. Everything else looks good.

At this point my only concern is that maybe some of the weirder (more internal) parameters may be better off **not** being documented, to avoid encouraging their use. But we can always change exactly what we document in future PRs.

Thanks for your tireless work on this, it should prove a very useful improvement!

CONTRIBUTING.md
DataHearth: I think the transformer would be a good idea depending on how this implemented and presented with arguments. The `installPackage` is self explanatory, but I still don't know how to rephrase it outside of `Just install the package or not`. It mostly depends on side args in the same func. Maybe we could just stick to a very simple `Should the package be installed or not` for now, and then try to find a better way to customize the plugin via, as you said for example, a transformer.

DataHearth: It's a deal then 🤝!

CONTRIBUTING.md
DataHearth: I can't see a problem with the alignment on my end. I'm only on a codespace for this weekend, maybe it's the line wrapping on nvim?

MattSturgeon: Looks like the headings are off too

CONTRIBUTING.md
DataHearth: done!

MattSturgeon: I've opened #2524 to remove the option. I think regardless of the outcome of that PR we probably don't need this one documented.


```suggestion
```

CONTRIBUTING.md
DataHearth: I know this is very, very zoomed out, but I think this is the github diff that f**ked up the rendering. Here's how it is on my end
image

DataHearth: > I know this is very, very zoomed out, but I think this is the github diff that f**ked up the rendering. Here's how it is on my end image

Actually, this in not the first time Github does this... I've had some rendering issue with the diff on some projects with whitespaces and newlines.

CONTRIBUTING.md
```diff
@@ -31,36 +31,151 @@ Most of nixvim is dedicated to wrapping neovim plugins such that we can configur
...
+| ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ----------------------------------------------------------------------------------------------- |
+| **name**                     | The name of the plugin.                                                                                                                                                                                                                              | Yes      | N/A                                                                                             |
+| **url**                      | The URL of the plugin's repository.                                                                                                                                                                                                                  | Yes      | N/A                                                                                             |
```
MattSturgeon, line 62: https://github.com/nix-community/nixvim/pull/2462#discussion_r1818741351 is still outstanding. `url` defaults to "default package's `meta.homepage` if available".

```suggestion
| **url**                      | The URL of the plugin's repository.                                                                                                                                                                                                                  | Yes      | `package` parameter's `meta.homepage`                                                           |
```

CONTRIBUTING.md
```diff
@@ -31,36 +31,151 @@ Most of nixvim is dedicated to wrapping neovim plugins such that we can configur
...
+| ---------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ---------------- |
+| **name**                     | The name of the Vim plugin.                                                                                                                                                                                                                        | Yes      | N/A              |
+| **url**                      | The URL of the plugin repository.                                                                                                                                                                                                                  | Yes      | N/A              |
```
MattSturgeon, line 125: Same

```suggestion
| **url**                      | The URL of the plugin repository.                                                                                                                                                                                                                  | Yes      | `package` parameter's `meta.homepage` |
```

CONTRIBUTING.md
MattSturgeon: See #2524

CONTRIBUTING.md
DataHearth: Yup, forgot the modification when to void of my mind. Corrected

MattSturgeon (approved):
Awesome! Thanks for all your hard work and persistence with this one!

MattSturgeon:
@mergify queue

GaetanLepage:
Thanks a lot to you both for this very useful contribution !

PR #2465, opened 2024-10-28
noice: migrate to mkNeovimPlugin

khaneliman:
~Got the main migration done. But, there are tons of options without descriptions or not appropriately formatted descriptions, atm.~

### DIFF
--- a/plugins/by-name/noice/default.nix
+++ b/plugins/by-name/noice/default.nix
@@ -1,6 +1,5 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
@@ -8,10 +7,13 @@
 # TODO: This uses a lot of types.anything because noice.nvim types are quite complex.
 # It should be possible to map them to nix, but they would not map really well through
 # toLuaObject, we would maybe need some ad-hoc pre-processing functions.
-with lib;
+let
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
+in
 {
-  options.plugins.noice = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption ''
+  options.plugins.noice = lib.nixvim.neovim-plugin.extraOptionsOptions // {
+    enable = lib.mkEnableOption ''
       noice.nvim, an experimental nvim UI.
       Note that if treesitter is enabled you need the following parsers:
         vim, regex, lua, bash, markdown, markdown_inline
@@ -25,11 +27,11 @@ with lib;
     };
 
     cmdline = {
-      enabled = helpers.defaultNullOpts.mkBool true "enables Noice cmdline UI";
-      view = helpers.defaultNullOpts.mkStr "cmdline_popup" "";
-      opts = helpers.defaultNullOpts.mkAttrsOf types.anything { } "";
+      enabled = defaultNullOpts.mkBool true "enables Noice cmdline UI";
+      view = defaultNullOpts.mkStr "cmdline_popup" "";
+      opts = defaultNullOpts.mkAttrsOf types.anything { } "";
       format =
-        helpers.defaultNullOpts.mkAttrsOf types.anything
+        defaultNullOpts.mkAttrsOf types.anything
           {
             cmdline = {
               pattern = "^:";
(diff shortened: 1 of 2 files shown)


MattSturgeon:
> But, there are tons of options without descriptions or not appropriately formatted descriptions, atm.

I haven't looked, but if these options are not too important and do not have any type coercion, perhaps we could consider dropping them and relying on the freeform type.

Especially given they do not even have documentation...

plugins/by-name/noice/default.nix
```diff
@@ -1,35 +1,103 @@
 {
   lib,
-  helpers,
-  config,
-  pkgs,
   ...
 }:
-# TODO: This uses a lot of types.anything because noice.nvim types are quite complex.
-# It should be possible to map them to nix, but they would not map really well through
-# toLuaObject, we would maybe need some ad-hoc pre-processing functions.
```
MattSturgeon, line 10: Is this comment still relevant? Or was it removed intentionally?

plugins/by-name/noice/default.nix
```diff
@@ -1,35 +1,103 @@
...
+  originalName = "noice.nvim";
+  package = "noice-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    `noice.nvim`, an experimental Neovim UI.
+
+    > [!NOTE] If treesitter is enabled you need the following parsers:
+    >  vim, regex, lua, bash, markdown, markdown_inline
```
MattSturgeon, line 20:
```suggestion
    > [!NOTE]
    > If treesitter is enabled you need the following parsers:
    > vim, regex, lua, bash, markdown, markdown_inline
```

I'm unsure if your initial syntax is valid, but elsewhere we have the admonition syntax on a separate line from the actual text body

plugins/by-name/noice/default.nix
```diff
@@ -1,35 +1,103 @@
...
 
-    package = lib.mkPackageOption pkgs "noice" {
-      default = [
-        "vimPlugins"
-        "noice-nvim"
+  # TODO: added 2024-10-27 remove after 24.11
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings =
+    let
+      mkOptionPaths = paths: map (p: lib.splitString "." p) paths;
```
MattSturgeon, line 27:
```suggestion
      mkOptionPaths = map (lib.splitString ".");
```

No need to specify the arg if it is passed as the final arg anyway. We can just use a partially applied function.

Same for the mapping function.

plugins/by-name/noice/default.nix
```diff
@@ -1,35 +1,103 @@
...
+        "presets"
+        "throttle"
+        "views"
+        "routes"
+        "status"
+        "format"
       ];
-    };
+    in
+    mkOptionPaths rawPaths;
```
MattSturgeon, line 81: Stylistic: you _could_ inline `rawPaths` here if you wanted.

MattSturgeon (commented):
Just some initial thoughts, as I don't have time to properly review right now.

plugins/by-name/noice/default.nix
khaneliman: Doh, you're right 
image

plugins/by-name/noice/default.nix
khaneliman: I guess I wasn't sure of its relevance anymore if we go down the route of allowing freeform settings and only type the ones that seem obvious. I'm double checking types.

plugins/by-name/noice/default.nix
```diff
@@ -165,78 +239,93 @@ with lib;
...
 
-        throttle = helpers.defaultNullOpts.mkNum (literalExpression "1000 / 30") "frequency to update lsp progress message";
+        format_done =
+          defaultNullOpts.mkNullableWithRaw (with types; either str anything) "lsp_progress_done"
+            ''
+              Lsp Progress is formatted using the builtins for lsp_progress.
+            '';
 
-        view = helpers.defaultNullOpts.mkStr "mini" "";
+        throttle = defaultNullOpts.mkNum (lib.literalExpression "1000 / 30") "Frequency to update lsp progress message.";
```
MattSturgeon, line 274: I think this is blamed on me, but I wonder why we're saying `1000 / 30` instead of the simpler `100 / 3`?

MattSturgeon (approved):
I don't see any issues after a quick read through the diff.

Thanks again :rocket:

plugins/by-name/noice/default.nix
khaneliman: I thought I saw this in the upstream default so maybe just respecting what they had?

MattSturgeon: Yeah, I guess so. Just seems strange to have the extra `0` given `(100 * N) / (3 * N) = 100 / 3 ≈ 33.33`

khaneliman:
@mergify queue

plugins/by-name/noice/default.nix
traxys: The context is that noice types are horrible to map to nix (mixed list/attrs, recursive,...)

With freeform options this is much less of an issue

MattSturgeon: > With freeform options this is much less of an issue

Unless we inadvertently impose type-restrictions by declaring settings options 🙃

PR #2468, opened 2024-10-29
flake/dev: add nixvim-init command (under 'init' shell command)

GaetanLepage:
A small helper to quickly add a new plugin.
This is inspired from [nix-init](https://github.com/nix-community/nix-init/).

It will be a small interactive program that prompts the name of the plugin, the package name and that will initialize the module and test files.

- [ ] Information prompting
- [ ] Generation of the module file `plugins/by-name/<NAME>/default.nix`
  - [ ] `mkVimPlugin` support
  - [ ] `mkNeovimPlugin` support
- [ ] Generation of the test file `tests/test-sources/plugins/by-name/<NAME>/default.nix`


flake-modules/dev/init-script/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    println!("Hello, world!");
```
khaneliman, line 2: oooh we doin rust now :)

GaetanLepage: I'm a huge noob, but I just want to do a simple `nix-init` clone for nixvim.

flake-modules/dev/init-script/default.nix
```diff
@@ -0,0 +1,13 @@
+{
+  pkgs,
+}:
+pkgs.rustPlatform.buildRustPackage {
+  pname = "nixvim-init";
+  version = "nightly";
+
+  src = ./.;
```
MattSturgeon, line 8: Generally, `src = ./.` is a bad idea because `default.nix` itself will also be considered a "source" of the derivation.

This means if `default.nix` is modified, the derivation needs to be rebuilt. Even if no source _actually_ changed and the change to the nix file wouldn't actually _need_ a rebuild.

Usually, you'd do something like `src = ./src`. I think you could also use a fileset or filtered source to filter out nix files, but those options seem a little overkill.

That said, this is a minor issue and we're already doing this in most of our other derivations, so it shouldn't block the PR if you'd rather keep it as-is.

flake-modules/dev/init-script/src/main.rs
MattSturgeon: Not opposed to learning a bit of rust for this, just conscious I already have to improve my python to properly be capable of maintaining our other in-tree scripts... :sweat_smile:

flake-modules/dev/init-script/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "nixvim-init"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+clap = "4.5.20"
```
traxys, line 7: I'd suggest using `clap = { version = "4.5.20", features = ["derive"] }`, it's quite a bit more ergonomic

GaetanLepage: Ok thanks ! Actually, I don't even think that we will need clap at all.
Good to know though.

MattSturgeon: > Actually, I don't even think that we will need clap at all.

I think we'll probably end wanting some decent argument parsing and help-text at some point...

Even if we have a mostly interactive UX it'd still be nice to be able to do things via CLI args.

flake-modules/dev/init-script/default.nix
MattSturgeon: I still prefer having sources and nix files separate, (I don't think it has any downsides?), but I won't press on the issue.

PR #2474, opened 2024-10-31
plugins/scrollview: init

### LINKED ISSUE #2467: [PLUGIN REQUEST] nvim-scrollview
| Field    | Description |
|----------|-------------|
| Plugin   | `nvim-scrollview`  | 
| Homepage | https://github.com/dstein64/nvim-scrollview     | 
| Nixpkgs  | https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/editors/vim/plugins/generated.nix#L9212 | 

## Extra Information

nvim-scrollview is a Neovim plugin that displays interactive vertical scrollbars and signs. The plugin is customizable (see :help scrollview-configuration).

GaetanLepage:
Add [nvim-scrollview](https://github.com/dstein64/nvim-scrollview), a plugin that displays interactive vertical scrollbars and signs.

Fixes #2467.

### DIFF
--- a/plugins/by-name/scrollview/default.nix
+++ b/plugins/by-name/scrollview/default.nix
@@ -1,0 +1,439 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts mkNullOrStr mkNullOrOption;
+  inherit (lib) types;
+in
+lib.nixvim.vim-plugin.mkVimPlugin {
+  name = "scrollview";
+  originalName = "nvim-scrollview";
+  package = "nvim-scrollview";
+  globalPrefix = "scrollview_";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  settingsOptions =
+    let
+      mkPriorityOption =
+        name: default:
+        defaultNullOpts.mkUnsignedInt default ''
+          The priority for the ${name}.
+
+          Considered only when the plugin is loaded.
+        '';
+
+      symbolCommonDesc = ''
+        A list of strings can also be used; the symbol is selected using the index equal to the
+        number of lines corresponding to the sign, or the last element when the retrieval would be
+        out-of-bounds.
+
+        Considered only when the plugin is loaded.
+      '';
+
+      symbolType = with types; maybeRaw (either str (listOf (maybeRaw str)));
+    in
+    {
+      always_show = defaultNullOpts.mkBool false ''
+        Specify whether scrollbars and signs are shown when all lines are visible.
+      '';
+
+      base =
+        defaultNullOpts.mkEnumFirstDefault
+          [
+            "right"
+            "left"
+            "buffer"
+          ]
+          ''
+            Specify where the scrollbar is anchored.
+
+            Possible values are `"left"` or `"right"` for corresponding window edges, or `"buffer"`.
+          '';
+
+      byte_limit = defaultNullOpts.mkInt 1000000 ''
+        The buffer size threshold (in bytes) for entering restricted mode, to prevent slow operation.
+
+        Use `-1` for no limit.
+      '';
+
+      character = defaultNullOpts.mkStr "" ''
+        A character to display on scrollbars.
+
+        Scrollbar transparency (via `|scrollview_winblend|`) is not possible when a scrollbar
+        character is used.
+      '';
+
+      column = defaultNullOpts.mkPositiveInt 1 ''
+        The scrollbar column (relative to `|scrollview_base|`).
+
+        Must be an integer greater than or equal to `1`.
+      '';
+
+      consider_border = defaultNullOpts.mkBool false ''
+        (experimental)
+
+        Specify whether floating window borders are taken into account for positioning scrollbars and
+        signs.
+        - When set to `true`, borders are considered as part of the window; scrollbars and signs can
+          be positioned where there are borders.
+        - When set to `false`, borders are ignored.
+
+        The setting is only applicable for floating windows that have borders, and only relevant when
+        `floating_windows` is turned on and `base` is set to `"left"` or `"right"`.
+      '';
+
+      current_only = defaultNullOpts.mkBool false ''
+        Whether scrollbars and signs should only be displayed in the current window.
+      '';
+
+      excluded_filetypes = defaultNullOpts.mkListOf types.str [ ] ''
+        optional file types for which scrollbars should not be displayed.
+      '';
+
+      floating_windows = defaultNullOpts.mkBool false ''
+        Whether scrollbars and signs are shown in floating windows.
+      '';
+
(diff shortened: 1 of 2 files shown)


GaetanLepage:
I might have leaned towards the exhaustive side...

plugins/by-name/scrollview/default.nix
```diff
@@ -0,0 +1,444 @@
...
+        The symbol for diagnostic hint signs.
+
+        Defaults to the trimmed sign text for `DiagnosticSignHint` if defined, or `"H"` otherwise.
+
+        ${symbolCommonDesc}
+      '';
+
+      diagnostics_info_priority = mkPriorityOption "the diagnostic info signs" 40;
+
+      diagnostics_info_symbol = mkNullOrOption symbolType ''
```
MattSturgeon, line 313: IIRC, `mkNullOrOption` is one of the weird special cases that don't inject `maybeRaw` support? If so there may be a `mkNullOrOptionWithRaw` variant or you could manually add `rawLua` or `maybeRaw` support.

MattSturgeon (commented):
Looks good. Very minor, but a few options I think don't have proper `rawLua` support?

I don't see any obvious issues otherwise :grin:

plugins/by-name/scrollview/default.nix
GaetanLepage: You are right, I have fixed that and adjusted the test too.

plugins/by-name/scrollview/default.nix
```diff
@@ -0,0 +1,450 @@
...
+
+      loclist_symbol = mkNullOrOption symbolType ''
+        The symbol for loclist signs.
+
+        Defaults to a small circle.
+
+        ${symbolCommonDesc}
+      '';
+
+      marks_characters = mkNullOrOption (with types; either rawLua (listOf str)) ''
```
MattSturgeon, line 383:
```suggestion
      marks_characters = mkListOf types.str ''
```

plugins/by-name/scrollview/default.nix
```diff
@@ -0,0 +1,450 @@
...
+
+      conflicts_top_symbol = defaultNullOpts.mkNullable symbolType "<" ''
+        The symbol for conflict top signs.
+
+        ${symbolCommonDesc}
+      '';
+
+      cursor_priority = mkPriorityOption "cursor signs" 0;
+
+      cursor_symbol = mkNullOrStr ''
```
MattSturgeon, line 291:
```suggestion
      cursor_symbol = defaultNullOpts.mkStr ''
```

IIRC, mkNullOrStr doesn't inject rawLua support?

Perhaps you're using the non-defaultNullOpts helper so that you don't need to specify a pluginDefault? I wonder if it'd be better to use the `mkStr'` "prime" variant to have more flexibile arguments then?

plugins/by-name/scrollview/default.nix
```diff
@@ -0,0 +1,450 @@
...
+
+      changelist_current_priority = mkPriorityOption "the current changelist sign" 10;
+
+      changelist_current_symbol = defaultNullOpts.mkStr "@" ''
+        The symbol for the current item changelist sign.
+      '';
+
+      changelist_next_priority = mkPriorityOption "the next item changelist sign" 5;
+
+      changelist_next_symbol = mkNullOrStr ''
```
MattSturgeon, line 259: Same

plugins/by-name/scrollview/default.nix
GaetanLepage: > IIRC, mkNullOrStr doesn't inject rawLua support?

[I checked and it does](https://github.com/nix-community/nixvim/blob/42ea1626cb002fa759a6b1e2841bfc80a4e59615/lib/options.nix#L90).

> Perhaps you're using the non-defaultNullOpts helper so that you don't need to specify a pluginDefault?

Yes, this is the reason.



> I wonder if it'd be better to use the `mkStr'` "prime" variant to have more flexibile arguments then?

Sure, but wouldn't this be more verbose for nothing ?

plugins/by-name/scrollview/default.nix
MattSturgeon: I think this one is still relevant, `maybeRaw (listOf (maybeRaw str)` is produced by `defaultNullOpts.mkListOf str`.

If you don't want a pluginDefault you could use `mkListOf'`

```nix
defaultNullOpts.mkListOf' {
  type = types.str;
  description = "";
}
```

If you really don't want that, you can add maybeRaw to the element type yourself: `maybeRaw (listOf (maybeRaw str)`.

GaetanLepage: I went with the latter as for me the goal of `defaultNullOpts` helper is to provide a plugin default.
Using them and deleting their purpose (the plugin default) doesn't seem right to me.

To be honest, your solution is elegant too. I don't feel too hard about it.

MattSturgeon: > the goal of `defaultNullOpts` helper is to provide a plugin default.

I disagree with that sentiment, but I don't want to get into a debate here :grin: 

(I think that was the original purpose, but they have evolved over time to relate to anything that is maybeRaw and defaults to null)

MattSturgeon (approved):
Awesome, thanks! :tada:

plugins/by-name/scrollview/default.nix
GaetanLepage: Actually, I'm curious, I changed it to `mkListOf` but I can't manage to get the _Plugin default: `null`_ out of the documentation.

GaetanLepage:
```nix
      marks_characters = defaultNullOpts.mkListOf' {
        type = types.str;
        # default = null;
        description = ''
          List of strings specifying characters for which mark signs will be shown.
          Defaults to characters `a-z` and `A-Z`.

          Considered only when the plugin is loaded.
        '';
      };
```

MattSturgeon: > but I can't manage to get the _Plugin default: `null`_ out of the documentation.

Interesting... That sounds like a bug I should investigate :thinking:

GaetanLepage: I guess I will merge as is then.

GaetanLepage: My apologies. I think that I got tricked by `direnv`. The `mkListOf'` implementation does work fine (and does not show _plugin default: `null`_):
image

Sorry for the false alarm.

MattSturgeon (approved):
Thanks for being thorough!

GaetanLepage:
> Thanks for being thorough!

Thanks for the careful review :)

GaetanLepage:
@mergify queue

PR #2479, opened 2024-11-04
plugins/luasnip: fix example

amalgame21:
Fix `fromLua` example in this page:
https://nix-community.github.io/nixvim/plugins/luasnip/fromLua.html


MattSturgeon (approved):
Great, thanks!

PR #2481, opened 2024-11-04
plugins/blink-cmp: update configuration

Balssh:
Update the default configuration and add some tests.

What I'm not sure is if it's ok to keep the keymap type as it is, given in the [plugin's default config](https://github.com/Saghen/blink.cmp) is like `keymap = 'default'`, but it can also be like:
```lua
keymap = {
   preset = 'default',
   ['<Up>'] = { 'select_prev', 'fallback' },
   ['<Down>'] = { 'select_next', 'fallback' },
},
```
Also the test I added I'm not sure is being run with `tests -i -> plugins-by-name-blink-cmp-default`


plugins/by-name/blink-cmp/default.nix
```diff
@@ -21,23 +21,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
-            "<C-n>"
-          ];
-          show_documentation = "<C-space>";
-          hide_documentation = "<C-space>";
-          scroll_documentation_up = "<C-b>";
-          scroll_documentation_down = "<C-f>";
-          snippet_forward = "<Tab>";
-          snippet_backward = "<S-Tab>";
+          preset = "default";
         }
```
MattSturgeon, line 25: Maybe this would benefit from an `example`?

This can be done using `defaultNullOpts.mkNullableWithRaw'`, e.g.

```nix
defaultNullOpts.mkNullableWithRaw' {
  type = with types; either (attrsOf anything) (enum [ false ]);
  pluginDefault = { preset = "default"; };
  example = {
    # Some example! 😁
  };
  description = ''
    # Some description!
  '';
}
```

---

Also: is `enum [false]` still correct?

Balssh: Makes sense, forgot the example part. Will correct this tomorrow, thanks!

plugins/by-name/blink-cmp/default.nix
```diff
@@ -61,16 +45,18 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+        auto_show = defaultNullOpts.mkBool false ''
+          Enables automatically showing documentation when typing.
+        '';
+        auto_show_delay_ms = defaultNullOpts.mkUnsignedInt 500 ''
+          Delay, in milliseconds, after which documentation is shown.
+        '';
+        update_delay_ms = defaultNullOpts.mkUnsignedInt 50 ''
+          Delay, in milliseconds, after which documentation is updated.
+        '';
+      };
```
MattSturgeon, line 59: While not strictly necessary, it'd be nice to have some warnings added when the old attrs are defined in `cfg.settings`...

@khaneliman has been spearheading this technique. Here's an example of a recent deprecation he made (although for several more options than we need here): https://github.com/nix-community/nixvim/blob/aa06b176e78c9ae9e779e605cab61c9d8681a54e/plugins/by-name/rest/default.nix#L380-L428

MattSturgeon: In this scenario a warning would alert users who had followed our docs and defined `settings.documentation` that they should actually be using `settings.window.documentation`.

It could probably be as simple as 
```nix
warnings = lib.optional (cfg.settings ? documentation) ''
  Nixvim(plugins.blink): some warning about using settings.window.documentation instead...
'';
```

plugins/by-name/blink-cmp/default.nix
```diff
@@ -18,30 +18,62 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
-          scroll_documentation_down = "<C-f>";
-          snippet_forward = "<Tab>";
-          snippet_backward = "<S-Tab>";
-        }
-        ''
-          Customize the keymaps for blink.
-        '';
+    keymap = defaultNullOpts.mkNullableWithRaw' {
+      type = with types; either (attrsOf anything) (enum [ false ]);
+      pluginDefault = "default";
```
MattSturgeon, line 23:
```suggestion
      pluginDefault = {
        preset = "default";
      };
```

plugins/by-name/blink-cmp/default.nix
```diff
@@ -18,30 +18,62 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+        ];
+        "<S-Tab>" = [
+          "snippet_backward"
+          "fallback"
+        ];
+
+      };
+      description = ''
+        The keymap can be:
+        - A preset ('default' | 'super-tab' | 'enter')
```
MattSturgeon, line 71:
```suggestion
        - A preset (`'default'` | `'super-tab'` | `'enter'`)
```

Using markdown's `` `backticks` ``, we can mark this as a code snippet

plugins/by-name/blink-cmp/default.nix
```diff
@@ -18,30 +18,62 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+        - A table of keys => command[] (optionally with a "preset" key to merge with a preset)
```
MattSturgeon, line 72:
```suggestion
        - A table of `keys => command[]` (optionally with a "preset" key to merge with a preset)
```

plugins/by-name/blink-cmp/default.nix
```diff
@@ -131,4 +155,11 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
       };
     };
   };
+
+  extraConfig = cfg: {
+    warnings = lib.optional (cfg.settings ? documentation) ''
+      Nixvim(plugins.blink): settings.documentation was not properly defined, use settings.windows.documentation
+      instead
```
MattSturgeon, line 162:
```suggestion
      Nixvim(plugins.blink): `settings.documentation` does not correspond to a known setting, use `settings.windows.documentation` instead.
```

1. Slight re-wording suggestion (if you agree)
2. Ideally, put everything on one line

Warnings get printed verbatum to stdout, without any markdown formatting. Therefore (unlike option descriptions) we can't rely on markdown to un-wrap lines that we would prefer to have wrapped, if we don't want them to be printed on multiple lines.

MattSturgeon (commented):
Looks good. I just have a couple minor suggestions

Balssh:
Implemented your requests.

MattSturgeon (approved):
Great thanks, LGTM! :tada:

MattSturgeon:
@mergify queue

plugins/by-name/blink-cmp/default.nix
MattSturgeon: Ah, I see I misinterpreted the upstream docs! I think I was looking at an example instead of the default value...

Luckily this is _effectively_ the same, but your original `default = 'default'` was actually more correct. Apologies!

Balssh: Don't worry, it does look better with the way you suggested and it also works. Appreciate the follow up!

PR #2486, opened 2024-11-07
plugins/lsp: enable auto-installing rustfmt

AxerTheAxe:
Adds an option to automatically install rustfmt when using rust-analyzer. This is a draft as I was not sure exactly how to define the option so any suggestions on how to better implement it are welcome. Thanks!

### DIFF
--- a/plugins/lsp/language-servers/rust-analyzer.nix
+++ b/plugins/lsp/language-servers/rust-analyzer.nix
@@ -30,6 +30,16 @@ in
 
     # TODO: make nullable
     rustcPackage = mkPackageOption pkgs "rustc" { };
+
+    installRustfmt = mkOption {
+      type = with types; nullOr bool;
+      default = null;
+      example = true;
+      description = "Whether to install `rustfmt`.";
+    };
+
+    # TODO: make nullable
+    rustfmtPackage = mkPackageOption pkgs "rustfmt" { };
   };
   config = mkIf cfg.enable {
     warnings =
@@ -56,7 +66,8 @@ in
 
     extraPackages =
       with pkgs;
-      (optional ((isBool cfg.installCargo) && cfg.installCargo) cfg.cargoPackage)
-      ++ (optional ((isBool cfg.installRustc) && cfg.installRustc) cfg.rustcPackage);
+      optional (isBool cfg.installCargo && cfg.installCargo) cfg.cargoPackage
+      ++ optional (isBool cfg.installRustc && cfg.installRustc) cfg.rustcPackage
+      ++ optional (isBool cfg.installRustfmt && cfg.installRustfmt) cfg.rustfmtPackage;
   };
 }


AxerTheAxe:
Thanks for the feedback. How about now?

plugins/lsp/language-servers/rust-analyzer.nix
```diff
@@ -30,6 +30,16 @@ in
...
+
+    installRustfmt = mkOption {
+      type = with types; nullOr bool;
+      default = null;
+      example = true;
+      description = "Whether to install `rustfmt`.";
+    };
+
+    # TODO: make nullable
+    rustfmtPackage = mkPackageOption pkgs "rustfmt" { };
```
MattSturgeon, line 42: Any reason this shouldn't be nullable yet?


```suggestion
    rustfmtPackage = mkPackageOption pkgs "rustfmt" {
      nullable = true;
    };
```

plugins/lsp/language-servers/rust-analyzer.nix
```diff
@@ -57,6 +67,7 @@ in
     extraPackages =
       with pkgs;
       (optional ((isBool cfg.installCargo) && cfg.installCargo) cfg.cargoPackage)
-      ++ (optional ((isBool cfg.installRustc) && cfg.installRustc) cfg.rustcPackage);
+      ++ (optional ((isBool cfg.installRustc) && cfg.installRustc) cfg.rustcPackage)
+      ++ (optional ((isBool cfg.installRustfmt) && cfg.installRustfmt) cfg.rustfmtPackage);
```
MattSturgeon, line 71: It pains me to see so many unnecessary parentheses, but it matches the style of the existing code 😅

Just to illustrate a point, this is equivalent to:
```nix
optional (isBool cfg.installCargo && cfg.installCargo) cfg.cargoPackage
++ optional (isBool cfg.installRustc && cfg.installRustc) cfg.rustcPackage
++ optional (isBool cfg.installRustfmt && cfg.installRustfmt) cfg.rustfmtPackage;
```

plugins/lsp/language-servers/rust-analyzer.nix
MattSturgeon: Ah, I guess this is because we have the `install` option too? So the package option being nullable is kinda redundant...

At least recently, we typically skip the install-toggle-options and instead just make the installable thing nullable. Either null by default with a useful example, or set to something by default but documented as nullable.

In this case I don't have a strong preference, since it matches the style of this module's other options.

MattSturgeon (approved):
I've left comments above, but I don't have any objections to this as is. Thanks for your contribution!

khaneliman (approved):
LGTM overall, I agree with Matt's points if you wanted to go above and beyond and clean stuff up. But, I won't hold this back for those.

AxerTheAxe:
Thanks! I removed the redundant parentheses. I’m not exactly sure what changes are being suggested for the nullable packages; is the plan to remove the installation flags and install if the package options are set to nonnull? I'd be happy to create another PR for that if you could clarify the desired changes :)

khaneliman:
> Thanks! I removed the redundant parentheses. I’m not exactly sure what changes are being suggested for the nullable packages; is the plan to remove the installation flags and install if the package options are set to nonnull? I'd be happy to create another PR for that if you could clarify the desired changes :)

Yeah, I think we'd have to rework this plugin's options. Right now, we have an option that says whether or not to install the extra package and then the package option. But, we could change it to be just the package option that is nullable. If a user doesn't want to install the package, they would set it to null.

MattSturgeon:
Agreed, further cleanup should be done in a dedicated pr. Don't feel pressured to work on that if you don't want to as I'm sure we'll get to it eventually 😅

When we do tackle it we'll have to consider how best to deprecate the enable options, without introducing and immediately breaking changes...

MattSturgeon:
@mergify queue

AxerTheAxe:
Ok! I am ready for this to be merged whenever.

PR #2488, opened 2024-11-08
plugins/prettierd: add disableTsServerFormatter like for prettier

Pyxels:
Add a new option (and warning) to disable the typescript-language-server (ts_ls) formatting just like with prettier.

```nix
none-ls = {
  enable = true;

  sources = {
    formatting.prettierd = {
      enable = true;
      disableTsServerFormatter = true;
    };
  };
};
```

I simply copied the implementation from prettier, to keep the source code more readable. If you'd like me to make this more generic instead I could create a function that reduces repetition.

### DIFF
--- a/plugins/by-name/none-ls/prettier.nix
+++ b/plugins/by-name/none-ls/prettier.nix
@@ -1,38 +1,41 @@
-{ lib, config, ... }:
+{
+  lib,
+  config,
+  options,
+  ...
+}:
 let
   cfg = config.plugins.none-ls.sources.formatting.prettier;
   ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+  opt = options.plugins.none-ls.sources.formatting.prettier;
+  defaultPrio = (lib.mkOptionDefault null).priority;
 in
 {
   options.plugins.none-ls.sources.formatting.prettier = {
     disableTsServerFormatter = lib.mkOption {
-      type = with lib.types; nullOr bool;
+      type = lib.types.bool;
       description = ''
         Disables the formatting capability of the `ts_ls` language server if it is enabled.
       '';
-      default = null;
+      default = false;
       example = true;
     };
   };
 
   config = lib.mkIf cfg.enable {
-    warnings = lib.optional ((cfg.disableTsServerFormatter == null) && ts-ls-cfg.enable) ''
-      You have enabled the `prettier` formatter in none-ls.
-      You have also enabled the `ts_ls` language server which also brings a formatting feature.
+    warnings =
+      lib.optional (opt.disableTsServerFormatter.highestPrio == defaultPrio && ts-ls-cfg.enable)
+        ''
+          You have enabled the `prettier` formatter in none-ls.
+          You have also enabled the `ts_ls` language server which also brings a formatting feature.
 
-      - To disable the formatter built-in the `ts_ls` language server, set
-        `plugins.none-ls.sources.formatting.prettier.disableTsServerFormatter` to `true`.
-      - Else, to silence this warning, explicitly set the option to `false`.
-    '';
+          - To disable the formatter built-in the `ts_ls` language server, set
+            `plugins.none-ls.sources.formatting.prettier.disableTsServerFormatter` to `true`.
+          - Else, to silence this warning, explicitly set the option to `false`.
+        '';
 
     plugins.lsp.servers.ts_ls =
-      lib.mkIf
-        (
-          cfg.enable
-          && ts-ls-cfg.enable
-          && (lib.isBool cfg.disableTsServerFormatter)
-          && cfg.disableTsServerFormatter
-        )
+      lib.mkIf (cfg.enable && ts-ls-cfg.enable && cfg.disableTsServerFormatter)
         {
           onAttach.function = ''
             client.server_capabilities.documentFormattingProvider = false
--- a/plugins/by-name/none-ls/prettierd.nix
+++ b/plugins/by-name/none-ls/prettierd.nix
@@ -1,0 +1,45 @@
+{
+  lib,
+  config,
+  options,
+  ...
+}:
+let
+  cfg = config.plugins.none-ls.sources.formatting.prettierd;
+  ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+  opt = options.plugins.none-ls.sources.formatting.prettierd;
+  defaultPrio = (lib.mkOptionDefault null).priority;
+in
+{
+  options.plugins.none-ls.sources.formatting.prettierd = {
+    disableTsServerFormatter = lib.mkOption {
+      type = lib.types.bool;
+      description = ''
+        Disables the formatting capability of the `ts_ls` language server if it is enabled.
+      '';
+      default = false;
+      example = true;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    warnings =
+      lib.optional (opt.disableTsServerFormatter.highestPrio == defaultPrio && ts-ls-cfg.enable)
+        ''
+          You have enabled the `prettierd` formatter in none-ls.
+          You have also enabled the `ts_ls` language server which also brings a formatting feature.
+
+          - To disable the formatter built-in the `ts_ls` language server, set
+            `plugins.none-ls.sources.formatting.prettierd.disableTsServerFormatter` to `true`.
+          - Else, to silence this warning, explicitly set the option to `false`.
+        '';
+
+    plugins.lsp.servers.ts_ls =
+      lib.mkIf (cfg.enable && ts-ls-cfg.enable && cfg.disableTsServerFormatter)
+        {
+          onAttach.function = ''
+            client.server_capabilities.documentFormattingProvider = false
+          '';
+        };
+  };
+}
--- a/plugins/by-name/none-ls/sources.nix
+++ b/plugins/by-name/none-ls/sources.nix
@@ -5,7 +5,10 @@ let
 in
 {
   imports =
-    [ ./prettier.nix ]
+    [
+      ./prettier.nix
+      ./prettierd.nix
+    ]
     ++ (lib.flatten (
       lib.mapAttrsToList (category: (lib.map (mkSourcePlugin category))) noneLsBuiltins
     ));


plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,42 @@
+{ lib, config, ... }:
+let
+  cfg = config.plugins.none-ls.sources.formatting.prettierd;
+  ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+in
+{
+  options.plugins.none-ls.sources.formatting.prettierd = {
+    disableTsServerFormatter = lib.mkOption {
+      type = with lib.types; nullOr bool;
```
MattSturgeon, line 9: Why use `nullOr` here? Isn't it clearer to end-users what the default behaviour is if we just have a plain `bool` type?

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,42 @@
+{ lib, config, ... }:
+let
+  cfg = config.plugins.none-ls.sources.formatting.prettierd;
+  ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+in
+{
+  options.plugins.none-ls.sources.formatting.prettierd = {
+    disableTsServerFormatter = lib.mkOption {
```
MattSturgeon, line 8: I'm not a fan of negatively-named booleans. I think it'd be better to have a `enable` option instead of a `disable` option.

It's fine to have a default-true enable option instead of a default-false disable option, for example :slightly_smiling_face:

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,42 @@
...
+        `plugins.none-ls.sources.formatting.prettierd.disableTsServerFormatter` to `true`.
+      - Else, to silence this warning, explicitly set the option to `false`.
+    '';
+
+    plugins.lsp.servers.ts_ls =
+      lib.mkIf
+        (
+          cfg.enable
+          && ts-ls-cfg.enable
+          && (lib.isBool cfg.disableTsServerFormatter)
```
MattSturgeon, line 33: Note: the parentheses aren't needed here, since function application takes precedent over most other operations, including `&&` operators.

However if we dropped the `nullOr` type then this line wouldn't be needed at all :grin:

MattSturgeon (commented):
This looks useful, thanks! I just have a few minor concerns below :slightly_smiling_face:

plugins/by-name/none-ls/prettierd.nix
Pyxels: Done!

plugins/by-name/none-ls/prettierd.nix
Pyxels: `nullOr bool` allows silencing the warning about `ts_ls` formatting conflict.

plugins/by-name/none-ls/prettierd.nix
Pyxels: Since this is a copy of the implementation for `prettier`, I kept the same naming scheme. We could diverge for this plugin, but I would suggest keeping the naming the consistent, since `prettierd` is "basically the same".

MattSturgeon: My preference would be to diverge here. Ideally, poor decisions in the past shouldn't influence decisions made today :upside_down_face: 

In separate PRs we could add rename (and/or alias) options for any existing negative booleans.

I don't feel too strongly though. As they're related I'd also be OK with being consistent, it just means there will be _more_ to deprecate/rename if we ever get around to fixing the other option names...

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  config = lib.mkIf cfg.enable {
+    warnings = lib.optional ((cfg.disableTsServerFormatter == null) && ts-ls-cfg.enable) ''
+      You have enabled the `prettierd` formatter in none-ls.
+      You have also enabled the `ts_ls` language server which also brings a formatting feature.
+
+      - To disable the formatter built-in the `ts_ls` language server, set
+        `plugins.none-ls.sources.formatting.prettierd.disableTsServerFormatter` to `true`.
+      - Else, to silence this warning, explicitly set the option to `false`.
+    '';
```
MattSturgeon, line 26: If this is the only reason we have `nullOr`, we can also use option priorities.

option-defaults have prio `1500` (aka `(lib.mkOptionDefault null).priority`), so we can check that a higher priority (well, lower number) is set for the option's _highest_ priority:

```nix
# Get access to this modules _options_ attrs
opt = options.plugins.none-ls.sources.formatting.prettierd;

# either
defaultPrio = (lib.mkOptionDefault null).priority;
# or
defaultPrio = 1500;

# Check isDefault instead of isNull
isDefault = opt.disableTsServerFormatter.highestPrio == defaultPrio;
```

_When designing the existing options, we probably didn't know about this technique..._

plugins/by-name/none-ls/prettierd.nix
MattSturgeon: Ah, I see. I've addressed this at https://github.com/nix-community/nixvim/pull/2488/files#r1838093274

plugins/by-name/none-ls/prettierd.nix
Pyxels: I'm sorry, I can't come up with a proper default-true name.

`enableTsServerFormatter` seems confusing, since this option only makes sense if the plugin `ts_ls` is also enabled, which IMO is not the base case.

I'd argue for keeping `disableTsServerFormatter`, but mostly since I can't come up with anything different :D

MattSturgeon (approved):
LGTM. Some minor suggestions below might be nice to have, but aren't important.

Thanks for your persistence working on this!

plugins/by-name/none-ls/prettierd.nix
MattSturgeon: Sure thing. I won't block on a minor naming issue that will need addressing _anyway_ for the other related option.

I'll give other maintainers chance to chime in too, in case they have any opinions or ideas.

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,48 @@
...
+      default = false;
+      example = true;
+    };
+  };
+
+  config =
+    let
+      opt = options.plugins.none-ls.sources.formatting.prettierd;
+      defaultPrio = (lib.mkOptionDefault null).priority;
+    in
```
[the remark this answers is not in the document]
Pyxels: Sure.
_done_

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,48 @@
...
+  ...
+}:
+let
+  cfg = config.plugins.none-ls.sources.formatting.prettierd;
+  ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+in
+{
+  options.plugins.none-ls.sources.formatting.prettierd = {
+    disableTsServerFormatter = lib.mkOption {
+      type = with lib.types; bool;
```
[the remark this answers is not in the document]
Pyxels: You're right, I missed the types one. _done_

Pyxels:
I also refactored the `prettier.nix` config to use the same priority system instead of checking against `null`. This should only be an issue if somebody has explicitly set their option to `null`, otherwise it should work the same.

plugins/by-name/none-ls/prettier.nix
```diff
@@ -1,38 +1,41 @@
...
+}:
 let
   cfg = config.plugins.none-ls.sources.formatting.prettier;
   ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+  opt = options.plugins.none-ls.sources.formatting.prettier;
+  defaultPrio = (lib.mkOptionDefault null).priority;
 in
 {
-  options.plugins.none-ls.sources.formatting.prettier = {
+  opt = {
```
MattSturgeon, line 14: This is wrong. It gets a little confusing, but `opts` is a local variable we've created that gets its value from an attr in the `options` module argument.

Here we're _defining_ an `options` that is part of our module outputs.

This change makes it so that we actually define a literal `opts` attr instead of an `options` attr :sweat_smile:

plugins/by-name/none-ls/prettierd.nix
```diff
@@ -0,0 +1,45 @@
...
+  ...
+}:
+let
+  cfg = config.plugins.none-ls.sources.formatting.prettierd;
+  ts-ls-cfg = config.plugins.lsp.servers.ts_ls;
+  opt = options.plugins.none-ls.sources.formatting.prettierd;
+  defaultPrio = (lib.mkOptionDefault null).priority;
+in
+{
+  opt = {
```
MattSturgeon, line 14: Same

plugins/by-name/none-ls/prettier.nix
```diff
@@ -1,38 +1,41 @@
...
     disableTsServerFormatter = lib.mkOption {
-      type = with lib.types; nullOr bool;
+      type = lib.types.bool;
```
MattSturgeon, line 16: > I also refactored the `prettier.nix` config to use the same priority system instead of checking against `null`. This should only be an issue if somebody has explicitly set their option to `null`, otherwise it should work the same.
> 
> Let me know if you want that reversed.

If we wanted to have some kinda graceful transition period we'd have to create a type along the lines of:
```nix
# A nullOr bool that pretends it is just a bool in its description
fakeBoolType = lib.types.nullOr lib.types.bool // {
  description = lib.types.bool.description;
}
```
and also have a separate warning for when the option is null.

In this case, I don't think that's worth the effort because (as you say) no one should be explicitly setting this option to null.

If anyone is explicitly setting it to null, they should get a fairly helpful error about using an incorrect type for this option.

MattSturgeon:
Also, minor nit:

I (just) noticed you're using conventional commits (which is great), but in nixvim we tend to use path/scope based commit messages instead. See the [file history for nonels](https://github.com/nix-community/nixvim/commits/7cd54a15f01377b2d7624a8360860ad25c3eef10/plugins/by-name/none-ls) for some prior examples.

This is the same format as you're already using in your PR title :grin: 

In this case, I might go with something like ``plugins/nonels/prettierd: add disableTsServerFormatter option`` and ``plugins/nonels/prettier: use option prio for warning`` ?

plugins/by-name/none-ls/prettier.nix
Pyxels: My bad!
_fixed_

MattSturgeon (approved):
Still no issues from me. :rocket: 

Just waiting to see if anyone else has any input on anything before merging (especially https://github.com/nix-community/nixvim/pull/2488/files#r1837089738).

plugins/by-name/none-ls/prettierd.nix
khaneliman: If we really wanted to not have a negative name, we could do `overrideTsServerFormatter`. But, I'm fine keeping it as-is until we migrate none-ls to new standards.

MattSturgeon:
@mergify queue

MattSturgeon:
Thanks again @Pyxels for this useful addition and related cleanup work!

Pyxels:
Thanks for the all the friendly guidance and extensive explanations @MattSturgeon !
I'm always happy to learn and the nits were very helpful in that regard.

Issue #2367, opened 2024-10-05
[BUG] jupytext not found when running with jupytext-nvim


jalil-salame:
| Field        | Description |
|--------------|-------------|
| Plugin       | `jupytext`  | 
| Nixpkgs      | `24.05 (some packages come from unstable)`     | 
| Home Manager | `24.05`     | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

When installing the `jupytext` plugin and trying to open a `*.ipynb` file, this fails with `jupytext ... : 127` (command not found)

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  plugins.jupytext.enable = true;
}
```

## Other things I've tried

I have tried:
- setting `programs.nixvim.extraPython3Packages = p: [p.jupytext];` (same error)
- entering a devshell with `python3.jupytext` available (works properly)

bkp5190:
Looks like you may need this:

https://search.nixos.org/packages?channel=24.05&show=python312Packages.jupytext&from=0&size=50&sort=relevance&type=packages&query=jupytext

~~~
home.packages = with pkgs; [
      python312Packages.jupytext
];
~~~

jalil-salame:
No need for homeManager, you can fix it inside `nixvim`: ` extraPackages = [ (pkgs.python3.withPackages (p: [ p.jupytext ])) ];`.

Still, the plugin should do this automatically.

bkp5190:
Merged a change for this dependency similar to other packages, can you retest this @jalil-salame ?

jalil-salame:
Thanks for the help! Sadly the fix doesn't work:

```
zsh:1: command not found: jupytext
Error detected while processing BufReadCmd Autocommands for "*.ipynb":
jupytext '/home/jalil/Dev/jupyter-test/test.ipynb' --to=md:markdown --output='/home/jalil/Dev/jupyter-test/test.md' : 127
Error executing lua callback: Vim:E484: Can't open file /home/jalil/Dev/jupyter-test/test.md
stack traceback:
	[C]: in function 'readfile'
	...ckages/start/vimplugin-plugin-pack/lua/jupytext/init.lua:90: in function 'read_from_ipynb'
	...ckages/start/vimplugin-plugin-pack/lua/jupytext/init.lua:176: in function <...ckages/start/vimplugin-plugin-pack/lua/jupytext/init.lua:175>
```

[...]

PR #2490, opened 2024-11-08
Add jupytext python3Dependency to jupytext plugin

bkp5190:
# Description
Add jupytext python3 Dependency for the juptext plugin

#2367

### DIFF
--- a/plugins/by-name/jupytext/default.nix
+++ b/plugins/by-name/jupytext/default.nix
@@ -78,4 +78,6 @@ helpers.neovim-plugin.mkNeovimPlugin {
     };
   };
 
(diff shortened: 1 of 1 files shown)


traxys (commented):
I think that by itself this does nothing, you need to add the value to the `extraPython3Packages` global option. I think molten.nvim does that

bkp5190:
Ah I see, thank you! Looks like I was missing a portion of the change. Does that override seem correct? - Similar to molten.nvim

GaetanLepage (approved):
LGTM, thanks !

traxys (approved):

traxys:
@mergifyio queue

Issue #2491, opened 2024-11-08
[PLUGIN REQUEST] quarto-nvim


bkp5190:
| Field    | Description |
|----------|-------------|
| Plugin   | `quarto-nvim`  |
| Homepage | `https://github.com/quarto-dev/quarto-nvim`     |
| Nixpkgs  | `true` | 

## Extra Information

Currently, quarto-nvim is available in nixpkgs. It would be nice to also have it available and configurable in nixvim

PR #2492, opened 2024-11-08
plugins/quarto: init

bkp5190:
# Description
Adding quarto-nvim

# Issue
#2491

### DIFF
--- a/plugins/by-name/quarto/default.nix
+++ b/plugins/by-name/quarto/default.nix
@@ -1,0 +1,101 @@
+{ lib, pkgs, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "quarto";
+  originalName = "quarto-nvim";
+  package = "quarto-nvim";
+
+  maintainers = [ lib.maintainers.BoneyPatel ];
+
+  settingsOptions = {
+    debug = defaultNullOpts.mkBool false ''
+      Enables or disables debugging.
+    '';
+
+    closePreviewOnExit = defaultNullOpts.mkBool true ''
+      Closes preview on exit.
+    '';
+
+    lspFeatures = {
+      enabled = defaultNullOpts.mkBool true ''
+        Enables LSP features.
+      '';
+
+      chunks = defaultNullOpts.mkStr "curly" ''
+        Sets chunk delimiter style.
+      '';
+
+      languages =
+        defaultNullOpts.mkListOf types.str
+          [
+            "r"
+            "python"
+            "julia"
+            "bash"
+            "html"
+          ]
+          ''
+            List of supported languages.
+          '';
+
+      diagnostics = {
+        enabled = defaultNullOpts.mkBool true ''
+          Enables diagnostics.
+        '';
+
+        triggers = defaultNullOpts.mkListOf types.str [ "BufWritePost" ] ''
+          Sets triggers for diagnostics.
+        '';
+      };
+
+      completion = {
+        enabled = defaultNullOpts.mkBool true ''
+          Enables completion.
+        '';
+      };
+    };
+
+    codeRunner = {
+      enabled = defaultNullOpts.mkBool false ''
+        Enables or disables the code runner.
+      '';
+
+      default_method =
+        defaultNullOpts.mkEnum
+          [
+            "molten"
+            "slime"
+          ]
+          null
+          ''
+            Sets the default code runner method. Either "molten", "slime", or `null`.
+          '';
+
+      ft_runners = defaultNullOpts.mkAttrsOf types.str { } ''
+        Specifies filetype to runner mappings.
+      '';
+
+      never_run = defaultNullOpts.mkListOf types.str [ "yaml" ] ''
+        List of filetypes that are never sent to a code runner.
+      '';
+    };
+  };
+
+  settingsExample = {
+    debug = false;
+    lspFeatures = {
+      enabled = false;
+      diagnostics = {
+        enabled = true;
+        triggers = [ "BufWritePost" ];
+      };
+    };
+    codeRunner = {
+      enabled = false;
+      default_method = "vim-slime";
+    };
+  };
+}
--- a/tests/test-sources/plugins/by-name/quarto/default.nix
+++ b/tests/test-sources/plugins/by-name/quarto/default.nix
@@ -1,0 +1,75 @@
+{
+  empty = {
+    plugins.quarto.enable = true;
+  };
+
+  defaults = {
+    plugins.quarto = {
+      enable = true;
+
+      settings = {
+        debug = false;
+        closePreviewOnExit = true;
+        lspFeatures = {
+          enabled = true;
+          chunks = "curly";
+          languages = [
+            "r"
+            "python"
+            "julia"
+            "bash"
+            "html"
+          ];
+          diagnostics = {
+            enabled = true;
+            triggers = [ "BufWritePost" ];
+          };
+          completion = {
+            enabled = true;
+          };
+        };
+        codeRunner = {
+          enabled = false;
+          default_method = null;
+          ft_runners = { };
+          never_run = [ "yaml" ];
+        };
+      };
+    };
+  };
+
+  example = {
+    plugins.quarto = {
+      enable = true;
+
+      settings = {
+        debug = true;
+        closePreviewOnExit = false;
+        lspFeatures = {
+          enabled = true;
+          chunks = "curly";
+          languages = [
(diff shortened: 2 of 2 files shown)


GaetanLepage:
Hello ! Thanks for this initiative, your PR looks in promising.

This plugin uses a `setup()` function for initialization/configuration.
Hence, you should use `neovim-plugin.mkNeovimPlugin` rather than `vim-plugin.mkVimPlugin`. This should not really impact the rest of the code.

plugins/by-name/quarto-nvim/default.nix
```diff
@@ -0,0 +1,108 @@
+{
+  lib,
+  helpers,
+  ...
+}:
+with lib;
+with helpers.vim-plugin;
+mkVimPlugin {
+  name = "quarto-nvim";
```
GaetanLepage, line 9:
```suggestion
  name = "quarto";
```
Also, move the source (and test) files to `quarto` instead of `quarto-nvim`.

plugins/by-name/quarto-nvim/default.nix
```diff
@@ -0,0 +1,108 @@
...
+
+    codeRunner.default_method =
+      helpers.defaultNullOpts.mkEnum
+        [
+          "molten"
+          "slime"
+        ]
+        null
+        ''
+          Sets the default code runner method. Either "molten", "slime", or null.
```
GaetanLepage, line 68:
```suggestion
          Sets the default code runner method. Either "molten", "slime", or `null`.
```

GaetanLepage (changes requested):

bkp5190:
Thank you for the quick review! I will make the required changes


Also, I am new to this repository but an avid nixvim user so excuse my ignorance. Are you able to point me to more docs on running tests locally. I have tried to use cachix, but they are still relatively slow. Is there a framework for testing a single plugin?

GaetanLepage:
> Thank you for the quick review! I will make the required changes
> 
> Also, I am new to this repository but an avid nixvim user so excuse my ignorance. Are you able to point me to more docs on running tests locally. I have tried to use cachix, but they are still relatively slow. Is there a framework for testing a single plugin?

Your participation is very welcome ! Thanks for taking the time to contributing.
In general, you can refer to the [CONTRIBUTING.md](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md) file which contains a good share of important information.

Running the tests can be done locally by running `nix develop --command tests` (or simply `tests` if you are already within the dev shell).
However, running the whole test suite is 1. useless and 2. very memory heavy (~60GB of RAM are necessary).
So, I would recommend you to run the specific quatro test by adding the `-i` flag which allows you to select a specific test to run.

plugins/by-name/quarto-nvim/default.nix
bkp5190: Moved to quarto/ for the source and test

plugins/by-name/quarto-nvim/default.nix
bkp5190: Updated description.

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,108 @@
...
+        [
+          "molten"
+          "slime"
+        ]
+        null
+        ''
+          Sets the default code runner method. Either "molten", "slime", or `null`.
+        '';
+
+    codeRunner.ft_runners = helpers.defaultNullOpts.mkAttrSet { } ''
```
GaetanLepage, line 71:
```suggestion
    codeRunner.ft_runners = helpers.defaultNullOpts.mkAttrsOf types.str { } ''
```

bkp5190: Updated! Thank you!

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,114 @@
+{
+  lib,
+  pkgs,
+  helpers,
+  ...
+}:
+with lib;
+with helpers.neovim-plugin;
```
GaetanLepage, line 8: The usage of `with` is discouraged as it can lead to non-intuitive behaviors.
Prefer a `let`-`in` block.

Have a look at [this example](https://github.com/nix-community/nixvim/blob/aabbd60633947baba11db44df84f402edc241440/plugins/by-name/notebook-navigator/default.nix#L2-L6).

bkp5190: Thank you for the reference! I have updated the convention.

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,111 @@
...
+        python = "molten";
+      };
+      never_run = [ "yaml" ];
+    };
+  };
+
+  extraPlugins = with pkgs.vimPlugins; [
+    nvim-lspconfig
+    otter-nvim
+  ];
```
GaetanLepage, line 110:
```suggestion

  extraPlugins = with pkgs.vimPlugins; [
    nvim-lspconfig
    otter-nvim
  ];
```

Such plugin dependencies should not be added in nixvim.
Either they are strictly required to run the plugin and, in this case, we would need to upstream this constraint in nixpkgs, or they are not. Then, we could add a warning/option for telling the user that this integrations are possible.

From my testing, it appears that only `nvim-lspconfig` is strictly required.

bkp5190: I will investigate nixpkgs then! When running locally, I would see quarto fail to build based on an otter.nvim dependency, but it should not be strictly required. Agreed.

GaetanLepage: https://github.com/NixOS/nixpkgs/pull/354634

bkp5190: Update: Still waiting on this to be in nixos-unstable.

bkp5190: I see this is now available in the nixos-unstable and nixpkgs-unstable branches, what else needs to be done to propagate those changes here? @GaetanLepage

GaetanLepage: I have just merged the `flake.lick` bump. It should work now.

bkp5190:
Checks are passing now! Thanks @GaetanLepage. Please add any review comments/changes if required.

MattSturgeon (approved):
No issues from me

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,103 @@
...
+      Closes preview on exit.
+    '';
+
+    lspFeatures.enabled = defaultNullOpts.mkBool true ''
+      Enables LSP features.
+    '';
+
+    lspFeatures.chunks = defaultNullOpts.mkStr "curly" ''
+      Sets chunk delimiter style.
+    '';
```
GaetanLepage, line 27:
```suggestion
    lspFeatures = {
      enabled = defaultNullOpts.mkBool true ''
        Enables LSP features.
      '';

      chunks = defaultNullOpts.mkStr "curly" ''
        Sets chunk delimiter style.
      '';
      
      ...
```

Same for `codeRunner` and within `codeRunner` too.

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,103 @@
...
+    };
+    codeRunner = {
+      enabled = false;
+      default_method = "vim-slime";
+      ft_runners = {
+        python = "molten";
+      };
+      never_run = [ "yaml" ];
+    };
+  };
```
GaetanLepage, line 102: Make the `settingsExample` configuration at least half shorter.

GaetanLepage (changes requested):

plugins/by-name/quarto/default.nix
bkp5190: Updated with nested structures for the options!

plugins/by-name/quarto/default.nix
bkp5190: Reduced the size of this configuration example!

plugins/by-name/quarto/default.nix
```diff
@@ -0,0 +1,100 @@
+{ lib, pkgs, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "quarto";
+  package = "quarto-nvim";
```
GaetanLepage, line 8:
```suggestion
  name = "quarto";
  originalName = "quarto-nvim";
  package = "quarto-nvim";
```

bkp5190: Added!

GaetanLepage (approved):
LGTM, thank you for this contribution :)

GaetanLepage:
@mergifyio queue

Issue #2318, opened 2024-09-25
[PLUGIN REQUEST] kulala.nvim


SymphonySimper:
| Field    | Description |
|----------|-------------|
| Plugin   | `kulala.nvim`  | 
| Homepage | `https://github.com/mistweaverco/kulala.nvim`     | 
| Nixpkgs  | `false` | 

## Extra Information
Requires Neovim 0.10.0+ and cURL ([complete requirements](https://kulala.mwco.app/docs/getting-started/requirements/)).
>[!NOTE]
>There is already an open [issue](https://github.com/NixOS/nixpkgs/issues/325970) ([PR](https://github.com/NixOS/nixpkgs/pull/332758)) for the package.

PR #2494, opened 2024-11-09
plugins/kulala: init

bkp5190:
# Description
Add kulala.nvim

# Issue
#2318

### DIFF
--- a/plugins/by-name/kulala/default.nix
+++ b/plugins/by-name/kulala/default.nix
@@ -1,0 +1,190 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "kulala";
+  originalName = "kulala.nvim";
+  package = "kulala-nvim";
+
+  maintainers = [ lib.maintainers.BoneyPatel ];
+
+  settingsOptions = {
+    curl_path = defaultNullOpts.mkStr "curl" ''
+      Path to the cURL binary. Set this if cURL is installed in a non-standard location.
+    '';
+
+    display_mode =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "split"
+          "float"
+        ]
+        ''
+          Defines how the request/response is displayed. Options are "split" or "float".
+        '';
+
+    split_direction =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "vertical"
+          "horizontal"
+        ]
+        ''
+          Defines the direction of the split window. Options are "vertical" or "horizontal".
+        '';
+
+    default_view =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "body"
+          "headers"
+          "headers_body"
+        ]
+        ''
+          Sets the default view mode for responses. Options are "body", "headers", or "headers_body".
+        '';
+
+    default_env = defaultNullOpts.mkStr "dev" ''
+      Sets the default environment for requests. Common options are "dev", "test", or "prod".
+    '';
+
+    debug = defaultNullOpts.mkBool false ''
+      Enables or disables debug mode.
+    '';
+
+    contenttypes =
+      defaultNullOpts.mkAttrsOf types.anything
+        {
+          "application/json" = {
+            ft = "json";
+            formatter = [
+              "jq"
+              "."
+            ];
+            pathresolver.__raw = "require('kulala.parser.jsonpath').parse";
+          };
+          "application/xml" = {
+            ft = "xml";
+            formatter = [
+              "xmllint"
+              "--format"
+              "-"
+            ];
+            pathresolver = [
+              "xmllint"
+              "--xpath"
+              "{{path}}"
+              "-"
+            ];
+          };
+          "text/html" = {
+            ft = "html";
+            formatter = [
+              "xmllint"
+              "--format"
+              "--html"
+              "-"
+            ];
+            pathresolver = [ ];
+          };
+        }
+        ''
+          Defines formatters and path resolvers for specific content types.
+        '';
+
+    show_icons =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "on_request"
+          "above_request"
+          "below_request"
+        ]
+        ''
+          Determines where icons are displayed relative to requests. Options are `"on_request"`, `"above_request"`, `"below_request"`, or `{__raw = "nil";}` to disable.
+        '';
+
+    icons =
+      defaultNullOpts.mkAttrsOf types.anything
+        {
+          inlay = {
+            loading = "⏳";
+            done = "✅";
+            error = "❌";
+          };
+          lualine = "🐼";
+        }
+        ''
+          Sets default icons for loading states and lualine.
+        '';
+
+    scratchpad_default_contents =
+      defaultNullOpts.mkListOf types.anything
+        [
+          "@MY_TOKEN_NAME=my_token_value"
+          ""
+          "# @name scratchpad"
+          "POST https://httpbin.org/post HTTP/1.1"
+          "accept: application/json"
+          "content-type: application/json"
+        ]
+        ''
+          Default contents for the scratchpad feature.
+        '';
+
+    winbar = defaultNullOpts.mkBool false ''
+      Enables or disables the winbar for displaying request status.
+    '';
+
+    default_winbar_panes =
+      defaultNullOpts.mkListOf types.str
+        [
+          "body"
+          "headers"
+          "headers_body"
+        ]
+        ''
+          Specifies which panes are displayed by default in the winbar.
+        '';
+
+    vscode_rest_client_environmentvars = defaultNullOpts.mkBool false ''
+      Enables reading environment variables from VSCode’s REST client.
+    '';
+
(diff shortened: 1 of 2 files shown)


GaetanLepage (changes requested):
This plugin can be configured using its `setup` function. You should then add the necessary code to allow for this configuration in nixvim. See [here](https://github.com/nix-community/nixvim/blob/31364af1990067d5529846a2ebf17a42c5ab22ff/plugins/by-name/guess-indent/default.nix) for an example.

bkp5190:
> This plugin can be configured using its `setup` function. You should then add the necessary code to allow for this configuration in nixvim. See [here](https://github.com/nix-community/nixvim/blob/31364af1990067d5529846a2ebf17a42c5ab22ff/plugins/by-name/guess-indent/default.nix) for an example.

Agreed! I missed the configuration options sections as it was a link in the README.md instead of directly included for kulala.nvim!

MattSturgeon:
> This plugin can be configured using its `setup` function. You should then add the necessary code to allow for this configuration in nixvim. 

Just to point out, this should already be possible with the PR as-is, since `mkNeovimPlugin` defaults `hasSettings` to `true`.

More work should only be necessary if you want to declare corresponding nix-options with descriptions and type-checking, which is [entirely optional](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md#declaring-plugin-options) and has its own pros, cons, and maintenance costs.

bkp5190:
I see, I have added some options. I am unsure how to handle options in the following format: https://kulala.mwco.app/docs/getting-started/configuration-options/#certificates

Are there any examples of this configuration or should be leave out these options as it may be more maintenance?

Thank you for the pointers!

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,251 @@
...
+        ''
+          Sets the scope for environment and request variables. Options are "b" (buffer) or "g" (global).
+        '';
+
+    certificates = defaultNullOpts.mkAttrsOf lib.types.anything { } ''
+      Configures SSL/TLS certificates.
+    '';
+  };
+
+  settingsExample = {
```
GaetanLepage, line 179: Ideally, the example should be more concise.
It will be embedded in the description of the `settings` option.

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,251 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
```
GaetanLepage, line 4:
```suggestion
  inherit (lib.nixvim) defaultNullOpts;
  inherit (lib) types;
in
```

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,251 @@
...
+    default_env = defaultNullOpts.mkStr "dev" ''
+      Sets the default environment for requests. Common options are "dev", "test", or "prod".
+    '';
+
+    debug = defaultNullOpts.mkBool false ''
+      Enables or disables debug mode.
+    '';
+
+    contenttypes =
+      defaultNullOpts.mkAttrsOf lib.types.anything
```
GaetanLepage, line 60:
```suggestion
      defaultNullOpts.mkAttrsOf types.anything
```
Same for the other occurrences.

GaetanLepage (dismissed):

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,251 @@
...
+      Path to the cURL binary. Set this if cURL is installed in a non-standard location.
+    '';
+
+    display_mode =
+      defaultNullOpts.mkEnum
+        [
+          "split"
+          "float"
+        ]
+        "split"
```
MattSturgeon, line 23: Minor suggestion: you may prefer using `mkEnumFirstDefault` when the plugin-default is the first enum value:

```suggestion
    display_mode =
      defaultNullOpts.mkEnumFirstDefault
        [
          "split"
          "float"
        ]
```

plugins/by-name/kulala/default.nix
bkp5190: I removed a section for the contenttypes, are there any other recommendations on making it more concise?

plugins/by-name/kulala/default.nix
bkp5190: Updated!

plugins/by-name/kulala/default.nix
bkp5190: Good call! I did not realize this was available. I have updated my enum types as suggested.

plugins/by-name/kulala/default.nix
GaetanLepage: I was more thinking of ~5-6 options only.
You can look for
`language:lua require('kulala').setup` on GitHub to look at people's config.

bkp5190: Got it, I reduced the settingsExample with fewer options similar to other example configs. Thanks @GaetanLepage !

bkp5190:
Any additional thoughts on this @GaetanLepage ?

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,191 @@
...
+    contenttypes =
+      defaultNullOpts.mkAttrsOf types.anything
+        {
+          "application/json" = {
+            ft = "json";
+            formatter = [
+              "jq"
+              "."
+            ];
+            pathresolver = "require('kulala.parser.jsonpath').parse";
```
MattSturgeon, line 66: This looks like it should be rawLua?
```suggestion
            pathresolver.__raw = "require('kulala.parser.jsonpath').parse";
```

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,191 @@
...
+          Defines formatters and path resolvers for specific content types.
+        '';
+
+    show_icons =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "on_request"
+          "above_request"
+          "below_request"
+          null
```
MattSturgeon, line 103: `defaultNullOpts` will already wrap the option's type with `nullOr`, so having `null` in the enum values is kinda redundant.

Further, nixvim currently has a (horrible but necessary) behaviour where nulls, empty `{}` and empty `[]` are ignored when producing the lua settings table.

This means if users actually need to define a `nil` lua value, they need to do `mkRaw "nil"` or `{__raw = "nil";}`... 

I'm guessing this means you need to tweak the description for this option a little theno

MattSturgeon (commented):
Spotted a couple small issues:

plugins/by-name/kulala/default.nix
bkp5190: Added a change for this and modified the description to account for disabling the option with `{__raw = "nil";}`. Please let me know if that matches your suggested change.

plugins/by-name/kulala/default.nix
```diff
@@ -0,0 +1,190 @@
...
+
+    show_icons =
+      defaultNullOpts.mkEnumFirstDefault
+        [
+          "on_request"
+          "above_request"
+          "below_request"
+        ]
+        ''
+          Determines where icons are displayed relative to requests. Options are "on_request", "above_request", "below_request", or "{__raw = "nil";}" to disable.
```
MattSturgeon, line 105:
```suggestion
          Determines where icons are displayed relative to requests. Options are "on_request", "above_request", "below_request", or `{__raw = "nil";}` to disable.
```

Better without quotes, and instead marked as `` `code` ``? Also begs the question: whether the others should be `` `"string_val"` ``?

MattSturgeon (approved):
LGTM :rocket:

plugins/by-name/kulala/default.nix
bkp5190: I like this better! Modified to code and added ``"string_val"``

GaetanLepage:
Since I have merged https://github.com/nix-community/nixvim/pull/2495, could you rebase on `main` and get rid of the `maintainers: add BoneyPatel` commit please ?

MattSturgeon:
> Since I have merged #2495, could you rebase on `main` and get rid of the `maintainers: add BoneyPatel` commit please ?

That should be automatic. It should happen when mergify rebases :upside_down_face: 

But if you want to be the sole author without @nix-infra-bot also being marked as "committer", then you can rebase yourself.

bkp5190:
I dropped the maintainers commit from this PR, should be good to merge now!

MattSturgeon:
@bkp5190 do you want to rebase on our latest main to avoid @nix-infra-bot doing it for you?

Currently you're 33 commits behind our main:
image

bkp5190:
> @bkp5190 do you want to rebase on our latest main to avoid @nix-infra-bot doing it for you?
> 
> Currently you're 33 commits behind our main: image

Rebased!

MattSturgeon:
Just to clarify, a rebase looks like:

```shell
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

I looks like you've just done a merge instead :slightly_smiling_face: 

(To clarify, none of this is too important. We can handle it for you, you just end up with slightly less attribution than if you do it all yourself)

MattSturgeon:
@mergify queue

khaneliman:
@mergify refresh

khaneliman:
@mergify queue

PR #2516, opened 2024-11-13
modules/output: expose other providers

HeitorAugustoLN:
### Changes

* Added `withPerl` option to enable Perl provider.
* Added `withPython3` option to enable Python 3 provider.

### DIFF
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -99,4 +99,54 @@
         }
       ];
     };
+
+  with-providers = {
+    withNodeJs = true;
+    withPerl = true;
+    withPython3 = true;
+    withRuby = true;
+
+    extraConfigLua = ''
+      if vim.fn.executable("nvim-node") ~= 1 then
+        print("Unable to find Node.js provider.")
+      end
+
+      if vim.fn.executable("nvim-perl") ~= 1 then
+        print("Unable to find Perl provider.")
+      end
+
+      if vim.fn.executable("nvim-python3") ~= 1 then
+        print("Unable to find Python3 provider.")
+      end
+
+      if vim.fn.executable("nvim-ruby") ~= 1 then
+        print("Unable to find Ruby provider.")
+      end
+    '';
+  };
+
+  without-providers = {
+    withNodeJs = false;
+    withPerl = false;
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Thanks! This looks great. Ideally we'd have a test to ensure it works correctly (and continues working in the future).

I guess this could be a normal "module" test in [`tests/test-sources/modules`](https://github.com/nix-community/nixvim/tree/main/tests/test-sources/modules), however it doesn't look like we have a test case for the current providers either, so we'd have to come up with something new.

I suppose it is enough to just have a module that uses each provider to add some custom package. If nixvim builds and runs then the test is successful.

Bonus points if the test can also assert that the package is _actually available_ while running neovim. This could be done with [`vim.fn.executable`](https://neovim.io/doc/user/builtin.html#executable()) and some `print()` calls in `extraConfigLua`.

If neovim `print()`s anything, then we consider the test to have failed.

---

I'm thinking something along these lines (unless anyone else has other suggestions):

```nix
# A new test file
{
  with-nodejs =
    { pkgs, ... }:
    {
      # Make node accessible to nvim
      withNodeJs = true;
      
      # When running nvim, print an error if node isn't accessible
      extraConfigLua = ''
        if vim.fn.executable('node') != 1 then
          print('unable to find node...')
        end
      '';
    };
}
```

I don't really mind if we have one test-case per option or if all options are tested in one test-case. I think the latter would be less work for the test suit though.

tests/test-sources/modules/output.nix
```diff
@@ -99,4 +99,54 @@
...
     };
+
+  with-providers = {
+    withNodeJs = true;
+    withPerl = true;
+    withPython3 = true;
+    withRuby = true;
+
+    extraConfigLua = ''
+      if vim.fn.executable("nvim-node") != 1 then
```
MattSturgeon, line 110:
```
error parsing: error occurred while tokenizing: unexpected character !
```

Looks like I got my lua operators wrong:

```suggestion
      if vim.fn.executable("nvim-node") ~= 1 then
```

MattSturgeon (approved):
LGTM, thanks for this 🎉

MattSturgeon:
@mergify queue

PR #2517, opened 2024-11-13
options: add defaultNullOpts.mkRaw

GaetanLepage:
As we plan to move away from `mkLua` and `mkLuaFn` options, declaring options that only accept a `rawLua` value will be more and more common.
This PR adds the `defaultNullOpts.mkRaw` and `defaultNullOpts.mkRaw'` helpers to create such options easily.

Example:
```nix
# Before
custom_func = defaultNullOpts.mkNullable types.rawLua { __raw = "function(x) return x + 1 end"; } ''
  Description.
'';

# After
custom_func = defaultNullOpts.mkRaw "function(x) return x + 1 end" ''
  Description.
'';
```

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -147,6 +147,19 @@ rec {
       mkLuaFn' = args: mkNullOrLuaFn' (processDefaultNullArgs args);
       mkLuaFn = pluginDefault: description: mkLuaFn' { inherit pluginDefault description; };
 
+      mkRaw' =
+        args:
+        mkNullable' (
+          args
+          // {
+            type = types.rawLua;
+          }
+          // lib.optionalAttrs (args ? pluginDefault) {
+            pluginDefault = lib.nixvim.mkRaw args.pluginDefault;
+          }
+        );
+      mkRaw = pluginDefault: description: mkRaw' { inherit pluginDefault description; };
+
       mkNum' = args: mkNullableWithRaw' (args // { type = types.number; });
       mkNum = pluginDefault: description: mkNum' { inherit pluginDefault description; };
       mkInt' = args: mkNullableWithRaw' (args // { type = types.int; });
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Q: even if the type is otherwise identical to `rawLua`, would there still be value in having text like "lua function" in the types description (when appropriate)?

In the same way we currently have a `strLua` and a `strLuaFn`, is there value in having a `rawLuaFn`?

GaetanLepage:
> Q: even if the type is otherwise identical to `rawLua`, would there still be value in having text like "lua function" in the types description (when appropriate)?
> 
> In the same way we currently have a `strLua` and a `strLuaFn`, is there value in having a `rawLuaFn`?

I am not sure. I don't have a strong feeling on this.
On the one hand, the strive for simplicity would push me to say no, let's not multiply helpers for such small differences.
The description should precise that a function is expected anyway.
Now, we could also do it later if we encounter more use cases.

lib/options.nix
```diff
@@ -147,6 +147,17 @@ rec {
       mkLuaFn' = args: mkNullOrLuaFn' (processDefaultNullArgs args);
       mkLuaFn = pluginDefault: description: mkLuaFn' { inherit pluginDefault description; };
 
+      mkRaw' =
+        args:
+        mkNullable' (
+          args
+          // {
+            type = types.rawLua;
+            pluginDefault = lib.nixvim.mkRaw args.pluginDefault;
```
MattSturgeon, line 156: `args.pluginDefault` should be optional. We'll need something like:

```nix
mkRaw' =
  args:
  mkNullable' (
    args
    // { type = types.rawLua; }
    // lib.optionalAttrs (args ? pluginDefault) {
      pluginDefault = lib.nixvim.mkRaw args.pluginDefault;
    }
  );
```

GaetanLepage: Good catch, done.

MattSturgeon (approved):
LGTM, thanks!

GaetanLepage:
@mergifyio queue

PR #2520, opened 2024-11-15
lib/utils: add `literalLua` & `nestedLiteralLua` option docs

MattSturgeon:
Creates a `literalExpression` equivalent to using `lib.nixvim.mkRaw`.

This can be used in place of `{ __raw = "foo"; }` in option docs and provides us with a central place where we would be able to switch to documenting [`lib.generators.mkLuaInline`](https://noogle.dev/f/lib/generators/mkLuaInline) instead.

Design considerations:
- The name: I went with `literalLua` instead of `literalMkRaw` because it seems likely we'll want to transition away from `mkRaw` in the future.
  I am a little concerned that end-users will stumble upon `literalLua` and not realize it is intended for documentation though...
- I've used the full `lib.nixvim.mkRaw` path for clarity. But this has its own issues since non-standalone users will have a different path.
  Perhaps just `mkRaw` would be better? Either way, we're relying on users understanding other parts of our docs to interpret the context.
    - If we switch to [`lib.generators.mkLuaInline`](https://noogle.dev/f/lib/generators/mkLuaInline) then the above problem goes away, but there's more prep work to do before we can do that.
      See #1935
- The function returns a literalExpression. This can be used directly with `pluginDefault`, `defaultText`, and `example`, but _cannot_ be nested within a non-larger value.
    - If we need that, we'll need another function that produces a `__pretty` printer, as supported by [`lib.generators.toPretty`](https://noogle.dev/f/lib/generators/toPretty).
      Such a function could wrap this one, and could be called something like `nestedLiteralLua`.
    - EDIT: **Done** in a follow up commit, included in this PR

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -92,6 +92,7 @@ lib.fix (
       hasContent
       ifNonNull'
       listToUnkeyedAttrs
+      literalLua
       mkIfNonNull
       mkIfNonNull'
       mkRaw
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -97,6 +97,8 @@ lib.fix (
       mkIfNonNull'
       mkRaw
       mkRawKey
+      nestedLiteral
+      nestedLiteralLua
       override
       overrideDerivation
       toRawKeys
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -155,7 +155,7 @@ rec {
             type = types.rawLua;
           }
           // lib.optionalAttrs (args ? pluginDefault) {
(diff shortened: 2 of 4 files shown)


traxys (approved):
> I've used the full lib.nixvim.mkRaw path for clarity. But this has its own issues since non-standalone users will have a different path.

I feel that either users already know about mkRaw, or they need to read the mkRaw documentation as it explains in more details, so I think this is completly acceptable

khaneliman (approved):

MattSturgeon:
@mergify queue

MattSturgeon:
@mergify queue

PR #2521, opened 2024-11-15
templates/simple: configure `nixpkgs.pkgs`

HeitorAugustoLN:
This change removes the existing default, which is planned for deprecation in a future version.

### DIFF
--- a/templates/simple/config/default.nix
+++ b/templates/simple/config/default.nix
@@ -1,4 +1,7 @@
+{ inputs, system, ... }:
 {
   # Import all your configuration modules here
   imports = [ ./bufferline.nix ];
+
+  nixpkgs.pkgs = import inputs.nixpkgs { inherit system; };
 }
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Thanks for this.

I haven't yet fully figured out how the standalone build should work once we transition away from taking pkgs as an argument, but I suspect we may end up leaving `makeNixvimWithModule` as is and introduce some new functions that better match how `evalNixvim` works behind the scenes...

Therefore I'm not sure if we're ready to start implementing this transition with user-facing changes or not.

HeitorAugustoLN:
> Thanks for this.
> 
> I haven't yet fully figured out how the standalone build should work once we transition away from taking pkgs as an argument, but I suspect we may end up leaving `makeNixvimWithModule` as is and introduce some new functions that better match how `evalNixvim` works behind the scenes...
> 
> Therefore I'm not sure if we're ready to start implementing this transition with user-facing changes or not.

Got it, thanks for the context! I’ve been using it this way in my configs and personally find it works better. That said, I’m happy to hold off if it’s not the right time to make this change.

PR #2524, opened 2024-11-15
lib/{vim,neovim}-plugin: `installPackage` -> `packageDecorator`

MattSturgeon:
Replace `mkNeovimPlugin`'s `installPackage` parameter with a new internal module option: `packageDecorator`.

`installPackage` was originally added in #1823 and is only used by treesitter in order to apply its grammar packages to the package.

The option is a function that applies some transformation to `cfg.package`'s value before it is installed.

As discussed in https://github.com/nix-community/nixvim/pull/2462#discussion_r1843050121


GaetanLepage (approved):
LGTM, let's declutter those helpers !

khaneliman (approved):
I like it :)

MattSturgeon:
@mergify queue

PR #2526, opened 2024-11-15
plugins/lsp: remove buf-language-server

anthonycicc:
buf-language-server is officially deprecated and removed from `nixpkgs/master` [here](https://github.com/NixOS/nixpkgs/pull/356199). Removing it here to allow nixvim to continue to build.

I don't know enough about the main-project `buf` to know how to make the language server actually work, if someone else does feel free to close this PR!

### DIFF
--- a/plugins/lsp/lsp-packages.nix
+++ b/plugins/lsp/lsp-packages.nix
@@ -16,6 +16,7 @@
     "bright_script"
     "bsl_ls"
     "buddy_ls"
+    "bufls"
     "bzl"
     "c3_lsp"
     "cadence"
@@ -194,7 +195,6 @@
     bitbake_language_server = "bitbake-language-server";
     blueprint_ls = "blueprint-compiler";
     buck2 = "buck2";
-    bufls = "buf-language-server";
     ccls = "ccls";
     clangd = "clang-tools";
     clojure_lsp = "clojure-lsp";


anthonycicc:
For sure - Does that happen on a cadence? Or just vibes based "when someone gets around to it"?

khaneliman:
It is scheduled every week, but we only updated to latest nixos-unstable. We'll never track master.

anthonycicc:
Ah so it will be a couple more days. I'm fine if this wants to be left open until the `buf-language-server` change makes its way all the way down here, but also feel free to close this. Your choice!

MattSturgeon:
If the package is going to be deprecated soon anyway, I'm not opposed to merging this early.

But removing the line is not enough as we have a test that all nvim-lspconfig servers are handled in some way by the servers list file:

```
 lsp: The following are not declared in plugins/lsp/lsp-packages.nix:
- bufls
```

What this means is that we need to add `"bufls"` to the "unpacked" section of the file.

Once that's done, this LGTM.

(Note: we try to keep the lists sorted)

khaneliman:
> If the package is going to be deprecated soon anyway, I'm not opposed to merging this early.
> 
> But removing the line is not enough as we have a test that all nvim-lspconfig servers are handled in some way by the servers list file:
> 

Sorry, Matt's correct. I saw the failed check and my brain went to new plugin mindset. If we want to deprecate it early we just need to tell our module to ignore it.

anthonycicc:
Appreciate the feedback! I _think_ I made the other requested change (to the unpackaged array) correctly

MattSturgeon (approved):
LGTM. If you're able to rebase onto our latest `main` and then squash your commits to a single commit titled `"plugins/lsp: remove buf-language-server"` that'd be great

MattSturgeon:
@mergify queue

MattSturgeon:
@anthonycicc thanks for dealing with this! :rocket:

MattSturgeon:
> Remove `buf-language-server` from your list of packages and place `buf` in there instead. You’ll then need to configure your editor to use `buf` instead of `buf-language-server`.
> _~[comment](https://discourse.nixos.org/t/package-name-has-changed/56324/2?u=mattsturgeon)_

Looks like anyone using this lsp should switch to [`buf_ls`](https://nix-community.github.io/nixvim/plugins/lsp/servers/buf_ls/index.html)

PR #2530, opened 2024-11-16
lib/types: allow inline-lua in `rawLua` type

MattSturgeon:
Another step towards #1935, allowing `types.rawLua` and `strLua` to accept `_type = "lua-inline"` values as an alternative to our custom `__raw` values.

It may be nice to also implement some "merge" functionality allowing otherwise "equal" definitions to not conflict based on which wrapper type they're using. I.e. currently having an option defined as `mkRaw "foo"` in multiple places merges fine using `mergeEqualOption`, but having a `mkInlineLua "foo"` would not merge.

Maybe we'll deal with that when we get around to coercing `__raw` -> `lua-inline`? As currently we'd need to only coerce if there's multiple definitions of different types.

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -14,7 +14,7 @@ let
       name = "strLua";
       inherit description;
       descriptionClass = "noun";
-      check = v: lib.isString v || isRawType v;
+      check = v: lib.isString v || types.rawLua.check v;
       merge =
         loc: defs:
         lib.pipe defs [
@@ -24,7 +24,8 @@ let
           (lib.options.mergeEqualOption loc)
         ];
     };
-  isRawType = v: v ? __raw && lib.isString v.__raw;
+
+  isRawType = v: lib.isString (v.__raw or null);
 in
 rec {
   # TODO: deprecate in favor of types.rawLua.check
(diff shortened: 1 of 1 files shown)


khaneliman:
I like the idea, just looks like there's something that needs to be tweaked with the changes.

>              error: expected a function but found a list: [ «thunk» «thunk» «thunk» ]

lib/types.nix
```diff
@@ -14,7 +14,10 @@ let
       name = "strLua";
       inherit description;
       descriptionClass = "noun";
-      check = v: lib.isString v || isRawType v;
+      check = lib.any [
```
khaneliman, line 17: still need the v: references in these lib.any changes?

MattSturgeon: Yeah, I forgot what [`builtins.any`](https://noogle.dev/f/builtins/any) actually does.

To do this I'd need `x: any (pred: pred x) [ /* predicates */ ]`, or fallback to plain old `x: pred1 x || pred2 x || pred3 x`

MattSturgeon:
@mergify queue

PR #2536, opened 2024-11-18
docs/search: refactor to use `override`

MattSturgeon:
This should hopefully make branching off 24.11 easier, and may also be a step towards having additional docs versions available and/or docs linking to each other.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -50,35 +50,30 @@ let
     }).optionsJSON;
 
 in
-{
-  inherit options-json;
-  inherit (pkgs) nixos-render-docs;
-
-  gfm-alerts-to-admonitions = pkgs.python3.pkgs.callPackage ./gfm-alerts-to-admonitions { };
-
-  man-docs = pkgs.callPackage ./man { inherit options-json; };
-}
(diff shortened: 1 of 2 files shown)


docs/default.nix
```diff
@@ -50,35 +50,30 @@ let
...
-  in
-  {
+    man-docs = pkgs.callPackage ./man { inherit options-json; };
+  }
+  // lib.optionalAttrs (!pkgs.stdenv.isDarwin) {
     # NuschtOS/search does not seem to work on darwin
-    search = mkSearch "/";
+    search = nuschtosSearch.packages.mkSearch {
+      optionsJSON = options-json + "/share/doc/nixos/options.json";
+      urlPrefix = "https://github.com/nix-community/nixvim/tree/main";
```
traxys, line 67: We should maybe the github URL overrideable too?

MattSturgeon: I'm more thinking of end-users who may wish to build the docs locally themselves. 

Or if we want to locally build them to test out a backport or whatever.

But sure, it's not too important to get perfect straight away

PR #2537, opened 2024-11-18
docs: only set base-href in CI built docs

MattSturgeon:
By shifting the responsibility to set the `baseHref` from the docs to the CI workflow, we remove the need for the docs to know where they will be installed.

- This reverts #1930
- This relies on #2538

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -26,6 +26,9 @@ jobs:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
+      repo: "${{ github.repository }}"
+      repoName: "${{ github.event.repository.name }}"
+      out: docs-build
     runs-on: ubuntu-latest
     timeout-minutes: 40
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_documentation.yml
```diff
@@ -40,14 +43,33 @@ jobs:
...
+          # 2: dest-dir (relative to /nixvim/)
+          build() {
+            flake="github:${repo}${1:+/$1}"
+            baseHref="/${repoName}/${2:+$2/}"
+            destDir="${out}${baseHref}"
+
+            # Build docs for the given flake ref, overriding baseHref in the derivation args
+            nix build --impure --expr '(builtins.getFlake "'"$flake"'").outputs.packages.${builtins.currentSystem}.docs.override {
+              baseHref = "'"$baseHref"'";
+            }'
```
MattSturgeon, line 60: Note: overriding `baseHref` relies on the 24.05 branch _having_ a `baseHref` argument. This is added by #2538

.github/workflows/build_documentation.yml
```diff
@@ -40,14 +43,33 @@ jobs:
...
+            mkdir -p "$destDir"
+            cp -r result/share/doc/* "$destDir"
+          }
+
+          # Install main-branch docs at the top-level
+          build 'main' ''
+
+          # Install 24.05 docs under 'stable'
+          # TODO: consider having `<release>` instead of `stable`
+          build 'nixos-24.05' 'stable'
```
MattSturgeon, line 72: To better scale during transition periods where some users may still need docs for the previous stable version, I think it'd be good to have `/24.05/` and `/24.11/` in the URL, instead of `/stable/`.

One way we could provide some backward compatibility wound be to create an HTML meta-redirect file for each HTML file in the stable docs;

```bash
cd $destDir/24.05
mkdir ../stable
for file in **/*.html; do
  echo "TODO HTML template for $file" > ../stable/$file
done
# TODO maybe have special handling for ../stable/404.html
```

(unfortunately we need meta-redirect files because gh-pages doesn't support proper HTTP redirects)

Obviously this isn't in scope for this PR, but I hope it offers a potential direction we can aim for?

.github/workflows/build_documentation.yml
MattSturgeon: (Yes it's messy that we need to use `getFlake` and `currentSystem` in an expr, but unfortunately `nix build` can't take both a flakeref _and_ an apply function in the same way `nix eval` can) :cry:

.github/workflows/build_documentation.yml
traxys: Yeah I agree that stable URLs would be better, we could avoid breaking them whenever possible

traxys (approved):

MattSturgeon:
@mergify queue

PR #2539, opened 2024-11-18
ci/docs: fix `env` vs `environment` error

MattSturgeon:
Apparently `environment` is something entirely unrelated to `env`, and the workflow was actually invalid!


MattSturgeon:
I just tried testing this by pushing it to a test branch and running the workflow on that branch, but we have a protection rule that only allows the workflow to be run from `main`.

I guess we'll just have to merge, test, and iterate like that?

Unless @GaetanLepage can allow the test branch to deploy gh-pages temporarily, for debugging purposes?

MattSturgeon:
@mergify queue

PR #2540, opened 2024-11-18
docs/ci: fix docs being installed to `nixvim/nixvim/<subdir>`

MattSturgeon:
Everything now seems to be working, other than the docs being incorrectly nested at `/nixvim/nixvim/*`!

- https://nix-community.github.io/nixvim/ (404)
- https://nix-community.github.io/nixvim/nixvim/
- https://nix-community.github.io/nixvim/nixvim/stable/

We don't need to specify `/nixvim/` as part of our install path as that is where gh-pages uploads to by default.

Rather, we only need to specify the full `/nixvim/foo/` in the base-href and the install path should just be `/foo`.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -54,8 +54,9 @@ jobs:
           # 2: dest-dir (relative to /nixvim/)
           build() {
             flake="github:${repo}${1:+/$1}"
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):

MattSturgeon:
@mergify queue

PR #2543, opened 2024-11-19
plugins/lsp/gopls: Only add goPackage to extraPackages if enabled

azuwis:
Currently unconditional, and blow the closure size.

CC @khaneliman

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -20,7 +20,7 @@ let
         };
       };
       extraConfig = cfg: {
-        extraPackages = [ cfg.goPackage ];
+        extraPackages = optionals cfg.enable [ cfg.goPackage ];
       };
     };
     idris2_lsp = {


MattSturgeon:
`extraConfig` is already conditional on `cfg.enable`, so this PR doesn't actually change anything.

Perhaps you have some other issue?

MattSturgeon:
```
$ nix path-info --recursive --closure-size --human-readable result/ | grep go
/nix/store/bavnchxi7v6xs077jxv7fl5rrqc3y87w-go-1.23.3 
```

All this implies is that `go` is included in the closure, it doesn't say _why_ it was included. It could be a dependency of something, or could have been enabled by the wrapper. What we need to pin this down is figure out some steps to _not_ produce a closure that has `go`.

- For instance, does the base `nvim-unwrapped` package have `go` in its closure?
- How about the default (wrapped) `nvim` package?
- How about an nvim wrapped using `wrapNeovimUnstable`, with various options enabled/disabled?
  (this is what nixvim uses under the hood)
- How about nixvim with absolutely nothing enabled?

MattSturgeon:
> After this PR, the `go` package is gone from the closure.

Ah, thanks for checking again! This looks to be caused by a bug in `mkLsp`.

Looking at the `imports`, it unconditionally uses `extraConfig` to produce a config module.

https://github.com/nix-community/nixvim/blob/2017830a2c81a97f6b7679ea5fa0d921cd0f4535/plugins/lsp/language-servers/_mk-lsp.nix#L171

This should be something like:

```nix
{
  config = lib.mkIf cfg.enable (extraConfig cfg);
}
```

This is unlike other instances of `extraConfig` in (e.g.) `mkVimPlugin` and `mkNeovimPlugin`, so I'm happy to call it a "bug". That said, it's possible fixing the bug will break some module that assumes it can define configuration for when `cfg.enable == false`, so we should double check that too.

MattSturgeon:
@azuwis thanks for bringing this to our attention and proposing a fix! @traxys has resolved the issue in #2545.

Let us know if you run into any other issues, rough edges, etc :grinning:

PR #2545, opened 2024-11-19
plugins/lsp: Fix enable conditions

traxys:
There were two bugs in the lsp server modules:
- They unconditionally evaluated extraConfig
- They did not take into account if the global lsp plugin was enabled or not

Fixes #2543


traxys:
I've not checked that, I'll check beforce queueing

MattSturgeon:
I had a very quick scan through, all I could find was one server that has a now-redundant `mkIf`: https://github.com/nix-community/nixvim/blob/main/plugins%2Flsp%2Flanguage-servers%2Fdefault.nix#L29

traxys:
I saw another place where there was a redundant mkIf, but I saw no other problem

traxys:
@mergify queue

PR #2548, opened 2024-11-19
plugins/telescope: refactor `mkExtension`

MattSturgeon:
Allow importing `mkExtension` without needing to supply any dependencies.

All dependencies can be accessed via module args anyway.

This is similar to how `mkVimPlugin` and `mkNeovimPlugin` now work.

This is intended to enable future work adding an `applyExtraConfig` function:

```nix
# Allows `extraConfig` to be either an attrset, a 1-param function, or a 2-param function:
applyExtraConfig =
  {
    extraConfig,
    cfg,
    opts,
    enabled ? cfg.enable,
  }:
  lib.pipe extraConfig [
    # First try and apply `cfg` to the fn
    (fn: if isFunction fn then fn cfg else fn)
    # Then try and apply `opts` to it
    (fn: if isFunction fn then fn opts else fn)
    # Finally wrap the result with `mkIf`
    (lib.mkIf enabled)
  ];
```


traxys (approved):

MattSturgeon:
@mergify queue

PR #2549, opened 2024-11-19
lib/modules: add `applyExtraConfig`

MattSturgeon:
Adds a function to `lib.nixvim.modules` for applying "`extraConfig`" style configurations.

This allows us to be less rigid with how `extraConfig` is defined, without having to re-implement complex logic everywhere we use an `extraConfig`-style option.

Specifically, `extraConfig` is now allowed to be defined as:
- A plain attrset, `AttrSet`
- A 1-param function, `cfg -> AttrSet`
- A 2-param function, `cfg -> opts -> AttrSet`

Where `cfg` and `opts` are the sub-section of `config` or `options` relevant to this specific context.

This flexibility allows most plugins to continue using the `cfg: { }` pattern, but those that do not need `cfg` need not specify it and those that also need `opts` no longer need to define it themselves via a `let-in` block.


lib/modules.nix
```diff
@@ -36,6 +36,28 @@ in
...
+
+    The result will be wrapped using a `mkIf` definition.
+  */
+  applyExtraConfig =
+    {
+      extraConfig,
+      cfg,
+      opts,
+      enabled ? cfg.enable or throw "`enabled` argument not provided and no `cfg.enable` option found.",
+    }:
```
MattSturgeon, line 52: Thoughts on naming and how generic this should be?

E.g. we could refer to a `varArgFn` (or simply `fn` or `function`) instead of an `extraConfig`

Perhaps this should be broken down into two pieces:
- a generic function living in `utils` that can apply a variable number of arguments to a fn
- a extraConfig-specific helper that applies varargs and then wraps the result in `mkIf`

A generic apply helper might look like:

<p>


```nix
# While fn is a function, apply aruments from args to it
# Once fn is no-longer a function, its value is retuned
applyArgsIfFn =
  args: fn:
  lib.pipe fn (map (arg: f: if isFunction f then f arg else f) args);
```

```nix
# Using recursion _may_ be more efficient,
# although `lib.drop` looks more expensive than you'd expect...
applyArgsIfFn =
  args: fn:
  if isFunction fn then
    applyArgsIfFn (drop 1 args) (fn (head args))
  else
    fn;
```

```nix
# This recursive impl avoids needing `lib.drop`,
# by doing the recursion on an index instead of the list itself
applyArgsIfFn =
  args: fn:
  let
    len = length args;
    go =
      index: acc:
      if index < len && isFunction acc then
        # If acc is still a fn, pass it args[i], then go a step deeper...
        go (index + 1) (acc (elemAt args index))
      else
        acc;
  in
    go 0 fn;
```

</p>
</details>

khaneliman: I can't think of anything more generic for the current implementation since it's specific to extra config and its arguments. In the future, if we find use for the function and want to break it up it could make sense to go that generic route.

khaneliman (approved):
LGTM I think it cleans some stuff up nicely.

MattSturgeon:
@mergify queue

GaetanLepage:
Good job !

PR #2552, opened 2024-11-20
modules/opts: fix code generation condition to prevent false positives

GaetanLepage:
Currently, we generate the lua code as soon as the value of the option is not an empty attrs.
However, a case where the attrs only contains `null` values (`nord_colorscheme = null; nord_foo = null;}` would pass the `if` condition, but the generated lua table would still be empty because `toLua` filters out those values.

Hence, we should directly check if the attrs would be translated to an empty lua table or not to decide whether we should generate the lua snippet or not.

### DIFF
--- a/modules/opts.nix
+++ b/modules/opts.nix
@@ -67,12 +67,12 @@ in
             }:
             let
               varName = "nixvim_${luaVariableName}";
-              optionDefinitions = config.${optionName};
+              optionDefinitions = helpers.toLuaObject config.${optionName};
             in
-            lib.optionalString (optionDefinitions != { }) ''
+            lib.optionalString (optionDefinitions != "{ }") ''
               -- Set up ${prettyName} {{{
               do
-                local ${varName} = ${helpers.toLuaObject optionDefinitions}
+                local ${varName} = ${optionDefinitions}
 
                 for k,v in pairs(${varName}) do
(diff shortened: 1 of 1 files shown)


modules/opts.nix
```diff
@@ -67,12 +67,12 @@ in
             }:
             let
               varName = "nixvim_${luaVariableName}";
-              optionDefinitions = config.${optionName};
+              optionDefinitions = helpers.toLuaObject config.${optionName};
             in
-            lib.optionalString (optionDefinitions != { }) ''
+            lib.optionalString (optionDefinitions != "{ }") ''
```
khaneliman, line 72: I see in matt's pr the string is `{}`, what do we generate with toLuaObject?

GaetanLepage: We generate `"{ }"` which gets formatted by `luafmt` to `"{}"`.
Originally, I tried with `"{}"` and it was still showing.

GaetanLepage:
```
nix-repl> lib.x86_64-linux.helpers.toLuaObject {}
"{ }"
```

khaneliman (approved):

MattSturgeon (approved):
LGTM, thanks!

Closes #2551

GaetanLepage:
@mergifyio queue

PR #2553, opened 2024-11-20
docs/ci: use `nix build --argstr` to safely supply input

MattSturgeon:
Rather than interpolating input (`flakeref` & `baseHref`) into the build expression, take advantage of `--argstr` which avoids potential string-escape pitfalls.

See "Common evaluation options" in `nix build --help` for more detail.

Additionally, refactor the expression slightly to make it more readable; using (de)structured args and a let-in block to improve readability.

Having a more maintainable expression will help in adding additional overrides, such as `branch` and a list of "other" docs to link to.


MattSturgeon:
@mergify queue

PR #2555, opened 2024-11-20
docs: no longer using `master` but `main`

elsbrock:
Tiny update. Noticed the `master` branch does not exist.


MattSturgeon (approved):
I also spotted this the other day, but forgot about it agin soon after... Thanks for the fix!

MattSturgeon:
@mergify queue

PR #2561, opened 2024-11-22
plugins/vim-colemak: init

kalbasit:
Integrate the [vim-colemak](https://search.nixos.org/packages?channel=24.05&show=vimPlugins.vim-colemak&from=0&size=50&sort=relevance&type=packages&query=vim-colemak) plugin.

### DIFF
--- a/plugins/by-name/vim-colemak/default.nix
+++ b/plugins/by-name/vim-colemak/default.nix
@@ -1,0 +1,5 @@
+{ lib, ... }:
+lib.nixvim.vim-plugin.mkVimPlugin {
+  name = "vim-colemak";
+  maintainers = [ lib.maintainers.kalbasit ];
+}
--- a/tests/test-sources/plugins/by-name/vim-colemak/default.nix
+++ b/tests/test-sources/plugins/by-name/vim-colemak/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.vim-colemak.enable = true;
+  };
+}


GaetanLepage:
What is the difference with https://github.com/linduxed/colemak.nvim ?
If both are doing the same, maybe we should choose which one to add (and have it named `plugins.colemak`).

kalbasit:
@GaetanLepage I have not used that plugin. This plugin ([mine](https://github.com/kalbasit/vim-colemak)) is forked from [this one](https://github.com/jooize/vim-colemak) with the motion (hjkl on qwerty) moved to `neio` the keys on the home row for the right hand on the colemak keyboard layout. I submitted a [PR years ago](https://github.com/jooize/vim-colemak/pull/10) but it was never merged.

I have not evaluated the colemak.nvim but I took a quick look at it, and from the README:
> **unei** (**ijkl** on QWERTY) replaces **hjkl**

Makes it immediately a no-go for me as it moves the home-row motion into a reverse T style arrow key movement (the u is above e, n to the left of e and i to the right) and that's moving away too far from how Vim typically works with the flat-style motion.

MattSturgeon:
> @GaetanLepage I have not used that plugin. This plugin ([mine](https://github.com/kalbasit/vim-colemak)) is forked from [this one](https://github.com/jooize/vim-colemak) with the `hjkl` motions moved to `neio` the keys on the home row [...]
> 
> I have not evaluated the colemak.nvim but I took a quick look at it, and from the README:
> > **unei** (**ijkl** on QWERTY) replaces **hjkl**
> 
> Makes it immediately a no-go for me as it moves the home-row motion into a reverse T style arrow key movement [...]
> 
> I also do not see it packaged in [nixpkgs](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=colemak).

So to summarise, we have three plugins with the same goal (making vim more colemak friendly) but different strategies. Of those, this is the only one packaged in nixpkgs?

I can see a case for simplicity; naming the plugin simply "colemak"... But it also seems conceivable that (down the line) some users may wish to package and use the _other_ competing plugins instead. On that basis, I think the current "vim-colemak" name is the best choice.

---

As an aside; `neio` are all fairly fundamental actions in vim, many with a mnemonic relation to what they do; `n` next, `e` end of word, `i` insert, `o` open new line.

Therefore, even as a colemak user myself, I've avoided these kinda rebindings. I'm curious how you find this affects your other keybinds?

MattSturgeon (approved):
No issues from me 👍

kalbasit:
@MattSturgeon Thanks for your review and approval.
 
> As an aside; `neio` are all fairly fundamental actions in vim, many with a mnemonic relation to what they do; `n` next, `e` end of word, `i` insert, `o` open new line.
> 
> Therefore, even as a colemak user myself, I've avoided these kinda rebindings. I've also held back to maintain consistency with _other_ programs that have vi-style keybindings, e.g. pagers like `less` that are often used by things like `git`.
> 
> I'm curious how you find this affects your other keybinds?

It hasn't been easy, but doable. Most programs allows some level of custom keybing, including less using the less-known `lesskey` program that can adjust its keybinding; See this module I created for it: https://github.com/SoxinOS/soxin/blob/main/modules/programs/less/default.nix

The `vim-colemak` moves things around to allow me to use `neio` for motion, and I do the same for all other applications I use. This also forces me to change some settings for `(n)vim` plugins in order to change their default keybinding, for instance NeoTree, (or multicursors that I'm trying to integrate currently); I'm consolidating my work on the plugins [here](https://github.com/kalbasit/nixvim/blob/main/config/keymaps/colemak.nix).

GaetanLepage:
Ok, this was indeed the motivation behind my question: If we could want at some point to have several of those plugins in nixvim, it is better to anticipate and have it named `vim-colemak`.

GaetanLepage (approved):

MattSturgeon:
@mergify queue

PR #2575, opened 2024-11-27
plugins/image.nvim: ueberzugPackage default to ueberzugpp

samos667:
ueberzug is not maintained anymore and is not working.

[ueberzugpp](https://github.com/jstkdng/ueberzugpp) is a active fork, that I think , should be used as default

### DIFF
--- a/plugins/by-name/image/default.nix
+++ b/plugins/by-name/image/default.nix
@@ -26,7 +26,7 @@ in
       nullable = true;
     };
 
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
LGTM, thanks

khaneliman (approved):
Thanks again

khaneliman:
@mergify queue

PR #2577, opened 2024-11-27
lazy-simple

khaneliman:


MattSturgeon:
> Maybe a good first step would be supporting only init and none, doing a second step for the lazy loading provider

I'm wondering about maybe allowing `either bool submodule`?

For now it could be just `bool`, where if true we add the config to `init.lua` and if false we do not.

In the future we can add `either submodule` or `either (coercedTo str _ submodule)` to allow users to define a lazy spec specifying which lazy loader to use and what events will trigger loading the plugin.

I also wonder whether which lazy loading implementation to use should be a global option or a per-plugin one? Surely it wouldn't make sense for multiple lazy loaders to be used simultaneously?

Perhaps it would be enough to just figure out which lazy loading plugin is enabled? And assert that multiple are not enabled at the same time?

traxys:
The issue I could forsee is that not all lazy providers could present the same API, meaning that we would not easily provide a RFC-42 style submodule, though this may not be a large issue if all lazy loading providers use a common interface

Eveeifyeve:
When enabling lz-n, it should automatic in each plugin enable lazyloading then you have a choice to disable lazyloading if need to (for issues,) but I feel like it would save like between 1-50 lines of code plus having a benefit of cleaner code in a nixvim config.

lib/neovim-plugin.nix
```diff
@@ -113,6 +113,30 @@
                 '';
                 internal = true;
               };
+
+              lazyLoad = lib.mkOption {
+                default = {
+                  enable = false;
+                };
```
MattSturgeon, line 120: If `enable` has its own default, this top-level default can simply be `{}`

lib/neovim-plugin.nix
```diff
@@ -113,6 +113,30 @@
...
+                description = ''
+                  Lazy load configuration settings.
+                '';
+                type = lib.types.submodule {
+                  options = {
+                    enable = lib.mkOption {
```
MattSturgeon, line 126: I'm wondering, if it is likely we will have a different "enable" in the future, should this be named something else?

Having it named something like `dontCallSetup` may be more explicit and would also allow us to more easily deprecate this option if we need to in the future?

lib/neovim-plugin.nix
```diff
@@ -113,6 +113,30 @@
...
+                      default = false;
+                      type = lib.types.bool;
+                      description = ''
+                        Whether to lazy load the plugin.
+
+                        If you enable this, the plugin's lua configuration will need to be manually loaded by other means.
+
+                        A usage would be passing the plugin's luaConfig to the `plugins.lz-n.plugins` configuration.
```
MattSturgeon, line 134: Q: I like this initial implementation, but do you think it is too much of a stretch to go straight to implementing this ourselves in this PR?

That way we never have the interim stage of end-users having to enable a "dont setup" option and passing their plugin's settings to another plugin which lazy-loads them.

khaneliman: I have another branch i'm playing around with right now for setting up lz-n for a user following the work done in https://github.com/nix-community/nixvim/pull/1875

khaneliman: Yeah... I thought about that when I started going through and implementing my own lazy loading. It felt tedious enough, let alone if I had to redo it again.

PR #2580, opened 2024-11-28
plugins/vim-suda: init

MarcelCoding:
Read/Write already opened files using sudo.

https://github.com/lambdalisue/vim-suda/

### DIFF
--- a/plugins/by-name/vim-suda/default.nix
+++ b/plugins/by-name/vim-suda/default.nix
@@ -1,0 +1,41 @@
+{ lib, helpers, ... }:
+
+lib.nixvim.vim-plugin.mkVimPlugin {
+  name = "vim-suda";
+  globalPrefix = "suda#";
+  maintainers = [ lib.maintainers.marcel ];
+
+  settingsOptions = {
+    executable = helpers.defaultNullOpts.mkStr "sudo" ''
+      Path to the sudo binary.
+    '';
+
+    noninteractive = helpers.defaultNullOpts.mkFlagInt 0 ''
+      If set, suda will not prompt you for a password before saving a file.
+      It is supposed to support a setup with passwordless sudo or doas.
+      Use with care.
+    '';
+
+    prompt = helpers.defaultNullOpts.mkStr "Password: " ''
+      A prompt string used to ask password.
+    '';
+
+    # TODO: til https://github.com/lambdalisue/vim-suda/pull/84
+    #       makes it into nixos-unstable
+    # NOTE: also update in `settingsExample` and test cases
+    #smart_edit = helpers.defaultNullOpts.mkFlagInt 0 ''
+    #  If set, an `|autocmd|` is created that performs a heuristic check on
+    #  every buffer and decides whether to replace it with a suda buffer.
+    #  The check is done only once for every buffer and it is designed to be
+    #  optimized as possible so you shouldn't feel any slowdown when opening
+    #  buffers.
+    #'';
+  };
+
+  settingsExample = {
+    path = "doas";
+    noninteractive = 1;
+    prompt = "Pass: ";
+    #smart_edit = 1;
+  };
+}
--- a/tests/test-sources/plugins/by-name/vim-suda/default.nix
+++ b/tests/test-sources/plugins/by-name/vim-suda/default.nix
@@ -1,0 +1,29 @@
+{
+  empty = {
+    plugins.vim-suda.enable = true;
+  };
+
+  defaults = {
+    plugins.vim-suda = {
+      enable = true;
+      settings = {
+        path = "sudo";
+        noninteractive = 0;
+        prompt = "Password: ";
+        #smart_edit = 0;
+      };
+    };
+  };
+
+  example = {
+    plugins.vim-suda = {
+      enable = true;
+      settings = {
+        path = "doas";
+        noninteractive = 1;
+        prompt = "Pass: ";
+        #smart_edit = 1;
+      };
+    };
+  };
+}


traxys:
I see that the are some settings, I think we should specify `globalPrefix` as `suda#` to allow users to configure the plugin

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,6 @@
+{ lib, ... }:
+lib.nixvim.vim-plugin.mkVimPlugin {
+  name = "vim-suda";
+  globalPrefix = "suda#";
```
MattSturgeon, line 4: Looking at https://github.com/lambdalisue/vim-suda/blob/b97fab52f9cdeabe2bbb5eb98d82356899f30829/doc/suda.txt#L171-L194, most globals are using the `suda#` prefix, but there is also `suda_smart_edit` which uses `suda_`...

Kinda frustrating, but I guess this is fine?

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,6 @@
...
+  maintainers = [ lib.maintainers.marcel ];
+}
```
MattSturgeon, line 6: I'm happy to keep `settingsOptions` empty/undefined, but I wonder if it'd be nice to have a small `settingsExample` using some real-world example? Or maybe one of the examples from [upstream's docs](https://github.com/lambdalisue/vim-suda/blob/master/doc/suda.txt)?

MarcelCoding: Considering that the prefix is not constant would it be viable to just define all options explicitly?

MattSturgeon: Are you replying to https://github.com/nix-community/nixvim/pull/2580#discussion_r1862521732 ?

I think it'd be valid in this case to have the prefix set to `""` and do exactly that, just a little frustrating.

We'd also need to double check the auto-generated `settings` option's documentation, as it _might_ make reference to a prefix that we're not using...

plugins/by-name/vim-suda/default.nix
MarcelCoding: I've added a pull request to add fix the mismatched option prefix. https://github.com/lambdalisue/vim-suda/pull/84

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,39 @@
...
+  name = "vim-suda";
+  globalPrefix = "suda#";
+  maintainers = [ lib.maintainers.marcel ];
+
+  settingsOptions = {
+    executable = helpers.defaultNullOpts.mkStr "sudo" ''
+      Path to the sudo binary.
+    '';
+
+    noninteractive = helpers.defaultNullOpts.mkBool false ''
```
MattSturgeon, line 13: I don't think these are _proper_ booleans, i.e. `set foo` and `set nofoo`. Based on the default `0` listed in the docs, I think these are bit flag style _fake_ booleans.

If so:

```suggestion
    noninteractive = helpers.defaultNullOpts.mkFlagInt 0 ''
```

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,39 @@
...
+      It is supposed to support a setup with passwordless sudo or doas.
+      Use with care.
+    '';
+
+    prompt = helpers.defaultNullOpts.mkStr "Password: " ''
+      A prompt string used to ask password.
+    '';
+
+    smart_edit = helpers.defaultNullOpts.mkBool false ''
+      If set, an |autocmd| is created that performs a heuristic check on
```
MattSturgeon, line 24: We can use markdown in descriptions, so probably best to wrap `` `|doclinks` `` in backticks?

```suggestion
      If set, an `|autocmd|` is created that performs a heuristic check on
```

Also: if you want a paragraph break, you need to have an empty/blank line between paragraphs; single line-breaks are ignored by markdown :slightly_smiling_face:

tests/test-sources/plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.vim-suda.enable = true;
+  };
```
MattSturgeon, line 4: If we have settings and an example now, there should be corresponding test cases here.

In total we should have `empty`, `defaults` and `example`

- `defaults` should set all values explicitly to upstream's defaults
- - `example` should ideally matching the settingsExample in the plugin module

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    '';
+  };
+
+  settingsExample = {
+    path = "doas";
+    noninteractive = true;
+    prompt = "Pass: ";
+    smart_edit = true;
+  };
+
```
MattSturgeon, line 38: Minor nit:

```suggestion
```

plugins/by-name/vim-suda/default.nix
```diff
@@ -0,0 +1,39 @@
...
+      If set, suda will not prompt you for a password before saving a file.
+      It is supposed to support a setup with passwordless sudo or doas.
+      Use with care.
+    '';
+
+    prompt = helpers.defaultNullOpts.mkStr "Password: " ''
+      A prompt string used to ask password.
+    '';
+
+    smart_edit = helpers.defaultNullOpts.mkBool false ''
```
MattSturgeon, line 23: I think if this setting should just be a `# TODO` until https://github.com/lambdalisue/vim-suda/pull/84 makes it to nixos-unstable?

plugins/by-name/vim-suda/default.nix
MarcelCoding: As there are no paragraph breaks in the upstream docs, I'm just going not to insert some.

MattSturgeon (approved):
LGTM, thanks :rocket:

MattSturgeon:
@mergify queue

PR #2588, opened 2024-11-30
ci/docs: remove on 24.05 branch

MattSturgeon:
This doesn't work because we can only deploy docs from the main branch.


GaetanLepage (approved):

MattSturgeon:
@mergify queue

Issue #2587, opened 2024-11-30
RAW Lua support for plugins.dashboard.settings.config.header


freerig:
Currently, the type of [plugins.dashboard.settings.config.header](https://nix-community.github.io/nixvim/plugins/dashboard/settings/config/index.html#pluginsdashboardsettingsconfigheader) is `null or (list of (string or raw lua code))`. I want it to be `null or (list of (string or raw lua code)) OR RAW LUA CODE`.
The idea is to, instaid of having multiple lua function for each line, having one lua function that returns a list of string.
I think this is not super hard to do, I'll try to do it myself if I manage to understand the repo, but feel free to make a pull request, or to tell me what to do.

PR #2592, opened 2024-11-30
plugins/dashboard: better header rawlua support

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2587

Looks like `mkListOf` is `mkNullable` instead of `mkNullableWithRaw`. Do we want to change that instead?


MattSturgeon (approved):
Horrible that we need this, maybe it's worth having a "TODO remove once underlying issue is fixed"?

Either way, this is fine as is and solves a specific issue, so I'm happy.

Long term, I think we should make it easier to have raw-lua support everywhere by default.

khaneliman:
@mergify queue

PR #2595, opened 2024-12-01
plugin/lsp: automatic keymap description

ein-shved:
Add keymaps default descriptions for lsp buf and diagnostics

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -198,6 +198,14 @@ in
               let
                 actionStr = action.action or action;
                 actionProps = optionalAttrs (isAttrs action) (removeAttrs action [ "action" ]);
+                desc =
+                  let
+                    split = splitString "." prefix;
+                    splitlen = length split;
+                    prefixFinal = (elemAt split (splitlen - 2)) + " ";
+                    optPrefix = optionalString (splitlen > 2) prefixFinal;
+                  in
+                  "Lsp " + optPrefix + actionStr;
               in
               {
                 mode = "n";
@@ -206,6 +214,7 @@ in
 
                 options = {
                   inherit (cfg.keymaps) silent;
+                  inherit desc;
                 } // actionProps;
               }
             );
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -192,20 +192,12 @@ in
       keymapsOnEvents.LspAttach =
         let
           mkMaps =
-            prefix:
+            prefix: descPrefix:
             mapAttrsToList (
               key: action:
               let
                 actionStr = action.action or action;
                 actionProps = optionalAttrs (isAttrs action) (removeAttrs action [ "action" ]);
-                desc =
-                  let
-                    split = splitString "." prefix;
-                    splitlen = length split;
-                    prefixFinal = (elemAt split (splitlen - 2)) + " ";
-                    optPrefix = optionalString (splitlen > 2) prefixFinal;
-                  in
-                  "Lsp " + optPrefix + actionStr;
               in
               {
                 mode = "n";
@@ -214,13 +206,13 @@ in
 
                 options = {
                   inherit (cfg.keymaps) silent;
-                  inherit desc;
+                  desc = "${descPrefix} ${actionStr}";
                 } // actionProps;
               }
             );
         in
-        mkMaps "vim.diagnostic." cfg.keymaps.diagnostic
-        ++ mkMaps "vim.lsp.buf." cfg.keymaps.lspBuf
+        mkMaps "vim.diagnostic." "Lsp diagnostic" cfg.keymaps.diagnostic
+        ++ mkMaps "vim.lsp.buf." "Lsp buf" cfg.keymaps.lspBuf
         ++ cfg.keymaps.extra;
 
       # Enable inlay-hints


ein-shved:
@GaetanLepage, I think, it looks good already in generic way:
image
image

But if you prefer straight way, I can change it.

GaetanLepage:
> @GaetanLepage, I think, it looks good already in generic way: image image
> 
> But if you prefer straight way, I can change it.

The results is indeed great. I think that having such a generic implementation for computing two strings is a bit overkilled.

ein-shved:
> I think that having such a generic implementation for computing two strings is a bit overkilled

Yep, you are right! I've updated implementation, but did not squash commits to left generic implementation in history. Just in case.

plugins/lsp/default.nix
```diff
@@ -192,12 +192,13 @@ in
...
           mkMaps =
-            prefix:
+            prefix: descPrefix:
             mapAttrsToList (
               key: action:
               let
                 actionStr = if isString action then action else action.action;
-                actionProps = if isString action then { } else filterAttrs (n: v: n != "action") action;
+                actionProps = optionalAttrs (!(isString action)) (filterAttrs (n: v: n != "action") action);
+                desc = descPrefix + " " + actionStr;
```
GaetanLepage, line 201:
```suggestion
                desc = "${descPrefix} ${actionStr}";
```
I guess this looks simpler...

plugins/lsp/default.nix
```diff
@@ -206,12 +207,13 @@ in
 
                 options = {
                   inherit (cfg.keymaps) silent;
+                  inherit desc;
```
GaetanLepage, line 210:
```suggestion
                  desc = lib.mkDefault desc;
```
Maybe we should do that so that the user can override the description without having to use `lib.mkForce`.

plugins/lsp/default.nix
```diff
@@ -192,12 +192,13 @@ in
...
           mkMaps =
-            prefix:
+            prefix: descPrefix:
             mapAttrsToList (
               key: action:
               let
                 actionStr = if isString action then action else action.action;
-                actionProps = if isString action then { } else filterAttrs (n: v: n != "action") action;
+                actionProps = optionalAttrs (!(isString action)) (filterAttrs (n: v: n != "action") action);
+                desc = "${descPrefix} ${actionStr}";
```
MattSturgeon, line 201: Just to avoid confusion, perhaps name this something like `autoDesc` or `generatedDesc`?

MattSturgeon: Or since it is only used on line 210, maybe just inline it there instead of defining it in a let-in block

plugins/lsp/default.nix
```diff
@@ -192,12 +192,13 @@ in
...
         let
           mkMaps =
-            prefix:
+            prefix: descPrefix:
             mapAttrsToList (
               key: action:
               let
                 actionStr = if isString action then action else action.action;
-                actionProps = if isString action then { } else filterAttrs (n: v: n != "action") action;
+                actionProps = optionalAttrs (!(isString action)) (filterAttrs (n: v: n != "action") action);
```
MattSturgeon, line 200: This seems unrelated?

Especially if https://github.com/nix-community/nixvim/pull/2595/files#r1865065311 is done, we wouldn't need _any_ changes to this let-in block for this PR.

I don't mind having general cleanup if it's in a separate commit or another PR. If that was done, perhaps `actionStr` could be simplified to `action.action or action`? But both changes really feel unrelated to this PR :slightly_smiling_face:

plugins/lsp/default.nix
```diff
@@ -206,12 +207,13 @@ in
...
                   inherit (cfg.keymaps) silent;
+                  inherit desc;
                 } // actionProps;
               }
             );
         in
-        (mkMaps "vim.diagnostic." cfg.keymaps.diagnostic)
-        ++ (mkMaps "vim.lsp.buf." cfg.keymaps.lspBuf)
+        (mkMaps "vim.diagnostic." "Lsp diagnostic" cfg.keymaps.diagnostic)
+        ++ (mkMaps "vim.lsp.buf." "Lsp buf" cfg.keymaps.lspBuf)
```
MattSturgeon, line 216: If you're touching these lines, could you drop the unnecessary parens?
```suggestion
        mkMaps "vim.diagnostic." "Lsp diagnostic" cfg.keymaps.diagnostic
        ++ mkMaps "vim.lsp.buf." "Lsp buf" cfg.keymaps.lspBuf
```

plugins/lsp/default.nix
GaetanLepage: #2598

plugins/lsp/default.nix
MattSturgeon: Resolved in https://github.com/nix-community/nixvim/pull/2598

plugins/lsp/default.nix
MattSturgeon: Actually, not sure this is needed? If the user has defined a `desc` it'll be part of the `actionProps` that overrides anything set here, right?

MattSturgeon:
I've just pushed a commit to resolve conflicts with #2598 and also https://github.com/nix-community/nixvim/pull/2595#discussion_r1865065311. I left https://github.com/nix-community/nixvim/pull/2595#discussion_r1864929279 unresolved for now, since I think it's actually handled by the `// actionProps` attr-update.

MattSturgeon (approved):
Thanks for this!

plugins/lsp/default.nix
ein-shved: > If the user has defined a `desc` it'll be part of the `actionProps` that overrides anything set here, right?

Exactly!

Issue #2596, opened 2024-12-01
[PLUGIN REQUEST] wezterm.nvim


samos667:
| Field    | Description |
|----------|-------------|
| Plugin   | `wezterm.nvim`  | 
| Homepage | [github.com/willothy/wezterm.nvim](https://github.com/willothy/wezterm.nvim)     | 
| Nixpkgs  | `I use github:wez/wezterm?dir=nix input because otherwise wezterm is broken on wayland.` | 

## Extra Information

Allow to interact with wezterm within nvim, provide [various features](https://github.com/willothy/wezterm.nvim/blob/main/doc/wezterm.nvim.txt) nice to have for wezterm users.
[I have tested](https://github.com/samos667/nivis/commit/a3da8881a6132d58b7875d8c34cc177375695459) with pkgs.vimUtils and all seems to work nicely.

By the way I think wezterm should be a possible choice for molten.nvim image provider https://github.com/nix-community/nixvim/blob/80e49e7fd3fa720b93d18e6d859d9b9e7aad4a62/plugins/by-name/molten/default.nix#L78

GaetanLepage:
Would you be interested in giving it a shot ?
We would obviously help you with make it work !

samos667:
Of course I'd love to, I will try it and open a PR !

GaetanLepage:
> Of course I'd love to, I will try it and open a PR !

Great ! Also, feel free to join the [development matrix room](https://matrix.to/#/#nixvim-dev:matrix.org) if you wish :)

PR #2597, opened 2024-12-01
plugins/wezterm: init

samos667:
This pull request introduces a new Neovim plugin for `wezterm` and integrates it into the existing plugin system. The changes include adding the `wezterm` plugin to the `molten` plugin list, defining the `wezterm` plugin in its own file, and adding test cases for the new plugin.

### New Plugin Addition:

* [`plugins/by-name/molten/default.nix`](diffhunk://#diff-230d660b843085745a52be69db8e4aa925f8c0297745d7b7bccbefc9c9ad2e29R83-R87): Added the `wezterm` plugin to the list of plugins and enabled it by default.
* [`plugins/by-name/wezterm/default.nix`](diffhunk://#diff-f8bd7524f04fe0ddf164336d51c3b20e34fcdc9c647a9948b32f068a4ffc3dd8R1-R21): Defined the `wezterm` plugin with its settings and maintainers.

### Testing:

* [`tests/test-sources/plugins/by-name/wezterm/default.nix`](diffhunk://#diff-62ec656d1f81a46b27b8ceb653a8e1db164bbc350bf6f34f9af9383d60b3c190R1-R12): Added test cases for the `wezterm` plugin to ensure it is enabled and its settings are correctly applied.

fixes https://github.com/nix-community/nixvim/issues/2596

### DIFF
--- a/plugins/by-name/molten/default.nix
+++ b/plugins/by-name/molten/default.nix
@@ -80,9 +80,12 @@ mkVimPlugin {
         [
           "none"
           "image.nvim"
+          "wezterm"
         ]
         ''
           How images are displayed.
+
+          If set to `"wezterm"`, `plugins.wezterm.enable` must be true.
         '';
 
     open_cmd = helpers.mkNullOrOption types.str ''
--- a/plugins/by-name/wezterm/default.nix
+++ b/plugins/by-name/wezterm/default.nix
@@ -1,0 +1,39 @@
+{
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "wezterm";
+  originalName = "wezterm.nvim";
+  package = "wezterm-nvim";
+
+  maintainers = [ lib.maintainers.samos667 ];
+
+  extraOptions = {
+    weztermPackage = lib.mkPackageOption pkgs "wezterm" {
+      nullable = true;
+      default = null;
+      example = [ "wezterm" ];
+    };
+  };
+
+  settingsOptions = {
+    create_commands = defaultNullOpts.mkBool true ''
+      Whether to create plugin commands.
+    '';
+  };
+
+  settingsExample = {
+    create_commands = false;
+  };
+
+  extraConfig = cfg: {
+    extraPackages = [
+      cfg.weztermPackage
+    ];
+  };
+}
--- a/tests/test-sources/plugins/by-name/wezterm/default.nix
+++ b/tests/test-sources/plugins/by-name/wezterm/default.nix
@@ -1,0 +1,23 @@
+{
+  empty =
+    { pkgs, ... }:
+    {
+      plugins.wezterm = {
+        enable = true;
+        weztermPackage = pkgs.wezterm;
+      };
+    };
+
+  defaults =
+    { pkgs, ... }:
+    {
+      plugins.wezterm = {
+        enable = true;
+        weztermPackage = pkgs.wezterm;
+
+        settings = {
+          create_commands = true;
+        };
+      };
+    };
+}
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -124,4 +124,10 @@
     githubId = 38228372;
     name = "Katie Janzen";
   };
+  samos667 = {
+    email = "sammyetur11@gmail.com";
+    github = "samos667";
+    githubId = 50653464;
+    name = "samos667";
+  };
 }


tests/test-sources/plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.wezterm.enable = true;
+  };
```
khaneliman, line 4: Can you add a default test set with the default value for the provided option.

plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,21 @@
...
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "wezterm";
+  originalName = "wezterm.nvim";
+  package = "wezterm-nvim";
+
+  maintainers = [ lib.maintainers.samos667 ];
+
+  settingsOptions = {
+    create_commands = defaultNullOpts.mkBool true ''
+      Whether to create plugin commands. Plugin default is true.
```
khaneliman, line 14:
```suggestion
      Whether to create plugin commands.
```
We don't need to specify the default in the description, the `true` from the arguments will generate that in the documentation.

plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "wezterm";
+  originalName = "wezterm.nvim";
+  package = "wezterm-nvim";
```
MattSturgeon, line 8: > `error: vimPlugins.wezterm-nvim cannot be found in pkgs`

Looks like this [is in nixos-unstable now](https://search.nixos.org/packages?channel=unstable&show=vimPlugins.wezterm-nvim&from=0&size=50&sort=relevance&type=packages&query=wezterm-nvim). Rebasing on our latest `main` branch should bring a more up-to-date lock file and resolve this issue :upside_down_face:

plugins/by-name/molten/default.nix
```diff
@@ -80,9 +80,11 @@ mkVimPlugin {
         [
           "none"
           "image.nvim"
+          "wezterm"
         ]
         ''
           How images are displayed.
+          wezterm require plugins.wezterm.enable = true
```
MattSturgeon, line 87: This feels a little clunky. I think the indended wording is something like:

> ``If set to `"wezterm"`, `plugins.wezterm.enable` must be true.``

I'd also have a blankline (to trigger a paragraph break)

---

However it feels strange to have this notice for wezterm but not image.nvim? Maybe the original description was fine as it was?

samos667: idk because image.nvim directly refer to a plugin but not wezterm. But it's as you want, I don't really care

plugins/by-name/wezterm/default.nix
samos667: Should be good now

MattSturgeon:
We still have `error: vimPlugins.wezterm-nvim cannot be found in pkgs`, so I guess we need to do another bump to the flake lock.

plugins/by-name/wezterm/default.nix
MattSturgeon: We still have `error: vimPlugins.wezterm-nvim cannot be found in pkgs`, so I guess we the most recent lock bump was slightly too old.

#2610 should resolve :crossed_fingers:

khaneliman:
@samos667 We updated flake lock today so you should be able to rebase this, now.

GaetanLepage (changes requested):

samos667:
> error: attribute 'samos667' missing

Sadge :'(

khaneliman:
> > error: attribute 'samos667' missing
> 
> 
> 
> Sadge :'(

If you're not in nixpkgs maintainers, you can add yourself to our maintainers file in another commit.

MattSturgeon:
We also have a test failure:

```
defaults> ERROR: Error detected while processing /nix/store/XXX-init.lua:
defaults> Wezterm failed to find 'wezterm' executable
defaults> Wezterm failed to find 'wezterm' executable
```

Perhaps the nixpkgs package doesn't have wezterm marked as a dependency?

MattSturgeon:
The following patch causes the test to pass, however I believe this should be fixed upstream in nixpkgs.

```diff
diff --git a/plugins/by-name/wezterm/default.nix b/plugins/by-name/wezterm/default.nix
index aecb332e..e95b36a4 100644
--- a/plugins/by-name/wezterm/default.nix
+++ b/plugins/by-name/wezterm/default.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ lib, pkgs, ... }:
 let
   inherit (lib.nixvim) defaultNullOpts;
 in
@@ -9,6 +9,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
 
   maintainers = [ lib.maintainers.samos667 ];
 
+  extraPackages = [
+    pkgs.wezterm
+  ];
+
   settingsOptions = {
     create_commands = defaultNullOpts.mkBool true ''
       Whether to create plugin commands.
```

Or if not\*, then we should have a nullable `weztermPackage` option.

```nix
weztermPackage = lib.mkPackageOption pkgs "wezterm" {
  nullable = true;
  # maybe also `default = null` and `example = [ "wezterm" ]`?
};
```

\* Perhaps nixpkgs decide that bundling the dep leads to too many potential issues, such as if users have an overridden/custom wezterm installed to their PATH?

GaetanLepage:
> The following patch causes the test to pass, however I believe this should be fixed upstream in nixpkgs.
> 
> ```diff
> diff --git a/plugins/by-name/wezterm/default.nix b/plugins/by-name/wezterm/default.nix
> index aecb332e..e95b36a4 100644
> --- a/plugins/by-name/wezterm/default.nix
> +++ b/plugins/by-name/wezterm/default.nix
> @@ -1,4 +1,4 @@
> -{ lib, ... }:
> +{ lib, pkgs, ... }:
>  let
>    inherit (lib.nixvim) defaultNullOpts;
>  in
> @@ -9,6 +9,10 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
>  
>    maintainers = [ lib.maintainers.samos667 ];
>  
> +  extraPackages = [
> +    pkgs.wezterm
> +  ];
> +
>    settingsOptions = {
>      create_commands = defaultNullOpts.mkBool true ''
>        Whether to create plugin commands.
> ```

This is still something we have not clearly agreed on in nixpkgs...
I think that the right approach here is having the suggested `weztermPackage` option.

tests/test-sources/plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,23 @@
+{
+  empty =
+    { pkgs, ... }:
+    {
+      plugins.wezterm = {
+        enable = true;
+        package = pkgs.wezterm;
```
GaetanLepage, line 7:
```suggestion
        weztermPackage = pkgs.wezterm;
```

tests/test-sources/plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,23 @@
...
+      };
+    };
+
+  defaults =
+    { pkgs, ... }:
+    {
+      plugins.wezterm = {
+        enable = true;
+        package = pkgs.wezterm;
```
GaetanLepage, line 16:
```suggestion
        weztermPackage = pkgs.wezterm;
```

GaetanLepage (changes requested):

tests/test-sources/plugins/by-name/wezterm/default.nix
MattSturgeon: 🤣

plugins/by-name/wezterm/default.nix
```diff
@@ -0,0 +1,39 @@
...
+
+  maintainers = [ lib.maintainers.samos667 ];
+
+  extraOptions = {
+    weztermPackage = lib.mkPackageOption pkgs "wezterm" {
+      nullable = true;
+      default = null;
+      example = [ "wezterm" ];
+    };
```
MattSturgeon, line 21: What's the best default here? The package or null?

I'm assuming most users who enable this plugin will already have wezterm installed 🤔

MattSturgeon:
@samos667 I've pushed up a couple tweaks. Let us know if you're happy and we should be good to approve & merge.

plugins/by-name/wezterm/default.nix
GaetanLepage: Usually we set the package.

plugins/by-name/molten/default.nix
```diff
@@ -80,9 +80,12 @@ mkVimPlugin {
         [
           "none"
           "image.nvim"
+          "wezterm"
         ]
         ''
           How images are displayed.
+
+          If set to `"wezterm"`, `plugins.wezterm.enable` must be true.
```
GaetanLepage, line 88: Why not adding a warning to check this ? This is usually what we do.

samos667:
Seems good to me !

plugins/by-name/wezterm/default.nix
MattSturgeon: This would achieve that:

```suggestion
    weztermPackage = lib.mkPackageOption pkgs "wezterm" {
      nullable = true;
    };
```

Is there any reason not to do that here?

MattSturgeon:
I believe @GaetanLepage is hoping you'll change the following before we approve/merge:

1. Introduce a `warning` for https://github.com/nix-community/nixvim/pull/2597#discussion_r1876816570
2. Change the `weztermPackage` option to no longer default to null https://github.com/nix-community/nixvim/pull/2597#discussion_r1876736098

khaneliman:
@samos667 Uninterested in continuing this?

samos667:
Sorry, I can not invest myself as a contributor in this project, feel free to take the hand on this

PR #2602, opened 2024-12-02
mkNeovimPlugin: add lazyLoad with lz-n support

khaneliman:
Closes https://github.com/nix-community/nixvim/pull/1875 and https://github.com/nix-community/nixvim/pull/2577


lib/neovim-plugin.nix
```diff
@@ -157,7 +164,37 @@
...
+                ])
+                ++ (lib.optionals hasConfigAttrs [
+                  (lib.mkIf (cfg.lazyLoad.backend == "lz.n") {
+                    plugins.lz-n = {
+                      # infinite recursion?
+                      # enable = true;
+                      plugins = [
+                        {
+                          # TODO: handle this for every plugin properly
+                          __unkeyed-1 = originalName;
```
khaneliman, line 179: I think I see why https://github.com/nix-community/nixvim/pull/1875/files#diff-639400266149b8173dbdc1cf49e1622129bfdc587b52ac1547fc4544e96e415aR46-R49 was implemented, that way... when I was running this code without checking if a lazyLoad.backend was defined, I saw errors due to rtp path depending on which attribute was used.

tests/test-sources/plugins/lazy-load.nix
```diff
@@ -0,0 +1,85 @@
...
+    {
+      plugins = {
+        lz-n = {
+          enable = true;
+        };
+
+        neotest = {
+          enable = true;
+          lazyLoad = {
+            enable = true;
```
Eveeifyeve, line 13: https://github.com/nix-community/nixvim/pull/2577#issuecomment-2509783543

tests/test-sources/plugins/lazy-load.nix
```diff
@@ -0,0 +1,85 @@
...
+      # TODO: assert config exists how we expect
+    };
+
+  lazy-load-lz-n =
+    { config, ... }:
+    {
+      plugins = {
+        codesnap = {
+          enable = true;
+          lazyLoad.enable = true;
```
Eveeifyeve, line 35: https://github.com/nix-community/nixvim/pull/2577#issuecomment-2509783543

tests/test-sources/plugins/lazy-load.nix
MattSturgeon: Global toggle and/or on by default should not be part of the initial implementation.

Down the line, I could envision the per-plugin options _defaulting_ to the value of some global option.

But the initial implementation should be as simple to implement as possible

Eveeifyeve: I think it should be by default when you have lazy loading through `lz-ln` or another one, enabled because It will just make things cleaner in codebases.

MattSturgeon: I'm not saying we shouldn't do this. I'm saying we can have that debate and come up with a solution _after_ this PR.

This PR should be the minimal viable product, so to speak.

lib/neovim-plugin.nix
MattSturgeon: Isn't this all we need?
```suggestion
                          __unkeyed-1 = luaName;
```
I don't think we need a separate lazy-specific name, we just need to use the same name as would be used in `require()` calls. Right?

lib/options.nix
```diff
@@ -333,5 +333,120 @@ rec {
...
+            oneOf [
+              rawLua
+              str
+              (listOf str)
+            ];
+        in
+        types.submodule {
+          options = with defaultNullOpts; {
+
+            enable = lib.mkEnableOption ''
```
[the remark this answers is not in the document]
khaneliman: While working with it, it just felt clunky to have to both enable AND select a backend otherwise no effect. I didn't want to automatically setup the config regardless of lazy loading attempt so we dont run into config breaking behavior if anything happens with the lazy modules and a user didn't enable the feature. But, I agree enable being a boolean is just simpler and then maybe we assert that you've enabled it but haven't selected a backend?

lib/neovim-plugin.nix
khaneliman: I'll try it again, last time I saw failures on some plugins. Which is why I think they had an argument  in case something needed an override.

khaneliman: I think the main issue is it doesn't use the lua module name, its using the pack path directory name. I just tested on my own config and most plugins fail to be detected with `luaName` because we add the derivation to the packDir with the source name. 
image

lib/options.nix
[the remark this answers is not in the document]
khaneliman: I do like that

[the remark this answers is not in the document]
traxys: We could also have a global `programs.nixvim.lazy-loading = { enabled = ....; };` option, and use the `config.lazy-loading.enable` as the default?

lib/neovim-plugin.nix
MattSturgeon: > I think the main issue is it doesn't use the lua module name, its using the pack path directory name.

Interesting - I must've misunderstood how the rtp & lua modules work because I assumed the lua module name was derived from the directory name!

khaneliman: It basically looks like rtp directory is the `originalName` and the `luaName` is the actual lua module from that directory.

lib/options.nix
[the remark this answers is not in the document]
MattSturgeon: That also works, or perhaps the default would be based on the global enable _and_ whether a lazyload trigger has been defined for this specific plugin?

Perhaps the exact defaults for the enable option are drifting out of scope for this PR then. As discussed https://github.com/nix-community/nixvim/pull/2602#discussion_r1866616042, the exact details of how to handle "globally" enabling lazyloading should be figured out after this PR.

Although we still need to think about those details a _little_ bit as they influence decisions like this one!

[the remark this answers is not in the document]
traxys: I think that what we can say is that we will likely have a global toggle that will enable lazy-loading, and that plugins will by default opt-in into that, expect if they set _some set of TBD options_, in which case they will opt-out

lib/neovim-plugin.nix
MattSturgeon: Making me think they should get better names on our end then... "dirName" and "moduleName" perhaps?

lib/options.nix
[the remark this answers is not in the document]
traxys: What I'm not sure about is do we want to support _different_ lazy providers for different plugins. I'm not sure about that, I think we should architecture the initial implementation as different providers exist, but that there will be a global option that allows to choose one

lib/options.nix
```diff
@@ -333,5 +333,120 @@ rec {
...
+            extraSettings = mkSettingsOption {
+              description = ''
+                Extra settings to pass to the lazy loader backend.
+              '';
+              example = {
+                dependencies = {
+                  __unkeyed-1 = "nvim-lua/plenary.nvim";
+                  lazy = true;
+                };
+              };
```
traxys, line 445: It would maybe be better to only support a `backendSettings` option of type `attrsOf anything`, that will be passed to the code that implements the lazy loading provider.
I'm not confident that we can find a common API for all possible lazy loading schemes

lib/options.nix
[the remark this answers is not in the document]
khaneliman: I do have the per plugin backend option added, atm. I was about to remove it because it felt redundant if we just tell someone to enable a provider. But, if we want to support different providers per plugin, it becomes a good option again.

lib/options.nix
khaneliman: That makes sense. I was just changing the options nesting path and this would fit nicely. We can wrap all the `backendSettings` in a `mkSettingsOption` with the known options specified for documentation.

lib/options.nix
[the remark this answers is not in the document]
traxys: I think we should only add it if someone comes with a compelling reason. I don't see any use for it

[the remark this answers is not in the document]
MattSturgeon: Having multiple lazyload implementations at the same time seems pretty silly, but I guess it is a technical possibility.

Would it be unreasonable to have an assertion that only one is ever enabled? It'd certainly simplify things for us...

I personally think it makes the most sense to not have per-plugin "backend" options and instead use either "whatever backend plugin is enabled" or "whatever backend is configured, via a global backend option".

As for the per-plugin enable option: I'm wondering if it'd be intuitive for some plugins to default to `false` and others to default to `true` (depending on if we know a plugin is problematic when lazyloaded).

You would then need _both_ the global and per-plugin options enabled for it to be lazy-loaded.

lib/options.nix
MattSturgeon: > I'm not confident that we can find a common API for all possible lazy loading schemes

Perhaps. But I think we will need _some_ level of abstraction even if it is just knowing where to put our setup call.

That could look like a `setupOptionLoc` option with a default based on which backend is enabled?

E.g.
```nix
setupOptionLoc = mkOption {
  type = listOf str; # maybe a non-merging list instead?
  default =
    if config.pligins.lz-n.enable then
      [ "after" ]
    else if config.lazy.enable then
      [ "config" ]
    else
      throw "setupOptionLoc cannot be used when no lazyloader is enabled";
  defaultText = literalMD "TODO";
}

Comment (traxys) [reply lib/options.nix:445]: I was thinking of a function that takes as an input all the required information.... But that will end up in an infinite recursion.

We *do* need an enumeration of supported backends, but we may not need more than that

File: lib/options.nix [hunk 6]
```diff
@@ -333,5 +333,109 @@ rec {
...
+                      Lazy-load on colorscheme.
+
+                      Equivalence: lz.n => colorscheme; lazy.nvim => None
+                    '';
+                  };
+                };
+            };
+          };
+        };
+      default = lazyLoadDefaults;
```
MattSturgeon, line 438: This feels wrong. If we have defaults they should probably be set on a per-option basis, either when declaring the option or using `lib.mkOptionDefault`.

Generally, the top-level default should be `{}` or `[]`

lib/options.nix
```diff
@@ -333,5 +333,109 @@ rec {
...
+              (listOf str)
+            ];
+        in
+        types.submodule {
+          options = with defaultNullOpts; {
+            enable = lib.mkEnableOption ''
+              lazy-loading for ${originalName}
+            '';
+
+            backendSettings = lib.mkOption {
```
MattSturgeon, line 362: Maybe just name this settings?

lib/neovim-plugin.nix
```diff
@@ -157,7 +166,55 @@
...
+                                "before"
+                                "beforeAll"
+                                "event"
+                                "cmd"
+                                "ft"
+                                "keys"
+                                "colorscheme"
+                                "after"
+                              ])
+                            ) cfg.lazyLoad.backendSettings;
```
MattSturgeon, line 212: 1. This can use `removeAttrs`
2. Does lz-n not have a freeform settings option?

lib/neovim-plugin.nix
```diff
@@ -157,7 +166,55 @@
                 ])
                 ++ (lib.optionals hasConfigAttrs [
                   (lib.optionalAttrs callSetup { ${namespace}.${name}.luaConfig.content = setupCode; })
-                  (lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content))
+                  (lib.mkIf (!cfg.lazyLoad.enable) (
+                    lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content)
+                  ))
+                ])
+                ++ (lib.optionals hasConfigAttrs [
```
MattSturgeon, line 173:
```suggestion
                ++ lib.optionals hasConfigAttrs [
```
:sob: Maybe it's just me, but I actually find it easier to read when there's less parens; it makes the remaining parens more meaningful

lib/neovim-plugin.nix
```diff
@@ -157,7 +166,55 @@
...
+                  (lib.mkIf (config.plugins.lz-n.enable && cfg.lazyLoad.enable && cfg.lazyLoad.backendSettings != { })
```
MattSturgeon, line 174: Maybe this should be nested conditions?

A top-level `mkIf cfg.lazyLoad.enable` that then has assertions and conditionals within the definition?
```nix
(mkIf cfg.lazyLoad.enable {
  plugins.lz-n = mkIf config.plugins.lz-n.enable {};
  plugins.lazy = mkIf config.plugins.lazy.enable {};
  assertions = [
    { assertion = cfg.lazyLoad.backendSettings != { }; }
    { assertion = length (filter (x: config.plugins.${x}.enable) [ "lz-n" "lazy" ]) == 1; }
  ];
})
```

lib/neovim-plugin.nix
```diff
@@ -89,6 +89,13 @@
           options.${namespace}.${name} =
             {
               enable = lib.mkEnableOption originalName;
+              lazyLoad = lib.nixvim.mkLazyLoadOption originalName;
+              # TODO:
+              # lazyLoadDefaults = {
+              #   # FIXME: infinite recursion
+              #   # enable = lib.mkDefault (cfg.lazyLoad.backendSettings != { });
+              # };
+              # };
```
MattSturgeon, line 98: Looks like this is infinitely recursive because you can't use the `lazyload` option's value to define the `lazyload` option itself.

Unless I'm misinterpreting what `lazyLoadDefaults` is/does?

---

I believe the inf-rec could be avoided by working _within_ the submodule, e.g.:

```nix
types.submodule (
  { config, ... }:
  {
    # e.g. using a config def, but maybe an option default makes more sense?
    config.enable = lib.mkDefault (config.backendSettings != {});
  }
)
```

lib/neovim-plugin.nix
```diff
@@ -152,13 +159,62 @@
...
+                ]
+                ++ lib.optionals hasConfigAttrs [
                   (lib.optionalAttrs callSetup { ${namespace}.${name}.luaConfig.content = setupCode; })
-                  (lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content))
-                ])
+                  (lib.mkIf (!cfg.lazyLoad.enable) (
+                    lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content)
+                  ))
+                ]
+                ++ lib.optionals hasConfigAttrs [
```
MattSturgeon, line 171:
```suggestion
```

The block above is already `optionals hasConfigAttrs`

lib/neovim-plugin.nix
```diff
@@ -152,13 +159,62 @@
                     }
                   ))
                 ]
-                ++ (lib.optionals (!hasConfigAttrs) [
+                ++ lib.optionals (!hasConfigAttrs) [
                   (lib.optionalAttrs callSetup (setLuaConfig setupCode))
-                ])
-                ++ (lib.optionals hasConfigAttrs [
+                ]
```
MattSturgeon, line 164: How should the `!hasConfigAttrs` case be handled when lazyloading?

lib/neovim-plugin.nix
```diff
@@ -152,13 +159,62 @@
...
+                      {
+                        assertion =
+                          lib.length (
+                            lib.filter (x: config.plugins.${x}.enable) [
+                              "lz-n"
+                              "lazy"
+                            ]
+                          ) == 1;
+                        message = "You have enabled multiple lazy providers. Enable only one.";
+                      }
```
MattSturgeon, line 187: My mistake: this should not be a per-plugin assertion.

Instead, we should probably have a global "lazyload" module somewhere that defines this assertion only once.

Perhaps such a module could end up being the future home of any _global_ lazyloading options too?

Might tackle that in a separate PR, since it shouldn't depend on anything in this one.

lib/neovim-plugin.nix
khaneliman: Looks like we only have one instance of no config attrs, in rustaceanvim, because they use global options for configuration. Seems like something you don't get to really lazy load through a manager and is handled by the plugin itself?

MattSturgeon: Certainly for lz-n, there's nothing to lazyload if there's no setup function. lazy.nvim is a bit fancier though.

If we _did_ go down that route, maybe the `plugins.*.lazyload` option should only exist when `hasSetup` is true? Although I'm happy for that to be tackled in future work if it is too much for this PR.

lib/neovim-plugin.nix
MattSturgeon: https://github.com/nix-community/nixvim/pull/2609

lib/options.nix
```diff
@@ -333,5 +333,114 @@ rec {
...
+              str
+              (listOf str)
+            ];
+        in
+        types.submodule (
+          { config, ... }:
+          {
+            options = with defaultNullOpts; {
+              enable = lib.mkOption {
+                default = config.settings != { };
```
khaneliman, line 361: Gotta tweak this a bit. Realized it will be true if

```nix
neotest = {
  enable = true;
  lazyLoad = {};
};

**Comment by khaneliman** [comment]
Started working on lazy.nvim integration but it's going to be a pretty involved change, it seems. Would rather handle that separately.

File: lib/options.nix [hunk 16]
```diff
@@ -333,5 +333,45 @@ rec {
...
+      description = ''
+        Lazy-load settings for ${originalName}.
+      '';
+      default = { };
+      type = types.submodule (
+        { config, ... }:
+        {
+          options = {
+            enable = lib.mkOption {
+              default = lib.any (x: x != null) (builtins.attrValues config.settings);
```
MattSturgeon, line 349: This should have a `literalExpression` or `literalMD` type defaultText. Maybe something like:

```nix
defaultText = literalMD ''
  `true` when `settings` has a non-null attribute
'';
```

----

But while we're discussing the option:

Is this actually what we want long-term, or would we rather have most/all plugins default to `lazy.enable = true` but have it _also_ require a global `lazy.enable`?

If we did go with a global enable, should it be an explicit (user-set) option, or should it just be an internal option based on having a lazyload impl plugin enabled?

tests/test-sources/plugins/lazyloading/lz-n.nix
```diff
@@ -0,0 +1,307 @@
...
+          assertion = (builtins.length config.plugins.lz-n.plugins) == 1;
+          message = "`lz-n.plugins` should have contained a single plugin configuration, but contained ${builtins.toJSON config.plugins.lz-n.plugins}";
+        }
+        {
+          assertion =
+            let
+              plugins = config.plugins.lz-n.plugins or [ ];
+              plugin = if builtins.length plugins > 0 then builtins.head plugins else null;
+              keys = if plugin != null && builtins.isList plugin.keys then plugin.keys else [ ];
+            in
```
MattSturgeon, line 38: Hm... I wonder if it's worth having a small util file that can be imported by the relevant tests? This seems to be getting fairly repetitive?

Not a big deal though, if you're happy with it.

khaneliman:
Ripped out the lazy.nvim implementation, for now. Until I can figure out why the errors are being thrown. Creating a new follow up PR for that.

lib/options.nix
khaneliman: I think it makes more sense, for now. I think we'd need to update plugins with sane defaults of cmd/ft/etc triggers to support enabling by default without user configuration.

khaneliman:
@GaetanLepage Not sure of a good way to implement the global warning. It feels messy atm with my local wip implementation and breaks my existing lazy loading tests due to throwing a warning just for usage.

GaetanLepage:
> @GaetanLepage Not sure of a good way to implement the global warning. It feels messy atm with my local wip implementation and breaks my existing lazy loading tests due to throwing a warning just for usage.

Well, I would then propose to do a per-plugin warning. I will surely be more verbose (who cares ?), but at least the implem will be more straightforward.

khaneliman:
> > @GaetanLepage Not sure of a good way to implement the global warning. It feels messy atm with my local wip implementation and breaks my existing lazy loading tests due to throwing a warning just for usage.
> 
> Well, I would then propose to do a per-plugin warning. I will surely be more verbose (who cares ?), but at least the implem will be more straightforward.

Actually, forgot we can just disable checking warnings and the assertions should still trigger and fail the test.

GaetanLepage:
> > 
> 
> Actually, forgot we can just disable checking warnings and the assertions should still trigger and fail the test. I was thinking it was all or none for some reason.

Ok great. Both solutions are fine to me.

modules/lazyload.nix
```diff
@@ -26,5 +27,31 @@ in
...
+          x:
+          !(lib.elem x ignoredPackages)
+          && lib.hasAttr "lazyLoad" config.plugins.${x}
+          && config.plugins.${x}.lazyLoad.enable
+        ) (builtins.attrNames config.plugins);
+        count = builtins.length pluginsWithLazyLoad;
+      in
+      lib.optionals (count > 0) [
+        ''
+          You have enabled experimental lazy loading support.
```
GaetanLepage, line 51:
```suggestion
          [Nixvim] WARNING: You have enabled experimental lazy loading support.
```
Something like this ?

khaneliman: I'll add the Nixvim prefix, similar to how we do throughout the codebase. But, I don't know if I want to make it sound scary :P

GaetanLepage: Sure, this is completely fair.

lib/options.nix
MattSturgeon: Agreed. I hadn't fully thought this one through. Of course we can't lazy load plugins that have no "trigger" events configured.

modules/lazyload.nix
```diff
@@ -20,11 +20,37 @@ in
...
+        ) (builtins.attrNames config.plugins);
+        count = builtins.length pluginsWithLazyLoad;
+      in
+      lib.optionals (count > 0) [
+        ''
+          Nixvim: You have enabled experimental lazy loading support:
+          ${lib.concatImapStringsSep "\n" (i: x: "${toString i}. plugins.${x}") pluginsWithLazyLoad}
+          This is not a stable API and may have breaking changes.
+        ''
+      ];
```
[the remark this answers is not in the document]
GaetanLepage: Wouldn't putting the warning in `mkNeovimPlugin` (hence on a per-plugin basis) significantly simplify the implementation ?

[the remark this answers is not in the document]
khaneliman: > This is a non-issue for the assertion, since a lazyload impl plugin will always be in the plugins namespace. However for lazy loaded plugins I believe colorschemes are also supported?

Yes, this misses colorschemes.

I'm not truly a fan of the warning, to begin with, since I like warnings to be actionable to remove from my flake switches. Just from using this branch for a while it kinda irks me seeing it all the time. :S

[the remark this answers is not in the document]
GaetanLepage: Ok, I think that you are right on this @khaneliman.
Let's have a warning in the option description. Also, we wil be very clear in our announcement that this is experimental. No need to clutter people's terminal.

MattSturgeon (commented):
Looks good. Just last minute discussions

lib/neovim-plugin.nix
```diff
@@ -161,8 +162,41 @@
...
+                      {
+                        assertion = (isColorscheme && colorscheme != null) || cfg.lazyLoad.settings != { };
+                        message = "You have enabled lazy loading for ${originalName} but have not provided any configuration.";
+                      }
+                      {
+                        assertion = cfg.lazyLoad.enable && (config.plugins.lz-n.enable || config.plugins.lazy.enable);
+                        message = "You have enabled lazy loading for ${originalName} but have not enabled any lazy loading plugins.";
+                      }
+                    ];
+                    plugins.lz-n = lib.mkIf config.plugins.lz-n.enable {
```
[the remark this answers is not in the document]
MattSturgeon: Not sure I follow? Defining `lz-n.plugins` is effectivly a noop if `lz-n.enable = false`

lib/neovim-plugin.nix
```diff
@@ -161,8 +162,43 @@
...
+
+                  # When lazy loading is enabled for this plugin, route its configuration to the enabled provider
+                  (lib.mkIf cfg.lazyLoad.enable {
+                    assertions = [
+                      {
+                        assertion = (isColorscheme && colorscheme != null) || cfg.lazyLoad.settings != { };
+                        message = "You have enabled lazy loading for ${originalName} but have not provided any configuration.";
+                      }
+                      {
+                        assertion = cfg.lazyLoad.enable && (config.plugins.lz-n.enable || config.plugins.lazy.enable);
```
traxys, line 176: For a future PR it could be usefull to have a way to enable lazyLoading without a provider, in order to implement lazy  loading for a  provider not yet supported by nixvim

lib/neovim-plugin.nix
```diff
@@ -161,8 +162,43 @@
...
+                        (
+                          {
+                            __unkeyed-1 = originalName;
+                            # Use provided after, otherwise fallback to normal function wrapped lua content
+                            after =
+                              let
+                                after = cfg.lazyload.settings.after or null;
+                                default = ("function()\n " + cfg.luaConfig.content + " \nend");
+                              in
+                              if lib.isString after then after else default;
```
traxys, line 191: shouldn't we check if after is defined? If a user does `after = lib.mkRaw { .... }`  this will be surprising

lib/neovim-plugin.nix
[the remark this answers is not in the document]
khaneliman: If i change the option type inside plugins.lz-n that is a breaking change, it doesn't matter if lz-n is enabled or not if we try to pass an incorrect type to an option.

lib/neovim-plugin.nix
khaneliman: Thinking this should just be a warning instead of an assertion, then?

khaneliman: I think the reason it's an assertion right now, is because if you use this option we handle the plumbing and we can only guarantee it works for things we have wired up.

lib/neovim-plugin.nix
MattSturgeon: Good point re-raw types... However I don't think simply checking it is defined is enough...

My intention was to check if defined via `or null`, and exclude unexpected types (permitted by `freeform = attrsOf anything`) using `isString`.

I'm also unsure whether blindly passing a raw type through to `lz-n.plugins.*.after` may or may not be handled correctly?

If we can handle raw, then maybe we want to check "`isString` or `types.rawLua.check`"? Or if we're being strict about not coercing strings->lua, maybe we should only accept rawLua?

lib/neovim-plugin.nix
MattSturgeon: I think assertion is best for now, but we could address ways to bypass the assertion in a future PR?

khaneliman: Hmm.. I just switched to a warning just because I saw the potential for enabling lazy loading just to pass the luaConfig around yourself. ie: Someone wants to try using it with lazy before we officially support it.

lib/neovim-plugin.nix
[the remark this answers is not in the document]
khaneliman: That's true, just felt like this was good defensive coding to avoid wasting time evaluating something that has no positive effect and only has possible negatives.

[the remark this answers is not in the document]
MattSturgeon: > avoid wasting time evaluating something that has no positive effect and only has possible negatives

Worth noting that it will only be evaluated if `plugins.lz-n.plugins` itself gets evaluated, which I believe should only happen when `plugins.lz-n` is enabled? :upside_down_face:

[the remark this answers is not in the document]
khaneliman: I created a test case to verify it will bomb out if i removed the condition and just passed configuration with an invalid option type. But, will use that as a smoke test if we do it in the future, for whatever reason.

lib/neovim-plugin.nix
```diff
@@ -161,8 +162,48 @@
...
+                                after = cfg.lazyLoad.settings.after or null;
+                                default = "function()\n " + cfg.luaConfig.content + " \nend";
+                              in
+                              if (lib.isString after || lib.types.rawLua.check after) then after else default;
+                            colorscheme = lib.mkIf isColorscheme (cfg.lazyLoad.settings.colorscheme or colorscheme);
+                          }
+                          // (lib.removeAttrs cfg.lazyLoad.settings [
+                            "after"
+                            "colorscheme"
+                          ])
```
MattSturgeon, line 202:
```suggestion
                          // lib.removeAttrs cfg.lazyLoad.settings [
                            "after"
                            "colorscheme"
                          ]
```

lib/options.nix
```diff
@@ -333,5 +333,54 @@ rec {
...
           example;
     };
+
+  mkLazyLoadOption =
+    originalName:
+    lib.mkOption {
+      description = ''
+        Lazy-load settings for ${originalName}.
+
+
```
MattSturgeon, line 343:
```suggestion

```

Double blank line intentional?

tests/test-sources/plugins/lazyloading/lz-n.nix
```diff
@@ -0,0 +1,337 @@
...
+                  __unkeyed-3 = "<CMD>Neotest summary<CR>";
+                  desc = "Summary toggle";
+                }
+              ];
+            };
+          };
+        };
+      };
+
+      assertions =
```
MattSturgeon, line 37: Q: is it intended that (all?) these assertion-based tests also run `nvim` as part of the test? Or should we set `test.runNvim = false` to make the tests cheaper?

lib/neovim-plugin.nix
traxys: Yeah my first thoughts was doing it in a subsequent PR, but no problem if it is being done here.

tests/test-sources/plugins/lazyloading/lz-n.nix
khaneliman: Good callout, we don't really need to run nvim for these test scenarios since we're just trying to make sure we pass things around properly.

lib/neovim-plugin.nix
```diff
@@ -161,8 +162,48 @@
...
+                      }
+                    ];
+
+                    warnings = lib.optional (!config.plugins.lz-n.enable) ''
+                      You have enabled lazy loading for ${originalName} but have not enabled any lazy loading plugins.
+                      We will not insert the lua for this plugin and you must handle loading it yourself.  
+
+                      Currently supported lazy providers:
+                        - lz-n
+                    '';
```
traxys, line 183: This means that if someone wants to use an unsupported lazy loader they will have warnings for each of their plugins right?

In another PR we should consider adding a global option that will allow to silence the warnings, something like `useUnsupportedLazyLoader`

lib/neovim-plugin.nix
MattSturgeon: I believe the new formatting would also permit
```nix
// lib.removeAttrs cfg.lazyLoad.settings [ "after" "colorscheme" ]
```

Although whether that's bettor or worse in this case is up for debate. If there's a chance the list well grow with time, the current formatting is preferable to minimize  noise in future patches.

GaetanLepage (approved):
LGTM !!!

MattSturgeon (approved):
Thanks for working on this!

GaetanLepage:
@mergifyio queue

PR #2609, opened 2024-12-05
modules/lazyload: init w/ assertion

MattSturgeon:
Adds a global lazyloading assertion. No module options yet, and none are likely to be added until after we have per-plugin options.


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #2616, opened 2024-12-08
plugins/colorizer: rename, switch to mkNeovimPlugin

GaetanLepage:
Rework the `nvim-colorizer` plugin:
- Rename it to `colorizer` (proper user warnings have been introduced)
- Switch to `mkNeovimPlugin`


khaneliman (commented):
Took me far too long to realize which `nvim-colorizer-lua` we're packaging... found a few different variations searching lol.

GaetanLepage:
> Took me far too long to realize which `nvim-colorizer-lua` we're packaging... found a few different variations searching lol.

Yes, me too. We are packaging https://github.com/catgoose/nvim-colorizer.lua.

khaneliman (approved):

plugins/by-name/colorizer/default.nix
```diff
@@ -0,0 +1,251 @@
...
+  imports =
+    let
+      oldPluginPath = [
+        "plugins"
+        "nvim-colorizer"
+      ];
+      newPluginPath = [
+        "plugins"
+        "colorizer"
+      ];
```
khaneliman, line 26: Actually, just realized... we should move this to lib.deprecation.nix

khaneliman: https://github.com/nix-community/nixvim/blob/4b5364a66bffd22536e358214b37068b34287a7a/plugins/deprecation.nix#L20-L23

GaetanLepage: Well, this won't really let me do those internal option renames, will it ?

khaneliman: Ah, I suppose you had modernization stuff too. No, that's setup just for top level

khaneliman (approved):

plugins/by-name/colorizer/default.nix
GaetanLepage: > Ah, I suppose you had modernization stuff too. No, that's setup just for top level. This is probably cleaner, then. Since you'd have to include this stuff anyways.

Yes, and I would be afraid that it would clash.
What happens if you have both:
- rename `x.a` -> `x.b`
- rename `x` -> `y`

What happens if you set `y.a` ? `x.a` ?

MattSturgeon: > What happens if you have both:
> 
>     * rename `x.a` -> `x.b`
> 
>     * rename `x` -> `y`
> 
> 
> What happens if you set `y.a` ? `x.a` ?

I believe that alias options are smart enough to handle those scenarios, but I'd feel more comfortable seeing it empirically.

plugins/by-name/colorizer/default.nix
```diff
@@ -0,0 +1,251 @@
...
+          (
+            newPluginPath
+            ++ [
+              "settings"
+              "user_default_options"
+              optionName
+            ]
+          );
+    in
+    (map renameOption optionsToRename)
```
MattSturgeon, line 63: Nit: no need for parens here
```suggestion
    map renameOption optionsToRename
```

plugins/by-name/colorizer/default.nix
```diff
@@ -0,0 +1,251 @@
...
+    ++ (lib.mapAttrsToList
```
MattSturgeon, line 64: Same
```suggestion
    ++ lib.mapAttrsToList
```

MattSturgeon (approved):
No issues. Nits below are meant as FYI, no action needed for merge :upside_down_face:

plugins/by-name/colorizer/default.nix
MattSturgeon: Yeah, a single entry in `deprecation.nix` may be nicer, but since we have to move all the options to `settings` anyway, it makes little difference.

GaetanLepage:
> No issues. Nits below are meant as FYI, no action needed for merge 🙃

I applied your suggestions. Thanks for the review :)

GaetanLepage:
@mergifyio queue

GaetanLepage:

PR #2623, opened 2024-12-09
mkNeovimPlugin: refactor lua code generation logic

GaetanLepage:
Follow-up of #2619

This keeps the current behavior of not adding a `plugins.*.luaConfig` option if it doesn't make sense (for `rustacenvim` for e.g.).
What has been removed is the ability/expectation for `configLocation` to be `null`.


lib/neovim-plugin.nix
```diff
@@ -143,21 +143,26 @@
...
-                ++ (lib.optionals hasConfigAttrs [
+                # Lua configuration code generation
+                ++ (lib.optionals hasLuaConfig [
+
+                  # Add the plugin setup code `require('foo').setup(...)` to the lua configuration
                   (lib.optionalAttrs callSetup { ${namespace}.${name}.luaConfig.content = setupCode; })
-                  (lib.optionalAttrs (configLocation != null) (setLuaConfig cfg.luaConfig.content))
+
+                  # Write the lua configuration `luaConfig.content` to the config file
+                  (setLuaConfig cfg.luaConfig.content)
```
MattSturgeon, line 165: Q: why is it ok to remove the null check? (`configLocation != null`)

GaetanLepage: Because it should never be `null`.
`null` is a forbidden value. Or at least, it is fine as long as `hasLuaConfig` is `false`.

MattSturgeon (approved):

traxys:
I made #2624 to see how we could implement a `luaConfig` attribute for plugins that are configured with globals. If the design is good enough maybe we should have an argument of mkNeovimPlugin which would allow rustaceanvim to use that?

GaetanLepage:
@mergifyio queue

PR #2624, opened 2024-12-09
lib/vim-plugin: Add support for `luaConfig`

traxys:
This series allows to use `luaConfig` for vim plugins that are configured through global variables

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -137,47 +137,63 @@ rec {
       }";
     };
 
-  pluginLuaConfig = types.submodule (
-    { config, ... }:
+  pluginLuaConfig =
+    { hasContent }:
     let
       inherit (builtins) toString;
       inherit (lib.nixvim.utils) mkBeforeSection mkAfterSection;
-    in
-    {
-      options = {
-        pre = lib.mkOption {
-          type = with types; nullOr lines;
-          default = null;
-          description = ''
-            Lua code inserted at the start of the plugin's configuration.
-            This is the same as using `lib.nixvim.utils.mkBeforeSection` when defining `content`.
-          '';
-        };
-        post = lib.mkOption {
-          type = with types; nullOr lines;
-          default = null;
-          description = ''
-            Lua code inserted at the end of the plugin's configuration.
-            This is the same as using `lib.nixvim.utils.mkAfterSection` when defining `content`.
-          '';
-        };
-        content = lib.mkOption {
-          type = types.lines;
-          default = "";
-          description = ''
-            Configuration of the plugin.
 
-            If `pre` and/or `post` are non-null, they will be merged using the order priorities
-            ${toString (mkBeforeSection null).priority} and ${toString (mkBeforeSection null).priority}
-            respectively.
-          '';
+      commonModule = {
+        options = {
+          pre = lib.mkOption {
+            type = with types; nullOr lines;
+            default = null;
+            description =
+              ''
+                Lua code inserted at the start of the plugin's configuration.
+              ''
+              + lib.optionalString hasContent ''
+                This is the same as using `lib.nixvim.utils.mkBeforeSection` when defining `content`.
+              '';
+          };
+          post = lib.mkOption {
+            type = with types; nullOr lines;
+            default = null;
+            description =
+              ''
+                Lua code inserted at the end of the plugin's configuration.
+              ''
+              + lib.optionalString hasContent ''
+                This is the same as using `lib.nixvim.utils.mkAfterSection` when defining `content`.
+              '';
+          };
         };
       };
 
-      config.content = lib.mkMerge (
-        lib.optional (config.pre != null) (mkBeforeSection config.pre)
-        ++ lib.optional (config.post != null) (mkAfterSection config.post)
-      );
-    }
-  );
+      contentModule =
+        { config, ... }:
+        {
+          options = {
+            content = lib.mkOption {
+              type = types.lines;
+              default = "";
+              description = ''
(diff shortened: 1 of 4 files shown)


lib/types.nix
```diff
@@ -137,47 +137,60 @@ rec {
...
+              description =
+                ''
+                  Lua code inserted at the end of the plugin's configuration.
+                ''
+                + lib.optionalString hasContent ''
+                  This is the same as using `lib.nixvim.utils.mkAfterSection` when defining `content`.
+                '';
+            };
+          }
+          // (lib.optionalAttrs hasContent) {
```
MattSturgeon, line 174:
```suggestion
          // lib.optionalAttrs hasContent {
```

I don't think this would cause eval errors, but it seems like a typo left over from some previous code?

lib/types.nix
```diff
@@ -137,47 +137,60 @@ rec {
...
+            };
+          };
 
-      config.content = lib.mkMerge (
-        lib.optional (config.pre != null) (mkBeforeSection config.pre)
-        ++ lib.optional (config.post != null) (mkAfterSection config.post)
-      );
-    }
-  );
+        config = lib.optionalAttrs hasContent {
```
MattSturgeon, line 188: This feels like it'd be simpler to have two modules, e.g. using `types.submoduleWith`

```nix
types.submoduleWith {
  modules =
    [
      {/* mod1 has common options */}
    ]
    ++ lib.optionals hasContent [
      {/*mod2 has content option */}
      ({config, ...}: {/* mod3 has config definitions*/})
    ];
};
```

Or maybe using a let-in?
```nix
let
  commonModule = {};
  configOptionsModule = {};
  configMergingModule = {config, ...}: {};
in
types.submoduleWith {
  modules =
    [ commonModule ]
    ++ lib.optionals hasContent [
      configOptionsModule
      configMergingModule
    ];
};
```

lib/types.nix
```diff
@@ -137,47 +137,63 @@ rec {
...
+
+          config = {
+            content = lib.mkMerge (
+              lib.optional (config.pre != null) (mkBeforeSection config.pre)
+              ++ lib.optional (config.post != null) (mkAfterSection config.post)
+            );
+          };
+        };
+    in
+    types.submodule ([ commonModule ] ++ (lib.optional hasContent contentModule));
```
MattSturgeon, line 198: Does this work? I didn't think `types.submodule` could accept a list :thinking: 

```suggestion
    types.submoduleWith {
      modules = [ commonModule ] ++ lib.optional hasContent contentModule;
    };
```

traxys: https://nixos.org/manual/nixos/stable/#sec-option-types-submodule
> This is equivalent to types.submoduleWith { modules = toList o; shorthandOnlyDefinesConfig = true; }

MattSturgeon: > > This is equivalent to `modules = toList o`

Thanks for educating me! :heart:

traxys: For me the `let in` seems clearer

modules/opts.nix
```diff
@@ -36,15 +36,29 @@ let
...
+        default = "";
+        internal = true;
+      };
+
+      globalsPost = lib.mkOption {
+        type = lib.types.lines;
+        default = "";
+        internal = true;
+      };
+    };
```
MattSturgeon, line 61: I guess the intention is that this is lua code injected pre/post the lua impl for the `globals` option.

But reading the option name I'd half-expect to be able to assign `globals` earlier or later than normal.

This feels like working around a poor design rather than an objective improvement.

MattSturgeon: I'm wondering if instead, each plugin should be responsible for using `vim.g[k] = ` itself. Perhaps within `luaConfig.init` ?

Yes, this would mean vim-plugins aren't taking advantage of the `globals` option, and consiquently end-users won't see the configured globals when inspecting the option's value. But perhaps that is worth it in order to locate the globals in a more predictable location?

MattSturgeon: Another idea: Perhaps the `luaConfig` submodule could also be used for non-plugin modules too? E.g. we could have a `globalsLuaConfig` option, and move the lua impl to `globalsLuaConfig.init`?

_(maybe rename `init` something like `main`, `content`, `text`, etc?)_

MattSturgeon: I'm not really a fan of any of the three solutions, tbh. So open to alternative proposals too.

MattSturgeon (commented):
Sorry it took me until now to read past `lib/types.nix` and actually look at the main part of the PR...

modules/opts.nix
traxys: > I'm wondering if instead, each plugin should be responsible for using `vim.g[k] = ` itself. Perhaps within `luaConfig.init` ?
> 
> Yes, this would mean vim-plugins aren't taking advantage of the `globals` option, and consiquently end-users won't see the configured globals when inspecting the option's value. But perhaps that is worth it in order to locate the globals in a more predictable location?

I though about that, but I think setting the global options in nix is a bit better, as it allows to do more compile time stuff (though you could _still_ do compile time stuff by accessing the plugin's settings)

traxys: > Another idea: Perhaps the `luaConfig` submodule could also be used for non-plugin modules too? E.g. we could have a `globalsLuaConfig` option, and move the lua impl to `globalsLuaConfig.init`?
> 
> _(maybe rename `init` something like `main`, `content`, `text`, etc?)_

There _could_ be a use for it, for example to define helpers for keymaps, or options

PR #2626, opened 2024-12-10
flake-modules/new-plugin: init

khaneliman:
Allows quickly creating new plugins to help fulfill plugin creation requests.

### DIFF
--- a/flake-modules/dev/new-plugin.py
+++ b/flake-modules/dev/new-plugin.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+
+import os
+import re
+from argparse import ArgumentParser
+
+# Template for default.nix
+# TODO: conditionally include parts of the template based on args
+default_nix_template = """{{ lib, ... }}:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {{
+  name = "{name}";
+  originalName = "{originalName}";
+  package = "{package}";
+
+  maintainers = [ lib.maintainers.YOUR_NAME ];
+
+  settingsOptions = {{
+    exampleOption = lib.nixvim.defaultNullOpts.mkBool false ''
+      Example option for the {name} plugin.
+    '';
+  }};
+}}
+"""
+
+# Template for test file
+test_nix_template = """{{
+  empty = {{
+    plugins.{name}.enable = true;
+  }};
+}}
+"""
+
+
+def kebab_case(input_string):
+    """
+    Convert a string to kebab-case.
+
+    Args:
+        input_string (str): The input string to convert.
+
+    Returns:
+        str: The converted kebab-case string.
+    """
+    # Replace non-alphanumeric characters with hyphens
+    input_string = re.sub(r"[\W_]+", "-", input_string).lower()
+
+    # Remove leading and trailing standalone 'nvim'
+    input_string = re.sub(r"(^nvim-|-nvim$|^nvim$)", "", input_string)
+
+    return input_string.strip("-")
+
+
+def create_nix_file(file_path, template, name, originalName, package):
+    """
+    Create a nix file from a template.
+
+    Args:
+        file_path (str): The path to the file to create.
+        template (str): The template string to use for the file content.
+        name (str): The name of the plugin.
+        originalName (str): The original name of the plugin.
+        package (str): The package name of the plugin.
+    """
+    content = template.format(name=name, originalName=originalName, package=package)
+    write_to_file(file_path, content)
+
+
+def create_test_file(file_path, template, name):
+    """
+    Create a test file from a template.
+
+    Args:
+        file_path (str): The path to the file to create.
+        template (str): The template string to use for the file content.
+        name (str): The name of the plugin.
+    """
+    content = template.format(name=name)
+    write_to_file(file_path, content)
+
+
+def write_to_file(file_path, content: str):
+    """
+    Makes sure directories exist and write content to a file.
+
+    Args:
+    file_path (str): The path to the file to write.
+    content (str): The content to write to the file.
+    """
+    os.makedirs(os.path.dirname(file_path), exist_ok=True)
+    with open(file_path, "w") as f:
+        f.write(content)
+
+
+# TODO: support interactive unmanaged args
+def main():
+    """
+    Main function to generate default.nix and test files for a new plugin.
+    """
+    parser = ArgumentParser(
+        description="Generate default.nix and test files for a new plugin"
+    )
+    parser.add_argument(
+        "originalName", type=str, help="Original name of the new plugin"
+    )
+    parser.add_argument("package", type=str, help="Package name of the new plugin")
+    args = parser.parse_args()
+
+    # Calculate name
+    name = kebab_case(args.originalName)
+
+    # Define paths
+    plugin_path = f"plugins/by-name/{name}/default.nix"
+    test_path = f"tests/test-sources/plugins/by-name/{name}/default.nix"
+
+    # Create files
+    create_nix_file(
+        plugin_path,
+        default_nix_template,
+        name,
+        args.originalName,
+        args.package,
+    )
+    create_test_file(
+        test_path,
+        test_nix_template,
+        name,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/flake-modules/dev/new-plugin.py
+++ b/flake-modules/dev/new-plugin.py
@@ -91,6 +91,18 @@ def write_to_file(file_path, content: str):
         f.write(content)
 
 
+def find_project_root(root_identifier):
+    current_path = os.getcwd()
+    while True:
+        if root_identifier in os.listdir(current_path):
+            return current_path
+        parent_path = os.path.dirname(current_path)
+        if parent_path == current_path:
+            return None
+        os.chdir("..")
+        current_path = os.getcwd()
+
+
 # TODO: support interactive unmanaged args
 def main():
     """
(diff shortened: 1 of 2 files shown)


flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,79 @@
...
+  empty = {{
+    plugins.{name}.enable = true;
+  }};
+}}
+"""
+
+
+def snake_case(input_string):
+    # Convert to snake_case and remove trailing underscores
+    return re.sub(r"[\W_]+", "_", input_string).lower().replace("nvim", "").rstrip("_")
```
MattSturgeon, line 35: Feels like this may benefit from some unit tests.

One edge case, for example: how does it handle leading `nvim-` prefixes?

Also, don't we normally use kebab-case for plugin names?

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,79 @@
...
+
+def main():
+    parser = ArgumentParser(
+        description="Generate default.nix and test files for a new plugin"
+    )
+    parser.add_argument(
+        "originalName", type=str, help="Original name of the new plugin"
+    )
+    parser.add_argument("package", type=str, help="Package name of the new plugin")
+    args = parser.parse_args()
```
MattSturgeon, line 58: As a TODO item, it'd be nice to have an interactive option for any unspecified args

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,79 @@
...
+  originalName = "{originalName}";
+  package = "{package}";
+
+  maintainers = [ lib.maintainers.YOUR_NAME ];
+
+  settingsOptions = {{
+    exampleOption = lib.nixvim.defaultNullOpts.mkBool false ''
+      Example option for the {name} plugin.
+    '';
+  }};
```
MattSturgeon, line 20: As a TODO item, it'd be nice to have a way to toggle whether a plugin has settings, should call `setup` etc.

Having settings could also mean a `defaults` test case is added.

GaetanLepage:
Good initiative, my rust attempt is likely to stale for a while !

khaneliman:
> Good initiative, my rust attempt is likely to stale for a while !

Oh yeah... sorry forgot about that one

flake-modules/dev/new-plugin.py
khaneliman: Yeah, late night brain fart lol.

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def kebab_case(input_string):
+    # Convert to snake_case and remove trailing underscores
+    return (
+        re.sub(r"[\W_]+", "-", input_string)
+        .lower()
+        .replace("nvim", "")
+        .rstrip("-")
+        .lstrip("-")
```
MattSturgeon, line 40:
```suggestion
        .strip("-")
```
:upside_down_face: https://docs.python.org/3.4/library/stdtypes.html#str.strip

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,85 @@
...
+test_nix_template = """{{
+  empty = {{
+    plugins.{name}.enable = true;
+  }};
+}}
+"""
+
+
+def kebab_case(input_string):
+    # Convert to snake_case and remove trailing underscores
```
MattSturgeon, line 34: Comment is out of date; it now converts to kebab case.

Wondering if we should use doc-strings too? https://www.programiz.com/python-programming/docstrings

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,85 @@
...
+        .replace("nvim", "")
+        .rstrip("-")
+        .lstrip("-")
+    )
+
+
+def create_nix_file(file_path, template, name, originalName, package):
+    content = template.format(name=name, originalName=originalName, package=package)
+    with open(file_path, "w") as f:
+        f.write(content)
```
MattSturgeon, line 47: I'm thinking we should extract creating dirs + writing text to a file to a helper fn.

That way the `main()` fn doesn't have to take responsibility for ensuring dirs exist and without having to make these `create_*_file` fns more repetitive.

```py
def write_to_file(file_path, content: str):
    os.makedirs(os.path.dirname(file_path), exist_ok=True)
    with open(file_path, "w") as f:
        f.write(content)
```

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,85 @@
...
+    plugins.{name}.enable = true;
+  }};
+}}
+"""
+
+
+def kebab_case(input_string):
+    # Convert to snake_case and remove trailing underscores
+    return (
+        re.sub(r"[\W_]+", "-", input_string)
```
MattSturgeon, line 36: `\W` is non-word chars, right? I'm wondering if we'll end up also needing to convert camelCase -> kebab-case? If so we'll need a capture group for `((?<![A-Z])[A-Z])` or something.

MattSturgeon:
> One thing that came up in https://github.com/nix-community/nixvim/pull/2643 made me wonder about how to handle multiple instances of `nvim` in a name. It stripped them all out and made it `remote` but out plugin creation script in nixpkgs generated `remote-nvim-nvim` and the plugin itself refers to itself in lua as `remote-nvim`

I have two thoughts on this.

1. This script doesn't need to be perfect. The contributor should review and tweak the generated file anyway.
2. If we later add an interactive mode, this will be even easier
    (I'm imagining something similar to the GitHub CLIs interactive commands, e.g. `gh repo create` and `gh pr create` - those make use of Charm's bubblegum TUI widgets)
3. We can probably use a smarter regex here; only removing `nvim` if it is at the start/end: `'^nvim|nvim$'`
  Or we could use a `removePrefix('nvim')` and `removeSuffix('nvim')` combo

khaneliman:
Added the TODO items and tweaked replace to only remove leading and trailing `nvim` occurrences so that we can preserve them when it's part of the name.

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        input_string (str): The input string to convert.
+
+    Returns:
+        str: The converted kebab-case string.
+    """
+    input_string = re.sub(r"(?<!^)(?=[A-Z])", "-", input_string)
+
+    return re.sub(r"[\W_]+", "-", input_string).lower().strip("nvim").strip("-")
```
MattSturgeon, line 46: Can `strip()` accept a regex? If so we could leverage word boundaries to remove some edge-cases. If not, no big deal.

 Maybe also catch the `-` in one pass using `-*`, although that may be less readable than chaining multiple `.strip()` calls...

```suggestion
    return re.sub(r"[\W_]+", "-", input_string).lower().strip(r"\bnvim\b").strip("-")
```

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,119 @@
...
+    name = kebab_case(args.originalName)
+
+    # Define paths
+    plugin_path = f"plugins/by-name/{name}/default.nix"
+    test_path = f"tests/test-sources/plugins/by-name/{name}/default.nix"
+
+    # Create files
+    create_nix_file(
+        plugin_path, default_nix_template, name, args.originalName, args.package
+    )
```
GaetanLepage, line 114:
```suggestion
    create_nix_file(
        file_path=plugin_path,
        template=default_nix_template,
        name=name,
        originalName=args.originalName,
        package=args.package,
    )
```
I tend to write this kind of function calls like this.
This is 100% subjective: feel free to ignore.

flake-modules/dev/new-plugin.py
khaneliman: I went with chaining calls in case stripping `nvim` caused there to be a leftover `-`.

But, looks like `strip` doesn't accept a regex so we have to use `sub`

khaneliman:
Alright, handled cases where `nvim` might be in a word and then formatted the function calls differently.

khaneliman:
@GaetanLepage our checks dont like it though :P

GaetanLepage:
> @GaetanLepage our checks dont like it though :P

Hmm, weird. Have you appended a comma after the last arg ?

khaneliman:
> > @GaetanLepage our checks dont like it though :P
> 
> Hmm, weird. Have you appended a comma after the last arg ?

Ah, I think that was it. I didn't have a trailing comma so the formatter reverted it back.

MattSturgeon (approved):
All my feedback is addressed. IDK if @GaetanLepage has anything else to add as our resident python guy :grin: 

Note to self for followup work: 

- [ ] Interactive mode
- [ ] Dynamic template sections
- [ ] Document in `CONTRIBUTING`

flake-modules/dev/new-plugin.py
```diff
@@ -0,0 +1,131 @@
...
+    parser.add_argument("package", type=str, help="Package name of the new plugin")
+    args = parser.parse_args()
+
+    # Calculate name
+    name = kebab_case(args.originalName)
+
+    # Define paths
+    plugin_path = f"plugins/by-name/{name}/default.nix"
+    test_path = f"tests/test-sources/plugins/by-name/{name}/default.nix"
```
traxys, line 113: We should maybe get the repo root in order to make those absolute paths? Except if the devshell command automatically cd's at the root, but I'm not sure about that

khaneliman: Oh, are you thinking if someone is `cd`-ing around the repo while in the devshell? Didn't try that..

khaneliman: Pushed an update to make sure we generate based on the project root.

traxys (approved):

khaneliman:
@GaetanLepage any other concerns before we merge first implementation?

khaneliman:
@mergify queue

PR #2628, opened 2024-12-10
flake-modules/dev: cosmetic refactor

GaetanLepage:


flake-modules/dev/default.nix
```diff
@@ -12,9 +12,6 @@
       system,
       ...
     }:
-    let
-      fmt = pkgs.nixfmt-rfc-style;
-    in
```
MattSturgeon, line 17: This was leftover from when we defined a nixfmt package for both `treefmt` and the pre-commit hook.

MattSturgeon (approved):

GaetanLepage:
@mergifyio queue

PR #2630, opened 2024-12-10
docs/user-guide: add lazy-loading.md

khaneliman:
Just getting this started and creating a draft PR if anyone has input on what they'd like mentioned.

### DIFF
--- a/docs/user-guide/lazy-loading.md
+++ b/docs/user-guide/lazy-loading.md
@@ -1,0 +1,93 @@
+# Lazy Loading
+
+> [!WARNING]
+> This is an experimental option and may not work as expected with all plugins.
+> The API may change without notice.
+>
+> Please report any issues you encounter.
+
+> [!NOTE]
+> When the lazy loading functionality is enabled, the plugin's configuration is
+> routed to the enabled lazy provider. Options mentioned will be in relation to
+> the `${namespace}.${name}.lazyLoad` context.
+
+## Supported Lazy Loading Providers
+
+It is recommended to become familiar with lazy loading strategies supported by
+your provider in their upstream documentation and the corresponding plugin's
+support.
+
+Currently we only support **one** lazy loader:
+
+- lz.n: [`plugins.lz-n`](../plugins/lz-n/index.md) -
+  [upstream docs](https://github.com/nvim-neorocks/lz.n?tab=readme-ov-file#plugin-spec)
+
+## Enabling Lazy Loading Per Plugin
+
+You can enable lazy loading for a plugin by passing a configuration to
+`lazyLoad.settings`. This configuration is passed through to the lazy loading
+provider's loading mechanism.
+
+If you are just wanting to store potential configuration without enabling it,
+you can explicitly disable it setting `lazyLoad.enable = false`.
+
+By default, we route the generated setup code to your lazy loading provider, but
+it can be overridden by using the `lazyLoad.settings` option.
+
+For `Lz.n`, the `lazyLoad.settings.after` option contains what it will call when
+a trigger condition is met.
+
+## Configuring Triggers
+
+You need to define the trigger conditions in which a plugin will be loaded. This
+is done through the `lazyLoad.settings` option.
+
+```nix
+plugins = {
+  grug-far = {
+    enable = true;
+    lazyLoad = {
+      settings = {
+        cmd = "GrugFar";
+      };
+    };
+  };
+};
+```
+
+```nix
+plugins = {
+  glow = {
+    enable = true;
+    lazyLoad.settings.ft = "markdown";
+  };
+```
+
+```nix
+plugins.toggleterm = {
+  enable = true;
+  lazyLoad = {
+    settings = {
+      cmd = "ToggleTerm";
+      keys = [
+        "<leader>tg"
+        "<leader>gg"
+      ];
+    };
+  };
+};
+```
+
+### Colorschemes
+
+Colorschemes do not require explicit settings configuration. In `lz-n`, we will
+set up the `colorscheme` trigger to the name of the `colorscheme` so that it is
+lazy loaded when the `colorscheme` is requested.
+
+```nix
+colorscheme = "catppuccin";
+colorschemes.catppuccin = {
+  enable = true;
+  lazyLoad.enable = true;
+};
+```
--- a/docs/mdbook/SUMMARY.md
+++ b/docs/mdbook/SUMMARY.md
@@ -8,6 +8,7 @@
 - [Helpers](./user-guide/helpers.md)
 - [FAQ](./user-guide/faq.md)
 - [Configuration examples](./user-guide/config-examples.md)
+- [Lazy Loading](./user-guide/lazy-loading.md)
 
 # Platforms
 


khaneliman:
Marking as ready.. curious if there's anything else other's want added.

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,88 @@
...
+routed to the enabled lazy provider. Options mentioned will be in relation to
+the `${namespace}.${name}.lazyLoad` context.
+
+## Supported Lazy Loading Providers
+
+It is recommended to become familiar with lazy loading strategies supported by
+your provider in their upstream documentation and the corresponding plugin's
+support.
+
+- [Lz.n](https://github.com/nvim-neorocks/lz.n?tab=readme-ov-file#plugin-spec)
```
MattSturgeon, line 17:
```suggestion
Currently we only support **one** lazy loader:
- [Lz.n](https://github.com/nvim-neorocks/lz.n?tab=readme-ov-file#plugin-spec)
```

MattSturgeon: Also, wondering if this should have links to our docs too, e.g:

```suggestion
- lz.n: [`plugins.lz-n`](/plugins/lz-n/index.md) - [upstream docs](https://github.com/nvim-neorocks/lz.n?tab=readme-ov-file#plugin-spec)
```

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Enabling Lazy Loading Per Plugin
+
+There are a couple different ways to enable lazy loading for a particular
+plugin.
+
+- `enable` set to `true`.
+- `settings` containing some configuration.
```
MattSturgeon, line 25:
```suggestion
- `lazyLoad.enable` set to `true`.
- `lazyLoad.settings` containing some configuration.
```
?

MattSturgeon: Maybe we don't need to really mention `enable` here though; we could say something like 

> Configuring `lazyLoad.settings.*` will enable lazyloading. If needed, it can be explicitly disabled using `lazyLoad.enable = false`.

?

EDIT: I see you have something similar below
> If you are just wanting to store potential configuration without enabling it, you can explicitly disable it setting `enable = false`.

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,88 @@
...
+will be passed along to the corresponding lazy provider instead. This behavior
+can be overridden by using the `settings` option that will be passed to the
+provider.
+
+For `Lz.n`, that is the `settings.after` option.
+
+## Configuring Triggers
+
+Currently, you need to define the trigger conditions in which a plugin will be
+loaded. This is done through the `settings` option.
```
MattSturgeon, line 40: I'd suggest moving away from "Currently".

Maybe one-day we could decide to offer some sane defaults, but I also wonder if that'd end up being an anti-pattern.

https://github.com/nix-community/nixvim/issues/2632#issuecomment-2540912650

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,88 @@
...
+        "<leader>gg"
+      ];
+    };
+  };
+};
+```
+
+The only exception to this is `colorschemes`. In `lz-n`, we will set up the
+`colorscheme` trigger to the name of the `colorscheme` so that it is lazy loaded
+when the `colorscheme` is requested.
```
MattSturgeon, line 80: Maybe this could be re-framed as a dedicated `## Colour schemes` section, rather than a footnote paragraph in the triggers section?

We could rephrase _"The only exception is..."_ to something more positive like _"Additionally, plugins in the `colorschemes.*` namespace will..."_ :thinking:

MattSturgeon (commented):
Some suggestions below.

Do you have any goals in mind before you undraft?

My preference would be to get an initial version of this guide merged _soon_ and then make improvements to it iteratively.

Thanks for working on this, it's looking amazing! :rocket:

docs/user-guide/lazy-loading.md
khaneliman: I mention at the top that all the settings are in relation to the lazyLoad setting to reduce repetition everywhere. 
>When the lazy loading functionality is enabled, the plugin's configuration is
routed to the enabled lazy provider. Options mentioned will be in relation to
the `${namespace}.${name}.lazyLoad` context.

If this is easy to miss I can just inline it everywhere.

docs/user-guide/lazy-loading.md
khaneliman: I can't run docs on my mac, so will have to verify later that path works :P

khaneliman:
> Do you have any goals in mind before you undraft?
> 
> My preference would be to get an initial version of this guide merged _soon_ and then make improvements to it iteratively.
> 
> Thanks for working on this, it's looking amazing! 🚀

Thanks, I realized I had been letting it draft too long so I undrafted it just before your review. :P

docs/user-guide/lazy-loading.md
MattSturgeon: > If this is easy to miss I can just inline it everywhere.

Yeah, I think `lazyLoad.settings` is clearer than `settings`, especially to anyone skim-reading.[^1] Having it mentioned explicitly at the top is great, but let's not rely _exclusively_ on that :upside_down_face: 

[^1]: Because we all _totally_ read third-party docs fully and in detail while rushing around trying to get our stuff to work :grin:

khaneliman: While proofreading after the change, I do agree it is easier to follow.

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Supported Lazy Loading Providers
+
+It is recommended to become familiar with lazy loading strategies supported by
+your provider in their upstream documentation and the corresponding plugin's
+support.
+
+Currently we only support **one** lazy loader:
+
+-lz.n: [`plugins.lz-n`](/plugins/lz-n/index.md) -
```
MattSturgeon, line 20: My bad, this seems to skip past the base-href. I guess we need a relative link.

```suggestion
- lz.n: [`plugins.lz-n`](../plugins/lz-n/index.md) -
```

Also: need a ` ` space after the `-` to render as a list item

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,94 @@
...
+[upstream docs](https://github.com/nvim-neorocks/lz.n?tab=readme-ov-file#plugin-spec)
```
MattSturgeon, line 21: Not sure, but I believe this needs to stay on the same line, as list items are parsed differently to normal paragraphs.

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,94 @@
...
+## Enabling Lazy Loading Per Plugin
+
+There are a couple different ways to enable lazy loading for a particular
+plugin.
+
+- `lazyLoad.enable` set to `true`.
+- `lazyLoad.settings` containing some configuration.
+
+If you are just wanting to store potential configuration without enabling it,
+you can explicitly disable it setting `lazyLoad.enable = false`.
```
MattSturgeon, line 32: Continuing from https://github.com/nix-community/nixvim/pull/2630#discussion_r1884715931

I wonder if this would be clearer without mention that `enable` can be used to enable... Seeing is you should never really need to enable without also configuring `settings`, right?

Therefore I'd suggest we remove the list and replace the first para with an explanation about what `settings` is and the fact that configuring it will auto-enable.

The final para already covers use cases for `enable = false`, so I think that is ok as-is?

docs/user-guide/lazy-loading.md
MattSturgeon: Turns out, this renders fine as-is.

docs/user-guide/lazy-loading.md
khaneliman: doh, thanks missed that

docs/user-guide/lazy-loading.md
khaneliman: That's fair, I was about to add a mention about requiring the settings and then realized it felt weird to even mentioning enabling, by itself.

docs/user-guide/lazy-loading.md
```diff
@@ -0,0 +1,91 @@
+# Lazy Loading
+
+> [!WARNING]
+> This is an experimental option and may not work as expected with all plugins.
+> The API may change without notice. > Please report any issues you encounter.
```
MattSturgeon, line 5:
```suggestion
> The API may change without notice.
>
> Please report any issues you encounter.
```
or
```suggestion
> The API may change without notice. \
> Please report any issues you encounter.
```

Was this indented?

MattSturgeon: image

khaneliman: Hmm my conform kept moving it up, for some reason.

khaneliman: Alright, fixed it.
image

MattSturgeon (approved):
This is an excellent addition to the docs. Thanks! :bookmark_tabs:

khaneliman:
@mergify queue

PR #2633, opened 2024-12-11
plugins/hmts: fix warning

khaneliman:


plugins/by-name/hmts/default.nix
```diff
@@ -24,9 +23,9 @@ in
   };
 
   config = mkIf cfg.enable {
-    warnings = optional (!config.plugins.treesitter.enable) [
-      "Nixvim: hmts needs treesitter to function as intended"
-    ];
+    warnings = optional (
+      !config.plugins.treesitter.enable
+    ) "Nixvim: hmts needs treesitter to function as intended";
```
MattSturgeon, line 28: We could achieve the same thing with a smaller diff by doing `optional` -> `optionals`

MattSturgeon (approved):

plugins/by-name/hmts/default.nix
khaneliman: True, just thought you preferred the optional for single lists :P

khaneliman:
@mergify queue

plugins/by-name/hmts/default.nix
MattSturgeon: No real preference. `optional` is often neater, however nixfmt seems to not be a massive fan looking at this :grin:

khaneliman: They definitely aren't... that's why I rarely used it because it's often discouraged in nixpkgs.

khaneliman:
@mergify refresh

khaneliman:
@mergify dequeue

khaneliman:
Sorry, just realized this could be upgraded.

MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2637, opened 2024-12-11
plugins/lint: migrate to mkNeovimPlugin

HeitorAugustoLN:


plugins/by-name/lint/default.nix
```diff
@@ -227,41 +218,46 @@ in
...
+              if builtins.isString linterConfig then
+                ''lint.linters["${linter}"] = ${toLuaObject lib.nixvim.mkRaw linterConfig}''
+              else
+                lib.mapAttrsToList (
+                  propName: propValue:
+                  lib.optionalString (
+                    propValue != null
+                  ) ''lint.linters["${linter}"]["${propName}"] = ${toLuaObject propValue}''
+                ) linterConfig
+            ) cfg.settings.linters
```
MattSturgeon, line 258: I haven't read this yet to see what it does (I'll wait for you to undraft and/or ask for review), but as a more general tip:

Anything under `settings.*` should be options defined and handled by the actual lua plugin. If we have to manipulate the option value in some way and/or use it for something, then it should probably be a nixvim-specific option that lives outside of `settings`.

Such custom options can be declared by using `extraOptions` instead of `settingsOptions`.

HeitorAugustoLN:
I guess this is ready by now. The CI will probably be failing until my upstream fix (mfussenegger/nvim-lint#704) gets merged and packages get updated. I merged `customLinters` and `linters` into just `linters`, to be closer as it is in upstream

plugins/by-name/lint/default.nix
```diff
@@ -86,113 +84,49 @@ let
...
-                {
-                  type = types.nullOr type;
-                  default = null;
-                }
-            )
-          )
-        )
-      ) linterOptions;
-    };
+  mkLinterOpts = types.submodule {
```
MattSturgeon, line 90: If this is no longer a factory-function, it should not be named `mk*`. As this is a concrete submodule type, maybe something like `linterType` or `linterOptsType` is more appropriate?

plugins/by-name/lint/default.nix
```diff
@@ -1,20 +1,18 @@
...
-  cfg = config.plugins.lint;
+  inherit (lib)
+    mkOption
+    mkRemovedOptionModule
+    mkRenamedOptionModule
+    types
+    ;
+  inherit (lib.nixvim) defaultNullOpts toLuaObject;
 
   linterOptions = with types; {
```
MattSturgeon, line 6: I'd like to avoid having `with` used with too wide a scope.

I'm fine with using it at a "final"/"leaf node" assignment,[^1] but I'm less comfortable having it before a large attrset.

[^1]: e.g. `type = with types; foo bar baz` is ok, but `with types; { foo = "bar"; baz = 123; # ...` is not ideal.

MattSturgeon (commented):
This is a fairly complex plugin to port over to mkNeovimPlugin and the PR is in a pretty good state. Well done!

plugins/by-name/lint/default.nix
```diff
@@ -225,43 +159,70 @@ in
...
+            linter: linterConfig:
             let
-              linterConfig' = if isString linterConfig then helpers.mkRaw linterConfig else linterConfig;
+              linterConfig' =
+                if builtins.isString linterConfig then lib.nixvim.mkRaw linterConfig else linterConfig;
             in
-            ''__lint.linters["${customLinter}"] = ${helpers.toLuaObject linterConfig'}''
-          ) cfg.customLinters
+            ''lint.linters.${linter} = ${toLuaObject linterConfig'}''
+          ) cfg.linters
```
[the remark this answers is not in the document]
HeitorAugustoLN: >Do all this plugin's options take the form `require('lint')[optname] = optval?` And there is no `.setup()` or similar function?

Yes, this plugin doesn't have `.setup()`. Which I quite like, to be honest, that's why I even thought than putting into settings, since these options were the settings of it.

> I'm wondering if we can still have a freeform settings option, but have its toplevel attrs relate to keys in the `require('lint')[key] = val` assignment?

I don't think a freeform will be needed, since the plugin is quite simple, and it will probably not add anything else.

plugins/by-name/lint/default.nix
```diff
@@ -225,43 +159,70 @@ in
...
+        lint.linters_by_ft = ${toLuaObject cfg.linters_by_ft}
+      '')
+      + (lib.optionalString (cfg.linters != null) (
+        lib.concatLines (
+          lib.mapAttrsToList (
+            linter: linterConfig:
             let
-              linterConfig' = if isString linterConfig then helpers.mkRaw linterConfig else linterConfig;
+              linterConfig' =
+                if builtins.isString linterConfig then lib.nixvim.mkRaw linterConfig else linterConfig;
```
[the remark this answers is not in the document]
HeitorAugustoLN: You're right, from my testing this doesn't look like it is needed anymore

plugins/by-name/lint/default.nix
```diff
@@ -86,113 +79,49 @@ let
...
-        phpcs.args = [
-          "-q"
-          "--report=json"
-          "-"
-        ];
-      };
-    };
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "lint";
+  originalName = "nvim-lint";
```
khaneliman, line 113:
```suggestion
  packPathName = "nvim-lint";
```

plugins/by-name/lint/default.nix
```diff
@@ -225,43 +154,57 @@ in
           callback = defaultCallback;
         };
       };
+
+    linters =
+      defaultNullOpts.mkAttrsOf linterType
```
khaneliman, line 159: Might be nicer to move the let block down here so it's usage is close to its definition.

plugins/by-name/lint/deprecations.nix
```diff
@@ -0,0 +1,13 @@
...
+let
+  inherit (lib) mkRemovedOptionModule;
+in
+{
+  imports = [
+    (mkRemovedOptionModule [
+      "plugins"
+      "lint"
+      "customLinters"
+    ] "plugins.lint.customLinters option was removed, use linters instead.")
```
khaneliman, line 11: I haven't looked too close at the implementation, does this change require your upstream PR to be valid?

HeitorAugustoLN: Yes, it does. The current nvim-lint would consider any linter that uses an table assignment (e.g. `lint.linters.name = { prop = value }`), a new linter, even if it has a builtin one with the same name. That is probably why the `customLinters` option was needed.

khaneliman: I'd recommend pulling that change out until we get an updated plugin, then.

HeitorAugustoLN: I thought we could wait for the plugin to be updated to merge this PR

khaneliman: Just not sure how fast upstream is to respond to PRs and then there would be a wait time until it gets updated in nixpkgs and that package makes it to nixos-unstable. If you don't want to handle it in a separate PR, then probably just mark it a draft and come back to it later.

HeitorAugustoLN: I don't think it will take too long, but while it doesn't get merged I will convert the PR to a draft.

HeitorAugustoLN: Plugin author didn't want to merge because of breaking changes, so reverted to the old format.

plugins/by-name/lint/default.nix
HeitorAugustoLN: Not needed anymore

khaneliman (approved):
LGTM, thanks

plugins/by-name/lint/default.nix
```diff
@@ -225,43 +172,81 @@ in
...
+  };
 
-    extraConfigLua =
+  extraConfig = cfg: {
+    autoCmd = lib.optionals (cfg.autoCmd != null) [ cfg.autoCmd ];
+    plugins.lint.luaConfig.content =
       ''
-        __lint = require('lint')
-        __lint.linters_by_ft = ${lib.nixvim.toLuaObject cfg.lintersByFt}
+        local lint = require('lint')
```
khaneliman, line 219: Only concern is changing the naming here. I believe we've been doing __ prefixes to prevent messing with a user's config.

HeitorAugustoLN: Reverted back to `__lint`.

khaneliman:
@mergify queue

PR #2639, opened 2024-12-11
plugins/otter: fix treesitter warning

khaneliman:
Doesn't actually check if treesitter is enabled.

### DIFF
--- a/plugins/by-name/otter/default.nix
+++ b/plugins/by-name/otter/default.nix
@@ -101,10 +101,12 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings =
-      lib.optional (cfg.enable && config.plugins.treesitter.settings.highlight.enable == null)
+      lib.optional
+        (config.plugins.treesitter.enable -> config.plugins.treesitter.settings.highlight.enable == null)
+
         ''
-          NixVim(plugins.otter): you have enabled otter, but `plugins.treesitter.settings.highlight.enable` is not enabled.
-          Otter functionality might not work as expected without it and `plugins.treesitter.enable` enabled.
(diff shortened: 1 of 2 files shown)


plugins/by-name/otter/default.nix
```diff
@@ -101,10 +101,16 @@ helpers.neovim-plugin.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings =
-      lib.optional (cfg.enable && config.plugins.treesitter.settings.highlight.enable == null)
+      lib.optional
+        (
+          cfg.enable
+          && (
+            !config.plugins.treesitter.enable || config.plugins.treesitter.settings.highlight.enable == null
```
MattSturgeon, line 108:
```suggestion
            config.plugins.treesitter.enable -> config.plugins.treesitter.settings.highlight.enable == null
```

`!x || y` is the same as `x -> y`

---

Would need to double check [operator precedence](https://nix.dev/manual/nix/2.17/language/operators) though, as these mean different things:

```nix
(x -> y) ==  null
x -> (y == null)
```
One is the same as `x -> y == null`, but which one?

MattSturgeon: Ok, equality has prio 11 while logical `||` and `->` have 13 and 14 respectively. Therefore parens are not needed.

`x -> (y == null)` and `x -> y == null` are equivalent. Which makes intuitive sense, too.

MattSturgeon (commented):
Can take advantage of `->` logical implication, but otherwise looks good

plugins/by-name/otter/default.nix
khaneliman: Thanks, that looks cleaner. :)

traxys (approved):

MattSturgeon (approved):

khaneliman:
@mergify queue

khaneliman:

Issue #2133, opened 2024-08-31
[PLUGIN REQUEST] grug-far


tknawara:
| Field    | Description |
|----------|-------------|
| Plugin   | `grug-far`  | 
| Homepage | `https://github.com/MagicDuck/grug-far.nvim`     | 
| Nixpkgs  | `https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/editors/vim/plugins/generated.nix#L4684` | 

## Extra Information

it is a nice plugin for doing global find/replace utilizing ripgrep, I'm not sure if you want to have ripgrep installed as part of enabling the plugin though. I put nixpkgs as false for now, but ripgrep is available in nixpkgs

0x006E:
@tknawara Currently available in nixpkgs.  [vimPlugins.grug-far-nvim](https://search.nixos.org/packages?channel=24.11&show=vimPlugins.grug-far-nvim&from=0&size=50&sort=relevance&type=packages&query=grug-far). It would if I could use this instead of nvim-spectre

PR #2641, opened 2024-12-11
plugins/grug-far: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2133

### DIFF
--- a/plugins/by-name/grug-far/default.nix
+++ b/plugins/by-name/grug-far/default.nix
@@ -1,0 +1,23 @@
+{ lib, ... }:
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "grug-far";
+  originalName = "grug-far.nvim";
+  package = "grug-far-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsExample = {
+    debounceMs = 1000;
+    minSearchChars = 1;
+    maxSearchMatches = 2000;
+    normalModeSearch = false;
+    maxWorkers = 8;
+    engine = "ripgrep";
+    engines = {
+      ripgrep = {
+        path = "rg";
+        showReplaceDiff = true;
+      };
+    };
+  };
+}
--- a/tests/test-sources/plugins/by-name/grug-far/default.nix
+++ b/tests/test-sources/plugins/by-name/grug-far/default.nix
@@ -1,0 +1,57 @@
+{
+  empty = {
+    plugins.grug-far.enable = true;
+  };
+
+  # Just a small subset of the settings
+  defaults = {
+    plugins.grug-far = {
+      enable = true;
+
+      settings = {
+        debounceMs = 500;
+        minSearchChars = 2;
+        maxSearchMatches = 2000;
+        normalModeSearch = false;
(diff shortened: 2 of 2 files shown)


tests/test-sources/plugins/by-name/grug-far/default.nix
```diff
@@ -0,0 +1,37 @@
+{
+  empty = {
+    plugins.grug-far.enable = true;
+  };
+
+  # Just a small subset of the settings
+  defaults = {
+    plugins.grug-far = {
+      enable = true;
+      settings = {
```
MattSturgeon, line 10:
```suggestion
      enable = true;

      settings = {
```

@GaetanLepage would make this suggestion. Personally I don't care too much, but I guess we should be consistent :joy:

plugins/by-name/grug-far/default.nix
```diff
@@ -0,0 +1,24 @@
...
+    searchOnInsertLeave = true;
+    maxWorkers = 8;
+    engine = "ripgrep";
+    engines = {
+      ripgrep = {
+        path = "rg";
+        showReplaceDiff = true;
+      };
+    };
+  };
```
MattSturgeon, line 23: I'm assuming it's intentional to have this (excellent) settings example and the defaults test case, but not actually declare any settings sub-options?

Not complaining! Just double checking. :heart_eyes:

tests/test-sources/plugins/by-name/grug-far/default.nix
MattSturgeon, line 1: Would there be any value in also having an `example` test case based on the settingsExample? Or is that a waste of resources?

plugins/by-name/grug-far/default.nix
khaneliman: I just saw the wealth of options they had and don't have the energy to document any atm until I spend any time with the plugin and see what's really useful.

MattSturgeon: I actually prefer not having them as nix options, especially if it seems they may be difficult to understand/maintain.

traxys: My heuristic is to look at the readme, if they explicitly mention ~5 options I'd add them, else I would not add anything

GaetanLepage: > I'm assuming it's intentional to have this (excellent) settings example and the defaults test case, but not actually declare any settings sub-options?

I see myself in this picture and I don't like it.

tests/test-sources/plugins/by-name/grug-far/default.nix
GaetanLepage: > @GaetanLepage would make this suggestion.

I would indeed.
ADD THIS SPACE.

NOW.
Or I'm blocking this PR

GaetanLepage: (And please, declare all the 420 options of this plugin with proper types and sub-element-rawtype-support, or this doesn't get merged.)

traxys (approved):

khaneliman:
@mergify queue

Issue #2400, opened 2024-10-10
[PLUGIN REQUEST] dap-lldb


Sk7Str1p3:
| Field    | Description |
|----------|-------------|
| Plugin   | `nvim-dap-lldb`  | 
| Homepage | `https://github.com/julianolf/nvim-dap-lldb`     | 
| Nixpkgs  | `false` | 

## Extra Information

Extention for nvim-dap providing C/C++/Rust support

PR #2644, opened 2024-12-12
plugins/dap-lldb: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2400


mathjiajia:
would this be an extension of `dap` like `dap-python`?

khaneliman:
> would this be an extension of `dap` like `dap-python`?

Yeah, the dap extensions basically just try and make it easier to setup nvim-dap with various languages/debuggers.

MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2648, opened 2024-12-12
plugins/lz-n: earlier lua loading

khaneliman:
Discovered during testing lazy loading colorscheme that we aren't loading the config early enough to catch all triggers.

### DIFF
--- a/plugins/pluginmanagers/lz-n.nix
+++ b/plugins/pluginmanagers/lz-n.nix
@@ -11,6 +11,8 @@ nixvim.neovim-plugin.mkNeovimPlugin {
   name = "lz-n";
   originalName = "lz.n";
   maintainers = [ maintainers.psfloyd ];
+  # NOTE: We want to load lz.n as early as possible so that triggers are respected
+  configLocation = "extraConfigLuaPre";
 
   settingsDescription = ''
     Options provided to `vim.g.lz_n`.


traxys:
You could I believe use `configLocation` & set it to `extraConfigLuaPre`

khaneliman:
> You could I believe use `configLocation` & set it to `extraConfigLuaPre`

Forgot about that... I'll try that real quick.

traxys (approved):

GaetanLepage (approved):

plugins/pluginmanagers/lz-n.nix
```diff
@@ -11,6 +11,8 @@ nixvim.neovim-plugin.mkNeovimPlugin {
   name = "lz-n";
   originalName = "lz.n";
   maintainers = [ maintainers.psfloyd ];
+  # NOTE: We want to load lz.n as early as possible so that triggers are respected
+  configLocation = "extraConfigLuaPre";
```
MattSturgeon, line 15: Makes me wonder if `configLocation` should support `mkOrder` wrappers?

To allow us to more explicitly say things like `configLocation = mkOrder 69 "extraConfigLuaPre"`?

MattSturgeon (approved):

plugins/pluginmanagers/lz-n.nix
traxys: I think it would be useful

khaneliman:
@mergify queue

PR #2649, opened 2024-12-12
colorschemes/catppuccin: add originalName

khaneliman:
Supports lazy loading

Lz-n will look for the colorscheme in packpath by originalName. Without change it will throw an error when trying to lazy load. This plugin has a unique `originalName` because the upstream repo is just `nvim` and upstream package is being added to packpath as `catppuccin-nvim`.


traxys (approved):

khaneliman:
@mergify queue

PR #2650, opened 2024-12-12
plugins/neorg: switch to mkNeovimPlugin

GaetanLepage:


plugins/by-name/neorg/default.nix
```diff
@@ -1,173 +1,249 @@
...
+        basePluginPath
+        ++ [
+          "telescopeIntegration"
+          "package"
+        ]
+      ))
+    ];
+
+  extraOptions = {
+    telescopeIntegration = {
```
khaneliman, line 82: If we're already creating a bunch of renames maybe follow suite of the others `enableTelescope`?

plugins/by-name/neorg/default.nix
```diff
@@ -1,173 +1,249 @@
...
-
-    package = lib.mkPackageOption pkgs "neorg" {
-      default = [
-        "vimPlugins"
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "neorg";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  # TODO introduced 2024-12-12: remove after 25.05
```
khaneliman, line 21: One thing we talked about recently was creating a separate `deprecations.nix` that we import next to each file in our `by-name` structure so that it doesn't pollute the main file. 

`plugins/by-name/neorg/deprecations.nix`

GaetanLepage: Good point. I didn't move the `optionsRenamedToSettings` though as it is less straightforward.

plugins/by-name/neorg/default.nix
```diff
@@ -1,173 +1,222 @@
...
+
+  extraOptions = {
+    telescopeIntegration = {
+      enable = lib.mkEnableOption "`neorg-telescope` plugin for telescope integration.";
+
+      package = lib.mkPackageOption pkgs "neorg-telescope" {
+        default = [
+          "vimPlugins"
+          "neorg-telescope"
+        ];
```
khaneliman, line 62:
```suggestion
        ];
        nullable = true;
```

GaetanLepage: I don't really want it nullable...
Well actually maybe it is stupid to have `enable` + non-`null`able-`package`.
Maybe we should only have a `null`able `neorg-telescopePackage` ?

khaneliman: Oh, is the only thing we're doing in here installing the package? Normally we have to add the module name to our telescope module

GaetanLepage: Here it's a plugin install indeed.

khaneliman: If it doesn't require `require('telescope').load_extension(extension)` and we're just adding the package, then I agree it doesn't make sense to be nullable. It also feels too small to be its own module.

plugins/by-name/neorg/default.nix
```diff
@@ -1,173 +1,222 @@
...
-
-      extraPlugins = [
-        cfg.package
-      ] ++ (optional (telescopeSupport && cfg.neorgTelescopePackage != null) cfg.neorgTelescopePackage);
-
-      extraConfigLua = ''
-        require('neorg').setup(${helpers.toLuaObject setupOptions})
-      '';
-    };
+    extraPlugins = [ cfg.telescopeIntegration.package ];
```
khaneliman, line 203:
```suggestion
    extraPlugins = lib.mkIf cfg.telescopeIntegration.enable [ cfg.telescopeIntegration.package ];
```

If we keep the enable option, we should check that they even want it installed.

plugins/by-name/neorg/default.nix
GaetanLepage: Ok, then I would say we keep the non-`null`able `package` + `enable` options.
Indeed, if a user wants to enable the feature, he only has to `telescopeIntegration.enable = true;` and not have to manually set a plugin value.

GaetanLepage:
@mergifyio queue

PR #2653, opened 2024-12-13
plugins/nix-develop: migrate to mkNeovimPlugin

HeitorAugustoLN:


plugins/by-name/nix-develop/deprecations.nix
```diff
@@ -0,0 +1,16 @@
...
+in
+{
+  imports = [
+    (mkRenamedOptionModule
+      [ "plugins" "nix-develop" "ignoredVariables" ]
+      [ "plugins" "nix-develop" "ignored_variables" ]
+    )
+    (mkRenamedOptionModule
+      [ "plugins" "nix-develop" "separatedVariables" ]
+      [ "plugins" "nix-develop" "separated_variables" ]
```
khaneliman, line 13: Everything looks pretty good overall, but I actually think we might not want to do renames for these options. They do match the lua table they are going to populate, but we are using them in extraOptions to do custom processing to insert lua config. Curious what others think about it, though...

HeitorAugustoLN: I personally think it’s better to keep the names the same as upstream. It makes it easier to figure out what’s being configured since everything matches up.

MattSturgeon: I can see a case to be made either way.

These are direct mappings to upstream options, so it makes sense to use the upstream name.

However these are also options declared _outside_ of any freeform option, so traditionally we would use a camelCase name.

Personally, my preference would be to avoid a rename if the option schema itself has not changed.

We can always look into either having a freeform option that these options can be moved into at a later date, and/or reconsider renaming them in another PR.

However that's not a strong opinion. I can still be persuaded otherwise.

GaetanLepage: Indeed, please revert to using `camelCase` please. Once this will be done, I guess that we will be able to merge this PR.

plugins/by-name/nix-develop/default.nix
```diff
@@ -1,61 +1,69 @@
...
-  cfg = config.plugins.nix-develop;
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts toLuaObject;
 in
-{
-  options.plugins.nix-develop = helpers.neovim-plugin.extraOptionsOptions // {
-    enable = mkEnableOption "nix-develop.nvim";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "nix-develop";
+  originalName = "nix-develop.nvim";
```
khaneliman, line 8:
```suggestion
  packPathName = "nix-develop.nvim";
```

plugins/by-name/nix-develop/deprecations.nix
HeitorAugustoLN: Fixed

plugins/by-name/nix-develop/default.nix
HeitorAugustoLN: Updated

khaneliman (approved):

khaneliman:
@mergify queue

PR #2654, opened 2024-12-13
{lightline,nvim-autopairs,treesitter}: extraConfigLua -> luaConfig

khaneliman:
Just replacing the occurrences of extraConfigLua in plugins that have been migrated to use the recommended approach instead of the deprecated approach.

### DIFF
--- a/plugins/by-name/lightline/default.nix
+++ b/plugins/by-name/lightline/default.nix
@@ -22,20 +22,20 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
        settings.component_function = {
          readonly = "LightlineReadonly";
        };
-     };
 
-     extraConfigLua = '''
-       function LightlineReadonly()
-         local is_readonly = vim.bo.readonly == 1
-         local filetype = vim.bo.filetype
+       luaConfig.pre= '''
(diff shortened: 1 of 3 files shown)


traxys (approved):

khaneliman:
@mergify queue

plugins/by-name/lightline/default.nix
```diff
@@ -22,20 +22,20 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
        settings.component_function = {
          readonly = "LightlineReadonly";
        };
-     };
 
-     extraConfigLua = '''
-       function LightlineReadonly()
-         local is_readonly = vim.bo.readonly == 1
-         local filetype = vim.bo.filetype
+       luaConfig.pre= '''
```
MattSturgeon, line 26: I'm sure the tests should have picked up if this was wrong, but shouldn't this be nested under the plugin's namespace?

E.g. `plugins.ligntline.luaConfig.pre`?

(Same for other two)

MattSturgeon: Oh, sorry. I missed the indentation change and the removed `}` above.

Issue #2381, opened 2024-10-06
[PLUGIN REQUEST] tiny-devicons-auto-colors.nvim


imoBooze:
| Field    | Description |
|----------|-------------|
| Plugin   | `tiny-devicons-auto-colors.nvim`  | 
| Homepage | [https://github.com/rachartier/tiny-devicons-auto-colors.nvim](https://github.com/rachartier/tiny-devicons-auto-colors.nvim)    | 
| Nixpkgs  | `false` | 

## Extra Information

This plugin allows `nvim-web-devicons` to use the colors set by `colorscheme`. An optional colour palette can be defined to allow for better colour matching.

Issue #2292, opened 2024-09-21
Implement DAG-based Plugin Ordering in init.lua


khaneliman:
### Context:

- Similar to Home Manager's dagOf option type, Nixvim currently lacks a built-in mechanism to explicitly define dependencies between plugins in init.lua.
-  This can lead to issues where plugins are loaded in an incorrect order, causing unexpected behavior or even errors.

### Problem:

- Without a clear way to specify plugin dependencies, init.lua ordering is fragile and can break easily when new plugins are added or existing ones are updated. We've had recent issues with certain plugins misbehaving after our by-name importing due to an automatic import mechanism instead of manual ordering. 

### Proposed Solution:

- Introduce a new configuration option, perhaps within the plugins table, to express plugin dependencies using a DAG-like structure. 
- This would allow users to specify which plugins a given plugin depends on, ensuring they are loaded in the correct order. We can set sane defaults for things we know are required dependency orders. 

### Benefits:

- A more robust and maintainable way to manage plugin dependencies in init.lua.
- Reduced likelihood of plugin conflicts and errors due to incorrect load order.
-  Improved user experience, especially for complex configurations with many plugins.

erahhal:
Is there a way to do this now in a more brute-force manner? I'm running into issues with the avante plugin.

PR #2655, opened 2024-12-13
plugins/tiny-devicons-auto-colors: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2381

~Attempted to see if it would work with mini mockDevicons and it still seemed to throw an error.~

Looked like a load order issue from dependencies, manually inserting setupCode in corresponding plugin's `luaConfig.post` until we can handle `mkOrder` or `dag` https://github.com/nix-community/nixvim/issues/2292 is available

### DIFF
--- a/plugins/by-name/mini/default.nix
+++ b/plugins/by-name/mini/default.nix
@@ -5,6 +5,8 @@
 lib.nixvim.neovim-plugin.mkNeovimPlugin {
   name = "mini";
   packPathName = "mini.nvim";
+  # Just want it before most other plugins for the icons provider.
+  configLocation = lib.mkOrder 800 "extraConfigLua";
 
   maintainers = [ lib.maintainers.khaneliman ];
 
--- a/plugins/by-name/tiny-devicons-auto-colors/default.nix
+++ b/plugins/by-name/tiny-devicons-auto-colors/default.nix
@@ -1,0 +1,109 @@
+{ config, lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "tiny-devicons-auto-colors";
+  packPathName = "tiny-devicons-auto-colors.nvim";
+  package = "tiny-devicons-auto-colors-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsOptions = {
+    colors =
+      defaultNullOpts.mkNullableWithRaw (with lib.types; either (attrsOf anything) (listOf str)) [ ]
+        ''
+          A table of color codes that the plugin will use to assign colors to devicons.
+          If not provided, the plugin will fetch highlights from the current theme to generate a color palette.
+        '';
+
+    factors = {
+      lightness = defaultNullOpts.mkNum 1.75 ''
+        Adjust the lightness factor.
+      '';
+      chroma = defaultNullOpts.mkNum 1 ''
+        Adjust the chroma factor.
+      '';
+      hue = defaultNullOpts.mkNum 1.25 ''
+        Adjust the hue factor.
+      '';
+    };
+
+    cache = {
+      enabled = defaultNullOpts.mkBool true ''
+        Enable or disable caching to improve performance.
+      '';
+      path = defaultNullOpts.mkStr (lib.nixvim.literalLua ''vim.fn.stdpath("cache") .. "/tiny-devicons-auto-colors-cache.json"'') ''
+        Path to the cache file.
+      '';
+    };
+
+    precise_search = {
+      enabled = defaultNullOpts.mkBool true ''
+        Enable or disable precise search for better color matching.
+      '';
+      iteration = defaultNullOpts.mkNum 10 ''
+        Number of iterations for precise search.
+      '';
+      precision = defaultNullOpts.mkNum 20 ''
+        Precision level for the search.
+      '';
+      threshold = defaultNullOpts.mkNum 23 ''
+        Threshold to consider a color as a match.
+      '';
+    };
+
+    ignore = defaultNullOpts.mkListOf lib.types.str [ ] ''
+      A list of icon names to ignore.
+    '';
+
+    autoreload = defaultNullOpts.mkBool false ''
+      Automatically reload colors when the colorscheme changes.
+    '';
+  };
+
+  settingsExample = {
+    colors = {
+      red = "#ff0000";
+      green = "#00ff00";
+    };
+    factors = {
+      lightness = 1.5;
+      chroma = 1.2;
+      hue = 1.1;
+    };
+    cache = {
+      enabled = true;
+      path = "/path/to/cache.json";
+    };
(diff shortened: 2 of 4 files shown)


khaneliman:
I think it may be a load order thing. Gonna try something out.

plugins/by-name/tiny-devicons-auto-colors/default.nix
```diff
@@ -0,0 +1,124 @@
...
+      {
+        assertion =
+          config.plugins.web-devicons.enable
+          || (
+            config.plugins.mini.enable
+            && config.plugins.mini.modules ? icons
+            && config.plugins.mini.mockDevIcons
+          );
+        message = ''Nixvim: The plugin `web-devicons` must be enabled to use `tiny-devicons-auto-colors`'';
+      }
```
MattSturgeon, line 105: I'm imagining having a central "icon provider" module that's responsible for a single assertion and an option where providers can register themselves.

Maybe this would go in top-level `plugins/icons.nix`, or maybe `plugins/by-name/webdevicons/alternatives.nix` ?

Probably doesn't fit within this PR, I'm guessing best as a post-merge refactor?

plugins/by-name/tiny-devicons-auto-colors/default.nix
```diff
@@ -0,0 +1,124 @@
...
+    ];
+    plugins =
+      let
+        setupCode = ''
+          require('tiny-devicons-auto-colors').setup(${
+            lib.optionalString (cfg ? settings) (lib.nixvim.toLuaObject cfg.settings)
+          })
+        '';
```
MattSturgeon, line 113: No need for `optional (cfg?settings)` here, that's used in the generic helper utils because there may not be a `settings` option declared. If an option exists it will _always_ be in `config`.

plugins/by-name/tiny-devicons-auto-colors/default.nix
```diff
@@ -0,0 +1,124 @@
...
+      in
+      {
+        mini.luaConfig.post = lib.mkIf (
+          config.plugins.mini.enable
+          && config.plugins.mini.modules ? icons
+          && config.plugins.mini.mockDevIcons
+        ) setupCode;
+        web-devicons.luaConfig.post = lib.mkIf config.plugins.web-devicons.enable setupCode;
```
MattSturgeon, line 121: Q: are you happy with this or were you planning to explore a `mkOrder` based solution?

plugins/by-name/tiny-devicons-auto-colors/default.nix
khaneliman: True, that makes sense.... I was just lazy and copy pasted this over. :P

plugins/by-name/tiny-devicons-auto-colors/default.nix
khaneliman: I do like that actually, I had thought about something similar recently with the introduction of different `cmp` plugins for the same reason.

plugins/by-name/tiny-devicons-auto-colors/default.nix
khaneliman: I can switch to using that I think, essentially we would set the priority for the icons providers to the same level and set this below that priority. Does that sound correct?

MattSturgeon: Yeah, sounds correct. Obviously, higher number is lower priority.

My only concern it that by setting a priority for these three plugins, they'll _all_ end up in a different position from where they are currently.

Currently they're merged pseudorandomly within the "default" merge order prio.

khaneliman: The only thing that gets a little messy, right now, is delegating the mini priority in the current form of the module. The `mini.icons` stuff gets bundled in the same luaConfig as the rest of their modules. Should we isolate that config and move it up or just move all the mini stuff to a higher order?

MattSturgeon: My initial instinct is that it'd be nice to still keep the mini stuff together. I just hope it doesn't cause issues.

This will be no worse than we we migrated everything to `plugins/by-name`, effectively shuffling the merge order.

plugins/by-name/mini/default.nix
```diff
@@ -5,6 +5,8 @@
 lib.nixvim.neovim-plugin.mkNeovimPlugin {
   name = "mini";
   packPathName = "mini.nvim";
+  # Just want it before most other plugins for the icons provider.
+  configLocation = lib.mkBefore "extraConfigLua";
```
MattSturgeon, line 9: I'm minded to suggest we avoid using default prio numbers internally, that way end users can reliably go "before" us if we consistently use a prio after `(mkBefore null).priority`, for example

IIRC, before is 500, default is 1000, and after is 1500? So perhaps we should aim to be in the 7xx & 8xx range?

plugins/by-name/tiny-devicons-auto-colors/default.nix
```diff
@@ -0,0 +1,107 @@
...
+    assertions = [
+      {
+        assertion =
+          config.plugins.web-devicons.enable
+          || (
+            config.plugins.mini.enable
+            && config.plugins.mini.modules ? icons
+            && config.plugins.mini.mockDevIcons
+          );
+        message = ''Nixvim: The plugin `web-devicons` must be enabled to use `tiny-devicons-auto-colors`'';
```
MattSturgeon, line 103:
```suggestion
        message = ''
          Nixvim: Either `plugins.web-devicons` or `plugins.mini`* must be enabled to use `tiny-devicons-auto-colors`.
            *If using `plugins.mini`, you must enable the `icons` module and the `mockDevIcons` option.'';
```

MattSturgeon (approved):
I don't see any issues.

Left a suggestion for being less prescriptive in the assertion message. Do with it what you will :grin: 

Maybe let's give other maintainers chance to look over this before merging, especially regarding the re-ordered `configLocation`?

plugins/by-name/tiny-devicons-auto-colors/default.nix
traxys: I think we could have a similar thing for lazy loading

GaetanLepage (approved):
LGTM

plugins/by-name/tiny-devicons-auto-colors/default.nix
```diff
@@ -0,0 +1,110 @@
...
+      threshold = 20;
+    };
+    ignore = [
+      "lua"
+      "vim"
+    ];
+    autoreload = true;
+  };
+
+  # We want to make sure its after either of the dev icons providers
```
MattSturgeon, line 92:
```suggestion
```

This comment is leftover from before the icon providers were re-ordered?

traxys (approved):

khaneliman:
@mergify queue

khaneliman:
@mergify dequeue

khaneliman:
@mergify queue

PR #2656, opened 2024-12-13
plugins/luasnip: migrate to mkNeovimPlugin

HeitorAugustoLN:


plugins/by-name/luasnip/default.nix
```diff
@@ -104,31 +309,30 @@ in
           require("luasnip.loaders.from_vscode").lazy_load({['paths'] = {'/nix/store/.../path/to/snippets'}})
         ```
       '';
-      type = types.listOf loaderSubmodule;
     };
 
-    fromSnipmate = mkOption {
+    from_snipmate = mkOption {
```
khaneliman, line 314: Similar to https://github.com/nix-community/nixvim/pull/2653#discussion_r1885069208 we want extraOptions to be `camelCase`

khaneliman (approved):

plugins/by-name/luasnip/default.nix
```diff
@@ -39,54 +33,265 @@ let
...
+    '';
+
+    update_events =
+      defaultNullOpts.mkNullableWithRaw (with types; either str (listOf str)) "InsertLeave"
+        ''
+          Choose which events trigger an update of the active nodes' dependents.
+        '';
+
+    region_check_events =
+      defaultNullOpts.mkNullableWithRaw (with types; either str (listOf str)) { __raw = "nil"; }
```
MattSturgeon, line 79:
```suggestion
      defaultNullOpts.mkNullableWithRaw (with types; either str (listOf str)) (literalLua "nil")
```

Going forward, we should aim to use `lib.nixvim.litearalLua` instead of `__raw` in examples & defaultText.

plugins/by-name/luasnip/default.nix
```diff
@@ -161,49 +365,41 @@ in
...
-                options = attrsets.getAttrs [
+                options = lib.getAttrs [
                   "paths"
                   "exclude"
                   "include"
                 ] loader;
               in
               ''
-                require("luasnip.loaders.from_${name}").${optionalString loader.lazyLoad "lazy_"}load(${lib.nixvim.toLuaObject options})
+                require("luasnip.loaders.from_${name}").${lib.optionalString loader.lazyLoad "lazy_"}load(${toLuaObject options})
```
MattSturgeon, line 392: IDK how important this is, but it feels like it may be safer to escape `name` by passing it through `toJSON` or "toLuaObject`?

```nix
/*lua*/ ''
  loaders["from_" .. ${toLuaObject name}]
  -- or
  loaders[${toLuaObject "from_${name}"}]
  -- or
  loaders[${toLuaObject ("from_" + name)}]
''
```

plugins/by-name/luasnip/default.nix
HeitorAugustoLN: Should be fixed now

plugins/by-name/luasnip/default.nix
HeitorAugustoLN: I don't think this is really needed, imo.

MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2660, opened 2024-12-14
lib/neovim-plugin: allow `configLocation` to be wrapped using `mkOrder`

MattSturgeon:
https://github.com/nix-community/nixvim/pull/2648#discussion_r1883632261

For example, a plugin can choose a specific order-prio within `extraConfigLuaPre`:

```nix
mkNeovimPlugin {
  name = "foo";
  # ...

  configLocation = lib.mkOrder 69 "extraConfigLuaPre";
}
```


khaneliman:
Would be nice to have some examples of how we could use this since it sounds like it could be useful with https://github.com/nix-community/nixvim/pull/2655 and https://github.com/nix-community/nixvim/issues/2292

MattSturgeon:
> Would be nice to have some examples of how we could use this

If you know of any existing plugins that'd benefit from this, I'm happy to come up with something. If it's trivial it can be part of this PR, otherwise a follow PR.

If we can't think of anything, maybe this isn't as useful as it looks? :joy:

khaneliman:
> > Would be nice to have some examples of how we could use this
> 
> If you know of any existing plugins that'd benefit from this, I'm happy to come up with something. If it's trivial it can be part of this PR, otherwise a follow PR.
> 
> If we can't think of anything, maybe this isn't as useful as it looks? 😂

I guess I was curious about implementation details that could be used to control being before / after a specific plugin.

MattSturgeon:
> I guess I was curious about implementation details

Added a small example to the PR desc.

`mkOrder` is very trivial, so if you wanted to explicitly order two plugins you'd _likely_ want to define order-prios for both of them manually :thinking:

> If this is exposed to users, it could be useful in arranging your generated lua.

I don't think it's possible to have `configLocation` itself be a module option without running into infinite recursion, since its value defines part of the _shape_ of the toplevel `config`.

I think @traxys originally wanted a `luaConfig.loc`, but ran into the above issue.

In theory we could expose the actual prio as a module option, but the API would probably be kinda ugly :thinking: ... Maybe something like `luaConfig.destOrder`?

khaneliman:
> > I guess I was curious about implementation details
> 
> Added a small example to the PR desc.
> 
> `mkOrder` is very trivial, so if you wanted to explicitly order two plugins you'd _likely_ want to define order-prios for both of them manually 🤔
> 

Ah just saw mention of `loc` and thought it was going to allow name based. This is still an improvement imo Because we can control the order for specific plugins we know need to be in a certain order through their mkOrder

MattSturgeon:
> Ah just saw mention of `loc` and thought it was going to allow name based.

Yes, loc is name/path based. But that's what we have already. And that's the bit that likely can't be a configurable module option without inf-rec.

khaneliman (approved):
Might benefit from a small section in our contributing docs? But LGTM

CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ A template plugin can be found in (plugins/TEMPLATE.nix)[https://github.com/nix-
...
 | **colorscheme**              | The name of the colorscheme.                                                                                                                                                                                                                         | No       | `name` parameter                                                                                |
-| **configLocation**           | The location for the Lua configuration.                                                                                                                                                                                                              | No       | `"extraConfigLuaPre"` if `isColorscheme` then `extraConfigLuaPre`, otherwise `"extraConfigLua"` |
+| **configLocation**           | The option location where the lua configuration should be installed. Nested option locations can be represented as a list. The location can also be wrapped using `lib.mkBefore`, `lib.mkAfter`, or `lib.mkOrder`.                                   | No       | `"extraConfigLuaPre"` if `isColorscheme` then `extraConfigLuaPre`, otherwise `"extraConfigLua"` |
```
MattSturgeon, line 65: @khaneliman added this doc

khaneliman: Thanks, appreciate it!

MattSturgeon:
@mergify queue

PR #2663, opened 2024-12-14
flake-modules/dev: add deadnix

khaneliman:
Adding https://github.com/astro/deadnix through our git-hooks.nix pre-commit hooks.
https://github.com/cachix/git-hooks.nix/blob/c2b3567b03baf0999a1dd14f7e7ab34b46297d68/modules/hooks.nix#L283-L338

### DIFF
--- a/flake-modules/dev/default.nix
+++ b/flake-modules/dev/default.nix
@@ -66,6 +66,15 @@
         check.enable = false;
 
         settings.hooks = {
+          deadnix = {
+            enable = true;
+
+            settings = {
+              noLambdaArg = true;
+              noLambdaPatternNames = true;
+              edit = true;
+            };
+          };
           treefmt.enable = true;
           typos = {
             enable = true;
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -168,9 +168,6 @@ in
       let
         pluginToLua =
           plugin:
-          let
-            keyExists = keyToCheck: attrSet: lib.elem keyToCheck (lib.attrNames attrSet);
-          in
           if isDerivation plugin then
             { dir = "${lazyPath}/${lib.getName plugin}"; }
           else
--- a/wrappers/nixos.nix
+++ b/wrappers/nixos.nix
@@ -4,16 +4,10 @@ self:
   config,
   lib,
   ...
-}@args:
+}:
 let
   inherit (lib)
-    mkEnableOption
-    mkOption
-    mkOptionType
-    mkForce
-    mkMerge
     mkIf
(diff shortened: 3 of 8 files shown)


GaetanLepage (approved):
Cleaning is always nice.

wrappers/nixos.nix
```diff
@@ -4,16 +4,10 @@ self:
   config,
   lib,
   ...
-}@args:
```
khaneliman, line 7: I'm not as familiar with the language enough to know if this will have unexpected results @MattSturgeon

MattSturgeon: I've been meaning to remove that every time I open the file with nixd running... The cleanup was never related to what I was working on though.

`args` is literally just a param/variable being defined here. If it's unused, its safe to delete. No hidden black magic.

plugins/pluginmanagers/lazy.nix
```diff
@@ -168,9 +168,6 @@ in
       let
         pluginToLua =
           plugin:
-          let
-            keyExists = keyToCheck: attrSet: lib.elem keyToCheck (lib.attrNames attrSet);
-          in
```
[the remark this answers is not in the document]
khaneliman: I'm not sure what it was for, in my lazy branch I think I removed it or just ignored it because I wanted to keep diff small.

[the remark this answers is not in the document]
khaneliman: That file is getting rewritten anyways in the mkNeovimPlugin PR.

flake-modules/dev/default.nix
```diff
@@ -66,6 +66,15 @@
...
         settings.hooks = {
+          deadnix = {
+            enable = true;
+
+            settings = {
+              noLambdaArg = true;
+              noLambdaPatternNames = true;
+              edit = true;
+            };
+          };
```
MattSturgeon, line 77: I'm on the fence whether this is good as a pre-commit hook. This could be frustrating while working on unfinished PRs.

My instinct is that this may be better as a separate flake check and/or a _github_ workflow.

The benefit of a github workflow over buildbot here is that it can provide more integrated feedback inline within PRs and be more easily distinguished from other failures.

That said, I'm fine to merge this and later refactor/revert if needed.

MattSturgeon (approved):

flake-modules/dev/default.nix
khaneliman: That's fair, I think with these flags it shouldn't be too disruptive. But, if we all start to find it annoying we can pivot pretty easily, at least.

khaneliman:
@mergify queue

PR #2665, opened 2024-12-14
docs/config-examples: generate dynamically from a toml list

MattSturgeon:
This should help to maintain uniformity and hopefully make contributions easier.

I would like to add tests for robustness and to avoid regressions. Any ideas on what to test and/or how would be great.

If there's any toml validation that needs to happen, maybe we could have a test for that too? This could then be included in the pre-commit hooks.

EDIT: I added some regression tests that ensure it is able to handle a few edge-cases.

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -1,5 +1,6 @@
 {
   pkgs,
+  callPackage,
   runCommand,
   lib,
   evaledModules,
@@ -349,6 +350,10 @@ pkgs.stdenv.mkDerivation (finalAttrs: {
     substituteInPlace ./index.md \
       --replace-fail "@DOCS_VERSIONS@" "$(cat ${finalAttrs.passthru.docs-versions})"
 
+    # Patch user-configs
+    substituteInPlace ./user-guide/config-examples.md \
+      --replace-fail "@USER_CONFIGS@" "$(cat ${finalAttrs.passthru.user-configs})"
+
     mdbook build
     cp -r ./book/* $dest
     mkdir -p $dest/search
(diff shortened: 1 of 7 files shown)


user-configs/list.toml
MattSturgeon, line 1: `/user-configs/list.toml` is meant to be easy for users to discover, but maybe putting it in `/docs` is neater?

GaetanLepage: Yes, I think it makes more sense in `/docs` and would pollute less the root of the tree.

GaetanLepage (dismissed):
Very good idea Matt !

MattSturgeon:
> Does nixpkgs have a two derivations and/or files should produce the same output test helper?

Looks like the answer is [`testEqualContents`](https://nixos.org/manual/nixpkgs/unstable/#tester-testEqualContents)

docs/mdbook/default.nix
```diff
@@ -398,5 +403,6 @@ pkgs.stdenv.mkDerivation (finalAttrs: {
             echo "- The $link, for use with nixpkgs \`$nixpkgs\`$suffix" >> "$out"
           done
         '';
+    user-configs = callPackage ../user-configs { };
```
MattSturgeon, line 406: Hm... Not sure how I feel about naming here.

The page in the docs is "config-examples.md" while the package that's generating them is "user-configs".

GaetanLepage: No hard feeling, both are fine.

khaneliman (approved):
Love it, thanks :)

docs/mdbook/default.nix
MattSturgeon: Ok, I'll leave it as-is then. Having a different name for the actual page in the docs and the packaged used to construct the main part of that page isn't the end of the world.

MattSturgeon:
@mergify queue

PR #2666, opened 2024-12-14
plugins/lsp: modernize implem of language-servers

GaetanLepage:


plugins/lsp/language-servers/ltex-settings.nix
```diff
@@ -212,52 +215,50 @@ with lib;
...
+    This setting is a multi-scope setting. See the documentation for details.
+    This setting supports external files. See the documentation for details.
+    If this list is very large, performance may suffer.
+
+    Example:
+    ```nix
+      {
+        "en-US" = [ ":/path/to/externalFile.txt" ];
+      }
+    ```
```
MattSturgeon, line 258: Can we promote this to be a "real" example?

MattSturgeon (commented):
Looked through and found no issues.

I spotted one manually added markdown example that would be good to migrate to a nixos option example while we're doing this cleanup work.

plugins/lsp/language-servers/ltex-settings.nix
GaetanLepage: Good idea ! Done.

MattSturgeon (approved):
:broom:

GaetanLepage:
@mergifyio queue

PR #2667, opened 2024-12-14
list-plugins: improve script

GaetanLepage:
- **dev/list-plugins: fix spacing in plugin printing (deprecation warning)**
- **dev/list-plugins: add symbol for 'MISC' type plugins**
- **dev/list-plugins: update EXCLUDES and KNOWN_PATHS variables**

### DIFF
--- a/flake-modules/dev/list-plugins.py
+++ b/flake-modules/dev/list-plugins.py
@@ -160,6 +160,9 @@ class Plugin:
             f"| {kind_icon}\033[0m  | {state_icon}  | {deprecation_icon} | {self.path}"
         )
 
+    def print_markdown(self) -> None:
+        print(f"- [ ] {self.path} ({self.kind.name.lower()})")
+
 
 def has_deprecation_warnings(string: str) -> bool:
     for regex in DEPRECATION_REGEX:
@@ -218,13 +221,17 @@ def _is_excluded(path: str) -> bool:
     return True
 
 
-def main(args):
+def main(args) -> None:
     paths: list[str] = glob.glob(pathname="plugins/**/*.nix", recursive=True)
     filtered_paths: list[str] = list(filter(_is_excluded, paths))
     filtered_paths.sort()
 
-    print("| Typ | Sty | DW | path")
-    print("|-----|-----|----|--------------------------------------------------------")
+    if not args.markdown:
+        print("| Typ | Sty | DW | path")
+        print(
+            "|-----|-----|----|--------------------------------------------------------"
+        )
+
     for plugin_path in filtered_paths:
         plugin: Optional[Plugin] = parse_file(path=plugin_path)
         if plugin is not None:
@@ -233,7 +240,10 @@ def main(args):
                 and (args.state is None or plugin.state.name.lower() == args.state)
                 and (not args.deprecation_warnings or plugin.dep_warnings)
             ):
-                print(plugin)
+                if args.markdown:
+                    plugin.print_markdown()
+                else:
+                    print(plugin)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


khaneliman:
Ooh nice, gonna test real quick. I've been using this for the tracking issue and just general usage for a while.

khaneliman:
Does look like unknown is shifted atm 
image
Don't know if that's like that on main

flake-modules/dev/list-plugins.py
```diff
@@ -11,10 +11,38 @@
 
 EXCLUDES: list[str] = [
     # Not plugin files
```
MattSturgeon, line 13: Maybe in the future we can detect plugins files? Or maybe that's non-trivial...

We could use the same strategy as the "by-name" and "package option" tests, i.e. scan the toplevel option set recursively, collect all options, then filter for `enable` options.

The option itself contains info on where it was declared, so we get the filename that way.

Obviously beyond this PR, and would need some nix code to complement the python.

flake-modules/dev/list-plugins.py
```diff
@@ -31,10 +59,71 @@ class State(Enum):
...
+        bool,  # Has deprecation warnings
+    ],
+] = {
+    "plugins/by-name/chadtree/default.nix": (State.OLD, Kind.NEOVIM, False),
+    "plugins/by-name/commentary/default.nix": (State.OLD, Kind.VIM, False),
+    "plugins/by-name/conjure/default.nix": (
+        State.OLD,
+        Kind.VIM,  # configured with weird globals `vim.g["conjure#mapping#doc_word"] = "gk"`
+        False,
+    ),
```
MattSturgeon, line 76: Doesn't having this hard-coded defeat the purpose of this test? Can you try and explain why we're doing things this way?

khaneliman:
Latest commit seems to have gotten rid of those question marks, now the only weird output I see 
image

flake-modules/dev/list-plugins.py
GaetanLepage: Well, for me, having a plugin listed as "unknown" is not helping at all.
So we have two choices:
1. improve the matching logic to automatically identify more plugins
2. Add exceptions

But anyway, I pushed a commit that completely get rid of `kind.UNKNOWN`.

GaetanLepage:
> Latest commit seems to have gotten rid of those question marks, now the only weird output I see

I have fixed it so that it looks good in my terminal. But the issue might be on my end, so I will revert this.

khaneliman:
> > Latest commit seems to have gotten rid of those question marks, now the only weird output I see
> 
> I have fixed it so that it looks good in my terminal. But the issue might be on my end, so I will revert this.

Oh, maybe we can't handle all different terminal/font/etc combinations.

khaneliman (approved):
LGTM, thanks for updated functionality!

The amount of manual workarounds indicates we might benefit from some more meta attributes in our modules that could help identify plugins better.

GaetanLepage:
> The amount of manual workarounds indicates we might benefit from some more meta attributes in our modules that could help identify plugins better.

Honestly, I hope for a bug push to convert all plugins and deprecate this script haha !

GaetanLepage:
@mergifyio queue

khaneliman:
> > The amount of manual workarounds indicates we might benefit from some more meta attributes in our modules that could help identify plugins better.
> 
> Honestly, I hope for a bug push to convert all plugins and deprecate this script haha !

I could still see a usage for it for querying our plugins easily and allow searching / filtering them by neovim/vim plugin/colorscheme etc

GaetanLepage:
> I could still see a usage for it for querying our plugins easily and allow searching / filtering them by neovim/vim plugin/colorscheme etc

For sure.

PR #2668, opened 2024-12-15
lib: remove dependency on `pkgs`

MattSturgeon:
We had a few things in `lib.nixvim.builders` that depend on `pkgs` and are therefore also per-system.

These were deprecated in #2231, with a view to fully remove after 24.11 was released.

With these fully removed, the lib now has zero dependency on a `pkgs` instance or a `system`.

Fixes #2186


GaetanLepage (approved):
LGTM

traxys (approved):

khaneliman (approved):

MattSturgeon:
@mergify queue

PR #2670, opened 2024-12-15
lib/modules: pass `inputs.nixpkgs` into `evalNixvim`

MattSturgeon:
Expose our locked nixpkgs as the `nixpkgs.source` module options.

This only happens when `evalNixvim` is part of a lib that was provided `flake` as an argument.

Stubbed the `nixpkgs.source` option for now. Eventually, this will be used to construct `pkgs` internally. For now, it's purely informational.

This is another small step towards #1784. See also #2022

### DIFF
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -1,11 +1,15 @@
 {
+  self,
   config,
   lib,
   withSystem,
   ...
 }:
 {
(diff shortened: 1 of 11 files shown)


GaetanLepage (approved):
LGTM

MattSturgeon:
@mergify queue

PR #2676, opened 2024-12-15
plugins/friendly-snippets: switch to mkVimPlugin

GaetanLepage:


plugins/by-name/friendly-snippets/default.nix
```diff
@@ -1,31 +1,13 @@
...
-    enable = mkEnableOption "friendly-snippets";
-
-    package = lib.mkPackageOption pkgs "friendly-snippets" {
-      default = [
-        "vimPlugins"
-        "friendly-snippets"
-      ];
-    };
-  };
+lib.nixvim.vim-plugin.mkVimPlugin {
```
khaneliman, line 5: This is a lua plugin, though https://github.com/rafamadriz/friendly-snippets

GaetanLepage: Yes, but as it is not configured using lua code, we don't consider it as a "neovim" plugin.
Plugins that have no configuration are implemented using `mkVimPlugin`.

Maybe long term we can find a better abstraction. In the short-term we'll stick with this methodology.

khaneliman: We've been using `hasSettings = false` and `callSetup = false`

GaetanLepage: Sure, but this makes not much more sense to me...
The `mkNeovimPlugin` name is a bit misleading, it should be `mkPluginThatIsConfiguredWithRequireLuaOrSomethingSimilar`.

I am open to discussing how to rethink those helpers and their respective field.
However, using `mkVimPlugin` here is fine in my opinion.

khaneliman:
Want to cleanup the entry in https://github.com/nix-community/nixvim/pull/2681

GaetanLepage:
> Want to cleanup the entry in #2681

Good point, done

plugins/by-name/friendly-snippets/default.nix
MattSturgeon: > We've been using `hasLuaConfig = false`, `hasSettings = false`, and `callSetup = false`

The fact we can now use either `mkNeovimPlugin` or `mkVimPlugin` makes me wonder about combining them together into on single function?

Or alternatively, maybe a lot of the modes available via args should actually be special-case wrapper functions that wrap a single lower-level `mkPluginWith`?

MattSturgeon (approved):

GaetanLepage:
@mergifyio queue

PR #2678, opened 2024-12-15
plugins/lsp: Add some LSP servers

traxys:


khaneliman (approved):
Nice :) Wonder if we can communicate these in announcements somehow for people who might be interested

traxys:
It may be a bit spammy :/

traxys:
@mergify queue

MattSturgeon:
> Nice :) Wonder if we can communicate these in announcements somehow for people who might be interested



> It may be a bit spammy :/

I think a single post with a brief summary is ok

PR #2679, opened 2024-12-15
tests: use our extended lib in tests

GaetanLepage:


khaneliman:
Looks like it breaks the maintainer list test. Must be doing some lib comparison between our extended lib and upstream lib maintainers list.

tests/default.nix
```diff
@@ -11,11 +11,11 @@ let
   autoArgs = pkgs // {
     inherit
       helpers
-      lib
       pkgsUnfree
       self
       system
       ;
+    lib = helpers.extendedLib;
```
MattSturgeon, line 18:
```suggestion
    lib = helpers.extendedLib;
    prevLib = lib;
```

? This can then be used in the maintainers test to access `prevLib.maintainers`.


Alt names:
- `prevLib`
- `oldLib`
- `nixpkgsLib`
- `upstreamLib`

khaneliman: I was just wondering if we should do something like that just to store it for usage when we need to use it.

khaneliman (approved):

PR #2682, opened 2024-12-15
plugins/quickmath: migrate to mkVimPlugin

GaetanLepage:


plugins/by-name/quickmath/default.nix
```diff
@@ -1,42 +1,33 @@
...
-    extraPlugins = [ cfg.package ];
-
+  extraConfig = cfg: {
     keymaps =
       with cfg.keymap;
-      optional (key != null) {
+      lib.optional (key != null) {
         mode = "n";
         inherit key;
         action = ":Quickmath<CR>";
```
MattSturgeon, line 33: Should probably be, as `<cmd>` covers a few more edge cases than `:`. Though not too important here since we're enforcing normal-mode mappings.
```suggestion
        action = "<CMD>Quickmath<CR>";
```

MattSturgeon (approved):

plugins/by-name/quickmath/default.nix
GaetanLepage: Good to know, done.

GaetanLepage:
@mergifyio queue

PR #2683, opened 2024-12-16
lib/mkSettingsRenamedOptionModules: allow attrs value with 'old' and 'new' keys

GaetanLepage:
The `optionRenamedToSettings` attribute allows to automatically create `mkRenamedOptionModule`s for simple renames (options that move from plugin top-level to `settings`).
This helper is implemented using the `lib.nixvim.mkSettingsRenamedOptionModules` function.

This PR improves this function by making it accept `attrs` elements in the form
```nix
{
  old = ["one" "fooBar"];
  new = "one_foo_bar";
}
```
This makes it much more flexible and allows us to get rid of several calls to the native `lib.mkRenamedOptionModule` in the plugin modules.

- **lib/mkSettingsRenamedOptionModules: allow attrs value with 'old' and 'new' keys**
  -> Makes the change to `mkSettingsRenamedOptionModules`
- **plugins/*: use new mkSettingsRenamedOptionModules**
  -> Takes advantage of this implementation tree-wide

### DIFF
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -47,16 +47,45 @@ rec {
         '';
     };
 
+  /*
+    Returns a function that maps
+      [
+        "someOption"
+        ["fooBar" "someSubOption"]
+        { old = "someOtherOption"; new = ["foo_bar" "some_other_option"]}
+      ]
+
+    to
+      [
+        (lib.mkRenamedOptionModule
+          (oldPath ++ ["someOption"])
+          (newPath ++ ["some_option"])
+        )
+        (lib.mkRenamedOptionModule
+          (oldPath ++ ["fooBar" "someSubOption"])
+          (newPath ++ ["foo_bar" "some_sub_option"])
+        )
+        (lib.mkRenamedOptionModule
+          (oldPath ++ ["someOtherOption"])
+          (newPath ++ ["foo_bar" "some_other_option"])
+        )
+      ]
+  */
   mkSettingsRenamedOptionModules =
     oldPrefix: newPrefix:
     map (
-      option':
+      spec:
       let
-        option = lib.toList option';
-        oldPath = oldPrefix ++ option;
-        newPath = newPrefix ++ map lib.nixvim.toSnakeCase option;
+        finalSpec =
+          if lib.isAttrs spec then
+            lib.mapAttrs (_: lib.toList) spec
+          else
+            {
+              old = lib.toList spec;
+              new = map lib.nixvim.toSnakeCase finalSpec.old;
+            };
       in
-      lib.mkRenamedOptionModule oldPath newPath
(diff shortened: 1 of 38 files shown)


lib/deprecation.nix
```diff
@@ -47,14 +47,43 @@ rec {
...
       option':
+      # If the value is an attrs, it must have both `old` and `new` attributes.
+      assert lib.isAttrs option' -> ((option' ? old) && (option' ? new));
       let
-        option = lib.toList option';
-        oldPath = oldPrefix ++ option;
-        newPath = newPrefix ++ map lib.nixvim.toSnakeCase option;
+        oldSubPath = lib.toList (option'.old or option');
+        newSubPath =
+          if lib.isAttrs option' then lib.toList option'.new else map lib.nixvim.toSnakeCase oldSubPath;
```
MattSturgeon, line 83: How about flipping the logic a little by coercing non-attrs to a normalised form?

```nix
option'' =
  if isAttrs option' then
    mapAttrs (_: toList) option'
  else
    {
      old = toList option';
      new = map lib.nixvim.toSnakeCase option''.old;
    };
```

Maybe better names than `option'` and `option''` :grin:

lib/deprecation.nix
```diff
@@ -47,14 +47,43 @@ rec {
...
+
+        oldPath = oldPrefix ++ oldSubPath;
+        newPath = newPrefix ++ newSubPath;
```
MattSturgeon, line 86: Maybe just inline these?

MattSturgeon (commented):
Like this!

lib/deprecation.nix
MattSturgeon: Maybe rename `option'` to `rename`, `spec`, or something?
And `option''` to `normalized`, `final` or something

plugins/by-name/barbar/default.nix
```diff
@@ -125,14 +125,35 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
+    }
+    {
+      old = [
+
+        "icons"
+        "filetype"
+        "enable"
+      ];
+      new = [
+
```
MattSturgeon, line 144:
```suggestion
```

plugins/by-name/rest/default.nix
```diff
@@ -146,49 +40,20 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
...
-        )
-        (
-          settingsPath
-          ++ [
-            "highlight"
-            "enable"
-          ]
-        )
+    ]
+    ++ (map
```
MattSturgeon, line 44:
```suggestion
    ++ map
```
No need to wrap concatenation segments with `()` usually

MattSturgeon (approved):
This is much better! Great idea, thanks for doing this! :broom: :fire:

GaetanLepage:
@mergifyio queue

PR #2685, opened 2024-12-16
dev/list-plugins: various improvements

GaetanLepage:
- **dev/list-plugins: properly error out when parsing is unsuccessfull**
- **dev/list-plugins: make it a proper package**
- **dev/list-plugins: add --root-path argument**
- **dev/list-plugins: add flake check**

### DIFF
--- a/flake-modules/dev/list-plugins/default.nix
+++ b/flake-modules/dev/list-plugins/default.nix
@@ -1,6 +1,8 @@
+{ self, ... }:
 {
   perSystem =
     {
+      self',
       config,
       lib,
       inputs',
@@ -8,8 +10,8 @@
       pkgs,
       ...
     }:
-    {
-      packages.list-plugins = pkgs.writers.writePython3Bin "list-plugins" {
+    let
+      package = pkgs.writers.writePython3Bin "list-plugins" {
         # Disable flake8 checks that are incompatible with the ruff ones
         flakeIgnore = [
           # line too long
@@ -18,6 +20,17 @@
           "W503"
         ];
       } (builtins.readFile ./list-plugins.py);
+    in
+    {
+      packages.list-plugins = package;
+      checks.list-plugins-test =
+        pkgs.runCommand "list-plugins-test"
+          {
+            nativeBuildInputs = [ package ];
+          }
+          ''
+            list-plugins --root-path ${self} > $out
+          '';
 
       devshells.default.commands = [
         {
--- a/flake-modules/dev/list-plugins/default.nix
+++ b/flake-modules/dev/list-plugins/default.nix
@@ -1,0 +1,30 @@
+{
+  perSystem =
+    {
+      config,
+      lib,
+      inputs',
+      system,
+      pkgs,
+      ...
+    }:
+    {
+      packages.list-plugins = pkgs.writers.writePython3Bin "list-plugins" {
+        # Disable flake8 checks that are incompatible with the ruff ones
+        flakeIgnore = [
+          # line too long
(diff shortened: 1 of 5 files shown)


flake-modules/dev/list-plugins/list-plugins.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
```
MattSturgeon, line 2: I believe this can stay. Sure, `writePython3Bin` will add its own, but shell interpreters will only use the first and others will be treated as "normal" comments

flake-modules/dev/list-plugins/list-plugins.py
```diff
@@ -252,6 +256,12 @@ def main(args) -> None:
     """,
         formatter_class=RawTextHelpFormatter,
     )
+    parser.add_argument(
+        "--root-path",
+        type=str,
+        default="./",
```
MattSturgeon, line 262: Maybe as a TODO comment:

This could default to looking recursively for the flake.lock like some of the other scripts do?

MattSturgeon: Maybe we want to have some lib/util for common logic used in most/all our scripts?

flake-modules/dev/list-plugins/default.nix
```diff
@@ -19,6 +21,17 @@
...
+      checks.list-plugins-test =
+        pkgs.runCommand "list-plugins-test"
+          {
+            nativeBuildInputs = [
+              self'.packages.list-plugins
+            ];
+          }
+          ''
+            list-plugins --root-path ${toString self} > $out
+          '';
```
MattSturgeon, line 33: This feels more like a test _of_ the package itself rather than a general test of our repo?

Therefore I'd _prefer_ to see this defined locally to the package, i.e. as `list-plugins.passthru.test`

We could still add it to the flake checks; in the same way we do for the `docs.user-configs.tests` tests.

flake-modules/dev/list-plugins/list-plugins.py
GaetanLepage: Well, overkilled for this PR, but why not.
I thought about setting the `devShell` command to `--root-path ${self}`, but then it would not have been dynamic anymore.

flake-modules/dev/list-plugins/default.nix
GaetanLepage: yes, I wanted to do this too, but:
1) I need to pass the absolute path to the flake (`self`)
2) I don't think that I can add a test or `passthru` with `writePython3Bin`.

flake-modules/dev/list-plugins/list-plugins.py
MattSturgeon: Good point. Normally we can't use `self` because that's the in-store flake and we often want the flake in the local git repo instead.

I'm guessing here it would be ok to use the in-store flake (`self`)?

flake-modules/dev/list-plugins/default.nix
MattSturgeon: > 1. I need to pass the absolute path to the flake (`self`)

Hm, I'm not sure why _that_ is a blocker?

> * I don't think that I can add a test or `passthru` with `writePython3Bin`.

Hm, maybe that's not made accessible? But it should be possible via `.overrideAttrs` in the worst case.

MattSturgeon (approved):

GaetanLepage:
@mergifyio queue

Issue #2693, opened 2024-12-17
[BUG] Build failures in darwin and nixos


bassforce86:
| Field        | Description |
|--------------|-------------|
| Plugin       | `nixvim`  | home manager module issue
| Nixpkgs      | `unstable` | unstable
| Home Manager | `unstable` | unstable

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

I have encountered what I _think_ is a bug in the home manager module.

When running switch (from a flake, although I suspect the may also occur using channels) is a build failure when trying to create `devshells` which are imported from a flake, see error below/

```txt
 … while evaluating definitions from `/nix/store/18li13082730qixvnqc8wdvjlwckq532-source/wrappers/hm.nix':
       … while evaluating definitions from `/nix/store/s6agnzjr3kg5sayjy0b0ms52bkvifqrv-source/modules/transposition.nix':
       … while evaluating definitions from `/nix/store/s6agnzjr3kg5sayjy0b0ms52bkvifqrv-source/modules/perSystem.nix':
       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `perSystem.aarch64-darwin.devshells' does not exist. Definition values:
       - In `/nix/store/18li13082730qixvnqc8wdvjlwckq532-source/flake-modules/dev/list-plugins, via option perSystem':
           {
             default = {
               commands = [
                 {
                   command = "/nix/store/m83ppxx1b0m7kn75mdqph66y2zixqw9l-list-plugins/bin/list-plugins \"$@\"";
           ...
```
this is on my Macbook using nix-darwin, however this also happens on my nixos system (x86_64-linux). I believe this is just a typo somewhere referencing `devshells` instead of `devShells`, but i'm not entirely sure where yet.

i'm able to mitigate this error by pinning nixvim to `ec24d496d52c2620b5ce3d237a2a1029a197b412`

error occurs when using nixvim at `30895485c3a31bb16ace513def4f3a36bfeb68c6`

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
}
```

bassforce86:
i think i've found the offending line causing this break: https://github.com/nix-community/nixvim/blob/main/flake-modules/dev/list-plugins/default.nix#L35

bassforce86:
I've created a PR and confirmed that this fixes my issue on Darwin.

[...]

PR #2694, opened 2024-12-17
Fix minor typo causing build failures.

bassforce86:
Relates to: #2693 

@khaneliman @MattSturgeon

### DIFF
--- a/flake-modules/dev/list-plugins/default.nix
+++ b/flake-modules/dev/list-plugins/default.nix
@@ -32,7 +32,7 @@
             list-plugins --root-path ${self} > $out
           '';
 
-      devshells.default.commands = [
+      devShells.default.commands = [
         {
           name = "list-plugins";
           command = ''${lib.getExe config.packages.list-plugins} "$@"'';


GaetanLepage (approved):
I think that `devshells` is the correct name:
https://flake.parts/options/devshell

bassforce86:
Yes, for flake-parts it is the correct name.

but it's then being referenced a an "actual" flake here: https://github.com/nix-community/nixvim/commit/bef9feb446a9203a1343a5026970396bcae60f6f#diff-53345b746dda1748bd7ce380e4c321cdaa5030738a87a73d320ac95e1a02981fR11

which is what's causing the break i think. I don't have enough background knowledge on why we want to reference the entire flake itself to comment on which area is best to change, but this seemed like to change with the smallest scope to mitigate the problem 🙂 

See flake output schema here: https://nixos.wiki/wiki/Flakes#Output_schema

GaetanLepage (changes requested):
To be investigated further.

HeitorAugustoLN:
I guess what is happening here, is that list-plugins/default.nix is depends on numtide/devshell, and is not disconsidered when input is removed

MattSturgeon:
> I guess what is happening here, is that list-plugins/default.nix is depends on numtide/devshell, and is not disconsidered when input is removed

Correct, well deduced!

The issue is here:
https://github.com/nix-community/nixvim/blob/ba43815abf6b101c13fe1d8d93323ca2e1689315/flake-modules/dev/default.nix#L3-L9

The `./list-plugins` import should be conditional on the relevant inputs being present.

HeitorAugustoLN:
#2699 should fix it

PR #2696, opened 2024-12-17
plugins/palette: fix minor typo;

bitterlox:
the option is declared as settings.customPalettes, but is accessed as settings.custom_palettes;
corrected the tests as well;

### DIFF
--- a/plugins/colorschemes/palette.nix
+++ b/plugins/colorschemes/palette.nix
@@ -123,7 +123,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
         (
           name: defaultPaletteNames:
           let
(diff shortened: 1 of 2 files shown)


plugins/colorschemes/palette.nix
```diff
@@ -123,7 +123,7 @@ lib.nixvim.neovim-plugin.mkNeovimPlugin {
         (
           name: defaultPaletteNames:
           let
-            customPalettesNames = lib.attrNames cfg.settings.custom_palettes.${name};
+            customPalettesNames = lib.attrNames cfg.settings.customPalettes.${name};
```
khaneliman, line 126: this should stay the same and the option should be changed

khaneliman:

PR #2698, opened 2024-12-17
plugins/openscad: migrate to mkVimPlugin

GaetanLepage:
- **lib/vim-plugin: expose new helpers `settingsOptionDescription` and `processPrefixedGlobals`**
- **modules: let extraPlugins accept null values**
- **plugins/openscad: migrate to mkVimPlugin**

### DIFF
--- a/lib/vim-plugin.nix
+++ b/lib/vim-plugin.nix
@@ -1,5 +1,29 @@
 { lib }:
-{
+rec {
+  mkSettingsOptionDescription =
+    { name, globalPrefix }:
+    ''
+      The configuration options for **${name}** without the `${globalPrefix}` prefix.
+
+      For example, the following settings are equivialent to these `:setglobal` commands:
+      - `foo_bar = 1` -> `:setglobal ${globalPrefix}foo_bar=1`
+      - `hello = "world"` -> `:setglobal ${globalPrefix}hello="world"`
+      - `some_toggle = true` -> `:setglobal ${globalPrefix}some_toggle`
+      - `other_toggle = false` -> `:setglobal no${globalPrefix}other_toggle`
+    '';
+
+  mkSettingsOption =
+    {
+      options ? { },
+      example ? { },
+      name,
+      globalPrefix,
+    }:
+    lib.nixvim.mkSettingsOption {
+      inherit options example;
+      description = mkSettingsOptionDescription { inherit name globalPrefix; };
+    };
+
   mkVimPlugin =
     {
       name,
@@ -36,18 +60,10 @@
       createSettingsOption = (lib.isString globalPrefix) && (globalPrefix != "");
 
       settingsOption = lib.optionalAttrs createSettingsOption {
-        settings = lib.nixvim.mkSettingsOption {
+        settings = mkSettingsOption {
           options = settingsOptions;
           example = settingsExample;
-          description = ''
-            The configuration options for **${name}** without the `${globalPrefix}` prefix.
-
-            For example, the following settings are equivialent to these `:setglobal` commands:
-            - `foo_bar = 1` -> `:setglobal ${globalPrefix}foo_bar=1`
-            - `hello = "world"` -> `:setglobal ${globalPrefix}hello="world"`
-            - `some_toggle = true` -> `:setglobal ${globalPrefix}some_toggle`
-            - `other_toggle = false` -> `:setglobal no${globalPrefix}other_toggle`
-          '';
+          inherit name globalPrefix;
         };
       };
 
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -29,9 +29,10 @@ in
 {
   options = {
     extraPlugins = mkOption {
-      type = with types; listOf (either package pluginWithConfigType);
+      type = with types; listOf (nullOr (either package pluginWithConfigType));
       default = [ ];
       description = "List of vim plugins to install";
+      apply = builtins.filter (p: p != null);
     };
 
     extraPackages = mkOption {
--- a/plugins/by-name/openscad/default.nix
+++ b/plugins/by-name/openscad/default.nix
@@ -1,81 +1,93 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-with lib;
 let
-  defaultFuzzyFinder = "skim";
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts applyPrefixToAttrs;
+  inherit (lib.nixvim.vim-plugin) mkSettingsOptionDescription;
+
+  name = "openscad";
+  globalPrefix = "openscad_";
 in
-{
-  options.plugins.openscad = {
-    enable = mkEnableOption "openscad.nvim, a plugin to manage OpenSCAD files";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  inherit name;
+  packPathName = "openscad.nvim";
+  package = "openscad-nvim";
 
-    package = lib.mkPackageOption pkgs "openscad.nvim" {
-      default = [
-        "vimPlugins"
-        "openscad-nvim"
-      ];
-    };
+  maintainers = [ lib.maintainers.GaetanLepage ];
 
-    fuzzyFinder = helpers.defaultNullOpts.mkEnum [
-      "skim"
-      "fzf"
-    ] defaultFuzzyFinder "fuzzy finder to find documentation";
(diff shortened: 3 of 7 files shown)


plugins/by-name/openscad/default.nix
```diff
@@ -1,81 +1,170 @@
...
+      {
+        old = "helpManualTrigger";
+        new = "help_manual_trig_key";
+      }
+      {
+        old = "execOpenSCADTrigger";
+        new = "exec_openscad_trig_key";
+      }
+      "topToggle"
+    ]);
```
MattSturgeon, line 64: This can all go in `optionsRenamedToSettings` now, right?

plugins/by-name/openscad/default.nix
```diff
@@ -1,81 +1,170 @@
...
-    };
+    exec_openscad_trig_key = defaultNullOpts.mkStr "<A-o>" ''
+      Keyboard shortcut for opening the current file in OpenSCAD.
+    '';
+
+    top_toggle = defaultNullOpts.mkStr "<A-c>" ''
+      Keyboard shortcut for toggling `htop` filtered for OpenSCAD processes.
+    '';
+  };
+  settingsExample = {
```
MattSturgeon:
```suggestion
  };

  settingsExample = {
```

plugins/by-name/openscad/default.nix
```diff
@@ -1,81 +1,170 @@
...
+    load_snippets = true;
+    fuzzy_finder = "fzf";
+    cheatsheet_window_blend = 15;
+    auto_open = true;
   };
+  settingsDescription = settingsOptionDescription { inherit name globalPrefix; };
```
MattSturgeon:
```suggestion
  };

  settingsDescription = settingsOptionDescription { inherit name globalPrefix; };
```

Also: consider having `settingsDescription` declared at the top of the file, with other trivial metadata?

lib/vim-plugin.nix
```diff
@@ -1,5 +1,34 @@
 { lib }:
+let
+  settingsOptionDescription =
```
MattSturgeon, line 3:
```suggestion
  mkSettingsOptionDescription =
```

I think a `mk` prefix is appropriate for factory functions

plugins/by-name/openscad/default.nix
```diff
@@ -1,81 +1,170 @@
...
+              defaultPluginName
+            ];
+      in
+      lib.mkPackageOption pkgs "fuzzy finder" {
+        nullable = true;
+        inherit default;
+        extraDescription = ''
+          If ${optionName} is set explicitly, this option will default to
+        '';
+      };
```
MattSturgeon: I'd much rather have a `literalMD` type `defaultText` than explain the default in the `description`

Maybe `mkPackageOption` supports `defaultText`? Looks like it does not: https://noogle.dev/f/lib/mkPackageOption

Alternatively, we can define the option manually, or just do `// { defaultText = ... }`

```nix
lib.mkPackageOption pkgs "fuzzy finder" {
  nullable = true;
  inherit default;
}
// {
  defaultText = literalMD ''
    If `${optionName}` is set explicitly, then either
    ${concatStringsSep ", " (attrValues defaultPluginNames)}
    as appropriate. Otherwise, `null`.
  '';
}
```

MattSturgeon (commented):
My first reaction was "should this be three PRs", but the changes are trivial enough to lump in with the plugin migration :grin:

lib/vim-plugin.nix
```diff
@@ -1,5 +1,34 @@
...
+    }:
+    lib.nixvim.mkSettingsOption {
+      inherit options example;
+      description = settingsOptionDescription { inherit name globalPrefix; };
+    };
+
+  processPrefixedGlobals = prefix: lib.mapAttrs' (n: lib.nameValuePair (prefix + n));
+in
 {
+  inherit settingsOptionDescription mkSettingsOption processPrefixedGlobals;
```
MattSturgeon, line 30: Curious why you used a `let-in` block here? Wouldn't it be neater to define all public functions in the output attrset, and then if need access them via the `lib` input argument?

GaetanLepage: I wanted to avoid using `rec` but was not too sure. What do you prefer ?

plugins/by-name/openscad/default.nix
GaetanLepage: I think you reviewed just before I pushed something in this style.

GaetanLepage: We had the same idea at the same time haha. Any feedback on the wording ?

plugins/by-name/openscad/default.nix
GaetanLepage: Hmm, not sure about this one. It is solely used there, so I don't see the point of "spreading" the information.
If you strongly prefer moving it, I will do it though.

plugins/by-name/openscad/default.nix
GaetanLepage: image

MattSturgeon: Since `settings.fuzzy_finder` defaults to null, it is actually better not mentioning "explicit" definitions. So this is perfect.

I wasn't sure how a multi-line value would render, but it looks great in your screenshot.

lib/vim-plugin.nix
MattSturgeon: Not a fan of using `rec`, as it'll block us from ever making the `lib` properly extendible/overridable. We have recursive access to `lib.nixvim`. But I guess it's fine for now.

GaetanLepage: That's why I went with `let`-`in` in the first place. Do you want me to revert to that ?

plugins/by-name/openscad/default.nix
```diff
@@ -1,81 +1,93 @@
...
-    enable = mkEnableOption "openscad.nvim, a plugin to manage OpenSCAD files";
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  inherit name;
+  packPathName = "openscad.nvim";
+  package = "openscad-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  # TODO: Added 2024-12-17; remove after 25.05
+  optionsRenamedToSettings = import ./renamed-options.nix;
```
MattSturgeon, line 23: Shall we make `optionsRenamedToSettings` auto-import if needed?

```
fn = maybeList:
if isPath maybeList then fn (import maybeList) else # do thing...
```

GaetanLepage: No, to me that's a bit too overkilled. It would hide even more semantics. I would prefer to KISS in this case.

lib/vim-plugin.nix
MattSturgeon: let-in is also not overridable. So in that scenario its just as bad.

What I would suggest is `inherit (lib.nixvim) settingsOptionDescription mkSettingsOption processPrefixedGlobals`, but as I say. Not currently a big deal.

plugins/by-name/openscad/default.nix
MattSturgeon: It's pretty common practice in nixpkgs for utils like `callPackage` to accept _either_ a value or a path that contains a value when imported. But no issues with not doing that.

lib/vim-plugin.nix
```diff
@@ -1,5 +1,31 @@
...
+      example ? { },
+      name,
+      globalPrefix,
+    }:
+    lib.nixvim.mkSettingsOption {
+      inherit options example;
+      description = mkSettingsOptionDescription { inherit name globalPrefix; };
+    };
+
+  applyPrefixToAttrs = prefix: lib.mapAttrs' (n: lib.nameValuePair (prefix + n));
```
MattSturgeon, line 27: I'd rather this be in `lib/utils.nix`

MattSturgeon (approved):
Nice work!

modules/output.nix
```diff
@@ -29,9 +29,10 @@ in
 {
   options = {
     extraPlugins = mkOption {
-      type = with types; listOf (either package pluginWithConfigType);
+      type = with types; listOf (nullOr (either package pluginWithConfigType));
       default = [ ];
       description = "List of vim plugins to install";
+      apply = builtins.filter (p: p != null);
```
khaneliman, line 35: <3

GaetanLepage:
Are we good ?

MattSturgeon (approved):

GaetanLepage:
@mergify queue

PR #2700, opened 2024-12-17
plugins/lsp: migrate to mkNeovimPlugin

HeitorAugustoLN:
- Migrated LSP to mkNeovimPlugin
- Added support for setting mode in non-extra lsp keymaps

It should probably be fine, since I didn't change anything besides support for mode

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -163,13 +163,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
             key: action:
             let
               actionStr = action.action or action;
+              mode = action.mode or "n";
               actionProps = lib.optionalAttrs (builtins.isAttrs action) (
-                builtins.removeAttrs action [ "action" ]
+                builtins.removeAttrs action [
+                  "action"
+                  "mode"
+                ]
               );
             in
             {
-              inherit key;
-              mode = "n";
+              inherit key mode;
               action = lib.nixvim.mkRaw (prefix + actionStr);
 
               options = {
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -1,231 +1,222 @@
-{
-  lib,
-  helpers,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
+{ lib, ... }:
 let
-  cfg = config.plugins.lsp;
+  inherit (lib) mkOption types;
 in
-{
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lsp";
+  packPathName = "nvim-lspconfig";
+  package = "nvim-lspconfig";
+
+  callSetup = false;
+  hasSettings = false;
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
   imports = [ ./language-servers ];
 
-  options = {
-    plugins.lsp = {
-      enable = mkEnableOption "neovim's built-in LSP";
-
-      package = lib.mkPackageOption pkgs [
-        "vimPlugins"
-        "nvim-lspconfig"
-      ] { };
-
-      keymaps = {
-        silent = mkOption {
-          type = types.bool;
-          description = "Whether nvim-lsp keymaps should be silent";
-          default = false;
-        };
-
-        diagnostic = mkOption {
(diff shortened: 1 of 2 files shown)


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -22,16 +22,16 @@
...
   ...
 }:
-with lib;
 let
   cfg = config.plugins.lsp.servers.${name};
   opts = options.plugins.lsp.servers.${name};
 
   enabled = config.plugins.lsp.enable && cfg.enable;
+
+  inherit (lib) mkOption types;
```
GaetanLepage, line 34:
```suggestion

  inherit (lib.nixvim) defaultNullOpts mkNullOrOption;
  inherit (lib) mkOption types;
```
Just at a glance, this will allow you to get rid of `lib.nixvim.` in this file.

HeitorAugustoLN: It looks like there is more than that

plugins/lsp/default.nix
```diff
@@ -1,231 +1,225 @@
...
 
-      postConfig = mkOption {
-        type = types.lines;
-        description = "Code to be run after loading the LSP. This is an internal option";
-        default = "";
-      };
+    preConfig = mkOption {
+      type = types.lines;
+      description = "Code to be run before loading the LSP. Useful for requiring plugins";
+      default = "";
```
MattSturgeon, line 147: I wonder if `preConfig` and `postConfig` can be rename-alias modules pointing to `lubConfig.pre` and `luaConfig.post` respectively?

plugins/lsp/default.nix
```diff
@@ -1,231 +1,225 @@
...
-              action.__raw = "require('telescope.builtin').lsp_definitions";
-            }
-            {
-              key = "K";
-              action = "<CMD>Lspsaga hover_doc<Enter>";
-            }
-          ];
-          default = [ ];
+      lspBuf = mkOption {
+        type = with types; attrsOf (either str (attrsOf anything));
```
MattSturgeon, line 36: I'm thinking we can a) use type coercion to reduce the post-processing and b) use our standard `lib.nixvim.keymaps.mapOptionSubmodule` for increased consistency.

We may need to extend `mapOptionSubmodule` to support `key` defaulting to the attr-name though.

I'm thinking something like
```nix
type = coercedTo str (action: {inherit action;}) mapOptionSubmodule;
```
(Although we'd actually need to use the map-option submodule factory to produce a slightly customised type)

plugins/lsp/language-servers/_mk-lsp.nix
MattSturgeon: I'm in favour of inheriting anything that is used _a lot_. I'm not against being explicit, but `defaultNullOpts` for example gets used a ton.

plugins/lsp/default.nix
HeitorAugustoLN: I think it could, but it would need a little refactor

plugins/lsp/default.nix
HeitorAugustoLN: I like the idea, but I don't know if it would be possible to make it backwards-compatible. But if it is not a problem, then I think this is better.

traxys: When migrating plugins to mkNeovimPlugin it's acceptable to introduce some breaking changes. You can mark the old option as removed with instructions on how to migrate

MattSturgeon: > You can mark the old option as removed with instructions on how to migrate

This only really applies when the new option has a different name/location.

If the old "removed" option has the same name as the new option that doesn't work.

In this case, I think we can do a migration without breaking changes. But since the option isn't being renamed, it can wait until after this PR.

traxys: Yeah I meant for options that we move from the top-level to `settings`, but if this option does not move then there's not much we can do

khaneliman (approved):
Don't wanna let too many of these rot, we can make improvements in follow ups.

khaneliman:
@mergifyio queue

PR #2702, opened 2024-12-17
lib: add applyPrefixToAttrs

GaetanLepage:


lib/vim-plugin.nix
```diff
@@ -113,7 +113,7 @@
                 extraPlugins = extraPlugins ++ [
                   (cfg.packageDecorator cfg.package)
                 ];
-                globals = lib.mapAttrs' (n: lib.nameValuePair (globalPrefix + n)) (cfg.settings or { });
+                globals = lib.applyPrefixToAttrs globalPrefix (cfg.settings or { });
```
MattSturgeon, line 116:
```suggestion
                globals = lib.nixvim.applyPrefixToAttrs globalPrefix (cfg.settings or { });
```

Or `lib.nixvim.utils.applyPrefixToAttrs` if we don't alias it to the top-level in `lib/default.nix`

MattSturgeon (approved):

GaetanLepage:
@mergify queue

PR #2705, opened 2024-12-18
plugins/cmp/sources: use mkNeovimPlugin instead of mkVimPlugin for source plugins

GaetanLepage:


MattSturgeon (approved):

GaetanLepage:
@mergify queue

PR #2706, opened 2024-12-18
plugins/cmp-tabby: refactor

GaetanLepage:


MattSturgeon:
Do you think this would be more maintainable as its own `plugins/by-name` module?

It can still be a cmp plugin in the same way that (e.g.) otter does it.

MattSturgeon:
This can be rebased without bf6a68f406577718d31c67103769b7816e46b420, right?

plugins/by-name/cmp-tabby/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "cmp-tabby";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  imports = [
+    { cmpSourcePlugins.cmp_tabby = "cmp-tabby"; }
```
MattSturgeon, line 12: IDK why, I need to double check this every time...

It is correct though :+1: 

https://github.com/nix-community/nixvim/blob/5abe382c54eff3cd3977733e9d268ab8994f397b/plugins/cmp/auto-enable.nix#L38

MattSturgeon (approved):
Awesome

GaetanLepage:
@mergify queue

Issue #2711, opened 2024-12-18
[BUG] Can't override "K" with `plugins.lsp.keymaps.lspBuf`


jfly:
| Field        | Description |
|--------------|-------------|
| Plugin       | lsp  | 
| Nixpkgs      | N/A | 
| Home Manager | N/A | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

I'm trying to specify the `K` normal mode mapping when dealing with LSPs, as documented [here](https://nix-community.github.io/nixvim/plugins/lsp/keymaps/index.html#pluginslspkeymapslspbuf).

This binding doesn't take precedence over [nvim's builtin mapping](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp.lua/#L737-L739).

## Root cause

Here's what happens:

1. [`Client:on_attach` runs](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp/client.lua/#L1009)
2. [It invokes `lsp._set_defaults`](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp/client.lua/#L1012), which [sees that there's no normal mode mapping for `K`](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp.lua/#L735), so it [registers a *buffer-specific* normal mode mapping for `K`](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp.lua/#L737-L739).
3. Next, [`Client:on_attach` triggers the 'LspAttach' autocmd](https://github.com/neovim/neovim/blob/160cbd0ef4c905b849beb969e8b5a56064c7cea7/runtime/lua/vim/lsp/client.lua/#L1014-L1018). That triggers the [`keymapsOnEvents.LspAttach` that nixvim's lsp plugin uses to register its `lspBuf` keymappings](https://github.com/nix-community/nixvim/blob/eaa20846279654fa038427ce078d4cd9609e5f2f/plugins/lsp/default.nix#L215).
[...]

PR #2712, opened 2024-12-18
modules/keymap: include buffer when registering `keymapsOnEvents`

jfly:
This feels more correct to me: I believe we were registering a keymap for all buffers, even if only some of them have experienced the relevant autocmd.

I don't know if it's guaranteed here that `args.buf` always exists (but maybe it's not a problem if it doesn't exist, because we'll still register a keymap for all buffers?).

This fixes https://github.com/nix-community/nixvim/issues/2711

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -97,11 +97,13 @@
       inherit event;
       group = "nixvim_binds_${event}";
       callback = helpers.mkRaw ''
-        function()
+        function(args)
           do
             local __nixvim_binds = ${lib.nixvim.toLuaObject (map helpers.keymaps.removeDeprecatedMapAttrs mappings)}
+
             for i, map in ipairs(__nixvim_binds) do
-              vim.keymap.set(map.mode, map.key, map.action, map.options)
+              local options = vim.tbl_extend("error", map.options or {}, { buffer = args.buf })
+              vim.keymap.set(map.mode, map.key, map.action, options)
             end
           end
         end


MattSturgeon (approved):
Makes sense to me, but I'll hold off for a second opinion @nix-community/nixvim

traxys (approved):

MattSturgeon:
@mergify queue

jfly:
Thanks for the quick merge, @matt

khaneliman:
Bisecting a lua error points this commit as the cause. 

```bash
Error executing lua callback: vim/shared.lua:0: key found in more than one map: buffer
stack traceback:
	[C]: in function 'error'
	vim/shared.lua: in function 'tbl_extend'
	/nix/store/d7h1rgpfh4ds52p7mj71gpj1yvgs00cg-init.lua: in function </nix/store/d7h1rgpfh4ds52p7mj71gpj1yvgs00cg-init.lua:0>
	[C]: in function 'nvim_exec_autocmds'
	...ompiled-0.10.2/share/nvim/runtime/lua/vim/lsp/client.lua: in function '_on_attach'
	...ompiled-0.10.2/share/nvim/runtime/lua/vim/lsp/client.lua: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```

jfly:
@khaneliman, I opted to use `vim.tbl_extend("error", ...)` to try to avoid suprising people that we're now specifying a buffer for them.

Are you specifying an `autoCmd` that specifies a buffer?

khaneliman:
I've got only a handful of autoCmds https://github.com/search?q=repo%3Akhaneliman%2Fkhanelivim%20autoCmd&type=code and then a bunch of plugins with will create their own.

jfly:
I guess the easiest way to track this down would be to look at your generated init.lua file. Can you give me a command to repro the issue?

MattSturgeon:
Hm, maybe extending with "error" mode isn't the best approach? If a user has explicitly configured something, that should override whatever defaults we're setting. :thinking:

jfly:
> If a user has explicitly configured something, that should override whatever defaults we're setting.

Generally I'd agree, but in this case, I think we're talking about a buffer hardcoded at build time? Which doesn't make any sense to me. Perhaps I'm fundamentally misunderstanding what's happening here. Having a repro would be nice to set me straight.

MattSturgeon:
> a buffer hardcoded at build time

I believe you can use "buffer 0" to reference the "current" buffer? So maybe some people are explicitly doing that?

I haven't looked too closely, so it's likely I'm missing something fundamental too. But my first thought is that if we don't want to allow setting buffnr through nixvim options, that should be a eval time filter and/or warning/assertion.

jfly:
Ah yeah, using "buffer 0" could make some sense.

Sorry for the breakage. I'm leaving for a trip soon, and won't have any real time to spend on this until the new year. Here's a quickfix: https://github.com/nix-community/nixvim/pull/2719.

PR #2715, opened 2024-12-18
lib/plugins: structual refactor

MattSturgeon:
- **lib/plugins: organise plugin-factory functions in a subdir**
- **lib/plugins: split vim utils into separate file**
- **lib/plugins: separate main factory functions**
- **lib/plugins: take ownership of `modules` utils**
- **lib/plugins: extract common logic for `package` options**
- **lib/plugins: call sub-components with relevant args**

This PR moves all logic related to `mkVimPlugin` and `mkNeovimPlugin` into a subdirectory `lib/plugins`.

Further, it moves some (very internal) helper functions out of `lib.nixvim.modules` and into `lib.nixvim.plugins.utils`.

To aid in focusing the review, I have not done any tree-wide migration in this PR, although most migration to the new `lib.nixvim.plugins.mk*Plugin` names should be trivial with a regex match.

In future, this new structure should make adding common helpers/utilities easier. It should also make combining `mkVimPlugin` and `mkNeovimPlugin` together _slightly_ simpler.


khaneliman (approved):
I like the organization and ability to start abstracting things into utils.

MattSturgeon:
@mergify queue

PR #1205, opened 2024-03-04
 plugins/crates: init

GaetanLepage:
Add support for [crates.nvim](https://github.com/Saecki/crates.nvim/tree/main).

I am not convinced on whether I should explicitly declare some options or none...

Fixes #1149

PR #2717, opened 2024-12-18
plugins/crates: rename from crates-nvim to crates, move to by-name, migrate to mkNeovimPlugin

### LINKED ISSUE #1149: [PLUGIN REQUEST] crates.nvim

GaetanLepage:
#1205 will have to be rebases on this.

Fixes #1149


plugins/by-name/crates/default.nix
```diff
@@ -0,0 +1,30 @@
...
+    lib.nixvim.mkSettingsRenamedOptionModules [ "plugins" "crates-nvim" ]
+      [ "plugins" "crates" ]
+      [
+        "enable"
+        "package"
+        {
+          old = "extraOptions";
+          new = "settings";
+        }
+      ];
```
MattSturgeon, line 22: This can be simplified to use `optionsRenamedToSettings`, right?

MattSturgeon: Oh, nevermind. I missed that it's doing the plugin rename

MattSturgeon: I'm not sure ottomh if this works. Doesn't `mkSettingsRenamedOptionModules` automatically nest into `settings`?

GaetanLepage: No, I need to also rename `enable` and `package` . This is one level upper.

GaetanLepage: > I'm not sure ottomh if this works. Doesn't `mkSettingsRenamedOptionModules` automatically nest into `settings`?

No. It takes two paths `old` and `new. Then, it does all the renames using the provided list of sub-paths, each being relative to `old` (for the rename source) and `new` (for the rename destination).

MattSturgeon (approved):

GaetanLepage:
@mergify queue

MattSturgeon:
> #1205 will have to be rebases on this.

I'm confused. Isn't that PR "adding" the same plugin this PR is dealing with?

GaetanLepage:
> I'm confused. Isn't that PR "adding" the same plugin this PR is dealing with?

You are right. I was meaning that there could be a follow up to add option declarations, LSP integration...

PR #2718, opened 2024-12-19
Revert "plugins/cmp/sources: use mkNeovimPlugin instead of mkVimPlugin for source plugins"

GaetanLepage:
This reverts commit 6a9d840370647abb0c980ff1bb8e47edd1255789.

It is not really needed anymore.

### DIFF
--- a/plugins/cmp/sources/_mk-cmp-plugin.nix
+++ b/plugins/cmp/sources/_mk-cmp-plugin.nix
@@ -14,12 +14,8 @@
   imports ? [ ],
   ...
 }@args:
-lib.nixvim.neovim-plugin.mkNeovimPlugin (
-  {
-    hasSettings = false;
-    callSetup = false;
-  }
-  // builtins.removeAttrs args [
+lib.nixvim.vim-plugin.mkVimPlugin (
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Let's hold off on flip-flopping back and forth unless you think there's major downsides to using `mkNeovimPlugin` here?

Maybe it'd be better to invest effort into _merging_ the plugin factories? `mkCmp` could then use a `mkSimplePlugin` wrapper that doesn't support any settings options.

GaetanLepage:
> Let's hold off on flip-flopping back and forth unless you think there's major downsides to using `mkNeovimPlugin` here?
> 
> Maybe it'd be better to invest effort into _merging_ the plugin factories? `mkCmp` could then use a `mkSimplePlugin` wrapper that doesn't support any settings options.

Well, now that I have finished the cmp refactoring (for now), I would rather close the topic rather than leaving this "TODO".
I don't see why we would have to re-enable this, but even in this case, we could re-revert.
P(flip-flop-flip) < 0.1 according to me.

MattSturgeon (approved):

GaetanLepage:
@mergify queue

PR #2719, opened 2024-12-20
modules/keymap: quickfix, use "keep" rather than "error"

jfly:
There are some codepaths out there in the wild that seem to be explicitly setting a buffer at build time. I'm not convinced that's a sane thing to do, but I don't have time to dig any further right now.

See the discussion starting here:
https://github.com/nix-community/nixvim/pull/2712#issuecomment-2556330101 for more context.


traxys (approved):

traxys:
@mergifyio queue

PR #2727, opened 2024-12-21
lib: mk overridable & extendable; also move flake output

MattSturgeon:
- make overrideable (`override` function is added by `makeOverridable`)
- make nixvim-lib extendable (`extend` function is added by `makeExtensible`)
- treewide: access nixvim-lib via flake instead of directly importing it

Add a non-system specific `<flake>.lib.nixvim` output, which is equivalent to the existing `<flake>.lib.<system>.helpers` output.
A warning is traced when the per-system output is used.

This is now also wrapped with `lib.makeOverridable` to allow overriding the function args used to construct the nixvim-lib. Additionally, `lib.fix` was replaced with `lib.makeExtensible`, allowing nixvim's sub-lib to be directly extended via overlays.

Throughout the codebase, nixvim-lib now comes from the flake outpets, rather than being imported manually. Where `_nixvimTests` needs to be passed, we use `.override`.

---

Note, any overlays will be reset if you subsequently use `override`. E.g.
```nix
{ lib }:
rec {
  extended = lib.nixvim.extend (final: prev: {
    # Overlay
    hello-world = "hi there";
  });

  overridden = extended.override {/* args */};

  extendedHasHello = extended ? hello-world; # true

  overriddenHasHello = overridden ? hello-world; # false
}
```

---

In follow up work I'd also like to offer a `<flake>.lib.overlay` which should replace the current `lib.nixvim.extendedLib` hack. This will allow wrapper modules and end-users to apply nixvim's lib extensions to any compatible nixpkgs lib, rather than always using the `lib` that provided by flake-parts.

Perhaps this could also be done automatically by `evalNixvim` when users supply their own `specialArgs.lib`?

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -1,4 +1,5 @@
 {
+  helpers,
   system,
   nixpkgs,
   nuschtosSearch,
@@ -7,7 +8,6 @@ let
   # We overlay a few tweaks into pkgs, for use in the docs
   pkgs = import ./pkgs.nix { inherit system nixpkgs; };
   inherit (pkgs) lib;
-  helpers = import ../lib { inherit lib; };
 
   nixvimPath = toString ./..;
 
--- a/flake-modules/lib.nix
+++ b/flake-modules/lib.nix
@@ -6,24 +6,29 @@
   ...
 }:
 {
-  _module.args.helpers = import ../lib {
-    inherit lib;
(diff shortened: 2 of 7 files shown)


khaneliman (approved):
LGTM thanks

MattSturgeon:
@mergify queue

PR #2729, opened 2024-12-21
dev/new-plugin: originalName -> packPathName

khaneliman:


MattSturgeon:
Hm, I think we'll need an integration test to prevent similar regressions from happening

MattSturgeon (approved):

khaneliman:
> Hm, I think we'll need an integration test to prevent similar regressions from happening.
> 
> E.g. we could test that the script creates valid plugin-modules that can eval, build, & run for a handful of example inputs.
> 
> We'd probably need to use a fake plugin name but a real package name? The fake name to avoid conflicts, the real package to ensure it can eval/build/run.

Yeah, that's a good idea.

khaneliman:
@mergifyio queue

PR #2731, opened 2024-12-21
lib/overlay: init

MattSturgeon:
The overlay allows extending any instance of nixpkgs-lib into a nixvim-compatible "extended" instance.

For example, this enables users to apply their own extensions, and then apply ours.

In the future, we could automatically extend any lib supplied to us in `evalNixvim`. If that's desired...

- Refactored some internals of the bootstrapping to enable this.
- Removed the somewhat hacky `extendedLib` attr

Closes #2328


lib/default.nix
```diff
@@ -9,14 +10,17 @@ lib.makeExtensible (
     # Used when importing parts of our lib
     call = lib.callPackageWith {
       inherit call self;
-      lib = self.extendedLib;
+      # If `prevLib` is provided, assume `lib` is already extended
+      # TODO: consider keeping lib & self explicitly separate
+      # This should improve overridability and simplify bootstrapping?
+      lib = if prevLib == null then lib.extend flake.lib.overlay else lib;
+      ${if prevLib == null then null else "prevLib"} = prevLib;
     };
```
khaneliman, line 18: I dont know if it's just me, but it feels hard to track what our if conditions are doing. Maybe a default behavior in the arguments to `callPackageWith` and merge in the other values when we see `prevLib == null`?

MattSturgeon: The complexity mainly comes from us using the "final" lib throughout our `lib/*.nix`.

I believe doing this also kinda bypasses `makeExtensible`, such that overlaying stuff may actually affect other dependent parts of our lib transitively.

---

My ideal would be to have a clean separation between `final`/`self` and `prev`/`lib` within out nixvim-lib.

But that would inflate the diff needlessly (for now) and also has some complications with getting access to the final merged `lib.types`.

---

`${null} = foo` is just an alternative to using `optionalAttrs`, but it avoids the need for an `//` update. Currently, we don't use `prevLib` anywhere, so we could just drop that line.

The line above is just responsible for making sure the `lib` pass to other parts of our nixvim-lib is an extended lib. As I say, my preference would be to simply not do this at all and have a clean separation internally.

MattSturgeon: I've simplified this to a more understandable `_isExtended` bool, but I still think this highlights some underlying issues with the current approach. https://github.com/nix-community/nixvim/pull/2731/files#diff-84f980eccc1d3f99c10f8b0e6c5c5fc2af2069517533c4c4d73264a190e85188R15

khaneliman (approved):
LGTM

MattSturgeon:
@mergify queue

PR #2732, opened 2024-12-22
treewide: migrate & deprecate `neovim-plugin` & `vim-plugin` aliases

MattSturgeon:
- **treewide: `neovim-plugin` -> `plugins`**
- **treewide: `vim-plugin` -> `plugins`**
- **lib/plugins: deprecate `neovim-plugin` & `vim-plugin` aliases**

Follow up to #2715, this PR deprecates the old attr names by showing a warning when they are used.

This PR includes the treewide migration, which was a fairly simple substitution pattern.


MattSturgeon:
@mergify queue

PR #2736, opened 2024-12-23
modules: add `flake` and `build.manDocsPackage`

MattSturgeon:
- **modules/context: `flake` option, provides access to our flake**
- **modules/output: add `manDocsPackage`**

Rather than only providing the configuration with access to `<flake>.inputs.nixpkgs`, we should probably pass the entire `<flake>` in. This is made available by a new `flake` option declared in `modules/misc/context.nix` and the default is set within `evalNixvim` in `lib/modules.nix`.

We then use this context option to define the default for `nixpkgs.source` and to enable a new `build.manDocsPackage`, exposing the man-docs package as part of the configuration.

Having the `manDocsPackage` as part of the module configuration will enable having the standalone package also be part of the configuration and/or having the locally installed docs be based on the local modules, rather than only what's in our flake.

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -30,11 +30,9 @@ in
     lib.evalModules {
       modules = modules ++ [
         ../modules/top-level
-
-        # Pass our locked nixpkgs into the configuration
         {
           _file = "<nixvim-flake>";
-          nixpkgs.source = lib.mkOptionDefault flake.inputs.nixpkgs;
+          flake = lib.mkOptionDefault flake;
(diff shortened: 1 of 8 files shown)


traxys (approved):

MattSturgeon:
@mergify queue

MattSturgeon:
@mergify refresh

MattSturgeon:
@mergify requeue

MattSturgeon:
@mergify queue

PR #2737, opened 2024-12-23
lib/modules: remove assertion message for removed `check` arg

MattSturgeon:
- **lib/modules: remove assertion message for removed `check` arg**
- **lib/modules: `flake` is now required**


traxys (approved):

MattSturgeon:
@mergify queue

PR #2738, opened 2024-12-23
modules/nixpkgs: complete the MVP

### LINKED ISSUE #1784: modules/misc: nixpkgs module

MattSturgeon:
This PR finally gets the `nixpkgs` module to a fully functional state. If `nixpkgs.pkgs` is not defined, then an instance of nixpkgs is constructed from `nixpkgs.source`.

Constructing nixpkgs also requires knowing which system to target, so the `hostPlatform` and `buildPlatform` options are included.

I also included the `nixpkgs.config` option. This could just be `attrsOf anything`, but I've used the upstream option-type because it handles some special cases while merging definitions.

I've also added the non-standalone option `nixpkgs.useGlobalPackages`, which automatically sets `nixpkgs.pkgs` to the "host" or "global" packages; i.e. the `pkgs` arg in use by the outer module eval.

Most implementation is directly based on https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/nixpkgs.nix

Follow up to:
- #2835
- #2740
- #2670
- #2430
- #2301

Closes #2022
Closes #2437
Fixes #1784

#### `system` agnostic Nixvim factory

- #2210

Constructing `pkgs` within the module system allows us to have a system-agnostic replacement for `makeNixvimWithModule` and somewhat simplify the [standalone template](https://github.com/nix-community/nixvim/blob/main/templates/simple/flake.nix)

Note: `lib.nixvim.modules.evalNixvim` is already system-agnostic. We simply need to provide documentation and/or new wrapper-functions for use by end-users.

#### Migrating `useGlobalPackages` default

- #2839

I'm not 100% sure I know a good way to warn about that planned transition though. Maybe we'll just have to have an obnoxious but temporary warning shown to all users? Could still be missed by users who don't update very often.

### DIFF
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -1,11 +1,10 @@
 {
   self,
-  pkgs,
-  lib ? pkgs.lib,
-  ...
+  system,
+  lib,
 }:
 let
-  defaultPkgs = pkgs;
+  defaultSystem = system;
 
   # Create a nix derivation from a nixvim executable.
   # The build phase simply consists in running the provided nvim binary.
@@ -31,7 +30,7 @@ let
     {
       name ? null,
       pkgs ? null,
-      system ? defaultPkgs.stdenv.hostPlatform.system,
+      system ? defaultSystem,
       module,
       extraSpecialArgs ? { },
     }:
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -30,7 +30,8 @@ let
   mkTestDerivationFromNixvimModule =
     {
       name ? null,
-      pkgs ? defaultPkgs,
+      pkgs ? null,
+      system ? defaultPkgs.stdenv.hostPlatform.system,
       module,
       extraSpecialArgs ? { },
     }:
@@ -42,14 +43,18 @@ let
         _nixvimTests = true;
       };
 
+      systemMod =
+        if pkgs == null then
+          { nixpkgs.hostPlatform = lib.mkDefault { inherit system; }; }
+        else
+          { nixpkgs.pkgs = lib.mkDefault pkgs; };
+
       result = helpers.modules.evalNixvim {
         modules = [
           module
           (lib.optionalAttrs (name != null) { test.name = name; })
           { wrapRc = true; }
-          # TODO: Only do this when `args?pkgs`
-          # Consider deprecating the `pkgs` arg too...
-          { nixpkgs.pkgs = lib.mkDefault pkgs; }
+          systemMod
         ];
         inherit extraSpecialArgs;
       };
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -34,20 +34,6 @@ in
 {
   options.nixpkgs = {
     pkgs = lib.mkOption {
-      # TODO:
-      # defaultText = lib.literalExpression ''
-      #   import "''${nixos}/.." {
-      #     inherit (cfg) config overlays localSystem crossSystem;
-      #   }
-      # '';
-      defaultText = lib.literalMD ''
-        If `useGlobalPackages` is true, `pkgs` is inherited from your host config
-        (i.e. NixOS, home-manager, or nix-darwin).
-        Or the `pkgs` supplied to `makeNixvimWithModule` when building a standalone nixvim.
-
-        > [!CAUTION]
-        > This default will be removed in a future version of nixvim
-      '';
       type = lib.types.pkgs // {
         description = "An evaluation of Nixpkgs; the top level attribute set of packages";
       };
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -164,7 +164,8 @@ in
       apply = lib.systems.elaborate;
       defaultText = lib.literalMD ''
         - Inherited from the "host" configuration's `pkgs`
-        - Must be specified manually when building a standalone nixvim
+        - Or `evalNixvim`'s `system` argument
+        - Otherwise, must be specified manually
       '';
       description = ''
         Specifies the platform where the Nixvim configuration will run.
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -172,6 +172,52 @@ in
       '';
     };
 
+    hostPlatform = lib.mkOption {
+      type = with lib.types; either str attrs;
+      example = {
+        system = "aarch64-linux";
+      };
+      apply = lib.systems.elaborate;
+      defaultText = lib.literalMD ''
+        - Inherited from the "host" configuration's `pkgs`
+        - Must be specified manually when building a standalone nixvim
+      '';
+      description = ''
+        Specifies the platform where the Nixvim configuration will run.
+
+        To cross-compile, also set `nixpkgs.buildPlatform`.
+
+        Ignored when `nixpkgs.pkgs` is set.
+      '';
+    };
+
+    buildPlatform = lib.mkOption {
+      type = with lib.types; either str attrs;
+      default = cfg.hostPlatform;
+      example = {
+        system = "x86_64-linux";
+      };
+      apply =
+        value:
+        let
+          elaborated = lib.systems.elaborate value;
+        in
+        # If equivalent to `hostPlatform`, make it actually identical so that `==` can be used
+        # See https://github.com/NixOS/nixpkgs/issues/278001
+        if lib.systems.equals elaborated cfg.hostPlatform then cfg.hostPlatform else elaborated;
+      defaultText = lib.literalMD ''
+        Inherited from the "host" configuration's `pkgs`.
+        Or `config.nixpkgs.hostPlatform` when building a standalone nixvim.
+      '';
+      description = ''
+        Specifies the platform on which Nixvim should be built.
+        By default, Nixvim is built on the system where it runs, but you can change where it's built.
+        Setting this option will cause Nixvim to be cross-compiled.
+
+        Ignored when `nixpkgs.pkgs` is set.
+      '';
+    };
+
     # NOTE: This is a nixvim-specific option; there's no equivalent in nixos
     source = lib.mkOption {
       type = lib.types.path;
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -2,11 +2,34 @@
   config,
   options,
   lib,
+  pkgs,
   ...
 }:
 let
   cfg = config.nixpkgs;
   opt = options.nixpkgs;
+
+  isConfig = x: builtins.isAttrs x || lib.isFunction x;
+
+  mergeConfig =
+    lhs_: rhs_:
+    let
+      optCall = maybeFn: x: if lib.isFunction maybeFn then maybeFn x else maybeFn;
+      lhs = optCall lhs_ { inherit pkgs; };
+      rhs = optCall rhs_ { inherit pkgs; };
+    in
+    lib.recursiveUpdate lhs rhs
+    // lib.optionalAttrs (lhs ? packageOverrides) {
+      packageOverrides =
+        pkgs:
+        optCall lhs.packageOverrides pkgs // optCall (lib.attrByPath [ "packageOverrides" ] { } rhs) pkgs;
+    }
+    // lib.optionalAttrs (lhs ? perlPackageOverrides) {
+      perlPackageOverrides =
+        pkgs:
+        optCall lhs.perlPackageOverrides pkgs
+        // optCall (lib.attrByPath [ "perlPackageOverrides" ] { } rhs) pkgs;
+    };
 in
 {
   options.nixpkgs = {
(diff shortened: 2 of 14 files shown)


modules/top-level/nixpkgs.nix
```diff
@@ -165,10 +275,21 @@ in
...
       # don't need to evaluate `finalPkgs`.
       _module.args.pkgs = lib.mkOverride lib.modules.defaultOverridePriority finalPkgs.__splicedPackages;
-
-      # FIXME: This is a stub option for now
-      warnings = lib.optional (
-        opt.source.isDefined && opt.source.highestPrio < (lib.mkOptionDefault null).priority
-      ) "Defining the option `nixpkgs.source` currently has no effect";
     };
+
+  assertions = [
```
khaneliman, line 280: does this need to go up a level into the {} above

MattSturgeon: Yeah, my bad. Just pushed a fix as you commented :sweat_smile:

modules/top-level/nixpkgs.nix
```diff
@@ -56,6 +56,29 @@ in
...
+    config = lib.mkOption {
+      default = { };
+      example = {
+        allowBroken = true;
+        allowUnfree = true;
+      };
+      type = lib.mkOptionType {
+        name = "nixpkgs-config";
+        description = "nixpkgs config";
+        check = x: if isConfig x then true else lib.traceSeqN 1 x false;
```
MattSturgeon, line 68:
```suggestion
        check = x: isConfig x || lib.traceSeqN 1 x false;
```

Is this better or worse?

khaneliman: Either is fine to me. Or condition maybe cleaner?

wrappers/_shared.nix
```diff
@@ -23,21 +23,35 @@ let
...
+  nixpkgsModule =
+    { config, ... }:
+    {
+      _file = ./_shared.nix;
+
+      config = {
+        nixpkgs = {
+          # Use global packages by default in nixvim's submodule
+          # FIXME: probably doesn't need to be mkForce
+          # What should happen if the user configures both `nixpkgs.pkgs` and `nixpkgs.useGlobalPackages` ?
```
traxys, line 40: I think that we should either raise a warning and ignore one of them (I don't know which ...) or raise an error

lib/tests.nix
```diff
@@ -30,7 +29,8 @@ let
   mkTestDerivationFromNixvimModule =
     {
       name ? null,
-      pkgs ? defaultPkgs,
+      pkgs ? null,
```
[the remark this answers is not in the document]
MattSturgeon: Yes, for backwards compatibility. I've tried to make this a non-breaking change.

IMO rather than deprecating specific args, it makes more sense to deprecate all functions in `lib/tests.nix` and `wrappers/standalone.nix`. That way they can be replaced by functions that have no baggage, do not need to be per-system in the flake output, and can more closely wrap the underlying `evalNixvim`.

I'm not ready to do that in this PR, and it's already large enough as it is. Before deprecating the per-system functions, I think we should move the "standalone package" into the module configuration; `modules/top-level/output.nix`.

modules/top-level/nixpkgs.nix
```diff
@@ -2,46 +2,46 @@
...
-      #     inherit (cfg) config overlays localSystem crossSystem;
-      #   }
-      # '';
-      defaultText = lib.literalMD ''
-        The `pkgs` inherited from your host config (i.e. NixOS, home-manager, or nix-darwin),
-        or the `pkgs` supplied to `makeNixvimWithModule` when building a standalone nixvim.
-
-        > [!CAUTION]
-        > This default will be removed in a future version of nixvim
-      '';
```
MattSturgeon, line 26: Should probably move this admonition to the `useGlobalPackages` option.

tests/main.nix
```diff
@@ -27,8 +28,10 @@ let
           inherit self system;
           inherit (self) inputs;
         };
+        nixpkgs.config = {
+          allowUnfree = true;
+        };
```
MattSturgeon, line 33: In a future PR, this should be moved to whichever tests actually require it.

tests/test-sources/modules/nixpkgs.nix
```diff
@@ -71,4 +74,25 @@
         }
       ];
     };
+
+  # Test that a nixpkgs revision can be specified using `nixpkgs.source`
+  source =
```
MattSturgeon, line 79: Let's also test that `config` has an effect

wrappers/modules/shared.nix
```diff
@@ -3,6 +3,19 @@
...
+    nixpkgs.useGlobalPackages = lib.mkOption {
+      type = lib.types.bool;
+      default = true; # TODO: Added 2024-12-23; switch to false one release after adding a deprecation warning
+      defaultText = lib.literalMD ''`true`, but will change to `false` in a future version.'';
+      description = ''
+        Whether Nixvim should use the host configuration's `pkgs` instance.
+        When false, `pkgs` will be constructed from `nixpkgs.source`.
+
+        > [!TIP]
+        > The host configuration is usually home-manager, nixos, or nix-darwin.
```
MattSturgeon, line 15: Instead of a generic "tip", this could actually use a context option to dynamically produce platform-specific docs. This is possible because we actually show the per-platform options separately.

Probably out of scope for this PR though.

MattSturgeon: Trivial with the existing `meta.config.name` option that was added for the per-platform option-docs.

lib/tests.nix
```diff
@@ -42,14 +42,18 @@ let
...
+        else
+          { nixpkgs.pkgs = lib.mkDefault pkgs; };
+
       result = helpers.modules.evalNixvim {
         modules = [
           module
           (lib.optionalAttrs (name != null) { test.name = name; })
           { wrapRc = true; }
-          # TODO: Only do this when `args?pkgs`
-          # Consider deprecating the `pkgs` arg too...
```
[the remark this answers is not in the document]
MattSturgeon: Same as https://github.com/nix-community/nixvim/pull/2738#discussion_r1896477703

modules/top-level/nixpkgs.nix
```diff
@@ -2,11 +2,34 @@
...
+      packageOverrides =
+        pkgs:
+        optCall lhs.packageOverrides pkgs // optCall (lib.attrByPath [ "packageOverrides" ] { } rhs) pkgs;
+    }
+    // lib.optionalAttrs (lhs ? perlPackageOverrides) {
+      perlPackageOverrides =
+        pkgs:
+        optCall lhs.perlPackageOverrides pkgs
+        // optCall (lib.attrByPath [ "perlPackageOverrides" ] { } rhs) pkgs;
+    };
```
traxys, line 32: I have not found the documentation for this override in the nixpkgs manual, you got it from upstream's config type? Should we leave it?

tests/nixpkgs-module.nix
```diff
@@ -28,7 +28,54 @@ let
...
+        ];
+      };
+
+      nixpkgs.hostPlatform = {
+        inherit (stdenv.hostPlatform) system;
+      };
+
+      assertions = [
+        {
+          assertion = pkgs.config.permittedInsecurePackages == [ "foobar123" ];
```
traxys, line 87: There would be no added value to try to actually install an insecure package right? If we observe the config value it should mean that the package would be installed?

MattSturgeon: Probably not. All this stuff is well tested in upstream nixpkgs too, so our tests are only really sanity checks to ensure the module isn't massively broken.

I've tried to make most of them lightweight unit tests rather than realistic integration tests.

modules/top-level/nixpkgs.nix
MattSturgeon: Yeah, this is taken from upstream: https://github.com/NixOS/nixpkgs/blob/c7f73ed1ad47009a95815327d478dba0db03c844/nixos/modules/misc/nixpkgs.nix#L16-L33

I assume it is there to cover some edge cases when merging two `nixpkgs.config` definitions. I.e. one def having a `perlPackageOverrides` _attrset_ and the other having a `perlPackageOverrides` _function_.

wrappers/modules/shared.nix
MattSturgeon: image
image
image

traxys (approved):

khaneliman:
Added some comments to matrix that I'll move here:

```bash
… while calling the 'intersectAttrs' builtin
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/lib/customisation.nix:219:17:
          218|       # This includes automatic ones and ones passed explicitly
          219|       allArgs = intersectAttrs fargs autoArgs // args;
             |                 ^
          220|

       … in the right operand of the update (//) operator
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/pkgs/top-level/splice.nix:152:54:
          151|
          152|   newScope = extra: lib.callPackageWith (pkgsForCall // extra);
          153|

       … in the left operand of the update (//) operator
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/lib/customisation.nix:649:32:
          648|       };
          649|       spliced = extra spliced0 // spliced0 // keep self;
             |                                ^
          650|       self = f self // {

       … from call site
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/lib/customisation.nix:649:17:
          648|       };
          649|       spliced = extra spliced0 // spliced0 // keep self;
             |                 ^
          650|       self = f self // {

       … while calling 'extra'
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/pkgs/top-level/darwin-packages.nix:48:11:
           47|   otherSplices = generateSplicesForMkScope "darwin";
           48|   extra = spliced: spliced.apple_sdk.frameworks;
             |           ^
           49|   f = lib.extends aliases (

       … while evaluating the attribute 'apple_sdk.frameworks'
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/pkgs/top-level/darwin-packages.nix:90:13:
           89|           inherit
           90|             apple_sdk
             |             ^
           91|             apple_sdk_10_12

       … while calling the 'throw' builtin
         at /nix/store/4hpdrd3qvj7nks3rrimqm2jdmcga8isc-source/pkgs/top-level/darwin-packages.nix:85:15:
           84|         .${stdenv.hostPlatform.darwinSdkVersion}
           85|           or (throw "Unsupported sdk: ${stdenv.hostPlatform.darwinSdkVersion}");
             |               ^
           86|

       error: Unsupported sdk: 11.3
```

The line of code is present in the sourced nixpkgs archive in the test

```nix
  nixpkgs.source = builtins.fetchTarball {
        url = "https://github.com/NixOS/nixpkgs/archive/1807c2b91223227ad5599d7067a61665c52d1295.tar.gz";
        sha256 = "0xnj86751780hg1zhx9rjkbjr0sx0wps4wxz7zryvrj6hgwrng1z";
      };
```

Is it a mismatch between the pkgs coming in for the test case and the nixpkgs.source being used in the test? Does it show an issue with the impelmentation? Or would that be an expected failure?

MattSturgeon:
Thanks for the help @khaneliman 

I've reduced flakyness by testing with a _mock_ nixpkgs source rather than downloading a real instance.

I've also narrowed the scope of the tests to focus on only the nixpkgs module and its dependencies. That way we aren't accidentally _using_ the pkgs instance elsewhere in unrelated modules.

The `testWrappers` tests still test against a full nixvim configuration with all modules, due to them using the actual wrapper modules.

traxys (approved):

MattSturgeon:
@mergify queue

PR #2740, opened 2024-12-24
tests/nixpkgs: move nixpkgs module test to dedicated drv

MattSturgeon:
We don't need a full test that actually _builds_ or _runs_ nixvim. All we need is some assertions on the result of a nixvim configuration.

Currently #2738 needs to change the module tests to construct their own `pkgs` instance, so that we can test how the module behaves. However this is likely to have a performance hit since we are repeatedly instanciating `pkgs` again for each test case. It's probably better to use a single `pkgs` instance for all test case modules, but move the nixpkgs-module tests to their own dedicated test derivation.

This PR does that, moving the existing tests to a dedicated link-farm.

This also allows creating some bespoke helpers to streamline writing the tests and _may_ improve performance slightly because we don't need to also build `config.build.package` for the nixpkgs-module unit tests.

The bespoke helpers are a little cumbersome, so I'm open to feedback on them. Some of them may also be useful more generally, so may be worth moving to some kinda "test utils" file. This can be done now or later.

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -43,6 +43,7 @@ in
   no-flake = callTest ./no-flake.nix { };
   lib-tests = callTest ./lib-tests.nix { };
   maintainers = callTest ./maintainers.nix { };
+  nixpkgs-module = callTest ./nixpkgs-module.nix { };
   plugins-by-name = callTest ./plugins-by-name.nix { };
   generated = callTest ./generated.nix { };
   package-options = callTest ./package-options.nix { };
--- a/tests/main.nix
+++ b/tests/main.nix
@@ -8,7 +8,6 @@
   pkgs,
   pkgsUnfree,
   self,
-  system,
 }:
 let
   fetchTests = callTest ./fetch-tests.nix { };
@@ -19,14 +18,12 @@ let
     file: name: module:
     mkTestDerivationFromNixvimModule {
       inherit name;
+      # Use a single common instance of nixpkgs, with allowUnfree
+      # Having a single shared instance should speed up tests a little
       module = {
         _file = file;
         imports = [ module ];
-        _module.args = {
-          # Give tests access to the flake
-          inherit self system;
-          inherit (self) inputs;
-        };
+        nixpkgs.pkgs = pkgsUnfree;
       };
       pkgs = pkgsUnfree;
     };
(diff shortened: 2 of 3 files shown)


MattSturgeon:
FYI https://github.com/nix-community/nixvim/pull/2738 is _kinda_ blocked on this. Or at least blocked on deciding whether we will merge this.

I'd like to re-write the tests in that PR based on this PR.

traxys:
I'd be in favour of introducing in this PR the framework for tests that don't launch nvim, it can be useful to add tests for assertions or warnings (#2751 reminded me that I wanted to add a diagnostic & such a test)

MattSturgeon:
> I'd be in favour of introducing in this PR the framework for tests that don't launch nvim, it can be useful to add tests for assertions or warnings (#2751 reminded me that I wanted to add a diagnostic & such a test)

That's still on my radar, but I don't see how it relates to _this_ PR

traxys:
Those tests don't require to launch nvim, so I guess what I am mostly saying is that it would be nice to create in this PR a test suite that does not eval nvim, and allows to easily add tests to it

MattSturgeon:
Ok, now that we've tackled negative tests can we revisit this PR and discuss whether we will merge it or something like it?

While we _could_ set `test.buildNixvim = false` and keep the nixpkgs module tests in `test-sources`, that'd require all the test-spurces modules to _not define_ `nixpkgs.pkgs`.

IDK how performance intensive instantiating an instance of nixpkgs is, or how well nix is able to cache similar instances for re-use...

Given the sheer number of test case modules we have, I feel it's probably a good idea to explicitly use the same `pkgs` instance for all of them.

If so, that means keeping the nixpkgs module tests separate, as per this PR (or similar).

traxys:
Yeah it addresses my comment, so I have no idea about the performance though :/

traxys (approved):

MattSturgeon:
I've re-worked this PR with a much nicer implementation.

It still achieves the same separation, but is much cleaner by using `config.build.test` instead of writing `runCommand` tests from scratch.

traxys (approved):
Indeed, much cleaner!

MattSturgeon:
@mergify queue

PR #2741, opened 2024-12-25
plugins/blink-cmp-copilot: init

HeitorAugustoLN:


plugins/by-name/blink-cmp-copilot/default.nix
```diff
@@ -0,0 +1,32 @@
...
+          completions properly appearing in blink-cmp-copilot.
+        ''
+      ]
+      ++ lib.optionals (isEnabled copilot-lua-cfg.panel.enabled) [
+        ''
+          It is recommended to disable copilot's `panel` module, as it can interfere with completions
+          properly appearing in blink-cmp-copilot.
+        ''
+      ];
+    plugins.copilot-lua.enable = lib.mkDefault true;
```
traxys, line 30: Why put it as `lib.mkDefault true` and not simply `true`? I'd think a user would be warned if they explicitely disable the plugin, I would prefer requiring `lib.mkForce`

plugins/by-name/blink-cmp-copilot/default.nix
```diff
@@ -0,0 +1,32 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "blink-cmp-copilot";
+  package = "blink-cmp-copilot";
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  callSetup = false;
+  hasLuaConfig = false;
+  hasSettings = false;
+
```
traxys, line 11: Could you add `description = '...'` that says that this plugin is configured through blink.cmp's `sources.providers` ?

plugins/by-name/blink-cmp-copilot/default.nix
HeitorAugustoLN: To be the same behavior as current copilot-cmp: https://github.com/nix-community/nixvim/blob/8938e09db14d510dcc2f266e8b2e738ee527d386/plugins/by-name/copilot-cmp/default.nix#L72

plugins/by-name/blink-cmp-copilot/default.nix
HeitorAugustoLN: Done

traxys (approved):

traxys:
@mergify queue

PR #2647, opened 2024-12-12
plugins/fidget: migrate to mkNeovimPlugin

HeitorAugustoLN:
- Migrated fidget.nvim to mkNeovimPlugin
- Added renames from old options to new ones (with the exception of the submodules, because I didn't find a way to)
- Added new options, etc.
- Updated tests to new format too

PR #2742, opened 2024-12-25
plugins/copilot-lua: migrate to mkNeovimPlugin

HeitorAugustoLN:


plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,269 @@
...
-          example = {
-            trace = "verbose";
-            settings = {
-              advanced = {
-                listCount = 10; # number of completions for panel
-                inlineSuggestCount = 3; # number of completions for getCompletions
-              };
-            };
-          };
+  settingsExample = {
```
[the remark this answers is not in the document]
HeitorAugustoLN: Made it shorter.

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,269 @@
...
+        next = keymapOption "<M-]>" "Keymap for accepting the next suggestion.";
+        prev = keymapOption "<M-[>" "Keymap for accepting the previous suggestion.";
+        dismiss = keymapOption "<C-]>" "Keymap to dismiss the suggestion.";
+      };
+    };
 
-          keymap = {
-            accept = keymapOption "<M-l>" "Keymap for accepting the suggestion.";
+    filetypes =
+      defaultNullOpts.mkAttrsOf types.bool
```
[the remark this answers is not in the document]
HeitorAugustoLN: Added examples in example key instead of description.

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,224 @@
...
+      keymap = {
+        accept_word = false;
+        accept_line = false;
       };
+    };
   };
 
-  config = mkIf cfg.enable {
+  deprecateExtraOptions = true;
+  optionsRenamedToSettings =
```
khaneliman, line 161: Didn't we have an example of how to move these out to a separate file recently?

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,224 @@
...
-
-            accept = keymapOption "<CR>" "Keymap to accept the proposed suggestion.";
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
 
-            refresh = keymapOption "gr" "Keymap to refresh the suggestions.";
-
-            open = keymapOption "<M-CR>" "Keymap to open.";
-          };
+  mkKeymapOption = defaultNullOpts.mkNullableWithRaw (with types; either (enum [ false ]) str);
```
khaneliman, line 11: Would have probably preferred this in a let block for the `settingsOptions`

plugins/by-name/copilot-lua/default.nix
khaneliman: Looks like https://github.com/nix-community/nixvim/pull/2647 is the one that had the change

plugins/by-name/copilot-lua/default.nix
HeitorAugustoLN: Done.

plugins/by-name/copilot-lua/default.nix
HeitorAugustoLN: Yeah, I forgot about that. Moved it to deprecations.nix

plugins/by-name/copilot-lua/default.nix
```diff
@@ -176,45 +169,7 @@ in
...
-            server_opts_overrides = serverOptsOverrides;
-          }
-          // cfg.extraOptions;
-      in
-      ''
-        require('copilot').setup(${lib.nixvim.toLuaObject setupOptions})
-      '';
   };
+
+  inherit (import ./deprecations.nix) deprecateExtraOptions optionsRenamedToSettings;
```
khaneliman, line 174: Looks like file didn't get checked in

khaneliman:
@mergifyio queue

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,165 @@
...
-          pluginDefault = { };
-          description = ''
-            Override copilot lsp client `settings`.
-            The settings field is where you can set the values of the options defined in
-            https://github.com/zbirenbaum/copilot.lua/blob/master/SettingsOpts.md.
-            These options are specific to the copilot lsp and can be used to customize its behavior.
+      copilot_node_command = lib.mkOption {
+        type = types.str;
+        default = lib.getExe pkgs.nodejs-18_x;
+        example = lib.literalExpression "lib.getExe pkgs.nodejs";
```
MattSturgeon, line 111: This needs a `defaultText`, otherwise we'll end up evaluating that package to render the docs. And the result won't look great anyway.

```suggestion
        default = lib.getExe pkgs.nodejs-18_x;
        defaultText = lib.literalExpression "lib.getExe pkgs.nodejs-18_x";
        example = lib.literalExpression "lib.getExe pkgs.nodejs";
```

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,165 @@
...
-        serverOptsOverrides = helpers.defaultNullOpts.mkAttrsOf' {
-          type = types.anything;
-          pluginDefault = { };
-          description = ''
-            Override copilot lsp client `settings`.
-            The settings field is where you can set the values of the options defined in
-            https://github.com/zbirenbaum/copilot.lua/blob/master/SettingsOpts.md.
-            These options are specific to the copilot lsp and can be used to customize its behavior.
+      copilot_node_command = lib.mkOption {
+        type = types.str;
```
MattSturgeon, line 109: Is it intended to have a non-nullable option in the freeform `settings` submodule? Usually we like all settings options to be nullable and default to null.

khaneliman: Thanks, didn't notice that.

MattSturgeon:
@mergify dequeue

plugins/by-name/copilot-lua/default.nix
HeitorAugustoLN: Well, I intended to mirror the previous behavior. And since it the plugin depends on node, I thought it would be fine to keep it as it was.

HeitorAugustoLN: I ended up putting back the default value again, it is now nullable, but I still think it should have a default value.

khaneliman: Freeform settings options should not be set automatically. They are supposed to be plugin defaults from upstream that a user can change.

MattSturgeon (approved):

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,167 @@
...
-          type = types.anything;
-          pluginDefault = { };
-          description = ''
-            Override copilot lsp client `settings`.
-            The settings field is where you can set the values of the options defined in
-            https://github.com/zbirenbaum/copilot.lua/blob/master/SettingsOpts.md.
-            These options are specific to the copilot lsp and can be used to customize its behavior.
+      copilot_node_command = lib.nixvim.mkNullOrOption' {
+        type = types.str;
+        default = lib.getExe pkgs.nodejs-18_x;
```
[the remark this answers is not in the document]
MattSturgeon: We discussed this briefly above https://github.com/nix-community/nixvim/pull/2742#discussion_r1904748878

Maybe a nullable package option would be better; I think the main reason it was done this way is because that's how it was before. But maybe having this de-coupled would be better.

e.g.

```nix
{
  settingsOptions.copilot_node_command = defaultNullOpts.mkStr' {
    defaultText = lib.literalMD ''
      `lib.getExe nodePackage` if `nodePackage` is not-null, otherwise `null`
    '';
    pluginDefault = "node";
    # ...
  };

  extraOptions.nodePackage = lib.mkPackageOption pkgs "nodejs" {
    default = [ "nodejs-18_x" ];
    example = [ "nodejs" ];
    nullable = true;
    extraDescription = ''
      If non-null, will provide a default for `settings.copilot_node_command`.
    '';
  };

  extraConfig = cfg: {
    plugins.copilot-lua.settings.copilot_node_command = lib.mkIf (cfg.nodePackage != null) (lib.mkDefault (
      lib.getExe cfg.nodePackage
    ));
  };
}
```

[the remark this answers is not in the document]
traxys: I think it's ok to have settings options not defaulting to null, this is something that happens quite often in nixpkgs where they are affected by the values of other modules, though I think there must be a good reason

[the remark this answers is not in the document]
MattSturgeon: > this is something that happens quite often in nixpkgs

Let's not forget that nixos is very different. Primarily because they usually don't do any "null culling" like we do in `toLuaObject`.

Therefore, if a sub-options exists in nixos, it is likely there for the express _purpose_ of providing a sane default.



> I think it's ok to have settings options not defaulting to null

I don't have a strong opinion on that.

Currently nixvim tries to avoid doing so. Where we do do so, we try to have it based on the value of a non-settings option that can be configured separately in an attempt to make things more modular & explicit.

---

In this case we've already agreed to setting the default, the discussion is now whether there should be a separate `nodePackage` option like we've been using recently.

I think having a `nodePackage` option makes it clearer to the end-user that we will be installing & configuring a node package.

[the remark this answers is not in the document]
HeitorAugustoLN: Should be fine now.

plugins/by-name/copilot-lua/default.nix
```diff
@@ -1,172 +1,181 @@
...
+      hide_during_completion = false;
+      debounce = 90;
+      keymap = {
+        accept_word = false;
+        accept_line = false;
+      };
+    };
+  };
+
+  extraOptions.nodePackage = lib.mkPackageOption pkgs "nodejs" {
```
MattSturgeon, line 169: To be clear, I'm fine with either.

But this is how we'd _normally_ format it, for extra clarity and to minimise diff size if additional options are added down the line

```suggestion
  extraOptions = {
    nodePackage = lib.mkPackageOption pkgs "nodejs" {
```

MattSturgeon (approved):
Thanks for being patient with us! :sweat_smile:

PR #2743, opened 2024-12-25
lib/mk-neovim-plugin: allow lazy-loading without `luaConfig`

HeitorAugustoLN:
Currently, lazy-loading only happens when the plugin has luaConfig, but it shouldn't be like that. Since some plugins such as `blink-cmp-copilot` can be set to be lazy-loaded with:
```nix
{
  plugins.blink-cmp.copilot  = {
    enable = true;
    lazyLoad.settings.lazy = true;
  };
}
```

and then only be loaded when `blink-cmp` loads with:

```nix
{
  plugins.blink-cmp = {
    enable = true;
    lazyLoad.settings = {
      before.__raw = ''
        require("lz.n").trigger_load("blink-cmp-copilot")
      '';
      event = "InsertEnter";
    };
  };
}
```


lib/plugins/mk-neovim-plugin.nix
```diff
@@ -139,7 +131,7 @@ let
                         after =
                           let
                             after = cfg.lazyLoad.settings.after or null;
-                            default = "function()\n " + cfg.luaConfig.content + " \nend";
+                            default = "function()\n " + lib.optionalString hasLuaConfig cfg.luaConfig.content + " \nend";
```
MattSturgeon, line 134: When hasLuaConfig is false, this will produce an empty function.

Therefore lazyloading it is pointless, right? At least when using lz.n, where only the "setup call" (i.e. hasLuaConfig) is done lazily.

MattSturgeon (commented):
I'm struggling to understand the motivation? With the current lz.n implementation this is pointless, right? So is the idea that this may become useful when we support lazy.nvim? Or am I missing something?

lib/plugins/mk-neovim-plugin.nix
HeitorAugustoLN: Gaetan's comment explains it. It is basically a way to make a the plugin load only when another plugin loads.

MattSturgeon: I still don't follow, because lz.n doesn't actually affect how plugins are _loaded_ from the rtp. My understanding is it _only_ controls when the "lua config"/"`setup` call" is run.

HeitorAugustoLN: It actually does, if the plugins are set to be optional, which currently is not done by nixvim. But I am going to do a PR doing this.

MattSturgeon: I've just read through the upstream README again 

https://github.com/nvim-neorocks/lz.n

I can't find an `optional` field mentioned. I see some talk of transitively loading dependencies, but it sounds to me like it's just talking about nvim's built-in loading mechanisms.

It's mentioned several times that lz.n itself does not manipulate the rtp at all. They even link to some examples that use other utilities like rtp.n.

HeitorAugustoLN: The `optional` field is not in `lz.n`, I was talking about `optional` attr when declaring plugins in `makeNeovimConfig`. So they are added to the `opt` instead of `start` in the `packpath`. If you take a look at the paq-nvim and home-manager examples, in both of them, the plugins are set to optional. In that way `lz.n` will then load the plugins with `packadd`

MattSturgeon: Ah, I see. I had missed that behaviour when reading the docs in the past.

For the record, here's the relevant upstream example:

```nix
programs.neovim = {
  enable = true;
  plugins = with pkgs.vimPlugins [
    lz-n
    {
      plugin = pkgs.vimPlugins.telescope-nvim;
      config = ''
        require("lz.n").load {
          "telescope.nvim",
          cmd = "Telescope",
        }
      '';
      type = "lua";
      optional = true;
    }
    {
      plugin = pkgs.vimPlugins.sweetie-nvim;
      config = ''
        require("lz.n").load {
          "sweetie.nvim",
          colorscheme = "sweetie",
        }
      '';
      type = "lua";
      optional = true;
    }
  ];
};
```

Makes me wonder why we aren't leveraging nixpkgs' vim plugin spec with config pattern; perhaps simply because it doesn't work well with `print-nixvim-init`? It does seem to offer much neater encapsulation.

HeitorAugustoLN: Also #2748 for making the plugins optional when lazy-loading

HeitorAugustoLN: > Makes me wonder why we aren't leveraging nixpkgs' vim plugin spec with config pattern

https://github.com/NixOS/nixpkgs/blob/a7c66bac5f126f2340b32f6ba5ed15d4b39d79e9/pkgs/applications/editors/neovim/wrapper.nix#L91

Looks like plugins configurations would prepend the rest of the init.lua, and we wouldn't really be able to set where it would be written in the init.lua

khaneliman (approved):
LGTM, seems like we have a few more tweaks that can be made to support more flexible lazy loading.

PR #2745, opened 2024-12-25
docs/contributing: fix hyperlink issues

RottenFishbone:
Fixes a couple issues I found while learning how to add a plugin.

direnv had a dead link to it (moved to `plugins/by-name/direnv/default.nix`)

lsp-lines linked to a commit with several changes, including the move to `plugins/by-name/lsp-lines/default.nix`, instead of the file itself.

I also added a template link immediately following the example code, as that is logically when someone would want it.

Something of note: most files are linked by reference but the two under `mkVimPlugin` are permalinks. Would it be preferable if all the files were linked as permalinks or that these are switched to relative?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ To add a new plugin you need to do the following.
 
 - _vim plugins_: They are configured through **global variables** (`g:plugin_foo_option` in vimscript and `vim.g.plugin_foo_option` in lua).\
   For those, you should use the `lib.nixvim.plugins.mkVimPlugin`.\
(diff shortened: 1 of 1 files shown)


MattSturgeon:
> > Something of note: most files are linked by reference but the two under `mkVimPlugin` are permalinks. Would it be preferable if all the files were linked as permalinks or that these are switched to relative?
> 
> I would say that relative links make more sense.

The problem is that we use this both directly on GitHub _and also_ use it in the documentation website. Therefore we actually need full URLs rather than relative URIs.

Maybe we could do some preprocessing in the docs build script to convert relative URIs to full URLs though?

PR #2748, opened 2024-12-26
lib/plugins: make plugin optional when lazy-loading

HeitorAugustoLN:
I thought nixvim was already doing this when lazy-loading were implemented, since it was in `lz.n` documentation. But few days ago, I noticed it was not being done. So this PR makes plugins optional when lazy-loading is enable, instead of just lazy-loading plugins when they require setup for loading, which some don't.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -46,6 +46,7 @@ lib.makeExtensible (
 
     inherit (self.options)
       defaultNullOpts
+      mkAutoLoadOption
       mkCompositeOption
       mkCompositeOption'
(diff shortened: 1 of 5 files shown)


MattSturgeon (approved):
LGTM. Thanks for enlightening me about optional plugins and the fact that lz.n handles them via `packadd`.

I'll leave this open in case anyone else has feedback, then merge tomorrow.

@GaetanLepage @traxys @khaneliman

khaneliman (approved):
LGTM, thanks

PR #2752, opened 2024-12-27
modules/test: allow arbitrary warning/assertion expecations

MattSturgeon:
- **modules/test: allow not building nixvim config**
- **modules/test: allow arbitrary warning/assertion expecations**
- **tests: use warning-expectations options**

Long overdue options for "expecting" warnings or assertions to match arbitrary predicates.

Common use-cases can de defined declaratively using an enum of pre-defined predicates, while more advanced use-cases can provide an arbitrary predicate function.

Also made it possible to write a test that does not build nixvim's "wrapped" nvim w/config, although I wonder if such tests are better off written manually using runCommand & evalNixvim?

### DIFF
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -94,12 +94,15 @@ in
     checkWarnings = lib.mkOption {
       type = lib.types.bool;
       description = "Whether to check `config.warnings` in the test.";
+      apply = x: lib.warnIfNot x "`test.checkWarnings = false` is replaced with `test.warnings = [ ]`." x;
       default = true;
     };
 
     checkAssertions = lib.mkOption {
       type = lib.types.bool;
       description = "Whether to check `config.assertions` in the test.";
+      apply =
+        x: lib.warnIfNot x "`test.checkAssertions = false` is replaced with `test.assertions = [ ]`." x;
       default = true;
     };
 
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -39,7 +39,14 @@ let
           default = null;
         };
         value = lib.mkOption {
-          type = lib.types.anything;
+          type =
+            if config.expect == null then
+              lib.types.unspecified
+              // {
+                description = "defined by `expect`";
+              }
+            else
+              namedPredicate.valueType;
           description = ''
             If defined, will be used together with `expect` to define `predicate`.
           '';
@@ -137,6 +144,12 @@ in
                 Expectation message supplier, matching `(value) -> String` or `(value) -> [(message)] -> String`.
               '';
             };
+            valueType = lib.mkOption {
+              type = lib.types.optionType;
+              description = ''
+                The type to use for `value` when this expectation is used.
+              '';
+            };
           };
         });
       description = ''
@@ -196,14 +209,17 @@ in
           count = {
             predicate = v: l: builtins.length l == v;
             message = v: l: "Expected length to be ${toString v} but found ${toString (builtins.length l)}.";
+            valueType = lib.types.ints.unsigned;
           };
           any = {
             predicate = v: builtins.any (lib.hasInfix v);
             message = v: "Expected ${lib.toJSON v} infix to be present.";
+            valueType = lib.types.str;
           };
           anyExact = {
             predicate = builtins.elem;
             message = v: "Expected ${lib.toJSON v} to be present.";
+            valueType = lib.types.str;
           };
         };
       };
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -19,10 +19,17 @@ in
       description = "The test derivation's name.";
     };
 
+    buildNixvim = lib.mkOption {
+      type = lib.types.bool;
+      description = "Whether to build the nixvim config in the test.";
+      default = true;
+    };
+
     runNvim = lib.mkOption {
       type = lib.types.bool;
       description = "Whether to run `nvim` in the test.";
-      default = true;
+      defaultText = lib.literalExpression "config.test.buildNixvim";
+      default = cfg.buildNixvim;
     };
 
     checkWarnings = lib.mkOption {
@@ -68,9 +75,12 @@ in
     in
     {
       build.test =
+        assert lib.assertMsg (cfg.runNvim -> cfg.buildNixvim) "`test.runNvim` requires `test.buildNixvim`.";
         pkgs.runCommandNoCCLocal cfg.name
           {
-            nativeBuildInputs = [ config.build.packageUnchecked ];
+            nativeBuildInputs = lib.optionals cfg.buildNixvim [
+              config.build.packageUnchecked
+            ];
 
             # Allow inspecting the test's module a little from the repl
             # e.g.
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -8,8 +8,60 @@
 let
   cfg = config.test;
 
-  inherit (config) warnings;
-  assertions = builtins.concatMap (x: lib.optional (!x.assertion) x.message) config.assertions;
+  # Expectation submodule used for checking warnings and/or assertions
+  expectationType = lib.types.submodule (
+    { config, ... }:
+    let
+      # NOTE: ensure `config.expect != null` before evaluating this!
+      namedPredicate = cfg.namedExpectationPredicates.${config.expect};
+    in
+    {
+      options = {
+        predicate = lib.mkOption {
+          type = with lib.types; functionTo bool;
+          description = ''
+            A predicate of that determines whether this expectation is met.
+
+            Type
+            ```
+            [String] -> Boolean
+            ```
+
+            Parameters
+            1. values - all warnings or matched assertions
+          '';
+        };
+        expect = lib.mkOption {
+          type = with lib.types; nullOr (enum (builtins.attrNames cfg.namedExpectationPredicates));
+          description = ''
+            If non-null, will be used together with `value` to define `predicate`.
+          '';
+          default = null;
+        };
+        value = lib.mkOption {
+          type = lib.types.anything;
+          description = ''
+            If defined, will be used together with `expect` to define `predicate`.
+          '';
+        };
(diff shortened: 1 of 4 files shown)


modules/top-level/test.nix
```diff
@@ -8,8 +8,72 @@
...
+          type = with lib.types; nullOr (enum (builtins.attrNames cfg.predefinedExpectationPredicates));
+          description = ''
+            If non-null, will be used together with `value` to define `predicate`.
+          '';
+          default = "any";
+        };
+        value = lib.mkOption {
+          type = lib.types.anything;
+          description = ''
+            If defined, will be used together with `assertion` to define `assertionFunction`.
```
traxys: assertion -> expect?
assertionFunction -> predicate?

modules/top-level/test.nix
```diff
@@ -8,8 +8,72 @@
...
+            Parameters
+            1. values - all warnings or matched assertions
+          '';
+        };
+        expect = lib.mkOption {
+          type = with lib.types; nullOr (enum (builtins.attrNames cfg.predefinedExpectationPredicates));
+          description = ''
+            If non-null, will be used together with `value` to define `predicate`.
+          '';
+          default = "any";
```
traxys, line 35: I'm not too keen on having a default value for this field, it requires remembering (or searching) for the default

modules/top-level/test.nix
```diff
@@ -8,8 +8,72 @@
...
+
+  expectationType =
+    with lib.types;
+    coercedTo (either str ints.unsigned) (
+      value:
+      {
+        inherit value;
+      }
+      // lib.optionalAttrs (builtins.isInt value) {
+        expect = "count";
```
traxys, line 81: For the same reason of the default I'm not too keen on having this coerction. I'd prefer we always specify what kind of assertion we want. If needs be we can define helper functions to reduce the boilerplate, like `expectCount = n: { expect = "count"; value = n; }`, or even `mkExpect = expect: value: { inherit expect value; }`;

modules/top-level/test.nix
```diff
@@ -19,23 +90,79 @@ in
...
+    };
+
+    assertions = lib.mkOption {
+      type = lib.types.listOf expectationType;
+      description = "A list of expectations for `assertions`.";
+      defaultText = lib.literalMD "Expect `count == 0`";
+      default = [ 0 ];
+    };
+
+    predefinedExpectationPredicates = lib.mkOption {
```
traxys, line 135: I think `predefined` is a bit of mis-nomer, a test could add more predicates as the option is not read-only. Maybe something like `named` or `registered`?

modules/top-level/test.nix
```diff
@@ -19,23 +90,79 @@ in
...
+      apply =
+        x: lib.warnIfNot x "`test.checkAssertions = false` is replaced with `test.assertions = [ ]`." x;
       default = true;
     };
+
+    warnings = lib.mkOption {
+      type = lib.types.listOf expectationType;
+      description = "A list of expectations for `warnings`.";
+      defaultText = lib.literalMD "Expect `count == 0`";
+      default = [ 0 ];
```
traxys, line 125: Why isn't the default merged with the values set by the test?
In particular why doesn't it result in : `[ { expect = "count"; value = 0;} { expect = "count"; value = 1;} { expect = "any", value ="The rust-tools project has been abandoned."; } ]` in one of the tests?

modules/top-level/test.nix
```diff
@@ -8,8 +8,79 @@
...
+        };
+        value = lib.mkOption {
+          type =
+            if config.expect == null then
+              lib.types.unspecified
+              // {
+                description = "defined by `expect`";
+              }
+            else
+              cfg.predefinedExpectationPredicates.${config.expect}.valueType;
```
traxys, line 45: That's some advanced depenedent type checking!

modules/top-level/test.nix
```diff
@@ -19,23 +90,79 @@ in
...
+              description = ''
+                Expectation message supplier, matching `(value) -> String` or `(value) -> [(message)] -> String`.
+              '';
+            };
+            valueType = lib.mkOption {
+              type = lib.types.optionType;
+              description = ''
+                The type to use for `value` when this expectation is used.
+              '';
+              default = lib.types.str;
```
traxys, line 157: Not sure if I agree with specifying a default type for this, I don't really see why a `str` is a "default" type for an assertion value

modules/top-level/test.nix
```diff
@@ -50,27 +177,61 @@ in
...
+        # If checkWarnings or checkAssertions are disabled, ensure the default expectations are overridden
+        assertions = lib.mkIf (!cfg.checkAssertions) [ ];
+        warnings = lib.mkIf (!cfg.checkWarnings) [ ];
+
+        # Predefined expectation predicates available via the `expect` enum-option
+        predefinedExpectationPredicates = {
+          count.predicate = v: l: builtins.length l == v;
+          count.message =
+            v: l: "Expected length to be ${toString v} but found ${toString (builtins.length l)}.";
+          count.valueType = lib.types.ints.unsigned;
```
traxys, line 218:
```suggestion
          count = {
              predicate = v: l: builtins.length l == v;
              message =
                v: l: "Expected length to be ${toString v} but found ${toString (builtins.length l)}.";
              valueType = lib.types.ints.unsigned;
        };
```

I find this a bit easier to read

modules/top-level/test.nix
```diff
@@ -50,27 +177,61 @@ in
...
+          any.predicate = v: builtins.any (lib.hasInfix v);
+          any.message = v: "Expected ${lib.toJSON v} infix to be present.";
+          anyExact.predicate = builtins.elem;
+          anyExact.message = v: "Expected ${lib.toJSON v} to be present.";
```
traxys, line 222: Same, but without the suggestion as I got lazy x)

traxys:
Thanks a lot for doing this! Maybe we should also document the predicates available by default in the CONTRIBUTING guide? Or at least link to the parts of the codebase where they are defined?

modules/top-level/test.nix
MattSturgeon: Only definitions of the same override-priority get merged together.

The default is prio 1500 (mkOptionDefault) while normal definitions are prio 100 (`lib.modules.defaultOverridePriority`)

This is intentional; the default list asserts that no warnings are present, but if you manually define expectations we don't want this.

modules/top-level/test.nix
MattSturgeon: I had a similar thought process but went in a different direction. I'd rather avoid introducing such helper functions as I feel `coercedTo` is somewhat self-documenting.

Although I agree type-dependent coercion could be kinda confusing. I think this can still be documented well enough? Not convinced either way.

modules/top-level/test.nix
MattSturgeon: I'd like this to to be better documented, we could actually fold `cfg.predefinedExpectationPredicates` into a bullet-point list of type-descriptions, but I figured that'd needlessly bloat an initial PR.

modules/top-level/test.nix
MattSturgeon: This was the default because it is the most commonly use. As it's an internal option I didn't think it was too important, but you're right it should probably be explicit.

MattSturgeon:
> Maybe we should also document the predicates available by default in the CONTRIBUTING guide? Or at least link to the parts of the codebase where they are defined?

I agree this'd be nice, but I'd rather any documentation for this be auto-generated.

It should already be documented _to an extend_ via the `expect` option's enum type. But I agree it could be better documented. IDK if I want to inflate this PR with that though :thinking:

modules/top-level/test.nix
traxys: Yeah I saw the intention, but I did not know enough of the module system to understand how it works

traxys:
There seems to be some infinite recursion:
```
error: derivation 'x86_64-linux.search' does not have valid outputs: error:
              … while calling the 'getAttr' builtin
                at <nix/derivation-internal.nix>:44:19:
                  43|       value = commonAttrs // {
                  44|         outPath = builtins.getAttr outputName strict;
                    |                   ^
                  45|         drvPath = strict.drvPath;
                at <nix/derivation-internal.nix>:34:12:
                  33|
                  34|   strict = derivationStrict drvAttrs;
                    |            ^
                  35|
              … while evaluating the default value of option `test.predefinedExpectationPredicates.<name>.valueType`
              … while evaluating an attribute `functor`
              … while evaluating an attribute `type`
              (1994 duplicate frames omitted)
              … while evaluating an attribute `_type`
              error: stack overflow; max-call-depth exceeded
              at /nix/store/8fwsiv0hd7nw1brkvka0jf1frk3m7qkr-source/lib/generators.nix:511:25:
                 510|       in
                 511|         if multiline && length lines > 1 then multilineResult else singlelineResult
                    |                         ^
                 512|     else if true  ==   v then "true"
                 ```

modules/top-level/test.nix
traxys: I quite dislike automatic coercions in most cases, I always prefer explicit ways, as I find myself always forgetting about coercions I rarely use

modules/top-level/test.nix
MattSturgeon: This is actually a bug; the default should be `null`

MattSturgeon:
> There seems to be some infinite recursion:

This is `lib.generators.toPretty` causing issues trying to pretty-print an option-type, since option-types have a recursive `functor`.

Providing a non-recursive `defaultText` or dropping the `default` as [suggested already](https://github.com/nix-community/nixvim/pull/2752#discussion_r1898582140) will resolve.

> `` while evaluating the default value of option `test.predefinedExpectationPredicates.<name>.valueType` ``

modules/top-level/test.nix
traxys: Then the str coercion (if kept) will need to be updated right?

traxys (approved):

MattSturgeon:
@mergify queue

PR #2755, opened 2024-12-27
tests: omit `default` suffix in test names

MattSturgeon:
If the test file is `default.nix`, we don't need to append the filename
to the test namespace.


MattSturgeon:
@mergify queue

PR #2756, opened 2024-12-27
modules/test: document possible types for expectation `value`

MattSturgeon:
Small followup to #2752, kept separate to avoid further bloating that PR

image


traxys (approved):

MattSturgeon:
@mergify queue

PR #2758, opened 2024-12-27
docs: treat internal options as invisible

MattSturgeon:
`internal` options are intended to be implicitly not `visible`. This is already handled correctly by the nixos tooling, but our custom tooling did not correctly handle the edge case.

This lead to strange cases where an internal option is not included in the docs, but its sub-options were still visible.

For example, the new internal option `test.namedExpectationPredicates` is not in the docs, but its sub-options are:

<details><summary>Without this PR</summary>
<p>

image

</p>
</details> 

<details><summary>With this PR</summary>
<p>

image

</p>
</details>

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -41,9 +41,9 @@ let
   isVisible =
     opts:
     if lib.isOption opts then
-      opts.visible or true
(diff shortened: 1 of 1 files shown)


traxys (approved):

MattSturgeon:
@mergify queue

PR #2759, opened 2024-12-27
modules/test: hide the deprecated `check*` options

MattSturgeon:
TBH these should probably be removed entirely and/or throw rather than warn, but for now we should at least hide them from the public docs.


traxys (approved):

MattSturgeon:
@mergify queue

PR #2761, opened 2024-12-27
plugins/blink-cmp: force version for blink.cmp >= 0.8.2

HeitorAugustoLN:
When using nix to install blink-cmp plugin version 0.8.2 or higher, the plugin fails to load because it cannot find the git hash or tag used during build. So we force define the current version based on the current package version.

### DIFF
--- a/plugins/by-name/blink-cmp/default.nix
+++ b/plugins/by-name/blink-cmp/default.nix
@@ -162,5 +162,11 @@ lib.nixvim.plugins.mkNeovimPlugin {
     warnings = lib.optional (cfg.settings ? documentation) ''
       Nixvim(plugins.blink): `settings.documentation` does not correspond to a known setting, use `settings.windows.documentation` instead.
     '';
+
+    # After version 0.8.2, if we don't force the version, the plugin return an error after being loaded.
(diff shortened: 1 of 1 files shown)


khaneliman:
I was thinking about this yesterday and how it was best to approach it. I was thinking we'd rather we upstream a patch for the nixpkg derivation.

MattSturgeon (approved):
Defining `settings` options really goes against nixvim's principles. We should try to find a better solution upstream in blink and/or nixpkgs. But as a temporary solution I'm fine with this.

khaneliman:
> Defining `settings` options really goes against nixvim's principles. We should try to find a better solution upstream in blink and/or nixpkgs. But as a temporary solution I'm fine with this.

We should at least make it a `mkDefault` priority, no?

traxys:
> Defining `settings` options really goes against nixvim's principles. We should try to find a better solution upstream in blink and/or nixpkgs. But as a temporary solution I'm fine with this.

I'm not entirely sure about that, I'd say it's OK for us to set "real" default values if the plugin requires them for some reason (path to an executables, special values due to nixvim, ...)
Though I agree sometimes the right place to fix this is in nixpkgs

HeitorAugustoLN:
NixOS/nixpkgs#368677 should fix it.

PR #2773, opened 2024-12-31
Replace runCommandNoCCLocal alias with runCommandLocal

SuperSandro2000:


traxys:
I don't know enough about Nix to know why this change is useful, could you add a commit description?

khaneliman:
If they are actually aliases instead of the real command, some users will disable allowing aliases in their configs.

khaneliman (approved):
LGTM, using the actual command instead of the alias per https://github.com/NixOS/nixpkgs/blob/27c1df61f71d637fdfddb17388776294c0fbeb0c/pkgs/top-level/aliases.nix#L1170-L1171 to support https://github.com/NixOS/nixpkgs/blob/2f6435980795cc253d84fe4956a2497b8f14d4b5/pkgs/top-level/config.nix#L77-L93

khaneliman:
@mergifyio queue

SuperSandro2000:
> If they are actually aliases instead of the real command, some users will disable allowing aliases in their configs so it would improve compatibility with more configurations. Similar to removing IFD

Yep, I have aliases turned off  to not accidentally using them when creating PRs.

PR #2774, opened 2024-12-31
plugins/lazydev: init

HeitorAugustoLN:
Adds [lazydev.nvim](https://github.com/folke/lazydev.nvim) to nixvim

### DIFF
--- a/plugins/by-name/lazydev/default.nix
+++ b/plugins/by-name/lazydev/default.nix
@@ -1,0 +1,167 @@
+{ config, lib, ... }:
+let
+  inherit (lib) types;
+  inherit (lib.nixvim)
+    defaultNullOpts
+    literalLua
+    mkNullOrOption'
+    nestedLiteralLua
+    ;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lazydev";
+  packPathName = "lazydev.nvim";
+  package = "lazydev-nvim";
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  description = ''
+    ### lazydev.nvim as a blink.cmp source
+
+    ```nix
+    {
+      plugins = {
+        lazydev.enable = true;
+
+        blink-cmp.settings = {
+          sources.providers = {
+            lazydev = {
+              name = "LazyDev";
+              module = "lazydev.integrations.blink";
+              # make lazydev completions top priority (see `:h blink.cmp`)
+              score_offset = 100;
+            };
+          };
+        };
+      };
+    }
+    ```
+  '';
+
+  settingsOptions =
+    let
+      libraryType = types.submodule {
+        freeformType = with types; attrsOf anything;
+        options = {
+          files = mkNullOrOption' {
+            type = with types; listOf str;
+            example = [ "xmake.lua" ];
+            description = "A list of files that triggers the library to be loaded when required.";
+          };
+          mods = mkNullOrOption' {
+            type = with types; listOf str;
+            example = [ "wezterm" ];
+            description = "A list of modules that triggers the library to be loaded when required.";
+          };
+          path = lib.mkOption {
+            type = types.str;
+            example = "~/projects/my-awesome-lib";
+            description = "The path to the library.";
+          };
+          words = mkNullOrOption' {
+            type = with types; listOf str;
+            example = [
+              "wezterm"
+              "LazyVim"
+            ];
+            description = "A list of words that triggers the library to be loaded.";
+          };
+        };
+      };
+    in
+    {
+      enabled =
+        defaultNullOpts.mkBool
+          (literalLua ''
+            function(root_dir)
+              return vim.g.lazydev_enabled == nil and true or vim.g.lazydev_enabled
+            end
+          '')
+          ''
+            Whether lazydev.nvim is enabled.
+          '';
+
+      integrations = {
+        cmp = defaultNullOpts.mkBool true ''
+          Whether lazydev.nvim should integrate with nvim-cmp.
+
+          Adds a completion source for:
+            - `require "modname"`
+            - `---@module "modname"`
+        '';
+
+        coq = defaultNullOpts.mkBool false ''
+          Whether lazydev.nvim should integrate with coq.nvim.
+
+          Adds a completion source for:
+            - `require "modname"`
+            - `---@module "modname"`
+        '';
+
+        lspconfig = defaultNullOpts.mkBool true ''
+          Whether lazydev.nvim should integrate with nvim-lspconfig.
+
+          Fixes lspconfig's workspace management for LuaLS.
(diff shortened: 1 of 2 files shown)


plugins/by-name/lazydev/default.nix
```diff
@@ -0,0 +1,134 @@
...
+          function(root_dir)
+            return vim.g.lazydev_enabled == nil and true or vim.g.lazydev_enabled
+          end
+        '')
+        ''
+          Whether lazydev.nvim is enabled.
+        '';
+
+    integrations = {
+      cmp = defaultNullOpts.mkBool true ''
```
khaneliman, line 83: We could benefit from adding some warnings when a user enables any of these and checking if they've enabled the corresponding module.

HeitorAugustoLN: Don't really think this is needed, `lazydev.nvim` only loads the integration if the module of what it is integrating is available

HeitorAugustoLN: I added warnings, since I thought more about it, and it makes sense.

plugins/by-name/lazydev/default.nix
```diff
@@ -0,0 +1,162 @@
...
+          ]
+      ++
+        lib.optionals
+          (
+            builtins.isBool cfg.settings.integrations.coq && cfg.settings.integrations.coq
+            -> config.plugins.coq-nvim.enable
+          )
+          [
+            "Nixvim(plugins.lazydev): you have enabled coq integration but plugins.coq-nvim is not enabled."
+          ];
```
traxys, line 160:
```
              … while evaluating definitions from `/nix/store/0f0hsgi0zsnqq4xw3jw076b33xc7ndxq-source/modules/top-level/test.nix':
              … while evaluating definitions from `/nix/store/0f0hsgi0zsnqq4xw3jw076b33xc7ndxq-source/modules/top-level/test.nix':
              … while evaluating definitions from `/nix/store/0f0hsgi0zsnqq4xw3jw076b33xc7ndxq-source/plugins/by-name/lazydev':
              error: attribute 'integrations' missing
              at /nix/store/0f0hsgi0zsnqq4xw3jw076b33xc7ndxq-source/plugins/by-name/lazydev/default.nix:138:26:
                 137|       lib.optionals
                 138|         (builtins.isBool cfg.integrations.cmp && cfg.integrations.cmp -> config.plugins.cmp.enable)
                    |                          ^
                 139|         [ "Nixvim(plugins.lazydev): you have enabled nvim-cmp integration but plugins.cmp is not enabled." ]
```

You need to check if `cfg.settings ? integrations`, then `cfg.settings.integrations ? <xxx>`. Maybe there's a short way, I don't know the helpers well

HeitorAugustoLN: This error should be fixed after: https://github.com/nix-community/nixvim/compare/105f64b315821699c645aa74a833531eca14bccf..0896a2dcbbe7ae5d18b542575bd4e8df4c611bcf

MattSturgeon: I don't think that addresses the underlying issue.

`builtins.isBool cfg.settings.integrations.coq` can fail because `cfg.settings` is freeform, and `integrations` (etc) is not guaranteed to be present.

Therefore we either need some `or` alternation, or `?` checks.

I would use `builtins.isBool (cfg.settings.integrations.coq or null) for a safe check.

---

The other way to avoid this would be to declare a settings sub-option. Because actual sub-options are not freeform, they are guaranteed to exist in `cfg.settings.`

MattSturgeon: Also, could be me, but is the logic for these warnings inverted? Remember, the condition being true means the warning will show (unlike assertions, which are the other way around)! :sob:

HeitorAugustoLN: I think I fixed it now, I never thought i would struggle with warnings lol

HeitorAugustoLN: My logic on the conditions weren't right, now I think it is.

MattSturgeon: > `builtins.isBool cfg.settings.integrations.coq` can fail because `cfg.settings` is freeform, and `integrations` (etc) is not guaranteed to be present.
>
> The other way to avoid this would be to declare a settings sub-option. Because actual sub-options are not freeform, they are guaranteed to exist in `cfg.settings.`

Ah, my mistake, it _is_ declared as a sub-option.



> My logic on the conditions weren't right, now I think it is.
> I never thought i would struggle with warnings lol

I always get mixed up whenever there's two styles of negative and positive conditions :sweat_smile:

plugins/by-name/lazydev/default.nix
```diff
@@ -0,0 +1,165 @@
...
+
+  extraConfig = cfg: {
+    warnings =
+      lib.optionals
+        (
+          builtins.isBool cfg.settings.integrations.cmp
+          && !config.plugins.cmp.enable
+          && cfg.settings.integrations.cmp
+        )
+        [ "Nixvim(plugins.lazydev): you have enabled nvim-cmp integration but plugins.cmp is not enabled." ]
```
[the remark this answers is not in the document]
HeitorAugustoLN: At least on blink-cmp there is not:

https://github.com/nix-community/nixvim/blob/0ebc64a2328fc0a0532f9544eb6c6e297135962e/plugins/by-name/blink-cmp/default.nix#L45

plugins/by-name/lazydev/default.nix
```diff
@@ -0,0 +1,165 @@
...
+      lspconfig = defaultNullOpts.mkBool true ''
+        Whether lazydev.nvim should integrate with nvim-lspconfig.
+
+        Fixes lspconfig's workspace management for LuaLS
+        Only create a new workspace if the buffer is not part
+        of an existing workspace or one of its libraries.
+      '';
+    };
+
+    library = defaultNullOpts.mkNullableWithRaw (with types; listOf (either str libraryType)) [ ] ''
```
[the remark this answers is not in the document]
HeitorAugustoLN: It can!

plugins/by-name/lazydev/default.nix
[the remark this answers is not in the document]
MattSturgeon: I don't know which formatting we prefer, but we're definitely not very consistent about it currently! 🙈

I'll hold my hands up to contributing some of the inconsistency on this, too.

If we really want to be consistent, we probably need some systematic method of achieving that...

plugins/by-name/lazydev/default.nix
khaneliman: I always have to think way too hard with warnings vs assertions...

MattSturgeon (approved):

khaneliman (approved):

khaneliman:
@mergify queue

PR #2776, opened 2025-01-01
plugins/flit: init

jolars:


jolars:
Great, thanks for the edits! After reading the documentation I was unsure about whether you would prefer options to be set explicitly or not through the nixvim interface, but I suppose then that you want options to be defined explicitly?

PR #2791, opened 2025-01-04
plugins/project-nvim: deprecate

alisonjenkins:
Deprecate project-nvim as it hasn't been updated in a long time and it's functionality is breaking on recent Neovim versions.

### DIFF
--- a/plugins/deprecation.nix
+++ b/plugins/deprecation.nix
@@ -6,6 +6,13 @@ let
       The `rust-tools` project has been abandoned.
       It is recommended to use `rustaceanvim` instead.
     '';
+    # TODO: added 2025-01-04, move to removed after 25.11
+    "project-nvim" = ''
+      The project-nvim project has not been updated in 2 years
+      and is broken on recent Neovim versions (it no longer adds projects
+      automatically and provides no manual method to add them) use either
+      telescope-project or projections-nvim.
+    '';
   };
   removed = {
     # Added 2023-08-29
--- a/plugins/deprecation.nix
+++ b/plugins/deprecation.nix
@@ -1,7 +1,7 @@
 { lib, ... }:
 let
   deprecated = {
-    # TODO: added 10-23-2024, move to removed after 24.11
+    # TODO: added 2024-10-23, move to removed after 24.11
     "rust-tools" = ''
       The `rust-tools` project has been abandoned.
       It is recommended to use `rustaceanvim` instead.


plugins/deprecation.nix
```diff
@@ -1,11 +1,18 @@
 { lib, ... }:
 let
   deprecated = {
-    # TODO: added 10-23-2024, move to removed after 24.11
+    # TODO: added 2024-10-23, move to removed after 24.11
     "rust-tools" = ''
       The `rust-tools` project has been abandoned.
       It is recommended to use `rustaceanvim` instead.
     '';
+    # TODO: added 2025-01-04, move to removed after 24.11
```
khaneliman, line 9:
```suggestion
    # TODO: added 2025-01-04, move to removed after 25.05
```

khaneliman:
What version does this not work in? I still have this plugin, but haven't used it in a while.

plugins/deprecation.nix
alisonjenkins: Thanks fixed.

khaneliman:
Doesn't seem to be the case for me.
image

khaneliman:
> @khaneliman could you see if it works in this for you?
> 
> ```
> nix run github:alisonjenkins/neovim-nix-flake/9b5846e53627846e19c456993926643349d082fa
> ```
> 
> This is the last commit of my Nixvim before I removed it in: [alisonjenkins/neovim-nix-flake@c5fd2f7](https://github.com/alisonjenkins/neovim-nix-flake/commit/c5fd2f74e8048a5fa6aeb4469b7461690a50bab0)

Yeah, it doesn't generate it from your config but as soon as I entered with mine it did.

plugins/deprecation.nix
```diff
@@ -1,11 +1,18 @@
...
     "rust-tools" = ''
       The `rust-tools` project has been abandoned.
       It is recommended to use `rustaceanvim` instead.
     '';
+    # TODO: added 2025-01-04, move to removed after 25.11
+    "project-nvim" = ''
+      The project-nvim project has not been updated in 2 years
+      and is broken on recent Neovim versions (it no longer adds projects
+      automatically and provides no manual method to add them) use either
+      telescope-project or projections-nvim.
```
MattSturgeon, line 14: Nit: this isn't markdown, so we can't wrap lines and assume they'll be formatted automatically.

Instead we have to manually wrap the lines as we want them to appear in the console output.

Issue #2346, opened 2024-09-30
[PLUGIN REQUEST] neoconf.nvim


GGORG0:
| Field    | Description |
|----------|-------------|
| Plugin   | `neoconf.nvim`  | 
| Homepage | https://github.com/folke/neoconf.nvim     | 
| Nixpkgs  | https://github.com/NixOS/nixpkgs/blob/4e0c36e4dd53f35d5a6385bdae88895ec5832f70/pkgs/applications/editors/vim/plugins/generated.nix#L7314 | 

## Extra Information

This plugin allows you to override some configuration options using config files in projects (kinda like with `.vscode/settings.json`). It might be a bit tricky to implement in Nixvim though, especially the fact that:

> It's important that you set up neoconf.nvim BEFORE nvim-lspconfig.

( https://github.com/folke/neoconf.nvim#-setup )

bkp5190:
Opened https://github.com/NixOS/nixpkgs/pull/354673. Requires lspconfig from upstream nixpkgs to be defined, but unsure how to handle the "BEFORE" dependency.

FirelightFlagboy:
Hello, I'm using this plugin but I'm getting the following error:

```
Setup `require("neoconf").setup()` should be run **BEFORE** setting up any lsp server with lspconfig
```

And if I inspect the generated `init.lua` the lsp is configured before `neoconf`.

Did you found a workaround for the ordering problem?

PR #2807, opened 2025-01-08
plugins/neoconf: setup before lspconfig

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2346


khaneliman:
@mergify queue

PR #2841, opened 2025-01-17
lib/modules: clean `specialArgs`: drop `lib`, add `modulesPath`

MattSturgeon:
`lib.evalModules` will automatically add `lib` to the module args internally. Doing it explicitly is redundant.

`specialArgs.modulesPath` can be used to allow users to manually import modules relative to our `./modules` directory.
It is also used by the module system to provide a base path for relative `disabledModules` paths.

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -36,7 +36,6 @@ in
         }
       ];
       specialArgs = {
-        inherit lib;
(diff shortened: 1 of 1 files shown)


MattSturgeon:
@mergify queue

PR #2843, opened 2025-01-17
modules/test: provide access to `expect` function

MattSturgeon:
Allow `test.warnings` and `test.assertions` to be defined as either a list, or a function coerced to a list.

When defined as a function, it is supplied an `expect` function which provides some syntactic-sugar for defining simple expectations.

This is an alternative to the current approach of defining that `expect` function on an ad-hoc basis.

I prefer this to adding `expect` to nixvim's lib because:
1. That would require having access to `lib`
2. IDK where in `lib` such a specialized function should live

Note: this can also be used in #2738

image

### DIFF
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -76,6 +76,19 @@ let
       };
     }
   );
+
+  expectationListType =
+    let
+      fnType = lib.mkOptionType {
+        name = "function";
+        description = "function";
+        descriptionClass = "noun";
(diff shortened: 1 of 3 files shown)


khaneliman (approved):
Very nice, thank you

MattSturgeon:
@mergify queue

MattSturgeon:
@mergify refresh

MattSturgeon:
@mergify queue

PR #2848, opened 2025-01-17
lib/utils: add mkWarnings

GaetanLepage:
This aims to provide more consistency for our warnings.

- **lib/utils: add mkWarnings**
- **treewide: use mkWarnings wherever possible**

Before:
```nix
warnings = lib.optional COND ''
  Nixvim (plugins.foo): MESSAGE
'';
```
After:
```nix
warnings = lib.nixvim.mkWarnings "plugins.foo" {
  condition = COND; # Or `cond = `, to be discussed
  message = "MESSAGE";
};
```

### Notes:
- The second argument of `mkWarnings` can either be a single _conditional warning_ or a list of several ones.
- A _conditional warning_ is either a string (i.e. non-conditional) or a `{ condition = ...; message = ...; }` attrs.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -86,6 +86,7 @@ lib.makeExtensible (
       mkIfNonNull'
       mkRaw
       mkRawKey
+      mkWarnings
       nestedLiteral
       nestedLiteralLua
       override
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -158,6 +158,7 @@ rec {
     ```
 
     # Type
+
     ```
     literalLua :: String -> AttrSet
     ```
@@ -171,6 +172,34 @@ rec {
       exp = "lib.nixvim.mkRaw " + builtins.toJSON raw.__raw;
     in
     lib.literalExpression exp;
+
+  /**
+    Convert one or several conditional warnings to a final warning list.
+    The second argument can either be a list of _conditional warnings_ or a single one.
+
+    # Example
+
+    ```nix
+    warnings = mkWarnings "plugins.foo" {
+      when = plugins.foo.settings.barIntegration && (!plugins.bar.enable);
+      message = "`barIntegration` is enabled but the `bar` plugin is not."
+    }
+    ```
+
+    # Type
+
+    ```
+    mkWarnings :: String -> List -> List
+    ```
+  */
+  mkWarnings =
+    scope: warnings:
+    let
+      processWarning =
+        warning:
+        lib.optional (warning.when or true) "Nixvim (${scope}): ${lib.trim (warning.message or warning)}";
+    in
+    builtins.concatMap processWarning (lib.toList warnings);
 
   /**
     Convert the given string into a `__pretty` printed mkRaw expression.
@@ -190,6 +219,7 @@ rec {
     ```
 
     # Type
+
     ```
     nestedLiteralLua :: String -> AttrSet
     ```
@@ -221,6 +251,7 @@ rec {
     ```
 
     # Type
+
     ```
     nestedLiteral :: (String | literalExpression) -> AttrSet
     ```
--- a/plugins/cmp/auto-enable.nix
+++ b/plugins/cmp/auto-enable.nix
@@ -58,22 +58,29 @@ in
   config = lib.mkIf (cfg.enable && cfg.autoEnableSources) (
     lib.mkMerge [
       {
-        warnings =
+        warnings = lib.nixvim.mkWarnings "plugins.cmp" [
           # TODO: expand this warning to ft & cmd sources lists and `showDefs` the offending definitions
-          lib.optional (lib.types.isRawType cfg.settings.sources) ''
-            Nixvim (plugins.cmp): You have enabled `autoEnableSources` that tells Nixvim to automatically
-            enable the source plugins with respect to the list of sources provided in `settings.sources`.
-            However, the latter is proveded as a raw lua string which is not parseable by Nixvim.
+          {
+            when = lib.types.isRawType cfg.settings.sources;
+            message = ''
+              You have enabled `autoEnableSources` that tells Nixvim to automatically
+              enable the source plugins with respect to the list of sources provided in `settings.sources`.
+              However, the latter is proveded as a raw lua string which is not parseable by Nixvim.
 
-            If you want to keep using raw lua for defining your sources:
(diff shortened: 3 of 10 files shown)


khaneliman:
I like it just to be consistent treewide. Error is because you didn't inherit it in the lib/default.nix to get it available at top-level .

khaneliman:
> > I like it just to be consistent treewide.
> 
> Yes I plan to do it in this very PR. Was just waiting for feedback first ;)

Sorry, that was what I was trying to convery. I liked it to just get that consistency treewide without having to think about what our current warning formatting is and we only have to update it in one place to affect everywhere.

lib/utils.nix
```diff
@@ -172,6 +172,40 @@ rec {
...
+
+    # Type
+    ```
+    mkWarnings :: String -> List -> List
+    ```
+  */
+  mkWarnings =
+    scope: w:
+    let
+      warnings = if !lib.isList w then [ w ] else w;
```
MattSturgeon, line 197: Can be simplified to:
```suggestion
      warnings = lib.toList w;
```

lib/utils.nix
```diff
@@ -172,6 +172,41 @@ rec {
...
+      processWarning =
+        warning:
+        lib.optional (warning.condition or true) (
+          let
+            message = if lib.isString warning then warning else warning.message;
+          in
+          "Nixvim (${scope}): ${message}"
+        );
+    in
+    lib.flatten (map processWarning warnings);
```
MattSturgeon, line 208:
```suggestion
    builtins.concatMap processWarning (lib.toList w);
```

Lets not use a seldgehammer (`flatten`) to drive in a nail (list of lists).

Also, I think we can inline `warnings` here without loosing any readability.

lib/utils.nix
```diff
@@ -172,6 +172,41 @@ rec {
...
+  mkWarnings =
+    scope: w:
+    let
+      warnings = if !lib.isList w then [ w ] else w;
+
+      processWarning =
+        warning:
+        lib.optional (warning.condition or true) (
+          let
+            message = if lib.isString warning then warning else warning.message;
```
MattSturgeon, line 203:
```suggestion
            message = warning.message or warning;
```
?

MattSturgeon: Also: this would be a great place to strip trailing whitespace, e.g. using [`lib.trim`](https://noogle.dev/f/lib/strings/trim)

lib/utils.nix
```diff
@@ -172,6 +172,41 @@ rec {
...
+          in
+          "Nixvim (${scope}): ${message}"
```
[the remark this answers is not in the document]
MattSturgeon:
```suggestion
          "Nixvim (${scope}): ${lib.trim (warning.message or warning)}"
```

lib/default.nix
```diff
@@ -82,6 +82,7 @@ lib.makeExtensible (
       mkIfNonNull'
       mkRaw
       mkRawKey
+      mkWarnings
```
MattSturgeon, line 85: Can we add a test to `tests/lib-test.nix`?

lib/utils.nix
```diff
@@ -172,6 +172,41 @@ rec {
...
+    ```
+  */
+  mkWarnings =
+    scope: w:
+    let
+      warnings = if !lib.isList w then [ w ] else w;
+
+      processWarning =
+        warning:
+        lib.optional (warning.condition or true) (
```
MattSturgeon, line 201: Q: should the condition be _inverted_ so that it matches assertion-style logic?

I always get confused when assertions have negative conditions and warnings have positive conditions.

Although "normal" warnings having positive conditions while nixvim-warnings having negative conditions would _also_ be confusing.

lib/utils.nix
```diff
@@ -172,6 +172,41 @@ rec {
...
+    ```nix
+    warnings = mkWarnings [
+      {
+        condition = plugins.foo.settings.barIntegration && (!plugins.bar.enable);
+        message = "`barIntegration` is enabled but the `bar` plugin is not."
+      }
+    ];
+    ```
+
+    # Type
```
MattSturgeon, line 189:
```suggestion
    # Type

```
Nit

lib/utils.nix
MattSturgeon: One way we could solve this is by having two well-known condition attrs: e.g. `expect` and `unless` or `when` and `whenNot`, etc

khaneliman: > One way we could solve this is by having two well-known condition attrs: e.g. `expect` and `unless` or `when` and `whenNot`, etc

I kinda like this... if we're going to create our own helper might as well make it feel more robust. I think we either have it follow upstream if just a plain condition or allow defining both postive/negative cases.

MattSturgeon: Could do something like:
```nix
processCondition = warning: warning.when or !(warning.expect or false);
```
?

lib/utils.nix
```diff
@@ -172,6 +173,35 @@ rec {
     in
     lib.literalExpression exp;
 
+  /**
+    Convert one or several conditional warnings to a final warning list.
+
+    # Example
+
+    ```nix
+    warnings = mkWarnings [
```
MattSturgeon, line 182:
```suggestion
    warnings = mkWarnings "plugins.foo" [
```

> ``error: expected a set but found a function: «lambda mkWarnings @ lib/utils.nix:197:5»``

plugins/cmp/auto-enable.nix
```diff
@@ -58,22 +58,29 @@ in
   config = lib.mkIf (cfg.enable && cfg.autoEnableSources) (
     lib.mkMerge [
       {
-        warnings =
+        warnings = lib.nixvim.mkWarnings.mkWarnings "plugins.cmp" [
```
MattSturgeon, line 61:
```suggestion
        warnings = lib.nixvim.mkWarnings "plugins.cmp" [
```

lib/utils.nix
MattSturgeon: > Maybe we can start with that and improve it later.

Agreed, if we feel the need for more options we can add them at the time.

lib/default.nix
MattSturgeon: bump: any chance of a unit test?

MattSturgeon (approved):
😍

PR #2852, opened 2025-01-17
lib,lsp,none-ls: Add a standard way to handle automatic installation of packages that may not exist in nixpkgs

traxys:
If such a package is used by an end user an error will be raised with a message of the form:

```
       error: Nixvim (plugins.none-ls.sources.formatting.textlint.package): No package is known for textlint, install externally and set this option to null
```

I'm open to changing that to a warning and returning null instead, though this would complicate the implementation as we would now need to check the priorities

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -403,5 +403,43 @@ rec {
         if cfg ? lazyLoad then lib.literalMD "`false` when lazy-loading is enabled." else true;
       example = false;
     };
+
+  /**
+    Create an option for a package not currently available in nixpkgs.
+
+    The option will throw an error if a value is not explicitly set by the end user.
+  */
+  mkUnpackagedOption =
+    optionName: packageName:
+    lib.mkOption {
+      type = types.nullOr types.package;
+      description = ''
+        Package to use for ${packageName}.
+        Nixpkgs does not include this package, and as such an external derivation or null must be provided.
+      '';
+      default = throw ''
+        Nixvim (${optionName}): No package is known for ${packageName}, to resolve this either:
+          - install externally and set this option to `null`
+          - or provide a derviation to install this package
+      '';
+      defaultText = lib.literalMD "No package, throws when undefined";
+    };
+
+  /**
+    Create an option for a package that may not be currently available in nixpkgs.
+
+    See `mkUnpackagedOption`
+  */
+  mkMaybeUnpackagedOption =
+    optionName: pkgs: packageName: package:
+    if lib.isOption package then
+      package
+    else if package != null then
+      lib.mkPackageOption pkgs packageName {
+        nullable = true;
+        default = package;
+      }
+    else
+      mkUnpackagedOption optionName packageName;
 }
 // removed
--- a/plugins/lsp/language-servers/_mk-lsp.nix
+++ b/plugins/lsp/language-servers/_mk-lsp.nix
@@ -37,7 +37,8 @@ in
 {
   meta.nixvimInfo = {
     # TODO: description
-    url = args.url or opts.package.default.meta.homepage or null;
+    # The package default can throw when the server is unpackaged, so use tryEval
+    url = args.url or (builtins.tryEval opts.package.default).value.meta.homepage or null;
     path = [
       "plugins"
       "lsp"
@@ -51,21 +52,8 @@ in
       enable = lib.mkEnableOption description;
 
       package =
-        if lib.isOption package then
-          package
-        else if args ? package then
-          lib.mkPackageOption pkgs name {
-            nullable = true;
-            default = package;
-          }
-        else
-          # If we're not provided a package, we should provide a no-default option
-          lib.mkOption {
-            type = types.nullOr types.package;
-            description = ''
-              The package to use for ${name}. Has no default, but can be set to null.
-            '';
-          };
+        lib.nixvim.mkMaybeUnpackagedOption "plugins.lsp.servers.${name}.package" pkgs name
+          package;
 
       cmd = mkOption {
         type = with types; nullOr (listOf str);
--- a/tests/lsp-servers.nix
+++ b/tests/lsp-servers.nix
@@ -82,7 +82,10 @@ let
           {
             enable = !(lib.elem server disabled);
           }
-          // lib.optionalAttrs (opts ? package && !(opts.package ? default)) { package = null; }
+          # Some servers are defined using mkUnpackagedOption whose default will throw
+          // lib.optionalAttrs (opts ? package && !(builtins.tryEval opts.package.default).success) {
+            package = null;
+          }
         ))
         (lib.filterAttrs (server: _: !(lib.elem server renamed)))
       ];
--- a/tests/test-sources/plugins/by-name/none-ls/default.nix
+++ b/tests/test-sources/plugins/by-name/none-ls/default.nix
@@ -100,7 +100,6 @@
 
   with-sources =
     {
-      config,
       options,
       lib,
       pkgs,
@@ -158,8 +157,10 @@
                 # Enable unless disabled above
                 enable = !(lib.elem sourceName disabled);
               }
-              # Some sources have a package option with no default
-              // lib.optionalAttrs (opts ? package && !(opts.package ? default)) { package = null; }
+              # Some sources are defined using mkUnpackagedOption whose default will throw
+              // lib.optionalAttrs (opts ? package && !(builtins.tryEval opts.package.default).success) {
+                package = null;
+              }
             )
           ) options.plugins.none-ls.sources;
       };
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -54,6 +54,7 @@ lib.makeExtensible (
       mkCompositeOption
       mkCompositeOption'
       mkLazyLoadOption
+      mkMaybeUnpackagedOption
       mkNullOrLua
       mkNullOrLua'
       mkNullOrLuaFn
@@ -69,6 +70,7 @@ lib.makeExtensible (
       mkPackageOption
       mkPluginPackageOption
       mkSettingsOption
+      mkUnpackagedOption
       pluginDefaultText
       ;
 
--- a/plugins/by-name/none-ls/_mk-source-plugin.nix
+++ b/plugins/by-name/none-ls/_mk-source-plugin.nix
@@ -9,7 +9,6 @@ sourceType: sourceName:
 let
   inherit (import ./packages.nix lib) packaged;
   pkg = packaged.${sourceName};
-  loc = lib.toList pkg;
 
   cfg = config.plugins.none-ls;
   cfg' = config.plugins.none-ls.sources.${sourceType}.${sourceName};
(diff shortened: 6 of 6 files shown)


lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
...
+          Package to use for ${name}.
+        '';
+      })
+      // {
+        _unpackaged = true;
+        _packageName = name;
+      };
+
+  useOptionalPackage =
+    options: config: opt:
```
MattSturgeon, line 416:
```suggestion
    options: config: loc:
```

or

```suggestion
    options: config: path:
```

(not sure we really need `config` either, because we have `opt.value` available)

lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
...
+    let
+      def = lib.getAttrFromPath opt options;
```
MattSturgeon, line 418:
```suggestion
      opt = lib.getAttrFromPath loc options;
```

lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
...
+    in
+    if !def.isDefined && def ? _unpackaged then
+      throw ''
+        Nixvim (${lib.concatStringsSep "." opt}): No package is known for ${def._packageName}, install externally and set this option to null
```
MattSturgeon, line 422:
```suggestion
        Nixvim (${opt}): No package is known for ${opt._packageName}, install externally and set this option to null
```

options have a `__toString` which uses `showOption opt.loc`, which is a bit nicer than using the `loc` we have passed in, because it also gets things like the prefix (`programs.nixvim`).

Not sure where `opt._packageName` comes from?

lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
...
+      ''
+    else
+      lib.getAttrFromPath opt config;
```
MattSturgeon, line 425:
```suggestion
      opt.value;
```

lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
...
+      (lib.mkOption {
+        type = types.nullOr types.package;
+        description = ''
+          Package to use for ${name}.
+        '';
+      })
+      // {
+        _unpackaged = true;
+        _packageName = name;
+      };
```
MattSturgeon, line 413: Does this survive the module system? I'm pretty sure particular parts of the module system that elaborate options will only inherit known attrs.

lib/options.nix
```diff
@@ -391,6 +391,39 @@ rec {
       );
     };
 
+  mkOptionalPackage =
```
MattSturgeon, line 394: initial impression: this feels like an anti-pattern. Can we find a way to "use" an optional package without having to declare the option in a special way?

lib/options.nix
traxys: From what I tested it does (as I managed to trigger the warnkngs

lib/options.nix
```diff
@@ -403,5 +403,31 @@ rec {
...
+      }
+    else
+      (lib.mkOption {
+        type = types.nullOr types.package;
+        description = ''
+          Package to use for ${name}.
+          Nixpkgs does not include this package, and as such an external derivation or null must be provided.
+        '';
+        default = throw ''
+          Nixvim (${loc}): No package is known for ${name}, to resolve this either:
```
MattSturgeon, line 424:
```suggestion
          Nixvim (${lib.showOption loc}): No package is known for ${name}, to resolve this either:
```

MattSturgeon: Oh, I see. `loc` isn't a list.

Maybe we should have a different name, e.g. `fullName`?

tests/lsp-servers.nix
```diff
@@ -82,7 +82,7 @@ let
           {
             enable = !(lib.elem server disabled);
           }
-          // lib.optionalAttrs (opts ? package && !(opts.package ? default)) { package = null; }
+          // lib.optionalAttrs (opts ? package && opts.package ? _unpackaged) { package = null; }
```
MattSturgeon, line 85: One (hacky) way we could avoid needing `_unpackaged` here is by using [`builtins.tryEval`](https://noogle.dev/f/builtins/tryEval) on the default...

```nix
# Use tryEval to test for a default that throws
lib.optionalAttrs (!(builtins.tryEval opts.package.default).success) { package = null; }
```

I don't love this, but in lieu of https://github.com/NixOS/nixpkgs/pull/341199, I _think_ I prefer it to `_unpackaged` :thinking:

lib/options.nix
```diff
@@ -403,5 +403,31 @@ rec {
         if cfg ? lazyLoad then lib.literalMD "`false` when lazy-loading is enabled." else true;
       example = false;
     };
+
+  mkMaybeUnpackagedPackage =
```
MattSturgeon, line 407: Could we split this into `mkMaybeUnpackagedPackageOption` and `mkUnpackagedPackageOption`?

That way calling code could decide for itself whether to use `mkPackageOption` or `mkUnpackagedPackageOption` if needed.

lib/options.nix
MattSturgeon, line 1: Could we add doc-comments to the new functions?

MattSturgeon (commented):
I agree, this is looking nicer.

Thanks for continuing to work on this!

lib/options.nix
```diff
@@ -403,5 +403,31 @@ rec {
...
+        default = package;
+      }
+    else
+      (lib.mkOption {
+        type = types.nullOr types.package;
+        description = ''
+          Package to use for ${name}.
+          Nixpkgs does not include this package, and as such an external derivation or null must be provided.
+        '';
+        default = throw ''
```
MattSturgeon, line 423: Looking at CI ([e.g.](https://buildbot.nix-community.org/#/builders/684/builds/3767)), I'm wondering if this fails because the module system does type-checking on all definitions, even those that are overridden?

I.e. when calling `opt.type.check` with the default definition, this will throw.

traxys: At least one of the issues is related to the documentation generation

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
...
+      description = ''
+        Package to use for ${name}.
+        Nixpkgs does not include this package, and as such an external derivation or null must be provided.
+      '';
+      default = throw ''
+        Nixvim (${fullName}): No package is known for ${name}, to resolve this either:
+          - install externally and set this option to `null`
+          - or provide a derviation to install this package
+      '';
+      defaultText = lib.literalExpression "No package";
```
MattSturgeon, line 424: Nit: 
```suggestion
      defaultText = lib.literalMD "No package";
```

Non-code defaults are usually better as markdown than expression.

Could even say something like `"throws when undefined"`?

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
         if cfg ? lazyLoad then lib.literalMD "`false` when lazy-loading is enabled." else true;
       example = false;
     };
+
+  /**
+    Create an option for an executable currently unpackaged in nixpkgs.
+    This option will throw an error if a value is not explicitly set by the end user
+  */
+  mkUnpackagedOption =
+    fullName: name:
```
MattSturgeon, line 412: Q: would `pluginName` and `packageName` be better?

traxys: Not sure, maybe `optionName` and `packageName`?

lib/options.nix
MattSturgeon: Hm, even with `defaultText` we still have this in the [latest CI run](https://buildbot.nix-community.org/#/builders/1017/builds/3762):

```
error: Nixvim (plugins.lsp.servers.ada_ls.package): No package is known for ada_ls, to resolve this either:
  - install externally and set this option to `null`
  - or provide a derviation to install this package
```

MattSturgeon: Ah, I bet it's coming from `url` attempting to get `meta.homepage`.

traxys: And it only occurs in the mdbook documentation, not for the generation of man pages

traxys:
```
while evaluating the attribute 'package.default.meta.homepage'
```

:(

MattSturgeon: > ```
> while evaluating the attribute 'package.default.meta.homepage'
> ```

See above: https://github.com/nix-community/nixvim/pull/2852#discussion_r1936381943

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -37,7 +37,7 @@ in
 {
   meta.nixvimInfo = {
     # TODO: description
-    url = args.url or opts.package.default.meta.homepage or null;
+    url = args.url or (builtins.tryEval opts.package.default).value.meta.homepage or null;
```
MattSturgeon, line 40: Nit: we should probably comment _why_ were using `tryEval`


```suggestion
    # The package default can throw when undefined, so use tryEval
    # See mkUnpackagedOption
    url = args.url or (builtins.tryEval opts.package.default).value.meta.homepage or null;
```

tests/test-sources/plugins/by-name/none-ls/default.nix
```diff
@@ -159,7 +158,9 @@
                 enable = !(lib.elem sourceName disabled);
               }
               # Some sources have a package option with no default
```
MattSturgeon, line 160: This comment is no longer strictly accurate.

It should probably mention throwing and `mkUnpackagedOption` instead?

tests/lsp-servers.nix
```diff
@@ -82,7 +82,9 @@ let
           {
             enable = !(lib.elem server disabled);
           }
-          // lib.optionalAttrs (opts ? package && !(opts.package ? default)) { package = null; }
+          // lib.optionalAttrs (opts ? package && !(builtins.tryEval opts.package.default).success) {
+            package = null;
+          }
```
MattSturgeon, line 87: This should probably also have a comment, matching none-ls

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
         if cfg ? lazyLoad then lib.literalMD "`false` when lazy-loading is enabled." else true;
       example = false;
     };
+
+  /**
+    Create an option for an executable currently unpackaged in nixpkgs.
+    This option will throw an error if a value is not explicitly set by the end user
```
MattSturgeon, line 409:
```suggestion

    The option will throw an error if a value is not explicitly set by the end user.
```

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
...
+        Nixvim (${optionName}): No package is known for ${packageName}, to resolve this either:
+          - install externally and set this option to `null`
+          - or provide a derviation to install this package
+      '';
+      defaultText = lib.literalMD "No package, throws when undefined";
+    };
+
+  /**
+    Create an option for an executable that may not be currently packaged in nixpkgs.
+    See mkUnpackagedOption
```
MattSturgeon, line 429:
```suggestion

    See `mkUnpackagedOption`.
```

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
...
+      default = throw ''
+        Nixvim (${optionName}): No package is known for ${packageName}, to resolve this either:
+          - install externally and set this option to `null`
+          - or provide a derviation to install this package
+      '';
+      defaultText = lib.literalMD "No package, throws when undefined";
+    };
+
+  /**
+    Create an option for an executable that may not be currently packaged in nixpkgs.
```
MattSturgeon, line 428:
```suggestion
    Create an option for a package that may not be currently packaged in nixpkgs.
```

lib/options.nix
```diff
@@ -403,5 +403,41 @@ rec {
         if cfg ? lazyLoad then lib.literalMD "`false` when lazy-loading is enabled." else true;
       example = false;
     };
+
+  /**
+    Create an option for an executable currently unpackaged in nixpkgs.
```
MattSturgeon, line 408:
```suggestion
    Create an option for a package not currently available in nixpkgs.
```

MattSturgeon (commented):
LGTM. Final nits and wording suggestions below:

lib/options.nix
```diff
@@ -403,5 +403,43 @@ rec {
...
+      default = throw ''
+        Nixvim (${optionName}): No package is known for ${packageName}, to resolve this either:
+          - install externally and set this option to `null`
+          - or provide a derviation to install this package
+      '';
+      defaultText = lib.literalMD "No package, throws when undefined";
+    };
+
+  /**
+    Create an option for a package that may not be available packaged in nixpkgs.
```
MattSturgeon, line 429:
```suggestion
    Create an option for a package that may not be currently available in nixpkgs.
```

Oops, I think I suggested _"available packaged"_ :grin:

MattSturgeon (approved):
LGTM, thanks!

traxys:
@mergify queue

PR #2854, opened 2025-01-18
experimental configuration-based nixvim template

MattSturgeon:
Adds an experimental flake template that configures Nixvim directly using `evalNixvim` rather than the current standalone functions.

### Related issues
- #2210
- #2858
- #2867
- #2868

### Real world usage
- https://github.com/khaneliman/khanelivim/pull/12

### Todo

- [ ] Document new template/approach
- [x] Ensure **all** template's `checks` are inherited into the main flake's `checks`
- [x] Include `nixvim-print-init` in `build.package` #2874
- [x] Include man-docs in `build.package` #2874

Since this is experimental/draft, we don't need to tick off everything above before merging.

### DIFF
--- a/flake/templates.nix
+++ b/flake/templates.nix
@@ -25,7 +25,7 @@
             args@{
               inputs,
               outputs,
-              sourceInfo,
+              sourceInfo ? { },
             }:
             let
               outputs = args.outputs (inputs // { self = result; });
@@ -39,18 +39,21 @@
             in
             result;
 
-          templateFlakeOutputs = callFlake {
-            inputs = {
-              inherit (inputs) flake-parts nixpkgs;
-              nixvim = self;
-            };
-            # Import and read the `outputs` field of the template flake.
-            inherit (import ../templates/simple/flake.nix) outputs;
-            sourceInfo = { };
-          };
-
-          templateChecks = templateFlakeOutputs.checks.${system};
+          flakes = lib.mapAttrs (
+            name: template:
+            callFlake {
+              # Use inputs from our flake
+              inputs = {
+                inherit (inputs) flake-parts nixpkgs;
+                nixvim = self;
+              };
+              # Use outputs from the template flake
+              inherit (import "${template.path}/flake.nix") outputs;
+            }
+          ) self.templates;
         in
-        lib.concatMapAttrs (checkName: check: { "template-${checkName}" = check; }) templateChecks;
+        lib.concatMapAttrs (name: flake: {
+          "template-${name}" = pkgs.linkFarm name flake.checks.${system};
+        }) flakes;
     };
 }
--- a/flake/templates.nix
+++ b/flake/templates.nix
@@ -5,6 +5,10 @@
       path = ../templates/simple;
       description = "A simple nix flake template for getting started with nixvim";
     };
+    new = {
+      path = ../templates/experimental-flake-parts;
+      description = "An experimental flake template for configuring nixvim using evalNixvim and flake.parts";
+    };
   };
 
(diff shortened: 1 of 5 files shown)


traxys:
As we are defining a flake parts module could we make it such that it automatically adds a `package.${system}.nvim` entry set to `         config.nixvimConfigurations.default.config.build.package` and a `check.${system}.nvim` set to `config.nixvimConfigurations.default.config.build.test` if they exist?

We could add flake module options to disable this behaviour as needed. It's maybe a bit too magic?

MattSturgeon:
I'm moving the flake-parts module stuff to a dedicated PR: #2858

flake/flake-modules/auto.nix
```diff
@@ -0,0 +1,72 @@
...
+let
+  cfg = config.nixvim;
+in
+{
+  options.nixvim = {
+    packages = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        description = "Whether to automatically install packages for each nixvimConfiguration.";
+        default = false;
```
traxys, line 11: Are we sure to default false for it?

traxys (approved):
This is a MUCH cleaner way to handle standalone modules!
It should also be _much_ simpler to have a flake which uses standalone modules for rapid iteration & and exposes neovim through an home manager module with this kind of architecture

flake/flake-modules/auto.nix
MattSturgeon: I wouldn't say I'm sure...

My rationale was;  adding `packages` _by default_ may be more controversial than adding `checks` _by default_

traxys: I can't manage to convince myself one way or the other, so I'll defer to anyone with an opinion

MattSturgeon: > Why wouldn't someone define a nixvim config in his flake without wanting to have the `packages` output ?

Maybe they have several nixvim configurations and only want package outputs for some of them.

Maybe they want to use the nixvim configuration packages _only_ within other configurations (NixOS, home-manager).

Maybe they want bespoke/different names for the package attrs, and `nameFunction` is impractical for what they want to achieve.

I'm not _convinced_ that we should be messing with users flake outputs without them being aware, i.e. having opted in.

`checks` feel less special than `packages`, but maybe both should be default-false for sanity & consistency?

khaneliman (commented):
I'm really liking this. Right now my standalone setup feels hacky for creating the actual package output and this keeps it all modules and configuration without worrying about the flake output aspect, as much.

templates/new/flake.nix
```diff
@@ -0,0 +1,59 @@
...
+              modules = [
+                self.nixvimModules.default
+              ];
+            };
+          };
+
+          packages = {
+            # Lets you start nixvim using `nix run`
+            default = config.packages.nixvim-default;
+          };
```
MattSturgeon, line 56: This could go if we merge #2875

flake/templates.nix
```diff
@@ -5,6 +5,10 @@
       path = ../templates/simple;
       description = "A simple nix flake template for getting started with nixvim";
     };
+    new = {
+      path = ../templates/new;
+      description = "An experimental flake template for configuring nixvim directly with evalNixvim";
+    };
```
MattSturgeon, line 11: How do we feel about the template name & description?

traxys: Maybe we can mention flake-parts?

traxys (approved):
I'm not too concerned about the name personally.

We should figure out a way to document the templates, maybe by adding their README to sections in the main book

MattSturgeon:
@mergify queue

PR #2858, opened 2025-01-19
flake: initial flake-parts modules

MattSturgeon:
- **flake: add initial flake-parts module**
- **flake: add `nixvimConfigurations` flake-parts module**
- **flake: add `nixvimModules` flake-parts module**

This adds flake-parts modules for declaring `flake.nixvimModules` and `perSystem.nixvimConfigurations`. A `default` module imports both of these.

This is currently undocumented. It should be possible to document these modules in a similar way to how we document platform-wrappers. We may also want to document our flake-parts modules on the flake.parts website?

@traxys also [proposed](https://github.com/nix-community/nixvim/pull/2854#issuecomment-2600121408) having some auto-configuration, which I agree is a good idea. However I'd like to keep the scope of this PR small, so let's postpone that for now.

### DIFF
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -4,6 +4,7 @@
     ./flake-modules
     ./lib.nix
     ./legacy-packages.nix
+    ./nixvim-configurations.nix
     ./overlays.nix
     ./packages.nix
     ./templates.nix
--- a/flake/default.nix
+++ b/flake/default.nix
@@ -1,6 +1,7 @@
 {
   imports = [
     ./dev
+    ./flake-modules
     ./lib.nix
     ./legacy-packages.nix
     ./overlays.nix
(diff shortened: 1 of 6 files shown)


MattSturgeon:
Re documentation:

I've sent https://github.com/hercules-ci/flake.parts-website/pull/1288 upstream that should allow us to leverage the flake-parts-website tooling without having to also have extra stuff in our flake's `packages` outupt.

It would require adding `flake-parts-website` as a flake input though, or fetching it with a fetcher _I guess_.

The alternative is to use `flake-parts-lib.evalFlakeModule` manually, but scrolling through [`render-module.nix`](https://github.com/hercules-ci/flake.parts-website/blob/main/render/render-module.nix) there's a few edge-cases & botstrapping we may need to handle too if we take that approach.

MattSturgeon:
@mergify queue

PR #2859, opened 2025-01-19
ci: drop 24.05 support

MattSturgeon:
24.05 has reached end-of-life. Let's not auto-update its lockfile or generated docs for it anymore.

I've kept the ability to manually trigger updating its lockfile.

Closes #2856


MattSturgeon:
@mergify queue

PR #2862, opened 2025-01-19
modules/nixpkgs: `useGlobalPackages` default to false

MattSturgeon:
Rather than waiting for a technically impossible migration strategy, let's just bite the bullet and change the default.

This PR also adds a "breaking changes" notice to the docs homepage, since we don't have a better way to inform users.

### DIFF
--- a/docs/mdbook/index.md
+++ b/docs/mdbook/index.md
@@ -8,6 +8,14 @@ Documentation is currently available for the following versions:
 
 @DOCS_VERSIONS@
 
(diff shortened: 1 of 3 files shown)


traxys (approved):

traxys:
I guess we will see how many people complain

MattSturgeon:
@mergify queue

fjij:
I am complaining :)

https://github.com/nix-community/nixvim/issues/2876

PR #2865, opened 2025-01-19
flake.nix: follow nixpkgs-unstable instead of nixos-unstable

GaetanLepage:
Since #2862, `useGlobalPackages` defaults to `false`.
This will hopefully significantly reduce the number of 'out-of-sync' issues for users.
It also allows us to move to a _fresher_ channel without disrupting the workflow of users.

Hence, I propose to switch our reference channel from `nixos-unstable` to `nixpkgs-unstable`.
This proposition is motivated by two reasons:
1. `nixpkgs-unstable` makes more sense as our reference channel as the job set required for its bump involves building a bunch of packages **on all four platforms**.
    Conversely, `nixos-unstable`'s job set is linux-only and include additional NixOS-related tests that do not really matter for us.
2. `nixpkgs-unstable` is fresher in average compared to `nixos-unstable`.
    In conjunction to increasing our _bump frequency_ (to be done in a follow-up PR), choosing this channel will allow us to follow `nixpkgs`'s `master` channel from closer.
    Thus, we will be able to upstream eventual fixes faster and reduce the average length of the cycle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ If you have any question, please use the [discussions page](https://github.com/n
 
 > [!WARNING]
 > NixVim needs to be installed with a compatible nixpkgs version.
-> This means that the `main` branch of NixVim requires to be installed with `nixos-unstable`.
+> This means that the `main` branch of NixVim requires to be installed with `nixpkgs-unstable`.
 >
 > If you want to use NixVim with nixpkgs 24.11 you should use the `nixos-24.11` branch.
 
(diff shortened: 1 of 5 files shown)


MattSturgeon:
I wonder if we should delay this until standalone users are also migrating away from the current standalone factory function?

My reasoning is that the current function encourages users to provide an externally constructed `pkgs` instance, so those users will likely still run into out-of-sync errors.

khaneliman:
Basically have to hold off until we get the same flake update issues resolved, anyways.

khaneliman (approved):

PR #2866, opened 2025-01-19
ci/update: remove nixos-24.05 update action

GaetanLepage:


MattSturgeon:
We've already removed scheduled updates for 24.05, why remove the tickbox entirely?

MattSturgeon:
I had deliberately left it, but...

On the one hand, the tickbox does no harm.

On the other hand, if we really need to update the old branch we can manually run the workflow _on_ that branch.

And we probably will not update it anyway.

So I'm fine removing it as a "cleanup" measure.

MattSturgeon (approved):

PR #2867, opened 2025-01-20
flake: add `auto` flake-parts module

MattSturgeon:
When enabled, automatically sets up packages and/or checks for all `nixvimConfigurations`

Both options currently default to false. See https://github.com/nix-community/nixvim/pull/2854#discussion_r1921623854

A `nameFunction` can be configured for transforming nixvimConfiguration names into package/check names. Maybe prefix & suffix string options would be simpler?

Originally from #2854


khaneliman (approved):

MattSturgeon:
@mergify queue

PR #2874, opened 2025-01-20
modules/output: move `symlinkJoin` to `build.package`

MattSturgeon:
- **modules/output: move `symlinkJoin` to `build.package`**
- **wrappers/standalone: simplify `extend`**

Should work better with #2854

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -113,6 +113,15 @@ in
         readOnly = true;
       };
 
+      nvimPackage = mkOption {
+        type = types.package;
+        description = ''
+          Wrapped Neovim (without man-docs, printInitPackage, etc).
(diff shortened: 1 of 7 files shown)


MattSturgeon:
I don't understand how these changes are causing combine-plugins to break like this: 
```
error: attribute 'packpathDirs' missing
at /nix/store/XXX-test-sources/modules/performance/combine-plugins.nix:8:54:
     7|     pkg: files: type:
     8|     builtins.length (builtins.filter (p: p != files) pkg.packpathDirs.myNeovimPackages.${type});
     9| in
```
https://buildbot.nix-community.org/#/builders/2795/builds/3129

MattSturgeon:
@mergify queue

PR #2875, opened 2025-01-20
flake/auto: `nameFunction` default to `lib.id`

MattSturgeon:
This seems like a simpler/saner default.

See https://github.com/khaneliman/khanelivim/pull/12#discussion_r1921824152 for prior discussion.


MattSturgeon:
@mergify queue

PR #2882, opened 2025-01-22
ci/update: schedule main branch daily, other branches weekly

MattSturgeon:
Split scheduling other branches into a dedicated workflow. The main update workflow is no longer responsible for triggering the workflow on _other_ branches.

It may also be nice to have the workflow stop when nixpks is not one of the inputs being bumped, to reduce force-push spam to the update branches. That is not part of this PR.


MattSturgeon:
@mergify queue

PR #2885, opened 2025-01-23
ci/update: cancel scheduled runs that don't bump nixpkgs

MattSturgeon:
This should help mitigate spammy force-pushes that don't actually update the nixpkgs input.

This only affects _scheduled_ runs. Manually triggered updates will always force-push to the update branch.

I've tested out the `nix eval` commands locally with some canned examples, but as ever with CI workflows I'm unable to fully verify this works in production.


MattSturgeon:
@mergify queue

PR #2887, opened 2025-01-23
ci/update: fix `getFlake` expecting an absolute path

MattSturgeon:
Pass in `$PWD` instead of `.`, because `builtins.getFlake` does not permit relative paths.

See https://github.com/nix-community/nixvim/actions/runs/12928918623/job/36057233050#step:6:56


khaneliman (approved):

MattSturgeon:
@mergify queue

PR #2888, opened 2025-01-23
plugins/gx: init

jolars:
Initialize [gx.nvim](https://github.com/chrishrb/gx.nvim), a plugin to replace netrw's gx functionality. 

I added an extra option for actually disabling netrw's gx setting as per the example from the gx docs.

### DIFF
--- a/plugins/by-name/gx/default.nix
+++ b/plugins/by-name/gx/default.nix
@@ -1,0 +1,122 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gx";
+  packPathName = "gx.nvim";
+  package = "gx-nvim";
+  description = ''
+    Implementation of gx without the need of netrw.
+  '';
+
+  maintainers = [ lib.maintainers.jolars ];
+
+  extraOptions = {
+    disableNetrwGx = lib.mkOption {
+      default = true;
+      description = "Disable the `gx` mapping in netrw.";
+      type = types.bool;
+    };
+  };
+
+  extraConfig = cfg: {
+    globals = lib.mkIf cfg.disableNetrwGx {
+      netrw_nogx = lib.mkDefault true;
+    };
+  };
+
+  settingsOptions = {
+    open_browser_app = defaultNullOpts.mkStr null ''
+      Specify your browser app; default for macOS is `"open"`, Linux `"xdg-open"`
+      and Windows `"powershell.exe"`.
+    '';
+
+    open_browser_args = defaultNullOpts.mkListOf types.str [ ] ''
+      Arguments provided to the browser app, such as `--background` for macOS's `open`.
+    '';
+
+    handlers = defaultNullOpts.mkAttrsOf types.anything { } ''
+      Enable built-in handlers and configure custom ones. By default, all
+      handlers are disabled. To enable a built-in handler, set it to `true`.
+    '';
+
+    handler_options = {
+      search_engine = defaultNullOpts.mkStr "google" ''
+        Search engine to use when searching for text. Built-in options are
+        `"google"`, `"duckduckgo"`, `"ecosia"`, `"yandex"`, and `"bing"`. You
+        can also pass a custom search engine, such as
+        `"https://search.brave.com/search?q="`.
+      '';
+
+      select_for_search = defaultNullOpts.mkBool false ''
+        If your cursor is e.g. on a link, the pattern for the link AND for the word
+        will always match. This disables this behaviour for default so that the link is
+        opened without the select option for the word AND link.
+      '';
+
+      git_remotes = defaultNullOpts.mkListOf' {
+        type = types.str;
+        pluginDefault = [
+          "upstream"
+          "origin"
+        ];
+        description = ''
+          List of git remotes to search for git issue linking, in priority.
+          You can also pass a function.
+        '';
+        example = lib.nixvim.literalLua ''
+          function(fname)
+            if fname:match("myproject") then
+              return { "mygit" }
+            end
+            return { "upstream", "origin" }
+          end
+        '';
+      };
+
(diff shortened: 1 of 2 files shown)


plugins/by-name/gx/default.nix
```diff
@@ -0,0 +1,117 @@
...
+        filename = "Cargo.toml";
+        handle.__raw = ''
+          function(mode, line, _)
+            local crate = require("gx.helper").find(line, mode, "(%w+)%s-=%s")
+
+            if crate then
+              return "https://crates.io/crates/" .. crate
+            end
+          end
+        '';
```
MattSturgeon, line 107: I think this will end up rendering poorly in the docs.

Maybe it'd be better to have the settingExample be a literal expression?

```nix
  {
    # ...
    handlers = {
      rust = {
        name = "rust";
        filetype = [ "toml" ];
        filename = "Cargo.toml";
        handle.__raw = '''
          function(mode, line, _)
            local crate = require("gx.helper").find(line, mode, "(%w+)%s-=%s")
            if crate then
              return "https://crates.io/crates/" .. crate
            end
          end
        ''';
    # ...
```

MattSturgeon (approved):
No major issues from me, thanks!

plugins/by-name/gx/default.nix
jolars: Thanks, I updated the example. By the way, is there some way to preview the documentation locally? I couldn't find anything about this in the contributing section.

plugins/by-name/gx/default.nix
```diff
@@ -0,0 +1,118 @@
...
+            return fname:match("myproject")
+          end
+        \'\';
+        ```
+      '';
+    };
+  };
+
+  settingsExample = lib.literalExpression ''
+      open_browser_args = [ "--background" ];
```
MattSturgeon, line 93:
```suggestion
  settingsExample = lib.literalExpression ''
    {
      open_browser_args = [ "--background" ];
```

plugins/by-name/gx/default.nix
```diff
@@ -0,0 +1,119 @@
...
+        git_remotes.__raw = \'\'
+          function(fname)
+            if fname:match("myproject") then
+                return { "mygit" }
+            end
+            return { "upstream", "origin" }
+          end
+        \'\';
+        ```
+      '';
```
MattSturgeon, line 74: These examples should probably use mkOption's `example` attr. `defaultNullOpts` offers `'` variants of its functions that allow passing arbitrary args through to `mkOption`:

```nix
git_remotes = defaultNullOpts.mkListOf' {
  type = types.str;
  pluginDefault = [ "upstream" "origin" ];
  description = ''
        List of git remotes to search for git issue linking, in priority.
        You can also pass a function.
      '';
  example = lib.nixvim.literalLua ''
    function(fname)
      if fname:match("myproject") then
          return { "mygit" }
      end
      return { "upstream", "origin" }
    end
  '';
```

Also: nix uses `'''` to escape `''`, not `\'\'`.

plugins/by-name/gx/default.nix
```diff
@@ -0,0 +1,119 @@
...
+        ```nix
+        git_remote_push.__raw = \'\'
+          function(fname) 
+            return fname:match("myproject")
+          end
+        \'\';
+        ```
+      '';
+    };
+  };
```
MattSturgeon, line 90: Same

MattSturgeon (changes requested):
Sorry, I glossed over the `examples` embedded in `description` on my prior reviews

plugins/by-name/gx/default.nix
jolars: Great, thanks for the pointer!

plugins/by-name/gx/default.nix
jolars: Thanks, should be fixed now.

MattSturgeon (approved):

jolars:
Now when I build the documentation locally, the examples using literalLua look like this:

```
Example: lib.nixvim.mkRaw "function(fname)\n if fname:match(\"myproject\") then\n return { \"mygit\" }\n end\n return { \"upstream\", \"origin\" }\nend\n"
```

jolars:
Should I just use a multi-line string?

MattSturgeon:
> Now when I build the documentation locally, the examples using literalLua look like this:
> 
> ```
> Example: lib.nixvim.mkRaw "function(fname)\n if fname:match(\"myproject\") then\n return { \"mygit\" }\n end\n return { \"upstream\", \"origin\" }\nend\n"
> ```

That's a known issue with literalLua. My preference would be to still use it and I'll try to get around to fixing it soon.

jolars:
Oh okay, sorry, I went too fast there. I'll switch back.

plugins/by-name/gx/default.nix
```diff
@@ -0,0 +1,122 @@
...
+        example = lib.nixvim.literalLua ''
+          function(fname)
+            return fname:match("myproject")
+          end
+        '';
+      };
+    };
+  };
+
+  settingsExample = lib.literalExpression ''
```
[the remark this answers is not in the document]
MattSturgeon: > I would not have this as a `literalExpression`...

The issue is passing in a nix value means it'll be rendered by `lib.generators.toPretty`.

This means two things:

- `x.__raw` is rendered as `{ x = { __raw }; }`
- Multiline indented strings are never used, so multi-line strings render as `"foo\nbar\n"`.

> If this is actually better for the documentation, then I would rather make it the same for all plugin.
> It would then go in `mk[Neo]vimPlugin` directly.

I don't know how technically feasible that is. Nor do I think we will _universally_ prefer literalExpressions over nix values.

For examples that don't require manual formatting, such as trivial examples or examples without `.__raw`, I'd still prefer not using literalExpression.

---

Regardless, the wider discussion should probably happen on matrix. As far as this PR is concerned, literalExpression was used at my request (https://github.com/nix-community/nixvim/pull/2888#discussion_r1927368630) so I think we should avoid going back-and-forth on this here :slightly_smiling_face:

PR #2896, opened 2025-01-24
plugins/markdown-preview: modernize

khaneliman:
Was looking at stuff and saw it could use some cleaning up with statements and moving deprecations to separate file.


MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2897, opened 2025-01-24
plugins/dap: migrate to mkNeovimPlugin

khaneliman:


plugins/by-name/dap-python/default.nix
```diff
@@ -0,0 +1,94 @@
...
+  package = "nvim-dap-python";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  extraOptions = {
+    adapterPythonPath = lib.mkOption {
+      default = lib.getExe (pkgs.python3.withPackages (ps: with ps; [ debugpy ]));
+      defaultText = lib.literalExpression ''
+        lib.getExe (pkgs.python3.withPackages (ps: with ps; [ debugpy ]))
+      '';
```
MattSturgeon, line 30: Nit: let's simplify `with ps; [ debugpy ]` -> `[ ps.debugpy ]`

plugins/by-name/dap-python/default.nix
```diff
@@ -0,0 +1,94 @@
...
+      description = "Path to the python interpreter. Path must be absolute or in $PATH and needs to have the debugpy package installed.";
```
MattSturgeon, line 31: Nit: 
```suggestion
      description = "Path to the python interpreter. Path must be absolute or in `$PATH` and needs to have the `debugpy` package installed.";
```

Also: feels like the wording could be improved?

Might be getting long enough to justify using an `'' ''` indented string?

plugins/by-name/dap-python/default.nix
```diff
@@ -0,0 +1,94 @@
...
+    resolvePython = defaultNullOpts.mkLuaFn null ''
+      Function to resolve path to python to use for program or test execution.
+      By default the `VIRTUAL_ENV` and `CONDA_PREFIX` environment variables are used if present.
+    '';
+
+    testRunner = mkNullOrOption (types.either types.str types.rawLua) ''
+      The name of test runner to use by default.
+      The default value is dynamic and depends on `pytest.ini` or `manage.py` markers.
+      If neither is found "unittest" is used.
+    '';
```
MattSturgeon, line 46: Can we convey the "dynamic" default value using a `literalMD` pluginDefault?

plugins/by-name/dap-python/default.nix
```diff
@@ -0,0 +1,94 @@
...
+
+    testRunners = mkNullOrOption (with lib.types; attrsOf strLuaFn) ''
+      Set to register test runners.
+      Built-in are test runners for unittest, pytest and django.
+      The key is the test runner name, the value a function to generate the
+      module name to run and its arguments.
+      See |dap-python.TestRunner|.
+    '';
```
MattSturgeon, line 54: Feels like this description would benefit from some line breaks?

Also, lets add some `` ` `` formatting to `|dap-python.TestRunner|`

plugins/by-name/dap-ui/default.nix
```diff
@@ -0,0 +1,197 @@
...
+        disconnect = defaultNullOpts.mkStr "" "";
+        pause = defaultNullOpts.mkStr "" "";
+        play = defaultNullOpts.mkStr "" "";
+        run_last = defaultNullOpts.mkStr "" "";
+        step_into = defaultNullOpts.mkStr "" "";
+        step_over = defaultNullOpts.mkStr "" "";
+        step_out = defaultNullOpts.mkStr "" "";
+        step_back = defaultNullOpts.mkStr "" "";
+        terminate = defaultNullOpts.mkStr "" "";
+      };
```
MattSturgeon, line 83: Feels like this could be a mapAttrs? Also, is it possible to have descriptions?

plugins/by-name/dap-virtual-text/default.nix
```diff
@@ -0,0 +1,65 @@
...
+  name = "dap-virtual-text";
+  moduleName = "nvim-dap-virtual-text";
+  package = "nvim-dap-virtual-text";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsOptions = {
+    enabled_commands = defaultNullOpts.mkBool true ''
+      Create commands `DapVirtualTextEnable`, `DapVirtualTextDisable`, `DapVirtualTextToggle`.
+      (`DapVirtualTextForceRefresh` for refreshing when debug adapter did not notify its termination).
```
MattSturgeon, line 18:
```suggestion
      Create commands `DapVirtualTextEnable`, `DapVirtualTextDisable`, `DapVirtualTextToggle`.

      (`DapVirtualTextForceRefresh` for refreshing when debug adapter did not notify its termination).
```

plugins/by-name/dap/default.nix
```diff
@@ -62,32 +55,32 @@ with dapHelpers;
...
+        };
+      } // cfg.settings;
     in
-    mkIf cfg.enable {
-      extraPlugins = [ cfg.package ];
-
+    {
       extraConfigLua =
         (optionalString (cfg.adapters != null) ''
           require("dap").adapters = ${lib.nixvim.toLuaObject options.adapters}
```
MattSturgeon, line 86: Using `optionalString` here for what should be an optional _definition_ is means we will always add at least one line to extraConfigLua.

Not a big issue, but we've had complaints from OCD users in the past :grin: 

Ideally, we'd use `mkMerge` and `mkIf` here, unless that proves ridiculously ugly in practice.

Also: the current impl does not need the parenthesis for string concatenation.

Also: should we wrap this in a `do` block?

plugins/by-name/dap-go/default.nix
```diff
@@ -0,0 +1,52 @@
...
+      Additional dap configurations.
+      See `:h dap-configuration` for more detail.
+    '';
+
+    delve = {
+      path = defaultNullOpts.mkStr "dlv" "The path to the executable dlv which will be used for debugging.";
+
+      initialize_timeout_sec = defaultNullOpts.mkInt 20 "Time to wait for delve to initialize the debug session.";
+
+      port = defaultNullOpts.mkStr "$\{port}" ''
```
MattSturgeon, line 28: Is this `"$\{port}"` correct?

I don't think that's how you escape a nix interpolation, which feels like it may have been the original intent.

Same for the description below, although we probably shouldn't be documenting the default within the description anyway.

MattSturgeon (commented):
I've not cross-referenced the old/new impl. Nor am I familiar with DAP in general. But on the whole everything looks good.

I've left some comments below, mostly on existing code that could be improved as part of this PR.

plugins/by-name/dap-python/default.nix
khaneliman: Not familiar with it enough to add extra detail. Looks like a copy pasta from the comments inside dap-python and it refers to stuff in pytest.ini

MattSturgeon: I'm fine with keeping the wording intact. I guess I was thinking something like:

```nix
{
  description = "The name of test runner to use by default.";
  pluginDefault = literalMD ''
    The default value is dynamic and depends on `pytest.ini` or `manage.py` markers.
    If neither is found "unittest" is used.
  '';
}
```

khaneliman: Ah, I see. Sure.

plugins/by-name/dap-go/default.nix
khaneliman: I think the description explaining the default just adds detail explaining what it means to have that default.

plugins/by-name/dap-go/deprecations.nix
MattSturgeon, line 1: Could you add dates and/or PR-link comments to your deprecation files?


E.g. something like

```
# NOTE: Renames added in https://github.com/nix-community/nixvim/pull/2897 (2025-01-26)
```

khaneliman: I was starting to question the practice if we are going to keep things around, you will see that basic information in a git blame.

plugins/by-name/dap/default.nix
khaneliman: Updated it, let me know what ya think. I think the generated config looks good.

plugins/by-name/dap-go/deprecations.nix
MattSturgeon: I get where you're coming from, but

1. `git blame` can be a pain when you need to step back through intermediate commits to find the original source of the changes
2. `git blame` doesn't show the PR (just the commit)
3. We may have a change of heart, or decide to clean up _really_ old warnings

While my preference is to keep deprecation warnings around indefinitely, we haven't assessed how much of a performance impact they have on the module eval.

Having comments isn't necessary, but I think it makes this kinda thing _easier_ to reason about in the future when we're assessing how recent/important various warnings are.

plugins/by-name/dap/default.nix
MattSturgeon: I was thinking 

```nix
{
  extraConfigLua = mkMerge [
    (mkIf true "")
    (mkIf false "")
    # ...
  ];
}
```

?

khaneliman: I thought I did that it complained.. redid it and it worked.

plugins/by-name/dap-go/deprecations.nix
khaneliman: Fair enough, we can keep doing it, I guess.

MattSturgeon (approved):
I'm not too confident with this part of the codebase, but I don't see any issues.

I'd feel better if @GaetanLepage or @traxys could also review before merging, assuming they have time.

khaneliman:
Was testing with this branch for a bit and then wanted to test lazy loading. Had a couple tweaks for the packPathNames but also need to make some more tweaks to support lazy loading properly, still.

image

plugins/by-name/dap/default.nix
```diff
@@ -62,45 +55,51 @@ with dapHelpers;
...
+        };
+      } // cfg.settings;
     in
-    mkIf cfg.enable {
-      extraPlugins = [ cfg.package ];
-
-      extraConfigLua =
-        (optionalString (cfg.adapters != null) ''
+    {
+      plugins.dap.luaConfig.content = lib.mkMerge [
```
khaneliman, line 84: Forgot to switch to use the `luaConfig` so lazy loading wasn't putting it in `lz.n` config. :P

khaneliman:
Alright, lazy loading is working with this PR.

image
 vs
image

plugins/by-name/dap-virtual-text/default.nix
```diff
@@ -0,0 +1,68 @@
...
+
+    commented = defaultNullOpts.mkBool false "Prefix virtual text with comment string.";
+
+    only_first_definition = defaultNullOpts.mkBool true "Only show virtual text at first definition (if there are multiple).";
+
+    all_references = defaultNullOpts.mkBool false "Show virtual text on all references of the variable (not only definitions).";
+
+    clear_on_continue = defaultNullOpts.mkBool false "Clear virtual text on `continue` (might cause flickering when stepping).";
+
+    display_callback = defaultNullOpts.mkLuaFn ''
```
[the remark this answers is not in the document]
MattSturgeon: Bearing in mind this is a migration, wouldn't that be a breaking change?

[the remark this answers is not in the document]
khaneliman: I intentionally left these alone.

PR #2900, opened 2025-01-24
ci/update: check whether a PR is already open

MattSturgeon:
If there's no open PR, there's no reason to run the "check if the input changed" step.

Knowing about the open PR also allows us to link to it in the summary.

Follow up: it may be nice for _cancelling_ the workflow to not show up as an error. This could be done by having the follow-up steps be conditional on a `steps.check.outputs.stop` output, _instead of_ exiting non-zero.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -31,6 +31,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 40
     if: github.event_name != 'schedule' || github.repository == 'nix-community/nixvim'
+    env:
+      repo: ${{ github.repository }}
+      branch: update/${{ github.ref_name }}
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


MattSturgeon:
I've now tested this in production (on a `test/update-pr` branch). After ironing out a few kinks I'm happy it seems to be working.

.github/workflows/update.yml
```diff
@@ -49,6 +52,38 @@ jobs:
...
           git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
 
+      - name: Get info on the current PR
+        id: open_pr_info
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Query for info about the already open update PR
+          info=$(
+            gh api graphql -F owner='{owner}' -F repo='{repo}' -F branch="$branch" -f query='
```
khaneliman, line 62: Didn't even know github had a graphql endpoint. nice.

khaneliman (approved):

.github/workflows/update.yml
MattSturgeon: Yeah, you can play with it in the online repl too: https://docs.github.com/en/graphql/overview/explorer

MattSturgeon:
@mergify queue

PR #2905, opened 2025-01-24
docs: ensure `pkgs` is only used for `meta` during module eval

MattSturgeon:
This PR sets a `defaultText` for all option-defaults, examples, etc that depend on packages. Additionally, we no longer need `allowUnfree`. :tada: 

A "stub" `pkgs` instance is constructed, replacing all packages with stubs that throw when evaluated.

As an exception to the rule, `meta` attrs are _not_ replaced, meaning the docs can continue to depend on `meta.homepage` for now.

Some other exceptions are allowed, mainly derivation factory functions used within the `outputs` module.

With the new approach, forgetting to define an appropriate example/defaultText, would result in an error such as:

```
error: Attempted to access `pkgs.coreutils` while rendering the docs.
```

Building with `--show-trace` would show the offending code somewhere in the middle of the trace. In this case:

```
… while evaluating the attribute 'cmd'
 at /nix/store/g2g7rznbqh5mqa4ywqzhlhsqy7pc0540-source/plugins/by-name/fzf-lua/default.nix:92:13:
   91|           settings = {
   92|             previewers.cat.cmd = "${pkgs.coreutils}/bin/cat";
     |             ^
   93|             winopts.height = 0.5;

… from call site
 at /nix/store/g2g7rznbqh5mqa4ywqzhlhsqy7pc0540-source/plugins/by-name/fzf-lua/default.nix:92:35:
   91|           settings = {
   92|             previewers.cat.cmd = "${pkgs.coreutils}/bin/cat";
   93|             winopts.height = 0.5;
```


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -78,6 +79,25 @@ in
...
+              else if lib.hasInfix "\n" text || lib.hasInfix "``" text then
+                "\n\n```\n${text}\n```"
+              else
+                "`` ${text} ``"
+            else if cmd == null then
+              "null"
+            else
+              "`[ ${lib.concatMapStringsSep " " builtins.toJSON cmd} ]`"
+          }
+        '';
```
MattSturgeon, line 100: Internally, I believe we only ever supply `cmd` as a list. So only the last two branches here (`if null` and `else`) will ever be used.

It'd be a breaking change for anyone using mkLsp out-of-tree, but for simplicity we could drop support for `cmd` being a function?

plugins/by-name/efmls-configs/default.nix
```diff
@@ -44,10 +44,13 @@ in
...
-          opt
-        else
-          builtins.removeAttrs opt [ "defaultText" ] // { default = null; }
+        opt
+        // {
+          default = if lib.meta.availableOn pkgs.stdenv.hostPlatform opt.default then opt.default else null;
+          defaultText = lib.literalMD ''
+            `${opt.defaultText.text}` if available on the current system, otherwise null
+          '';
+        }
```
MattSturgeon, line 53: Do we _really_ need to nullify the default on unsupported platforms? It'd be much simpler to just set the default (unconditionally) and let users nullify it themselves if need be.

cc @traxys

plugins/by-name/dap/dap-python.nix
```diff
@@ -23,6 +23,9 @@ in
 
     adapterPythonPath = mkOption {
       default = "${pkgs.python3.withPackages (ps: with ps; [ debugpy ])}/bin/python3";
+      defaultText = lib.literalExpression ''
+        (pkgs.python3.withPackages (ps: with ps; [ debugpy ])) + "/bin/python3"
```
MattSturgeon, line 27:
```suggestion
        lib.getExe (pkgs.python3.withPackages (ps: with ps; [ debugpy ]))
```

plugins/by-name/copilot-vim/default.nix
```diff
@@ -25,6 +25,7 @@ lib.nixvim.plugins.mkVimPlugin {
     node_command = mkOption {
       type = with types; nullOr str;
       default = "${pkgs.nodejs-18_x}/bin/node";
+      defaultText = lib.literalExpression ''pkgs.nodejs-18_x + "/bin/node"'';
```
MattSturgeon, line 28:
```suggestion
      defaultText = lib.literalExpression "lib.getExe pkgs.nodejs-18_x";
```

plugins/by-name/codeium-vim/default.nix
```diff
@@ -58,6 +58,7 @@ lib.nixvim.plugins.mkVimPlugin {
     bin = mkOption {
       type = with types; nullOr str;
       default = "${pkgs.codeium}/bin/codeium_language_server";
+      defaultText = lib.literalExpression ''pkgs.codeium + "/bin/codeium_language_server"'';
```
MattSturgeon, line 61:
```suggestion
      defaultText = lib.literalExpression ''lib.getExe' pkgs.codeium "codeium_language_server"'';
```

plugins/by-name/clipboard-image/default.nix
```diff
@@ -95,17 +95,18 @@ in
...
+      extraDescription = ''
+        ${"\n\n"}
 
         Recommended:
           - X11: `pkgs.xclip`
           - Wayland: `pkgs.wl-clipboard`
           - MacOS: `pkgs.pngpaste`
       '';
-      example = pkgs.wl-clipboard;
     };
```
MattSturgeon, line 110: Was it intended that this option not have a default? I don't see any reason why we'd want that. Defaulting to null makes sense to me.

docs/default.nix
```diff
@@ -9,6 +9,56 @@ let
...
+      stubPackage =
+        prefix: name: package:
+        let
+          loc = prefix ++ [ name ];
+        in
+        if isPackageSet name then
+          lib.mapAttrs (stubPackage loc) package
+        else
+          lib.mapAttrs (_: throwAccessError loc) package
+          // lib.optionalAttrs (package ? meta) { inherit (package) meta; };
```
MattSturgeon, line 36: This would be much simpler (and more efficient) if we didn't need access to `meta.homepage` in the docs.

But realistically, I don't think this is _too_ bad.

plugins/by-name/clipboard-image/default.nix
traxys: I seems like the plugin _requires_ a package, else it raises a checkhealth issue:

```
local packages = {
  x11 = { name = "xclip", binary = "xclip" },
  tty = { name = "xclip", binary = "xclip" },
  wayland = { name = "wl-clipboard", binary = "wl-paste" },
  darwin = { name = "pngpaste", binary = "pngpaste" },
  windows = { name = "powershell", binary = "powershell.exe" },
  wsl = { name = "powershell", binary = "powershell.exe" },
}
```

Though we should maybe add a warning if it's a default prio null instead of not providing a value

plugins/by-name/codeium-vim/default.nix
traxys: Agree.
We should maybe also modify the default to that? But it's a bit orthogonal to this

plugins/by-name/efmls-configs/default.nix
traxys: I think it was @GaetanLepage that added that? I would not be opposed to changing the behaviour.
The tests may rely on it though

plugins/lsp/language-servers/_mk-lsp.nix
traxys: I'm not sure if we expose in an easily accessible manner `mkLsp`, so I don't think it's a big breaking change

traxys: > Also, is it really necessary to change `cmd`'s default to null when `package` is null? Why not just set `cmd`'s default unconditionally?

The implemention is quite old, maybe before I started to contribute. The direction we took recently was to try and make most things less magic, so I'd be in favour of less complex behaviour

docs/default.nix
traxys: Yeah I agree it's not _too_ bad

docs/pkgs.nix
```diff
@@ -3,14 +3,6 @@
   nixpkgs,
 }:
 let
-  # FIXME:
-  # Building the docs evaluates many package-option defaults, some of which are unfree.
-  # This usually happens when we include the package option value in another option's default without
-  # using a literalExpression defaultText.
-  config = {
-    allowUnfree = true;
```
khaneliman, line 11: Always thought this felt wrong. :P

plugins/by-name/clipboard-image/default.nix
khaneliman: I'd handle this in another PR. Would be hard to default it sanely so should probably be some kind of warning/assertion that it needs to be configured if the plugin doesn't check itself.

khaneliman (approved):
LGTM imo, I appreciate the using literals in the context of showing how you can override packages since it renders better in the docs than a nix store path.

plugins/by-name/clipboard-image/default.nix
traxys: No issue with that

MattSturgeon:
I've make the simple changes. The other suggestions can be tackled by anyone interested in follow up PRs :slightly_smiling_face: 

If there's no objections I'll merge this shortly.

MattSturgeon:
@mergify queue

PR #2908, opened 2025-01-25
docs/user-guide: Add an entry for collisions with combinePlugins

traxys:


MattSturgeon (approved):

khaneliman (approved):
This is something I thought about when seeing another issue around it, thanks!

traxys:
@mergify queue

PR #2916, opened 2025-01-27
dev/list-plugins: fix dap state

khaneliman:
Follow up when I was updating tracking issue noticed this was still showing as old.


khaneliman:
@mergify queue

PR #2917, opened 2025-01-27
docs/lazy-loading: add extra examples

khaneliman:
Add another example of using before loading to trigger dependencies when lazy loading to avoid race conditions.


khaneliman:
@mergify queue

PR #2922, opened 2025-01-28
ci/update: also check base branch, +cleanup

MattSturgeon:
As discussed https://github.com/nix-community/nixvim/pull/2921#issuecomment-2618861242, it'd be good for the CI to check for differences against the base branch when there is no PR currently open.

Additionally, I've cleaned up some stuff:

- Added a checkbox for manually enabling the "check for `nixpkgs` changes" (previously enabled only on scheduled runs)
- Replaced `exit 1` with a graceful cancellation
- Improved the "print summary" steps

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -14,6 +14,10 @@ on:
         type: boolean
         default: true
         description: Update generated files
+      check_for_changes:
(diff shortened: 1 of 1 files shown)


MattSturgeon:
@mergify queue

PR #2924, opened 2025-01-28
plugins/zellij-nav.nvim: init

### LINKED ISSUE #1915: [PLUGIN REQUEST] zellij-nav.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `zellij-nav`  | 
| Homepage | `https://github.com/swaits/zellij-nav.nvim`     | 
| Nixpkgs  | `false` | 

## Extra Information

The current zellij navigation plugin in Nixvim, `zellij.nvim`, [is archived](https://github.com/Lilja/zellij.nvim?tab=readme-ov-file#archived), and the creator suggests migrating to a different plugin. `zellij-nav.nvim` seems to be a suitable substitute.

PR to add `zellij-nav.nvim` to `nixpkgs` [is in the works](https://github.com/NixOS/nixpkgs/pull/329264).

GaetanLepage:
Add support for [zellij-nav.nvim](https://github.com/swaits/zellij-nav.nvim), a plugin granting seamless navigation between Neovim windows and Zellij panes.

Fixes #1915

I explicitly used `mkNeovimPlugin` although this plugin does not have any option.
It does have a `setup(opts)` function, so let's expose it just in case.

### DIFF
--- a/plugins/by-name/zellij-nav/default.nix
+++ b/plugins/by-name/zellij-nav/default.nix
@@ -1,0 +1,8 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "zellij-nav";
+  packPathName = "zellij-nav.nvim";
+  package = "zellij-nav-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+}
--- a/tests/test-sources/plugins/by-name/zellij-nav/default.nix
+++ b/tests/test-sources/plugins/by-name/zellij-nav/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.zellij-nav.enable = true;
+  };
+}


MattSturgeon:
> I explicitly used `mkNeovimPlugin` although this plugin does not have any option.
> It does have a `setup(opts)` function, so let's expose it just in case.

Agreed. This should also help with lazyloading too.

Does the `setup` fn actually accept an `opts` param? If not, we should set `hasSettings = false`.

khaneliman:
> > I explicitly used `mkNeovimPlugin` although this plugin does not have any option.
> > It does have a `setup(opts)` function, so let's expose it just in case.
> 
> Agreed. This should also help with lazyloading too.
> 
> Does the `setup` fn actually accept an `opts` param? If not, we should set `hasSettings = false`.

https://github.com/swaits/zellij-nav.nvim/blob/main/lua/zellij-nav/init.lua#L60

MattSturgeon (approved):

PR #2925, opened 2025-01-28
lib/plugins: introduce `mkMetaModule`

MattSturgeon:
Cherry-picked from #2884

### DIFF
--- a/lib/plugins/mk-neovim-plugin.nix
+++ b/lib/plugins/mk-neovim-plugin.nix
@@ -5,7 +5,7 @@
 {
   name,
   maintainers,
-  url ? throw "default",
+  url ? null,
   imports ? [ ],
   description ? null,
   # deprecations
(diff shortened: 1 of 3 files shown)


traxys:
> Just for me to understand, what is the point of using a **module** here ? I get that you want to factor out this logic, but couldn't you use a function for that ?

I'd say it's to ease the usage, as it's easier to set `imports = [ ... ];` than `extraConfig = {_: meta = ..... }`, as we have `imports` pretty much everywhere

MattSturgeon:
Yes, producing a module simplifies both the usage and implementation.

For usage, we can just dump the value into `imports`, as with any other module.

For implementation, we have full access to all module args. Especially `config` and `options`. In this case, that's useful for getting the default-package only when a `url` is not explicitly supplied.

At the end of the day, we're working with a module system. Using and producing modules will usually be the idiomatic approach.

😁

traxys (approved):

MattSturgeon:
@mergify queue

PR #2926, opened 2025-01-29
plugins/lexima: init

### LINKED ISSUE #1880: [PLUGIN REQUEST] Lexima
| Field    | Description |
|----------|-------------|
| Plugin   | `Lexima`  | 
| Homepage | `https://github.com/cohama/lexima.vim`     | 
| Nixpkgs  | [Nixpkgs](https://search.nixos.org/packages?channel=unstable&show=vimPlugins.lexima-vim&from=0&size=50&sort=relevance&type=packages&query=lexima) | 

## Extra Information

Vim plugin for vscode-like auto-closing brackets and quotes. Including logic for handling new lines, customizable, and doesn't break existing plugin.

GaetanLepage:
Add support for [lexima.vim](https://github.com/cohama/lexima.vim), a plugin to auto close parentheses.

Fixes #1880

### DIFF
--- a/plugins/by-name/lexima/settings-options.nix
+++ b/plugins/by-name/lexima/settings-options.nix
@@ -1,0 +1,51 @@
+lib:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
+{
+  no_default_rules = defaultNullOpts.mkFlagInt 0 ''
+    Whether to disable the default rules.
+
+    You can explicitly set default rules by calling `|lexima#set_default_rules()|`.
+  '';
+
+  map_escape = defaultNullOpts.mkStr "<Esc>" ''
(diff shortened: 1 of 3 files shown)


plugins/by-name/lexima/settings-options.nix
traxys, line 1: I'm not too keen on adding all the plugin options, I would prefer limiting ourselves to the ones explicitly mentionned in the README (lexima_enable_basic_rules, lexima_enable_newline_rules, lexima_enable_endwise_rules)

MattSturgeon: I'd still prefer to stick to options documented upstream, as that feels like a more stable base we'll be able to track easier. But not a _big_ deal in this case.

PR #2928, opened 2025-01-29
treewide: use mkAssertions where possible

GaetanLepage:


plugins/by-name/neotest/adapters.nix
```diff
@@ -37,12 +37,10 @@ let
...
 
-          assertions = [
-            {
-              assertion = config.plugins.neotest.enable;
-              message = "Nixvim: you have enabled `plugins.neotest.adapters.${name}` but `plugins.neotest.enable` is `false`.";
-            }
-          ];
+          assertions = lib.nixvim.mkAssertions "plugins.neotest.adapters.${name}" {
+            assertion = config.plugins.neotest.enable;
+            message = "You must set `plugins.neotest.enable` to `true` for using this extension.";
```
khaneliman, line 42:
```suggestion
            message = "You must set `plugins.neotest.enable` to `true` to use this extension.";
```

However, maybe it would be good to standardize the verbiage for all of these `you have enabled x but not y` messages separately.

plugins/by-name/git-worktree/default.nix
```diff
@@ -59,12 +59,11 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      assertions = [
-        {
-          assertion = cfg.enableTelescope -> config.plugins.telescope.enable;
-          message = ''Nixvim: The git-worktree telescope integration needs telescope to function as intended'';
-        }
-      ];
+      assertions = lib.nixvim.mkAssertions "plugins.git-worktree" {
+        assertion = cfg.enableTelescope -> config.plugins.telescope.enable;
+
+        message = "The git-worktree telescope integration needs telescope to function as intended.";
```
khaneliman, line 65: These telescope messages should probably be rewritten, too. Feels redundant with prefixing the plugin name.

khaneliman (approved):
LGTM, I think we can standardize each message type verbiage too eventually.

PR #2929, opened 2025-01-29
lib/utils: add isTrue and isFalse

GaetanLepage:
- **lib/utils: add isTrue and isFalse**
- **treewide: use isTrue/isFalse where possible**

### DIFF
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -130,6 +130,66 @@ rec {
 
   ifNonNull' = x: y: if (x == null) then null else y;
 
+  /**
+    A predicate that tests if a value is exactly true, i.e. `v == true`.
+
+    Useful when working with nullable booleans.
+
+    #Examples
+
+    ```
+    isTrue null
+    => false
+    ```
+
+    ```
+    isTrue true
+    => true
+    ```
+
+    ```
+    isTrue "true"
+    => false
+    ```
+
+    # Type
+
+    ```
+    Any -> Boolean
+    ```
+  */
+  isTrue = x: lib.isBool x && x;
+
+  /**
+    A predicate that tests if a value is exactly false, i.e. `v == false`.
+
+    Useful when working with nullable booleans.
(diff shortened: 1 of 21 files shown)


lib/utils.nix
```diff
@@ -130,6 +130,9 @@ rec {
 
   ifNonNull' = x: y: if (x == null) then null else y;
 
+  isTrue = x: lib.isBool x && x;
+  isFalse = x: lib.isBool x && !x;
```
MattSturgeon, line 134: If we do go down this route, could we have some doc comments and some motivational examples?

```md
    A predicate that tests if a value is exactly true, i.e. `v == true`.

    Useful when working with nullable booleans.

    #Examples

    ```
    isTrue null
    => false
    ```

    ```
    isTrue true
    => true
    ```

    ```
    isTrue "true"
    => false
    ```

    # Type

    ```
    Any -> Boolean
    ```
```

MattSturgeon (commented):
IMO it would be better to just disable statix's `bool_comparison` rule, so that we can do the simpler and more efficient `v == true` and `v == false`.

IIRC there used to be a `lib.isTrue` (or maybe a builtin) and it was depreciated because `==` is always better.

lib/utils.nix
khaneliman: No real opinion on it

MattSturgeon:
> IMO it would be better to just disable statix's `bool_comparison` rule, so that we can do the simpler and more efficient `v == true` and `v == false`.


Seems we can use [`perSystem.treefmt.programs.statix.disabled-lints`](https://flake.parts/options/treefmt-nix.html#opt-perSystem.treefmt.programs.statix.disabled-lints) for this.

lib/utils.nix
traxys: I prefer using simpler constructs like `==` than ad-hoc functions where possible

PR #2931, opened 2025-01-29
misc: delete useless statix.toml config file

GaetanLepage:


MattSturgeon (approved):

statix.toml
```diff
@@ -1,3 +0,0 @@
-ignore = ["plugins/lsp/language-servers/rust-analyzer-config.nix"]
-# Frequent in nix module configurations
-disabled = ["repeated_keys"]
```
MattSturgeon, line 3: Should this config be copied to the _actual_ in-use config in the flake module?

PR #2935, opened 2025-01-30
ci/update: drop dependency on peter-evans/create-pull-request

MattSturgeon:
We can create/update PRs using the `gh pr` command, no need for a third party dependency.


traxys (approved):

MattSturgeon:
@mergify queue

PR #2940, opened 2025-01-30
ci/update: re-apply manual commits from open PR

MattSturgeon:
Using `git cherry-pick`, we can re-apply any manually committed changes in the already open PR. "Manually committed" is defined as any commit _after_ the most recent commit authored by `github-actions[bot]`.

We can also check if any files have differences before pushing.

This removes the `inputs.nixpkgs` revision checking added in #2885, because the new file-differences check is simpler and more useful.

A major benefit of re-applying the additional commits is that they are no longer lost when the CI force-pushes its updates.

This has been tested:
- https://github.com/nix-community/nixvim/actions/runs/13061369469/attempts/1
- https://github.com/nix-community/nixvim/actions/runs/13061542819/attempts/1
- https://github.com/nix-community/nixvim/actions/runs/13061542819/attempts/2

Also: the second commit fixes [a bug](https://github.com/nix-community/nixvim/actions/runs/13061369469/job/36445016824#step:12:64) in #2935 where the create-pr step was checking an undefined variable.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -177,6 +177,7 @@ jobs:
         if: steps.diff.outputs.count
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          pr_num: ${{ steps.open_pr_info.outputs.number }}
           title: |
             [${{ github.ref_name }}] Update flake.lock & generated files
           body: |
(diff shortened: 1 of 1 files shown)


traxys (approved):

MattSturgeon:
@mergify queue

khaneliman:
That's awesome, thanks!

PR #2942, opened 2025-01-31
plugins/blink-cmp: support raw lua for settings.sources.providers.enabled

VioletWoodETL:
Adds the raw lua type to `plugins.blink-cmp.settings.sources.providers.enabled` so that the given example in the [documentation](https://nix-community.github.io/nixvim/plugins/blink-cmp/settings/sources/providers.html) actually works.

TODO:
- [X] Add type to `enabled`.
- [x] Add test for feature.

### DIFF
--- a/plugins/by-name/blink-cmp/provider-config.nix
+++ b/plugins/by-name/blink-cmp/provider-config.nix
@@ -25,7 +25,7 @@ types.submodule {
     };
 
     enabled = mkNullOrOption' {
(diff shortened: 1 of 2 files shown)


VioletWoodETL:
Squashing commits is a new one for me, is this good?

khaneliman (approved):
Thanks for catching that!

khaneliman:
@mergify queue

PR #2944, opened 2025-01-31
plugins/diffview: migrate to mkNeovimPlugin

khaneliman:
WIP


plugins/by-name/diffview/default.nix
```diff
@@ -1,145 +1,138 @@
...
-          ''
-            Determines whether the window should be a float or a normal
-            split.
-          '';
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
+  name = "diffview";
+  originalName = "diffview.nvim";
```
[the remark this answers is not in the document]
MattSturgeon:
```suggestion
  packPathName = "diffview.nvim";
```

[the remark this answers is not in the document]
khaneliman: This is the property name before I renamed it from originalName to packPathName

plugins/by-name/diffview/default.nix
```diff
@@ -1,145 +1,138 @@
...
-          ]
-          type
-          ''
-            Determines whether the window should be a float or a normal
-            split.
-          '';
+  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib) types;
+in
+lib.nixvim.neovim-plugin.mkNeovimPlugin {
```
MattSturgeon, line 10: This was renamed to:

```suggestion
lib.nixvim.plugins.mkNeovimPlugin {
```

GNUqb114514:
@khaneliman Would you like to work on it further?

plugins/by-name/diffview/default.nix
```diff
@@ -822,20 +681,12 @@ in
...
-
-      extraPlugins = [ cfg.package ];
-
-      extraConfigLua = ''
-        require("diffview").setup(${lib.nixvim.toLuaObject setupOptions})
-      '';
+    {
     };
+
+  inherit (import ./deprecations.nix { inherit lib; })
```
GNUqb114514, line 687: It seems that you put an unexpected argument here which causes another [](url)

MattSturgeon: Correct, `deprecations.nix` is a plain attrset, not a function. Additionally, it has no `imports` attr to inherit.

plugins/by-name/diffview/default.nix
```diff
@@ -1,805 +1,114 @@
...
-          view_enter = viewEnter;
-          view_leave = viewLeave;
-          view_post_layout = viewPostLayout;
-          diff_buf_read = diffBufRead;
-          diff_buf_win_enter = diffBufWinEnter;
-        };
-
+      keymapCfg = cfg.settings.keymaps;
+      setupOptions = cfg.settings // {
         keymaps =
```
khaneliman: Honestly, didn't spend any time really trying to optimize this with latest standards.. just wanted it to stay backwards compatible.

MattSturgeon: I would prefer to drop the `setupOptions` thing entirely and use the module system to merge the custom keymaps in.

I.e. we define the `settings.keymaps` option instead of using `// { keymaps }` explicitly.

PR #2954, opened 2025-02-02
plugins/obsidian: remove with lib and helpers; move deprecations; rename options file

khaneliman:
Misc cleanup tasks to clean up tech debt and get file out of the list-plugins showing as old.


MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2955, opened 2025-02-02
treewide: options -> settings-options

khaneliman:
Reduce redundancy in excludes by adhering to the standard.


MattSturgeon (approved):
LGTM.

Any thoughts on `mk*Plugin` having some logic to support `settingsOptions` being a file and/or function? It could automatically import/call the file/function so that we can just declare `settingsOptions = ./settings-options.nix`.

khaneliman:
> LGTM.
> 
> Any thoughts on `mk*Plugin` having some logic to support `settingsOptions` being a file and/or function? It could automatically import/call the file/function so that we can just declare `settingsOptions = ./settings-options.nix`.

Yeah, that sound simplify it a little.

khaneliman:
@mergify queue

PR #2959, opened 2025-02-02
plugins/typescript-tools: use plugin's luaConfig

khaneliman:
When lazy loading, we need to isolate the plugin's config otherwise it will fail because the generated config is still loaded immediately in lspconfig's lua.


khaneliman:
@mergify queue

PR #2962, opened 2025-02-03
plugins/nvim-ufo: Set lsp capabilities

FirelightFlagboy:
Following the [installation procedure of `nvim-ufo`](https://github.com/kevinhwang91/nvim-ufo/blob/main/README.md#minimal-configuration), it require to configure the LSP capabilities

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -64,6 +64,16 @@ in
 rec {
   inherit pluginDefaultText;
 
+  # Create an option declaration with a default value of `true`, and can be defined to `false`.
+  mkEnabledOption =
+    name:
+    lib.mkOption {
+      default = true;
+      example = false;
+      description = "Whether to enable ${name}.";
+      type = types.bool;
+    };
+
   # Creates an option with a nullable type that defaults to null.
   mkNullOrOption' =
     {
--- a/plugins/by-name/nvim-ufo/default.nix
+++ b/plugins/by-name/nvim-ufo/default.nix
@@ -130,4 +130,20 @@ lib.nixvim.plugins.mkNeovimPlugin {
         end
       '';
   };
+
+  extraOptions = {
+    setupLspCapabilities = lib.nixvim.options.mkEnabledOption "setup LSP capabilities for nvim-ufo";
+  };
+
+  extraConfig = cfg: {
+    plugins.lsp.capabilities =
+      lib.mkIf cfg.setupLspCapabilities # lua
+        ''
+          -- Capabilities configuration for nvim-ufo
+          capabilities.textDocument.foldingRange = {
+            dynamicRegistration = false,
+            lineFoldingOnly = true
+          }
+        '';
+  };
 }
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,6 @@
 [*.sh]
 indent_style = space
 indent_size = 2
+
+[*.nix]
+indent_size = 2
--- a/tests/test-sources/plugins/by-name/nvim-ufo/default.nix
+++ b/tests/test-sources/plugins/by-name/nvim-ufo/default.nix
@@ -90,4 +90,12 @@
       };
     };
   };
+
+  lsp-compat = {
+    plugins.nvim-ufo = {
+      enable = true;
+      setupLspCapabilities = true;
+    };
+    plugins.lsp.enable = true;
+  };
 }


FirelightFlagboy:
I was thinking of that, do you have a suggestion for the name of the option ?

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -130,4 +130,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
         end
       '';
   };
+
+  extraConfig = cfg: {
```
MattSturgeon, line 134: Nit:

```suggestion
  extraConfig = {
```

No need for the `cfg` param if it's unused.

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -130,4 +130,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+    plugins.lsp.capabilities = # lua
```
MattSturgeon, line 135: I agree this should be gated behind an option. But I'm fine for that option to be `default=true`.

Maybe `setupCapabilities`, `lspCapabilities`, `setupLspCapabilities` or something?

plugins/by-name/nvim-ufo/default.nix
MattSturgeon: Although it will be used if we're about to add a `mkIf`

FirelightFlagboy: Now `cfg` is used with the added option as suggested

plugins/by-name/nvim-ufo/default.nix
FirelightFlagboy: I've added the option `settings.setupLspCapabilities`

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -130,4 +132,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
         end
       '';
   };
+
+  extraConfig = cfg: {
+    plugins.lsp.capabilities =
+      lib.mkIf cfg.settings.setupLspCapabilities # lua
```
MattSturgeon, line 138:
```suggestion
      lib.mkIf cfg.setupLspCapabilities # lua
```

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -84,6 +84,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
       mappings = lib.nixvim.mkNullOrOption lib.types.attrs "Mappings for preview window.";
     };
+
+    setupLspCapabilities = lib.options.mkEnableOption "Setup LSP capabilities for nvim-ufo";
```
MattSturgeon, line 88: This should be moved from `settingsOptions` to `extraOptions`.

`settings` is exclusively for the _upstream plugin's_ options. Out custom nixvim-specific options should live outside of `settings`.

MattSturgeon: I'd also prefer it to be a `default=true` option.

`mkEnableOption "Setup LSP capabilities for nvim-ufo"` will produce:

```nix
mkOption {
  type = lib.types.bool;
  description = "Whether to enable Setup LSP capabilities for nvim-ufo.";
  default = false;
  example = true;
}
```

I'd suggest instead having:
```nix
mkOption {
  type = lib.types.bool;
  description = "Whether to enable setting-up LSP capabilities for nvim-ufo.";
  default = true;
  example = false;
}
```

plugins/by-name/nvim-ufo/default.nix
FirelightFlagboy: Done by putting the option in `extraOptions` :+1:

plugins/by-name/nvim-ufo/default.nix
FirelightFlagboy: I've added the function `mkEnabledOption` to `lib/options.nix` since I'm planning to fix another plugin with a similar issue

lib/options.nix
```diff
@@ -64,6 +64,16 @@ in
...
 
+  # Create an option declaration with a default value of `true`, and can be defined to `false`.
+  mkEnabledOption =
+    name:
+    lib.mkOption {
+      default = true;
+      example = false;
+      description = "Whether to enable ${name}.";
+      type = types.bool;
+    };
```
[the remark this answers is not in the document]
MattSturgeon: Also: sorry to nit-pick, but could you put this in a dedicated commit, before your plugin changes?

plugins/by-name/nvim-ufo/default.nix
```diff
@@ -130,4 +130,22 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       '';
   };
+
+  extraOptions = {
+    setupLspCapabilities = lib.nixvim.options.mkEnabledOption "setup LSP capabilities for nvim-ufo";
+  };
+
+  extraConfig =
+    cfg:
+    lib.mkIf cfg.enable {
```
MattSturgeon, line 140:
```suggestion
  extraConfig = {
```

No need to check `cfg.enable` as that is already done by mkNeovimPlugin

lib/options.nix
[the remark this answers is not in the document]
FirelightFlagboy: I've put the change in a dedicated commit, that all you've asked?

plugins/by-name/nvim-ufo/default.nix
FirelightFlagboy: done

MattSturgeon (approved):
LGTM, thanks!

If you want, you can fetch `upstream` and rebase on our main branch.


Otherwise, I can queue as-is and @nix-infra-bot will rebase for you :slightly_smiling_face:

MattSturgeon:
@mergify queue

PR #2970, opened 2025-02-05
ci/update: fix re-apply commit order

MattSturgeon:
We need to reverse the `rev-list` output so that the commits are re-applied in the correct order; oldest first.


khaneliman (approved):

MattSturgeon:
Thanks for approving @khaneliman 

This should be trivial, but just to be safe I'll wait until I can test it _in practice_ using #2971

MattSturgeon:
Ok, I'm happy it's working:

image

MattSturgeon:
@mergify queue

PR #2972, opened 2025-02-05
ci/update: allow disabling re-applying commits

MattSturgeon:
Most of the time we want to re-apply "manual"/"additional" commits from the already open PR, however it is possible we may wish to run the CI without doing so.

Add an on-by-default tickbox to toggle the `re_apply` step.


khaneliman (approved):

MattSturgeon:
@mergify queue

PR #2973, opened 2025-02-05
[24.11] ci/update: backport changes from `main`

MattSturgeon:
Backport changes from `main`.

This keeps the 24.11 workflow in-sync with the workflow on main.

The main feature being backported is re-applying additional commits from an already open update PR.

Note: the `on.schedule` event is ignored on non-default branches, so backporting that part of the workflow has no real effect (other than keeping the file content identical, which may aid future backports). This is also why I haven't backported the `update-other` workflow; doing so would be pointless since it'll only ever trigger on the default branch (`main`).

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,8 +1,8 @@
-name: update
+name: Update
 on:
-  # Runs every Saturday at noon
+  # Runs everyday at noon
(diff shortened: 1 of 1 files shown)


MattSturgeon:
@mergify queue

Issue #2915, opened 2025-01-27
[PLUGIN REQUEST] blink-copilot


mathjiajia:
| Field    | Description |
|----------|-------------|
| Plugin   | `blink-copilot`  | 
| Homepage | [blink-copilot](https://github.com/fang2hou/blink-copilot)  | 
| Nixpkgs  | `false` | 

## Extra Information

A configurable GitHub Copilot `blink.cmp` source, which offers superior performance over `copilot.lua` with rewritten native LSP interaction, even when using the official Vim plugin.

PR #2974, opened 2025-02-05
plugins/blink-copilot: init

khaneliman:
Resolves https://github.com/nix-community/nixvim/issues/2915

### DIFF
--- a/plugins/by-name/blink-copilot/default.nix
+++ b/plugins/by-name/blink-copilot/default.nix
@@ -1,0 +1,62 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "blink-copilot";
+  package = "blink-copilot";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    This plugin should be configured through blink-cmp's `sources.providers` settings.
+
+    For example:
+
+    ```nix
+    plugins.blink-cmp = {
+      enable = true;
+      settings.sources.providers = {
+        copilot = {
+          async = true;
+          module = "blink-copilot";
+          name = "copilot";
+          score_offset = 100;
+          # Optional configurations
+          opts = {
+            max_completions = 3;
+            max_attempts = 4;
+            kind = "Copilot";
+            debounce = 750;
+            auto_refresh = {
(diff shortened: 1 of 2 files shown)


plugins/by-name/blink-copilot/default.nix
```diff
@@ -0,0 +1,62 @@
...
+
+  description = ''
+    This plugin should be configured through blink-cmp's source settings.
+
+    For example:
+
+    ```nix
+    plugins.blink-cmp = {
+      enable = true;
+      settings.sources = {
```
MattSturgeon, line 16:
```suggestion
      settings.sources.providers = {
```

Correct me if I'm wrong, but don't cmp [providers](https://cmp.saghen.dev/configuration/sources.html#providers) get registered under `sources.providers`?

plugins/by-name/blink-copilot/default.nix
```diff
@@ -0,0 +1,62 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "blink-copilot";
+  package = "blink-copilot";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    This plugin should be configured through blink-cmp's source settings.
```
MattSturgeon, line 9:
```suggestion
    This plugin should be configured through blink-cmp's `sources.providers` settings.
```

plugins/by-name/blink-copilot/default.nix
```diff
@@ -0,0 +1,62 @@
...
+              backward = true;
+              forward = true;
+            };
+          };
+        };
+      };
+    };
+    ```
+
+    And then you can add it as a source for blink-cmp:
```
MattSturgeon, line 38:
```suggestion
    And then you can add it to blink-cmp's `default` sources:
```
?

plugins/by-name/blink-copilot/default.nix
khaneliman: Yes, thank you. Just copied the other one and renamed stuff.

MattSturgeon (approved):

khaneliman:
@mergify queue

PR #2975, opened 2025-02-05
plugins/blink-emoji: init

khaneliman:


plugins/by-name/blink-emoji/default.nix
```diff
@@ -0,0 +1,50 @@
...
+          # Optional configurations
+          opts = {
+            insert = true;
+          };
+        };
+      };
+    };
+    ```
+
+    And then you can add it to blink-cmp's `default` sources:
```
MattSturgeon, line 30: Q: is "`default` source" or "`default` sources" more appropriate here?

MattSturgeon: Or maybe we just say "`sources.default` setting"?

khaneliman: Probably best to just pass the option name because both wording is weird imo

MattSturgeon (approved):

PR #2981, opened 2025-02-06
lib/modules: init `mkBlinkPluginModule`

MattSturgeon:
Initial proof of concept for how we could implement per-plugin auto-integration of blink completion providers.

So far I've drafted out some functions to produce the relevant option/modules, but I have not applied this to any plugins yet or tested it out. Posting this now for early design feedback and to discuss the concept in general.


lib/modules.nix
```diff
@@ -52,6 +53,286 @@ in
...
+      enableDefault ? true,
+      enableCmdline ? false,
+      enabledFiletypes ? { },
+      # Whether to include a `blink` option at all
+      offerBlinkCompatibility ? true,
+      # Defaults for the blink compatibility option
+      enableBlinkProvider ? false,
+      enableBlinkDefault ? false,
+      enableBlinkCmdline ? false,
+      enabledBlinkFiletypes ? { },
```
khaneliman, line 213: Should we just use the corresponding arguments above ?

MattSturgeon: Not sure what you're asking. These are probably badly named, but they're intended to be the default values for the corresponding module options.

MattSturgeon: Ah.

Because it felt unlikely that enabling a plugin as a nvim-cmp source by default should also enable it as a blink source by default.

We could have these defaults to the nvim-cmp equivalent, but I think it should still be possible to set these separately?

khaneliman:
Just checking branch out to test stuff out: 

```bash
… while evaluating the option `plugins.blink-cmp.settings.sources.providers.copilot.async':

       error: The option `plugins.blink-cmp.settings.sources.providers.copilot.async` is defined both null and not null, in `/nix/store/k3vjx39q6ksdh9m1yq9n9ikscd5d6387-source/plugins/by-name/blink-copilot' and `/nix/store/v309l0bfp6jis58v9v6a6b8rica9c72z-source/modules/nixvim/plugins/blink'.
```

lib/modules.nix
```diff
@@ -52,6 +53,286 @@ in
...
+                  (import ../plugins/by-name/blink-cmp/provider-config.nix lib)
+                ];
+              };
+            };
+          };
+        };
+      };
+      config = lib.mkIf (pluginCfg.enable && cfg.enable) {
+        plugins.blink-cmp.settings.sources = {
+          providers.${cfg.key} = cfg.settings;
```
khaneliman, line 178:
```suggestion
          providers.${cfg.key} = lib.mkDefault cfg.settings;
```

Probably want to make them default to not break existing configs and support overriding from blink module?

MattSturgeon:
> Just checking branch out to test stuff out: 
> 
> ```bash
> … while evaluating the option `plugins.blink-cmp.settings.sources.providers.copilot.async':
> 
>        (stack trace truncated; use '--show-trace' to show the full, detailed trace)
> 
>        error: The option `plugins.blink-cmp.settings.sources.providers.copilot.async` is defined both null and not null, in `/nix/store/k3vjx39q6ksdh9m1yq9n9ikscd5d6387-source/plugins/by-name/blink-copilot' and `/nix/store/v309l0bfp6jis58v9v6a6b8rica9c72z-source/modules/nixvim/plugins/blink'.
> ```

Ah, that'll be because I'm copying `settings` around without preserving the original definition priority.

lib/modules.nix
MattSturgeon: > Just checking branch out to test stuff out: 

Anything you're playing with that we could add as test cases?

> ```bash
> … while evaluating the option `plugins.blink-cmp.settings.sources.providers.copilot.async':
> 
>        (stack trace truncated; use '--show-trace' to show the full, detailed trace)
> 
>        error: The option `plugins.blink-cmp.settings.sources.providers.copilot.async` is defined both null and not null, in `/nix/store/k3vjx39q6ksdh9m1yq9n9ikscd5d6387-source/plugins/by-name/blink-copilot' and `/nix/store/v309l0bfp6jis58v9v6a6b8rica9c72z-source/modules/nixvim/plugins/blink'.
> ```

Ah, that'll be because I'm copying `settings` around without preserving the original definition priorities.

Should be fixed by using [`mkAliasDefinitions`](https://github.com/NixOS/nixpkgs/blob/ba2d98239e995de78369da2a9e280ecc3c402e50/lib/modules.nix#L1076-L1097)

tests/test-sources/plugins/cmp/all-sources.nix
MattSturgeon, line 1: Do we need an all-sources test for blink.cmp too? Unless that is covered by each plugin's own tests?

lib/modules.nix
khaneliman: Just using my neovim flake, as-is. Which enables the plugins but configures them through blink.

khaneliman:
Latest checkout:

```bash
… while evaluating attribute 'paths' of derivation 'nixvim'
         at /nix/store/gk2yzlr4wnjflnr8r8fddbjffy4jq38y-source/modules/top-level/output.nix:384:11:
          383|           name = "nixvim";
          384|           paths =
             |           ^
          385|             with config.build;

       … while evaluating the example of option `plugins.cmp-ai.cmp.cmdline`

       error: expected a Boolean but found a set: { }
       at /nix/store/gk2yzlr4wnjflnr8r8fddbjffy4jq38y-source/lib/modules.nix:313:28:
          312|                 default = enableCmdline;
          313|                 example = !enableCmdline;
          314|                 description = ''
```

MattSturgeon:
> ```
>        error: expected a Boolean but found a set: { }
>        at /nix/store/gk2yzlr4wnjflnr8r8fddbjffy4jq38y-source/lib/modules.nix:313:28:
>           312|                 default = enableCmdline;
>           313|                 example = !enableCmdline;
>              |                            ^
>           314|                 description = ''
> ```

Oops, leftover from when that option was type=bool. Didn't pick up on it as I'm not building the docs on my laptop.

Should be resolved.

khaneliman:
> > ```
> >        error: expected a Boolean but found a set: { }
> >        at /nix/store/gk2yzlr4wnjflnr8r8fddbjffy4jq38y-source/lib/modules.nix:313:28:
> >           312|                 default = enableCmdline;
> >           313|                 example = !enableCmdline;
> >              |                            ^
> >           314|                 description = ''
> > ```
> 
> Oops, leftover from when that option was type=bool. Didn't pick up on it as I'm not building the docs on my laptop.
> 
> Should be resolved.

Looks good for generating config same as before. Going to try using the new config style.

MattSturgeon:
I'm really struggling to have a warning/assertion for when the new & old systems are used in tandem.

`autoEnableCmpSources` is already so prone to inf-recursion, it is almost impossible.

Basically the same limitations that caused us to put the `cmpSourcePlugins` outside of `plugins`; you cannot dynamically define `plugins.*` attrs while using `plugins.<something>` to determine what attr names to define without running into inf-recursion. We also cannot have one impl writing to `settings.sources` when a plugin is enabled, while another impl enables plugins when their name is in `settings.sources`.

So we have two different forms of inf-rec here and I can't think of a way to avoid one without running into the other.

khaneliman:
> Looks good for generating config same as before. Going to try using the new config style.

Trying to enable blink-copilot configuration from it's own settings doesn't seem to generate the corresponding provider config. I see it's inserting the value into the defaults attributes, but it's not in the provider's attributes.

MattSturgeon:
> I'm really struggling to have a warning/assertion for when the new & old systems are used in tandem.

Managed to resolve the inf-rec, so now using the old+new impl together produces a sane (tested) warning.

I've also cleaned up a few more bits and fixed some tests that weren't enabling blink.

> EDIT: Nvm, removed all my sources config and it only generates the default sources.

Could you check if you're still running into any issues? It's possible the migration path for anyone manually setting up blink providers isn't perfect, maybe some provider settings will have merge conflicts for example?

MattSturgeon:
> Trying to enable blink-copilot configuration from it's own settings doesn't seem to generate the corresponding provider config. I see it's inserting the value into the defaults list, but it's not in the provider's attributes. ~Assuming it's because we're aliasing the entire existing provider config and not at a more granular level. ( I dont know if we can do anything about that, really.. :S )~
> 
> EDIT: Nvm, removed all my sources config and it only generates the default sources.

Reproduced, fixed, and added a test case.

Turned out that `mkAliasDefinitions` doesn't pick up on definitions that were already defined _within_ a submodule-type, because those definitions don't show up in the _option's_ `definitions`. Essentially such definitions belong to the _type_ rather than the _option_.

MattSturgeon:
I've included deprecating and removing `autoEnableSources` in this PR.

I spent some time looking for a way to gracefully transition with a warning rather than an error, however the problem is fundamentally circular and cannot be solved without disabling/removing one strategy over the other.

One concern I still have with this PR is how "auto enable" should be handled when the user wants `sources` to be raw-lua. Maybe instead of plugins directly adding themselves to the `settings` options, we should introduce an intermediate `internal` option and a global toggle? Such an approach could start getting cumbersome quickly, though. Perhaps its fine for users to simply disable `plugins.*.cmp.enable` when they want to use raw-lua?

lib/modules.nix
```diff
@@ -52,6 +53,282 @@ in
...
+          blink.settings = {
+            name = lib.mkDefault sourceName;
+            inherit module;
+            opts = lib.mkIf usePluginSettings (lib.modules.mkAliasDefinitions pluginOpts.settings);
+          };
+        })
+        (lib.mkIf (pluginCfg.enable && cfg.enable) {
+          plugins.blink-cmp.settings.sources = {
+            # Use mkAliasDefinitions to preserve override priorities
+            providers.${cfg.key} = lib.modules.mkAliasDefinitions opt.settings;
```
khaneliman, line 174: This doesn't work the way we need it to to support blink's configuration.

```nix
      blink-copilot = {
        enable = true;

        # Need these outside of  the provider's opts
        # But don't exist in the module
        async = true;
        score_offset = 100;

        # Freeform accepts, but goes into `opts` instead of `providers.${cfg.key}` top level
        settings = {
          async = true;
          score_offset = 100;
        };
      };
```

```lua
copilot = {
    module = "blink-copilot",
    name = "copilot",
    opts = { async = true, score_offset = 100 },
}
```

plugins/cmp/default.nix
```diff
@@ -17,48 +17,55 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      settings.sources = [
-        { name = "nvim_lsp"; }
-        { name = "path"; }
-        { name = "buffer"; }
-      ];
+    plugins = {
+      cmp.enable = true;
+      cmp-nvim-lsp.enable = true;
+      cmp-path.enable = true;
+      cmp-buffer.enable = true;
```
khaneliman, line 34:
```suggestion
```

tests/test-sources/example-configurations/issues.nix
```diff
@@ -125,23 +124,23 @@
...
+    cmp-path.enable = true;
+    cmp-buffer.enable = true;
+    cmp-nvim-lsp.enable = true;
+    cmp_luasnip = {
+      enable = true;
+      cmp.default = {
+        option.show_autosnippets = true;
+      };
+    };
+
```
khaneliman, line 143: These look outside the plugin block

lib/modules.nix
```diff
@@ -52,6 +53,282 @@ in
...
+      };
+      config = lib.mkMerge [
+        (lib.setAttrByPath loc {
+          # NOTE: this could be defined within the `blink.settings` submodule,
+          # but that would not populate the option's `definitions` list.
+          # Meaning we wouldn't be able to propagate the definitions further using `mkAliasDefinitions`.
+          blink.settings = {
+            name = lib.mkDefault sourceName;
+            inherit module;
+            opts = lib.mkIf usePluginSettings (lib.modules.mkAliasDefinitions pluginOpts.settings);
```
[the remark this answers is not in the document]
khaneliman: Yeah... so this looks like we should inherit `pluginOpts.settings` for the known top level options and then opts inherits from `pluginOpts.settings.opts`?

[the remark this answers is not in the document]
MattSturgeon: `plugins.*.settings` has always represented the `setup` function's `opts`, so I've attempted to be consistent with that here.

The only difference is that blink invokes the `setup` function for us.

_Provider_ settings can still be configured directly, via `plugins.*.blink.settings`.

It was intentional to keep these separate, but I can see how it could be confusing if you take a different perspective to me.

🤔

lib/modules.nix
```diff
@@ -52,6 +53,314 @@ in
...
+                type = lib.types.submodule [
+                  {
+                    options = {
+                      enabled = internal;
+                      module = internal;
+                    };
+                    config = {
+                      name = lib.mkDefault sourceName;
+                      inherit module;
+                      # One-way alias for `plugin.*.settings` -> `plugin.*.blink.settings.opts`
```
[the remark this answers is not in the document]
MattSturgeon: The original comment was correct.

`plugins.*.blink.settings` is itself aliased to `plugins.blink-cmp.settings.sources.providers.${cfg.blink.name}` elsewhere.

So it's effectively a two-step alias.

tests/test-sources/example-configurations/issues.nix
MattSturgeon: `plugins.*.cmp.default` maps to `plugins.cmp.settings.sources [ (it // { inherit (cfg.cmp) name; }) ]`

lib/modules.nix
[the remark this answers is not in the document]
khaneliman: > `plugins.*.settings` has always represented the `setup` function's `opts`, so I've attempted to be consistent with that here.
> 
> The only difference is that blink invokes the `setup` function for us. But that's an implementation detail that is encapsulated from users in both normal and blink plugins...
> 
> _Provider_ settings can still be configured directly, via `plugins.*.blink.settings`.

Doesn't make this feel like a gain UX wise if they need to split their provider configuration between 2 modules, though. Personally, I'd just keep my configuration through `blink-cmp` so I wouldn't have to jump back and forth between 2 attribute sets to see what i'm doing. 

> It was intentional to keep these separate, but I can see how it could be confusing if you take a different perspective to me.
> 
> 🤔

I'd perfer providing the top level options to each blink plugin if we want to keep `settings` forwarded specifically to `opts`. So someone can choose to configure their provider from each source plulgin module itself.

MattSturgeon:
I've noticed _some_ plugins that bundle nvim-cmp sources have their own settings options for enabling the source.

e.g. codeium-nvim

In these cases should we _not_ declare our `plugins.*.cmp` options? Or handle them somehow differently?

I'm guessing we'd still want `blink.compat` for most of these sources, too...

plugins/cmp/deprecated-auto-enable.nix
```diff
@@ -0,0 +1,141 @@
...
+      (
+        cfg.enable
+        && !pluginCfg.enable
+        && pluginOpt.enable.highestPrio == 1500
+        && pluginCfg.cmp.enable
+        && sourceDefs != { }
+      )
+      (
+        lib.nixvim.mkWarnings (lib.showOption pluginLoc) ''
+          The ${builtins.toJSON pluginCfg.cmp.name} nvim-cmp source has been defined via `${lib.showOption cmpLoc}`, howevew `${pluginOpt.enable}` is not enabled.
```
MattSturgeon, line 100: howevew

```suggestion
          The ${builtins.toJSON pluginCfg.cmp.name} nvim-cmp source has been defined via `${lib.showOption cmpLoc}`, however `${pluginOpt.enable}` is not enabled.
```

---

Also, I wonder if we can improve this warning by saying the actual settings option(s) where the source was enabled? Instead of just using `cmpLoc`.

This should be possible to work out from the `sourceDefs` attrset.

PR #2984, opened 2025-02-06
plugins/schemastore: set the correct jsonls option

cswimr:
The JSON language server has the following [configuration structure](https://github.com/microsoft/vscode/blob/6fef251f8f0e9f74b7adb4ed310e35eabb97afdf/extensions/json-language-features/server/src/jsonServer.ts#L167-L179):
```ts
    // The settings interface describes the server relevant settings part
    interface Settings {
        json?: {
            schemas?: JSONSchemaSettings[];
            format?: { enable?: boolean };
            validate?: { enable?: boolean };
            resultLimit?: number;
        };
        http?: {
            proxy?: string;
            proxyStrictSSL?: boolean;
        };
    }
```
So, we should be mkDefault'ing `validate.enable` instead of `validate`, as `validate` is expected to be an `Object` containing `enable` as a boolean.
This is also reflected in the schemastore readme, which offers this configuration example:
```lua
require('lspconfig').jsonls.setup {
  settings = {
    json = {
      schemas = require('schemastore').json.schemas(),
      validate = { enable = true },
    },
  },
}
```
Whereas nixvim currently does the following:
image

### DIFF
--- a/plugins/by-name/schemastore/default.nix
+++ b/plugins/by-name/schemastore/default.nix
@@ -145,7 +145,7 @@ lib.nixvim.plugins.mkVimPlugin {
 
         # The plugin recommends to enable this option in its README.
         # Learn more: https://github.com/b0o/SchemaStore.nvim/issues/8
-        validate = mkDefault true;
+        validate.enable = mkDefault true;
       };
 
       yamlls.settings = mkIf cfg.yaml.enable {


MattSturgeon (approved):
Thanks!

MattSturgeon:
@mergify queue

MattSturgeon:
> Is there any chance of this being backported to `24.11`?

We don't often backport things. But I'll try.

> Failing that, is there any easy way of incorporating this in my own config?

We haven't documented this much yet, but you can use nixvim's main branch on a stable system, so long as you allow nixvim to use an unstable nixpkgs. Allowing nixvim to manage its own nixpkgs instance is the easiest way to achieve this, which is the default behaviour.

So as long as you don't change `programs.nixvim.nixpkgs.useGlobalPackages` or set a `inputs.nixpkgs.follows` pointing to an incompatible nixpkgs revisions, you shouldn't have to _do_ anything.

MattSturgeon:
@mergify backport nixos-24.11

PR #2999, opened 2025-02-10
plugins/easy-dotnet: init

### LINKED ISSUE #2826: [PLUGIN REQUEST] easy-dotnet.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | easy-dotnet.nvim  | 
| Homepage | [easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim)    | 
| Nixpkgs  | `false` | 

## Extra Information

Plugin providing essential IDE features for dotnet developers

GaetanLepage:
Add support for [easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim), support for .Net projects in Neovim.

Fixes #2826

### DIFF
--- a/plugins/by-name/easy-dotnet/default.nix
+++ b/plugins/by-name/easy-dotnet/default.nix
@@ -1,0 +1,54 @@
+{ lib, config, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "easy-dotnet";
+  packPathName = "easy-dotnet.nvim";
+  package = "easy-dotnet-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  settingsExample = lib.literalExpression ''
+    {
+      test_runner = {
+        enable_buffer_test_execution = true;
+        viewmode = "float";
+      };
+      auto_bootstrap_namespace = true;
+      terminal.__raw = '''
+        function(path, action, args)
+          local commands = {
+            run = function()
+              return string.format("dotnet run --project %s %s", path, args)
+            end,
+            test = function()
+              return string.format("dotnet test %s %s", path, args)
+            end,
+            restore = function()
+              return string.format("dotnet restore %s %s", path, args)
+            end,
+            build = function()
+              return string.format("dotnet build %s %s", path, args)
+            end,
+          }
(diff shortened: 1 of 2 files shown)


plugins/by-name/easy-dotnet/default.nix
```diff
@@ -0,0 +1,52 @@
...
+
+  settingsExample = {
+    test_runner = {
+      enable_buffer_test_execution = true;
+      viewmode = "float";
+    };
+    auto_bootstrap_namespace = true;
+    terminal.__raw = ''
+      function(path, action, args)
+        local commands = {
```
MattSturgeon, line 17: the docs will render this horribly:

```
terminal = {
  __raw = "function(path, action, args)\n  local commands = {\n";
}
```

This is because `lib.generators.toPretty` will use `builtins.toJSON` to render any string values.

I'd recommend using a literalExpression for the example this time

plugins/by-name/easy-dotnet/default.nix
```diff
@@ -0,0 +1,52 @@
...
+      end
+    '';
+    picker = "fzf";
+  };
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.easy-dotnet" (
+      lib.mapAttrsToList
+        (pickerName: pluginName: {
+          when = (cfg.settings.picker or null == pickerName) && (!config.plugins.${pluginName}.enable);
```
MattSturgeon, line 43: This doesn't offer any way to suppress the warning if end-users explicitly don't want to enable our plugins.

Maybe that's not too important, as it seems like a contrived scenario.

But we could do something like:
```nix
when =
  cfg.settings.picker or null == pickerName
  && !config.plugins.${pluginName}.enable
  && options.plugins.${pluginName}.enable.highestPrio == 1500;
```

And say:
```
You can suppress this warning by explicitly setting `plugins.${pluginName}.enable = false`.
```

MattSturgeon (commented):
Looks good to me, but I think the `example` should be a literal expression this time

MattSturgeon (approved):

PR #3010, opened 2025-02-13
docs/fix-links: init

MattSturgeon:
The README/CONTRIBUTING files are authored with GitHub in mind, but we re-use them for the docs website.

We can use pandoc's AST-based [lua filters](https://pandoc.org/lua-filters.html) to modify link targets.

Knowing we are dealing with a link target allows us to also handle things like relative links, meaning we can fixup links targeting the github repo such that they also work from the docs website.

- Replace README's substitution with the new impl
- Also fix links in the CONTRIBUTING file

### DIFF
--- a/docs/fix-links/filter.lua
+++ b/docs/fix-links/filter.lua
@@ -1,0 +1,40 @@
+local len = pandoc.text.len
+local sub = pandoc.text.sub
+
+-- True if str starts with prefix
+local function hasPrefix(prefix, str)
+	local pfxLen = len(prefix)
+	local strLen = len(str)
+	if pfxLen == strLen then
+		return prefix == str
+	end
+	if pfxLen < strLen then
+		return prefix == sub(str, 1, pfxLen)
+	end
+	return false
+end
+
+function Link(link)
+	local target = link.target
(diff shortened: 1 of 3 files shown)


traxys (approved):
Indeed much more robust!

MattSturgeon:
@mergify queue

docs/fix-links/filter.lua
```diff
@@ -0,0 +1,40 @@
...
+	-- Check for relative links
+	-- TODO: handle ../
+	if hasPrefix("./", target) then
+		link.target = githubUrl .. sub(target, 3)
+		return link
+	end
+	if not hasPrefix("https://", target) then
+		link.target = githubUrl .. target
+		return link
+	end
```
MattSturgeon, line 28: This doesn't correctly handle `#` anchor targets on the same page.

We need a case that checks `hasPrefix("#")`

traxys: Wasn't a very effective review on my part

MattSturgeon: Don't be silly. These kinda things slip through easily and it's no worse than it was before

PR #3011, opened 2025-02-14
docs/fix-links: handle `#anchor` targets on the same page

MattSturgeon:
Follow up to https://github.com/nix-community/nixvim/pull/3010, in particular it resolves https://github.com/nix-community/nixvim/pull/3010#discussion_r1955547292.


MattSturgeon:
@mergify queue

Issue #3012, opened 2025-02-15
[BUG] Cannot set `ia`, `ca`, `!a`. as keymap modes in NixVim


npg418:
| Field        | Description |
|--------------|-------------|
| Plugin       | `None`  | 
| Nixpkgs      | `24.11` | 
| Home Manager | `24.11` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description  

Currently, when trying to define keymaps in NixVim, specifying certain modes such as `ia`, `ca`, or `!a` is not supported. However, these are valid mode specifiers in Neovim, as documented in [`nvim_set_keymap()`](https://neovim.io/doc/user/api.html#nvim_set_keymap()).  

It would be helpful if these modes could be properly supported in `programs.nixvim.keymaps`. Alternatively, introducing a separate property for abbreviations could be a solution.  

## Minimal, Reproducible Example (MRE)  

```nix
{
  programs.nixvim = {
    keymaps = [
      {
        mode = "ca";  # Does not work
        key = "s";
        action = "getcmdtype() .. getcmdline() ==# ':s' ? [getchar(), ''][1] .. \"%s///g<Left><Left>\" : 's'";
        options.expr = true;
      }
    ];
  };
}
```

This configuration results in an error or does not apply the keymap correctly.

Would it be possible to support these keymap modes directly, or provide an alternative way to handle abbreviations in NixVim?

npg418:
No, it's not combinations of modes. Mode specifier like `ca` means abbreviations.
In neovim, command abbreviations can be created by writing
```lua
vim.keymap.set("ca", "key", "action")
```
which equivalent of
```vimscript
cabbrev key action
```
in vimscript

MattSturgeon:
Ah I see, you're right. Those abbrev modes are simply not permitted by the emum type we're using. Thanks for reporting!

I'll open a PR to add support.

MattSturgeon:
This is fixed on nixvim's `main` branch. It won't be backported to our 24.11 branch, however it is possible to use nixvim's `main` branch on a 24.11 system if you allow nixvim to manage its own `pkgs` instance (which is the default behaviour).

PR #3015, opened 2025-02-15
{readme,lib/keymaps}: cleanup `modes` docs/impl

MattSturgeon:
- **readme: use `:h map-table` for mapping modes**
- **lib/keymaps: replace `modes` attrs with list**

Small cleanup in preparation for fixing #3012


MattSturgeon:
@mergify queue

PR #3016, opened 2025-02-15
lib/keymaps: add abbreviation support to `modes` enum

### LINKED ISSUE #3012: [BUG] Cannot set `ia`, `ca`, `!a`. as keymap modes in NixVim

MattSturgeon:
See https://neovim.io/doc/user/usr_24.html#_abbreviations
and https://learnvimscriptthehardway.stevelosh.com/chapters/08.html

Fixes #3012

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,8 +427,13 @@ You can provide several modes to a single mapping by using a list of strings.
 | `"o"`  |  -   |  -  |  -  |  -  |  -  | yes |  -   |  -   | Operator-pending mode |
 | `"t"`  |  -   |  -  |  -  |  -  |  -  |  -  | yes  |  -   | Terminal mode |
 | `"l"`  |  -   | yes | yes |  -  |  -  |  -  |  -   | yes  | Insert, command-line and lang-arg mode |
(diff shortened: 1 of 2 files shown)


MattSturgeon:
@mergify queue

PR #3017, opened 2025-02-15
lib/keymaps: make mode type description more readable

MattSturgeon:
- **lib/keymaps: add a link to `:h map-modes`**
  
- **lib/keymaps: make `mode` type's description more readable**
  Only list the enum values once, instead of twice.
  
image


MattSturgeon:
@mergify queue

PR #3020, opened 2025-02-16
plugins/lsp: enable dolmenls and ocamllsp

stepbrobd:
Not ready yet, flake.lock still tracking old nixpkgs (I'll rebase after a lock file update)

- dolmenls added in https://github.com/NixOS/nixpkgs/pull/380805
- ocamllsp was packaged long time ago

### DIFF
--- a/plugins/lsp/lsp-packages.nix
+++ b/plugins/lsp/lsp-packages.nix
@@ -42,7 +42,6 @@
     "debputy"
     "delphi_ls"
     "djlsp"
-    "dolmenls"
     "drools_lsp"
     "ds_pinyin_lsp"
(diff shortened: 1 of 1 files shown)


stepbrobd:
Also quick question on https://github.com/nix-community/nixvim/commit/c15e93c157277c4d9e53336ffe04ff15834a5f78, coq_lsp works just fine with package specified, why marking it as unavailable?

And not sure if this is a typo, yamlls shouldn't be marked as unavailable due to coq update?

stepbrobd:
sheesh dumb me forgot i have an override on nixpkgs 😂
thx!

PR #3029, opened 2025-02-22
flake: partition dev dependencies

### LINKED ISSUE #2948: New experimental nixvim configurations fail without devshell input
```
error:
       … in the left operand of the update (//) operator
         at «nix-internal»/call-flake.nix:69:13:
           68|             # This is shadowed in the next //
           69|             // sourceInfo
             |             ^
           70|             // {

       … while evaluating the attribute 'config.flake'
         at /nix/store/afzapn9d41mhkpw475dcaxwrmcz5yr3m-source/lib/modules.nix:334:9:
          333|         options = checked options;
          334|         config = checked (removeAttrs config [ "_module" ]);
             |         ^
          335|         _module = checked (config._module);

       error: stack overflow; max-call-depth exceeded
```

### LINKED ISSUE #3027: Partition flake inputs so they are fetched lazily
Currently our flake is being bloated by devshells and dev tools; these are useful to us but not to end users.

The same could be said for our `checks`, however due to buildbot limitations they can't be moved to another flake.

Currently, we work around this by allowing users to disable some of our flake inputs using `follows`. However IMO this is a clunky and fragile workaround.

See also https://github.com/nix-community/nixvim/issues/2948#issuecomment-2672966951

For prior art, flake.parts for example uses a `./dev` flake in order to keep dev tools separate to the user-facing flake: https://github.com/hercules-ci/flake-parts/tree/main/dev

MattSturgeon:
## Summary

By _partitioning_ our "dev" dependencies, they will not be added to end-user's lockfiles. Instead, they will be fetched lazily when flake outputs that _actually_ need them are evaluated.

E.g. evaluating `nixvim.nixvimModules` will not fetch dev dependencies, however evaluating `nixvim.checks` or `nixvim.devShells` will.

## Motivation

This removes the need for end-users to manually set `nixvim.inputs.devshell.follows = ""` (etc). In fact, these optional inputs are no longer part of the root flake at all. I believe any user that currently has such inputs will get an error or warning notifying them that they are following a non-existent input.

## Technical explanation

We offload evaluation of some of our flake modules into a "partition" submodule. When its not needed, this submodule is not evaluated. When it is needed, it fetches extra inputs from `flake/dev/flake.nix` as part of evaluating the submodule.

The submodule extends the top-level (root) configuration, so any root inputs and configuration are also present in the submodule's `config` and `inputs` args.

We use another `flake.nix` and `flake.lock` file (in `flake/dev`). This is purely to manage inputs, so that we can run `nix flake update --commit-lock-file --flake ./flake/dev`. The dev flake does not have any outputs of its own.

See reference docs: https://flake.parts/options/flake-parts-partitions.html

## Caveats

The only disadvantage of this approach is we now have two lockfiles to keep up to date.

The dev flake (`flake/dev/flake.nix`) has inputs that transitively depend on nixpkgs. In order to avoid depending on multiple nixpkgs revisions, we therefore need to specify nixpkgs as an input in both our root and dev flakes.

This is a little awkward, as we have to be careful when updating lock files that both stay in sync.

Eventually, we may be able to use relative paths to do something like:
```nix
{
  inputs = {
    nixvim.url = "path:../..";
    nixpkgs.follows = [ "nixvim" "inputs" "nixpkgs" ];
  };
}
```

However this depends on https://github.com/NixOS/nix/pull/10089 making it to a stable nix version, ideally making it to `nixVersions.minimum` which may take some time. I'm unsure how `lix` deals with relative path flake inputs, too.

For now, we can ensure thing stay in sync by carefully using `--override-input` when updating the dev lockfile.

## Not covered

The `nuschtosSearch` input was not (and still is not) optional. I'd like to change that at some point, but it will need some restructuring of the docs package set (`./docs/default.nix`) and is not in scope for this PR. See https://github.com/hercules-ci/flake-parts/pull/272#issuecomment-2676984478

----

Fixes #3027
Resolves #2948 :crossed_fingers:, by working-around whatever issue was causing that.

cc @HeitorAugustoLN for testing

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -6,10 +6,14 @@ on:
   # Allow manual triggering
   workflow_dispatch:
     inputs:
-      lock:
+      root_lock:
         type: boolean
         default: true
-        description: Update flake.lock
+        description: Update root flake.lock
+      dev_lock:
+        type: boolean
+        default: true
+        description: Update dev flake.lock
       generate:
         type: boolean
         default: true
@@ -100,12 +104,28 @@ jobs:
           git fetch origin "$pr_branch"
           git branch --set-upstream-to "origin/$pr_branch"
 
-      - name: Update flake.lock
-        id: flake_lock
-        if: inputs.lock || github.event_name == 'schedule'
+      - name: Update root flake.lock
+        id: root_flake_lock
+        if: inputs.root_lock || github.event_name == 'schedule'
         run: |
           old=$(git show --no-patch --format=%h)
           nix flake update --commit-lock-file
+          new=$(git show --no-patch --format=%h)
+          if [ "$old" != "$new" ]; then
+            echo "body<<EOF" >> "$GITHUB_OUTPUT"
+            git show --no-patch --format=%b >> "$GITHUB_OUTPUT"
+            echo "EOF" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Update dev flake.lock
+        id: dev_flake_lock
+        if: inputs.dev_lock || github.event_name == 'schedule'
+        run: |
+          root_nixpkgs=$(nix eval -f . 'inputs.nixpkgs.rev')
+          old=$(git show --no-patch --format=%h)
+          nix flake update --commit-lock-file \
+              --override-input nixpkgs "github:NixOS/nixpkgs/$root_nixpkgs" \
+              --flake './flake/dev'
           new=$(git show --no-patch --format=%h)
           if [ "$old" != "$new" ]; then
             echo "body<<EOF" >> "$GITHUB_OUTPUT"
@@ -185,9 +205,14 @@ jobs:
           title: |
             [${{ github.ref_name }}] Update flake.lock & generated files
           body: |
-            ## Flake lockfile
+            ## Root lockfile
             ```
-            ${{ steps.flake_lock.outputs.body || 'No changes' }}
+            ${{ steps.root_flake_lock.outputs.body || 'No changes' }}
+            ```
+
+            ## Dev lockfile
+            ```
+            ${{ steps.dev_flake_lock.outputs.body || 'No changes' }}
             ```
 
             ## Generate
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
 (import (
   let
-    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    lock = builtins.fromJSON (builtins.readFile ./flake/dev/flake.lock);
   in
   fetchTarball {
     url =
--- a/flake.nix
+++ b/flake.nix
@@ -9,59 +9,9 @@
       inputs.nixpkgs-lib.follows = "nixpkgs";
     };
 
-    /*
-      # NOTE: The inputs below this comment are optional
-      # You can remove them with `inputs.<input>.follows = ""`
-
-      For example:
-
-      ```
-      nixvim = {
-        url = "github:nix-community/nixvim";
-        inputs = {
-          nixpkgs.follows = "nixpkgs";
-          flake-parts.follows = "flake-parts";
-
-          devshell.follows = "";
-          flake-compat.follows = "";
-          git-hooks.follows = "";
-          home-manager.follows = "";
-          nix-darwin.follows = "";
-          treefmt-nix.follows = "";
-        };
-      };
-      ```
-    */
-
     nuschtosSearch = {
       url = "github:NuschtOS/search";
       inputs.nixpkgs.follows = "nixpkgs";
-    };
-
-    home-manager = {
-      url = "github:nix-community/home-manager";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    nix-darwin = {
-      url = "github:lnl7/nix-darwin";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-
-    devshell = {
-      url = "github:numtide/devshell";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    treefmt-nix = {
-      url = "github:numtide/treefmt-nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-
-    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
-
-    git-hooks = {
-      url = "github:cachix/git-hooks.nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-      inputs.flake-compat.follows = "flake-compat";
     };
   };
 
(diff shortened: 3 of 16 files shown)


.github/workflows/update.yml
```diff
@@ -113,6 +121,38 @@ jobs:
...
+          old=$(git show --no-patch --format=%h)
+          nix flake update --commit-lock-file \
+              --override-input nixpkgs "github:NixOS/nixpkgs/$root_nixpkgs" \
+              --flake './flake/dev'
+          new=$(git show --no-patch --format=%h)
+          if [ "$old" != "$new" ]; then
+            echo "body<<EOF" >> "$GITHUB_OUTPUT"
+            git show --no-patch --format=%b >> "$GITHUB_OUTPUT"
+            echo "EOF" >> "$GITHUB_OUTPUT"
+          fi
```
MattSturgeon, line 138: We essentially copy-paste this workflow step three times. I wonder if there's a better approach? Maybe we can [avoid duplication](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication) by adding a composite action? I think that'd be best in a follow-up PR as I like to keep workflow changes to a minimum per-PR

flake/default.nix
```diff
@@ -1,14 +1,66 @@
...
+  flake = lib.optionalAttrs (partitionStack == [ ]) (
+    builtins.mapAttrs (
+      attrName: partitions:
+      lib.mkMerge (
+        builtins.map (
+          partition: lib.mkForce config.partitions.${partition}.module.flake.${attrName}
+        ) partitions
+      )
+    ) partitionedAttrs
+  );
```
MattSturgeon, line 65: This is the implementation from https://github.com/hercules-ci/flake-parts/pull/272, if that gets merged before this we can bump our flake-parts input and drop this.

.github/workflows/update.yml
```diff
@@ -113,6 +121,38 @@ jobs:
...
+
+      - name: Update docs flake.lock
+        id: docs_flake_lock
+        if: inputs.docs_lock || github.event_name == 'schedule'
+        run: |
+          root_nixpkgs=$(nix eval -f . 'inputs.nixpkgs.rev')
+          old=$(git show --no-patch --format=%h)
+          nix flake update --commit-lock-file \
+              --override-input nixpkgs "github:NixOS/nixpkgs/$root_nixpkgs" \
+              --flake './docs'
```
MattSturgeon, line 148: This is _kinda_ awkward; we can't easily re-use the nixpkgs input from the root module because flake's don't work very well when referencing other flakes with relative paths. https://github.com/NixOS/nix/pull/10089 improves things, but we can't take advantage of it until nix 2.26.

For now we're stuck manually trying to keep nixpkgs in sync across lock files. We can do that fairly easily here using `--override-input`, but we have to remember to do the same if we ever update lock files manually.

That said, we should prefer using the ci workflow to update nixpkgs _anyway_, because the CI will also check if the generated files need to change.

traxys:
> However this depends on https://github.com/NixOS/nix/pull/10089 making it to a stable nix version, ideally making it to nixVersions.minimum which may take some time. I'm unsure how lix deals with relative path flake inputs, too.

For the dev flake I think we can add requirements on the nix version stricter than minimum (though I agree, not only latest)

default.nix
```diff
@@ -1,6 +1,6 @@
 (import (
   let
-    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    lock = builtins.fromJSON (builtins.readFile ./flake/dev/flake.lock);
```
traxys, line 3: Why do we need the dev lockfile for flake-compat?

flake.nix
```diff
@@ -9,60 +9,10 @@
...
-    devshell = {
-      url = "github:numtide/devshell";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-    treefmt-nix = {
-      url = "github:numtide/treefmt-nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-
-    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
```
traxys, line 59: I don't think this is an dev input, it's a third kind, only for users who dont't use flakes

MattSturgeon:
> For the dev flake I think we can add requirements on the nix version stricter than minimum (though I agree, not only latest)

Thinking about it, we only interact with the dev flake in two scenarios: when updating its lockfile and when evaluating its inputs for the partition.

The first is only done by us (usually by our ci), so we can require whatever version of nix we like.

The second is done by the flake.parts `partitions` module, which uses `builtins.getFlake` for string value flakes or https://github.com/ursi/get-flake for path value flakes. Although, this is technically an internal implementation detail; maybe at some point in the future `builtins.getFlake` could support both types and flake.parts will stop using https://github.com/ursi/get-flake.

We define the partition's `extraInputsFlake` as a path value; so, at least currently, we don't really care about the native nix impl while _evaluating_ the partition; instead we only care what https://github.com/ursi/get-flake supports.

default.nix
MattSturgeon: It's less that we need it in the dev lockfile, and more that we **don't** need it in the root lockfile.

My thought process for this PR was every input we previously claimed was optional, should be part of the partitioned off submodule's inputs.

For this specific input, we should probably not use flakes at all and just [`fetchTarball`](https://nix.dev/manual/nix/2.18/language/builtins#builtins-fetchTarball) it explicitly in the relevant file(s), but that felt out of scope for this PR.

flake.nix
MattSturgeon: If it's for users who don't use flakes, then it definitely doesn't belong in the root flake's inputs

default.nix
traxys: I'd prefer not removing it from the main flake then (for now). I made it a flake input in order to have it updated by CI, we could do another mechanism for that

MattSturgeon: The dev flake is also updated by CI, so I don't see the issue.

However isn't the URL versioned, and therefore not updatable, anyway? So the flake lock is only managing that hash for us, not the version.

flake.nix
MattSturgeon: The root flake should only contain inputs that need to be propagated to end-users' lockfiles

traxys (approved):

MattSturgeon:
@mergify queue

MattSturgeon:
> The `partitions` module uses `builtins.getFlake` for string values or [ursi/get-flake](https://github.com/ursi/get-flake?rgh-link-date=2025-02-24T09%3A47%3A39Z) for path values.

This is slightly out of date, flake.parts switched from [ursi/get-flake](https://github.com/ursi/get-flake?rgh-link-date=2025-02-24T09%3A47%3A39Z) to [edolstra/flake-compat](https://github.com/edolstra/flake-compat?rgh-link-date=2025-02-24T09%3A47%3A39Z) in https://github.com/hercules-ci/flake-parts/pull/261, but otherwise the point still stands.

PR #3031, opened 2025-02-24
ci/update: fix `inputs.nixpkgs.rev` eval

MattSturgeon:
Need to use raw output to avoid having a quoted string, otherwise we end up with a malformed flakeref:

```
error: in URL 'github:NixOS/nixpkgs/"2d068ae5c6516b2d04562de50a58c682540de9bf"',
'"2d068ae5c6516b2d04562de50a58c682540de9bf"' is not a commit hash or branch/tag name
```


MattSturgeon:
@mergify queue

PR #3032, opened 2025-02-24
docs/fix-links: generalise checks for links targeting `.`

MattSturgeon:
- Strip trailing `?query` and/or `#anchor`
- Strip leading `./` recursively
- Check if what's left is `""` or `"."`

Any link that targets the current page should be left as-is (no-op).

Follow up to https://github.com/nix-community/nixvim/pull/3031 and https://github.com/nix-community/nixvim/pull/3010

### DIFF
--- a/docs/fix-links/filter.lua
+++ b/docs/fix-links/filter.lua
@@ -16,16 +16,20 @@ end
 
 function Link(link)
 	local target = link.target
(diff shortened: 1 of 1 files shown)


MattSturgeon:
I've tested building the docs locally. All links in the readme and contributing pages still seem to be correct.

PR #3034, opened 2025-02-24
docs/fix-links: pandoc `markdown` -> `gfm`

### LINKED ISSUE #3033: [BUG] Contributing page does not render admonitions in docs
The contributing page does not render its GFM admonitions on the docs website.

This is due to a bug in the `fix-links` builder; pandoc is escaping "unused"\* `[` tokens, so the `[!TIP]` ends up being escaped to `\[!TIP\]` and rendered literally. 

\* unused as-in not forming part of an actual link

MattSturgeon:
We are essentially writing GFM e.g. inline HTML comments and GFM alerts. So tell pandoc to parse/render as such. This resolves issues such as `> [!TIP]` being escaped to `> \[!TIP\]`.

### Before:

image

### After:

image

Fixes #3033

### DIFF
--- a/docs/fix-links/default.nix
+++ b/docs/fix-links/default.nix
@@ -26,8 +26,8 @@ runCommand (src.name or (builtins.baseNameOf src))
 
     pandoc \
         --output $out \
-        --from markdown \
-        --to markdown-raw_attribute \
+        --from gfm \
+        --to gfm \
         --lua-filter filter.lua \
         $src
   ''


docs/fix-links/default.nix
```diff
@@ -26,8 +26,8 @@ runCommand (src.name or (builtins.baseNameOf src))
 
     pandoc \
         --output $out \
-        --from markdown \
-        --to markdown-raw_attribute \
+        --from gfm \
+        --to gfm \
```
MattSturgeon, line 30: Maybe it'd be best to keep `--from gfm` but have `--to markdown-raw_attribute+alerts`?

This probably better fits what we're actually trying to target with mdbook's markdown support :thinking:

traxys (approved):
Let's use gfm until it comes to bite us x)

Issue #3035, opened 2025-02-24
[BUG] Update CI cannot re-apply "empty" cherry-pick


MattSturgeon:
https://github.com/nix-community/nixvim/actions/runs/13501971853/job/37722390337#step:11:24

In this specific scenario, the cherry-pick was empty because we found the wrong "base" commit; the base commit should be the most-recent update/generate commit, which we detect based on the author being `github-actions[bot]`. In this case, the latest generated commits were authored by @GaetanLepage, so we incorrectly attempt to cherry-pick them.

There are two issues here:

- our "base" commit detection fails when generated commits are not authored by `github-actions[bot]`
- cherry-picking failes ungracefully when we attempt to re-apply an empty/no-op commit
  Fixed in #3036

PR #3036, opened 2025-02-24
ci/update: drop empty commits on re-apply

MattSturgeon:
Partial fix for #3035

Tell `git cherry-pick` to drop any commits that end up being empty, instead of aborting.


MattSturgeon:
@mergify queue

PR #3040, opened 2025-02-27
plugins/alpha: migrate to mkNeovimPlugin

HeitorAugustoLN:


plugins/by-name/alpha/default.nix
```diff
@@ -193,19 +230,8 @@ in
...
-                inherit layout opts;
-              });
-        in
-        ''
-          require('alpha').setup(${lib.nixvim.toLuaObject setupOptions})
-          require('alpha.term')
-        '';
+      plugins.alpha.luaConfig.content =
+        optionalString themeDefined "require('alpha').setup(${toLuaObject cfg.theme})\n"
+        + "require('alpha.term')";
```
MattSturgeon, line 235: I'm a little confused. You have a `settings` option, but you are also manually calling setup and passing it a `theme` option.

The idea of `settings` is that it is the nix representation of the lua passed to `setup`.

HeitorAugustoLN: `.setup` needs to be manually called when using `theme` option. Because that is the only way we can set pre-defined themes. That's why, `theme` and `settings.layout` options can't be used at the same time

MattSturgeon: If so, that also implies:

1. `settings` as a whole must not be used when a theme is defined
2. `setup` must _always_ be called manually, either passing it `settings` or `theme`

MattSturgeon: Interesting, the upstream examples show a different usage:

```lua
require('alpha').setup(require('alpha.themes.dashboard').config)
```

They are passing pre-defined settings to setup; getting those settings _from_ a theme module.

HeitorAugustoLN: It is the same usage, the theme option is being converted here:

https://github.com/nix-community/nixvim/blob/c83415c8c0125d22a7f79c3767d28775c696d5da/plugins/by-name/alpha/default.nix#L196-L201

plugins/by-name/alpha/default.nix
```diff
@@ -137,75 +123,113 @@ in
...
         {
-          assertion = themeDefined || layoutDefined;
+          assertion = layoutDefined || themeDefined;
           message = ''
             You have to either set a `theme` or define some sections in `layout`.
           '';
         }
         {
-          assertion = !(themeDefined && layoutDefined);
+          assertion = !(themeDefined && cfg.settings != { });
```
MattSturgeon, line 222: If we have any settingsOptions declared, then `settings` will never be `{}`.

That's because a submodule's options _always_ end up in the submodule's `config` (aka the final value of the submodule-option).

Instead, we could take one of the following approaches:

- check `toLuaObject settings == "{}"` (toLuaObject will filter null/empty values)
- check `opt.settings.highestPrio == 1500 && length opt.settings.definitions == 1` i.e. only the `default` definition is present

plugins/by-name/alpha/default.nix
```diff
@@ -137,75 +123,113 @@ in
...
-              (with cfg; {
-                inherit layout opts;
-              });
-        in
-        ''
-          require('alpha').setup(${lib.nixvim.toLuaObject setupOptions})
-          require('alpha.term')
-        '';
+      plugins.alpha.luaConfig.content = ''
+        require('alpha').setup(${toLuaObject (if themeDefined then cfg.theme else cfg.settings)});
```
MattSturgeon, line 231: If we're calling `setup` manually, we need to tell mkNeovimPlugin not to also call `setup`.

I.e. add `callSetup = false` somewhere, perhaps immediately before `extraConfig`?

plugins/by-name/alpha/default.nix
HeitorAugustoLN: Forgot that 😅

plugins/by-name/alpha/default.nix
```diff
@@ -1,86 +1,74 @@
...
+              example = "Some text";
+              description = ''
+                The value for the section.
+              '';
+            };
 
-      layout = mkOption {
-        type = with types; either (maybeRaw str) (listOf sectionType);
-        default = [ ];
-        description = "List of sections to layout for the dashboard";
```
MattSturgeon, line 83: > error: option `plugins.alpha.settings.layout` has no description

HeitorAugustoLN: Added description again.

plugins/by-name/alpha/default.nix
```diff
@@ -1,86 +1,74 @@
...
-        visible = false;
-      };
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "alpha";
+  packPathName = "alpha-nvim";
+  package = "alpha-nvim";
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  callSetup = false;
```
MattSturgeon: Typically we have an explanatory comment.

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,118 @@ in
...
+        The theme to use for the dashboard.
+      '';
+      example = "dashboard";
+      apply =
+        value: if builtins.isString value then mkRaw ''require("alpha.themes.${value}").config'' else value;
+    };
+  };
 
-      extraPlugins = [ cfg.package ];
+  optionsRenamedToSettings = [
```
MattSturgeon: This should have a comment saying when it was added.

Also, we need to update the test cases. See CI failures: https://buildbot.nix-community.org/#/builders/2885/builds/3656/steps/1/logs/stdio

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,118 @@ in
...
-              (with cfg; {
-                inherit layout opts;
-              });
-        in
-        ''
-          require('alpha').setup(${lib.nixvim.toLuaObject setupOptions})
-          require('alpha.term')
-        '';
+      plugins.alpha.luaConfig.content = ''
+        require('alpha').setup(${toLuaObject (if themeDefined then cfg.theme else cfg.settings)});
```
MattSturgeon: This feels like it could be confusing to users. I wonder if we should have it so that we do this in the module system?

```nix
settings = mkDerivedConfig opts.theme (
  value:
  if builtins.isString value then mkRaw ''require("alpha.themes.${value}").config'' else value
);
```

However, that'd require refactoring `settings` to a `maybeRaw` type.

As a minimum, the weird behaviour of discarding all configured settings if a theme is defined should be documented somewhere.

MattSturgeon:
Build failure was a "cache failure", so should fix itself on next push :crossed_fingers:

GaetanLepage:
Can you rebase @HeitorAugustoLN please?

HeitorAugustoLN:
> Can you rebase @HeitorAugustoLN please?

Rebased it

MattSturgeon:
Some of the tests are failing on assertions:

```
custom-layout> Failed 1 expectation:
custom-layout> - Expected length to be 0 but found 2.
custom-layout> For warnings:
custom-layout> - The option `plugins.alpha.layout' defined in `/nix/store/5fvdjp0a5r31cbkli9y1xi2fpk34kz1v-source/tests/test-sources/plugins/by-name/alpha/default.nix' has been renamed to `plugins.alpha.settings.layout'.
custom-layout> - The option `plugins.alpha.opts' defined in `/nix/store/5fvdjp0a5r31cbkli9y1xi2fpk34kz1v-source/tests/test-sources/plugins/by-name/alpha/default.nix' has been renamed to `plugins.alpha.settings.opts'.
```
```
no-icons> Failed 1 expectation:
no-icons> - Expected length to be 0 but found 1.
no-icons> For assertions:
no-icons> - Nixvim (plugins.alpha): You can't define both a `theme` and custom options.
no-icons> Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
```
```
terminal> Failed 1 expectation:
terminal> - Expected length to be 0 but found 1.
terminal> For warnings:
terminal> - The option `plugins.alpha.layout' defined in `/nix/store/5fvdjp0a5r31cbkli9y1xi2fpk34kz1v-source/tests/test-sources/plugins/by-name/alpha/default.nix' has been renamed to `plugins.alpha.settings.layout'.
```
```
theme-lua> Failed 1 expectation:
theme-lua> - Expected length to be 0 but found 1.
theme-lua> For assertions:
theme-lua> - Nixvim (plugins.alpha): You can't define both a `theme` and custom options.
theme-lua> Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
```

https://buildbot.nix-community.org/#/builders/4434/builds/509/steps/1/logs/stdio

Either need to satisfy the assertions by using the options as intended, fix the assertions if they're false-positives, or "expect" the assertion if the test is intending to trigger it.

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,118 @@ in
...
+  optionsRenamedToSettings = [
+    "opts"
+    "layout"
+  ];
 
-      assertions = lib.nixvim.mkAssertions "plugins.alpha" [
+  extraConfig =
+    cfg:
+    let
+      layoutDefined = cfg.settings.layout != null;
```
MattSturgeon: Here it's fine, because `layout` is a declared sub-option, but there's a small part of me that hurts whenever I see a settings attr accessed without `?` or `or`.

Doing:
```suggestion
      layoutDefined = cfg.settings.layout or null != null;
```

...now would mean this assertion doesn't break if we ever drop the explicit `layout` option in favour of relying on the freeformType.

Although this isn't really needed.

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,118 @@ in
...
         {
-          assertion = themeDefined || layoutDefined;
+          assertion = layoutDefined || themeDefined;
           message = ''
             You have to either set a `theme` or define some sections in `layout`.
           '';
         }
         {
-          assertion = !(themeDefined && layoutDefined);
+          assertion = !(themeDefined && (cfg.settings.layout != null || cfg.settings.opts != null));
```
MattSturgeon: `cfg.settings.layout != null` is already bound as `layoutDefined`; we can re-use that here.

Isn't the point of this assertion to say "if you've defined a theme, then `settings` is ignored? If so, should the assertion be checking something like `opts.settings.highestPrio == 1500 && length opts.settings.definitions == 1`?

plugins/by-name/alpha/default.nix
MattSturgeon: I still feel like this behaviour of `settings` only being used when there is no theme is a bad design 🤔

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,122 @@ in
...
 
-      extraPlugins = [ cfg.package ];
+  optionsRenamedToSettings = [
+    "opts"
+    "layout"
+  ];
 
-      assertions = lib.nixvim.mkAssertions "plugins.alpha" [
+  extraConfig =
+    opts: cfg:
```
MattSturgeon:
```suggestion
    cfg: opts:
```

plugins/by-name/alpha/default.nix
```diff
@@ -135,77 +123,122 @@ in
...
             You have to either set a `theme` or define some sections in `layout`.
           '';
         }
         {
-          assertion = !(themeDefined && layoutDefined);
+          assertion =
+            !(
+              themeDefined
+              && (opts.settings.highestPrio == 1500 && builtins.length opts.settings.definitions == 1)
+            );
```
MattSturgeon: I'm probably mis-understanding this logic. But should the assertion be `themeDefined` _implies_ `settings` not defined?

```suggestion
            themeDefined ->
              (opts.settings.highestPrio == 1500 && builtins.length opts.settings.definitions == 1);
```

GaetanLepage:
This should now be ready!

plugins/by-name/alpha/default.nix
GaetanLepage: This is now solved as `settings` can accept a `rawLua` type.
Hence, the user can do:
```nix
plugins.alpha = {
  enable = true;
  settings.__raw = "require('alpha.themes.dashboard').config";
};
```

plugins/by-name/alpha/default.nix
GaetanLepage: I moved `callSetup = false;` right above `extraConfig` to make its purpose obvious.

plugins/by-name/alpha/default.nix
GaetanLepage: Done.

plugins/by-name/alpha/deprecations.nix
```diff
@@ -0,0 +1,17 @@
...
+  imports = [
+    (lib.mkRemovedOptionModule [ "plugins" "alpha" "theme" ] ''
+      Directly use `settings` to set the theme.
+      Example:
+
+      plugins.alpha = {
+        enable = true;
+        settings.__raw = "require('alpha.themes.dashboard').config"
+      }
+    '')
```
MattSturgeon: This doesn't need to be a removal, it can be a `mkChangedOptionModule`.

Although, IIRC `mkChangedOptionModule` doesn't provide a way to define custom migration instructions without manually adding a separate warning.

MattSturgeon: Frankly, I'm not convinced this option needs removing. It seems like a useful abstraction with zero maintenance cost:

```nix
plugins.alpha.settings = mkDerivedConfig opts.theme (
  value:
  lib.mkIf (value != null) (
    if builtins.isString value then
      mkRaw "require('alpha.themes.' .. ${toLuaObject value}).config"
    else
      value
  )
);
```

Maybe we could deprecate `theme` allowing non-string values,[^1] since `settings` now supports those, but otherwise the `theme` option seems fine to me.

[^1]: This can be done using a warning predicated on `cfg.theme != null -> isString cfg.theme` and shadowing the option's `type.description` to avoid documenting its 'old' type.

GaetanLepage: Makes sense. I brought back the `theme` option. I changed its type to `nullOr str`.
You will probably tell me that this is not so great for users who were providing a raw lua string...

plugins/by-name/alpha/default.nix
```diff
@@ -1,177 +1,93 @@
...
-            Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
-          '';
-        }
-      ];
+  callSetup = false;
+  extraConfig = cfg: {
+    plugins.alpha = {
+      settings = lib.nixvim.mkIfNonNull' cfg.theme (
+        lib.nixvim.mkRaw "require('alpha.themes.${cfg.theme}').config"
+      );
```
MattSturgeon: We should use `mkDerivedConfig` to preserve override priority.

This allows a user to have `theme = mkDefault` and `settings = mkForce` (or vice-versa) and it behave as expected.
```suggestion
      settings = lib.mkDerivedConfig opts.theme (
        value:
        lib.nixvim.mkIfNonNull' value (
          lib.nixvim.mkRaw "require('alpha.themes.${value}').config"
        )
      );
```

Aside: I'm not convinced our `mkIfNonNull'` helper actually _helps_ vs just writing `mkIf (foo != null)`. IMO it's more readable to do the latter.

plugins/by-name/alpha/default.nix
```diff
@@ -1,177 +1,93 @@
...
+      }
+    ];
+  };
 
-      opts = lib.nixvim.mkNullOrOption (with types; attrsOf anything) ''
-        Optional global options.
-      '';
+  extraOptions = {
+    theme = lib.mkOption {
+      type = with lib.types; nullOr str;
```
MattSturgeon: For backwards compatibility, we should do:


```suggestion
      type =
        with types;
        let
          # TODO: deprecated 2025-10-30, remove after 26.05
          old = nullOr (maybeRaw str);
          new = nullOr str;
        in
        old // { inherit (new) description; };
```

and add a warning or assertion:

```nix
{
  when = cfg.theme ? __raw;
  warning = ''
    Defining `${opts.theme}` as raw lua is deprecated. You can define `${opts.settings}` as raw lua instead:
    ${opts.settings} = lib.nixvim.mkRaw ${lib.generators.toPretty { } cfg.theme.__raw};
  '';
}
```

plugins/by-name/alpha/default.nix
MattSturgeon: If allowing the "old" type, then `__raw` should be handled here too:
```nix
lib.mkDerivedConfig opts.theme (
  value:
  lib.mkMerge (
    lib.optional (value ? __raw) value
    ++ lib.optional (lib.isString value) {
      __raw = "require('alpha.themes.${value}').config";
    }
  )
)

File: plugins/by-name/alpha/default.nix (hunk 14 above)
Comment (GaetanLepage) [reply plugins/by-name/alpha/default.nix]: Neat! Done.

**Comment by GaetanLepage** [comment]
> Just blocking on `mkDerivedConfig` now: [#3040 (comment)](https://github.com/nix-community/nixvim/pull/3040#discussion_r2478200922)

Oops, forgot it. I added it.

File: tests/test-sources/plugins/by-name/alpha/default.nix [hunk 15]
```diff
@@ -1,22 +1,25 @@
+{ lib, ... }:
 {
   theme = {
     plugins.alpha = {
       enable = true;
+
       theme = "dashboard";
```
MattSturgeon, line 7: `config.layout: expected table, got nil`

I assume this relates to this test case.

```
building '/nix/store/hbxwcm0vhba0f7iw7vi039i6k0ybl810-theme.drv'...
theme> ERROR: Error detected while processing /nix/store/qwhhfpdh433siw540yn2q3nsy5g5448x-init.lua:
theme> E5113: Error while calling lua chunk: ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:732: config.layout: expected table, got nil
theme> stack traceback:
theme>  [C]: in function 'error'
theme>  vim/shared.lua: in function 'validate'
theme>  ...dir/pack/myNeovimPackages/start/alpha-nvim/lua/alpha.lua:732: in function 'setup'
theme>  /nix/store/qwhhfpdh433siw540yn2q3nsy5g5448x-init.lua:9: in main chunk
```

plugins/by-name/alpha/default.nix
```diff
@@ -1,177 +1,114 @@
...
-            You can't define both a `theme` and custom options.
-            Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
-          '';
-        }
-      ];
+    plugins.alpha = {
+      settings = lib.mkDerivedConfig opts.theme (
+        value:
+        lib.mkIf (value != null) (
+          if builtins.isString value then
```
MattSturgeon, line 100: Since you are _asserting_ `!(theme ? __raw)` above, we shouldn't need to handle `if isString else`. We only need this if you change the assertion to a warning.

plugins/by-name/alpha/default.nix
```diff
@@ -1,177 +1,114 @@
...
-          message = ''
-            You can't define both a `theme` and custom options.
-            Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
-          '';
-        }
-      ];
+    plugins.alpha = {
+      settings = lib.mkDerivedConfig opts.theme (
+        value:
+        lib.mkIf (value != null) (
```
MattSturgeon: Re: https://github.com/nix-community/nixvim/pull/3040#discussion_r2478429003

I'm unsure how this will work, but _maybe_ we need to lift this `mkIf` wrapper to outside of the `mkDerivedConfig`?

Sometimes nesting `mkOverride`, `mkMerge` and `mkIf` like Russian dolls can produce weird behavior, (IIRC).

Internally, `mkDerivedConfig` does a `mkIf opt.isDefined` and a `mkOverride opt.highestPrio`.

GaetanLepage: You are right! Moving it outside fixed it.

plugins/by-name/alpha/default.nix
```diff
@@ -1,177 +1,113 @@
...
-          message = ''
-            You can't define both a `theme` and custom options.
-            Set `plugins.alpha.theme = null` if you want to configure alpha manually using the `layout` option.
-          '';
-        }
-      ];
+  callSetup = false;
+  extraConfig = cfg: opts: {
+    assertions = lib.nixvim.mkAssertions "plugins.alpha" {
+      assertion = !(cfg.theme ? __raw);
```
MattSturgeon: Probably better to do:


```suggestion
      assertion = cfg.theme != null -> builtins.isString cfg.theme);
```

Because maybeRaw also allows nixpkgs' lua-inline type.

MattSturgeon (approved):

PR #3045, opened 2025-03-04
tests: cleanup

GaetanLepage:
- **tests: re-enable tests that work**
- **tests/none-ls: remove beautysh test (removed upstream)**


MattSturgeon (approved):
Awesome, thanks!

CI still passes and the diff looks good.

NOTE: if there's any conflicts with #2991, it may get partially reverted when the CI does its cherry-pick, so we may have to manually un-revert.

PR #3048, opened 2025-03-04
plugins/lsp: add gdscript package

MattSturgeon:
See https://github.com/nix-community/nixvim/issues/1359#issuecomment-2692475317

### DIFF
--- a/plugins/lsp/lsp-packages.nix
+++ b/plugins/lsp/lsp-packages.nix
@@ -51,7 +51,6 @@
     "flux_lsp"
     "foam_ls"
     "fsharp_language_server"
-    "gdscript"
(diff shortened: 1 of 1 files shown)


plugins/lsp/lsp-packages.nix
```diff
@@ -237,6 +236,7 @@
     fsautocomplete = "fsautocomplete";
     fstar = "fstar";
     futhark_lsp = "futhark";
+    gdscript = "gdtoolkit_4";
```
MattSturgeon, line 239: Actually, I'm not sure this is correct. GD Toolkit doesn't mention a language server.

Perhaps the correct package is `godot_4`?

PR #3049, opened 2025-03-05
docs/lib: generate documentation for nixvim's lib

MattSturgeon:
- **docs/lib: init**
- **cleanup & refactoring**

wip, based on nixpkg's lib docs

### DIFF
--- a/docs/lib/default.nix
+++ b/docs/lib/default.nix
@@ -1,0 +1,83 @@
+# Generates the documentation for library functions using nixdoc.
+{
+  lib,
+  runCommand,
+  writers,
+  nixdoc,
+  nixvim,
+  libsets ? [
+    # TODO: all lib sections
+    {
+      name = "utils";
+      description = "utility functions";
+    }
+    {
+      name = "options";
+      description = "NixOS / nixpkgs option handling";
+    }
+  ],
+}:
+
+runCommand "nixvim-lib-docs"
+  {
+    nativeBuildInputs = [
+      nixdoc
+    ];
+
+    locations = writers.writeJSON "locations.json" (
+      import ./function-locations.nix {
+        inherit lib libsets;
+        root = nixvim;
+        libToDoc = nixvim.lib.nixvim;
+        revision = nixvim.rev or "main";
+        prefix = "lib.nixvim";
+      }
+    );
+
+    passthru.menu = ''
+      # Lib
+      - [lib.nixvim](lib/index.md)
+      ${lib.concatMapStringsSep "\n" ({ name, ... }: "  - [${name}](lib/${name})") libsets}
+    '';
+
+  }
+  ''
+    export RUST_BACKTRACE=1
+
+    function docgen {
+      name=$1
+      baseName=$2
+      description=$3
+
+      file="${nixvim}/lib/$baseName"
+      if [[ -e "$file.nix" ]]; then
+        file+=".nix"
+      elif [[ -e "$file/default.nix" ]]; then
+        file+="/default.nix"
+      else
+        echo "Error: no file for $file"
+        exit 1
+      fi
+
+      nixdoc \
+        --prefix "lib.nixvim" \
+        --category "$name" \
+        --description "lib.nixvim.$name: $description" \
+        --locs $locations \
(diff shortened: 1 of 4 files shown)


docs/mdbook/SUMMARY.md
```diff
@@ -5,11 +5,12 @@
 # User guide
 
 - [Installation](./user-guide/install.md)
-- [Helpers](./user-guide/helpers.md)
 - [FAQ](./user-guide/faq.md)
 - [Configuration examples](./user-guide/config-examples.md)
 - [Lazy Loading](./user-guide/lazy-loading.md)
 
+@FUNCTIONS_MENU@
+
```
MattSturgeon, line 13: I originally put this higher up because the old `helpers.md` guide was near the top. However, it'd also be valid to have "Functions Reference" section _after_ the "Options" reference section, if preferred.

Thoughts?

docs/lib/default.nix
```diff
@@ -0,0 +1,103 @@
...
+        functionSet = lib.extend nixvim.lib.overlay;
+        pathsToScan = builtins.catAttrs "loc" pageSpecs;
+        revision = nixvim.rev or "main";
+      }
+    );
+
+    passthru.menu = ''
+      # Library functions
+
+      - [`lib`](${locToFile [ "index" ]})
```
MattSturgeon, line 42: Originally, I was allowing for having `lib.nixvim.*` and also a `lib.types` page, so I made the root node `lib` instead of `lib.nixvim`.

image

In hindsight, maybe that's confusing? Especially as the "index" page is _really_ about using `lib.nixvim` not `lib`...

MattSturgeon:
LMK if there's anything I can do to make this easier to review, or any parts that could be reviewed separately in another PR.

MattSturgeon:
@mergifyio rebase

khaneliman:
Everything looks pretty good to me stepping through commits. One thing that I was curious about was the separation of internal v public lib api. Is it just supposed to be to help document the most common lib functions that help with creating a config?

MattSturgeon:
> Everything looks pretty good to me stepping through commits.

🎉

> One thing that I was curious about was the separation of internal v public lib api. Is it just supposed to be to help document the most common lib functions that help with creating a config? 

Before this PR `utils.nix` contained a mix of functions intended for internal use and functions intended for use by end-users.

I was hesitant to add "internal" functions to the docs. I was also hesitant to remove rfc145 doc-comments from internal functions. My _workaround_ was to split the file into public & internal files. This is kinda ugly and ham-fisted, but it works...

I'm not _toatally_ opposed to having docs for internal functions too, but I think that should be addressed separately and clearly separated/hidden away in the docs. 

Anything we publicly document we're kinda saying "this is our public API, this is kinda stable and won't change without warning" so we should be careful which functions are documented publicly, just as we are careful which _options_ are visible/hidden/internal.

khaneliman:
>Anything we publicly document we're kinda saying "this is our public API, this is kinda stable and won't change without warning" so we should be careful which functions are documented publicly, just as we are careful which options are visible/hidden/internal.

This makes more sense to me about why we're being so conservative in the documentation. Thank you, I agree we should start slimmed down with the `public` api and we can expose more things later if we want.

khaneliman (approved):
LGTM, thanks Matt!

docs/lib/default.nix
```diff
@@ -0,0 +1,178 @@
...
+      markdown =
+        if builtins.isString markdown then
+          builtins.toFile "${lib.strings.replaceStrings [ "/" "-" ] (lib.lists.last loc)}.md" markdown
+        else
+          markdown;
+      outFile = lib.strings.concatStringsSep "/" (loc ++ [ "index.md" ]);
+    }
+    // lib.optionalAttrs (page ? pages) {
+      pages = elaboratePages loc pages;
+    };
```
MattSturgeon, line 46: I believe we could simplify this by using the module system to evaluate `pages.nix`.

Most of this "normalisation" could be done by a `page` submodule.

I'll plan to address this in a follow up PR.

MattSturgeon:
@mergifyio queue

PR #3058, opened 2025-03-09
plugins/image: add resolveImagePath option

xamcost:
Hi there! First of all, thanks for providing and maintaining Nixvim! This tiny contribution updates the _image_ plugin to expose the [resolve_image_path](https://github.com/3rd/image.nvim?tab=readme-ov-file#integrations) options, enabling users to modify how images path are resolved within their projects (typical use case: when using obsidian.nvim with a custom directory for images).

### DIFF
--- a/plugins/by-name/image/default.nix
+++ b/plugins/by-name/image/default.nix
@@ -66,6 +66,20 @@ in
           filetypes = helpers.defaultNullOpts.mkListOf types.str filetypesDefault ''
             Markdown extensions (ie. quarto) can go here.
           '';
+
+          resolveImagePath = helpers.mkNullOrLuaFn' {
+            description = "Configures how to resolve image paths.";
+            example = lib.literalExpression ''
+              lib.nixvim.mkRaw '''
+                function(document_path, image_path, fallback)
+                  -- document_path is the path to the file that contains the image
+                  -- image_path is the potentially relative path to the image. For markdown, it's `![](this text)`
+                  -- fallback is the default behavior
+                  return fallback(document_path, image_path)
+                end
+              ''';
+            '';
+          };
         };
       in
       mapAttrs mkIntegrationOptions {
@@ -150,6 +164,7 @@ in
                   download_remote_images = v.downloadRemoteImages;
                   only_render_image_at_cursor = v.onlyRenderImageAtCursor;
                   inherit (v) filetypes;
+                  resolve_image_path = v.resolveImagePath;
                 };
               in
               mapAttrs (_: processIntegrationOptions) integrations;


MattSturgeon:
This wouldn't be needed if the plugin was updated to use the freeform `settings` pattern. I.e. by porting the plugin to use `mkNeovimPlugin`.

Other than being one more option we'll have to deprecate and rename when we do eventually update this plugin, I don't see any issues with adding the option as a short-term solution though.

khaneliman:
Yeah, I agree. Since the name transition will be handled fine with `optionsRenamedToSettings` it should be fine to add without really taking on too much tech debt.

plugins/by-name/image/default.nix
```diff
@@ -66,6 +66,20 @@ in
...
+
+            						Example:
+            						```lua
+            							function(document_path, image_path, fallback)
+            								-- document_path is the path to the file that contains the image
+            								-- image_path is the potentially relative path to the image. For markdown, it's `![](this text)`
+            								-- fallback is the default behavior
+            								return fallback(document_path, image_path)
+            							end
+            						```
```
MattSturgeon, line 81: Can you remove the tabs please?
```suggestion
            Configures how to resolve image paths.

            Example:
            ```lua
              function(document_path, image_path, fallback)
                -- document_path is the path to the file that contains the image
                -- image_path is the potentially relative path to the image. For markdown, it's `![](this text)`
                -- fallback is the default behavior
                return fallback(document_path, image_path)
              end
            ```
```

MattSturgeon: Also, I'd usually prefer examples to be "proper" nixos-option examples:

```nix
resolveImagePath = helpers.mkNullOrLuaFn' {
  description = "Configures how to resolve image paths.";
    lib.nixvim.mkRaw '''
      function(document_path, image_path, fallback)
        -- document_path is the path to the file that contains the image
        -- image_path is the potentially relative path to the image. For markdown, it's `![](this text)`
        -- fallback is the default behavior
        return fallback(document_path, image_path)
      end
    '''
  '';
}
```

xamcost: Small note: the disgracious extra tabs are added by nix formatting. In 80cb5febac55 , I bypassed the pre-commit hooks to avoid their addition, but CI checks then fails. In 8dd136ba57ee , I just ran nix formatting, and now CI checks are all ok, but these annoying tabs are back...

MattSturgeon: It sounds like you have your editor setup to use tab-based indentation (80cb5febac55b4e09288b79a66950d2dfd7cd427 is using tabs) and then you're running into something similar to https://github.com/NixOS/nixfmt/issues/290 when running nixfmt.

Nixfmt isn't adding tabs, it's adding spaces but not _removing_ the tabs.

MattSturgeon (approved):

MattSturgeon:
@mergify queue

plugins/by-name/image/default.nix
xamcost: oh pff sorry about that! Indeed, I overlooked this setting upon migrating to Nixvim. Sorry for the noise, and thanks for your review!

Issue #3063, opened 2025-03-11
[BUG] failed to load zls activated in nix-shell


pseudocc:
| Field        | Description |
|--------------|-------------|
| Plugin       | `lsp`  | 
| Nixpkgs      | `24.11` | 
| Home Manager | `24.11` | 
| Project Flake| `master` |

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

I use zig and zls (0.13.0) on my system, but for some reason I have projects developed in zig (0.14.0) which is managed in flake.nix, but zls (0.13.0) is always loaded instead of zls (0.14.0).

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  enable = true;
  plugins.lsp = {
    enable = true;
    servers.zls.enable = true;
  };
}
```

project flake.nix
```nix
{
  description = "MRE";

  # zig 0.14.0
  inputs.nixpkgs.url = "github:nixos/nixpkgs";

  outputs = { self, nixpkgs, ... } @ inputs: let
    inherit (nixpkgs) lib;
    eachSystem = fn:
      lib.foldl' (
        acc: system:
          lib.recursiveUpdate
          acc
          (lib.mapAttrs (_: value: {${system} = value;}) (fn system))
      ) {}
      lib.platforms.unix;
  in
    eachSystem (
      system: let
        pkgs = import nixpkgs {inherit system;};
        zig = pkgs.zig_0_14;
      in {
        devShells.default = pkgs.mkShell {
          buildInputs = [ zig pkgs.zls ];
        };
      }
    );
}
```

project flake.lock
```json
{
  "nodes": {
    "nixpkgs": {
      "locked": {
        "lastModified": 1741673466,
        "narHash": "sha256-VVXaC6uiwDMabeJE0iz3PmPj6UPrUa2naPT3Yi6ucpA=",
        "owner": "nixos",
        "repo": "nixpkgs",
        "rev": "8bf6db44e9869a9bc97fb1a7ffb861a15882eae4",
        "type": "github"
      },
      "original": {
        "owner": "nixos",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "root": {
      "inputs": {
        "nixpkgs": "nixpkgs"
      }
    }
  },
  "root": "root",
  "version": 7
}
```

## Additional Informations

zls-0.13.0 comes first, echo $PATH in neovim:
```
[...]

PR #3064, opened 2025-03-11
fix(lsp): load the correct lsp in nix-shell

pseudocc:
Closes: #3063

Don't know if this will introduce regressions but it fixes my case like a charm.


pseudocc:
"I reviewed a few other PRs, and it appears that the failing check in CI is a common issue."

MattSturgeon (changes requested):
See https://github.com/nix-community/nixvim/issues/3063#issuecomment-2720859935

MattSturgeon:
IMO this can be superseded by https://github.com/nix-community/nixvim/pull/3450

We can move discussion there, unless you disagree and feel you have an alternative solution.

PR #3065, opened 2025-03-11
ci: add dependabot and track github-actions dependencies

GaetanLepage:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+
+version: 2
+updates:
+  # Enable version updates for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the default location of `.github/workflows`
+    # You don't need to specify `/.github/workflows` for `directory`. You can use `directory: "/"`.
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
khaneliman, line 14:
```suggestion
    schedule:
      interval: "weekly"
    commit-message:
      prefix: ".github"
```

If we want to make the commits to follow the filename prefix? or if you're going to do `ci` you could do that

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    commit-message:
+      prefix: ".github"
```
MattSturgeon, line 16: Re-opening https://github.com/nix-community/nixvim/pull/3065#discussion_r1989552962:

I'd prefer this without the `.` personally. Ideally, we'd set it to whatever the most-used commit prefix for our workflows was historically (`ci`?), but either way I don't think we often included the `.` in commit messages?

khaneliman (approved):
LGTM

PR #3091, opened 2025-03-20
plugins/image: migrate to mkNeovimPlugin

GaetanLepage:


plugins/by-name/image/default.nix
```diff
@@ -1,37 +1,53 @@
...
-        "image-nvim"
-      ];
-    };
-
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "image";
+  packPathName = "image.nvim";
+  package = "image-nvim";
+
+  maintainers = with lib.maintainers; [ GaetanLepage ];
```
MattSturgeon, line 15: Nit: I'd rather either avoid using `with` until we have multiple list items or formatting it such that adding additional items would minise the diff:

```suggestion
  maintainers = [ lib.maintainers.GaetanLepage ];
```

or

```suggestion
  maintainers = with lib.maintainers; [
    GaetanLepage
  ];
```

plugins/by-name/image/default.nix
```diff
@@ -45,143 +47,97 @@ in
...
-            tmux_show_only_in_active_window = tmuxShowOnlyInActiveWindow;
-            hijack_file_patterns = hijackFilePatterns;
-          }
-          // cfg.extraOptions;
-      in
-      ''
-        require('image').setup(${lib.nixvim.toLuaObject setupOptions})
-      '';
+
+    ] ++ lib.optional (cfg.settings.backend == "ueberzug") cfg.ueberzugPackage;
```
MattSturgeon, line 139: This behaviour should be noted in the `backend` option's description. Probably the `ueberzugPackage` option's extraDescription too?

plugins/by-name/image/default.nix
```diff
@@ -45,143 +47,97 @@ in
...
+
+    extraLuaPackages = ps: [ ps.magick ];
```
MattSturgeon, line 141: This feels like it should be conditional on some option?

plugins/by-name/image/default.nix
```diff
@@ -45,143 +47,97 @@ in
...
+      "cmp_docs"
+      ""
+    ];
+  };
 
+  extraConfig = cfg: {
     extraPackages = [
       # In theory, we could remove that if the user explicitly disables `downloadRemoteImages` for
       # all integrations but shipping `curl` is not too heavy.
       cfg.curlPackage
```
MattSturgeon, line 137: Do we want to address this old TODO comment?

MattSturgeon (commented):
Looks good overall. Just some nits and discussion points below

plugins/by-name/image/default.nix
khaneliman: Looks like it listed as required in the docs for NixOS. Wondering if we should upstream that.

MattSturgeon: That's reasonable. I could hypothetically imagine someone wanting to use a different `magick` package, e.g. with some overrides or a different version. Or maybe someone wanting it to be found impurely from the host system.

But this is somewhat contrived so I'm fine with not supporting that usage.

plugins/by-name/image/default.nix
MattSturgeon: I don't fully understand, but I'll trust your decision.

plugins/by-name/image/default.nix
```diff
@@ -42,146 +49,104 @@ in
           - kitty - best in class, works great and is very snappy
           - ueberzug - backed by ueberzugpp, supports any terminal, but has lower performance
             - Supports multiple images thanks to @jstkdng.
+
+          Note: when choosing the `"ueberzug"` backend, nixvim will automatically add `ueberzugpp`
+          as a dependency.
+          Set `ueberzugPackage = null` to disable this behavior.
```
MattSturgeon, line 55: Do we want an actual markdown admonition here?

```md
> [!Note]
> When choosing the `"ueberzug"` backend, nixvim will automatically add `ueberzugpp` as a dependency.
> Set `ueberzugPackage = null` to disable this behavior.
```

plugins/by-name/image/default.nix
MattSturgeon: Maybe if it's something we don't want to do, we should update/remove the comment?

plugins/by-name/image/default.nix
khaneliman: This looks like the magick lua package though, right? I think it's different than an overrideable imagemagick binary.

khaneliman: Looking upstream, we already pass this in the lua derivation.

khaneliman: Yeah, if it breaks it's an issue with how we fetch/wrap dependencies.

MattSturgeon: I wonder if this is another case that our combinedPlugins impl can't currently handle :grin:

khaneliman: > I wonder if this is another case that our combinedPlugins impl can't currently handle 😁

*PTSD triggers*

PR #3101, opened 2025-03-24
modules: add env option

GaetanLepage:
Add a top-level `env` option that allows to propagate environment variables in the wrapped neovim package.

```nix
env = {
  FOO = "1";
  BAR_PATH = "/home/me/.local/share/bar";
  BAZ_MAX_COUNT = "1000";
}
```

Fixes #3055

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -28,6 +28,26 @@ let
 in
 {
   options = {
+    env = mkOption {
+      type =
+        with types;
+        lazyAttrsOf (oneOf [
+          str
+          path
+          int
+          float
+        ]);
+      description = "Environment variables to set in the neovim wrapper.";
+      default = { };
+      example = {
+        FOO = 1;
+        PI = 3.14;
+        BAR_PATH = "/home/me/.local/share/bar";
+        INFER_MODE = "local";
+        BAZ_MAX_COUNT = 1000;
+      };
+    };
+
     extraPlugins = mkOption {
       type = with types; listOf (nullOr (either package pluginWithConfigType));
       default = [ ];
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -339,7 +339,16 @@ in
           initSource;
 
       extraWrapperArgs = builtins.concatStringsSep " " (
-        (optional (
+        # Setting environment variables in the wrapper
+        (lib.mapAttrsToList (
+          name: value:
+          lib.escapeShellArgs [
+            "--set"
+            name
+            value
+          ]
+        ) config.env)
+        ++ (optional (
           config.extraPackages != [ ]
         ) ''--prefix PATH : "${lib.makeBinPath config.extraPackages}"'')
         ++ (optional config.wrapRc ''--add-flags -u --add-flags "${initFile}"'')
--- a/tests/test-sources/env.nix
+++ b/tests/test-sources/env.nix
@@ -1,0 +1,42 @@
+{
+  env-variables = {
+    env = {
+      FOO = "1";
+      FOO_INT = 42;
+      FOO_FLOAT = 3.14;
+      STRING_PATH = "/home/me/.local/share/bar";
+      REAL_PATH = ./env.nix;
+      BAZ_MAX_COUNT = "1000";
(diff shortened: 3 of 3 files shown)


traxys (approved):

modules/top-level/output.nix
```diff
@@ -339,7 +339,9 @@ in
           initSource;
 
       extraWrapperArgs = builtins.concatStringsSep " " (
-        (optional (
+        # Setting environment variables in the wrapper
+        (lib.mapAttrsToList (var: value: "--set ${var} ${value}") config.env)
```
MattSturgeon, line 343: Do we need `lib.escapeShellArg` here?

tests/test-sources/env.nix
```diff
@@ -0,0 +1,15 @@
+{
+  env-variables = {
+    env = {
+      FOO = "1";
+      BAR_PATH = "/home/me/.local/share/bar";
+      BAZ_MAX_COUNT = "1000";
+    };
+
```
MattSturgeon, line 8: Let's add some tests for edge-cases, such as variable names or values that may need `lib.escapeShellArg`

modules/output.nix
```diff
@@ -28,6 +28,17 @@ let
 in
 {
   options = {
+    env = mkOption {
+      type = with types; attrsOf str;
```
MattSturgeon, line 32: Down the line we could expand the type to also accept e.g. `path` and `package`, but this is fine for an initial starting point.

Idk what type the NixOS or home-manager `environment` option uses?

MattSturgeon (commented):
Awesome 😎

modules/top-level/output.nix
MattSturgeon: I'm wondering if `escapeShellArgs` would be neater:

```nix
lib.escapeShellArgs [ "--set" name value ]
```

modules/output.nix
khaneliman: I thought environment variables could only be strings?

khaneliman: nvm https://github.com/nixos/nixpkgs/blob/master/nixos/modules/config/shells-environment.nix#L41-L41

MattSturgeon: I'm fine with keeping things simple in this PR and opening an issue regarding the other types we could support

khaneliman (approved):
LGTM

modules/top-level/output.nix
```diff
@@ -339,7 +339,9 @@ in
           initSource;
 
       extraWrapperArgs = builtins.concatStringsSep " " (
-        (optional (
+        # Setting environment variables in the wrapper
+        (lib.mapAttrsToList (var: value: "--set ${var} ${lib.escapeShellArg value}") config.env)
```
khaneliman, line 343: or should this be `escapeShellArgs` for the entire list? not too familiar with its usage

MattSturgeon: I'd personally prefer this:

```suggestion
        (lib.mapAttrsToList (name: value: lib.escapeShellArgs [ "--set" name value ]) config.env)
```

MattSturgeon (approved):
:rocket:

modules/output.nix
```diff
@@ -28,6 +28,26 @@ let
...
   options = {
+    env = mkOption {
+      type =
+        with types;
+        lazyAttrsOf (oneOf [
+          str
+          path
+          int
+          float
+        ]);
```
MattSturgeon, line 39: Just a note:

The [nixos impl](https://github.com/NixOS/nixpkgs/blob/3acdee779318c707cfb5735db379b735f314b094/nixos/modules/config/shells-environment.nix#L42-L44) has an `apply` function that coerces path values to `"${value}"` (ensuring they get copied to the nix store) and also does some stuff with list values (although we don't currently support list types).

```nix
apply =
  let
    toStr = v: if lib.isPath v then "${v}" else toString v;
  in
  lib.mapAttrs (n: v: if lib.isList v then lib.concatMapStringsSep ":" toStr v else toStr v);
```

MattSturgeon:
@mergify backport nixos-24.11

Issue #3104, opened 2025-03-24
Avante - typo/wrong attr name


repparw:
https://github.com/nix-community/nixvim/blob/9672b3d217c0340bc1a10a4f0856a3c9c0bd823d/plugins/by-name/avante/default.nix#L17-L22

This should be auto_suggestions_provider

repparw:
yeah already setup my config, just wanted to let you know

khaneliman:
Appreciate it!

PR #3105, opened 2025-03-24
plugins/avante: fix incorrect option name

khaneliman:
Closes https://github.com/nix-community/nixvim/issues/3104
Looks like it was initialized incorrectly from inception. Renaming to match the option name upstream.

Wasn't a valid option before, so not even putting a rename warning in.


MattSturgeon (approved):

khaneliman:
@mergify queue

PR #3121, opened 2025-03-28
flake/dev: rename `nixpkgs` input to avoid shadowing

MattSturgeon:
Rename `inputs.nixpkgs` to `inputs.dev-nixpkgs` to avoid shadowing the root flake's `inputs.nixpkgs`.

Having the dev flake shadow an input can be confusing, since overriding the root flake's input will have no effect when evaluating outputs that depend on the dev partition (i.e. outputs listed in [`partitionedAttrs`](https://flake.parts/options/flake-parts-partitions.html#opt-partitionedAttrs)).

The only reason we specify a nixpkgs input in the dev flake at all, is so the other inputs can follow it.

---

Once nix 2.26 is more prevalent, we can follow the root flake's inputs using a `"path:../.."` input. See [2.26 release notes](https://discourse.nixos.org/t/nix-2-26-released/59211) and https://github.com/NixOS/nix/pull/10089.

This may also require changes in [flake-compat](https://github.com/edolstra/flake-compat), since that is used by flake-parts to load the partition flake, see https://github.com/edolstra/flake-compat/pull/18. The nix-community [fork](https://github.com/nix-community/flake-compat) has [support for `path:` references](https://github.com/nix-community/flake-compat/commit/d1a6788892881f2d15836d0ac1849d844d7b01de), however idk if it is compatible with https://github.com/NixOS/nix/pull/10089 which introduced lockfile changes in nix 2.26. I'm also unsure exactly which version of flake-compat is in use by flake-parts.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -126,7 +126,7 @@ jobs:
           root_nixpkgs=$(nix eval --raw --file . 'inputs.nixpkgs.rev')
           old=$(git show --no-patch --format=%h)
           nix flake update --commit-lock-file \
-              --override-input nixpkgs "github:NixOS/nixpkgs/$root_nixpkgs" \
+              --override-input 'dev-nixpkgs' "github:NixOS/nixpkgs/$root_nixpkgs" \
               --flake './flake/dev'
           new=$(git show --no-patch --format=%h)
           if [ "$old" != "$new" ]; then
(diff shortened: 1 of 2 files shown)


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #3129, opened 2025-03-30
plugins/ltes-extra: support ltex_plus

Kirens:
ltex_extra reacently got support for ltex_plus, a more updated and still maintained fork of ltex (https://github.com/barreiroleo/ltex_extra.nvim/pull/66)

This PR adds support for both ltex, and ltex_plus, but defaults to using ltex_plus if neither is enabled. I've tried adding a warning of the change to notify users who depend on the implicit enabling of ltex.

I'm open to suggestions if there are other approaches to similar situations in Nixvim.


plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,59 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              ltex.onAttach.function != attachLua
+              || ltex.onAttach.override != false
+              ||
+                [ ] != collect (v: !(isAttrs v || v == null)) (
+                  ltex
+                  // {
+                    extraSettings = null;
+                    package = null;
+                    onAttach = null;
+                  }
```
GaetanLepage, line 78: Shouldn't those be inverted? Here, `extraSettings` & co will be overridden. Am I missing something?

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,59 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          message = ''
+            You have enabled `ltex-extra` but not the lsp (`plugins.lsp`).
+            You should set `plugins.lsp.enable = true` to make use of the LTeX_extra plugin's features.
+          '';
+        }
+        {
+          # TODO: Added 2025-03-30; remove after 25.05
+          # Warn if someone seem to have an externaly configured servers.ltex
+          when =
+            ltex.enable != true
```
GaetanLepage, line 67:
```suggestion
            (!ltex.enable)
```
Isn't this a boolean?

Kirens: Good catch, I thought it was a nullable value

plugins/by-name/ltex-extra/default.nix
Kirens: So I'm trying to find any `defaultNullOpts` that have been modified. This should mask options that are not a `defaultNullOpt` or, as in the case of `onAttach`, expected to be set by this module.

Kirens:
I've added options to the `ltex_plus` language server. Maybe I should split this into two PRs.

Tests previously failed due to ltex.enable not being masked properly. That is now fixed

plugins/by-name/ltex-extra/default.nix
Kirens: To be even more clear, I use `collect` to find all non-null properties of `lsp.server.ltex`. If there are any such cases, i.e. the result `!= []` then the user configuration has changes other than those previously caused by the `ltex-extra`-module.

GaetanLepage:
> I've added options to the `ltex_plus` language server. Maybe I should split this into two PRs.
> 
> Tests previously failed due to ltex.enable not being masked properly. That is now fixed

First, thanks for the good work, this is looking great!
When it comes to option, I don't think that it is worth it to explicitly declare them.
It is already possible for users to put whatever they want in `settings`. It adds a maintenance load on Nixvim.
I would recommend to drop this part.

GaetanLepage (approved):
LGTM! Thanks

GaetanLepage:
@nix-community/nixvim can someone double-check this PR please? Espacially the `warnings` logic.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-
-    plugins.lsp = {
-      servers.ltex = {
-        # Enable the ltex language server
-        enable = true;
+    in
+    {
+      warnings = lib.nixvim.mkWarnings "plugins.ltex-extra" [
+        {
+          when = !config.plugins.lsp.enable;
```
MattSturgeon, line 57: We're now using `config.plugins.lsp` a couple times. I'd consider having a `lspCfg` binding alongside the `cfg` binding?

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
 
-  extraConfig = cfg: {
-    warnings = lib.nixvim.mkWarnings "plugins.ltex-extra" {
-      when = !config.plugins.lsp.enable;
-      message = ''
-        You have enabled `ltex-extra` but not the lsp (`plugins.lsp`).
-        You should set `plugins.lsp.enable = true` to make use of the LTeX_extra plugin's features.
+  extraConfig =
+    cfg:
+    let
```
MattSturgeon, line 48: Very minor: having a "local" let..in block is causing nixfmt to indent the `extraConfig` value.

We could avoid formatting changes and minimise the git diff by having a let..in block at the top of the file, immediately after the module args.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              ||
+                [ ] != collect (v: !(isAttrs v || v == null)) (
+                  # Remove options that wouldn't be null in an empty ltex config
+                  removeAttrs ltex [
+                    "extraSettings"
+                    "package"
+                    "onAttach"
+                    "enable"
+                  ]
+                )
```
MattSturgeon, line 80: It would be helpful if we told users _which_ options they have configured to trigger this warning, and ideally in which files those options definitions are located.

If that's too much effort for now, could you leave a TODO comment and/or open an issue?

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          '';
+        }
+        {
+          # TODO: Added 2025-03-30; remove after 25.05
+          # Warn if servers.ltex seems to be configured outside of ltex-extra
+          when =
+            !ltex.enable
+            && (
+              ltex.onAttach.function != attachLua
+              || ltex.onAttach.override != false
```
MattSturgeon, line 70: I wonder if it's better to check these options' override priority, to assert that our default value isn't being overridden rather than explicitly checking the actual value.

E.g. `ltexOpt.highestPrio < (mkDefault null).priority` if we were comparing against a `mkDefault` definition.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      ];
 
-        onAttach.function = ''
-          require("ltex_extra").setup(${lib.nixvim.toLuaObject cfg.settings})
-        '';
+      plugins.lsp = {
+        servers.ltex.onAttach.function = attachLua;
+        servers.ltex_plus = {
+          # Enable ltex_plus if ltex is not already enabled
+          enable = mkIf (!ltex.enable) true;
```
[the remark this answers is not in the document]
MattSturgeon: This should also have a `mkDefault` or a `mkOverride 1400` or something, so that users can override it without needing `mkForce`.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -43,26 +43,60 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          onAttach.function = attachLua;
```
MattSturgeon, line 96: Both these `onAttach` definitions should also be between `mkDefault` and `mkOptionDefault` priority.

plugins/by-name/ltex-extra/default.nix
khaneliman: I feel like this is the opposite advice I've been given/giving. I would like to see variable usages scoped as close to usage, as possible. If it's used through top level options, then at the top let block is better. Otherwise, you end up a with giant files that are just variable definitions and then the actual code and you have to jump back to the top to understand it.

plugins/by-name/ltex-extra/default.nix
Kirens: I think I disagree with this. Having a different priority means a user may unintentionally disable this plugin by overriding this hook that's calling the setup.

plugins/by-name/ltex-extra/default.nix
Kirens: It can probably be done that way. But it is no longer necessary as I rewrote this to check for definitions in any unexpected files instead.

plugins/by-name/ltex-extra/default.nix
Kirens: Good feedback, but it seems to additionally be caused by `lib.nixvim.mkWarnings` taking a list of warnings now. But this let placement is a bit awkward and removed in the latest patch.

Kirens:
Changes based on feedback.

Made the warning logic identify additional definitions of `lsp.servers.ltex` to those produced by nixvim modules. This may produce false positives as it doesn't take priority into account, but keeps it at a reasonable complexity.

The formatting is a bit awkward for the warning. I guess it's sufficient this for temporary code, however.

plugins/by-name/ltex-extra/default.nix
MattSturgeon: > I feel like this is the opposite advice I've been given/giving

Yeah, there's pros and cons to both approaches.

In this case, I was thinking of the diff caused by formatting changes.

One pro for defining bindings with a more global scope is that they can usually be optimised and reused more efficiently by nix. You can observe this by adding a `trace` to your bindings; local bindings may print the trace many times while global bindings _often_ only print once.

But from a purely organisational perspective, local bindings are usually better. Sometimes if you get _too_ local, with nested let-blocks, it can end up being hard to read though.

Sometimes I consider not just where the binding is currently used, but also where it is likely to be used in the future. For example `cfg` is usually only used within the `config` definition, but is _often_ also used in `options` (e.g. for defaults); therefore `cfg` often makes sense as a top-level binding.

In this case I don't have any strong feelings.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -44,24 +52,65 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            if acc != null then
+              acc
+            else if e == null then
+              null
+            else
+              {
+                inherit name;
+                inherit (e) file;
+              };
+          external = foldlAttrs anyExternal null options.plugins.lsp.servers.ltex;
```
MattSturgeon, line 84: Interesting approach using the definition location 😎

I wonder if it'd be better to _filter_ for external definitions, instead of reducing to the _first_ match? That way we could list all offending definitions in the warning.

I'm also a little concerned about false-pisitives where there are irrelevant low-prio defs. I can't recall ottomh whether `definitions` includes _all_ definitions or only definitions of the `highestPrio`? I can check this in `nix repl` when I'm next at a PC.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -44,24 +52,65 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        in
+        {
+          # TODO: Added 2025-03-30; remove after 25.05
+          # Warn if servers.ltex seems to be configured outside of ltex-extra
+          when = !lspCfg.servers.ltex.enable && external != null;
+          message = ''
+            in ${external.file}
+            You seem to have configured `plugins.lsp.servers.ltex.${external.name}` for `ltex-extra`.
```
MattSturgeon, line 92: If we have a reference to the option itself, we can interpolate it here.

```nix
"${options.foo}"
=> "programs.nixvim.foo"
```

options come with a `__toString` attr with the value `lib.showOption option.loc`, so interpolating them will stringify to the option path.

This has the advantage of automatically including path prefixes, such as `programs.nixvim`.

plugins/by-name/ltex-extra/default.nix
```diff
@@ -2,16 +2,24 @@
...
+in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "ltex-extra";
   packPathName = "ltex_extra.nvim";
   package = "ltex_extra-nvim";
 
   maintainers = [ maintainers.loicreynier ];
 
+  description = ''
+    This plugin works with both the ltex or ltex_plus language servers and will enable ltex_plus if neither are.
```
MattSturgeon, line 20:
```suggestion
    This plugin works with either the ltex or ltex_plus language servers.
    Enabling this plugin will also enable ltex_plus, unless ltex is already enabled.
```

Is something like this clearer?

plugins/by-name/ltex-extra/default.nix
```diff
@@ -2,16 +2,24 @@
   lib,
   helpers,
   config,
+  options,
   ...
 }:
 with lib;
+let
+  lspCfg = config.plugins.lsp;
```
MattSturgeon, line 10: If you like, we could also have:

```nix
  lspOpts = options.plugins.lsp;
```

If you think that makes anything more readable?

No preference here.

MattSturgeon (commented):
Thanks, this is coming along nicely.

Some thoughts below, but nothing too major I hope.

plugins/by-name/ltex-extra/default.nix
MattSturgeon: Another consideration is if/how `mkIf false` definitions show up. E.g. a user might have a toggle somewhere between a manual configuration of the lsp and using this plugin, implemented using `mkIf`.

plugins/by-name/ltex-extra/default.nix
MattSturgeon, line 20: e.g. this

PR #3138, opened 2025-03-31
modules: refactor plugins code in top-level

GaetanLepage:
This should not change anything functionally. I find that the current state (everything dumped in `top-level.nix` is not ideal.

### DIFF
--- a/modules/top-level/plugins/byte-compile-plugins.nix
+++ b/modules/top-level/plugins/byte-compile-plugins.nix
@@ -1,0 +1,17 @@
+/*
+  Byte compiling of normalized plugin list
+
+  Inputs: List of normalized plugins
+  Outputs: List of normalized (compiled) plugins
+*/
+{ lib, pkgs }:
+let
+  builders = lib.nixvim.builders.withPkgs pkgs;
+
+  byteCompile =
+    p:
+    (builders.byteCompileLuaDrv p).overrideAttrs (
+      prev: lib.optionalAttrs (prev ? dependencies) { dependencies = map byteCompile prev.dependencies; }
+    );
+in
+map (p: p // { plugin = byteCompile p.plugin; })
--- a/modules/top-level/plugins/combine-plugins.nix
+++ b/modules/top-level/plugins/combine-plugins.nix
@@ -1,0 +1,50 @@
+{
+  lib,
+  pkgs,
+  pathsToLink,
+  standalonePlugins,
+}:
+let
+  inherit (import ./utils.nix lib)
+    getAndNormalizeDeps
+    removeDeps
+    ;
+  mkPluginPack = import ./mk-plugin-pack.nix { inherit lib pkgs; };
+
+in
+/*
+  *combinePlugins* function
+
+  Take a list of combined plugins, combine the relevant ones and return the resulting list of plugins
+*/
+normalizedPlugins:
+let
+  # Plugin list extended with dependencies
+  allPlugins =
+    let
+      pluginWithItsDeps = p: [ p ] ++ builtins.concatMap pluginWithItsDeps (getAndNormalizeDeps p);
+    in
+    lib.unique (builtins.concatMap pluginWithItsDeps normalizedPlugins);
+
+  # Separated start and opt plugins
+  partitionedOptStartPlugins = builtins.partition (p: p.optional) allPlugins;
+  startPlugins = partitionedOptStartPlugins.wrong;
+  # Remove opt plugin dependencies since they are already available in start plugins
+  optPlugins = removeDeps partitionedOptStartPlugins.right;
+
+  # Test if plugin shouldn't be included in plugin pack
+  isStandalone =
+    p:
+    builtins.elem p.plugin standalonePlugins || builtins.elem (lib.getName p.plugin) standalonePlugins;
+
+  # Separated standalone and combined start plugins
+  partitionedStandaloneStartPlugins = builtins.partition isStandalone startPlugins;
+  pluginsToCombine = partitionedStandaloneStartPlugins.wrong;
+  # Remove standalone plugin dependencies since they are already available in start plugins
+  standaloneStartPlugins = removeDeps partitionedStandaloneStartPlugins.right;
+
+  # Combine start plugins into a single pack
+  pluginPack = mkPluginPack { inherit pluginsToCombine pathsToLink; };
+in
+# Combined plugins
+[ pluginPack ] ++ standaloneStartPlugins ++ optPlugins
--- a/modules/top-level/plugins/default.nix
+++ b/modules/top-level/plugins/default.nix
@@ -1,0 +1,53 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  inherit (import ./utils.nix lib)
+    normalizedPluginType
+    normalizePlugins
+    ;
+  byteCompileCfg = config.performance.byteCompileLua;
+in
+{
+  options = {
+    build.plugins = lib.mkOption {
+      visible = false;
+      internal = true;
+      readOnly = true;
+      type = lib.types.listOf normalizedPluginType;
+      description = ''
+        Final list of (normalized) plugins that will be passed to the wrapper.
+
+        It notably implements:
+        - byte-compilation (performance.byteCompileLua) -> ./byte-compile-plugins.nix
+        - plugins combining (performance.combinePlugins) -> ./combine-plugins.nix
+      '';
+    };
+  };
+
+  config = {
+    build.plugins =
+      let
+        shouldCompilePlugins = byteCompileCfg.enable && byteCompileCfg.plugins;
+        byteCompilePlugins = import ./byte-compile-plugins.nix { inherit lib pkgs; };
+
+        shouldCombinePlugins = config.performance.combinePlugins.enable;
+        combinePlugins = import ./combine-plugins.nix {
+          inherit lib pkgs;
+          inherit (config.performance.combinePlugins)
+            standalonePlugins
+            pathsToLink
+            ;
+        };
+      in
+
+      lib.pipe config.extraPlugins (
+        [ normalizePlugins ]
+        ++ lib.optionals shouldCompilePlugins [ byteCompilePlugins ]
+        ++ lib.optionals shouldCombinePlugins [ combinePlugins ]
+      );
+  };
+}
--- a/modules/top-level/plugins/mk-plugin-pack.nix
+++ b/modules/top-level/plugins/mk-plugin-pack.nix
@@ -1,0 +1,60 @@
+{ lib, pkgs }:
+let
+  inherit (import ./utils.nix lib) normalizePlugin;
+in
+{ pluginsToCombine, pathsToLink }:
+let
+
+  overridePlugin =
+    plugin:
+    plugin.plugin.overrideAttrs (prev: {
+      nativeBuildInputs = lib.remove pkgs.vimUtils.vimGenDocHook prev.nativeBuildInputs or [ ];
+      configurePhase = ''
+        ${prev.configurePhase or ""}
+        rm -vf doc/tags'';
+    });
+
+  # Every plugin has its own generated help tags (doc/tags)
+  # Remove them to avoid collisions, new help tags
+  # will be generate for the entire pack later on
+  overriddenPlugins = map overridePlugin pluginsToCombine;
+
+  # Gather python 3 dependencies from every plugins
+  python3Dependencies =
+    ps:
+    lib.pipe pluginsToCombine [
+      (builtins.catAttrs "plugin")
+      (builtins.catAttrs "python3Dependencies")
+      (builtins.concatMap (f: f ps))
+    ];
+
+  # Combined plugin
+  combinedPlugin = pkgs.vimUtils.toVimPlugin (
+    pkgs.buildEnv {
+      name = "plugin-pack";
+      paths = overriddenPlugins;
+      inherit pathsToLink;
+
+      # Remove empty directories and activate vimGenDocHook
+      # TODO: figure out why we are running the `preFixup` hook in `postBuild`
+      postBuild = ''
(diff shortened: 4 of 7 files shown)


modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,52 @@
...
+  - byte-compilation (performance.byteCompileLua): in this file
+  - plugins combining (performance.combinePlugins): in ./combine-plugins.nix
+*/
+{
+  lib,
+  config,
+  pkgs,
+  builders,
+  ...
+}:
```
MattSturgeon, line 14: Initially, I'd expect top-level/plugins to be a module. Maybe we can imply a more functional name (verb), something like `normalise-plugins`?

If it is a function, we should avoid `...` to keep the arguments explicit.

---

If it is possible to implement as a module, I would prefer that.

Couldn't we just have an internal `normalisedPlugins` option that this module defines?

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,52 @@
...
+      standalonePlugins
+      pathsToLink
+      ;
+    inherit normalizedPlugins;
+  };
+
+  # Plugins to use in build.package
+  plugins = if config.performance.combinePlugins.enable then combinedPlugins else normalizedPlugins;
+in
+plugins
```
MattSturgeon, line 52: Instead of _returning_ `plugins`, we could turn this into a module and _define_ an internal `build.normalisedPlugins` option (or some other name)

```nix
{
  options.build.normalisedPlugins = mkOption {
    type = listOf package;
    internal = true;
    # ...
  };

  config.build.normalisedPlugins = if config.performance.combinePlugins.enable then combinedPlugins else normalizedPlugins;
}
```

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,102 @@
...
+      standalonePlugins,
+      pathsToLink,
+      normalizedPlugins,
+    }:
+    let
+      # Plugin list extended with dependencies
+      allPlugins =
+        let
+          pluginWithItsDeps =
+            p: [ p ] ++ builtins.concatMap pluginWithItsDeps (normalizePluginList p.plugin.dependencies or [ ]);
```
MattSturgeon, line 71: The one time where I like to have uneccessary parentheses is when passing `or` to a function.

TLDR; `or` is one of the only nix operators that has a higher priority than function application. Since you need parentheses for almost all other cases of applying an expression to a function, I think they aid readability when used with `or` too.

```suggestion
            p: [ p ] ++ builtins.concatMap pluginWithItsDeps (normalizePluginList (p.plugin.dependencies or [ ]));
```

Alternatively, it might be nice to have a separate "get dependencies" function:

```nix
getDeps = p: p.plugin.dependencies or [ ];
```

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,102 @@
+{ lib, pkgs, ... }:
+let
+  inherit (import ./utils.nix) normalizePlugin normalizePluginList;
+
+  # Remove dependencies from all plugins in a list
+  __removeDependencies = ps: map (p: p // { plugin = removeAttrs p.plugin [ "dependencies" ]; }) ps;
```
MattSturgeon, line 6: Why not move to `utils.nix`?

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,102 @@
...
+
+  # Combine a list of plugins into a single normalized plugin "pack"
+  __mkPluginPack =
```
MattSturgeon, line 9: Maybe introduce a new `mk-plugin-pack.nix` file?

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,102 @@
...
+      # Combine start plugins into a single pack
+      pluginPack = __mkPluginPack { inherit pluginsToCombine pathsToLink; };
+
+      # Combined plugins
+      combinedPlugins = [ pluginPack ] ++ standaloneStartPlugins ++ optPlugins;
+    in
+    combinedPlugins;
+in
+{
+  inherit combinePlugins;
```
MattSturgeon, line 101: Why not just return the function itself? Like we do in other single-util-function files (mkLsp, mkCmpPlugin, mkVimPlugin, mkNeovimPlugin, etc).

Returning an attrset implies this file may down-the-line be returning a bunch of util functions, which I think would not be ideal in this case.

I'd also suggest inlining the `combinePlugins` implementation, unless we need a recursive binding? If we do need recursion, we could also use `lib.fix`.

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,52 @@
...
+  byteCompilePlugins =
+    plugins:
+    let
+      byteCompile =
+        p:
+        (builders.byteCompileLuaDrv p).overrideAttrs (
+          prev: lib.optionalAttrs (prev ? dependencies) { dependencies = map byteCompile prev.dependencies; }
+        );
+    in
+    map (p: p // { plugin = byteCompile p.plugin; }) plugins;
```
MattSturgeon, line 29: Consider adding a new `byte-compile-plugins.nix` file too?

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,102 @@
+{ lib, pkgs, ... }:
```
MattSturgeon, line 1: Again, if this isn't a module we should avoid `...`.

We could turn this into a module by having it define an internal option, instead of returning a function. I don't have a strong opinion in this case.

MattSturgeon (commented):
This is much more organized. Thanks!

Seeing your approach has inspired some additional feedback for ways we could improve further, which I've left as comments below.

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,55 @@
...
+  inherit (import ./utils.nix lib)
+    normalizedPluginType
+    normalizePluginList
+    ;
+in
+{
+  options = {
+    finalPlugins = lib.mkOption {
+      visible = false;
+      internal = true;
```
MattSturgeon, line 17: Maybe we also want this to be `readOnly = true`?

It's probably _fine_ for users to override this if they want some custom impl, but since it's intended as internal and as part of the implementation of other options, it _feels_ like it should be read-only.

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,55 @@
...
+
+        # Normalized and optionally byte compiled plugin list
+        normalizedPlugins =
+          let
+            normalized = normalizePluginList config.extraPlugins;
+          in
+          if config.performance.byteCompileLua.enable && config.performance.byteCompileLua.plugins then
+            byteCompilePlugins normalized
+          else
+            normalized;
```
MattSturgeon, line 43: I think this would be neater with `lib.pipe`:

```nix
normalizedPlugins = lib.pipe config.extraPlugins [
  normalizePluginList
  (
    if config.performance.byteCompileLua.enable && config.performance.byteCompileLua.plugins then
      byteCompilePlugins
    else
      lib.id
  )
];
```

or

```nix
normalizedPlugins = lib.pipe config.extraPlugins (
  [
    normalizePluginList
  ]
  ++ lib.optionals (config.performance.byteCompileLua.enable && config.performance.byteCompileLua.plugins) [
    byteCompilePlugins
  ]
);
```

---

Also: maybe we should have a `cfg`-like binding for `config.performance.byteCompileLua`? Maybe `byteCompCfg = config.performance.byteCompileLua`?

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,103 @@
...
+  partitionedStandaloneStartPlugins = builtins.partition isStandalone startPlugins;
+  pluginsToCombine = partitionedStandaloneStartPlugins.wrong;
+  # Remove standalone plugin dependencies since they are already available in start plugins
+  standaloneStartPlugins = removeDependencies partitionedStandaloneStartPlugins.right;
+
+  # Combine start plugins into a single pack
+  pluginPack = __mkPluginPack { inherit pluginsToCombine pathsToLink; };
+
+  # Combined plugins
+  combinedPlugins = [ pluginPack ] ++ standaloneStartPlugins ++ optPlugins;
```
MattSturgeon, line 100: 1. we could just inline this, since the `combinedPlugins` is not used within the `let` block (right?)
2. The name is kinda confusing. IIUC, only `pluginPack` is _combined_ and the rest (`standaloneStartPlugins ++ optPlugins`) are all "normal" plugins?

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,103 @@
...
+            rm -vf doc/tags'';
+        })
+      ) pluginsToCombine;
+
+      # Gather python 3 dependencies from every plugins
+      python3Dependencies =
+        let
+          deps = map (p: p.plugin.python3Dependencies or (_: [ ])) pluginsToCombine;
+        in
+        ps: builtins.concatMap (f: f ps) deps;
```
MattSturgeon, line 31: Maybe another case where `lib.pipe` helps? `catAttrs` too, probably:

```nix
python3Dependencies = ps: lib.pipe pluginsToCombine [
  (builtins.catAttrs "plugin")
  (builtins.catAttrs "python3Dependencies")
  (builtins.concatMap (f: f ps))
];
```

modules/top-level/plugins/byte-compile-plugins.nix
```diff
@@ -0,0 +1,18 @@
+/*
+  Byte compiling of normalized plugin list
+  kljljlkjl
```
MattSturgeon, line 3:
```suggestion
```

modules/top-level/plugins/combine-plugins.nix
```diff
@@ -0,0 +1,103 @@
...
+      # Combined plugin
+      combinedPlugin = pkgs.vimUtils.toVimPlugin (
+        pkgs.buildEnv {
+          name = "plugin-pack";
+          paths = overriddenPlugins;
+          inherit pathsToLink;
+          # Remove empty directories and activate vimGenDocHook
+          postBuild = ''
+            find $out -type d -empty -delete
+            runHook preFixup
```
MattSturgeon, line 42: I know this is out of scope for this PR, but I find it very non-obvious that running `preFixup` within `postBuild` is "activate vimGenDocHook". This feels very very hacky.

MattSturgeon: Do you know the correct approach? If not would you consider adding some additional notes/comments/fixmes?

modules/top-level/plugins/utils.nix
```diff
@@ -0,0 +1,38 @@
...
+      };
+
+      optional = lib.mkOption {
+        type = lib.types.bool;
+      };
+    };
+  };
+
+  # Normalize a plugin in a standard { plugin, config, optional } attrs
+  normalizePlugin =
```
MattSturgeon, line 20: If we're going to use a submodule-type for normalizedPluginType, could we consider not having a "normalize plugins" function and instead rely on the module system (defaults, etc) to do the nomalisation for us?

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,55 @@
...
+  options = {
+    finalPlugins = lib.mkOption {
+      visible = false;
+      internal = true;
+      type = lib.types.listOf normalizedPluginType;
+      description = ''
+        Final list of (normalized) plugins that will be passed to the wrapper.
+
+        It notably implements:
+        - byte-compilation (performance.byteCompileLua): in this file
```
MattSturgeon, line 23: Is this correct, or should it say in `./byte-compile-plugins.nix`?

modules/top-level/plugins/utils.nix
MattSturgeon: Let's hold off and consider it in another PR? As I think we use `normalizePlugin` and `normalizePlugins` in a handful of unrelated places.

modules/top-level/plugins/default.nix
```diff
@@ -0,0 +1,56 @@
...
+    ;
+  byteCompileCfg = config.performance.byteCompileLua;
+in
+{
+  options = {
+    build.plugins = lib.mkOption {
+      visible = false;
+      internal = true;
+      readOnly = true;
+      type = lib.types.listOf normalizedPluginType;
```
MattSturgeon, line 20: We _could_ consider moving the byte-compile impl here, using `apply`:

```nix
apply =
  if byteCompileCfg.enable then
    byteCompilePlugins
  else
    lib.id;
```

MattSturgeon: Actually, I guess that doesn't work because we want to byte-comp _before_ combining?

MattSturgeon: Yeah, if it's a read-only option then there's _not much_ difference between doing stuff in the definition vs the apply-fn.

modules/top-level/plugins/mk-plugin-pack.nix
```diff
@@ -0,0 +1,58 @@
...
+      passthru = {
+        inherit python3Dependencies;
+      };
+    }
+  );
+
+  # Combined plugin configs
+  combinedConfig = builtins.concatStringsSep "\n" (
+    builtins.concatMap (x: lib.optional (x.config != null && x.config != "") x.config) pluginsToCombine
+  );
```
MattSturgeon, line 53: Could do:

```nix
combinedConfig = lib.pipe pluginsToCombine [
  (catAttrs "config")
  (filter (config: config != null && config != ""))
  (builtins.concatStringsSep "\n")
];
```

?

MattSturgeon (approved):
diff lgtm, _much_ better, thanks!

The impl is still kinda hacky, but at least we can mostly understand it now :grin:

PR #3139, opened 2025-04-01
tests: remove `type` passthrough

MattSturgeon:
Nix can only build attrsets with a `type = "derivation"` attr.

Having the configuration's option-type as a passthru attr breaks the ability to build specific test entries, since they are no longer `type = "derivation"`.

Renamed `passthru.type` -> `passthru.optionType` to work around this.


khaneliman (approved):

MattSturgeon:
@mergifyio queue

PR #3143, opened 2025-04-03
modules/spellfiles: init

### LINKED ISSUE #1919: Cannot find word list "fr.utf-8.spl" or "fr.ascii.spl"
| Plugin       | treesitter?  |  
| Nixpkgs      | unstable |  unstable
| Home Manager | unstable |  unstable

## Description

When I open a markdown file, and only those afaik, I get the message "Cannot find word list "fr.utf-8.spl" or "fr.ascii.spl"". How can I install them? I don't actually know what they are, but look like french language dictionaries, which would be nice. I'm an ancient vim user getting used to neovim, so don't really know my way around. I don't really know if this is related to treesitter at all, since it seems to deal with programming languages not natural ones.

I just installed nixvim recently so I've got a pretty vanilla setup.

Thank you.

Open a buffer with a md file
```nix
programs.nixvim = {
}
```

GaetanLepage:
### Motivation
Nix(vim) is all about offering the best out-of-the-box experience.
Hence, why not providing the spellfiles on-demand?
This PR implements the new `spellfiles` top-level option which allows users to decide which spellfiles should be automatically fetched from the [vim ftp server](https://ftp.nluug.nl/pub/vim/runtime/spell/).

Fixes #1919

Under the hood, this option populates `extraFiles."spell/en.utf-8.spl"` (for example) with the files fetched from `https://ftp.nluug.nl/pub/vim/runtime/spell/` with `pkgs.fetchurl`.

### Nixvim's vendored spellfile hashes

I added an extra script to the `update-scripts` folder which pre-fetches the hashes of all spell files hosted on the official FTP server.
Hence, while the spellfiles are downloaded at build-time when users deploy their configurations, they won't need to manually provide the expected hash. This information will already be available in the nixvim repo (in `generated/spellfiles.nix`).

> [!NOTE]  
> One could think that vendoring the hashes in Nixvim and hoping they will still be valid when a user will build his config is a bad idea. Indeed, it it changes, one could have a hash-mismatch error without any fallback to overwrite it.
> My answer to this is that those files are apparently very stable (last updated in 2019), so it is not a huge risk.

### Module options (`cfg = config.spellfiles`)

This behavior is opt-in (`cfg.enable` defaults to `false`).

By default, the module automatically infers the list of spellfiles `LANG.ENCODING.[spl,sug]` to install:
- `LANG`: Obtained from the following sources, in decreading priority:
  - `cfg.languages`
  - `opts.spelllang`
  - `["en"]` (default)
- `ENCODING`: Obtained from the following sources, in decreading priority:
  - `opts.encoding`
  - `"utf-8"` (default)
- `.sug` files: Actual spellfiles are `.spl` files and are always fetched. However, some `LANG.ENCODING` combos have a ["suggestion" file `.sug`](https://neovim.io/doc/user/spell.html#spell-NOSUGFILE) available. They are fetched if `cfg.includeSuggestions` is `true` (defaults to `false`).

Besides, he can also explicitly specify additionnal spellfiles to fetch:
```nix
spellfiles.spellfiles = [
  "en.utf-8.spl"
  "en.utf-8.sug"
  "fr.utf-8.spl"
];
```

_Note:_ At first, I made `cfg.spellfiles` and `cfg.languages` exclusive, but actually, it is better to allow both of them to be set concurrently.

### Design questions

- `cfg.languages` or `cfg.spelllangs`?
- Do we need to provide a `cfg.encoding` option?

### Resources

- Neovim documentation (https://neovim.io/doc/user/spell.html#spell)
- Official spellfiles repo (https://ftp.nluug.nl/pub/vim/runtime/spell/)
- NixOS wiki about vim spellfiles (https://nixos.wiki/wiki/Vim#Vim_Spell_Files)

### Credits
The idea is from [a very smart suggestion](https://github.com/nix-community/nixvim/issues/1919#issuecomment-2774609116) from @mirkolenz.

### DIFF
--- a/modules/spellfiles.nix
+++ b/modules/spellfiles.nix
@@ -1,0 +1,184 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) types;
+
+  spellfilesHashes = lib.importJSON ../generated/spellfiles.json;
+
+  availableSpellfiles = lib.attrNames spellfilesHashes;
+
+  # [ "foo<SEP>bar" "one<SEP>two<SEP>three" ] -> [ "foo" "one" ]
+  mapExtractBeforeSep = sep: map (x: builtins.head (builtins.split sep x));
+
+  availableLanguages = lib.pipe availableSpellfiles [
+    (mapExtractBeforeSep "\\.")
+    lib.unique
+  ];
+
+  cfg = config.spellfiles;
+in
+{
+  options.spellfiles = {
+    enable = lib.mkOption {
+      type = types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to automatically fetch and install vim spellfiles (`.spl` and `.sug`).
+
+        If enabled, Nixvim will automatically populate `spell/en.utf-8.spl`-type files by fetching
+        them from [the official vim FTP server](https://ftp.nluug.nl/pub/vim/runtime/spell/).
+
+        This module allows choosing which spellfiles to install.
+      '';
+    };
+
+    languages = lib.mkOption {
+      type = with types; either (enum [ "auto" ]) (listOf (enum availableLanguages));
+      default = "auto";
+      example = [
+        "en"
+        "fr"
+      ];
+      description = ''
+        By default, spellfiles from the languages listed in `opts.spelllang` will be fetched (or
+        `"en"` if `opts.spelllang` is not defined).
+
+        Set to `[ ]` to disable this behavior.
+      '';
+      apply =
+        value:
+        let
+          # If spelllang is not set, default to English
+          spelllang = config.opts.spelllang or "en";
+
+          /*
+            Infers the list of language ids from `opts.spelllang`:
+
+            Turns: "en_us,fr,de"
+            into: [ "en" "fr" "de" ]
+            # https://neovim.io/doc/user/spell.html#spell-load
+          */
+          languagesFromSpelllang = lib.pipe spelllang [
+            # "en_us,fr,de" -> [ "en_us" "fr" "de" ]
+            (lib.splitString ",")
+
+            # Remove region from language id ("en_us" -> "en")
+            # (map (lang: builtins.trace lang lib.head (builtins.match "^([^_]+)" lang)))
+            (mapExtractBeforeSep "_")
+
+            # Remove languages that do not have corresponding spellfiles (such as "medical")
+            (builtins.filter (lang: builtins.elem lang availableLanguages))
+          ];
+        in
+        if value == "auto" then languagesFromSpelllang else value;
+    };
+
+    includeSuggestions = lib.mkOption {
+      type = types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to also fetch suggestion files (`.sug`) when available.
+
+        Warning, using suggestion files can lead to higher memory usage.
+
+        See https://neovim.io/doc/user/spell.html#spell-NOSUGFILE.
+      '';
+    };
+
+    spellfiles = lib.mkOption {
+      type = with types; listOf (enum availableSpellfiles);
+      default = [ ];
+      description = ''
+        Explicitly specify spellfiles to install.
+
+        Possible values correspond to the spellfiles available at https://ftp.nluug.nl/pub/vim/runtime/spell/.
+      '';
+      example = [
+        "en.utf-8.spl"
+        "en.utf-8.sug"
+        "fr.utf-8.spl"
+      ];
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    extraFiles =
+      let
+        /*
+          Infers the spellfiles names from a language identifier.
+
+          The spellfiles associated with language `LL` are:
+          - `LL.<ENCODING>.spl`: main spell file
+          - `LL.<ENCODING>.sug` (optional): additional file for more precise suggestions
+
+          `<ENCODING>` is inferred from `opts.encoding` (if set) and defaults to `utf-8`.
+        */
+        getFilenamesFromLanguage =
+          lang:
+          let
+            # Default to utf-8 encoding
+            encoding = config.opts.encoding or "utf-8";
+
+            # Main spellfile (.spl)
+            baseName = "${lang}.${encoding}";
+
+            # Optional suggestion spellfile (.sug)
+            sugFilename = "${baseName}.sug";
+            includeSugSpellfile =
+              # user has enabled fetching suggestion spellfiles
+              cfg.includeSuggestions
+
+              # a suggestion spellfile is available for this language-encoding pair
+              && (lib.hasAttr sugFilename spellfilesHashes);
+          in
+          [
+            "${baseName}.spl"
+          ]
+          ++ lib.optional includeSugSpellfile sugFilename;
+
+        # [ "en.utf-8.spl" "en.utf-8.sug" "fr.utf-8.spl" ]
+        spellfiles = lib.unique (
+          # spellfiles from `cfg.languages`
+          (lib.concatMap getFilenamesFromLanguage cfg.languages)
+
+          # explicitly requested spellfiles
+          ++ cfg.spellfiles
+        );
+
+        /*
+          Maps "en.utf-8.spl" to:
+          {
+            name = "spell/en.utf-8.spl";
+
+            value.source = pkgs.fetchurl {
+              url = "https://ftp.nluug.nl/pub/vim/runtime/spell/en.utf-8.spl";
+              hash = "sha256-/sq9yUm2o50ywImfolReqyXmPy7QozxK0VEUJjhNMHA=";
+            };
+          }
+        */
+        fetchSpellFile = filename: {
+          # Destination path for this spellfile
+          name = "spell/${filename}";
+
+          value.source = pkgs.fetchurl (
+            spellfilesHashes.${filename}
+              or (throw "Spellfile `${filename}` is not available in https://ftp.nluug.nl/pub/vim/runtime/spell/")
+          );
+        };
+      in
+      /*
+        {
+          "spell/en.utf-8.spl".source = "/nix/store/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-en.utf-8.spl"
+          "spell/en.utf-8.sug".source = "/nix/store/yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy-en.utf-8.sug"
+          "spell/fr.utf-8.spl".source = "/nix/store/zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz-fr.utf-8.spl"
+        }
+      */
+      lib.listToAttrs (map fetchSpellFile spellfiles);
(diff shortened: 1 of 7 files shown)


mirkolenz:
Wow, this new module looks really nice! Regarding your questions:

> cfg.languages or cfg.spelllangs?

I think `cfg.languages` is more intuitive here.

> Do we need to provide a cfg.encoding option?

It could even be useful to have `cfg.encodings` to support multiple values here (e.g., users may want to install different encodings if they work in different environments).

modules/spellfiles.nix
```diff
@@ -0,0 +1,175 @@
...
+    extraFiles =
+      let
+        /*
+          Infers the list of language ids from `opts.spelllang`:
+
+          Turns: "en_us,fr,de"
+          into: [ "en" "fr" "de" ]
+          # https://neovim.io/doc/user/spell.html#spell-load
+        */
+        languagesFromSpelllang = lib.pipe (config.spelllang or "en") [
```
MattSturgeon, line 94: You can't use `or` as a fallback for an option value; if the option exists, it will always be present in `config`.

Instead, you can check `options.spelllang.isDefined`, introduce a separate boolean option, or define a `default` for the option.

In this case you could do:
```nix
lib.pipe options.spelllang [
  (opt: if opt.isDefined then opt.value else "en")
```
Although, isn't this effectively setting a default for the option, in which case that'd be better?

mirkolenz:
> Well, I don't know... At this point, users that want to customize both languages and encodings can directly explicitly list the spellfiles they want (with the spellfiles.spellfiles option), right?

Yeah that makes sense!

> Also, I am curious about why a user would want an encoding different than utf-8. Are there common situations where you would need to fallback to latin1?

I think that mostly happens when dealing with legacy systems or when opening old files (e.g., latex files from like ten years ago where often saved as latin1 due to encoding issues with utf-8).

modules/spellfiles.nix
```diff
@@ -0,0 +1,175 @@
...
+
+          # "en_us,fr,de" -> [ "en_us" "fr" "de" ]
+          (lib.splitString ",")
+
+          # Remove region from language id ("en_us" -> "en")
+          (map (lang: lib.head (builtins.match "^([^_]+)" lang)))
```
MattSturgeon, line 100: This is doing the same thing as L15, but with a slightly different approach.

Could we decide on a preferred approach and introduce a common function?

modules/spellfiles.nix
```diff
@@ -0,0 +1,175 @@
...
+        ];
+        languages = if cfg.languages == "auto" then languagesFromSpelllang else cfg.languages;
```
MattSturgeon, line 102: Fancy idea: could we move this to `cfg.languages` apply function?

That way users can _read_ the value to figure out what languages their "auto" definition actually ended up producing.

Format normalisation could also be done there.

Also makes me wonder if the option should be a list type? 🤔 And maybe auto should be a separate bool option?

modules/spellfiles.nix
```diff
@@ -0,0 +1,175 @@
...
+        /*
+          Maps "en.utf-8.spl" to:
+          {
+            name = "spell/en.utf-8.spl";
+
+            value.source = pkgs.fetchurl {
+              url = "https://ftp.nluug.nl/pub/vim/runtime/spell/en.utf-8.spl";
+              hash = "sha256-/sq9yUm2o50ywImfolReqyXmPy7QozxK0VEUJjhNMHA=";
+            };
+          }
```
MattSturgeon, line 154: I'm a little concerned about our pinning technique.

Firstly, I assume that spell files are occasionally modified, and don't offer a versioned/pinned mirror?

This leads to my issue with the assumption that we can generate a table of file hashes without running into issues.

E.g. a user rolls back to an old nixvim revision: now the hash is outdated \
Or: nixvim is slow to update: same outcome... \
Or: someone is bisecting their config \
Or: we are bisecting nixvim (which has a test for spell files)

I think we either need to download the actual spell files in our generate script, or find a way to have versioned URLs.

MattSturgeon (changes requested):
This looks great! I've left some feedback below.

My only major concern is the way we're pinning URLs and hashes; I think this would too easily get out of sync and result in "old" configs being un-buildable.

update-scripts/fetch-spellfiles.nix
```diff
@@ -0,0 +1,44 @@
...
+
+      # Special characters (like '%40', standing for `@`) are invalid for nix store paths
+      # -> We replace all non-alphanumeric characters with '_' and used this sanitized filename for
+      # the derivation name
+      sanitized_filename="''${filename//[^a-zA-Z0-9]/_}"
+      sha256=$(nix-prefetch-url "$url" --name "$sanitized_filename")
+
+      hash=$(nix hash convert --to sri --hash-algo sha256 "$sha256")
+
+      # Ugly hardcoding of the `%40` -> `@` substitution
```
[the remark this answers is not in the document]
MattSturgeon: There's some different approaches here: https://stackoverflow.com/questions/6250698/how-to-decode-url-encoded-string-in-shell

[One answer](https://stackoverflow.com/a/21693459) suggests using python's [`urllib`](https://docs.python.org/3/library/urllib.html#module-urllib):

```sh
python3 -c "import sys; from urllib.parse import unquote; print(unquote(sys.stdin.read()));"
```

Although, if we're depending on python it's probably tempting to write the whole script in python instead of bash?

modules/spellfiles.nix
MattSturgeon: > I don't really see a benefit there...

Definition merging, and just generally having a more structural config.

If the option is a list-type, then I can define `[ "en" ]` in one module, define `[ "fr" ]` in another module, and it will be merged as `[ "en" "fr" ]`. With a str-type option, merging is not supported.

modules/spellfiles.nix
MattSturgeon: Ah my mistake, I think I must've skimmed over the comment and thought it was doing the same thing.

modules/spellfiles.nix
MattSturgeon: Or maybe you meant the separate auto-option? I think the benefit of that separation is simplifying the types and the impl, making things more obvious to end-users.

If you had a separate auto-toggle, you wouldn't need any `if value == "auto"`; in fact, you wouldn't need an `apply` at all: you could just have an extra definition that sets `config.spellfiles.languages = lib.pipe ...`

modules/spellfiles.nix
MattSturgeon: > `opts.spelllang` is not an option. [...] There's no option we could set the default of here.

Yeah, I was thrown off by the missing `opts` in the option path. My above comment assumed we had a top-level option named `spelllang`, which we don't.

`?` and `or` _are_ correct when dealing with an attribute _of_ an option.

MattSturgeon:
> Of course, I have thought about this issue of "outdated hashed" (see the PR description).

Oops, I missed that.

> Indeed, **the spellfiles haven't been modified in the past 6+ years**.

That is reassuring, if nothing else :grinning: 

> This doesn't mean that they won't get modified again at some point, but I would consider them to be very stable.

For me the issue is that when they inevitably do change, all configs that had been using the spellfile module will suddenly break without having been changed. This means the module is technically non-deterministic.

It's the same reason why you would specify a package src using a commit or a tag, rather than a branch ref, in nixpkgs. We want _old_ revisions of nixpkgs to continue building indefinitely.

> I know that this feels _wrong_ or _flaky_, but when you think about it ~90% of Nixvim's code (if not more) is tracking waaaay more unstable upstream specifications (plugins, nixpkgs tooling and packages).

I think the key difference is that all those other things are handled by our `flake.lock`; e.g. nvim api changes are handled by having a specific revision of nvim pinned (via pinning nixpkgs)

---

All that said: I don't see a viable alternative, short of finding or hosting our own mirror. Which I don't want to do, either.

It's great that the files seem to have been stable for 6 years,[^1] but it isn't great that they seemingly don't offer a stable/versioned URL we can use.

I did see their [README](https://ftp.nluug.nl/pub/vim/runtime/spell/README.txt) has instructions for building spell files from source. That _may_ be a better alternative, especially if it can be done in such a way that the nixpkgs and/or nix-community binary cache has a cached build of the spellfile packages.

[^1]: This also makes me question whether this is the most up-to-date mirror...

update-scripts/fetch-spellfiles.nix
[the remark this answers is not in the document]
MattSturgeon: Yeah, I was just listing various options. I'd suggest using python's urllib as the simplest and most robust approach, without _too_ much over-engineering.

MattSturgeon:
> * We could expose the spellfiles in `nixpkgs` and have them cached by Hydra (`pkgs.[neo]vim.passthru.spellfiles`). That would be the best alternative I think. Although, (apart from building them from source) we would have the same URL-pinning issue there...

Vim spellfiles are apparently built from libreoffice dictionaries, which I assume are already packaged. If we can reproduce the from-source instructions as a derivation, it should be possible to package the vim spell files as derivations with the nixpkgs libreoffice dicts as nativeBuildInputs

<details><summary>Build steps</summary>
<p>

This involves downloading the files from the libreoffice github, applying a
patch, and running Vim to generate the `.spl` file.

To do this all in one go use the [Aap program](https://www.a-a-p.org).  It's simple to install, it only requires Python.

Before generating spell files, verify your system has the required locale
support.  Source the `check_locales.vim` script to find out.  If something is
missing, see LOCALE below.


You can also do it manually:
1. Fetch the right spell file from:
   https://github.com/LibreOffice/dictionaries

2. Unzip the archive:
   ```sh
   unzip LL_RR.zip
   ```

4. Apply the patch:
   ```sh
   patch < LL_RR.diff
   ```

5. If the language has multiple regions do the above for each region.  E.g.,
   for English there are five regions: US, CA, AU, NZ and GB.

6. Run Vim and execute `:mkspell`.  Make sure you do this with the correct
   locale, that influences the upper/lower case letters and word characters.
   On Unix it's something like:
   ```vim
   env LANG=en_US.UTF-8 vim
   mkspell! en en_US en_AU en_CA en_GB en_NZ
   ```

7. Repeat step 5 for other locales.  For English you could generate a spell
   file for latin1, utf-8 and ASCII.  ASCII only makes sense for languages
   that have very few words with non-ASCII letters.

**LOCALE**

For proper spell file generation the required locale must be installed. Otherwise Vim doesn't know what are letters and upper-lower case differences. Modern systems use UTF-8, but we also generate spell files for 8-bit locales
for users with older systems.

On Ubuntu the default is to only support locales for your own language.  To
add others you need to do this:

```sh
sudo vim /var/lib/locales/supported.d/local
# Add needed lines from /usr/share/i18n/SUPPORTED
sudo dpkg-reconfigure locales
```

When using the `check_locales.vim` script, you need to exit Vim and restart it
to pickup the newly installed locales.

</p>
</details> 

I guess we'd have to vendor the patch files within nixpkgs, unless there's a safe url from which they can be fetched (are they in a git repo?).

The thing I'm most unsure of is enabling all required locales within the nix builder sandbox. I guess we'd start by running `check_locales.vim` in the build sandbox and printing the output to `$out`?

PR #3152, opened 2025-04-06
plugins/snacks: add indent option

Mrid22:
Added option to enable indent plugin in snacks Note: this isn't support for the full config of that plugin, just the enable option, will add more later

### DIFF
--- a/plugins/by-name/snacks/default.nix
+++ b/plugins/by-name/snacks/default.nix
@@ -41,8 +41,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
     indent = {
       enabled = defaultNullOpts.mkBool true ''
-                Wether to enable `indent` plugin.
-        			'';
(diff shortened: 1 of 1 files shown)


Mrid22:
> Thanks! Although, we don't need to explicitly declare all `settings` options for users to be able to use them?

True, its just a convenience thing. That technically goes for all of these options, doesn't it?

Mrid22:
> > > Thanks! Although, we don't need to explicitly declare all `settings` options for users to be able to use them?
> > 
> > 
> > True, its just a convenience thing. That technically goes for all of these options, doesn't it?
> 
> Yes exactly. Now, all our `plugins.*.settings` options are "freeform", which means that users can put arbitrary values. The entire attrs is then converted to Lua automatically.

Oh, sorry, then I guess this is kinda uneccesary

PR #3162, opened 2025-04-06
modules/dependencies: init

GaetanLepage:
## Motivation

Currently, we manage "dependencies" (i.e. executable packages needed for certain plugins to work) by adding a `plugins.<PLUGIN>.fooPackage` `null`able option.
The issue is that we now have several of these options for the same package (`git` for instance) distributed across the code-base.

So, a user wanting to swap `pkgs.git` for `pkgs.gitMinimal` in his Nixvim closure would have to look for all instances of `gitPackage` options and set them to `pkgs.gitMinimal`.
It would be better to centralize this dependency management.

## Solution

Substitute these options for a `dependencies` top-level module.
Each attribute under `dependencies` is itself an attrs containing two options: `enable` (to install the package) and `package` to customize which package is installed:

```nix
dependencies = {
  git.package = pkgs.gitMinimal;
  curl.enable = false;
};
```

On the plugin module side, this change requires:
- Removing the declarations of the `plugins.*.fooPackage` options.
- Adding a _deprecation module_ (`mkRemovedOptionModule`) for each of the `plugins.*.fooPackage` options removed.
- Replacing `extraPackages = [ cfg.fooPackage ];` with `dependencies.foo.enable = lib.mkDefault true;`
    _Note:_ Later, we can add an argument to `mk{Neovim,Vim}Plugin`: `dependencies = ["foo"];` that would abstract this.

## PR content

- **modules/dependencies: init + add curl**
- **modules/dependencies: add ueberzug**
- **modules/dependencies: add git**

### DIFF
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -129,7 +129,10 @@ rec {
       packageName,
       oldPackageName ? packageName,
     }:
-    lib.mkRemovedOptionModule [ "plugins" plugin "${oldPackageName}Package" ] ''
+    let
+      optionPath = [ "plugins" ] ++ (lib.toList plugin) ++ [ "${oldPackageName}Package" ];
+    in
+    lib.mkRemovedOptionModule optionPath ''
       Please use the `dependencies.${packageName}` top-level option instead:
       - `dependencies.${packageName}.enable = false` to disable installing `${packageName}`
       - `dependencies.${packageName}.package` to choose which package to install for `${packageName}`.
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -122,4 +122,16 @@ rec {
       nestedTypes.coercedType = oldType;
       nestedTypes.finalType = newType;
     };
+
+  mkRemovedPackageOptionModule =
+    {
+      plugin,
+      packageName,
+      oldPackageName ? packageName,
+    }:
+    lib.mkRemovedOptionModule [ "plugins" plugin "${oldPackageName}Package" ] ''
+      Please use the `dependencies.${packageName}` top-level option instead:
+      - `dependencies.${packageName}.enable = false` to disable installing `${packageName}`
+      - `dependencies.${packageName}.package` to choose which package to install for `${packageName}`.
+    '';
 }
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -44,6 +44,7 @@ let
     typst.default = "typst";
     ueberzug.default = "ueberzugpp";
     websocat.default = "websocat";
+    wezterm.default = "wezterm";
     which.default = "which";
     yazi.default = "yazi";
     yq.default = "yq";
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -27,6 +27,7 @@ let
       default = "nodejs";
       example = "pkgs.nodejs_22";
     };
+    plantuml.default = "plantuml";
     ripgrep.default = "ripgrep";
     sd.default = "sd";
     sed.default = "gnused";
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -49,6 +49,10 @@ let
     websocat.default = "websocat";
     wezterm.default = "wezterm";
     which.default = "which";
+    xxd.default = [
+      "unixtools"
+      "xxd"
+    ];
     yazi.default = "yazi";
     yq.default = "yq";
     zk.default = "zk";
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -46,6 +46,7 @@ let
     websocat.default = "websocat";
     which.default = "which";
     yazi.default = "yazi";
+    zk.default = "zk";
   };
 
   mkDependencyOption = name: properties: {
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -9,6 +9,7 @@ let
 
   packages = {
     curl.default = "curl";
+    gcc.default = "gcc";
     git = {
       default = "git";
       example = [ "gitMinimal" ];
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -18,6 +18,7 @@ let
     };
     go.default = "go";
     lean.default = "lean4";
+    ledger.default = "ledger";
     nodejs = {
       default = "nodejs";
       example = "pkgs.nodejs_22";
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -28,6 +28,7 @@ let
     };
     glow.default = "glow";
     go.default = "go";
+    godot.default = "godot_4";
     lazygit.default = "lazygit";
     lean.default = "lean4";
     ledger.default = "ledger";
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -27,6 +27,7 @@ let
       default = "nodejs";
       example = "pkgs.nodejs_22";
     };
+    ripgrep.default = "ripgrep";
     texpresso.default = "texpresso";
     tinymist.default = "tinymist";
     tree-sitter.default = "tree-sitter";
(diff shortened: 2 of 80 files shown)


MattSturgeon:
Commends on the op; I haven't looked at the diff yet:

> * Removing the declarations of the `plugins.*.fooPackage` options.

Alternatively, these could be marked `visible = false`, have their default removed, and trigger a warning if `isDefined`. See below:

> * Adding a _deprecation module_ (`mkRemovedOptionModule`) for each of the `plugins.*.fooPackage` options removed.

Maybe we should try to use a rename alias instead?

Possible issues:

1. multiple distinct definitions would be a merge conflict - no better/worse than an assertion failure
2. if `dependencies.*.package` is not nullable, they'd be a type mismatch from old/new option

We could address #2 by implementing a custom/bespoke alias option instead of using `mkRenamedOptionModule`, if it's worth the effort.

E.g. we could address nullability by having any non-null definition trigger `dependencies.*.enable = true` while _all_ definitions being null would trigger `dependencies.*.enable = false`. If some are null while others are not, we could use either true or false and add an additional warning.

On the other hand, I assume that `plugins.*.*Package` options are not used _that_ often. If so, a removal assertion may not be too disruptive.

EDIT: Ah, this seems to be the goal of `mkRemovedPackageOptionModule`?

> * Replacing `extraPackages = [ cfg.fooPackage ];` with `dependencies.foo.enable = lib.mkDefault true;`

:+1:

modules/dependencies.nix
```diff
@@ -0,0 +1,30 @@
...
+      curl
+      git
+      ;
+    ueberzug = pkgs.ueberzugpp;
+  };
+
+  mkDependencyOption = name: default: {
+    enable = lib.mkEnableOption "Add ${name} to dependencies.";
+
+    package = lib.mkPackageOption pkgs name { };
```
MattSturgeon, line 21: `default` is unused.

---

IMO it is simpler to stick with string/list names/paths rather than trying to work with "real" packages, especially as that is how `mkPackageOption` is designed.

Additionally, do we anticipate a specific package warranting a bespoke `example`?

Maybe we need something like:

```nix
packages = {
  curl.default = ["curl"];
  git.default = ["git"];
  git.example = ["gitMinimal"];
  ueberzug.default = ["ueberzugpp"];
};

mkDependencyOption = name: args: {
  enable = lib.mkEnableOption "Install ${name} with nixvim. Its executables will be added to vim's `PATH`.";
  package = lib.mkPackageOption pkgs name args;
};
```

?

modules/dependencies.nix
```diff
@@ -0,0 +1,30 @@
...
+  };
+in
+{
+  options.dependencies = lib.mapAttrs mkDependencyOption packages;
+
+  config = {
+    extraPackages = lib.mapAttrsToList (_: p: if p.enable then p.package else null) cfg;
```
MattSturgeon, line 28: Feels simpler to filter on `enable`? Rather than adding a bunch of null definitions?

```nix
extraPackages = lib.pipe cfg [
  (lib.filterAttrs (_: dep: dep.enable))
  (builtins.catAttrs "package")
  # or (builtins.map (dep: dep.package))
]
```

tests/test-sources/modules/dependencies.nix
```diff
@@ -0,0 +1,10 @@
+{
+  override =
+    { pkgs, ... }:
+    {
+      dependencies.git = {
+        enable = true;
+        package = pkgs.gitMinimal;
+      };
+    };
+}
```
MattSturgeon, line 10: We should probably have an `all-plugins` test, just to check that all plugin names we declare actually exist in our pinned nixpkgs and can be installed on the current platform:

```nix
config.dependencies = mapAttrs (name: opt: {
  enable = true;
}) options.dependencies;
```

Bonus points if we have a second test for the examples: (note this would need to parse a `literalExpression`):

```nix
config.dependencies = lib.pipe options.dependencies [
  (filterAttrs (name: opt: opt.package ? example))
  (mapAttrs (name: opt: {
    enable = true;
    package = lib.pipe opt.package.example [
      # Map to text
      (
        example:
        assert example._type or null == "literalExpression" && example ? text;
        example.text
      )
      # Split human readable path into a list-path
      (lib.splitString ".")
      # Drop "pkgs" head from path
      (lib.drop 1)
      (loc: lib.attrByPath loc (throw "Example for `${opt.package}` (`${lib.showAttrPath loc}`) not found in pkgs") pkgs)
    ];
  }))
];
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -162,7 +163,7 @@ in
   config = mkIf cfg.enable {
     extraPlugins = [ cfg.package ];
 
-    extraPackages = [ cfg.gitPackage ];
+    dependencies.git.enable = lib.mkDefault true;
```
MattSturgeon, line 166: This ends up being undocumented at a per-plugin level. Before it was documented by the `plugins.*.fooPackage` option itself.

For now, we probably need a note in the plugin's description (or enable option?) mentioning that enabling this plugin will enable `dependencies.foo.enable` by default.

lib/deprecation.nix
```diff
@@ -122,4 +122,12 @@ rec {
...
       nestedTypes.finalType = newType;
     };
+
+  mkRemovedPackageOptionModule =
+    pluginName: packageName:
+    lib.mkRemovedOptionModule [ "plugins" pluginName "${packageName}Package" ] ''
+      Please use the `dependencies.${packageName}` top-level option instead:
+      - `dependencies.${packageName}.enable = false` to disable installing `${packageName}`
+      - `dependencies.${packageName}.package` to choose which package to install for `${packageName}`.
+    '';
```
MattSturgeon, line 132: It would not be an excessive amount of work to write our own module here that offers a more graceful migration period.

Such a change likely wouldn't affect the treewide diff.

Is this something you're opposed to in principal, or something you aren't interested in working on personally? I won't block the PR on this point, but I might be interested in tackling it as follow-up or parallel work.

modules/dependencies.nix
```diff
@@ -0,0 +1,52 @@
...
+  };
+in
+{
+  options.dependencies = lib.mapAttrs mkDependencyOption packages;
+
+  config = {
+    extraPackages = lib.pipe cfg [
+      (lib.filterAttrs (_: p: p.enable))
+      (lib.mapAttrsToList (_: p: p.package))
+    ];
```
MattSturgeon, line 50: Alt approach, since we never need the attr name here, we can be _slightly_ more simple & efficient by working with a list:
```nix
extraPackages = lib.pipe cfg [
  builtins.attrValues
  (builtins.filter (p: p.enable))
  (builtins.map (p: p.package))
];
```

modules/dependencies.nix
```diff
@@ -0,0 +1,52 @@
...
+    ueberzug = "ueberzugpp";
+    websocat = "websocat";
+    which = "which";
+    yazi = "yazi";
+  };
+
+  mkDependencyOption = name: default: {
+    enable = lib.mkEnableOption "Add ${name} to dependencies.";
+
+    package = lib.mkPackageOption pkgs name { inherit default; };
```
MattSturgeon, line 40: I did like the idea of being able to specify an `example` alongside the `default`. Happy to do it in a follow up PR if preferred.

If you just like not needing `.default =` for most cases, we could check the second param's type:

```nix
mkDependencyOption =
  name: arg:
  let
    # optional let-binding; could probably fit inline and still be readable.
    args = if builtins.isAttrs arg then arg else { default = arg; };
  in
```

MattSturgeon (approved):
This LGTM. It's a definite improvement to how we handle dependencies IMO.

I'd like another maintainer to have acknowledged the overall design before merging. cc @khaneliman & @traxys.

Potential future work:
- [ ] add a test for package examples (see [comment](https://github.com/nix-community/nixvim/pull/3162#discussion_r2031163873))
- [ ] add a "softer" deprecation implementation (see [comment](https://github.com/nix-community/nixvim/pull/3162#discussion_r2031191608))
- [ ] add per-plugin docs for the plugin's dependencies (see [comment](https://github.com/nix-community/nixvim/pull/3162#discussion_r2031183874))
- [ ] add an internal option for declaring packages

khaneliman (approved):

PR #3166, opened 2025-04-07
plugins/roslyn: init

KennethHoff:


plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  extraConfig = cfg: {
+    extraPlugins = [ pkgs.vimPlugins.roslyn-nvim ];
+    extraConfigLua = ''
+      require('roslyn').setup({
+        -- roslyn-ls adds this command
+        exe = 'Microsoft.CodeAnalysis.LanguageServer'
+      })
+    '';
+  };
+
```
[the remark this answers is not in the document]
KennethHoff: > This entire snippet[...]


Does that include this, because I believe it's important: 

```
exe = 'Microsoft.CodeAnalysis.LanguageServer'
```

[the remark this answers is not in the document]
MattSturgeon: > I think it would be ideal if `Microsoft.CodeAnalysis.LanguageServer` (and by extension `roslyn-ls`) is not a globally available command, and would be installed scoped to nix somehow; is that possible? `exe = "/nix/store/**/*/Microsoft.CodeAnalysis.LanguageServer"` perhaps, or is this a bad practice?

This is how nix operates by default. Everything is installed into the nix store (`/nix/store`) and is made accessible by modifying env vars (usually `PATH`).

In nixvim's case, the `nvim` exe is actually a "wrapper" script that sets env variables like `PATH` and then runs the _actual_ underlying `nvim` exe (making nixvim's packages available on the `PATH` only within neovim's scope).

Wrappers like this are standard practice for many (most?) nixpkgs packages too.

Random tidbit: _shells_ work in a similar way, by running a bash shell with env vars like `PATH` modified. NixOS and home-manager also do something similar, the difference being their modifications happen during boot/login, so they are "visible" in any scope/shell/context.

plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, pkgs, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "roslyn";
+  packPathName = "roslyn.nvim";
+  package = "roslyn-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  extraConfig = cfg: {
+    extraPlugins = [ pkgs.vimPlugins.roslyn-nvim ];
```
MattSturgeon, line 10:
```suggestion
```

`mkNeovimPlugin` already does this; that's what the `package` argument is for above.

plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,21 @@
...
+    extraConfigLua = ''
+      require('roslyn').setup({
+        -- roslyn-ls adds this command
+        exe = 'Microsoft.CodeAnalysis.LanguageServer'
```
MattSturgeon, line 14: As @GaetanLepage said, we don't normally set defaults for `setup()` options.

It can be done with the `plugins.roslyn.settings` option if needed, but usually such auto-config would be gated behind a non-settings option (e.g. `plugins.roslyn.configureLsp`).

Again, for enablign the language server: this isn't normally done by nixvim plugins, however there are a handful of exceptions. Unless it was essentially a requirement and it never makes sense to have the plugin without the lsp, we would usually want this gated behind a non-settings option.

FYI: when working with language servers in nixvim, you typically use the `plugins.lsp` and `plugins.lsp.servers.*` options.

Having the non-settings toggle option allows users to enable/disable our custom stuff if they'd rather do it themselves a different way.

plugins/by-name/roslyn/default.nix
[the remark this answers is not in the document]
KennethHoff: > > I think it would be ideal if `Microsoft.CodeAnalysis.LanguageServer` (and by extension `roslyn-ls`) is not a globally available command, and would be installed scoped to nix somehow; is that possible? `exe = "/nix/store/**/*/Microsoft.CodeAnalysis.LanguageServer"` perhaps, or is this a bad practice?
> 
> This is how nix operates by default. Everything is installed into the nix store (`/nix/store`) and is made accessible by modifying env vars (usually `PATH`).
> 
> In nixvim's case, the `nvim` exe is actually a "wrapper" script that sets env variables like `PATH` and then runs the _actual_ underlying `nvim` exe (making nixvim's packages available on the `PATH` only within neovim's scope).
> 
> Wrappers like this are standard practice for many (most?) nixpkgs packages too.
> 
> Random tidbit: _shells_ work in a similar way, by running a bash shell with env vars like `PATH` modified. NixOS and home-manager also do something similar, the difference being their modifications happen during boot/login, so they are "visible" in any scope/shell/context.

A lot of cool trivia there. Thanks 

I was aware of this symlink bonanza, but what I was referring to in particular is that if we install the `roslyn-ls` package (which I'm suggesting we do) the end user gets the `Microsoft.CodeAnalysis.LanguageServer` program added to their PATH. This is not a problem, but it's something they might not expect to happen from setting `nixvim.plugins.roslyn.enable = true;`.

So my question is; can we use the `roslyn-ls` package without adding that program to the PATH?

plugins/by-name/roslyn/default.nix
MattSturgeon, line 14: As @GaetanLepage said, we don't normally set defaults for `setup()` options.

It can be done with the `plugins.roslyn.settings` option if needed, but usually such auto-config would be gated behind a non-settings option (e.g. `plugins.roslyn.configureLsp`).

Again, for enablign the language server: this isn't normally done by nixvim plugins, however there are a handful of exceptions. Unless it was essentially a requirement and it never makes sense to have the plugin without the lsp, we would usually want this gated behind a non-settings option.

FYI: when working with language servers in nixvim, you typically use the `plugins.lsp` and `plugins.lsp.servers.*` options.

Having the non-settings toggle option allows users to enable/disable our custom stuff if they'd rather do it themselves a different way.

---

Finally, this call to `setup` is automatically done by `mkNeovimPlugin`. It converts `settings` from nix to lua, and calls `require('roslyn').setup(settings)`.

```suggestion
```

plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,21 @@
...
+      })
+    '';
+  };
+
+  # TODO: Figure out how to add this package.
+  # home.packages = [ pkgs.roslyn-ls ];
```
MattSturgeon, line 20: Assuming this is a language server, what we probably want is to enable `plugins.lsp.servers.roslyn.enable`. However, this option doesn't seem to exist.

Our lsp server options are generated from nvim-lspconfig, so perhaps the language server is not part of that project yet?

---

Once there _is_ a `plugins.lsp.servers.roslyn.enable` option, we'd _either_ want some kinda warning when it should be enabled but isn't, or we'd want some non-settings option that when enabled will enable the LSP automatically. Typically, we use warnings and assume the end-user is capable of configuring `plugins.lsp` themselves.

plugins/by-name/roslyn/default.nix
KennethHoff: Will do when I get internet back on my desktop machine.

Thanks!

plugins/by-name/roslyn/default.nix
KennethHoff: `roslyn-nvim` is essentially a language server with a few extra options, because Roslyn is very complex. It effectively supercedes the `csharp_ls` LSP.

The following is the summary on the Roslyn.nvim repo

```
This is an actively maintained & upgraded fork that interacts with the improved & open-source C# Roslyn language server, meant to replace the old and discontinued OmniSharp. This language server is currently used in the Visual Studio Code C# Extension, which is shipped with the standard C# Dev Kit.

This standalone plugin was necessary because Roslyn uses a non-standard method of initializing communication with the client and requires additional custom integrations, unlike typical LSP setups in Neovim.
```

I believe that this ideally  would be `lsp.servers.roslyn.enable`. Would that be possible?

plugins/by-name/roslyn/default.nix
KennethHoff: It absolutely does not make sense for this plugin to exist without the LSP. The only thing the plugin does besides adding the LSP is adding a few commands to stop, restart, and configure the LSP server.

plugins/by-name/roslyn/default.nix
MattSturgeon: Looks like nvim-lspconfig has https://github.com/razzmatazz/csharp-language-server and https://github.com/OmniSharp/omnisharp-roslyn, but not https://github.com/dotnet/vscode-csharp which the `roslyn-ls` package is based on?

MattSturgeon: > This standalone plugin was necessary because Roslyn uses a non-standard method of initializing communication with the client and requires additional custom integrations, unlike typical LSP setups in Neovim.

Ah interesting. So if I'm understanding correctly, the roslyn-nvim plugin exists as an alternative to nvim-lspconfig for setting up and configuring the language server?

I'm curious if the plugin does anything else and whether it is useful at all when the language server isn't installed?

KennethHoff: As I said [here](https://github.com/nix-community/nixvim/pull/3166#discussion_r2031991666), I believe this plugin is essentially noop without the LSP

plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  package = "roslyn-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  extraConfig = {
+    plugins.roslyn.settings.exe = lib.mkDefault "Microsoft.CodeAnalysis.LanguageServer";
+  };
+
+  # TODO: Figure out how to add this package, as it's the source of the `Microsoft.CodeAnalysis.LanguageServer` command, which is required for the LSP to function
+  # home.packages = [ pkgs.roslyn-ls ];
```
MattSturgeon, line 14: I'm unfamiliar with this LS, so I don't know if it is better to contribute to nvim-lspconfig (so that our `plugins.lsp.servers` options can be used) or to assume all configuration will be done by this plugin separately from nvim-lspconfig.

If the LS can't really be configured by nvim-lspconfig, the next best thing would be to use our new `dependencies` options.

To do this, you would edit `modules/dependencies.nix`, and add `"roslyn-ls"` to the list. Then you would edit `extraConfig` here, and define `dependencies.roslyn-ls.enable = lib.mkDefault true`.

PR #3171, opened 2025-04-11
plugins/spider: include luautf8

0x5a4:
spider [needs luautf8](https://github.com/chrisgrieser/nvim-spider?tab=readme-ov-file#utf-8-support) for UTF-8 support. this has been driving me nuts with german umlauts :smile:

### DIFF
--- a/plugins/by-name/spider/default.nix
+++ b/plugins/by-name/spider/default.nix
@@ -70,6 +70,7 @@ in
     in
     mkIf cfg.enable {
       extraPlugins = [ cfg.package ];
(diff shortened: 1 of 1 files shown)


MattSturgeon:
@GaetanLepage how/should this integrate with the new dependencies option-set?

Ideally, we'd still manage these lua deps via those options, however `extraLuaPackages` takes _functions_ accepting `luaPkgs`.

In this specific case this may be something better fixed in nixpkgs (right?), but we'll probably run into a case at some point where a lua dep should be managed in nixvim.

khaneliman (approved):

PR #3175, opened 2025-04-11
plugins/codeium-nvim: remove rename warnings, remove options, use dependencies

GaetanLepage:
- **plugins/codeium-nvim: remove rename warnings**
- **plugins/codeium-nvim: remove settings options**
- **modules/dependencies: add coreutils, gzip and util-linux**
- **plugins/codeium-nvim: use dependencies**

### DIFF
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -9,6 +9,11 @@ let
 
   packages = {
     bat.default = "bat";
+    codeium.default = "codeium";
+    coreutils = {
+      default = "coreutils";
+      example = "uutils-coreutils";
+    };
     cornelis.default = "cornelis";
     ctags.default = "ctags";
     curl.default = "curl";
@@ -29,6 +34,7 @@ let
     glow.default = "glow";
     go.default = "go";
     godot.default = "godot_4";
+    gzip.default = "gzip";
     lazygit.default = "lazygit";
     lean.default = "lean4";
     ledger.default = "ledger";
@@ -49,6 +55,7 @@ let
     tree-sitter.default = "tree-sitter";
     typst.default = "typst";
     ueberzug.default = "ueberzugpp";
+    util-linux.default = "util-linux";
     websocat.default = "websocat";
     wezterm.default = "wezterm";
     which.default = "which";
--- a/plugins/by-name/codeium-nvim/default.nix
+++ b/plugins/by-name/codeium-nvim/default.nix
@@ -16,40 +16,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   ];
 
   # TODO: added 2024-09-03 remove after 24.11
-  deprecateExtraOptions = true;
-  optionsRenamedToSettings = [
-    "configPath"
-    "binPath"
-    [
-      "api"
-      "host"
-    ]
-    [
-      "api"
-      "port"
-    ]
-    [
-      "tools"
-      "uname"
-    ]
-    [
-      "tools"
-      "uuidgen"
-    ]
-    [
-      "tools"
-      "curl"
-    ]
-    [
-      "tools"
-      "gzip"
-    ]
-    [
-      "tools"
-      "languageServer"
-    ]
-    "wrapper"
-  ];
+  inherit (import ./deprecations.nix) deprecateExtraOptions optionsRenamedToSettings;
 
   # Register nvim-cmp association
   imports = [
--- a/plugins/by-name/codeium-nvim/default.nix
+++ b/plugins/by-name/codeium-nvim/default.nix
@@ -2,9 +2,6 @@
   lib,
   ...
 }:
-let
-  inherit (lib.nixvim) defaultNullOpts;
-in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "codeium-nvim";
   packPathName = "codeium.nvim";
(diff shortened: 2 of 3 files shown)


plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +13,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-    [
-      "tools"
-      "gzip"
-    ]
-    [
-      "tools"
-      "languageServer"
-    ]
-    "wrapper"
-  ];
```
MattSturgeon, line 52: What's the motivation for removing these? The alternative of moving them to a deprecation file seems fine too, right?

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +13,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            let
+              dep = config.dependencies.${depName};
+              depPackage = dep.package;
+
+              execPath =
+                if builtins.isString execName then lib.getExe' depPackage execName else lib.getExe depPackage;
+            in
+            lib.mkIf dep.enable (lib.mkDefault execPath);
+        in
+        lib.mapAttrs exeIfEnabled dependencies;
```
MattSturgeon, line 52: This behaviour should be documented somewhere

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +13,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        coreutils = "uname";
+        util-linux = "uuidgen";
       };
-    }
-  '';
+    in
+    {
+      dependencies = lib.genAttrs (lib.attrNames dependencies) (_: {
+        enable = lib.mkDefault true;
+      });
```
MattSturgeon, line 37: `genAttrs (attrNames foo) (_: {})` is just a less efficient version of `mapAttrs (_: _: {}) foo`

modules/dependencies.nix
```diff
@@ -9,6 +9,11 @@ let
 
   packages = {
     bat.default = "bat";
+    codeium.default = "codeium";
+    coreutils = {
+      default = "coreutils";
+      example = "pkgs.uutils-coreutils";
```
MattSturgeon, line 15: Conflict with #3180:

```suggestion
      example = "uutils-coreutils";
```

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,38 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        depsCfg = config.dependencies;
+      in
+      {
+        curl = mkIf depsCfg.curl.enable (mkDefault (getExe depsCfg.curl.package));
+        gzip = mkIf depsCfg.gzip.enable (mkDefault (getExe depsCfg.gzip.package));
+        coreutils = mkIf depsCfg.coreutils.enable (mkDefault (getExe' depsCfg.coreutils.package "uname"));
+        util-linux = mkIf depsCfg.util-linux.enable (
+          mkDefault (getExe' depsCfg.util-linux.package "uuidgen")
+        );
+        language_server = mkIf depsCfg.codeium.enable (getExe depsCfg.codeium.package);
```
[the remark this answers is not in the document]
MattSturgeon: Your original impl is simpler and easier to follow, your new proposal is DRYer, but kinda hard to follow due to all the conditionals and nested let-blocks.

I'm fine with either, though.

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,38 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-    [
-      "tools"
-      "languageServer"
-    ]
-    "wrapper"
-  ];
+  description = ''
+    By default, Nixvim will install the `curl`, `gzip`, `coreutils`, `util-linux` and `codeium` packages and populate settings.tools.*` accordingly.
+
+    You are free to override `dependencies.*.enable` and `dependencies.*.package` to respectively disable and customize this behavior.
```
MattSturgeon, line 27: Wording suggestion, if you agree:
```suggestion
    You are free to configure `dependencies.*.enable` and `dependencies.*.package` to disable or customize this behavior, respectively.
```

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,38 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      "gzip"
-    ]
-    [
-      "tools"
-      "languageServer"
-    ]
-    "wrapper"
-  ];
+  description = ''
+    By default, Nixvim will install the `curl`, `gzip`, `coreutils`, `util-linux` and `codeium` packages and populate settings.tools.*` accordingly.
```
MattSturgeon, line 25: I'm struggling to come up with good wording, but I think _something_ like this is a little more explicit about what is actually happening:

```suggestion
    By default, enabling this plugin will also install the `curl`, `gzip`, `coreutils`, `util-linux` and `codeium` packages (via the `dependencies.*.enable` options`)

    Options under `settings.tools.*` with a known dependency will be configured automatically when the associated `dependencies.*.enable` option is enabled.
```

?

plugins/by-name/codeium-nvim/default.nix
MattSturgeon: > To reduce the clutter.

I believe moving the options to `./deprecation.nix` achieves this.

> I think that it is desirable for us to maintain our set of deprecations. I am in favor of cleaning the code base from (objectively?) useless warnings.

I don't necessarily disagree. I came around to the idea during our [discussion on matrix](https://matrix.to/#/!iBrKeqhFRZaBBGodLH:matrix.org/$AhfJL9W4d4HVJc7yuv16lWTd6Zo_mi05hH5T19KQKDU?via=matrix.org&via=vhack.eu&via=familleboyer.net).

But let's move this discussion to #3181.

Shall we postpone removal for now, so it doesn't block this PR?

plugins/by-name/codeium-nvim/default.nix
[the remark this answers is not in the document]
MattSturgeon: I do like that your proposed impl only defines the `dependencies` in a single place, which would prevent the two definitions from getting out of sync.

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,56 @@ lib.nixvim.plugins.mkNeovimPlugin {
     khaneliman
   ];
 
+  description = ''
+    By default, enabling this plugin will also install the `curl`, `gzip`, `coreutils`, `util-linux` and `codeium` packages (via the `dependencies.*.enable` options`)
```
MattSturgeon, line 25: My mistake: I forgot the `.`

```suggestion
    By default, enabling this plugin will also install the `curl`, `gzip`, `coreutils`, `util-linux` and `codeium` packages (via the `dependencies.*.enable` options`).
```

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,56 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-        # TODO: Added 2024-09-05; remove after 24.11
-        type = with lib.types; lib.nixvim.transitionType ints.positive toString (strMatching "[0-9]+");
-        pluginDefault = "443";
-        description = ''
-          The port of the API server to use.
-        '';
-      };
-    };
-
-    tools = {
```
MattSturgeon, line 83: I wonder if a `settings.tools` option(s) should remain, as it is a sensible place to document and/or implement the dependency-based stuff?

plugins/by-name/codeium-nvim/default.nix
```diff
@@ -15,108 +21,56 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       };
-    }
-  '';
+    in
+    {
+      dependencies = lib.mapAttrs (_: _: {
+        enable = lib.mkDefault true;
+      }) depPackages;
+
+      plugins.codeium-nvim.settings.tools = lib.mapAttrs' (
```
MattSturgeon, line 59: Nit:

Technically, because we are effectively doing `settings.tools = {}` here, we will always define the attrset, even if all deps are disabled and result in `mkIf false` attrs.

I.e. an empty attrset would be unconditionally added to the final `settings` value.

Currently this is handled by `toLuaObject` "culling" empties, however IMO it's bad practice.

---

A better approach might be to use `mkMerge`, since `mkMerge []` is effectively still undefined.

E.g.

```nix
plugins.codeium-nvim.settings.tools = lib.mkMerge (
  lib.mapAttrsToList (
    depName:
    {
      executable ? null,
      optionName ? depName,
    }:
    let
      # ...
    in
    mkIf depConfig.enable {
      ${optionName} = if executable != null then getExe' package executable else getExe package;
    }
  ) depPackages
);
```

MattSturgeon (commented):
I don't see any major blockers remaining :grin:

MattSturgeon (approved):
LGTM. @khaneliman any objections?

plugins/by-name/codeium-nvim/default.nix
khaneliman: I feel like `description` and `settingsExample` are good places to document usages without having to maintain options themselves.

khaneliman (approved):

plugins/by-name/codeium-nvim/default.nix
MattSturgeon: > Also, the packages will be in `$PATH` so setting these options is pretty much redundant anyway.

Hm. If the plugin's default behaviour is to get the exe from PATH does defining the settings options achieve anything? If there's no benefit to it, I'd rather not do it :grin:

PR #3178, opened 2025-04-11
plugins/parrot: init

### LINKED ISSUE #3137: [PLUGIN REQUEST] parrot.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `parrot.nvim`  | 
| Homepage | `https://github.com/frankroeder/parrot.nvim`     | 
| Nixpkgs  | `false` |

GaetanLepage:
Add support for [parrot.nvim](https://github.com/frankroeder/parrot.nvim), a stochastic parrot for Neovim.

Closes #3137

### DIFF
--- a/plugins/by-name/parrot/default.nix
+++ b/plugins/by-name/parrot/default.nix
@@ -1,0 +1,34 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "parrot";
+  packPathName = "parrot.nvim";
+  package = "parrot-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  extraConfig = {
+    dependencies.ripgrep.enable = lib.mkDefault true;
+  };
+
+  settingsExample = lib.literalExpression ''
+    {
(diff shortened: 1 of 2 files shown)


plugins/by-name/parrot/default.nix
```diff
@@ -0,0 +1,32 @@
...
+    };
+    hooks = {
+      Ask.__raw = ''
+        function(parrot, params)
+          local template = "Please, answer to this question: {{command}}."
+          local model_obj = parrot.get_model("command")
+          parrot.logger.info("Asking model: " .. model_obj.name)
+          parrot.Prompt(params, parrot.ui.Target.popup, model_obj, "🤖 Ask ~ ", template)
+        end
+      '';
```
MattSturgeon, line 29: FYI this will render horribly in the docs.

`renderOptionValue` uses `generators.toPretty` for non-literal values, which uses `toJSON` for string values.

Therefore you'll always get a single line string with `\n` for newlines.

I'd recommend using literalExpression when you want to manually enforce a specific formatting; in this case we want to enforce using an indented string.

MattSturgeon (approved):

PR #3180, opened 2025-04-12
modules/dependencies: refactor `all-examples` test

MattSturgeon:
Follow up to #3174

Embed the actual package as an attr in the `example`'s literal expression, so that we don't need to convert to and from human-readable attr paths in the test.

This also fixes the "foo not found in pkgs" error from incorrectly saying "pkgs.foo not found in pkgs" and looking up `pkgs.pkgs.foo` for the actual package.

Additionally, this means that weird edge-cases (like packages with non-alphabetic chars in their attr-path) will be handled correctly.

### DIFF
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -18,13 +18,13 @@ let
     flutter.default = "flutter";
     fzf = {
       default = "fzf";
-      example = "pkgs.skim";
+      example = "skim";
     };
     gcc.default = "gcc";
     gh.default = "gh";
     git = {
       default = "git";
-      example = [ "gitMinimal" ];
+      example = "gitMinimal";
     };
     glow.default = "glow";
     go.default = "go";
@@ -36,7 +36,7 @@ let
     manix.default = "manix";
     nodejs = {
       default = "nodejs";
-      example = "pkgs.nodejs_22";
+      example = "nodejs_22";
     };
     plantuml.default = "plantuml";
     ripgrep.default = "ripgrep";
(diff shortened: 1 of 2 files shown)


modules/dependencies.nix
```diff
@@ -64,7 +64,24 @@ let
...
+          let
+            path = lib.toList properties.example;
+            text = "pkgs.${lib.showAttrPath path}";
+          in
+          if properties.example._type or null == "literalExpression" then
+            properties.example
+          else
+            {
+              _type = "literalExpression";
+              inherit path text;
```
MattSturgeon, line 82: I guess we could embed the actual package instead of its path

MattSturgeon:
```
   … while evaluating attribute 'example'
     at /nix/store/5r0zii3rap2dyfyip31pdf9f0hpvv3d7-source/lib/options.nix:585:13:
      584|           // optionalAttrs (opt ? example) {
      585|             example = builtins.addErrorContext "while evaluating the example of option `${name}`" (
	 |             ^
      586|               renderOptionValue opt.example
```

For some reason `renderOptionValue opt.example` is evaluating `opt.example.package`. I had thought this would be lazy. I'll look into this a bit later.

MattSturgeon:
I understand the failure now.

`renderOptionValue` will convert non-literal expressions into literal expressions; but in our case it is a no-op as we already supply a literal expression.

The real issue is that the normalised example returned by `renderOptionValue` eventually gets passed to `builtins.toJSON`, so is evaluated deeply.

MattSturgeon:
@mergifyio queue

PR #3184, opened 2025-04-14
plugins/molten: add ipykernel python dependency

AsbjornOlling:
Without this dependency, commands like `:MoltenInit` causes the error message "[Molten] Unable to find any kernels to launch"

It seems like people (including myself) are implementing workarounds to this missing dep:

https://github.com/alisonjenkins/neovim-nix-flake/pull/51/commits/8b2b387b3d0d3fab465c81a662ed1240e99a2703

https://gitlab.com/p6s/vi/-/commit/792df862ea6f1a60683d3a870ea2593aa2533e36

Let's just include it with the default list of extra python dependencies.

### DIFF
--- a/plugins/by-name/molten/default.nix
+++ b/plugins/by-name/molten/default.nix
@@ -224,6 +224,7 @@ mkVimPlugin {
           cairosvg
           ipython
           nbformat
+          ipykernel
         ];
       defaultText = literalExpression ''
         p: with p; [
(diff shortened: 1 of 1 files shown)


MattSturgeon (approved):
Lgtm, thanks.

FYI @GaetanLepage this is another example of a function-to-deoendeicnes option that may need integrating into the new top-level `dependencies` paradigm (?)

MattSturgeon:
@mergifyio queue

PR #3185, opened 2025-04-14
lib/deprecation: implement aliases for `mkRemovedPackageOptionModule`

MattSturgeon:
Follow up to https://github.com/nix-community/nixvim/pull/3162, in particular https://github.com/nix-community/nixvim/pull/3162#discussion_r2031191608.

Allows most existing configs to continue building; defining or reading the old `plugins.*.*Package` options will now continue to work, but result in a warning.

An error is still possible when definitions cannot be merged. E.g: \
`plugins.a.fooPackage = null` and `plugins.b.fooPackage = pkgs.foo` would effectively result in `dependencies.foo.enable = mkMerge [ true false ]`.

Generally, `plugins.a.fooPackage = null` will be an alias for `dependencies.foo.enable = false` while `plugins.a.fooPackage = pkgs.foo` is an alias for `dpendencies.foo.enable = true` and `dependencies.foo.package = pkgs.foo`.

The alias is two-way; if a user _reads_ `plugins.a.fooPackage` they will actually see `if config.dependencies.foo.enable then config.dependencies.foo.package else null` (with a warning).

The alias respects override priority; if a user has `plugins.a.fooPackage = lib.mkDefault null` and `dependencies.foo.enable = true` the final value will be `true`. There may be some edge-cases I missed here, due to the alias involving three options instead of two; test case ideas welcome.

For reference, the implementation is based on [`lib.modules.doRename`](https://github.com/NixOS/nixpkgs/blob/9b78090fc935666a67fd6d2313b08bfa4ae6f8de/lib/modules.nix#L1748-L1855) _(the underlying impl used by `mkRenamedOptionModule`)_ and is conceptually similar to [`lib.modules.mkChangedOptionModule`](https://github.com/NixOS/nixpkgs/blob/9b78090fc935666a67fd6d2313b08bfa4ae6f8de/lib/modules.nix#L1649-L1689).

### DIFF
--- a/lib/deprecation.nix
+++ b/lib/deprecation.nix
@@ -129,23 +129,62 @@ rec {
       packageName,
       oldPackageName ? packageName,
     }:
-    { options, ... }:
+    { config, options, ... }:
     let
-      oldOptionPath = builtins.concatMap lib.toList [
+      cfg = config.dependencies.${packageName};
+      opt = options.dependencies.${packageName};
+      oldOptPath = builtins.concatMap lib.toList [
         "plugins"
         plugin
         "${oldPackageName}Package"
       ];
-      depOption = options.dependencies.${packageName};
-      depOptionLoc = lib.dropEnd 1 depOption.enable.loc;
+      oldOpt = lib.getAttrFromPath oldOptPath options;
 
-      instructions = ''
-        Please use the `${lib.showOption depOptionLoc}` top-level option instead:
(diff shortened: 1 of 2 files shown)


traxys (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergifyio queue

PR #3191, opened 2025-04-20
modules/dependencies: introduce top-level (internal) __depPackages option

GaetanLepage:
This adds flexibility to the `dependencies` module in prevision for future refactors.

Instead of statically declaring all the possible packages, this PR introduces an internal `__depPackages` option.
For now, it doesn't change much, but the idea is for plugin modules to "declare" their own dependencies.

### DIFF
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -5,68 +5,9 @@
   ...
 }:
 let
-  cfg = config.dependencies;
+  inherit (lib) types;
 
-  packages = {
-    bat.default = "bat";
-    codeium.default = "codeium";
-    coreutils = {
-      default = "coreutils";
-      example = "uutils-coreutils";
-    };
-    cornelis.default = "cornelis";
-    ctags.default = "ctags";
-    curl.default = "curl";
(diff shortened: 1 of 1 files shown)


modules/dependencies.nix
```diff
@@ -89,13 +30,102 @@ let
...
-  options.dependencies = lib.mapAttrs mkDependencyOption packages;
+  options = {
+    __depPackages = lib.mkOption {
+      type = types.attrsOf (
+        types.submodule {
+          options = {
+            default = lib.mkOption {
+              type = with types; either str (listOf str);
+              description = ''
+                TODO
```
khaneliman, line 41: ?

modules/dependencies.nix
```diff
@@ -89,13 +30,102 @@ let
...
+              '';
+            };
+
+            example = lib.mkOption {
+              type = with types; nullOr str;
+              description = ''
+                TODO
```
khaneliman, line 48: ?

khaneliman (approved):

PR #3193, opened 2025-04-21
modules/dependencies: minor cleanup

MattSturgeon:
Minor fixes and cleanup following #3191

- **modules/dependencies: add description to `__depPackages`**
- **modules/dependencies: add `__depPackages` example**
- **modules/dependencies: coerce `__depPackages` attr-paths to list**
- **modules/dependencies: restore literal expression example support**

cc @GaetanLepage


MattSturgeon:
@mergifyio queue

PR #3194, opened 2025-04-21
contributing: reduce emphasis on `settingsOptions`

MattSturgeon:
Seeing as we are (now) often advising contributors to remove the `settingsOptions` they have already spent time and effort writing, I think we should reduce the emphasis on this in our contributing guide & plugin template.

- Add an additional caution in `CONTRIBUTING.md#declaring-plugin-options`
- Remove `settingsOptions` from `plugins/TEMPLATE.nix`


MattSturgeon:
@mergifyio queue

PR #3196, opened 2025-04-21
contributing: add "Writing option examples" section

MattSturgeon:
We're putting more emphasis on examples than settings-options, so it may be helpful to have a section on this in our contributing guide.


MattSturgeon:
@mergifyio queue

PR #3197, opened 2025-04-21
tests: move away from string comparisons with stdenv.hostPlatform

GaetanLepage:


tests/lsp-servers.nix
```diff
@@ -36,6 +36,7 @@ let
       ...
     }:
     let
+      inherit (pkgs.stdenv) hostPlatform;
```
MattSturgeon, line 39:
```suggestion
      inherit (pkgs.stdenv) hostPlatform;

```

MattSturgeon (approved):
Kinda surprised there's only two instances, but not complaining :grin:

PR #3198, opened 2025-04-21
tests/dependencies: add disabled list to skip testing broken dependencies

GaetanLepage:


khaneliman (approved):

tests/test-sources/modules/dependencies.nix
MattSturgeon, line 1: Can we find a reasonable way to use the blacklist on both the `all` and `examples` tests?

PR #3203, opened 2025-04-22
modules/lsp: init

GaetanLepage:
Introduce a top-level `lsp` Nixvim module to wrap Neovim's `vim.lsp` API.

Example:
```nix
lsp.servers = {
  luals.enable = true;
  clangd = {
    enable = true;
    config = {
      cmd = [
        "clangd"
        "--background-index"
      ];
      root_markers = [
        "compile_commands.json"
        "compile_flags.txt"
      ];
      filetypes = [
        "c"
        "cpp"
      ];
    };
  };
};
```

References:
- https://gpanders.com/blog/whats-new-in-neovim-0-11/#lsp
- https://neovim.io/doc/user/lsp.html#lsp-quickstart

Note: this is distinct from our `plugins.lsp` module which is backed by [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig).

### DIFF
--- a/modules/lsp.nix
+++ b/modules/lsp.nix
@@ -1,0 +1,151 @@
+{
+  lib,
+  config,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+  inherit (lib.nixvim) toLuaObject;
+
+  cfg = config.lsp;
+
+  serverType = types.submodule {
+    options = {
+      enable = lib.mkEnableOption "the language server";
+
+      activate = lib.mkOption {
+        type = types.bool;
+        description = ''
+          Whether to call `vim.lsp.enable()` for this server.
+        '';
+        default = true;
+        example = false;
+      };
+
+      package = lib.mkOption {
+        type = with types; nullOr package;
+        default = null;
+        description = ''
+          Package to use for this language server.
+
+          Alternatively, the language server should be available on your `$PATH`.
+        '';
+      };
+
+      config = mkOption {
+        type = with types; attrsOf anything;
+        description = ''
+          Configurations for each language server.
+        '';
+        default = { };
+        example = {
+          cmd = [
+            "clangd"
+            "--background-index"
+          ];
+          root_markers = [
+            "compile_commands.json"
+            "compile_flags.txt"
+          ];
+          filetypes = [
+            "c"
+            "cpp"
+          ];
+        };
+      };
+    };
+  };
+in
+{
+  options.lsp = {
+    luaConfig = lib.mkOption {
+      type = lib.types.pluginLuaConfig;
+      default = { };
+      description = ''
+        Lua code configuring LSP.
+      '';
+    };
+
+    inlayHints = {
+      enable = lib.mkEnableOption "inlay hints globally";
+    };
+
+    servers = mkOption {
+      type = types.attrsOf serverType;
+      description = ''
+        LSP servers to enable and/or configure.
+
+        This option is implemented using neovim's `vim.lsp` lua API.
+        If you prefer to use [nvim-lspconfig], see [`plugins.lsp`].
+
+        [nvim-lspconfig]: https://github.com/neovim/nvim-lspconfig
+        [`plugins.lsp`]: https://nix-community.github.io/nixvim/plugins/lsp/index.html
+      '';
+      default = { };
+      example = {
+        "*".config = {
+          root_markers = [ ".git" ];
+          capabilities.textDocument.semanticTokens = {
+            multilineTokenSupport = true;
+          };
+        };
+        luals.enable = true;
+        clangd = {
+          enable = true;
+          config = {
+            cmd = [
+              "clangd"
+              "--background-index"
+            ];
+            root_markers = [
+              "compile_commands.json"
+              "compile_flags.txt"
+            ];
+            filetypes = [
+              "c"
+              "cpp"
+            ];
+          };
+        };
+      };
+    };
+  };
+
+  config =
+    let
+      enabledServers = lib.filterAttrs (_: v: v.enable) cfg.servers;
+    in
+    {
+      extraPackages = lib.pipe enabledServers [
+        builtins.attrValues
+        (builtins.catAttrs "package")
+      ];
+
+      lsp.luaConfig.content =
+        let
+          mkServerConfig =
+            name: props:
+            let
+              luaName = toLuaObject name;
+            in
+            ''
+              vim.lsp.config(${luaName}, ${toLuaObject props.config})
+            ''
+            + lib.optionalString props.activate ''
+              vim.lsp.enable(${luaName})
+            '';
+        in
+        lib.mkMerge (
+          lib.optional cfg.inlayHints.enable "vim.lsp.inlay_hint.enable(true)"
+          ++ lib.mapAttrsToList mkServerConfig enabledServers
+        );
+
+      extraConfigLua = lib.mkIf (cfg.luaConfig.content != "") ''
+        -- LSP {{{
+        do
+          ${cfg.luaConfig.content}
+        end
+        -- }}}
+      '';
+    };
+}
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -18,6 +18,7 @@
     ./highlights.nix
     ./keymaps.nix
     ./lazyload.nix
+    ./lsp.nix
     ./lua-loader.nix
     ./opts.nix
     ./output.nix
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -180,7 +180,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
           if wrappers == [ ] then s else (builtins.head wrappers) (runWrappers (builtins.tail wrappers) s);
       in
       ''
-        -- LSP {{{
+        -- nvim-lspconfig {{{
         do
           ${cfg.preConfig}
 
--- a/tests/test-sources/modules/lsp.nix
+++ b/tests/test-sources/modules/lsp.nix
@@ -1,0 +1,30 @@
+{
+  example = {
+    lsp.servers = {
+      "*".config = {
+        root_markers = [ ".git" ];
+        capabilities.textDocument.semanticTokens = {
+          multilineTokenSupport = true;
+        };
+      };
+      luals.enable = true;
+      clangd = {
+        enable = true;
+        config = {
+          cmd = [
+            "clangd"
+            "--background-index"
+          ];
+          root_markers = [
+            "compile_commands.json"
+            "compile_flags.txt"
+          ];
+          filetypes = [
+            "c"
+            "cpp"
+          ];
+        };
+      };
+    };
+  };
+}


modules/lsp.nix
```diff
@@ -0,0 +1,103 @@
...
+
+  cfg = config.lsp;
+
+  serverType = types.submodule {
+    options = {
+      enable = mkOption {
+        type = types.bool;
+        description = ''
+
+        '';
```
HeitorAugustoLN, line 17: Looks like it is missing a description

MattSturgeon:
> Note: this is distinct from our `plugins.lsp` module which is backed by [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig?rgh-link-date=2025-04-22T14%3A31%3A04Z).

Perhaps one day (when the top-level lsp module is more mature) we'll consider renaming `plugins.lsp` to something like `plugins.nvim-lspconfig` for clarity and to de-emphasize it?

modules/lsp.nix
```diff
@@ -0,0 +1,96 @@
...
+  inherit (lib) mkOption types;
+
+  cfg = config.lsp;
+
+  serverType = types.submodule {
+    options = {
+      enable = lib.mkEnableOption "the language server";
+
+      config = mkOption {
+        type = with types; attrsOf anything;
```
MattSturgeon, line 16: Perhaps going forward we should consider using `(pkgs.formats.lua { }).type`?

This is a bit more restrictive than `anything`: https://github.com/NixOS/nixpkgs/blob/55024d64cdf0cd2a0a2764857962a1da28a6ee23/pkgs/pkgs-lib/formats.nix#L722-L740

```nix
nullOr (oneOf [
  bool
  float
  int
  path
  str
  luaInline
  (attrsOf valueType)
  (listOf valueType)
])
```

That said, it only supports our non-standard "raw lua" by accident; `attrsOf str` _happens_ to match `{ __raw = ""; }`...

We should probably have this discussion outside of this PR, as using `attrsOf anything` is consistent with our current approach throughout nixvim.

modules/lsp.nix
```diff
@@ -0,0 +1,96 @@
...
+            "cpp"
+          ];
+        };
+      };
+    };
+  };
+in
+{
+  options.lsp = {
+    servers = mkOption {
```
MattSturgeon, line 41: Q: what other options do you see being introduced in the `lsp.*` namespace?

Would it ever make sense to rename `lsp.servers` to simply `lsp`?

HeitorAugustoLN: inlayHints, keymaps, etc. could be in there too

modules/lsp.nix
```diff
@@ -0,0 +1,96 @@
...
+      };
+    };
+  };
+
+  config = {
+    extraConfigLua =
+      let
+        serversToEnable = lib.pipe cfg.servers [
+          (lib.filterAttrs (_: props: props.enable))
+          (lib.mapAttrsToList (serverName: _: serverName))
```
MattSturgeon, line 81:
```suggestion
          builtins.attrNames
```

MattSturgeon (commented):
Looks good. Minor nits:

MattSturgeon:
@mergifyio rebase

khaneliman:
We'll want to update the warning in `plugins.clangd-extensions` to know about this module.

modules/lsp.nix
```diff
@@ -0,0 +1,96 @@
...
+        ];
+      in
+      ''
+        do
+          local __nixvimLspServersConfigs = ${lib.nixvim.toLuaObject cfg.servers}
+
+          for server_name, server_props in ipairs(__nixvimLspServersConfigs) do
```
khaneliman, line 88:
```suggestion
          for server_name, server_props in pairs(__nixvimLspServersConfigs) do
```
this fixes the config being passed for me

khaneliman: But seems to barf when i don't have a config... lol

modules/lsp.nix
```diff
@@ -0,0 +1,96 @@
...
+            vim.lsp.config[server_name] = server_props.config
+          end
```
khaneliman, line 90:
```suggestion
                    for server_name, server_props in pairs(__nixvimLspServersConfigs) do
            if server_props.config then
              vim.lsp.config[server_name] = server_props.config
            end
          end
```

MattSturgeon: Alternatively, we could filter out empty configs when writing `__nixvimLspServersConfigs`?

Might be slightly more efficient by moving the logic from runtime to buildtime.

MattSturgeon:
```suggestion
          for server_name, server_props in ipairs(__nixvimLspServersConfigs) do
            vim.lsp.config(server_name, server_props.config)
          end
```

IMO `__nixvimLspServersConfigs` should also be `mapAttrs (_: props: props.config)` so that we don't end up needlessly serialising the other props (like `enable`).

I also wonder if it makes sense that we are setting the config for _all_ defined servers, even ones with `enable = false`?

I know we're using `enable` later when deciding which servers to, well, _enable_; I just wonder if it's confusing that disabled servers still get configured?

modules/lsp.nix
```diff
@@ -0,0 +1,102 @@
...
+        do
+          local __nixvimLspServersConfigs = ${lib.nixvim.toLuaObject cfg.servers}
+
+          for server_name, server_props in ipairs(__nixvimLspServersConfigs) do
+            vim.lsp.config(server_name, server_props.config)
+          end
+
+          vim.lsp.enable(${lib.nixvim.toLuaObject serversToEnable})
+        end
+      '';
```
MattSturgeon, line 100: Since we're serialising the whole of `cfg.servers` anyway, we may as well move the `enable` into the loop too:

```suggestion
    extraConfigLua = ''
      do
        local __nixvimLspServers = ${lib.nixvim.toLuaObject cfg.servers}

        for server_name, server_props in ipairs(__nixvimLspServers) do
          vim.lsp.config(server_name, server_props.config)

          if server_props.enable then
            vim.lsp.enable(server_name)
          end
        end

      end
    '';
```

modules/lsp.nix
MattSturgeon: > > IMO `__nixvimLspServersConfigs` should also be `mapAttrs (_: props: props.config)`, so that we don't end up needlessly serialising the unrelated non-config props (like `enable`).
> 
> But, how do you get `server_name` then?

The name would be unchanged (that's the `_` in the `mapAttrs`). However I personally think https://github.com/nix-community/nixvim/pull/3203#discussion_r2055109527 is a better idea than this one.

> The user could want to enable the server somewhere else (in an autocmd for instance) but still have it configured through `lsp.servers.foo.config`.
> As the API itself differentiates configuring and enabling servers, I think that our wrapping should do the same.

Yeah, I was on the fence regarding the same argument. IMO this could be a justification for a global `lsp.enable` option, even if it is default-true, just to allow users to disable (e.g. conditionally) us configuring stuff for them

modules/lsp.nix
```diff
@@ -0,0 +1,162 @@
...
+      type = types.lines;
+      description = "Code to be run before configuring the LSP. Useful for requiring plugins";
+      default = "";
+    };
+
+    postConfig = mkOption {
+      type = types.lines;
+      description = "Code to be run after configuring the LSP. Useful for requiring plugins";
+      default = "";
+    };
```
MattSturgeon, line 71: Let's use a `pluginLuaConfig` type?

modules/lsp.nix
```diff
@@ -0,0 +1,159 @@
...
+      enable = lib.mkEnableOption "inlay hints globally";
+    };
+
+    servers = mkOption {
+      type = types.attrsOf serverType;
+      description = ''
+        LSP servers to enable and/or configure.
+
+        This option is implemented using neovim's `vim.lsp` lua API.
+        If you prefer to use [nvim-lspconfig], see [`plugins.lsp`].
```
MattSturgeon, line 79: At some point, we should tweak this wording to suggest users use _both_ `lsp` and `plugins.lsp` together.

modules/lsp.nix
```diff
@@ -0,0 +1,159 @@
...
+            ];
+          };
+        };
+      };
+    };
+  };
+
+  config =
+    let
+      enabledServers = lib.filterAttrs (_: v: v.enable) cfg.servers;
```
MattSturgeon, line 116: If we introduce a `name` option (similar to other submodule sets like `extraFiles`), then we can simplify this to `filter (v: v.enable) (attrValues cfg.servers)`

```nix
submodule (
  { name, ... }:
  {
    options = {
      name = lib.mkOption {
        type = lib.types.str;
        description = "The name of the language server, used as the first argument to `vim.lsp.enable` and `vim.lsp.config`.";
        default = name;
      };
      # ...
    };
  }
)
```

This would also allow users to configure the same server in different ways, using different attr-names, and decide which to use via the `enable` option.

modules/lsp.nix
```diff
@@ -0,0 +1,159 @@
...
+        lib.concatStringsSep "\n" (
+          [
+            cfg.preConfig
+          ]
+          ++ lib.optional cfg.inlay_hint.enable "vim.lsp.inlay_hint.enable(true)"
+          ++ [
+            serverConfig
+            cfg.postConfig
+          ]
+        );
```
MattSturgeon, line 149:
```suggestion
        in
        lib.mkMerge (
          lib.optional cfg.inlay_hint.enable "vim.lsp.inlay_hint.enable(true)"
          ++ lib.mapAttrsToList mkServerConfig enabledServers
        );
```

modules/lsp.nix
```diff
@@ -0,0 +1,159 @@
...
+  options.lsp = {
+    luaConfig = lib.mkOption {
+      type = lib.types.pluginLuaConfig;
+      default = { };
+      description = ''
+        Lua code configuring LSP.
+      '';
+    };
+
+    inlay_hint = {
```
HeitorAugustoLN, line 69: Why this doesn't follow the regular camelCase naming?

MattSturgeon: @GaetanLepage's argument was that it is directly representing the underlying `vim.lsp.inlay_hint` API.

Personally, I don't feel the representation is so "direct", so I'd also be in favour of:

```suggestion
    inlayHints = {
```

zowoq:
Any ideas why this PR has been causing multi hour darwin CI jobs on buildbot?

zowoq:
If the jobs aren't important I'd appreciate it if you killed the builds immediately instead of leaving it to me to notice.

Darwin CI usage by this repo isn't a problem yet but it is approaching that point.

Unfortunately improving CI scheduling isn't simple and may not happen anytime soon.

Disabling x86_64-darwin and only building aarch64-darwin is probably going to be the most workable option.

MattSturgeon:
> LSP is pretty central to Nixvim, so changing it causes many rebuilds in our CI test suite.

That really shouldn't be the case; this PR doesn't really modify anything used elsewhere in nixvim. It introduces a new module that is (as of yet) unused.

The only impact it should have is adding _one_ derivation to the test suite (the `example` test). Even this test doesn't install any LSP server packages, or any vim plugins.

Further, most of the CI runs that were problematic were failing with fairly simple eval errors (e.g., typos in function names).

I really don't see any reason why any builds would've continued beyond a few minutes 🤔

> If the jobs aren't important I'd appreciate it if you killed the builds immediately instead of leaving it to me to notice.

Apologies; we usually try to cancel jobs whenever they aren't important. I guess we were distracted this time.

> Darwin CI usage by this repo isn't a problem yet but it is approaching that point.

We appreciate being able to test our plugins on all platforms. 

However there are a handful of tests that aren't affected by platform; they are still declared via flake.parts `perSystem` due to the nature of how flake outputs work, and to allow them to be run locally on any system.

If we had a way to configure which tests are run by CI (other than simply omitting those tests from the `checks` flake output) we would certainly make use of it.

mathjiajia:
```
error: attribute 'preConfig' missing
at /nix/store/58faqxc9sbjxqq0vwpf9mlax6nhnf0gx-source/modules/lsp.nix:142:13:
    141|           [
    142|             cfg.preConfig
       |             ^
    143|           ]
```

zowoq:
> If we had a way to configure which tests are run by CI (other than simply omitting those tests from the `checks` flake output) we would certainly make use of it.

Unfortunately I think this project is the only one that would really benefit from this at the moment, same applies to the CI scheduling improvements.

modules/lsp.nix
```diff
@@ -0,0 +1,151 @@
...
+          ++ lib.mapAttrsToList mkServerConfig enabledServers
+        );
+
+      extraConfigLua = ''
+        -- LSP {{{
+        do
+          ${cfg.luaConfig.content}
+        end
+        -- }}}
+      '';
```
MattSturgeon, line 149: This unconditional definition is triggering this test: https://github.com/nix-community/nixvim/blob/6415ae4a9779286b4248a2dd2ed742d96f74bbbf/tests/test-sources/modules/output.nix#L81-L101

The test exists to ensure we don't add things to init.lua unless the user has actually configured something.

We want to add a `mkIf` here, not sure what the best condition should be... maybe `enabledServers != [ ]` or `luaConfig.content != ""`?

```suggestion
      extraConfigLua = lib.mkIf (cfg.luaConfig.content != "") ''
        -- LSP {{{
        do
          ${cfg.luaConfig.content}
        end
        -- }}}
      '';
```

PR #3204, opened 2025-04-22
plugins/lsp: use vim.lsp native API

### LINKED ISSUE #3201: [BUG] Use nvim 0.11+ LSP API
| Plugin       | `lsp`  | 
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

## Description

Since Neovim 0.11, there are new APIs for configuring LSP servers via `vim.lsp.config` and `vim.lsp.enable`.
As a consequence nvim-lspconfig have deprecated the old way of `require('lspconfig').server.setup()`.
Relevant issue: https://github.com/neovim/nvim-lspconfig/issues/3494.

We need to migrate our calls to using `vim.lsp.config` and `vim.lsp.enable`.

Currently we are still using the old APIs here:
https://github.com/nix-community/nixvim/blob/d4a0db2103d52b3b8e3ae502497ef9f19b048508/plugins/lsp/default.nix#L247-L261

```nix
programs.nixvim = {
  plugins.lsp.enable = true;
}
```

GaetanLepage:
Use Neovim's native `vim.lsp` API instead of legacy `require('lspconfig').*.setup()`.

[nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) aims to become a bank of configuration for popular language servers.
Its API has been deprecated in favor of Neovim 0.11's new `vim.lsp` native one.

Hence, we need to adapt how we configure `nvim-lspconfig` in lua and migrate it to the new API.

I have not yet really understand how `nvim-lspconfig` interacts with the native API to inject its configs.

> [!NOTE]
> This is a parallel effort of #3203
> Ultimately, we should be able to use our top-level `lsp` module in `plugins.lsp`'s (`nvim-lspconfig`) implementation.

## Open questions
- Quid of `plugins.lsp.servers.*.capabilities`? Do we need to provide them to `vim.lsp.config`?

## Resources
- https://gpanders.com/blog/whats-new-in-neovim-0-11/#lsp
- https://neovim.io/doc/user/lsp.html#lsp-quickstart
- https://github.com/neovim/nvim-lspconfig/issues/3494
- https://github.com/neovim/nvim-lspconfig

Fixes #3201

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -183,6 +183,12 @@ lib.nixvim.plugins.mkNeovimPlugin {
       ++ mkMaps "vim.lsp.buf." "Lsp buf" cfg.keymaps.lspBuf
       ++ cfg.keymaps.extra;
 
+    # Since https://github.com/nix-community/nixvim/pull/3204, we are now using the native vim.lsp
+    # API for configuring language servers with nvim-lspconfig.
+    # For some mysterious reason, `performance.combinePlugins` now prevent language servers from
+    # being properly configured (missing some keys: `cmd`, `filetypes`, `root_markers` etc.)
+    performance.combinePlugins.standalonePlugins = [ cfg.package ];
+
     plugins.lsp.onAttach = lib.mkIf cfg.inlayHints ''
       -- LSP Inlay Hints {{{
       if client.server_capabilities.inlayHintProvider and vim.lsp.inlay_hint then
@@ -242,15 +248,14 @@ lib.nixvim.plugins.mkNeovimPlugin {
           }"}
 
           for i, server in ipairs(__lspServers) do
-            local options = ${runWrappers cfg.setupWrappers "server.extraOptions"}
 
-            if options == nil then
-              options = __setup
-            else
-              options = vim.tbl_extend("keep", options, __setup)
+            vim.lsp.enable(server.name)
+
+            vim.lsp.config(server.name, __setup)
+
+            if server.extraOptions then
+              vim.lsp.config(server.name, server.extraOptions)
             end
-
-            require("lspconfig")[server.name].setup(options)
           end
 
           ${cfg.postConfig}


MattSturgeon:
> I have not yet really understand how `nvim-lspconfig` interacts with the native API to inject its configs.

I suspect they are using the new rtp `lsp` directory to merge in their configs.

Any file in `<rtp>/lsp/<foo>.lua` will be merged with `vim.lsp.config('foo')`. See [`:h lsp-config`](https://neovim.io/doc/user/lsp.html#_config)

khaneliman:
Seems to work fine on my end.

plugins/lsp/default.nix
```diff
@@ -246,7 +246,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
           for i, server in ipairs(__lspServers) do
             if type(server) == "string" then
-              require("lspconfig")[server].setup(__setup)
+              vim.lsp.config(server.name, __setup)
```
MattSturgeon, line 249: In this branch `server` is a string:

```suggestion
              vim.lsp.config(server, __setup)
```

This error didn't trip CI, though, so I wonder if we're missing a test-case for that?

plugins/lsp/default.nix
```diff
@@ -256,8 +256,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
                 options = vim.tbl_extend("keep", options, __setup)
```
MattSturgeon, line 256: I believe we can simplify this slightly, since multiple calls to `vim.lsp.config()` will use a table deep extend internally we can probably just call `vim.lsp.config()` multiple times.

MattSturgeon: I think we can simplify to something like:

```lua
local name = server
local options = nil
if type(name) != 'string' then
  name = server.name
  options = server.extraOptions
end

vim.lsp.enable(name)
vim.lsp.config(name, __setup)
if options != nil then
  vim.lsp.config(name, options)
end
```

MattSturgeon:
> ## Open questions
> * Quid of `plugins.lsp.servers.*.capabilities`? Do we need to provide them to `vim.lsp.config`?

Looking at [`:h lsp-attach`](https://neovim.io/doc/user/lsp.html#lsp-attach), I believe this API has not changed:

> To use LSP features beyond those provided by Nvim (see [`:h lsp-buf`](https://neovim.io/doc/user/lsp.html#lsp-buf)), you can set keymaps and options on [`Client:on_attach()`](https://neovim.io/doc/user/lsp.html#Client%3Aon_attach()) or [`LspAttach`](https://neovim.io/doc/user/lsp.html#LspAttach).
> Not all language servers provide the same capabilities; check `supports_method()` in your `LspAttach` handler.
> Example: Enable auto-completion and auto-formatting ("linting"):

?

plugins/lsp/default.nix
```diff
@@ -245,18 +245,17 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            local options = nil
+            if type(name) ~= 'string' then
+              name = server.name
+              options = server.extraOptions
+            end
 
-              require("lspconfig")[server.name].setup(options)
+            vim.lsp.enable(name)
+            vim.lsp.config(name, __setup)
+            if options ~= nil then
```
MattSturgeon, line 257: Ah my bad with `!=` vs `~=` (always trips me up).

In lua can we simplify to `if options then` or do we need to be explicit about which falsy cases we want to handle?

plugins/lsp/default.nix
```diff
@@ -245,18 +245,17 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-                options = __setup
-              else
-                options = vim.tbl_extend("keep", options, __setup)
-              end
+            local name = server
+            local options = nil
+            if type(name) ~= 'string' then
+              name = server.name
+              options = server.extraOptions
+            end
```
MattSturgeon, line 253: Nit: the earlier error with `server` being a string was not caught by the test suite. Is it possible to add a test-case for it being defined as a string?

Or maybe the option-type doesn't actually support it being a string, in which case this impl can be simplified?

MattSturgeon (commented):
The diff looks good.

One question and one nit below, but no blockers if CI passes.

I'll briefly test my config against this branch and report back

MattSturgeon (approved):
Tested my config with nixd, seems to be working. Above nits still relevant if you're interested in them.

plugins/lsp/default.nix
```diff
@@ -242,15 +248,14 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-              options = vim.tbl_extend("keep", options, __setup)
-            end
+            vim.lsp.enable(server.name)
 
-            require("lspconfig")[server.name].setup(options)
+            vim.lsp.config(server.name, __setup)
+
+            if options then
+              vim.lsp.config(server.name, server.extraOptions)
+            end
```
MattSturgeon, line 258: This looks wrong. You're checking `options` but using `server.extraOptions`.

I think we want:

```suggestion
            if server.extraOptions then
              vim.lsp.config(server.name, server.extraOptions)
            end
```

?

MattSturgeon (changes requested):

MattSturgeon (approved):

khaneliman (approved):
Still works here in latest setup

MattSturgeon:
@GaetanLepage I think we may have to revert this for a while, as not all nvim-lspconfig configs are updated to work with the new API.

Our users are running into issues like:

- https://github.com/nix-community/nixvim/issues/3217#issuecomment-2832906966
- https://github.com/nix-community/nixvim/issues/3224

Which may not be fully resolvable until https://github.com/neovim/nvim-lspconfig/issues/3705 is complete and propagated to nixpkgs.

jfly:
> Which may not be fully resolvable until https://github.com/neovim/nvim-lspconfig/issues/3705 is complete and propagated to nixpkgs.

I haven't looked at the relevant code at all, but would it be possible to have a knob for "use vim.lsp native API" that for now defaults to false? I've submitted a number of bugfixes recently (https://github.com/neovim/neovim/pull/33707, https://github.com/neovim/neovim/pull/33703, https://github.com/neovim/neovim/pull/33702) for the native API that I'd like to test out/benefit from.

khaneliman:
> > Which may not be fully resolvable until [neovim/nvim-lspconfig#3705](https://github.com/neovim/nvim-lspconfig/issues/3705) is complete and propagated to nixpkgs.
> 
> I haven't looked at the relevant code at all, but would it be possible to have a knob for "use vim.lsp native API" that for now defaults to false? I've submitted a number of bugfixes recently ([neovim/neovim#33707](https://github.com/neovim/neovim/pull/33707), [neovim/neovim#33703](https://github.com/neovim/neovim/pull/33703), [neovim/neovim#33702](https://github.com/neovim/neovim/pull/33702)) for the native API that I'd like to test out/benefit from.

This Pr was reverted, lspconfig doesn't use it again.

MattSturgeon:
> I've submitted a number of bugfixes recently ([neovim/neovim#33707](https://github.com/neovim/neovim/pull/33707), [neovim/neovim#33703](https://github.com/neovim/neovim/pull/33703), [neovim/neovim#33702](https://github.com/neovim/neovim/pull/33702)) for the native API that I'd like to test out/benefit from.

Thanks! :heart_eyes: 

> would it be possible to have a knob for "use vim.lsp native API" that for now defaults to false?

We kinda have that already; we're developing a new way of managing LSP in parallel to the old way staying mostly as-is for now.

The new design consists of `plugins.lspconfig` for installing nvim-lspconfg; this is basically just an `enable` option and a `package` option. Along with a top-level `lsp` module for managing the new `vim.lsp.*` lua API.

We're still calling the top-level `lsp` module experimental; e.g. [some options may be renamed](https://github.com/nix-community/nixvim/issues/3265) (feedback wanted there or on matrix!), but it should be working fairly well :crossed_fingers: 

See:

- [`lsp.servers`](https://nix-community.github.io/nixvim/lsp/servers/index.html)
- [`lsp.keymaps`](https://nix-community.github.io/nixvim/lsp/keymaps/index.html)
- [`lsp.inlayHints`](https://nix-community.github.io/nixvim/lsp/inlayHints.html)
- `lsp.onAttach` PR: https://github.com/nix-community/nixvim/pull/3280
- [`plugins.lspconfig`](https://nix-community.github.io/nixvim/plugins/lspconfig/index.html)

PR #3205, opened 2025-04-23
plugins/lsp: remove string support for plugins.lsp.enabledServers elements

GaetanLepage:


MattSturgeon (approved):
I'm not a _huge_ fan of dropping support for something without knowing how many users we're affecting, how helpful the error message will be, etc, etc.

I _believe_ users will get a fairly generic error that ``some definitions for `programs.nixvim.plugins.lsp.enabledServers` do not match type "listOf submodule"``

It is an `internal` option, though, which makes me feel better. On that basis I'll approve.

Issue #3208, opened 2025-04-24
How to merge unkeyed attrs?


jfly:
I'm curious how to merge multiple `listToUnkeyedAttrs`. For example:

```console
./ $ nix repl github:nix-community/nixvim
...
nix-repl> builtins.trace (lib.nixvim.toLuaObject ((lib.nixvim.listToUnkeyedAttrs ["a" "b"]) // (lib.nixvim.listToUnkeyedAttrs ["c" "d"]) // {foo = "bar";})) null
trace: { "c", "d", foo = "bar" }
null
```

Ideally, I would want the resulting lua table to be:

```lua
{ "a", "b", "c", "d", foo = "bar" }
```

It feels like this is a fundamental issue with how `listToUnkeyedAttrs` represents unkeyed attrs:

```console
./ $ nix repl github:nix-community/nixvim
nix-repl> lib.nixvim.listToUnkeyedAttrs ["a" "b"]
{
  __unkeyed-0 = "a";
  __unkeyed-1 = "b";
}
```

Have we considered one of the following?

- Reworking `lib.nixvim.listToUnkeyedAttrs` to represent unkeyed attrs in a different way? For example, rather than injecting integer indices into the keys, we could inject the value.
- Create a custom merge function that knows to treat these `--unkeyed-<N>` attrs specially.

MattSturgeon:
Yes, this is a limitation of the current impl.

Currently, unkeyed attrs are merged in the same way as any other attr-set, so `__unkeyed-0` from one definition will try to merge with `__unkeyed-0` from another definition.

In the vast majority of nixvim options, lua options are typed as `anything` or `attrsOf anything`. This means we don't have an obvious way to implement custom merge logic.

The two ways we could go about it are:

1. replace usage of `types.anything` treewide
2. replace `types.anything` with a custom impl in nixvim's "extended lib"

[...]

PR #3213, opened 2025-04-24
lib/lua-types: init

MattSturgeon:
The types in `lua-types` are inspired by `pkgs.formats`, allowing for a more appropriate option type than `types.anything`.

The intention is that we will replace usage of `types.anything` in 90% of cases treewide with `lua-types.anything` (see #3208). Currently there are ~270 instances of `anything` treewide.

All types in `lua-types` support inline-lua and raw-lua.

`lua-types.tableOf` will also serve as a good place to implement merging lists defs and attrs defs into a mixed table, at some point in the future. Ideally we can upstream something similar to nixpkgs too. Again, see #3208 and also https://github.com/NixOS/nixpkgs/compare/master...MattSturgeon:nixpkgs:table.

## TODO

- [ ] Move `types.rawLua` to `lua-types`
- [ ] Use `lua-types.anything` in `mkNvimPkugin`'s `settings` option
- [ ] Use `lua-types` in `defaultNullOpts` (?)

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -21,6 +21,7 @@ lib.makeExtensible (
     deprecation = call ./deprecation.nix { };
     keymaps = call ./keymap-helpers.nix { };
     lua = call ./to-lua.nix { };
+    lua-types = call ./lua-types.nix { };
     modules = call ./modules.nix { inherit flake; };
     options = call ./options.nix { };
     plugins = call ./plugins { };
--- a/lib/lua-types.nix
+++ b/lib/lua-types.nix
@@ -1,0 +1,66 @@
+{ lib }:
+let
+  inherit (lib) types;
+  inherit (lib.nixvim) lua-types;
+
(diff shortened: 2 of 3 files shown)


MattSturgeon:
> Why not relying on `lib.fix`? Is this approach (importing it back in and changing it in place) preferable? Would `lib.fix` (or `rec`) work at all?

I assume you mean the self-referencing recursive `lua-types.anything`?

As far as recursive laziness goes, both approaches are exactly the same. No matter how you write it, all self-referencing values do the same thing...

I originally wrote it with "local" recursion using `lib.fix`, but I found that clunkier due to the extra syntax.

```nix
lib.fix (self: types.oneOf [
  # ...
  (tableOf self)
])
// {
  description = "lua value";
  # ...
}
```

Then I realised that was unnecessary; we don't need `lib.fix` to get a `self` reference because we already have a `lua-types` reference in scope.

PR #3225, opened 2025-04-26
plugins/lsp: simplify implementation of inlayHints

GaetanLepage:
@MattSturgeon and I have observed that enabling inlay hints globally has the same effect as our current implementation, which enabled it locally for every server that supports it.

Simply adding `vim.lsp.inlay_hint.enable(true)` in the config file will enable inlay hints for all servers that support them.

### DIFF
--- a/plugins/lsp/default.nix
+++ b/plugins/lsp/default.nix
@@ -113,8 +113,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
       type = types.bool;
       default = false;
       description = ''
-        Whether to enable LSP inlay-hints.
-        Only affects language servers with inlay-hints support.
+        Whether to enable LSP inlay-hints globally.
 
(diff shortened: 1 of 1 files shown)


plugins/lsp/default.nix
```diff
@@ -113,8 +113,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
       type = types.bool;
       default = false;
       description = ''
-        Whether to enable LSP inlay-hints.
-        Only affects language servers with inlay-hints support.
+        Whether to enable LSP inlay-hints globally.
```
MattSturgeon, line 116: This may be a good place to document how to do this per-lsp and/or per-buffer:

`````md
	To use inlay hints only on specific servers, you should call `vim.lsp.inlay_hint.enable` in your `on_attach` function:

	```lua
	on_attach = function(client, bufnr)
	  vim.lsp.inlay_hint.enable(true, { bufnr = bufnr })
	end
```
`````

MattSturgeon (approved):

PR #3099, opened 2025-03-24
modules/performance/combinePlugins: propagate dependencies

MattSturgeon:
An attempt at having `combinePlugins` correctly propagate dependencies from the input plugins.

Marking as a draft for now until we can verify it works correctly, but otherwise ready for review.

PR #3232, opened 2025-04-27
modules/performance/combinePlugins: propagate lua dependencies

### LINKED ISSUE #3140: [BUG] removal of magick from image plugin breaks my neovim config with nixvim
| Plugin       | `image`  |
| Nixpkgs      |`unstable` | 
| Home Manager | `unstable` | 

- [ ] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

after recent update I got neovim complaining about `magick` missing:

```
im.lsp.get_active_clients() is deprecated. Run ":checkhealth vim.depre            
image.nvim: magick rock not found, please install it and restart your editor. Error
: "module 'magick' not found:\n\tno field package.preload['magick']\n\tcache_loader
: module 'magick' not found\n\tcache_loader_lib: module 'magick' not found\n\tno fi
le '/nix/store/q1lf0jbc4imxr2yz4m4xlf3vq7h0qhn0-luajit2.1-lua-utils.nvim-1.0.2-1/sh
are/lua/5.1/magick.lua'\n\tno file '/nix/store/q1lf0jbc4imxr2yz4m4xlf3vq7h0qhn0-lua
jit2.1-lua-utils.nvim-1.0.2-1/share/lua/5.1/magick/init.lua'\n\tno file '/nix/store
/714niwwfbij44f1h05x1w9hxcidwwihp-luajit-2.1.1741730670/share/lua/5.1/magick.lua'\n
\tno file '/nix/store/ws1yi3vfk9ylcps43h6i6j82mpa4z3q4-luajit2.1-nui.nvim-0.3.0-1/s
hare/lua/5.1/magick.lua'\n\tno file '/nix/store/ws1yi3vfk9ylcps43h6i6j82mpa4z3q4-lu
ajit2.1-nui.nvim-0.3.0-1/share/lua/5.1/magick/init.lua'\n\tno file '/nix/store/5zfy
6zfxqrxglwm6b2rz77bb6b8xs9vm-luajit2.1-nvim-nio-1.10.1-1/share/lua/5.1/magick.lua'\
n\tno file '/nix/store/5zfy6zfxqrxglwm6b2rz77bb6b8xs9vm-luajit2.1-nvim-nio-1.10.1-1
/share/lua/5.1/magick/init.lua'\n\tno file '/nix/store/zvikc9yvn9g9c4n68zfhvszvydbp
yzmm-luajit2.1-pathlib.nvim-2.2.3-1/share/lua/5.1/magick.lua'\n\tno file '/nix/stor
e/zvikc9yvn9g9c4n68zfhvszvydbpyzmm-luajit2.1-pathlib.nvim-2.2.3-1/share/lua/5.1/mag
ick/init.lua'\n\tno file '/nix/store/cbl8xqzl4sa19q3wf0k2l7mqf83y0jm4-luajit2.1-ple
nary.nvim-scm-1/share/lua/5.1/magick.lua'\n\tno file '/nix/store/cbl8xqzl4sa19q3wf0
k2l7mqf83y0jm4-luajit2.1-plenary.nvim-scm-1/share/lua/5.1/magick/init.lua'\n\tno fi
le '/nix/store/s6m053jn5s3pnd5gfahyplrb28rswfsa-luajit2.1-luassert-1.9.0-1/share/lu
a/5.1/magick.lua'\n\tno file '/nix/store/s6m053jn5s3pnd5gfahyplrb28rswfsa-luajit2.1
-luassert-1.9.0-1/share/lua/5.1/magick/init.lua'\n\tno file '/nix/store/8ihhpaaq1qg
v5hakhqzx1hmvsx5gxj32-luajit2.1-say-1.4.1-3/share/lua/5.1/magick.lua'\n\tno file '/
nix/store/8ihhpaaq1qgv5hakhqzx1hmvsx5gxj32-luajit2.1-say-1.4.1-3/share/lua/5.1/magi
ck/init.lua'\n\tno file '/nix/store/76sknx5rv9pg0zjwzffmahkc758clkwj-luajit-2.1.174
1730670-env/share/lua/5.1/magick.lua'\n\tno file '/nix/store/76sknx5rv9pg0zjwzffmah
kc758clkwj-luajit-2.1.1741730670-env/share/lua/5.1/magick/init.lua'\n\tno file '/ni
x/store/76sknx5rv9pg0zjwzffmahkc758clkwj-luajit-2.1.1741730670-env/lib/lua/5.1/magi
ck.so'\n\tno file '/nix/store/kl1zmhd8hgxkaz6xyv0nim0baky0ax6b-vim-pack-dir/pack/my
NeovimPackages/start/vimplugin-plugin-pack/lua/../build/magick.so'"  ```

digging into issues I found that 

```
extraLuaPackages = ps: [ ps.magick ];
```

had been removed from the source of the image plugin. If I put it back in my config this solves my error. 

I hoped that  https://github.com/NixOS/nixpkgs/pull/389851 would fix it -- should be in unstable by now -- but I still get this error.

```nix
{ inputs, lib, ... }:
{
  imports = [
    inputs.nixvim.homeManagerModules.nixvim
    ];
  programs.nixvim = {

    plugins = {
      diagram.enable = true;
      image = {
        enable = true;
        settings = {
          integrations = {
            neorg.enabled = true;
            markdown.enabled = true;
          };
          backend = "kitty";
        };
      };
  };
    # WIP
    #extraLuaPackages = ps: [ ps.magick ]; #thats the line...
    # END WIP

    enable = true;
    defaultEditor = true;

    performance = {
      combinePlugins = {
        enable = true;
      };
      byteCompileLua.enable = true;
    };

    viAlias = true;
    vimAlias = true;

    luaLoader.enable = true;
  };
}


```

stasjok:
Alternative to #3099.

Fixes #3140
Closes #3099

Plugins from luarocks (e.g. telescope-nvim) have dependencies specified in propagatedBuildInputs. These dependencies are not added as plugins in Nvim runtime. They are added to LUA_PATH env var for wrapped neovim. This PR collects all propagatedBuildInputs from input plugin list and puts them in the combined plugin.

Note that such dependencies are never combined or byte-compiled, because they are not plugins. Also it can lead to double plugins (e.g. one plenary-nvim as plugin dependency and another one in LUA_PATH). This is not nixvim issue, this is how it works in nixpkgs at this moment.

Most of the work for this PR was done to tests. Because of the changes in nixpkgs to plugin dependencies (especially to plenary.nvim) many tests weren't doing what was intended. To avoid such cases in the future, I'm not using plugins from nixpkgs anymore. I created dummy plugins with dependencies that should be tested (combinePlugins only for now). I hope this will avoid some of the unexpected breakages.

### DIFF
--- a/modules/top-level/plugins/mk-plugin-pack.nix
+++ b/modules/top-level/plugins/mk-plugin-pack.nix
@@ -28,24 +28,38 @@ let
       (builtins.concatMap (f: f ps))
     ];
 
-  # Combined plugin
-  combinedPlugin = pkgs.vimUtils.toVimPlugin (
-    pkgs.buildEnv {
-      name = "plugin-pack";
-      paths = overriddenPlugins;
-      inherit pathsToLink;
+  # propagatedBuildInputs contain lua dependencies
+  propagatedBuildInputs = lib.pipe pluginsToCombine [
+    (builtins.catAttrs "plugin")
+    (builtins.catAttrs "propagatedBuildInputs")
+    builtins.concatLists
+    lib.unique
+  ];
 
-      # Remove empty directories and activate vimGenDocHook
-      # TODO: figure out why we are running the `preFixup` hook in `postBuild`
-      postBuild = ''
-        find $out -type d -empty -delete
-        runHook preFixup
-      '';
-      passthru = {
-        inherit python3Dependencies;
-      };
-    }
-  );
+  # Combined plugin
+  combinedPlugin =
+    lib.pipe
+      {
+        name = "plugin-pack";
+        paths = overriddenPlugins;
+        inherit pathsToLink;
+
+        # buildEnv uses runCommand under the hood. runCommand doesn't run any build phases.
+        # To run custom commands buildEnv takes postBuild argument.
+        # fixupPhase is used for propagating build inputs and to trigger vimGenDocHook
+        postBuild = ''
+          find $out -type d -empty -delete
+          fixupPhase
+        '';
+        passthru = {
+          inherit python3Dependencies;
+        };
+      }
+      [
+        pkgs.buildEnv
+        pkgs.vimUtils.toVimPlugin
+        (drv: drv.overrideAttrs { inherit propagatedBuildInputs; })
+      ];
 
   # Combined plugin configs
   combinedConfig = lib.pipe pluginsToCombine [
--- a/tests/test-sources/modules/performance/byte-compile-lua.nix
+++ b/tests/test-sources/modules/performance/byte-compile-lua.nix
@@ -248,8 +248,8 @@ in
 
       extraPlugins = with pkgs.vimPlugins; [
         nvim-lspconfig
-        # Depends on plenary-nvim
-        telescope-nvim
+        # Depends on nui-nvim
+        noice-nvim
         # buildCommand plugin with python3 dependency
         ((pkgs.writeTextDir "/plugin/test.lua" "vim.opt.tabstop = 2").overrideAttrs {
           passthru.python3Dependencies = ps: [ ps.pyyaml ];
@@ -263,8 +263,8 @@ in
 
         -- Plugins are loadable
         require("lspconfig")
-        require("telescope")
-        require("plenary")
+        require("noice")
+        require("nui.popup")
         require("nvim-treesitter")
 
         -- Python modules are importable
@@ -276,16 +276,13 @@ in
         test_rtp_file("plugin/lspconfig.lua", true)
         test_rtp_file("doc/lspconfig.txt", false)
 
-        -- telescope-nvim
-        test_rtp_file("lua/telescope/init.lua", true)
-        test_rtp_file("lua/telescope/builtin/init.lua", true)
-        test_rtp_file("plugin/telescope.lua", true)
-        test_rtp_file("autoload/health/telescope.vim", false)
-        test_rtp_file("doc/telescope.txt", false)
+        -- noice-nvim
+        test_rtp_file("lua/noice/init.lua", true)
+        test_rtp_file("lua/noice/config/init.lua", true)
+        test_rtp_file("doc/noice.nvim.txt", false)
 
-        -- Dependency of telescope-nvim (plenary-nvim)
-        test_rtp_file("lua/plenary/init.lua", true)
-        test_rtp_file("plugin/plenary.vim", false)
+        -- Dependency of noice-nvim (nui-nvim)
+        test_rtp_file("lua/nui/popup/init.lua", true)
 
         -- Test plugin
         test_rtp_file("plugin/test.lua", true)
--- a/tests/test-sources/modules/performance/combine-plugins.nix
+++ b/tests/test-sources/modules/performance/combine-plugins.nix
@@ -2,10 +2,24 @@
 let
   inherit (pkgs) lib;
 
-  # Count plugins of given type excluding 'build.extraFiles'
-  pluginCount =
-    pkg: files: type:
-    builtins.length (builtins.filter (p: p != files) pkg.packpathDirs.myNeovimPackages.${type});
+  # Assertion for a number of plugins of given type defined in nvimPackage.packpathDirs
+  expectNPlugins =
+    config: type: n:
+    let
+      # 'build.extraFiles' must not be combined, so exclude it from counting
+      plugins = builtins.filter (
+        p: p != config.build.extraFiles
+      ) config.build.nvimPackage.packpathDirs.myNeovimPackages.${type};
+      numPlugins = builtins.length plugins;
+    in
+    {
+      assertion = numPlugins == n;
+      message = "Expected ${toString n} '${type}' plugins defined in 'nvimPackage.packpathDirs', got ${toString numPlugins}: ${
+        lib.concatMapStringsSep ", " lib.getName plugins
+      }.";
+    };
+  # Assertion that exactly one start plugin is defined in nvimPackage.packpathDirs
+  expectOneStartPlugin = config: expectNPlugins config "start" 1;
 
   # Stub plugins
   mkPlugin =
@@ -111,10 +125,7 @@ in
           assert(vim.fn.getcompletion("${name}", "help")[1], "no help tags for '${name}'")
         '') (map lib.getName extraPlugins);
       assertions = [
-        {
-          assertion = pluginCount config.build.nvimPackage config.build.extraFiles "start" == 1;
-          message = "More than one plugin is defined in packpathDirs, expected one plugin pack.";
-        }
+        (expectOneStartPlugin config)
       ];
     };
 
@@ -141,10 +152,7 @@ in
           end), "plugin '${name}' isn't found in runtime as a separate entry, expected not to be combined")
         '') (map lib.getName extraPlugins);
       assertions = [
-        {
-          assertion = pluginCount config.build.nvimPackage config.build.extraFiles "start" >= 2;
-          message = "Only one plugin is defined in packpathDirs, expected at least two.";
-        }
+        (expectNPlugins config "start" (builtins.length extraPlugins))
       ];
     };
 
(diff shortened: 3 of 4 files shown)


modules/top-level/plugins/mk-plugin-pack.nix
```diff
@@ -35,18 +35,28 @@ let
       paths = overriddenPlugins;
       inherit pathsToLink;
 
-      # Remove empty directories and activate vimGenDocHook
-      # TODO: figure out why we are running the `preFixup` hook in `postBuild`
+      # buildEnv uses runCommand under the hood. runCommand doesn't run any build phases.
+      # To run custom commands buildEnv takes postBuild argument.
+      # fixupPhase is used for propagating build inputs and to trigger vimGenDocHook
```
MattSturgeon, line 40: Thanks for improving this comment :+1:

modules/top-level/plugins/mk-plugin-pack.nix
```diff
@@ -35,18 +35,28 @@ let
...
       postBuild = ''
         find $out -type d -empty -delete
-        runHook preFixup
+        fixupPhase
```
MattSturgeon, line 43: Ah, this is the change that was alluding me. Stdenv creates the `propagated-build-inputs` file in the `fixupPhase`, so when I was collecting the `propagatedBuildInputs` if was having no effect because `fixupPhase` wasn't being run.

I'm slightly confused why my second attempt ([manually writing the `propagated-build-inputs` file](https://github.com/nix-community/nixvim/pull/3099/files#diff-69cce1aae191cd61ffbf31244f21b6fb71aa5dce7ff2fb8f9eb78500cbf54239R46-R48)) didn't work though. Maybe some of the tests needed additional work? Or maybe `_addToLuaPath` isn't as recursive as I thought?

modules/top-level/plugins/mk-plugin-pack.nix
```diff
@@ -35,18 +35,28 @@ let
...
   );
 
+  # propagatedBuildInputs contain lua dependencies
+  propagatedBuildInputs = lib.pipe pluginsToCombine [
+    (builtins.catAttrs "plugin")
+    (builtins.catAttrs "propagatedBuildInputs")
+    builtins.concatLists
+    lib.unique
+  ];
+  finalCombinedPlugin = combinedPlugin.overrideAttrs { inherit propagatedBuildInputs; };
```
MattSturgeon, line 58: I'm not too keen on the pattern of progressively building some value, assigning a slightly more final binding name to it each time the previous step is tweaked.

Sometimes it makes sense, other times I'd rather try to inline the steps.

In this case, `combinedPlugin` is already a `buildEnv` call wrapped in a `toVimPlugin` call... Why not also have the `overrideAttrs` call in the same expression?

Either wrapping the assignment with `( ... ).overrideAttrs {}` or using `lib.pipe`:

```nix
combinedPlugin =
  lib.pipe
    {
      name = "plugin-pack";
      paths = overriddenPlugins;
      inherit pathsToLink;
      # ...
    }
    [
      pkgs.buildEnv
      pkgs.vimUtils.toVimPlugin
      (p: p.overrideAttrs {
        inherit propagatedBuildInputs;
      })
    ];
```

?

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -248,8 +248,8 @@ in
 
       extraPlugins = with pkgs.vimPlugins; [
         nvim-lspconfig
-        # Depends on plenary-nvim
-        telescope-nvim
+        # Depends on nui-nvim
+        noice-nvim
```
MattSturgeon, line 252: I wonder if these tests are kinda fragile; if nixpkgs changes whatever plugin we're using as a test-case from `dependencies` to `propagatedBuildInputs` that'll break our test, right?

Would it be better to construct our own test-packages instead of using "real world" examples? Similar to the stub packages in the combine-plugins test.

tests/test-sources/modules/performance/combine-plugins.nix
```diff
@@ -6,31 +6,107 @@ let
...
+  pluginWithLuaDeps1 = ensureDep pkgs.vimPlugins.telescope-nvim "plenary-nvim";
+  pluginWithLuaDeps2 = ensureDep pkgs.vimPlugins.nvim-cmp "plenary-nvim";
+  pluginWithLuaDeps3 = ensureDep pkgs.vimPlugins.gitsigns-nvim "nui-nvim";
 in
 {
   # Test basic functionality
   default =
     { config, ... }:
-    {
+    rec {
```
MattSturgeon, line 88: Why do we need `rec` here? Can't we access things recursively via the `config` arg?

tests/test-sources/modules/performance/combine-plugins.nix
```diff
@@ -42,12 +118,22 @@ in
   # Test disabled option
   disabled =
     { config, ... }:
-    {
+    rec {
```
MattSturgeon, line 121: Same

MattSturgeon (commented):
Thanks you so much for fixing this and improving the test suite! :heart_eyes: 

I have some minor questions, feedback, and discussion, but overall this is looking really good.

I've only skimmed over the tests so far, I'll try and read them in more detail in the next review.

tests/test-sources/modules/performance/byte-compile-lua.nix
khaneliman: Although, I see something similar is done below... so wouldn't be much more effort to just move that logic up a level and share it between these tests.

khaneliman:
> Plugins from luarocks (e.g. telescope-nvim) have dependencies specified in propagatedBuildInputs. These dependencies are not added as plugins in Nvim runtime. They are added to LUA_PATH env var for wrapped neovim. This PR collects all propagatedBuildInputs from input plugin list and puts them in the combined plugin.
> 
> Note that such dependencies are never combined or byte-compiled, because they are not plugins. Also it can lead to double plugins (e.g. one plenary-nvim as plugin dependency and another one in LUA_PATH). This is not nixvim issue, this is how it works in nixpkgs at this moment.
> 
> Most of the work for this PR was done to tests. Because of the changes in nixpkgs to plugin dependencies (especially to plenary.nvim) many tests weren't doing what was intended. To avoid such cases in the future, I'm not using plugins from nixpkgs anymore. I created dummy plugins with dependencies that should be tested (combinePlugins only for now). I hope this will avoid some of the unexpected breakages.

Thanks for tackling this, I don't use the functionality so I've not been very motivated to work on it and appreciate your helping us out. I love the expanded tests using stubbed packages, I agree with Matt that it might be nice to include that package stubbing up a level so you can use it in the byteCompileLua test, as well. But, if you don't want to do it now, we can refactor that later.

modules/top-level/plugins/mk-plugin-pack.nix
stasjok: OK. Will do. First I wrote it with `( ... ).overrideAttrs {}`, but it introduced another level of indentation, so I tried to avoid it. But your second suggestion is also interesting. Not sure if it's readable though because of reverse order.

tests/test-sources/modules/performance/byte-compile-lua.nix
stasjok: You're completely right. I left it for another PR though. This PR is for combinePlugins. But I needed to fix tests to continue. In this PR I also changed tested plugins first. It was a nightmare. Hard to find suitable plugins. And for the last several months many changes were introduced to nixpkgs, It can break anytime in the future. So I dropped it in favor of specially crafted plugins (left the commit for history though).

MattSturgeon: > I left it for another PR though. This PR is for combinePlugins.

SGTM. I'll leave this thread open for easy reference, but it doesn't block the PR.

modules/top-level/plugins/mk-plugin-pack.nix
MattSturgeon: Nothing in this thread blocks the PR. Just musing my thoughts:

All three approaches (`( ... ).overrideAttrs {}`, `lib.pipe`, and `finalCombinedPlugin`) are _fine_.

MattSturgeon (approved):
LGTM, thanks again!

Do you want to rebase? Otherwise @nix-infra-bot will do it for us

MattSturgeon:
@mergifyio queue

PR #3234, opened 2025-04-28
plugins/lspconfig: init

MattSturgeon:
A simplified replacement for `plugins.lsp`, without any additional configuration options or initialisation.

This plugin module only has 5 options:

```nix
nix-repl> nixvimConfigurations.x86_64-linux.default.options.plugins.lspconfig
{
  autoLoad = { ... };
  enable = { ... };
  lazyLoad = { ... };
  package = { ... };
  packageDecorator = { ... }; # internal
}
```

Eventually, once the new `lsp` and `plugins.lspconfig` modules are mature, the old `plugins.lsp` module will be deprecated and/or aliased. I decided it was simpler to introduce the new plugin now and deprecate the old one in a _future_ PR to allow the new one to incubate before we instruct our users to migrate.

This is not very useful without https://github.com/nix-community/nixvim/pull/3203, although it's not technically _blocked_ by it.

### DIFF
--- a/plugins/by-name/lspconfig/default.nix
+++ b/plugins/by-name/lspconfig/default.nix
@@ -1,0 +1,66 @@
+{
+  lib,
+  config,
+  options,
+  ...
+}:
+let
+  # cfg = config.plugins.lspconfig;
+  opts = options.plugins.lspconfig;
+  oldCfg = config.plugins.lsp;
+  oldOpts = options.plugins.lsp;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lspconfig";
+  packPathName = "nvim-lspconfig";
+  package = "nvim-lspconfig";
+
+  maintainers = with lib.maintainers; [
+    GaetanLepage
+    MattSturgeon
+    khaneliman
+    traxys
+  ];
+
+  description = ''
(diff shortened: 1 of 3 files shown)


plugins/by-name/lspconfig/default.nix
```diff
@@ -0,0 +1,65 @@
...
+  name = "lspconfig";
+  packPathName = "nvim-lspconfig";
+  package = "nvim-lspconfig";
+
+  maintainers = with lib.maintainers; [
+    GaetanLepage
+    MattSturgeon
+    khaneliman
+    traxys
+  ];
```
MattSturgeon, line 23: @HeitorAugustoLN you're currently marked as the `plugins.lsp` maintainer, since you did the `mkNeovimPlugin` migration. Do you want to be listed as a maintainer in the new plugin?

Also: does anyone currently in the list _not_ want to be? (@GaetanLepage, @khaneliman, @traxys)

MattSturgeon:
@mergifyio queue

plugins/by-name/lspconfig/default.nix
HeitorAugustoLN: Sure! I will add myself after the merge

PR #3237, opened 2025-04-28
plugins/luasnip: remove redundant jsregexp dependency

stasjok:
This dependency is already handled in nixpkgs. And now it works with `combinePlugins` enabled too.

Also this is tested in nixpkgs (this happened to be one of primary examples where lua deps were needed in the first place):

https://github.com/NixOS/nixpkgs/blob/d99c55d7c10924d671c4fd61f264c207ae75695f/pkgs/applications/editors/neovim/tests/default.nix#L399-L404


MattSturgeon (approved):
Thanks

MattSturgeon:
@mergifyio queue

PR #3239, opened 2025-04-28
modules/lsp: add per-server `name` option

MattSturgeon:
As discussed in https://github.com/nix-community/nixvim/pull/3203#discussion_r2061620488

This allows users to override the name used for a given `lsp.servers.*` entry. The default is still the attr-name.

image

The `name` option is similar in principal to NixOS's [`environment.etc.<name>.target`](https://search.nixos.org/options?channel=unstable&show=environment.etc.%3Cname%3E.target) option, our [`extraFiles.<name>.target`](https://nix-community.github.io/nixvim/NeovimOptions/extraFiles/index.html#extrafilesnametarget) option, etc.

This may be useful to inject raw lua. Users could also use this to configure the same server in different ways, via different attrs; selecting one using its `enable` option.

To enable this change, the PR also moves the lsp module to a dedicated directory and moves the per-server module to a dedicated file.


MattSturgeon:
@mergifyio queue

PR #3242, opened 2025-04-28
plugins/dial: init

### LINKED ISSUE #3235: [PLUGIN REQUEST] dial.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `dial.nvim`  | 
| Homepage | `https://github.com/monaqa/dial.nvim`     | 
| Nixpkgs  | `false` | 

## Extra Information

This plugin extends C-A and C-X to e.g. toggle true/false, increment a date, increment a semver version etc.

GaetanLepage:
Add support for [dial.nvim](https://github.com/monaqa/dial.nvim), an extended increment/decrement plugin for Neovim.

Closes #3235

### DIFF
--- a/plugins/by-name/dial/default.nix
+++ b/plugins/by-name/dial/default.nix
@@ -1,0 +1,12 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "dial";
+  packPathName = "dial.nvim";
+  package = "dial-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  # This plugin does not have a conventional setup function
+  hasSettings = false;
+  callSetup = false;
+}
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):
It feels like we could come up with a way to represent the plugin's config in a structured nix rfc42 style.

In particular, we could map a freeform option to `register_group()`.

Perhaps you could add TODO comments with some ideas?

MattSturgeon:
> The API could break at some point. I don't know if we should add extra non-standard options for niche plugins like this one.

This argument applies (to varying degrees) to all plugins. It's not a question of whether the API _can_ break; that's always possible.

Rather it's a balancing act between factors like "_how_ stable is the API", "how likely are _we_ (the nixvim maintainers) to notice if something breaks", "how much effort is it to implement", "how many users will it benefit", etc.


> It is debatable, I could be convinced of the necessity of such options.

No options are truely necessary; everything can be done in raw lua. But structured/freeform options are the main value-add offered by nixvim.


> In the meantime, someone could do it, but I will leave this PR as is for the initialization.

That was the idea behind the suggestion of adding TODO comments; documenting in the code that someone interested may wish to open a PR

---

To be clear: I have no opinion on whether this plugin is worth the effort. I'd lean towards probably not.

Just documenting my general thoughts here for the record.

PR #3250, opened 2025-04-30
modules/lsp: don't enable wildcard server

jolars:
The wildcard server ("*") should have its config set, but not be enabled through `vim.lsp.enable()`. Calling `vim.lsp.enable("*")` causes an error.

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -93,10 +93,13 @@ in
       '';
       default = { };
       example = {
-        "*".settings = {
-          root_markers = [ ".git" ];
-          capabilities.textDocument.semanticTokens = {
-            multilineTokenSupport = true;
+        "*" = {
+          enable = true;
+          settings = {
+            root_markers = [ ".git" ];
+            capabilities.textDocument.semanticTokens = {
+              multilineTokenSupport = true;
+            };
           };
         };
         luals.enable = true;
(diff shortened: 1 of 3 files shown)


modules/lsp/default.nix
```diff
@@ -88,7 +88,7 @@ in
             ''
               vim.lsp.config(${luaName}, ${luaCfg})
             ''
-            + lib.optionalString server.activate ''
+            + lib.optionalString (server.activate && server.name != "*") ''
```
MattSturgeon, line 91: We should move this to the `activate` option's default, and also add a literal expression `defaultText`:


In modules/LSP/server.nix:
```nix
activate = mkOption {
  # ...
  default = config.name != "*";
  defaultText = lib.literalExression ''
    config.name != "*"
  '';
};
```

MattSturgeon (commented):
Thanks for identifying the issue and sending a fix!

If you're able to, could you move the fix to the `activate` option's default value?

modules/lsp/default.nix
jolars: Done!

MattSturgeon (commented):
Sorry, I think https://github.com/nix-community/nixvim/pull/3254 caused some confusion.

See comments:

MattSturgeon (approved):
@jolars I hope you don't mind, I force-pushed my suggestion.

Thanks for working on this.

MattSturgeon:
@mergifyio queue

PR #3254, opened 2025-04-30
modules/lsp/server: rename `config` -> `settings`

MattSturgeon:
`config` is a loaded term within modules and options.

`settings` is the name usually used for freeform config in nixvim and most other module configurations.

### DIFF
--- a/modules/lsp/server.nix
+++ b/modules/lsp/server.nix
@@ -2,9 +2,7 @@
 {
   name ? null,
   package ? null,
-  # Avoid naming conflict with the `config` module arg
-  # TODO: consider renaming the `config` option to something like `settings`?
-  configOption ? null,
+  settings ? null,
   pkgs ? { },
 }@args:
 { lib, name, ... }:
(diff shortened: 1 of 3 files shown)


modules/lsp/server.nix
```diff
@@ -70,5 +68,23 @@ in
...
+      visible = false;
+    };
   };
+
+  imports = [
+    # TODO: rename added 2025-04-30 (during the 25.05 cycle)
+    # The previous name `config` was introduced 2025-04-28 (during the 25.05 cycle)
+    # Because the previous name `config` never made it into a stable release,
+    # we could consider dropping this alias sooner than normal.
+    (lib.mkRenamedOptionModule [ "config" ] [ "settings" ])
```
khaneliman, line 88: Was gonna say it wasn't alive long enough to really need a rename.. but i supose to we can be nice :P

khaneliman (approved):

MattSturgeon:
@mergifyio queue

modules/lsp/server.nix
MattSturgeon: Maybe, but we've already posted an announcement in matrix and had feedback from users who are already using the new module.

On _that_ basis I figured a short-term warning was worth the small amount of effort.

Issue #3256, opened 2025-04-30
Consider making `lsp.servers.*.enable` default `true`


MattSturgeon:
Often in `attrsOf submodule` where the submodule has an `enable` option, we configure that `enable` to default to `true`.

This allows users to manually enable/disable specific attrs when needed, but means the majority of config does not need to explicitly configure it.

One potential issue with this approach is that we've added explicitly declared options for known servers with packages. These would all become enabled by default, which probably isn't what we want.

Another approach would be to have `enable`'s default depend on the value of other options in the submodule. E.g. the default could be `options.settings.highestPrio < 1500`, `config.settings != { }`, `config.activate`, etc.

Another approach may be to have the `enable` option default to true _only_ for the `"*"` attr. This means most servers would need to be explicitly enabled, but the "global" config in `lsp.servers."*"` would be enabled by default. \
The default could be `config.name == "*"`, or perhaps `config.name == "*" && config.settings != { }`. 

cc @nix-community/nixvim @jolars

khaneliman:
Yeah... I wouldn't want all of them enabled by default.

MattSturgeon:
> Why would all servers be enabled by default? Why deviating from the existing `plugins.lsp.servers.*.enable` defaulting to `false`?

The motivation would be that if a user has configured a lsp server (e.g., configured its `settings`) it is more likely they want it enabled than disabled.

> By
> 
> > Often in `attrsOf submodule` where the submodule has an `enable` option, we configure that `enable` to default to `true`.
> 
> you mean "by convention"?

Yes

> I probably miss some context here.

E.g., [`extraFiles.<name>.enable`](https://nix-community.github.io/nixvim/NeovimOptions/extraFiles/index.html#extrafilesnameenable) is default true, as is [`environment.etc.<name>.enable`](https://search.nixos.org/options?channel=unstable&show=environment.etc.<name>.enable).

MattSturgeon:
> Yeah... I wouldn't want all of them enabled by default.

That's what I was musing on in this comment:

> One potential issue with this approach is that we've added explicitly declared options for known servers with packages. These would all become enabled by default, which probably isn't what we want.

MattSturgeon:
I think this would make more sense if our option for configuring `vim.lsp.config` was de-coupled from our options for activating (`vim.lsp.enable`) and installing packages.

Let's say (for the sake of argument) that we had a `lsp.configs` option:

```nix
options.lsp.configs = mkOption {
[...]

PR #3260, opened 2025-04-30
modules/lsp: avoid empty config & enable `servers."*"` by default

MattSturgeon:
- **modules/lsp: check if server `settings` is empty**

Avoid writing `vim.lsp.config(...)` to `init.lua` when `server.settings == {}`

- **modules/lsp: enable `servers."*"` by default**

Change the `lsp.servers."*".enable` to default to true.

- Extracted from https://github.com/nix-community/nixvim/pull/3258
- Fixes https://github.com/nix-community/nixvim/issues/3256

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -93,6 +93,8 @@ in
             type = types.submodule (
               lib.modules.importApply ./server-base.nix {
                 displayName = "all servers";
+                enable.name = "the `*` server config";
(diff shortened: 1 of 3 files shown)


khaneliman (approved):
After reading your last comment and seeing the PR, I see what you mean about just enabling the `global config` server by default.

MattSturgeon:
@mergify queue

PR #3261, opened 2025-05-01
plugins/schemastore: adapt to new lsp module

jolars:
I have been seeing these warnings 

```
evaluation warning: Nixvim (plugins.schemastore): You have enabled `json` schemas, but `plugins.lsp.servers.jsonls` is not enabled.
evaluation warning: Nixvim (plugins.schemastore): You have enabled `yaml` schemas, but `plugins.lsp.servers.yamlls` is not enabled.
```

when using the new lsp module. This PR adapts the schemastore module to work with the new lsp module as well and updates the tests to include testing against the new lsp module.

### DIFF
--- a/plugins/by-name/schemastore/default.nix
+++ b/plugins/by-name/schemastore/default.nix
@@ -128,15 +128,22 @@ lib.nixvim.plugins.mkVimPlugin {
         message = "You have enabled the plugin, but neither `json` or `yaml` schemas are enabled.";
       }
       {
-        when = !(cfg.json.enable -> config.plugins.lsp.servers.jsonls.enable);
-        message = "You have enabled `json` schemas, but `plugins.lsp.servers.jsonls` is not enabled.";
+        when =
+          cfg.json.enable && !(config.plugins.lsp.servers.jsonls.enable || config.lsp.servers.jsonls.enable);
+        message = ''
+          You have enabled `json` schemas, but neither `plugins.lsp.servers.jsonls` nor `lsp.servers.jsonls` is enabled.
+        '';
       }
       {
-        when = !(cfg.yaml.enable -> config.plugins.lsp.servers.yamlls.enable);
-        message = "You have enabled `yaml` schemas, but `plugins.lsp.servers.yamlls` is not enabled.";
+        when =
+          cfg.yaml.enable && !(config.plugins.lsp.servers.yamlls.enable || config.lsp.servers.yamlls.enable);
+        message = ''
+          You have enabled `yaml` schemas, but neither `plugins.lsp.servers.yamlls` nor `lsp.servers.yamlls` is enabled.
+        '';
       }
     ];
 
+    # TODO: Remove once `plugins.lsp` is defunct.
     plugins.lsp.servers = {
       jsonls.settings = mkIf cfg.json.enable {
         schemas.__raw = ''
@@ -149,6 +156,33 @@ lib.nixvim.plugins.mkVimPlugin {
       };
 
       yamlls.settings = mkIf cfg.yaml.enable {
+        schemaStore = {
+          # From the README: "You must disable built-in schemaStore support if you want to use
+          # this plugin and its advanced options like `ignore`."
+          enable = mkDefault false;
+
+          # From the README: "Avoid TypeError: Cannot read properties of undefined (reading 'length')"
+          url = mkDefault "";
+        };
+
+        schemas.__raw = ''
+          require('schemastore').yaml.schemas(${lib.nixvim.toLuaObject cfg.yaml.settings})
+        '';
+      };
+    };
+
+    lsp.servers = {
+      jsonls.settings.settings = mkIf cfg.json.enable {
+        schemas.__raw = ''
+          require('schemastore').json.schemas(${lib.nixvim.toLuaObject cfg.json.settings})
+        '';
+
+        # The plugin recommends to enable this option in its README.
+        # Learn more: https://github.com/b0o/SchemaStore.nvim/issues/8
+        validate.enable = mkDefault true;
+      };
+
+      yamlls.settings.settings = mkIf cfg.yaml.enable {
         schemaStore = {
           # From the README: "You must disable built-in schemaStore support if you want to use
           # this plugin and its advanced options like `ignore`."
--- a/tests/test-sources/plugins/by-name/schemastore/default.nix
+++ b/tests/test-sources/plugins/by-name/schemastore/default.nix
@@ -1,5 +1,94 @@
 {
   empty = {
+    lsp = {
+      servers = {
+        jsonls.enable = true;
+        yamlls.enable = true;
+      };
+    };
+
+    plugins.schemastore.enable = true;
+  };
+
+  example = {
+    lsp = {
+      servers = {
+        jsonls.enable = true;
+        yamlls.enable = true;
+      };
+    };
+
+    plugins.schemastore = {
+      enable = true;
+
+      json = {
+        enable = true;
+        settings = {
+          replace."package.json" = {
+            description = "package.json overridden";
+            fileMatch = [ "package.json" ];
+            name = "package.json";
+            url = "https://example.com/package.json";
+          };
+          extra = [
+            {
+              description = "My custom JSON schema";
+              fileMatch = "foo.json";
+              name = "foo.json";
+              url = "https://example.com/schema/foo.json";
+            }
+            {
+              description = "My other custom JSON schema";
+              fileMatch = [
+                "bar.json"
+                ".baar.json"
+              ];
+              name = "bar.json";
+              url = "https://example.com/schema/bar.json";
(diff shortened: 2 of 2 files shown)


jolars:
On Thu May 1, 2025 at 4:21 PM CEST, Matt Sturgeon wrote:
> @MattSturgeon commented on this pull request.
>
> LGTM overall. Thanks for spotting this and dealing with it!
>
> One comment regarding whether `or` is needed and a question regarding whether nvim-lspconfig is needed:
>
>> @@ -128,15 +128,30 @@ lib.nixvim.plugins.mkVimPlugin {
>          message = "You have enabled the plugin, but neither `json` or `yaml` schemas are enabled.";
>        }
>        {
> -        when = !(cfg.json.enable -> config.plugins.lsp.servers.jsonls.enable);
> -        message = "You have enabled `json` schemas, but `plugins.lsp.servers.jsonls` is not enabled.";
> +        when =
> +          cfg.json.enable
> +          && !(
> +            (config.plugins.lsp.servers.jsonls.enable or false) || (config.lsp.servers.jsonls.enable or false)
>
> Note: we only need `or false` here when dealing with a freeform attr that can be absent from `config`. If there's an explicitly declared option, then the option's value is guaranteed to exist in `config`.
>
> Admittedly, this gets confusing when a submodule option includes _both_ a `freeformType` _and_ explicitly declared sub-options (like `lsp.servers` does)...
>
> In this case there _is_ a declared option: https://nix-community.github.io/nixvim/lsp/servers/jsonls.html#lspserversjsonlsenable
>
> Same for yamlls: https://nix-community.github.io/nixvim/lsp/servers/yamlls.html
>
> <details><summary>Caveats</summary>
> <p>
>
> That said, currently we only declare explicit sub-options for servers where we know the package name. If that ever changed and these packages were removed from [`lsp-packages.nix`](https://github.com/nix-community/nixvim/blob/c91753bbe57781e08226829b9e354facd50caf1b/plugins/lsp/lsp-packages.nix) then we wouldn't declare the relevant option anymore.
>
> Although, even in that scenario, our test suite would catch that this condition is reading a missing attr.

Thanks, yes I didn't think of that. I'll update it.

> </p>
> </details>
>
>> +          You have enabled `json` schemas, but neither
> +          `plugins.lsp.servers.jsonls` nor `lsp.servers.jsonls` is enabled.
>
> Is it enough to enable `lsp.servers.jsonls` without also installing nvim-lspconfig (via `plugins.lspconfig.enable`)?
>
> The old `plugins.lsp` approach handled _both_ installing nvim-lspconfig _and_ configuring the servers, but with the new approach this is separated.
>
> I assume that nvim-lspconfig isn't a hard requirement for schemastore, since users _could_ configure the server manually. But I also assume in most cases users will want it?

Yes, I forgot about lspconfig, but yes I think it should be enabled too.

> Q1: Do we need to update the warning condition (or add additional warnings) to also check if nvim-lspconfig is enabled?
> Q2: (even if not) should we update the warning message to recommend enabling `plugins.lspconfig`?
>
>> @@ -163,5 +178,32 @@ lib.nixvim.plugins.mkVimPlugin {
>          '';
>        };
>      };
> +
> +    lsp.servers = {
> +      jsonls.settings.settings = mkIf cfg.json.enable {
>
> Ah, I forgot that many servers use `settings` as a table key within their config... Or maybe this is even a standardised thing?
>
> Either way, this is making me rethink https://github.com/nix-community/nixvim/pull/3254 ... I can't decide if `settings.settings` or `config.settings` is worse :sob:

Yes, it is a standard part of the LSP config (https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig) 😁. I probably would have preferred `config.settings` personally,
even if `settings` is more standard in nixvim, at least for plugins. `config` is also closer to the argument name in the neovim LSP API, so there is a case
to be made in that sense too. But I don't think it's a big deal.

MattSturgeon:
> Yes, I forgot about lspconfig, but yes I think it should be enabled too.

The awkward thing here is most users will want lspconfig, so we should warn those who don't enable it. However some users will prefer to configure LSP servers manually _without_ installing lspconfig, and IDK how to avoid pestering them with warnings.

> > Ah, I forgot that many servers use `settings` as a table key within their config... Or maybe this is even a standardised thing?
> > Either way, this is making me rethink #3254 ... I can't decide if `settings.settings` or `config.settings` is worse 😭
> 
> Yes, it is a standard part of the LSP config ([`:h vim.lsp.ClientConfig`](https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig)) 😁.
>
> I probably would have preferred `config.settings` personally, even if `settings` is more standard in nixvim, at least for plugins. `config` is also closer to the argument name in the neovim LSP API, so there is a case to be made in that sense too. But I don't think it's a big deal.

Hm... Thanks for the info.

I'm not opposed to renaming it again, but I'd rather do it sooner than later before the new module gains more users and/or makes it into a 25.05 stable branch. Shall we open an new issue/discussion?

jolars:
> > Yes, I forgot about lspconfig, but yes I think it should be enabled too.
> 
> The awkward thing here is most users will want lspconfig, so we should warn those who don't enable it. However some users will prefer to configure LSP servers manually _without_ installing lspconfig, and IDK how to avoid pestering them with warnings.
> 
> > > Ah, I forgot that many servers use `settings` as a table key within their config... Or maybe this is even a standardised thing?
> > > Either way, this is making me rethink #3254 ... I can't decide if `settings.settings` or `config.settings` is worse 😭
> > 
> > 
> > Yes, it is a standard part of the LSP config ([`:h vim.lsp.ClientConfig`](https://neovim.io/doc/user/lsp.html#vim.lsp.ClientConfig)) 😁.
> > I probably would have preferred `config.settings` personally, even if `settings` is more standard in nixvim, at least for plugins. `config` is also closer to the argument name in the neovim LSP API, so there is a case to be made in that sense too. But I don't think it's a big deal.
> 
> Hm... Thanks for the info.
> 
> I'm not opposed to renaming it again, but I'd rather do it sooner than later before the new module gains more users and/or makes it into a 25.05 stable branch. Shall we open an new issue/discussion?

Hm, yes why not? I'll open one.

jolars:
> > Yes, I forgot about lspconfig, but yes I think it should be enabled too.
> 
> The awkward thing here is most users will want lspconfig, so we should warn those who don't enable it. However some users will prefer to configure LSP servers manually _without_ installing lspconfig, and IDK how to avoid pestering them with warnings.

Ah, yes of course. In that case I don't think we should warn if lspconfig is not enabled. At least I would not be happy myself if I was getting spurious warnings for perfectly good configurations.

plugins/by-name/schemastore/default.nix
```diff
@@ -128,15 +128,24 @@ lib.nixvim.plugins.mkVimPlugin {
         message = "You have enabled the plugin, but neither `json` or `yaml` schemas are enabled.";
       }
       {
-        when = !(cfg.json.enable -> config.plugins.lsp.servers.jsonls.enable);
-        message = "You have enabled `json` schemas, but `plugins.lsp.servers.jsonls` is not enabled.";
+        when =
+          cfg.json.enable && !(config.plugins.lsp.servers.jsonls.enable || config.lsp.servers.jsonls.enable);
+        message = ''
+          You have enabled `json` schemas, but neither
+          `plugins.lsp.servers.jsonls` nor `lsp.servers.jsonls` is enabled.
```
MattSturgeon, line 135: Nit: as-per https://github.com/nix-community/nixvim/pull/3262#discussion_r2072214733, we should wrap lines as we want them to appear in stderr output, not as we want them to appear in code.

```suggestion
          You have enabled `json` schemas, but neither `plugins.lsp.servers.jsonls` nor `lsp.servers.jsonls` is enabled.
```

plugins/by-name/schemastore/default.nix
```diff
@@ -128,15 +128,24 @@ lib.nixvim.plugins.mkVimPlugin {
...
+        '';
       }
       {
-        when = !(cfg.yaml.enable -> config.plugins.lsp.servers.yamlls.enable);
-        message = "You have enabled `yaml` schemas, but `plugins.lsp.servers.yamlls` is not enabled.";
+        when =
+          cfg.yaml.enable && !(config.plugins.lsp.servers.yamlls.enable || config.lsp.servers.yamlls.enable);
+        message = ''
+          You have enabled `yaml` schemas, but neither
+          `plugins.lsp.servers.yamlls` nor `lsp.servers.yamlls` is enabled.
```
MattSturgeon, line 143: Same

```suggestion
          You have enabled `yaml` schemas, but neither `plugins.lsp.servers.yamlls` nor `lsp.servers.yamlls` is enabled.
```

MattSturgeon (commented):
Two nits, otherwise looks good

MattSturgeon (approved):

MattSturgeon:
@mergify queue

PR #3262, opened 2025-05-01
plugins/clangd-extensions: adapt to new lsp module

jolars:
Adapt warnings, default settings, and tests to the new lsp module. Also change the previous warning, which I think should target the clangd server directly, and just not `plugins.lsp`, since the clangd server is necessary for clangd-extensions to function and a user might have disabled it on their end.

I made `lsp.servers.clangd` true by default now, but maybe it would be better to retain having `plugins.lsp.clangd` being set to default while the module matures?

### DIFF
--- a/plugins/by-name/clangd-extensions/default.nix
+++ b/plugins/by-name/clangd-extensions/default.nix
@@ -76,18 +76,27 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.clangd-extensions" {
-      when = !config.plugins.lsp.enable;
+      when = !(config.plugins.lsp.enable || config.lsp.servers.clangd.enable);
 
       message = ''
-        You have enabled `clangd-extensions` but not the lsp (`plugins.lsp`).
-        You should set `plugins.lsp.enable = true` to make use of the clangd-extensions' features.
+        You have enabled `clangd-extensions` but not the lsp (`plugins.lsp` or `lsp.servers.clangd`).
+        You should set `plugins.lsp.enable = true` or `lsp.servers.clangd.enable = true` to make use of the clangd-extensions' features.
       '';
+    };
+
+    lsp = {
+      servers.clangd = {
+        settings = lib.mkIf cfg.enableOffsetEncodingWorkaround {
+          capabilities = {
+            offsetEncoding = [ "utf-16" ];
+          };
+        };
+      };
     };
 
     plugins.lsp = {
       servers.clangd = {
         enable = lib.mkDefault true;
-
         extraOptions = lib.mkIf cfg.enableOffsetEncodingWorkaround {
           capabilities.__raw = ''
             vim.tbl_deep_extend(
--- a/tests/test-sources/plugins/by-name/clangd-extensions/default.nix
+++ b/tests/test-sources/plugins/by-name/clangd-extensions/default.nix
@@ -1,15 +1,16 @@
 {
   empty = {
+    lsp.servers.clangd.enable = true;
+
     plugins = {
-      lsp.enable = true;
       clangd-extensions.enable = true;
     };
   };
 
   default = {
-    plugins = {
-      lsp.enable = true;
+    lsp.servers.clangd.enable = true;
 
+    plugins = {
       clangd-extensions = {
         enable = true;
 
@@ -50,8 +51,9 @@
   };
 
   example = {
+    lsp.servers.clangd.enable = true;
+
     plugins = {
-      lsp.enable = true;
       clangd-extensions = {
         enable = true;
 
@@ -68,6 +70,13 @@
           };
         };
       };
+    };
+  };
+
+  emptyOld = {
+    plugins = {
+      lsp.enable = true;
+      clangd-extensions.enable = true;
     };
   };
 }


MattSturgeon:
> I made `lsp.servers.clangd` true by default now, but maybe it would be better to retain having `plugins.lsp.clangd` being set to default while the module matures?

One difference between `lsp.servers.clangd` and `plugins.lsp.clangd` is that `plugins.lsp.servers.clangd` is additionally gated behind `plugins.lsp.enable`. So enabling `plugins.lsp.servers.clangd` by default has less impact than enabling `lsp.servers.clangd` by default.

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,34 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
     warnings = lib.nixvim.mkWarnings "plugins.clangd-extensions" {
-      when = !config.plugins.lsp.enable;
+      when =
+        !(
+          (config.plugins.lsp.servers.clangd.enable or false) || (config.lsp.servers.clangd.enable or false)
+        );
 
       message = ''
-        You have enabled `clangd-extensions` but not the lsp (`plugins.lsp`).
-        You should set `plugins.lsp.enable = true` to make use of the clangd-extensions' features.
```
MattSturgeon, line 83: I believe the reason for the old wording was that `plugins.lsp.servers.clangd` gets enabled by default, but has no effect until the user manually enables `plugins.lsp`?

Therefore the new wording should suggest enabling _either_ `plugins.lsp` or `lsp.servers.clangd`.

Maybe the warning should also note that the user probably also wants `plugins.lspconfig` if they're using the latter option?

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,34 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          capabilities = {
+            offsetEncoding = [ "utf-16" ];
+          };
+        };
+      };
+    };
 
+    plugins.lsp = {
+      servers.clangd = {
+        enable = lib.mkIf (!config.lsp.clangd.enable or false) (lib.mkDefault true);
```
MattSturgeon, line 106: Nit: we shouldn't need `or false` here because there is an explicitly declared `config.lsp.servers.clangd.enable` option.

Typo: you missed `servers` from the attr path:

```suggestion
        enable = lib.mkIf (!config.lsp.servers.clangd.enable) (lib.mkDefault true);
```

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,34 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        `plugins.lsp.servers.clangd.enable = true` or
+        `lsp.servers.clangd.enable = true` to make use of the
+        clangd-extensions' features.
       '';
     };
 
-    plugins.lsp = {
+    lsp = {
       servers.clangd = {
         enable = lib.mkDefault true;
```
MattSturgeon, line 95: I would drop this, as outlined in my other comments since there isn't a `lsp.enable` option (like `plugins.lsp.enable`) this has more impact on the user's config than a plugin typically should.

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,34 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.clangd-extensions" {
-      when = !config.plugins.lsp.enable;
+      when =
+        !(
+          (config.plugins.lsp.servers.clangd.enable or false) || (config.lsp.servers.clangd.enable or false)
```
MattSturgeon, line 81:
```suggestion
          config.plugins.lsp.enable || config.lsp.servers.clangd.enable
```

Maybe we want an additional warning for:

```nix
{
  when = config.lsp.servers.clangd.enable && !config.plugins.lspconfig.enable;
  message = "TODO: you probably want nvim-lspconfig";
}
```

Although, IDK a sane way to suppress such a warning when users know how to configure the server without installing nvim-lspconfig to set defaults for them?

plugins/by-name/clangd-extensions/default.nix
jolars: You're right, I've changed back. But regarding lspconfig I think we should not default to it as per the discussion in #3261.

plugins/by-name/clangd-extensions/default.nix
jolars: Right, again I think we don't want to pester the user in this case.

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.clangd-extensions" {
-      when = !config.plugins.lsp.enable;
+      when = !((config.plugins.lsp.enable or false) || (config.lsp.servers.clangd.enable or false));
```
MattSturgeon, line 79: We still shouldn't need `or false` for either of these.

`or` use useful when an attr is allowed to be missing; the module system guarantees that all declared options will have a corresponding `config` attr.

Even if the value is undefined (which is impossible for options that have a default), that is implemented as the attr being present but defined as a `throw` expression. IIRC `throw "Option foo.bar.baz used but not defined"`.

Therefore `?` and `or` are only really useful in the module system, when you are dealing with the sub-attrs of some freeform option; e.g. an `attrsOf` option.

```suggestion
      when = !(config.plugins.lsp.enable || config.lsp.servers.clangd.enable);
```

plugins/by-name/clangd-extensions/default.nix
```diff
@@ -76,18 +76,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
 
       message = ''
-        You have enabled `clangd-extensions` but not the lsp (`plugins.lsp`).
-        You should set `plugins.lsp.enable = true` to make use of the clangd-extensions' features.
+        You have enabled `clangd-extensions` but not the lsp
+        (`plugins.lsp` or `lsp.servers.clangd`). You should set
+        `plugins.lsp.enable = true` or
+        `lsp.servers.clangd.enable = true` to make use of the
+        clangd-extensions' features.
```
MattSturgeon, line 86: Style nit: unlike option descriptions and other docs that use markdown formatting, this text is rendered as-is. Including the newlines.

While wrapping the lines like this looks nice in-code, and is usually encouraged, in this scenario it will result in oddly formatted warnings printed to stderr.

While style is subjective, this text must be formatted exactly as you want it to be shown to the user in stderr. Usually that means one line per sentence, or even one line per paragraph.

MattSturgeon (commented):
No issues other than some minor nits, with disproportionally long explanations :sweat_smile:

plugins/by-name/clangd-extensions/default.nix
jolars: Thanks, sorry I thought I had already fixed this.

plugins/by-name/clangd-extensions/default.nix
jolars: I forgot about, this, thanks!

MattSturgeon (approved):
Thanks for this and your patience through reviews!
🎉

MattSturgeon:
@mergify queue

PR #3277, opened 2025-05-03
modules/lsp/servers: move the `*` server to its own module

MattSturgeon:
Introduce a bespoke `global-server.nix` module. This is less DRY, but much simpler.

The `lsp.servers."*"` options are different enough[^1] from the other `lsp.servers.<name>` options that it is simpler to manually declare them separately, in a bespoke module.

Now that we have a dedicated `global-server.nix` module, we no longer need to split the normal server module into `server.nix`+`server-base.nix`

This partially reverts f2e96b67a30859ae21fc626bd33cc74c4d95d356

[^1]: E.g. differences include descriptions, examples, defaults, and some options being missing.


MattSturgeon:
@mergifyio queue

PR #3278, opened 2025-05-03
modules/lsp/keymaps: init

### LINKED ISSUE #3252: modules/lsp: Add keymaps option
It would be great if there was an option to set keymaps via the nixvim API for the new lsp module. 

Either exposing the same kind of functionality that there is for the `plugins.lsp` module, or perhaps use `lsp.servers.<server>.keymaps`, so that keymaps for all servers could be set via `lsp.servers."*".keymaps` and, for instance, keymaps that only should be set for `clangd` for `lsp.servers.clangd.keymaps`.

MattSturgeon:
Initial LSP keymaps option, inspired by `keymapsOnEvent` and `plugins.lsp.keymaps`.

Users can define `lspBufAction` instead of `action` if they wish to use a `vim.lsp.buf.<action>` callback.

---

I initially had a `diagnosticAction` option too, however looking at the manual there aren't many `vim.diagnostic.<action>` functions that'd be useful here.
Previously, `goto_prev` and `goto_next` fit well here, however 0.11 [deprecated](https://neovim.io/doc/user/deprecated.html#deprecated-0.11) those in favour of passing `count=1` or `count=-1` to `jump`, which can't easily be represented by a `diagnosticAction` option.

<details><summary><code>diagnosticAction</code></summary>

```nix
{ config, options, ... }:
{
  options = {
    lspBufAction = lib.mkOption {
      type = types.nullOr types.str;
      description = ''
        LSP buffer action to use for `action`.

        If non-null, the keymap's `action` will be defined as `vim.lsp.buf.<action>`.

        See [`:h lsp-buf`](https://neovim.io/doc/user/lsp.html#lsp-buf)
      '';
      default = null;
      example = "hover";
    };

    diagnosticsAction = lib.mkOption {
      type = types.nullOr types.str;
      description = ''
        Diagnostic action to use for `action`.

        If non-null, the keymap's `action` will be defined as `vim.diagnostic.<action>`.

        See [`:h diagnostic-api`](https://neovim.io/doc/user/diagnostic.html#diagnostic-api)
      '';
      default = null;
      example = "goto_next";
    };
  };

  config.action = lib.mkMerge [
    (lib.mkIf (config.lspBufAction != null) (
      lib.mkDerivedConfig options.lspBufAction (
        action: lib.nixvim.mkRaw "vim.lsp.buf[${lib.nixvim.toLuaObject action}]"
      )
    ))
    (lib.mkIf (config.diagnosticsAction != null) (
      lib.mkDerivedConfig options.diagnosticsAction (
        action: lib.nixvim.mkRaw "vim.diagnostic[${lib.nixvim.toLuaObject action}]"
      )
    ))
  ];
}
```

</details>

Maybe `vim.diagnostic.hide`, `show`, `enable`, etc could be useful, but many of those are often called with arguments too. 🤔

---

I'm on the fence with whether it is better to use `keymapsOnEvent.LspAttach` or to define a dedicated autocmd. I think it's neater to have a separate dedicated autocmd with its own dedicated group.

---

Fixes #3252

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -137,6 +137,10 @@ in
     };
   };
 
+  imports = [
+    ./keymaps.nix
+  ];
+
   config =
     let
       enabledServers = lib.pipe cfg.servers [
--- a/modules/lsp/keymaps.nix
+++ b/modules/lsp/keymaps.nix
@@ -1,0 +1,123 @@
+{ lib, config, ... }:
+let
+  inherit (lib) types;
+  inherit (lib.nixvim.keymaps) mkMapOptionSubmodule;
+
+  cfg = config.lsp;
+
+  # An extra module to include in the keymap option submodule
+  # Declares and implements a specialised `lspBufAction` option
+  extraKeymapModule =
+    { config, options, ... }:
+    {
+      options.lspBufAction = lib.mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          LSP buffer action to use for `action`.
+
+          If non-null, the keymap's `action` will be defined as `vim.lsp.buf.<action>`.
+
+          See [`:h lsp-buf`](https://neovim.io/doc/user/lsp.html#lsp-buf)
+        '';
+        default = null;
+        example = "hover";
+      };
+
+      config.action = lib.mkIf (config.lspBufAction != null) (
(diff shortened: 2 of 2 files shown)


MattSturgeon:
@mergifyio queue

PR #3280, opened 2025-05-03
modules/lsp: add `onAttach` option

MattSturgeon:
Similar to `plugins.lsp.onAttach`, implement a "global" equivalent to the per-server `on_attach` callback.

The implementation is simpler than `plugins.lsp.onAttach`, by using a `LspAttach` autocmd. The old impl was to propagate the global _onAttach_ lines into each server's `on_attach` callback.

The old impl should be able to migrate to using the new option under the hood, or even become an alias of it.

Alternatively, we might decide we don't want this at all? Although it seems convenient.

Perhaps in future it'd be nice to have a similar per-server `onAttach` option? When the per-server `onAttach` option is defined, `settings.on_attach` would be derived from it, using it as the function body.

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -20,6 +20,7 @@ in
   imports = [
     ./servers
     ./keymaps.nix
+    ./on-attach.nix
   ];
 
   config = {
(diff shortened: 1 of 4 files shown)


jfly (approved):
This would make my work in https://github.com/nix-community/nixvim/pull/3289 simpler!

MattSturgeon:
@mergify queue

jfly (commented):
Thanks!

PR #3282, opened 2025-05-04
colorschemes/kanagawa-paper: init

FKouhai:
this PR introduces the kanagawa-paper colorsheme config via nixvim.
the package is already on nixpkgs-unstable and nixos-unstable.
PR in nixpkgs: https://github.com/NixOS/nixpkgs/pull/403656

testing wise im encountering some issues with tests being "killed"
image

### DIFF
--- a/plugins/colorschemes/kanagawa-paper.nix
+++ b/plugins/colorschemes/kanagawa-paper.nix
@@ -1,0 +1,55 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "kanagawa-paper";
+  isColorscheme = true;
+  packPathName = "kanagawa-paper.nvim";
+  package = "kanagawa-paper-nvim";
+
+  description = ''
+    You can select the theme in two ways:
+    - Set `colorschemes.kanagawa-paper.settings.theme` AND explicitly unset `vim.o.background` (i.e. `opts.background = ""`).
+    - Set `colorschemes.kanagawa-paper.settings.background` (the active theme will depend on the value of `vim.o.background`).
+  '';
+
+  maintainers = [ lib.maintainers.FKouhai ];
+
+  settingsExample = {
+    background = "dark";
+    cache = false;
+    colors = {
+      palette = { };
+      theme = {
+        ink = { };
+        canvas = { };
+      };
+    };
+
+    undercurl = true;
+    styles = {
+      comments = {
+        italic = true;
+      };
+      functions = {
+        italic = true;
+      };
+      keywords = {
+        italic = true;
+      };
+      statement_style = {
+        bold = true;
+      };
+    };
+
+    transparent = true;
+    auto_plugins = false;
+    dim_inactive = false;
+    gutter = false;
+    compile = false;
+    overrides = lib.nixvim.nestedLiteralLua "function(colors) return {} end";
+    terminal_colors = false;
+    theme = "ink";
+  };
+}
--- a/tests/test-sources/plugins/colorschemes/kanagawa-paper.nix
+++ b/tests/test-sources/plugins/colorschemes/kanagawa-paper.nix
@@ -1,0 +1,48 @@
+{
+  empty = {
+    # disable test due to this error: ERROR: Cache updated: kanagawa-paper-plugins.json
+    # initial assessment was to disable cache via plugin config but that did not work
+    # then disabled the plugins loading on auto and that did not work either, might be a runtime setup
+    test.runNvim = false;
+    colorschemes.kanagawa-paper.enable = true;
+  };
+
+  defaults = {
+    test.runNvim = false;
+    colorschemes.kanagawa-paper = {
+      enable = true;
+
+      settings = {
+        cache = false;
+        colors = {
+          palette = { };
+          theme = {
+            ink = { };
+            canvas = { };
+          };
+        };
+        undercurl = true;
+        styles = {
+          comments = {
+            italic = true;
+          };
+          functions = {
+            italic = true;
+          };
+          keywords = {
+            italic = true;
+          };
+          statement_style = {
+            bold = true;
+          };
+        };
+        transparent = true;
+        auto_plugins = false;
+        dim_inactive = false;
+        terminal_colors = false;
+        gutter = false;
+        theme = "ink";
+      };
+    };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -12,6 +12,7 @@
     ./colorschemes/github-theme.nix
     ./colorschemes/gruvbox.nix
     ./colorschemes/kanagawa.nix
+    ./colorschemes/kanagawa-paper.nix
     ./colorschemes/melange.nix
     ./colorschemes/modus.nix
     ./colorschemes/monokai-pro.nix


FKouhai:
I'm not totally sure why it fails on CI
```
copying 1 paths...copying path '/nix/store/8rrj5nf429snckcl3x18987k21drxm0i-neovim-0.11.1' from 'ssh-ng://'...copying 1 paths...copying path '/nix/store/nzvd376xafkw479g3737w804zpsj3dmy-neovim-0.11.1' from 'ssh-ng://'...building '/nix/store/y1944shhn5qqvz2984ji1lajhjqanwsg-empty.drv' on 'ssh-ng://nix@darwin02.nix-community.org'...building '/nix/store/55fnhggmms55bn8kj2y7wm0by3jirrm6-defaults.drv' on 'ssh-ng://nix@darwin02.nix-community.org'...copying 0 paths...building '/nix/store/y1944shhn5qqvz2984ji1lajhjqanwsg-empty.drv'...copying 0 paths...empty> ERROR: Cache updated: kanagawa-paper-plugins.json
```
I took a quick look at https://buildbot.nix-community.org/#/builders/3864/builds/859/steps/1/logs/stdio but im very uncertain about what that error means

MattSturgeon:
> `empty> ERROR: Cache updated: kanagawa-paper-plugins.json`

This looks like an error from nixvim's `run nvim` test. If nvim spits out any stderr or stdout while running the nixvim config under test, we assume that is an error and report it with that `ERROR: ` format.

Just a guess: this could be happening because the plugin is trying to download something into cache, which fails because the test runs in a sandbox environment with no network access.

If so, there may be some settings/config we can pass to disable the plugin doing that. Alternatively, we can just disable the `run nvim` test if we can't find a sane way to make it work for this plugin. That can be done by setting `test.runNvim = false` in the `empty` test module.

plugins/colorschemes/kanagawa-paper.nix
```diff
@@ -0,0 +1,230 @@
...
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "kanagawa-paper";
+  isColorscheme = true;
+  packPathName = "kanagawa-paper.nvim";
+  package = "kanagawa-paper-nvim";
+
+  description = ''
+    You can select the theme in two ways:
+    - Set `colorschemes.kanagawa-paper.settings.theme` AND explicitly unset `vim.o.background` (i.e. `options.background = ""`).
```
MattSturgeon, line 17: I believe nixvim's `vim.o` option is `opts` not `options`

```suggestion
    - Set `colorschemes.kanagawa-paper.settings.theme` AND explicitly unset `vim.o.background` (i.e. `opts.background = ""`).
```

plugins/colorschemes/kanagawa-paper.nix
```diff
@@ -0,0 +1,230 @@
...
+      ''
+        Use `colorschemes.kanagawa-paper.settings.overrides` but you now have to add the full function definition:
+        ```
+          function(colors)
+            ...
+          end
+        ```
+      ''
+    )
+  ];
```
MattSturgeon, line 71: If this is a new plugin, why do we have deprecation warnings? Unless I'm missing something, these can go.

plugins/colorschemes/kanagawa-paper.nix
```diff
@@ -0,0 +1,230 @@
...
+                        parameter = "yellow";
+                    };
+                };
+                all = {
+                    ui = {
+                        bg_gutter = "none";
+                    };
+                };
+              }
+            ```
```
MattSturgeon, line 146: We usually avoid manually putting examples in the description. We should use `mkOption`'s `example` arg.

To get access to passing mkOption args in a defaultNullOpts function, use the `'` (prime) variant.

---

That said, we've been gradually transitioning away from declaring any `settingsOptions` _at all_. Preferring to _only_ define `settingsExample`.

plugins/colorschemes/kanagawa-paper.nix
```diff
@@ -0,0 +1,230 @@
...
+        ink.ui.float.bg = "none";
+        canvas.syn.parameter = "yellow";
+        all.ui.bg_gutter = "none";
+      };
+      palette = {
+        sumiInk0 = "#000000";
+        fujiWhite = "#FFFFFF";
+      };
+    };
+    overrides = "function(colors) return {} end";
```
MattSturgeon, line 227:
```suggestion
    overrides = lib.nixvim.nestedLiteralLua "function(colors) return {} end";
```

plugins/colorschemes/kanagawa-paper.nix
MattSturgeon: A tip regarding options:

If you can get a reference to the _actual_ option struct, you can stringify it to it's human-readable opition path.

E.g. `"${options.colorschemes.kanagawa-paper.settings}"` will stringify to `colorschemes.kanagawa-paper.settings` or `programs.nixvim.colorschemes.kanagawa-paper.settings`, depending on context.

Note this only works for _actual_ options, not sub-attrs within a freeform option or sets that contain options (e.g. `options.colorschemes.kanagawa-paper` is not an option, neither is `options.colorschemes.kanagawa-paper.settings.foo.bar`).

For sub-attrs you can uaually just type that bit manually though: e.g. `"${options.colorschemes.kanagawa-paper.settings}.foo.bar"`

The advantage of this approach are a) you get an error if you have a typo b) you get the correct context-dependant prefix in warnigns (doesn't usually affect docs).

FKouhai:
well I ended up just disabling that test thank you for all the suggestions :)

tests/test-sources/plugins/colorschemes/kanagawa-paper.nix
```diff
@@ -0,0 +1,45 @@
+{
+  empty = {
+    colorschemes.kanagawa-paper.enable = true;
+    test.runNvim = false;
```
MattSturgeon, line 4: Could you add a comment about _why_ it is disabled, e.g. the errors you see and/or the suspected underlying issue?

Also _very_ minor nit: if the other test has `test.runNvim` first, could it also be first in this test case? :sweat_smile:

MattSturgeon (approved):
> well I ended up just disabling that test thank you for all the suggestions :)

No problem.

---

If you're disabling the test, could you manually verify the plugin works with your nixvim config?

You can point your config's flake input to this PR branch using github's fancy `refs/pull/<number>/merge` ref:

```sh
nix flake lock --override-input nixvim 'github:nix-community/nixvim?ref=pull/3282/merge'
```

---

Also:

Next time you push to this branch, it may be an idea to fetch upstream and rebase onto it first, as once this is ready to merge it'll be faster if its already rebased.

FKouhai:
tested the plugin locally by adding it to my config and seems to work :)

tests/test-sources/plugins/colorschemes/kanagawa-paper.nix
FKouhai: yessir :)

FKouhai: yeah I should've added the comment earlier. I've added a comment on the current commit and re-ordered the test cases

MattSturgeon (approved):

MattSturgeon:
Thanks!

MattSturgeon:
@mergifyio refresh

MattSturgeon:
@mergifyio queue

PR #3284, opened 2025-05-04
plugins/blink-copilot: Mention in the description that copilot-lua is enabled by default

### LINKED ISSUE #3283: blink-copilot does not depend on copilot-lua
we should remove the code below since blink-copilot does not depend on copilot-lua since https://github.com/fang2hou/blink-copilot/commit/bdc45bbbed2ec252b3a29f4adecf031e157b5573

https://github.com/nix-community/nixvim/blob/c978122396a4208bf1965d346b7456e7256fe70c/plugins/by-name/blink-copilot/default.nix#L59C2-L61C5

GaetanLepage:
`copilot-lua` is not needed by `blink-copilot` anymore.

Reference: https://github.com/fang2hou/blink-copilot/commit/bdc45bbbed2ec252b3a29f4adecf031e157b5573

Fixes #3283.

### DIFF
--- a/plugins/by-name/blink-copilot/default.nix
+++ b/plugins/by-name/blink-copilot/default.nix
@@ -8,6 +8,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
   description = ''
     This plugin should be configured through blink-cmp's `sources.providers` settings.
 
+    `plugins.copilot-lua` will be enabled by default, to provide a working setup out-of-the-box.
+    You may disable it by explicitly adding `plugins.copilot-lua.enable = false` to your
+    configuration.
+
     For example:
 
     ```nix


MattSturgeon:
This looks good, but I'd like to understand the reasoning better:

I don't fully understand. Does the plugin depend on having a separate LSP installed? How is that managed now that the plugin iterates through all LSPs? Why does the new approach mean we don't have to enable copilot-lua?

mathjiajia:
Yes you have to setup a copilot lsp, so copilot-lua is not the only choice: you may use copilot-lsp, or just setup with the new lsp module like any other lsp servers.

khaneliman:
It's a module default for a sane OOB experience. It's using `defaultPrio` and can easily have a user `copilot-lua.enable = false` if they'd like to use a newer alternative provider. But, we should add a warning, if we don't want to enable a sane default, to notify them if they haven't enabled any copilot lsp provider.

mathjiajia:
> It's a module default for a sane OOB experience. It's using `defaultPrio` and can easily have a user `copilot-lua.enable = false` if they'd like to use a newer alternative provider. But, we should add a warning, if we don't want to enable a sane default, to notify them if they haven't enabled any copilot lsp provider.

That is great

MattSturgeon:
Maybe we should mention the default in the plugin's description, so users know about it and how to disable it. But otherwise I agree, we don't need to change anything on our end.

MattSturgeon (approved):
🎉

PR #3289, opened 2025-05-05
Port lsp-format to use non-deprecated setup instructions

jfly:
Upstream still documents the now-deprecated
`require("lspconfig").<name>.setup` mechanism. I believe the modern alternative is to register a `LspAttach` autocmd. For more context, see:

- https://github.com/lukas-reineke/lsp-format.nvim/issues/98
- https://github.com/lukas-reineke/lsp-format.nvim/pull/99

### DIFF
--- a/plugins/by-name/lsp-format/default.nix
+++ b/plugins/by-name/lsp-format/default.nix
@@ -78,8 +78,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
         - "all" (default): Enable formatting for all language servers
         - "none": Do not enable formatting on any language server.
           You might choose this if for some reason you want to manually call
-          `require("lsp-format").on_attach(client)` in the `onAttach` function of your language
-          servers.
+          `require("lsp-format").on_attach(client)`.
         - list of LS names: Manually choose the servers by name
       '';
       example = [
@@ -91,28 +90,41 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.lsp-format" {
-      when = !config.plugins.lsp.enable;
+      when = !(config.plugins.lsp.enable || config.plugins.lspconfig.enable);
       message = ''
-        This plugin requires `plugins.lsp` to be enabled.
+        This plugin requires either `plugins.lsp` or `plugins.lspconfig` to be enabled.
       '';
     };
 
-    plugins.lsp = {
-      onAttach =
-        lib.mkIf (cfg.lspServersToEnable == "all") # Lua
-          ''
-            require("lsp-format").on_attach(client)
+    autoGroups.nixvim_lsp_fmt_setup.clear = false;
+
+    autoCmd = lib.optionals (cfg.lspServersToEnable != "none") [
+      {
+        desc = "set up autoformatting on LSP attach";
+        event = "LspAttach";
+        group = "nixvim_lsp_fmt_setup";
+        callback =
+          let
+            enableCondition =
+              if cfg.lspServersToEnable == "all" then
+                "true"
+              else if lib.isList cfg.lspServersToEnable then
+                # Lua
+                "vim.list_contains(${lib.nixvim.toLuaObject cfg.lspServersToEnable}, client.name)"
+              else
+                throw "Unhandled value for `lspServersToEnable`: ${
+                  lib.generators.toPretty { } cfg.lspServersToEnable
+                }";
+          in
+          lib.nixvim.mkRaw ''
+            function(args)
+              local client = assert(vim.lsp.get_client_by_id(args.data.client_id))
+              if ${enableCondition} then
+                require("lsp-format").on_attach(client, args.buf)
+              end
+            end
           '';
-
-      servers = lib.optionalAttrs (lib.isList cfg.lspServersToEnable) (
-        lib.genAttrs cfg.lspServersToEnable (serverName: {
-          onAttach.function = # Lua
-            ''
-              require("lsp-format").on_attach(client)
-            '';
-        })
-      );
-
-    };
+      }
+    ];
   };
 }
--- a/tests/test-sources/plugins/by-name/lsp-format/default.nix
+++ b/tests/test-sources/plugins/by-name/lsp-format/default.nix
@@ -1,10 +1,10 @@
 {
-  empty = {
+  legacy-empty = {
     plugins.lsp.enable = true;
     plugins.lsp-format.enable = true;
   };
 
-  example = {
+  legacy-example = {
     plugins = {
       lsp = {
         enable = true;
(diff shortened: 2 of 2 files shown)


plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       message = ''
-        This plugin requires `plugins.lsp` to be enabled.
+        This plugin requires `plugins.lspconfig` to be enabled.
       '';
     };
 
-    plugins.lsp = {
-      onAttach =
-        lib.mkIf (cfg.lspServersToEnable == "all") # Lua
+    autoCmd = [
```
MattSturgeon, line 99: Should we have an autogroup?

```suggestion
    autoGroups.nixvim_lsp_fmt_setup.clear = false;

    autoCmd = [
```

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.lsp-format" {
-      when = !config.plugins.lsp.enable;
+      when = !config.plugins.lspconfig.enable;
```
MattSturgeon, line 93: We currently need to support both `plugins.lsp` and `plugins.lspconfig` side-by side:

```suggestion
      when = !(config.plugins.lsp.enable || config.plugins.lspconfig.enable);
```

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       message = ''
-        This plugin requires `plugins.lsp` to be enabled.
+        This plugin requires `plugins.lspconfig` to be enabled.
```
MattSturgeon, line 95:
```suggestion
        This plugin requires either `plugins.lsp` or `plugins.lspconfig` to be enabled.
```

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       '';
     };
 
-    plugins.lsp = {
-      onAttach =
-        lib.mkIf (cfg.lspServersToEnable == "all") # Lua
+    autoCmd = [
+      {
+        desc = "set up autoformatting on LSP attach";
+        event = "LspAttach";
```
MattSturgeon, line 102:
```suggestion
        event = "LspAttach";
        group = "nixvim_lsp_fmt_setup";
```

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              if cfg.lspServersToEnable == "all" then
+                "true"
+              else if lib.isList cfg.lspServersToEnable then
+                # lua
+                ''
+                  vim.list_contains(${lib.nixvim.toLuaObject cfg.lspServersToEnable}, client.name)
+                ''
+              else
+                assert false;
+                "Unhandled value for cfg.lspServersToEnable: ${cfg.lspServersToEnable}";
```
MattSturgeon, line 115:
```suggestion
                throw "Unhandled value for `lspServersToEnable`: ${lib.generators.toPretty { } cfg.lspServersToEnable}";
```

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        callback = lib.mkIf (cfg.lspServersToEnable != "none") (
+          let
+            enableCondition =
+              if cfg.lspServersToEnable == "all" then
+                "true"
+              else if lib.isList cfg.lspServersToEnable then
+                # lua
+                ''
+                  vim.list_contains(${lib.nixvim.toLuaObject cfg.lspServersToEnable}, client.name)
+                ''
```
MattSturgeon, line 112: This avoids a trailing newline:

```suggestion
                "vim.list_contains(${lib.nixvim.toLuaObject cfg.lspServersToEnable}, client.name)"
```

MattSturgeon (commented):
Looks good conceptually, thanks. May be simplified by #3280 as you wouldn't need to create a dedicated autocmd?

Other than nits, the main issue is that we aren't dropping support for `plugins.lsp` yet. The impl & warnings should support both the new and old way.

EDIT: the test may also need updating, currently it is tripping up on the warning:

> For warnings:
> - Nixvim (plugins.lsp-format): This plugin requires `plugins.lspconfig` to be enabled.

See https://buildbot.nix-community.org/#/builders/2791/builds/4458
and https://buildbot.nix-community.org/#/builders/2875/builds/4451

plugins/by-name/lsp-format/default.nix
jfly: Updated. On the use of the word "either": is nixvim enforcing uniqueness here? Or is OK to have both `plugins.lsp` and `plugins.lspconfig` enabled?

MattSturgeon: > is nixvim enforcing uniqueness here?

Yes, that's done in the `plugins.lspconfig` module:
https://github.com/nix-community/nixvim/blob/4016854bccf192d6a5d7fdd5860571dfa1ece55c/plugins/by-name/lspconfig/default.nix#L50-L64

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,20 +90,20 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       message = ''
-        This plugin requires `plugins.lsp` to be enabled.
+        This plugin requires either `plugins.lsp` or `plugins.lspconfig` to be enabled.
       '';
     };
 
     plugins.lsp = {
       onAttach =
-        lib.mkIf (cfg.lspServersToEnable == "all") # Lua
+        lib.mkIf (config.plugins.lsp.enable && cfg.lspServersToEnable == "all") # Lua
```
MattSturgeon, line 101: i.e., here

plugins/by-name/lsp-format/default.nix
```diff
@@ -114,5 +113,37 @@ lib.nixvim.plugins.mkNeovimPlugin {
       );
 
     };
+
+    autoGroups.nixvim_lsp_fmt_setup.clear = lib.mkIf config.plugins.lspconfig.enable false;
+
+    autoCmd = lib.mkIf config.plugins.lspconfig.enable [
```
MattSturgeon, line 119: and here

MattSturgeon (commented):
I'm unsure about a) the `mkIf`s and b) the dual impl

Regarding b) I think the new impl also works for `plugins.lsp`? Or does it rely on having not used `require('lspconfig')` anywhere else?

Regarding a) it feels like the autocmd would be useful regardless of whether nvim-lspconfig is installed? Theoretically, someone might use this plugin with a lspconfig alternative or manually configuring `vim.lsp.config()`?

Should we just drop the `mkIf` conditions and install the autocmd whenever _this_ plugin is enabled?

MattSturgeon:
> > EDIT: the test may also need updating, currently it is tripping up on the warning:
> 
> Investigating now.

Looks like adding the `|| plugins.lsp.enable` back has fixed the test

plugins/by-name/lsp-format/default.nix
```diff
@@ -91,28 +90,41 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-          onAttach.function = # Lua
-            ''
-              require("lsp-format").on_attach(client)
-            '';
-        })
-      );
+    autoGroups.nixvim_lsp_fmt_setup.clear = false;
 
-    };
+    autoCmd = lib.optionals (cfg.lspServersToEnable != "none") [
```
MattSturgeon, line 101: Was literally just typing a suggestion to move the condition here! :exploding_head:

(nit: I usually prefer to use `mkIf` as it actually avoids making any definitions, using `optionals` will always _define_ `autoCmd = [ ]`, potentially overriding lower-prio definitions)

MattSturgeon (approved):
Awesome! Thanks for the improvement and your patience through the review :heart:

MattSturgeon:
@mergifyio queue

plugins/by-name/lsp-format/default.nix
jfly: Ah, good to know! I'll keep that in mind for the future.

jfly:
> Thanks for the improvement and your patience through the review

Thank you for the quick feedback and thorough review!

PR #3292, opened 2025-05-06
plugins/oil-git-status: init

FKouhai:
this PR adds oil-git-status to nixvim, since it recently got into master branch of nixpkgs
https://github.com/refractalize/oil-git-status.nvim

### DIFF
--- a/plugins/by-name/oil-git-status/default.nix
+++ b/plugins/by-name/oil-git-status/default.nix
@@ -1,0 +1,81 @@
+{
+  lib,
+  config,
+  options,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "oil-git-status";
+  packPathName = "oil-git-status.nvim";
+  package = "oil-git-status-nvim";
+
+  description = ''
+    Add Git Status to oil.nvim directory listings.
+
+    Git status is added to the listing asynchronously after creating the oil directory
+
+    listing so it won't slow oil down on big repositories. The plugin puts the status in two new sign columns
+
+    the left being the status of the index, the right being the status of the working directory
+
+    `win_options.signcolumn` needs to be configured in `plugins.oil.settings`:
+    ```nix
+    plugins.oil = {
+      enable = true;
+      settings = {
+        win_options = {
+          signcolumn = "yes:2,";
+        };
+      };
+    };
+    ```
+  '';
+
+  maintainers = [ lib.maintainers.FKouhai ];
+
+  # Ensure oil-git-status is set up after oil is loaded
+  configLocation = lib.mkOrder 1100 "extraConfigLua";
+
+  dependencies = [
+    "git"
+  ];
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" [
+      {
+        when = !config.plugins.oil.enable;
+        message = ''
+          This plugin requires `plugins.oil` to be enabled
+          ${options.plugins.oil.enable} = true;
+        '';
+      }
+      {
+        when =
+          let
+            value = config.plugins.oil.settings.win_options.signcolumn or null;
+            isKnownBadStr = lib.lists.elem value [
+              "yes"
+              "yes:"
+              "yes:0"
+              "yes:1"
+            ];
+            hasYes = lib.strings.hasInfix "yes" value;
+          in
+          !(lib.strings.isString value && hasYes && !isKnownBadStr);
+
+        message = ''
+          This plugin requires the following `plugins.oil` setting:
+            ${options.plugins.oil.settings} = {
+              win_options = {
+                signcolumn = "yes:2";
+              };
+            };`
+        '';
+      }
+    ];
+  };
+
+  settingsExample = {
+    show_ignored = false;
+  };
+}
--- a/tests/test-sources/plugins/by-name/oil-git-status/default.nix
+++ b/tests/test-sources/plugins/by-name/oil-git-status/default.nix
@@ -1,0 +1,34 @@
+{
+  # empty test case is not needed since having it would make the warning throw an error
+  combine-plugins = {
+    performance.combinePlugins.enable = true;
+    plugins.oil = {
+      enable = true;
+      settings = {
+        win_options = {
+          signcolumn = "yes:2,";
+        };
+      };
+    };
+    plugins.oil-git-status = {
+      enable = true;
+      settings = {
+        show_ignored = false;
+      };
+    };
+  };
+
+  with-oil = {
+    # set settings.win_options.signcolumn since oil-git-status relies on that particular setting in order to function
+    plugins.oil = {
+      enable = true;
+      settings = {
+        win_options = {
+          signcolumn = "yes:2,";
+        };
+      };
+    };
+    plugins.oil-git-status.enable = true;
+  };
+
+}


FKouhai:
as of right now the build fails locally, I'm not totally sure if there's something I'm doing wrong, this plugin relies on a specific oil config to work, which I've set in the tests but these seem to not be working eventhough I've tried disabling test.runNvim

FKouhai:
okay I tested using this PR ref for my flake input and I'm getting the same error I saw on the tests.
image

FKouhai:
based on the generated init.lua
```lua
require("oil-git-status").setup({
    index = {
        ["[' ']"] = " ",
        ["['!']"] = "!",
        ["['?']"] = "?",
        ["['A']"] = "A",
        ["['C']"] = "C",
        ["['D']"] = "D",
        ["['M']"] = "M",
        ["['R']"] = "R",
        ["['T']"] = "T",
        ["['U']"] = "U",
    },
    show_ignored = true,
    working_tree = {
        ["[' ']"] = " ",
        ["['!']"] = "!",
        ["['?']"] = "?",
        ["['A']"] = "A",
        ["['C']"] = "C",
        ["['D']"] = "D",
        ["['M']"] = "M",
        ["['R']"] = "R",
        ["['T']"] = "T",
        ["['U']"] = "U",
    },
})

require("oil").setup({ win_options = { signcolumn = "yes:2" } })

```

I've set that correctly I think

FKouhai:
I'm gonna update the PR to specify that lz-n is needed for this plugin

FKouhai:
I think I'm wrong on lazy loading, I think there's something I'm doing wrong, I've been testing out the config locally on my computer but I dont see the git statuses, despite the lazy loading
image

MattSturgeon:
> okay I needed lazy loading for this

Could it just be that one plugin needs to be setup before the other?

FKouhai:
> > okay I needed lazy loading for this
> 
> Could it just be that one plugin needs to be setup before the other?
> 
> EDIT:
> 
> > I think I'm wrong on lazy loading, I think there's something I'm doing wrong, I've been testing out the config locally on my computer but I dont see the git statuses, despite the lazy loading
> 
> Oh, ok

yeah in theory I would need oil to be up first before oil-git-status, maybe I'm doing something wrong on my lazy loading im not sure

FKouhai:
finally I made it work
image

FKouhai:
pushed with a comment on how to successfully lazy load this plugin

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+
+    ```(nix)
+    # lazy loading
+    plugins.lz-n.enable = true;
+    plugins.oil-git-status = {
+      lazyLoad = {
+        settings = {
+          enable = true;
+          enabled = true;
+          before.__raw = \'\'
```
MattSturgeon, line 25: FYI: to include a `''` literal in an indented string, you use `'''` not `\'\'`.

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+            function()
+              require("oil").setup({win_options = { singcolumn = "yes:2" }})
```
MattSturgeon, line 27: This feels wrong. For example, this would result in `oil` potentially being setup multiple times. Once with the args passed here, and again with `plugins.oil.settings`.

I believe this section of lz-n's docs are relevant, but I've not taken the time to fully understand them: https://github.com/nvim-neorocks/lz.n#plugin-dependencies

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+
+  maintainers = [ lib.maintainers.FKouhai ];
+
+  dependencies = [
+    "git"
+  ];
+
+  settingsExample = {
+    show_ignored = true;
+    index = {
```
MattSturgeon, line 43:
```suggestion
    symbols.index = {
```

https://github.com/refractalize/oil-git-status.nvim#default-config

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+      "['A']" = "A";
+      "['C']" = "C";
+      "['D']" = "D";
+      "['M']" = "M";
+      "['R']" = "R";
+      "['T']" = "T";
+      "['U']" = "U";
+      "[' ']" = " ";
+    };
+    working_tree = {
```
MattSturgeon, line 55:
```suggestion
    symbols.working_tree = {
```

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+  maintainers = [ lib.maintainers.FKouhai ];
+
+  dependencies = [
+    "git"
+  ];
+
+  settingsExample = {
+    show_ignored = true;
+    index = {
+      "['!']" = "!";
```
MattSturgeon, line 44: `['foo']` in lua is equivalent to `foo` or `"foo"` in nix.

```suggestion
      "!" = "!";
```

Also, looking at https://github.com/refractalize/oil-git-status.nvim#default-config, this is the same as the plugin's defaults.

Usually an example should be some non-default value, as the user has no reason to copy an example that is just explicitly setting the defaults.

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+              require("oil").setup({win_options = { singcolumn = "yes:2" }})
+              end
+          \'\';
+        };
+      };
+    };
+  '';
+
+  maintainers = [ lib.maintainers.FKouhai ];
+
```
MattSturgeon, line 36: We may need to set a `configLocation` so that the plugin loads after oil when _not_ lazyloading:

```suggestion

    # Ensure we setup after plugins.oil, which uses the default prio of 1000
    configLocation = lib.mkOrder 1100 "extraConfigLua";

```

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,68 @@
...
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "oil-git-status";
+  packPathName = "oil-git-status.nvim";
+  package = "oil-git-status-nvim";
+
+  description = ''
+    Add Git Status to oil.nvim directory listings.
+    Git status is added to the listing asynchronously after creating the oil directory
```
MattSturgeon, line 12: Is this intended to render as a newline or paragraph break?

If so you need either:

```suggestion
    Add Git Status to oil.nvim directory listings. \
    Git status is added to the listing asynchronously after creating the oil directory
```

for a newline, or:

```suggestion
    Add Git Status to oil.nvim directory listings.

    Git status is added to the listing asynchronously after creating the oil directory
```

for a paragraph break.

plugins/by-name/oil-git-status/default.nix
FKouhai: yeah I have a similar concern with that when I took a look at the docs for plugin dependencies they seemed a bit cryptic

plugins/by-name/oil-git-status/default.nix
FKouhai: I'm going to try this out maybe that way I don't need lazy loading, my main thought was okay oil-git-status is loading before oil so I need to lazy load it but I'll try with this thank you very much :)

FKouhai: yep this did the trick

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,39 @@
...
+        win_options = {
+          signcolumn = "yes:2,";
+        };
+      };
+    };
+    ```
+  '';
+
+  maintainers = [ lib.maintainers.FKouhai ];
+  # Ensure oil-git-status is set up after oil is loaded
```
MattSturgeon, line 29: Nit: I'd have a blank line here:
```suggestion

  # Ensure oil-git-status is set up after oil is loaded
```

tests/test-sources/plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,49 @@
+{
+  empty = {
+    test.runNvim = false;
+    plugins.oil-git-status.enable = true;
+    plugins.oil.enable = true;
+  };
+
+  combine-plugins = {
+    performance.combinePlugins.enable = true;
```
MattSturgeon, line 9: Should we split this into a `defaults` and/or `example` test and test `combinePlugins` separately?

Most of the `settings` _look_ like the plugin defaults(?)

```suggestion
  combine-plugins = {
    performance.combinePlugins.enable = true;
    plugins.oil-git-status.enable = true;
    plugins.oil.enable = true;
  };

  defaults = {
```

tests/test-sources/plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,49 @@
+{
+  empty = {
+    test.runNvim = false;
```
MattSturgeon, line 3:
```suggestion
```

Is this needed? You don't have it on the other test.

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,39 @@
...
+  name = "oil-git-status";
+  packPathName = "oil-git-status.nvim";
+  package = "oil-git-status-nvim";
+
+  description = ''
+    Add Git Status to oil.nvim directory listings.\
+    Git status is added to the listing asynchronously after creating the oil directory\
+    listing so it won't slow oil down on big repositories. The plugin puts the status in two new sign columns\
+    the left being the status of the index, the right being the status of the working directory.
+    signcolumn needs to be configured in oil:
```
MattSturgeon, line 15: Might be clearer to break this up into paragraphs?

```suggestion
    Add Git Status to oil.nvim directory listings.

    Git status is added to the listing asynchronously after creating the oil
    directory listing so it won't slow oil down on big repositories.

    The plugin puts the status in two new sign columns, the left being the
    status of the index, the right being the status of the working directory.

    `win_options.signcolumn` must to be configured in `plugins.oil.settings`:
```

Also, I may have confused the point a little in https://github.com/nix-community/nixvim/pull/3292#discussion_r2078741604, but we don't _usually_ manually wrap lines with `\` unless there's a specific reason we want a manual linebreak without a paragraph break.

MattSturgeon (commented):
Thanks for your persistence, looks like this is nearly there :rocket:

tests/test-sources/plugins/by-name/oil-git-status/default.nix
FKouhai: I can try it out with the new priority change

FKouhai: I'd need to keep this I just tried it out without setting up needed config for oil and setting it up and the test fails

tests/test-sources/plugins/by-name/oil-git-status/default.nix
FKouhai: win_options.signcolumn is not a default option for oil, without that particular setting/toggle oil-git-status would throw

FKouhai:
> Thanks for your persistence, looks like this is nearly there 🚀

thank you :) you've been giving me really good reviews I really appreciate them

plugins/by-name/oil-git-status/default.nix
FKouhai: oh okay sounds good to me

tests/test-sources/plugins/by-name/oil-git-status/default.nix
MattSturgeon: The `defaults` case is about making sure that the defaults for _this_ plugin work as expected. It's not necessarily an issue to configure non-defaults for other plugins, although it's a good idea to add comments explaining _why_.

That said, if we aren't declaring `settingsOptions` with documented plugin-defaults, there's not really a reason to explicitly define the defaults in a test.

Maybe this should be a `with-oil` test case that actually runs nvim, while the `empty` test case does not? No need to define anything that is the default anyway.

FKouhai: hmm I think I get what you're saying I can try it out but I believe I'd still need the combinePlugins module.
In the empty test case I tried making a config for oil without combinePlugins and the test failed, that was my reasoning behind disabling the empty test and just keeping a combine-plugin since oil-git-status relies on that oil setting

FKouhai: I just tried out a defaults test case.
```nix
  defaults = {
    plugins.oil = {
      enable = true;
      settings = {
        win_options = {
          signcolumn = "yes:2,";
        };
      };
    };
    plugins.oil-git-status = {
      enable = true;
    };
  };
```
image

FKouhai: I guess I could drop the combine-plugins test case

FKouhai: or I could keep it, what would you recommend ?

FKouhai: yeah I mean I just added a change in config for it to test it out as well I guess it's significant enough

FKouhai: yeah I think it makes sense to have that name will do that

MattSturgeon:
Should this plugin warn when `!plugins.oil.enable`?

FKouhai:
> Should this plugin warn when `!plugins.oil.enable`?

yes 100% I've not thought of that

FKouhai:
> Should this plugin warn when `!plugins.oil.enable`?

would the warning be something like this.
```nix
  extraConfig = cfg: {
    warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" {
      when = (cfg.settings ? oil.enable);

      message = ''
        In order to use this plugin `plugins.oil` has to be enabled with the following setting
        -> `plugins.oil.settings.win_options.signcolumn = "yes:2,";`
      '';
    };
  };
```

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,56 @@
...
+    "git"
+  ];
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" {
+      when = !config.plugins.oil.enable;
+
+      message = ''
+        In order to use this plugin `plugins.oil` has to be enabled with the following setting
+        -> `plugins.oil.settings.win_options.signcolumn = "yes:2,";`
```
MattSturgeon, line 48: I'd put the settings bit in another warning:

```nix
warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" [
  {
    when = !config.plugins.oil.enable;
    # ...
  }
  {
    # TODO: are there other permitted values?
    # Should we do a regex match instead of equality check?
    when = config.plugins.oil.settings.win_options.signcolumn or null != "yes:2,";
    message = ''
      This plugin requires the following `plugins.oil` setting:
          plugins.oil.settings = {
            win_options.signcolumn = "yes:2,";
          }
    '';
  }
]

File: plugins/by-name/oil-git-status/default.nix [hunk 13]
```diff
@@ -0,0 +1,66 @@
...
+    "git"
+  ];
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" [
+      {
+        when = !config.plugins.oil.enable;
+        message = ''
+          This plugin requires `plugins.oil` to be enabled
+          plugins.oil.enable = true;
```
MattSturgeon, line 48: This is kinda subjective, but if we interpolate the _actual_ option, users will see either `plugins.oil.enable` or `programs.nixvim.plugins.oil.enable` depending on whether their nixvim configuration is a `programs.nixvim` submodule.

This also has the advantage of ensuring no typos.

```suggestion
              ${options.plugins.oil.enable} = true;
```

To do this, you need the `options` module arg of course.

Also, nit: I would indent the code example and add a colon to the end of the previous line, to be consistent with the other warning.

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,66 @@
...
+        '';
+      }
+      {
+        when = !config.plugins.oil.settings.win_options.signcolumn or null != "yes:2,";
+        message = ''
+          This plugin requires the following `plugins.oil` setting:
+            plugins.oil.settings={
```
MattSturgeon, line 55: Same optional/subjective suggestion:

```suggestion
            ${plugins.oil.settings} = {
```

Also, nit: spaces around the `=` please :slightly_smiling_face:

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    warnings = lib.nixvim.mkWarnings "plugins.oil-git-status" [
+      {
+        when = !config.plugins.oil.enable;
+        message = ''
+          This plugin requires `plugins.oil` to be enabled
+          plugins.oil.enable = true;
+        '';
+      }
+      {
+        when = !config.plugins.oil.settings.win_options.signcolumn or null != "yes:2,";
```
MattSturgeon, line 52: Q: is this the only supported valid string, or are there other values that'd work?

E.g. would it be more correct to check `lib.strings.hasInfix "yes:2" signcolumn`?

I'm unfamiliar with oil, but the value looks suspiciously like something that can be configured with more complex syntax and still work for oil-git-status?

MattSturgeon (commented):
Otherwise LGTM

plugins/by-name/oil-git-status/default.nix
MattSturgeon, line 52: CI failure, you have an extra `!`
```suggestion
        when = config.plugins.oil.settings.win_options.signcolumn or null != "yes:2,";
```

https://buildbot.nix-community.org/#/builders/2876/builds/4532

FKouhai: oh man I missed that thanks :)

plugins/by-name/oil-git-status/default.nix
FKouhai: yep I like that :)

plugins/by-name/oil-git-status/default.nix
MattSturgeon: Is it valid to e.g. have something like `"yes:2,3"` or `"yes:2,10"` to have a min & max?

If so, maybe we want `lib.strings.hasPrefix "yes:2,"`?

FKouhai: I don't think that's valid . the generated lua is this
```lua
 require("oil").setup({ win_options = { signcolumn = "yes:2" } })
```
it has to be a number bigger than 2 I think I could skip that `,` to avoid confusion

FKouhai: maybe we could use `lib.strings,hasPrefix "yes:"`

MattSturgeon: > it has to be a number bigger than 2

So would `"yes:3"` be valid?

What I'm getting at is we don't want to show a warning if the user configures something that _is_ valid but isn't explicitly what your warning is checking.

The warning should be somewhat flexible/permissive if possible, while still catching the majority of invalid cases.

FKouhai: yea "yes:3" would be valid, it would just add 1 more column

FKouhai: I have something like this now
```nix
      {
        when = lib.strings.hasInfix "yes:" config.plugins.oil.settings.win_options.signcolumn or (config.plugins.oil.settings.win_options != null);
        message = ''
          This plugin requires the following `plugins.oil` setting:
            ${options.plugins.oil.settings} = {
              win_options= {
                signcolumn = "yes:2";
              };
            };`
        '';
      }
```
but I'm getting an eval error:
```
       error: cannot coerce null to a string: null
```

MattSturgeon: You'll need a slightly more complex `when` because the option may or may not be a string.

```nix
when =
  let
    value = config.plugins.oil.settings.win_options.signcolumn or null;
    value' = lib.strings.removeSuffix ",";
    isKnownBadStr = lib.lists.elem value' [
      # I have no idea if these make sense,
      # but it's an example of using `builtins.elem`
      # to blacklist some possible values
      "yes"
      "yes:"
      "yes:0"
      "yes:1"
    ];
    hasYes = lib.strings.hasInfix "yes" value';
  in
  !(lib.strings.isString value && hasYes && !notKnownBadStr)
```

The key part here is checking `isString value` before using any string-functions on the value, the other parts are examples of how you can do potentially more thorough checks.

FKouhai: hmm I'm going to try that out, thank you very much

MattSturgeon: If whitespace is allowed, then `lib.strings.trim` may also be useful?

FKouhai: I tried with value prime and didn't seem to pass the test so I got rid of it and just kept value and the test seems to pass now

FKouhai:
```nix
      {
        when =
          let
            value = config.plugins.oil.settings.win_options.signcolumn or null;
            #value' = lib.strings.removeSuffix ",";
            isKnownBadStr = lib.lists.elem value [
              "yes"
              "yes:"
              "yes:0"
              "yes:1"
            ];
            hasYes = lib.strings.hasInfix "yes" value;
          in
          !(lib.strings.isString value && hasYes && !isKnownBadStr);

        message = ''
          This plugin requires the following `plugins.oil` setting:
            ${options.plugins.oil.settings} = {
              win_options= {
                signcolumn = "yes:2";
              };
            };`
        '';
      }
```

FKouhai: image

MattSturgeon: > image

The issue here is that you're not checking `isString value` before using `value'`. Maybe you accidentally had `isString value'`?

Still, this is only relevant if we need to strip `,` before using a `elem`-based blacklist.

If needed, you could do a more dynamic blacklist using `any`:

```nix
lib.lists.any (infix: lib.strings.hasInfix infix value) [
  "no"
  "yes:0"
  "yes:1" # This would false positive for e.g. "yes:10"
]
```

FKouhai: I don't think we need to strip the `,` since once it gets generated to lua it's already stripped

FKouhai: I just tried to run the tests without the `,` and they pass as well so I'd saw we are good

FKouhai: image
image

MattSturgeon: > once it gets generated to lua it's already stripped

Is there a custom option `apply` function doing that? :thinking:

plugins/by-name/oil-git-status/default.nix
```diff
@@ -0,0 +1,81 @@
...
+              "yes:1"
+            ];
+            hasYes = lib.strings.hasInfix "yes" value;
+          in
+          !(lib.strings.isString value && hasYes && !isKnownBadStr);
+
+        message = ''
+          This plugin requires the following `plugins.oil` setting:
+            ${options.plugins.oil.settings} = {
+              win_options= {
```
MattSturgeon, line 69: Whitespace:

```suggestion
              win_options = {
```

Also, is this neater with inline attrpath?

```nix
{
  win_options.signcolumn = "yes:2";
}
```

plugins/by-name/oil-git-status/default.nix
FKouhai: might be part of the translation layer, like since in this case it's only that value being set it might just get stripped to prevent a typo or something like that

MattSturgeon (approved):
LGTM, one whitespace nit and a rebase needed, then I think its good to merge.

Once the whitespace is fixed, can you fetch our repo and rebase onto main?

```sh
git fetch --all
git rebase upstream/main
git push --force-with-lease
```

Thanks for your hard work on this!

plugins/by-name/oil-git-status/default.nix
FKouhai: I'm not sure if it's neater I kinda like my brackets lol

plugins/by-name/oil-git-status/default.nix
MattSturgeon: I don't see any custom behaviour in the option declaration: https://github.com/nix-community/nixvim/blob/36d63a7c3e0b28d0a2a798826a3b990867079b21/plugins/by-name/oil/default.nix#L249



> might be part of the translation layer,

If you mean the lua translation layer, that's not relevant to your warning. You are reading the option value (i.e. a nix value) not the rendered lua.

FKouhai:
> LGTM, one whitespace nit and a rebase needed, then I think its good to merge.
> 
> Once the whitespace is fixed, can you fetch our repo and rebase onto main?
> 
> ```shell
> git fetch --all
> git rebase upstream/main
> git push --force-with-lease
> ```
> 
> Thanks for your hard work on this!

nah man thank you for bearing with me, sometimes I can be a bit intense, and thanks for the reviews they were extremely helpful and I got to learn a lot from this  :)

FKouhai:
@Mergifyio queue

MattSturgeon:
@mergifyio queue

PR #3295, opened 2025-05-07
plugins.lsp: alias `onAttach` to new `lsp.onAttach`

MattSturgeon:
Follow up to #3280

Simplify the impl by doing global on-attach logic in `lsp.onAttach` (i.e. a `LspAttach` autocmd) instead of adding lua lines to each server's individual `on_attach` callback function.

This is effectively a `mkAliasOptionModule [ "plugins" "lsp" "onAttach" ] [ "lsp" "onAttach" ]` alias, but it is only applied when `plugins.lsp.enable` is true.

### DIFF
--- a/plugins/by-name/rustaceanvim/default.nix
+++ b/plugins/by-name/rustaceanvim/default.nix
@@ -53,43 +53,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   callSetup = false;
   hasLuaConfig = false;
-  extraConfig =
-    cfg:
-    mkMerge [
-      {
-        globals.rustaceanvim = cfg.settings;
+  extraConfig = cfg: {
+    globals.rustaceanvim = cfg.settings;
 
-        assertions = lib.nixvim.mkAssertions "plugins.rustaceanvim" {
-          assertion = cfg.enable -> !config.plugins.lsp.servers.rust_analyzer.enable;
-          message = ''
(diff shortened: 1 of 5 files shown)


plugins/by-name/rustaceanvim/default.nix
```diff
@@ -53,43 +53,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      }
-      # If nvim-lspconfig is enabled:
-      (mkIf config.plugins.lsp.enable {
-        # Use the same `on_attach` callback as for the other LSP servers
-        plugins.rustaceanvim.settings.server.on_attach = mkDefault ''
-          function(client, bufnr)
-            return _M.lspOnAttach(client, bufnr)
-          end
-        '';
-      })
```
MattSturgeon, line 93: The actual change in this file is removing:
```nix
# If nvim-lspconfig is enabled:
(mkIf config.plugins.lsp.enable {
  # Use the same `on_attach` callback as for the other LSP servers
  plugins.rustaceanvim.settings.server.on_attach = mkDefault ''
    function(client, bufnr)
      return _M.lspOnAttach(client, bufnr)
    end
  '';
})
```

Otherwise, the large diff is just removing the now-redundent `mkMerge`, so it may be easier to read with whitespace changes hidden.

MattSturgeon:
@mergifyio queue

MattSturgeon:
@mergify requeue

PR #3297, opened 2025-05-08
plugins/lsp: correct motivation for `onAttach` alias impl

MattSturgeon:
This PR has no code changes, but it corrects an incorrect and misleading comment in #3295.

Previously I said we alias the definitions instead of the value to allow `mkOrder` to work correctly. That is incorrect, as order-priorities are already sorted in `<opt>.definitions`.

The actual reason to use the definitions instead of the final value is to avoid inf-recursion. The "from" option's `apply` function reads the "to" option's value, which would be defined based on the "from" option's value, which is set by the "from" option's apply function...

For a one-way binding, `mkDerivedConfig` is best. For a two-way binding, `mkAliasAndWrapDefsWithPriority` is necessary.


MattSturgeon:
@mergifyio queue

PR #3302, opened 2025-05-08
flake-modules/new-plugin: tweak template; support normalizing package name automatically

khaneliman:
Update for more modern settingsExample focused approach where we declare
fewer options. New default behavior accepts just the upstream plugin name and performs
the same package name normalization from nixpkgs.


MattSturgeon:
Is there a way we can enforce this stays in sync with plugins/TEMPLATE.nix?

Ideally, we'd use the actual template file and substitute in the generated metadata, so it isn't duplicated in two places. But if that's too hard, we could have a test case that a particular input to the command outputs the same content as plugins/TEMPLATE.nix

🤔

khaneliman:
> Is there a way we can enforce this stays in sync with plugins/TEMPLATE.nix?
> 
> Ideally, we'd use the actual template file and substitute in the generated metadata, so it isn't duplicated in two places. But if that's too hard, we could have a test case that a particular input to the command outputs the same content as plugins/TEMPLATE.nix
> 
> 🤔

Would have to think about it.. maybe we just generated the TEMPLATE.nix from the `new-plugin.py` with the arguments used to populate the template.

flake/dev/new-plugin.py
```diff
@@ -27,10 +26,32 @@
...
+    """
+    Convert name to a normalized package name format.
+
+    Args:
+        name (str): The original name to normalize.
+
+    Returns:
+        str: The normalized package name.
+    """
+    return name.replace(".", "-")
```
MattSturgeon, line 52: The kebab_case function _claims_ it will "replace non-alphanumeric chars with hyphens".

If that's accurate, is this redundant?

If that's inaccurate, maybe it should be fixed there instead? Or is there a reason I'm missing to do this in a separate function?

khaneliman:
Or we could update the docs to mention the `new-plugin.py` and remove the `TEMPLATE.nix`..

flake/dev/new-plugin.py
```diff
@@ -12,12 +12,11 @@
...
-  maintainers = [ lib.maintainers.YOUR_NAME ]; # TODO
+  # TODO
+  maintainers = [ lib.maintainers.YOUR_NAME ];
 
-  settingsOptions = {{
-    exampleOption = lib.nixvim.defaultNullOpts.mkBool false ''
-      Example option for the {name} plugin.
-    '';
+  settingsExample = {{
+    exampleOption = true;
```
MattSturgeon, line 19: This is pretty out of sync with https://github.com/nix-community/nixvim/blob/main/plugins/TEMPLATE.nix

MattSturgeon:
> maybe we just generate the TEMPLATE.nix

I feel like the simplest option may be to regex-replace the relevant lines of the main template?

Or we could remove the static template, but running the generate script could be a barrier to entry for some contributors. I'm unsure if that's a real issue or not, though.

flake/dev/new-plugin.py
khaneliman: It does more than that, though. It's our function for stripping `nvim` We don't do that upstream in nixpkgs. Maybe we can split that into 2 functions.

flake/dev/new-plugin.py
khaneliman: What do you mean out of sync?

khaneliman:
> > maybe we just generate the TEMPLATE.nix
> 
> I feel like the simplest option may be to regex-replace the relevant lines of the main template?
> 

Well, yeah... I thought you were talking about a maintenance burden thing.

flake/dev/new-plugin.py
khaneliman: Updated them to be more in line with comments and values showing different setting examples.

khaneliman:
While I was the mood for changing things, I added some of the stuff I thought I was gonna do shortly after introducing this.

flake/dev/new-plugin.py
```diff
@@ -9,15 +9,18 @@
 default_nix_template = """{{ lib, ... }}:
 lib.nixvim.plugins.mkNeovimPlugin {{
   name = "{name}";
+  moduleName = "LUA_MODULE_NAME"; # TODO replace (or remove entirely if it is the same as `name`)
```
MattSturgeon, line 12:
```suggestion
  moduleName = "LUA_MODULE_NAME"; # TODO Remove if it is the same as `name`
```

The "replace" case should be handled by the generate script, right?

flake/dev/new-plugin.py
```diff
@@ -9,15 +9,18 @@
...
   packPathName = "{originalName}";
   package = "{package}";
 
-  maintainers = [ lib.maintainers.YOUR_NAME ]; # TODO
+  {maintainer_todo}
```
MattSturgeon, line 16: Won't this leave an empty line with two spaces when `maintainer_todo == ''`?

I guess we could cheat and run treefmt and/or nixfmt on the finished file, but that seems excessive.

Should we move the `"\n  "` prefix into the `maintainer_todo` string?

MattSturgeon (approved):
LGTM, thanks

flake/dev/new-plugin.py
MattSturgeon: I don't think IFD is relevant, since we aren't within an nix eval here. This is a python script used at runtime. If we needed to eval some nix we could use `nix-instanciate` or `nix eval` while running the script. If we needed to build something we could do that at runtime too.

But if your point was the script doesn't currently check `moduleName` is correct, then that's a perfectly good reason to keep the comment as-is. :+1:

MattSturgeon (approved):

PR #3304, opened 2025-05-08
plugins/lz-n: add support for keymap(<plugin>).set API

HeitorAugustoLN:
Add a new `keymaps` option to the lz-n plugin that allows users to define
keymaps that use lz.n's keymap(<plugin>).set functionality. This enables 
users to create keymaps that lazy-load plugins when specific keys are pressed,
using the familiar vim.keymap.set() API style.

The implementation:
- Adds a new `keymaps` option using mkMapOptionSubmodule with a custom plugin property
- Exposes a simple and consistent interface for configuring lazy-loaded keymaps
- Processes these keymaps to generate corresponding Lua code with require('lz.n').keymap().set()
- Cleanly separates plugin specs from keymap definitions while preserving their relationship

### DIFF
--- a/plugins/pluginmanagers/lz-n.nix
+++ b/plugins/pluginmanagers/lz-n.nix
@@ -3,14 +3,25 @@
   options,
   ...
 }:
-with lib;
 let
-  inherit (lib.nixvim) defaultNullOpts;
+  inherit (lib)
+    id
+    literalMD
+    mkIf
+    mkOption
+    types
+    ;
+  inherit (lib.nixvim)
+    defaultNullOpts
+    mkNullOrLuaFn
+    mkNullOrOption'
+    toLuaObject
+    ;
 in
-nixvim.plugins.mkNeovimPlugin {
+lib.nixvim.plugins.mkNeovimPlugin {
   name = "lz-n";
   packPathName = "lz.n";
-  maintainers = [ maintainers.psfloyd ];
+  maintainers = [ lib.maintainers.psfloyd ];
   # NOTE: We want to load lz.n as early as possible so that triggers are respected
   configLocation = "extraConfigLuaPre";
 
@@ -44,32 +55,32 @@ nixvim.plugins.mkNeovimPlugin {
             '';
           };
 
-          enabled = nixvim.defaultNullOpts.mkStrLuaFnOr types.bool true ''
+          enabled = defaultNullOpts.mkStrLuaFnOr types.bool true ''
             When false, or if the function returns false, then this plugin will not be included in the spec.
             This option corresponds to the `enabled` property of lz.n.
           '';
 
-          beforeAll = nixvim.mkNullOrLuaFn ''
+          beforeAll = mkNullOrLuaFn ''
             Always executed before any plugins are loaded.
           '';
 
-          before = nixvim.mkNullOrLuaFn ''
+          before = mkNullOrLuaFn ''
             Executed before this plugin is loaded.
           '';
 
-          after = nixvim.mkNullOrLuaFn ''
+          after = mkNullOrLuaFn ''
             Executed after this plugin is loaded.
           '';
 
-          load = nixvim.mkNullOrLuaFn ''
+          load = mkNullOrLuaFn ''
             Can be used to override the `vim.g.lz_n.load()` function for this plugin.
           '';
 
-          priority = nixvim.defaultNullOpts.mkUnsignedInt (literalMD "`50` (or `1000` if `colorscheme` is set)") ''
+          priority = defaultNullOpts.mkUnsignedInt (literalMD "`50` (or `1000` if `colorscheme` is set)") ''
             Only useful for start plugins (not lazy-loaded) to force loading certain plugins first. 
           '';
 
-          event = nixvim.mkNullOrOption' {
+          event = mkNullOrOption' {
             type = types.anything;
             description = ''
               Lazy-load on event. Events can be specified as BufEnter or with a pattern like BufEnter *.lua
@@ -80,7 +91,7 @@ nixvim.plugins.mkNeovimPlugin {
             ];
           };
 
-          cmd = nixvim.mkNullOrOption' {
+          cmd = mkNullOrOption' {
             type = types.anything;
             description = ''
               Lazy-load on command.
@@ -91,7 +102,7 @@ nixvim.plugins.mkNeovimPlugin {
             ];
           };
 
-          ft = nixvim.mkNullOrOption' {
+          ft = mkNullOrOption' {
             type = types.anything;
             description = ''
               Lazy-load on filetype.
@@ -99,7 +110,7 @@ nixvim.plugins.mkNeovimPlugin {
             example = [ "tex" ];
           };
 
-          colorscheme = nixvim.mkNullOrOption' {
+          colorscheme = mkNullOrOption' {
             type = types.anything;
             description = ''
               Lazy-load on colorscheme.
@@ -107,7 +118,7 @@ nixvim.plugins.mkNeovimPlugin {
             example = "onedarker";
           };
 
-          keys = nixvim.mkNullOrOption' {
+          keys = mkNullOrOption' {
             type = types.listOf types.anything;
             description = ''
               Lazy-load on key mapping. Mode is `n` by default.
@@ -191,9 +202,9 @@ nixvim.plugins.mkNeovimPlugin {
     };
 
   extraConfig = cfg: {
-    globals.lz_n = modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
+    globals.lz_n = lib.modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
     plugins.lz-n.luaConfig.content = mkIf (cfg.plugins != [ ]) ''
-      require('lz.n').load( ${nixvim.toLuaObject cfg.plugins})
+      require('lz.n').load( ${toLuaObject cfg.plugins})
     '';
   };
 }
--- a/plugins/pluginmanagers/lz-n.nix
+++ b/plugins/pluginmanagers/lz-n.nix
@@ -7,16 +7,19 @@ let
   inherit (lib)
     id
     literalMD
-    mkIf
+    mkMerge
     mkOption
+    optional
     types
     ;
   inherit (lib.nixvim)
     defaultNullOpts
     mkNullOrLuaFn
     mkNullOrOption'
+    nestedLiteralLua
     toLuaObject
     ;
+  inherit (lib.nixvim.keymaps) mkMapOptionSubmodule;
 in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "lz-n";
@@ -146,6 +149,40 @@ lib.nixvim.plugins.mkNeovimPlugin {
       };
     in
     {
+      keymaps = mkOption {
+        type = types.listOf (mkMapOptionSubmodule {
+          extraOptions.plugin = mkOption {
+            type = with types; either str lzPluginType;
+            example = "telescope.nvim";
+            description = ''
+              The plugin (name or spec) to lazy-load when the keymap is triggered.
+
+              > [!NOTE]
+              > This must match the name used in the `plugins` list or be a valid plugin that can be loaded.
+            '';
+          };
+        });
+        default = [ ];
+        example = [
+          {
+            plugin = "telescope.nvim";
+            key = "<leader>ff";
+            action = nestedLiteralLua "function() require('telescope.builtin').find_files() end";
+            options.desc = "Find files";
+          }
+          {
+            plugin = "neo-tree.nvim";
+            key = "<leader>ft";
+            action = "<CMD>Neotree toggle<CR>";
+            options.desc = "NeoTree toggle";
+          }
+        ];
+        description = ''
+          Define keymaps that will use lz.n's `keymap(<plugin>).set` functionality.
+          This provides a more intuitive way to define lazy-loaded keymaps with the familiar `keymaps` option API style.
+        '';
+      };
+
       plugins = mkOption {
         description = ''
           List of plugin specs provided to the `require('lz.n').load` function.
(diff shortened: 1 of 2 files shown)


plugins/pluginmanagers/lz-n.nix
```diff
@@ -191,9 +201,9 @@ nixvim.plugins.mkNeovimPlugin {
     };
 
   extraConfig = cfg: {
-    globals.lz_n = modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
+    globals.lz_n = lib.modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
```
MattSturgeon, line 247: If you're doing a cleanup-commit (remove `with lib`) maybe we could have another cleanup commit that uses `extraConfig = cfg: opts: {` instead of using an explicit `options.plugins.lz-n`?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -6,6 +6,7 @@
 let
   inherit (lib)
     id
+    literalMD
```
MattSturgeon, line 9: Is this in the right commit? Should it be part of the "remove with lib" commit?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -80,7 +82,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
             Can be used to override the `vim.g.lz_n.load()` function for this plugin.
           '';
 
-          priority = defaultNullOpts.mkUnsignedInt (lib.literalMD "`50` (or `1000` if `colorscheme` is set)") ''
+          priority = defaultNullOpts.mkUnsignedInt (literalMD "`50` (or `1000` if `colorscheme` is set)") ''
```
MattSturgeon, line 85: Same; wrong commit?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -145,6 +147,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              This must match the name used in the `plugins` list or be a valid plugin name that can be loaded.
+            '';
+          };
+        });
+        default = [ ];
+        example = [
+          {
+            plugin = "telescope.nvim";
+            key = "<leader>ff";
+            action.__raw = "function() require('telescope.builtin').find_files() end";
```
MattSturgeon, line 166: May be better with `nestedLiteralLua` than `__raw`:

```suggestion
            action = lib.nixvim.nestedLiteralLua "function() require('telescope.builtin').find_files() end";
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -209,6 +247,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
     globals.lz_n = lib.modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
     plugins.lz-n.luaConfig.content = mkIf (cfg.plugins != [ ]) ''
       require('lz.n').load( ${toLuaObject cfg.plugins})
+
+      -- Set up lz.n keybinds {{{
+      do
+        local __lz_n_keymaps = ${toLuaObject (map removeDeprecatedMapAttrs cfg.keymaps)}
+        for _, map in ipairs(__lz_n_keymaps) do
+          require('lz.n').keymap(map.plugin).set(map.mode, map.key, map.action, map.options)
+        end
```
MattSturgeon, line 256: Thoughts on serializing the whole table and iterating in lua **vs** iterating in nix and rendering multiple `"require('lz.n').keymap(${plugin}).set"` lines?

I suspect the latter may be easier to read in `nixvim-print-init` and was the approach we ended up taking in more recent code, e.g. `lsp.servers` :thinking:

E.g:
```nix
''
do
  ${lib.concatMapStringsSep "\n" (
    { plugin, mode, key, action, options, ... }:
    "require('lz.n').keymap(${toLuaObject plugin}).set(${toLuaObject mode}, ${toLuaObject key}, ${toLuaObject action}, ${toLuaObject options})"
  ) cfg.keymaps}
end
''
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -209,6 +247,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
     globals.lz_n = lib.modules.mkAliasAndWrapDefsWithPriority id options.plugins.lz-n.settings;
     plugins.lz-n.luaConfig.content = mkIf (cfg.plugins != [ ]) ''
       require('lz.n').load( ${toLuaObject cfg.plugins})
+
+      -- Set up lz.n keybinds {{{
+      do
+        local __lz_n_keymaps = ${toLuaObject (map removeDeprecatedMapAttrs cfg.keymaps)}
```
MattSturgeon, line 253: Shouldn't need `removeDeprecatedMapAttrs` in new code; it's for removing the deprecated `lua` attr.

plugins/pluginmanagers/lz-n.nix
```diff
@@ -209,6 +247,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        for _, map in ipairs(__lz_n_keymaps) do
+          require('lz.n').keymap(map.plugin).set(map.mode, map.key, map.action, map.options)
+        end
+      end
+      -- }}}
```
MattSturgeon, line 258: I would wrap this with mkIf and mkMerge to avoid rendering it when there are no keybinds:

```nix
plugins.lz-n.luaConfig.content = mkMerge [
  (mkIf (cfg.plugins != [ ]) "require('lz.n').load(${toLuaObject cfg.plugins})")
  (mkIf (cfg.keymaps != [ ]) ''
    -- Set up lz.n keybinds {{{
    do
      ...
    end
    -- }}}
  '')
];
```

Or, this could be combined with the other suggestion to do something like:

```nix
plugins.lz-n.luaConfig.content = mkMerge (
  [
    (mkIf (cfg.plugins != [ ]) "require('lz.n').load(${toLuaObject cfg.plugins})")
  ]
  ++ map (
    { plugin, mode, key, action, options, ... }:
    "require('lz.n').keymap(${toLuaObject plugin}).set(${toLuaObject mode}, ${toLuaObject key}, ${toLuaObject action}, ${toLuaObject options})"
  ) cfg.keymaps
);
```

MattSturgeon (commented):
Looks good, nice idea!

plugins/pluginmanagers/lz-n.nix
HeitorAugustoLN: Done.

plugins/pluginmanagers/lz-n.nix
HeitorAugustoLN: I like it, changed the implementation

plugins/pluginmanagers/lz-n.nix
```diff
@@ -190,10 +241,20 @@ nixvim.plugins.mkNeovimPlugin {
...
+        {
+          plugin,
+          mode,
+          key,
+          action,
+          options,
+          ...
+        }:
+        "require('lz.n').keymap(${toLuaObject plugin}).set(${toLuaObject mode}, ${toLuaObject key}, ${toLuaObject action}, ${toLuaObject options})"
+      ) cfg.keymaps;
```
MattSturgeon, line 258: Needs parens to make the `++` apply to `mkMerge`'s arg, rather than concatenating the _result_ of calling `mkMerge`:

```suggestion
    mkMerge (
      [ (mkIf (cfg.plugins != [ ]) "require('lz.n').load(${toLuaObject cfg.plugins})") ]
      ++ map (
        {
          plugin,
          mode,
          key,
          action,
          options,
          ...
        }:
        "require('lz.n').keymap(${toLuaObject plugin}).set(${toLuaObject mode}, ${toLuaObject key}, ${toLuaObject action}, ${toLuaObject options})"
      ) cfg.keymaps
    );
```

HeitorAugustoLN: oops 😅

plugins/pluginmanagers/lz-n.nix
```diff
@@ -208,8 +247,19 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      ++ map (
+        {
+          plugin,
+          mode,
+          key,
+          action,
+          options,
+          ...
+        }:
+        "require('lz.n').keymap(${toLuaObject plugin}).set(${toLuaObject mode}, ${toLuaObject key}, ${toLuaObject action}, ${toLuaObject options})"
```
MattSturgeon, line 257: Q: does `set` support any named args pattern?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -146,6 +149,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
     in
     {
+      keymaps = mkOption {
+        type = types.listOf (mkMapOptionSubmodule {
+          extraOptions.plugin = mkOption {
+            type = with types; either str lzPluginType;
+            example = "telescope.nvim";
+            description = ''
+              The name of the plugin to lazy-load when the keymap is triggered.
+              This must match the name used in the `plugins` list or be a valid plugin name that can be loaded.
```
MattSturgeon, line 155: We could emphasize this as a note, warning, etc?

```suggestion

              > [!NOTE]
              > The name must match the name used in the `plugins` list or be a valid plugin name that can be loaded.
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -146,6 +149,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            plugin = "neo-tree.nvim";
+            key = "<leader>ft";
+            action = "<CMD>Neotree toggle<CR>";
+            options = {
+              desc = "NeoTree toggle";
+            };
+          }
+        ];
+        description = ''
+          Define keymaps that will use lz.n's keymap(<plugin>).set functionality.
```
MattSturgeon, line 179: Markdown inline code:

```suggestion
          Define keymaps that will use lz.n's `keymap(<plugin>).set` functionality.
```

plugins/pluginmanagers/lz-n.nix
```diff
@@ -146,6 +149,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        });
+        default = [ ];
+        example = [
+          {
+            plugin = "telescope.nvim";
+            key = "<leader>ff";
+            action = nestedLiteralLua "function() require('telescope.builtin').find_files() end";
+            options = {
+              desc = "Find files";
+            };
```
MattSturgeon, line 167: `toPretty` will render is as you have it regardless, but the in-code value may be neater as:

```suggestion
            options.desc = "Find files";
```

?

plugins/pluginmanagers/lz-n.nix
```diff
@@ -146,6 +149,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            key = "<leader>ft";
+            action = "<CMD>Neotree toggle<CR>";
+            options = {
+              desc = "NeoTree toggle";
+            };
+          }
+        ];
+        description = ''
+          Define keymaps that will use lz.n's keymap(<plugin>).set functionality.
+          This provides a more intuitive way to define lazy-loaded keymaps with the familiar vim.keymap.set() API style.
```
MattSturgeon, line 180:
```suggestion

          This provides a more intuitive way to define lazy-loaded keymaps with the familiar `vim.keymap.set()` API style.
```

Although, maybe it makes more sense to nixvim users if we compare to nixvim's top-level `keymaps` option instead of the vim API?

MattSturgeon (commented):
Diff looks good overall. Minor nits with docs.

Have you tested this in-practice, e.g. with your config?

plugins/pluginmanagers/lz-n.nix
HeitorAugustoLN: No, it doesn't

HeitorAugustoLN:
> Have you tested this in-practice, e.g. with your config?

Yes, I am re-creating my Neovim configuration currently, and I am using it [here](https://github.com/HeitorAugustoLN/nvim-config/blob/9cf3aef16d521d891ccf1a9807d6eaed0fb92d85/nvim/plugins/oil.nix)

MattSturgeon (approved):
LGTM!

Thanks @HeitorAugustoLN for another great PR :tada: 

Are you able to rebase again on `main`?

HeitorAugustoLN:
Not right now, I am not at home

MattSturgeon:
I wonder if we should add a per-plugin abstraction around this?

E.g. a `plugins.<name>.keymaps` option that, when the plugin is enabled:

1. If lazy-loading defines `plugins.lz-n.keymaps`
2. otherwise defines `keymaps`

HeitorAugustoLN:
> I wonder if we should add a per-plugin abstraction around this?
> 
> E.g. a `plugins.<name>.keymaps` option that, when the plugin is enabled:
> 
> 1. If lazy-loading defines `plugins.lz-n.keymaps`
> 2. otherwise defines `keymaps`

I like the idea, but not sure how that would be designed, it would also close #603

MattSturgeon:
> > I wonder if we should add a per-plugin abstraction around this?
> > 
> > E.g. a `plugins.<name>.keymaps` option that, when the plugin is enabled:
> > 
> > 1. If lazy-loading defines `plugins.lz-n.keymaps`
> > 2. otherwise defines `keymaps`
> 
> I like the idea, but not sure how that would be designed

I was thinking we add some stuff to `mkNeovimPlugin`, similar to how `luaConfig` and `lazyLoading` work.

> it would also close #603

Ah, #603 is kinda loaded, because it deals with both custom nixvim keymaps options _and_ plugin-defined keymaps settings. Although if we could have a unified approach to the nixvim options at least, I think we could close #603.

That does remind me that some plugins already have a `keymaps` option that'll conflict, though.

A migration here may be challenging, although maybe we could use differing _types_ to do the migration; `listOf` being the new way, `attrsOf` being deprecated?


> For example, `plugins.telescope.keymaps` `action` option are telescope actions, not regular actions. So it would need an extra option, or behave the same as before.

I think this can be approached with the same way as the new `lsp.keymaps` option added in https://github.com/nix-community/nixvim/pull/3278;

Here we have an extra module in the keymap-submodule that declares a `lspBufAction` option, which if non-null, is transformed into a definition for `action`.

PR #3311, opened 2025-05-11
flake: add overridable `systems` input

MattSturgeon:
Users can specify which systems to use by having our `systems` input follow a different file or flake:

```nix
{
  inputs = {
    # Here we list systems in a local file
    inputs.systems.url = "path:./systems.nix";
    inputs.systems.flake = false;

    inputs.nixvim.url = "github:nix-community/nixvim";
    # Here we override the list of systems with only our own
    inputs.nixvim.inputs.systems.follows = "systems";

    # ...
  };

  outputs = inputs: {
    # ...
  };
}
```

```nix
# systems.nix
[
  "x86_64-linux"
]
```

Alternatively, instead of users listing systems in a local file, they
can use an external flake, e.g.:

- `github:nix-systems/default`
  - Exposes aarch64 and x86_64 for linux and darwin
- `github:nix-systems/default-linux`
  - Exposes aarch64 and x86_64 for linux
- `github:nix-systems/default-darwin`
  - Exposes aarch64 and x86_64 for darwin
- `github:nix-systems/aarch64-darwin`
- `github:nix-systems/aarch64-linux`
- `github:nix-systems/x86_64-darwin`
- `github:nix-systems/x86_64-linux`

See https://github.com/nix-systems/nix-systems

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -22,7 +22,7 @@
     },
     "flake-utils": {
       "inputs": {
-        "systems": "systems"
+        "systems": "systems_2"
       },
       "locked": {
         "lastModified": 1731533236,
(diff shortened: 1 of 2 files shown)


MattSturgeon:
> I always find it weird to rely on an extra repo for this.

We don't have to. We could use `systems.url = "path:./systems.nix"` instead.

Using an external repo is just a minor convenience, in that we don't need our own list of systems.

Also, if use nix 2.26+ to create a lockfile with a relative-path input, then that lockfile isn't backwards compatible with nix <2.26.

MattSturgeon:
> > I always find it weird to rely on an extra repo for this.
> 
> We don't have to. We could use `systems.url = "path:./systems.nix"` instead.

Without https://github.com/edolstra/flake-compat/pull/71, we kinda _do_ need to use an external input.

Using a relative path input will break flake-compat.

In my personal config, I worked around this by using https://github.com/edolstra/flake-compat/pull/71 as my `flake-compat` input, but I'm guessing nixvim would prefer to wait for it to be reviewed and/or accepted upstream.

MattSturgeon:
@mergify queue

PR #3314, opened 2025-05-11
modules/performance: add ability to byte-compile lua packages and lua dependencies

stasjok:
This PR:

- Introduces `performance.byteCompileLua.luaLib` option. When enabled, it'll byte-compile lua libraries in LUA_PATH. Specifically: `extraLuaPackages` and plugins `propagatedBuildInputs`.
- Adds `tests/utils/plugin-stubs.nix` with a shared plugin stubs and generated lua code for using in performance tests.
- Once again refactors performance tests using shared plugin stubs.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -172,9 +172,10 @@ in
   config =
     let
       # Optionally byte compile lua library
+      inherit (import ./plugins/byte-compile-lua-lib.nix { inherit lib pkgs; }) byteCompileLuaPackages;
       extraLuaPackages =
         if config.performance.byteCompileLua.enable && config.performance.byteCompileLua.luaLib then
-          ps: map builders.byteCompileLuaDrv (config.extraLuaPackages ps)
+          ps: byteCompileLuaPackages (config.extraLuaPackages ps)
         else
           config.extraLuaPackages;
 
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -172,7 +172,9 @@ in
   config =
     let
       # Optionally byte compile lua library
-      inherit (import ./plugins/byte-compile-lua-lib.nix { inherit lib pkgs; }) byteCompileLuaPackages;
+      inherit (pkgs.callPackage ./plugins/byte-compile-lua-lib.nix { inherit lib; })
+        byteCompileLuaPackages
+        ;
       extraLuaPackages =
         if config.performance.byteCompileLua.enable && config.performance.byteCompileLua.luaLib then
           ps: byteCompileLuaPackages (config.extraLuaPackages ps)
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -171,11 +171,18 @@ in
 
   config =
     let
+      # Optionally byte compile lua library
+      extraLuaPackages =
+        if config.performance.byteCompileLua.enable && config.performance.byteCompileLua.luaLib then
+          ps: map builders.byteCompileLuaDrv (config.extraLuaPackages ps)
+        else
+          config.extraLuaPackages;
+
       neovimConfig = pkgs.neovimUtils.makeNeovimConfig (
         {
+          inherit extraLuaPackages;
           inherit (config)
             extraPython3Packages
-            extraLuaPackages
             viAlias
             vimAlias
             withRuby
--- a/modules/top-level/plugins/byte-compile-lua-lib.nix
+++ b/modules/top-level/plugins/byte-compile-lua-lib.nix
@@ -9,20 +9,35 @@ let
   builders = lib.nixvim.builders.withPkgs pkgs;
   inherit (import ./utils.nix lib) mapNormalizedPlugins;
 
+  luaPackages = pkgs.neovim-unwrapped.lua.pkgs;
+
+  # Applies a function to the derivation, but only if it's a lua module,
+  # otherwise returns it as is
+  mapLuaModule = f: drv: if luaPackages.luaLib.hasLuaModule drv then f drv else drv;
+
   # Byte-compile only lua dependencies
   byteCompileDeps =
     drv:
-    drv.overrideAttrs (
-      prev:
-      lib.optionalAttrs (prev ? propagatedBuildInputs) {
-        propagatedBuildInputs = map byteCompile prev.propagatedBuildInputs;
-      }
-    );
+    lib.pipe drv [
+      (
+        drv:
+        if drv.dependencies or [ ] != [ ] then
+          drv.overrideAttrs { dependencies = map byteCompileDeps drv.dependencies; }
+        else
+          drv
+      )
+      (
+        drv:
+        if drv.propagatedBuildInputs or [ ] != [ ] then
+          drv.overrideAttrs { propagatedBuildInputs = map byteCompile drv.propagatedBuildInputs; }
+        else
+          drv
+      )
+      # 'toLuaModule' updates requiredLuaModules attr
+      # It works without it, but update it anyway for consistency
+      (mapLuaModule luaPackages.toLuaModule)
+    ];
   # Byte-compile derivation (but only if it's a lua module) and its dependencies
-  byteCompile =
-    drv:
-    byteCompileDeps (
-      if pkgs.luaPackages.luaLib.hasLuaModule drv then builders.byteCompileLuaDrv drv else drv
-    );
+  byteCompile = drv: byteCompileDeps (mapLuaModule builders.byteCompileLuaDrv drv);
 in
 mapNormalizedPlugins byteCompileDeps
--- a/modules/top-level/plugins/byte-compile-lua-lib.nix
+++ b/modules/top-level/plugins/byte-compile-lua-lib.nix
@@ -1,0 +1,28 @@
+/*
+  Byte compiling of lua dependencies of normalized plugin list
+
+  Inputs: List of normalized plugins
+  Outputs: List of normalized plugins with all the propagatedBuildInputs byte-compiled
+*/
+{ lib, pkgs }:
+let
+  builders = lib.nixvim.builders.withPkgs pkgs;
+  inherit (import ./utils.nix lib) mapNormalizedPlugins;
+
+  # Byte-compile only lua dependencies
+  byteCompileDeps =
+    drv:
+    drv.overrideAttrs (
+      prev:
+      lib.optionalAttrs (prev ? propagatedBuildInputs) {
+        propagatedBuildInputs = map byteCompile prev.propagatedBuildInputs;
+      }
+    );
+  # Byte-compile derivation (but only if it's a lua module) and its dependencies
+  byteCompile =
+    drv:
+    byteCompileDeps (
+      if pkgs.luaPackages.luaLib.hasLuaModule drv then builders.byteCompileLuaDrv drv else drv
+    );
+in
+mapNormalizedPlugins byteCompileDeps
--- a/modules/top-level/plugins/byte-compile-lua-lib.nix
+++ b/modules/top-level/plugins/byte-compile-lua-lib.nix
@@ -1,10 +1,14 @@
 # Utilities for byte compiling of lua dependencies
-{ lib, pkgs }:
+{
+  lib,
+  pkgs,
+  neovim-unwrapped,
+}:
 let
   builders = lib.nixvim.builders.withPkgs pkgs;
   inherit (import ./utils.nix lib) mapNormalizedPlugins;
 
-  luaPackages = pkgs.neovim-unwrapped.lua.pkgs;
+  luaPackages = neovim-unwrapped.lua.pkgs;
 
   # Applies a function to the derivation, but only if it's a lua module,
   # otherwise returns it as is
--- a/modules/top-level/plugins/byte-compile-lua-lib.nix
+++ b/modules/top-level/plugins/byte-compile-lua-lib.nix
@@ -1,9 +1,4 @@
-/*
-  Byte compiling of lua dependencies of normalized plugin list
-
-  Inputs: List of normalized plugins
-  Outputs: List of normalized plugins with all the propagatedBuildInputs byte-compiled
-*/
+# Utilities for byte compiling of lua dependencies
 { lib, pkgs }:
 let
   builders = lib.nixvim.builders.withPkgs pkgs;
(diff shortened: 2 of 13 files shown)


tests/utils/plugin-stubs.nix
```diff
@@ -0,0 +1,364 @@
+/*
+  This file includes stub plugins with dependencies of various types.
+  It is primarily used in tests for the performance module.
+*/
+{ lib, pkgs }:
```
MattSturgeon, line 5: Better to use the `callPackages` pattern, IMO:

```suggestion
{
  lib,
  runCommand,
  neovim-unwrapped,
  neovimUtils,
  vimUtils, 
}:
```

Then import with `pkgs.callPackages ./utils/plugin-stubs.nix { }`

modules/top-level/plugins/byte-compile-lua-lib.nix
```diff
@@ -9,20 +9,32 @@ let
...
+            dependencies = map byteCompileDeps prev.dependencies;
+          }
+          // lib.optionalAttrs (prev ? propagatedBuildInputs) {
+            propagatedBuildInputs = map byteCompile prev.propagatedBuildInputs;
+          }
+        )
+      )
+      # 'toLuaModule' updates requiredLuaModules attr
+      # It works without it, but update it anyway for consistency
+      (mapLuaModule (drv: drv.lua.pkgs.toLuaModule drv))
```
MattSturgeon, line 30: Do we need `drv.lua`? If we had another way to reliably get the correct lua pkgs, we could do:

```suggestion
      (mapLuaModule lua.pkgs.toLuaModule)
```

No big deal if this is more correct.

tests/fetch-tests.nix
```diff
@@ -32,7 +32,7 @@ let
       handleEntry =
         name: type:
         let
-          file = "${path}/${name}";
+          file = /${path}/${name};
```
MattSturgeon, line 35: Unrelated change. I'll open a separate PR.

What's the motivation? Does this avoid additional copy-to-store actions or something?

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -1,5 +1,7 @@
-{ pkgs, ... }:
+{ lib, pkgs, ... }:
 let
+  pluginStubs = import ../../../utils/plugin-stubs.nix { inherit lib pkgs; };
```
MattSturgeon, line 3: As per above suggestion

```suggestion
  pluginStubs = pkgs.callPackages ../../../utils/plugin-stubs.nix { };
```

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -325,70 +327,62 @@ in
 
   # performance.byteCompileLua.luaLib for propagatedBuildInputs
   lua-lib-propagated-build-inputs =
-    {
-      config,
-      lib,
-      pkgs,
-      ...
-    }:
+    { config, lib, ... }:
```
MattSturgeon: Why?

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -1,5 +1,7 @@
-{ pkgs, ... }:
+{ lib, pkgs, ... }:
```
MattSturgeon, line 1: If this is just for the `pluginStubs`, we can revert the change if using `callPackages`.

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -325,70 +327,62 @@ in
...
+          # If there are two different derivations with the same name,
+          # then one of them probably isn't byte-compiled.
           {
-            assertion = lib.allUnique (map lib.getName requiredLuaModules);
+            assertion = lib.allUnique (map lib.getName propagatedBuildInputs);
             message = ''
-              Expected requiredLuaModules of all propagatedBuildInputs to have unique names.
-              Got the following derivations: ${builtins.concatStringsSep ", " requiredLuaModules}.
-              One possible reason is that not all dependencies are overridden the same way.
+              Expected propagatedBuildInputs and requiredLuaModules of all plugins to have unique names.
```
MattSturgeon, line 379: Message mentions _both_ `propagatedBuildInputs` _and_ `requiredLuaModules`, but the assertion only checks one. Is that correct?

EDIT: my mistake, I see they're concatenated above. Maybe the name should reflect that? E.g. `compiledLuaInputs` or `compiledLuaModules` or `allInputs` or something?

modules/top-level/output.nix
```diff
@@ -172,9 +172,10 @@ in
   config =
     let
       # Optionally byte compile lua library
+      inherit (import ./plugins/byte-compile-lua-lib.nix { inherit lib pkgs; }) byteCompileLuaPackages;
```
MattSturgeon, line 175: Ideally:

```suggestion
      inherit (pkgs.callPackages ./plugins/byte-compile-lua-lib.nix { }) byteCompileLuaPackages;
```

If it doesn't involve too much out-of-scope refactoring.

EDIT: I'm not sure how well that'd work with accessing `lib.nixvim.builders`, since the `lib` available via `callPackages` may not be the same lib used by nixvim's module eval.

modules/top-level/plugins/byte-compile-lua-lib.nix
```diff
@@ -1,9 +1,4 @@
-/*
-  Byte compiling of lua dependencies of normalized plugin list
-
-  Inputs: List of normalized plugins
-  Outputs: List of normalized plugins with all the propagatedBuildInputs byte-compiled
-*/
+# Byte compiling of lua dependencies
```
MattSturgeon, line 1:
```suggestion
# Utilities for byte compiling of lua dependencies
```

MattSturgeon (commented):
Awesome, thanks for continuing to work on these perf options!

modules/top-level/plugins/byte-compile-lua-lib.nix
```diff
@@ -0,0 +1,45 @@
...
+          prev:
+          # Process also dependencies
+          lib.optionalAttrs (prev ? dependencies) {
+            dependencies = map byteCompileDeps prev.dependencies;
+          }
+          // lib.optionalAttrs (prev ? propagatedBuildInputs) {
+            propagatedBuildInputs = map byteCompile prev.propagatedBuildInputs;
+          }
+        )
+      )
```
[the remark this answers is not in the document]
stasjok: > Q: does byteCompile handle non-lua inputs? dependencies and propagatedBuildInouts aren't guaranteed to exclusively contain lua derivations.

`byteCompileLuaDrv` supports any derivation. It just adds a hook that compiles all .lua files in `$out`. It should work with any derivation. E.g. it works with any kind of plugins. If there is no .lua files in `$out`, then nothing harmful is done. It should work similar to patchelf hook (patches any found binaries) which is executed by default for all drvs.

But this specific function has additional guard: [hasLuaModule](https://github.com/NixOS/nixpkgs/blob/53b40d0c37b9226d5136eb1d5e5b976b3408b37f/pkgs/development/lua-modules/lib.nix#L15). It won't override derivation if `hasLuaModule == false`. Note, that `byteCompile` never called for `dependencies`, so it doesn't override plugin themselves. `byteCompile` is called only on `propagatedBuildInputs`, then `byteCompile` ignores derivation itself if it doesn't `hasLuaModule`, but its `propagatedBuildInputs` are processed anyway.

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -321,6 +322,77 @@ in
...
+        test_lualib_file(argparse("test").parse, true)
+      '';
+
+      assertions =
+        let
+          requiredLuaModules = lib.pipe config.extraPlugins [
+            (builtins.catAttrs "requiredLuaModules")
+            builtins.concatLists
+            lib.unique
+          ];
```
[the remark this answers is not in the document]
stasjok: There was a flaw in this code: `config.extraPlugins` _are not updated plugins_! Latest version uses `config.build.plugins`. This value is already normalized. But anyway, this is a test code with a known input, so I don't think it's necessary to handle any possible input.

> Off topic: if we don't already, should we normalise plugins in the extraPlugins option's apply function, so that it always has a consistent shape when reading?

Interesting idea. Not sure if it's worth it though. Looks like this is pretty much the only place where this value is read: https://github.com/nix-community/nixvim/blob/69266437beaac970568d12a46766706c72f361b3/modules/top-level/plugins/default.nix#L47-L51

modules/top-level/plugins/byte-compile-lua-lib.nix
stasjok: Technically, various derivations can have different lua. But it won't work in runtime if the version is different (e.g. neovim is built with luajit, dependency is from lua5_2, import paths won't match). But I think we can take a lua from neovim-unwrapped as I pretty much always do. It can break only in very rare cases if the user for some reason overrode neovim package with another version of lua, but pkgs' neovim is standard luajit. I don't think we should consider this case, since doing this would be bad. On the other thought, nothing should break, because `toLuaModule` changes only passthru attributes, so let's take lua from pkgs' neovim-unwrapped.

modules/top-level/output.nix
stasjok: I prefer callPackage myself, but the latest refactor in #3138 used many imports, so thought that this is a recommended style.

> EDIT: I'm not sure how well that'd work with accessing lib.nixvim.builders, since the lib available via callPackages may not be the same lib used by nixvim's module eval.

callPackage by default sets arguments to the ones from pkgs, so I think the lib would be from pkgs.lib. If you doesn't populate it with your lib, I think it won't work without callPackageWith. I'll check.

MattSturgeon: > but the latest refactor in #3138 used many imports, so thought that this is a recommended style.

Generally we're moving towards using `callPackage`, `callPackages`, or custom callers like our `callTest` or `callLib` implementations of `callPackageWith`. Sometimes this gets overlooked even in new code, though :grin: 

> callPackage [and callPackages] by default sets arguments to the ones from pkgs, so I think the lib would be from pkgs.lib.

Correct

> If nixvim doesn't populate it with your nixvim's lib, I think it won't work without callPackageWith.

Kinda.

You don't necessarily need to set custom `autoArgs` via `callPackagesWith`, because you can also set specify args explicitly. E.g.:

```nix
pkgs.callPackages ./foo { bar = someOverride; }

# AKA
lib.callPackagesWith pkgs ./foo { bar = someOverride; }
```

The question is whether its worth it at that point.

tests/test-sources/modules/performance/byte-compile-lua.nix
MattSturgeon: Looks like the removed `pkgs` module arg was shadowing the `pkgs` test-file arg?

stasjok: > Looks like the removed pkgs module arg was shadowing the pkgs test-file arg?
No, I'm sure that I removed it after it became unused. Look at the commit:
https://github.com/nix-community/nixvim/pull/3314/commits/237d659c55e075c1274de9aa6b440c078f1a149c
After this there is no pkg usage in this test.

MattSturgeon: Ah, thank's for clarifying. Hard to spot that kinda thing without an LSP!

tests/test-sources/modules/performance/byte-compile-lua.nix
stasjok: I noticed a usage of `pkgs.lib.genAttrs` later on, so I'll leave a lib argument here. When I wrote this initially, lib wasn't passed to tests, but now it is.

stasjok: I want an opinion though: is it better to have a lib, pkgs on every test where it is used, or one top level is enough? Looking more closely, I noticed, that after this change I started to use a top-level lib in one of the tests. It worked and I didn't notice that. Should I care about it? So I should either remove a lib from test functions (and use a top-level one), or add a lib to every test where it's used. Which is better?

MattSturgeon: In practice, there should be very little difference.

The underlying `lib` and `pkgs` passed to test files _should_ be the same `lib` and `pkgs` available as module args within the test-case configurations.

However looking at the code, that isn't the case(!)

[`fetchTests`](https://github.com/nix-community/nixvim/blob/69266437beaac970568d12a46766706c72f361b3/tests/main.nix#L12) uses the `pkgs` instantiated by flake-parts, while the test-case configurations uses a [`pkgsForTest`](https://github.com/nix-community/nixvim/blob/69266437beaac970568d12a46766706c72f361b3/tests/main.nix#L16-L19) instance.

Still, in practice it usually won't matter, other than the small performance hit of having two pkgs instances, which likely can't be avoided _anyway_.

modules/top-level/plugins/byte-compile-lua-lib.nix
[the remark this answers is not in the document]
stasjok: It looks like `propagatedBuildInputs` attr always present. Empty `mkDerivation` creates a derivation with this attr == []. So I think in practice `overrideAttrs` can be avoided only with something like
```nix
if drv.propagatedBuildInputs or [] != [] then
```

Another variant:

```nix
      (
        drv:
        let
          newAttrs =
            lib.optionalAttrs (drv.dependencies or [ ] != [ ]) {
              dependencies = map byteCompileDeps drv.dependencies;
            }
            // lib.optionalAttrs (drv.propagatedBuildInputs or [ ] != [ ]) {
              propagatedBuildInputs = map byteCompile drv.propagatedBuildInputs;
            };
        in
        if newAttrs != { } then drv.overrideAttrs newAttrs else drv
      )
```

But I'm not sure if this is an improvement, because of all these comparisons. So I'm inclined to leave this as is.

Maybe I'll just change `(prev ? dependencies)` to `(drv.dependencies or [ ] != [ ])`. This should be slightly faster since the map won't be called.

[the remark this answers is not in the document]
stasjok: On the second thought

```nix
      (
        drv:
        if drv.dependencies or [ ] != [ ] then
          drv.overrideAttrs { dependencies = map byteCompileDeps drv.dependencies; }
        else
          drv
      )
      (
        drv:
        if drv.propagatedBuildInputs or [ ] != [ ] then
          drv.overrideAttrs { propagatedBuildInputs = map byteCompile drv.propagatedBuildInputs; }
        else
          drv
      )
```
is also readable. So let's be it.

modules/top-level/output.nix
stasjok: I've added a separate commit on top that changes all usages of import in top-level modules to `callPackage .. (inherit lib;)`. See if it's better. If not, I'll drop the commit. I didn't change `inherit (import ./utils.nix lib)` though.

stasjok:
I think all comments were addressed. I rebased it on top of c76112e47789f6ce74041f25bfbb99a0b191d593.

MattSturgeon (approved):
Awesome! The diff looks good to me, thanks again

MattSturgeon:
@mergifyio queue

PR #3317, opened 2025-05-12
Add explanation on follow

fzakaria:
The use of `follows` in a flake can cause issues even if `useGlobalPackages` is set.

Highlight this in the doc.


MattSturgeon (approved):
Thanks!

We always appreciate improvements to the docs! 🚀🎉

I've been meaning to flesh out the docs surrounding pkgs, inputs, follows, etc for a long time now.

Maybe it could do with a dedicated user guide?

Maybe the standalone examples should be using `inputs.nixpkgs.follows = "nixvim/nixpkgs"` to avoid having multiple instances of nixpkgs unnecessarily?

Etc

GaetanLepage (approved):

fzakaria:
@MattSturgeon i'm slowly making my way through it -- I can contribute as far as I get :)
(I like contributing upstream if changes are welcomed!)

PR #3321, opened 2025-05-13
modules/lsp/onAttach: fix `bufnr` and document `event` arg

### LINKED ISSUE #3319: [BUG] LSP module onAttach
| Module       | `lsp`       | 
| Nixpkgs      | `unstable`  | 
| Home Manager | `not used` | 

I'm using a standalone flake

## Description

In the lsp module we have this:
https://github.com/nix-community/nixvim/blob/ff0ccdf572ad6700a2a29a82cc5d17db29708988/modules/lsp/on-attach.nix#L28-L47
If I understand correctly, when using lsp.onAttach, we are allowed to refer to client and bufnr in the Lua code. However, I believe the assignment:
```
local bufnr = args.bufnr
```
is incorrect and should instead be:
```
local bufnr = args.buf
````

### Reasoning

While testing a code that relied on bufnr, I encountered an error (sorry, I don't have the exact log anymore), which disappeared once I changed the code to use args.buf directly. The minimal reproducible example below reproduces the same error.
if you try using the minimal example below, you I get this error message:

### Error Message

When using the current implementation:

```
Error detected while processing LspAttach Autocommands for "*":                                         
Error executing lua callback: /nix/store/mf79j9pggfgjacsfynplb1ls5lvvsg92-init.lua:690: attempt to concatenate local 'bufnr' (a nil value)
```

If I remove the bufnr usage and print args.buf instead, the output is correct:

> args.buf number is: 2

This confirms that args.buf is correctly populated while bufnr is nil.

Unless I’m misunderstanding something (which is very possible), this seems like a small bug in the module.

Apologies if this report is missing anything or doesn't follow conventions. Reporting issues in other people's code is new to me. I'm happy to provide the original code I used  if needed, but the minimal example below should be sufficient to reproduce the problem.

```nix
programs.nixvim = {
  lsp.onAttach = ''
    print("Buffer number is: " .. bufnr)
    print("args.buf number is: " .. args.buf)
  '';
}
```

MattSturgeon:
Fixes #3319

See [`:h event-args`](https://neovim.io/doc/user/api.html#event-args) and [`:h LspAttach`](https://neovim.io/doc/user/lsp.html#LspAttach)

I'm unsure whether we should consider this advice from the docs:

> Note: If the LSP server performs dynamic registration, capabilities may be registered any time _after_ `LspAttach`. In that case you may want to handle the "`registerCapability`" event.
>
> Example:
>
> ```lua
> vim.lsp.handlers['client/registerCapability'] = (function(overridden)
>   return function(err, res, ctx)
>     local result = overridden(err, res, ctx)
>     local client = vim.lsp.get_client_by_id(ctx.client_id)
>     if not client then
>       return
>     end
>     -- Call your custom on_attach logic...
>     -- my_on_attach(client, vim.api.nvim_get_current_buf())
>     return result
>   end
> end)(vim.lsp.handlers['client/registerCapability'])
> ```

But that's out of scope for this PR.

### DIFF
--- a/modules/lsp/on-attach.nix
+++ b/modules/lsp/on-attach.nix
@@ -29,14 +29,25 @@ in
       {
         event = "LspAttach";
         group = "nixvim_lsp_on_attach";
+        # `event` is documented in `:h event-args`:
+        #   • id:    (number)     autocommand id
+        #   • event: (string)     name of the triggered event
+        #   • group: (number|nil) autocommand group id, if any
+        #   • file:  (string)     <afile> (not expanded to a full path)
+        #   • match: (string)     <amatch> (expanded to a full path)
+        #   • buf:   (number)     <abuf>
+        #   • data:  (any)        arbitrary data passed from `:h nvim_exec_autocmds()`
+        #                         see `:h LspAttach`
+        #
+        # `:h event-args`: https://neovim.io/doc/user/api.html#event-args
+        # `:h LspAttach`: https://neovim.io/doc/user/lsp.html#LspAttach
         callback = lib.nixvim.mkRaw ''
-          function(args)
+          function(event)
             do
               -- client and bufnr are supplied to the builtin `on_attach` callback,
               -- so make them available in scope for our global `onAttach` impl
-              local client = vim.lsp.get_client_by_id(args.data.client_id)
-              local bufnr = args.bufnr
-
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
LGTM, thanks

MattSturgeon:
@mergify queue

PR #3339, opened 2025-05-17
plugins/copilot-lua: add curl and nodejs dependency

HeitorAugustoLN:


plugins/by-name/copilot-lua/default.nix
```diff
@@ -105,21 +110,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
       };
 
-      copilot_node_command =
-        let
-          inherit (config.plugins.copilot-lua) nodePackage;
-        in
-        lib.nixvim.mkNullOrOption' {
-          type = types.str;
-          default = if nodePackage == null then null else lib.getExe nodePackage;
-          defaultText = lib.literalMD "`lib.getExe nodePackage` if `nodePackage` is not null, otherwise `null`";
```
MattSturgeon, line 115: This old default was explicitly getting the package's exe path. The new approach is to install the package onto the vim wrapper's PATH.

I assume this is seamless from the user's perspective?

MattSturgeon (approved):
Thanks!

plugins/by-name/copilot-lua/default.nix
HeitorAugustoLN: Yes, it works the same as before

MattSturgeon:
@mergifyio queue

PR #3351, opened 2025-05-22
update-scripts: add "version-info" and run it in update CI

MattSturgeon:
- **update-scripts: get nixpkgs using flak-compat**
- **update-scripts/version-info: init**
- **update-scripts: move update logic from CI to dedicated script**
- **update-scripts/update: write version-info during update**
- **version-info: init**


GaetanLepage (approved):
Looks good overall. Is this to later use `nixvimOlder` & co?

MattSturgeon:
> Is this to later use `nixvimOlder` & co?

I've ran into wanting version info a few times. Sometimes it'd be acceptable to use `lib.trivial.release`, but other times we need to know the "definitely not modified by a `follows`" info.


I didn't want this PR to grow into also implementing various use-cases for version info.

Some uses include minor improvements to the docs, throwing assertions when the wrong nixpkgs channel is used, improving nixvim's "branching off a release" process, etc.

There's probably other version info worth tracking in the `version-info.toml` file too; or alternatively introducing a `lib.nixvim.version` that has a mixture of static and dynamic values? Either way, that's future work.

MattSturgeon:
@mergifyio queue

PR #3366, opened 2025-05-23
docs/mdbook: install directly to `$out`

MattSturgeon:
This fixes a regression introduced by https://github.com/nix-community/nixvim/pull/3349; that PR serves the docs from the root of the docs derivation, however docs are actually installed to `$out/share/doc`.

IMO there's no reason to put a MDBook website in `share/doc/`, so rather than fixing `serve-docs`' `--chdir` I've instead changed the docs to install to the root of `$out`.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -110,7 +110,7 @@ jobs:
 
             # Copy the result to the install dir
             mkdir -p "$installDir"
-            cp -r result/share/doc/* "$installDir"
+            cp -r result/* "$installDir"
           }
 
           # For each version of the docs...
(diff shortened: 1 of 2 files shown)


GaetanLepage (approved):
Good idea indeed.

MattSturgeon:
@mergifyio queue

MattSturgeon:
Just realised, this will need backporting because the CI needs to copy the docs from the old branch's `docs` derivation too. So it is currently copying _those_ from the wrong directory.

However I don't think we can backport without also backporting the changes to `serve-docs`

MattSturgeon:

PR #3368, opened 2025-05-23
plugins/mini-ai: init

HeitorAugustoLN:
This pull request introduces a new Neovim plugin module for the `mini.ai`, along with corresponding test cases.

Doing it for each `mini.nvim` module, so users can use them separately if they want to.


MattSturgeon:
Should we be breaking up the existing `mini` plugin module into smaller `mini.foo` modules?

khaneliman (approved):

PR #3378, opened 2025-05-24
plugins/mini-icons: init

HeitorAugustoLN:
This pull request introduces a new Neovim plugin module for the `mini.icons`, along with corresponding test cases.

### DIFF
--- a/plugins/by-name/mini-icons/default.nix
+++ b/plugins/by-name/mini-icons/default.nix
@@ -1,0 +1,42 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "mini-icons";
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):
Thanks for another great looking PR. LGTM!

This should simplify a lot of icon-related warnings treewide once they're updated to refer to this module instead of `plugins.mini`.

I'm still unsure how I feel about having duplicated effort for `plugins.mini` _and_ `plugins.mini-foo`, but if you're taking on the maintenance burden then I feel better about it :grin: 

I'd like to see what other @nix-community/nixvim maintainers think of splitting up the mini plugin, but if no one comments I'll start reviewing/merging this and your other PRs :rocket:

PR #3382, opened 2025-05-25
update-scripts/version-info: get channel status for supported versions

MattSturgeon:
## Motivation

This will be useful for simplifying the docs CI workflow and reducing the manual process of adding/dropping stable versions.

Having access to the channel's "status" (beta, stable, deprecated) will allow us to show the same info that NixOS Search does in _our_ docs.

We could also use it for the "update-other" workflow, although currently we don't need to checkout the repo in that workflow. That said, the "update-other" workflow _could_ fetch the individual file from githubusercontent and use that to build its job matrix.

I'd also like to add a "notification" of sorts to the main "update" CI workflow when the main branch transitions from one release to another, or when the available stable versions changes. This will be useful as a reminder to branch off the new stable version.

Eventually, we could probably even automate the branching off process, although maybe updating the flake inputs on the new branch and updating references in the `main` branch's docs should remain manual?

## Implementation

The `version-info` script, which is run by the `update` CI workflow, has been extended to also call a new "supported-versions" script.

The "supported-versions" script takes NixOS/infra's [`channels.nix`](https://raw.githubusercontent.com/NixOS/infra/refs/heads/main/channels.nix) file as input and spits out a TOML representation of the bits we care about in nixvim.

```toml
[versions."24.11"]
branch = "nixos-24.11"
channel = "nixos-24.11"
status = "deprecated"

[versions."25.05"]
branch = "nixos-25.05"
channel = "nixos-25.05"
status = "stable"

[versions."25.11"]
branch = "main"
channel = "nixpkgs-unstable"
status = "rolling"
```


MattSturgeon:
I think a lot of the impurity could be solved by refactoring `update-scripts` into a standalone sub-flake.

It's kinda ugly having to eval different nix files in different ways, depending on the scenario.

However, I don't want to do such a refactor as part of this PR, and I'd prefer not to block this PR on it being done as the merge conflicts would be a nightmare.

MattSturgeon:
A use-case I have for this is generating a version matrix in the `update-other` workflow:

```yml
jobs:
  prepare-matrix:
    runs-on: ubuntu-24.04-arm
    outputs:
      matrix: ${{ steps.set-matrix.outputs.result }}
    steps:
      - name: Get version info
        env:
          info_url: https://raw.githubusercontent.com/${{ github.repository }}/refs/heads/${{ github.ref_name }}/version-info.toml
          auth_header: Authorization: Bearer ${{ github.token }}
        run: |
          curl "$info_url" --header "$auth_header" > version-info.toml
      - name: Generate version matrix
        id: set-matrix
        uses: mikefarah/yq@master
        with:
          cmd: yq '.supported-versions' version-info.toml
```

We could also do something similar in the docs workflow when deciding which docs branches to build...

update-scripts/version-info/default.nix
```diff
@@ -1,36 +1,42 @@
...
+writeShellApplication {
+  name = "version-info";
 
-  info = {
-    inherit (lib.trivial) release;
-    nixpkgs_rev = lib.trivial.revisionWithDefault (throw "nixpkgs revision not available");
-    unstable = lib.strings.hasSuffix "-unstable" nixpkgsLock.ref;
+  runtimeEnv = {
+    NIX_CONFIG = ''
+      experimental-features = nix-command flakes pipe-operators
```
khaneliman, line 16: > pipe-operators

Reminds me I haven't tried them in a while... last time I enabled this I saw tons of weird warnings in the logs.

update-scripts/version-info/supported-versions.nix
```diff
@@ -0,0 +1,62 @@
...
+    |> builtins.filter (entry: builtins.elem entry.status supported)
+    |> builtins.map (entry: {
+      name = entry.channel |> builtins.match "nixos-(.+)" |> builtins.head;
+      value = {
+        inherit (entry) channel status;
+        # Currently, Nixvim stable branches match NixOS channel names
+        branch = entry.channel;
+      };
+    })
+    |> builtins.listToAttrs;
```
khaneliman, line 30: I agree this lib.pipe and pipe operator syntax is easier to logically follow without getting confused.

version-info.toml
```diff
@@ -3,3 +3,18 @@
...
+
+[versions."25.05"]
+branch = "nixos-25.05"
+channel = "nixos-25.05"
+status = "stable"
+
+[versions."25.11"]
+branch = "main"
+channel = "nixpkgs-unstable"
+status = "rolling"
```
khaneliman, line 20: Are we going be to be using this information for anything? I see we have the old info at the top about the specific revision we're on. Does it make sense to consolidate these into one section for the `current` rolling version information.

```toml
[versions."25.11"]
branch = "main"
channel = "nixpkgs-unstable"
status = "rolling"
rev = "fe51d34885f7b5e3e7b59572796e1bcb427eccb1"
```

and just get rid of the stuff at the top? 

Obviously, would be a follow up PR for consolidation since we'd have to touch whatever is using that current data.

khaneliman (commented):
Love the concept

>I'd also like to add a "notification" of sorts to the main "update" CI workflow when the main branch transitions from one release to another, or when the available stable versions changes. This will be useful as a reminder to branch off the new stable version.

This would be very nice indeed. Having a real time update when branch off occurs instead of manually having to check. 

>Eventually, we could probably even automate the branching off process, although maybe updating the flake inputs on the new branch and updating references in the main branch's docs should remain manual?

This is where my head went reading that, too. I think it could be automated. I don't know how much needs to be manual when we're just branching and updating references.

version-info.toml
MattSturgeon: I'm not 100% sure until we start trying to use the data. But there is a subtle difference.

The version info at the top comes from our pinned nixpkgs revision, while the version info in this array relates to the latest "channels" info from NixOS/infra. So there's potential for it to be _slightly_ out of sync.

EDIT: it could probably be combined if we had some assertions, however we still need a way to know which of these listed versions _this_ nixvim branch is on. Like a "current version" top-level attr.

MattSturgeon:
@mergifyio queue

update-scripts/update.nix
```diff
@@ -43,7 +43,7 @@ writeShellApplication {
 
     versionInfo() {
       nix-build ./update-scripts -A version-info
-      install -m 644 -T "$(realpath result)" ./version-info.toml
+      ./result
```
MattSturgeon, line 46: Oops, this is a directory. The script is actually at `./result/bin/version-info`

PR #3387, opened 2025-05-25
Use homeModules as flake output name for Home Manager modules

Aehmlo:
This PR contains a handful of changes centered around the flake output for Home Manager modules. My understanding is that, recently, [convention](https://github.com/nix-community/home-manager/pull/6406) [has](https://github.com/Mic92/sops-nix/pull/795) [been](https://github.com/catppuccin/nix/pull/562) [converging](https://github.com/nix-community/stylix/pull/1267) on using [`homeModules` as the name of the flake output](https://github.com/nix-community/home-manager/blob/901f8fef7f349cf8a8e97b3230b22fd592df9160/flake-module.nix#L17C7-L43C9) (as this is closer to `homeConfigurations` and more in line with `nixosModules`/`nixosConfigurations`), which is the focus of this PR.

This PR contains a few commits. The first simply adds `homeModules` as an alias of `homeManagerModules`. The second migrates documentation and internal uses of `homeManagerModules` to use `homeModules` instead, and the third inverts the first by defining `homeManagerModules` in terms of `homeModules`, treating it as a backward-compatibility layer. I'm not sure what the stance of the project is on this issue, but while I'd personally like to see all three land, just having `homeModules` available as an option (i.e., just the first commit here) would satisfy me.

### DIFF
--- a/flake/wrappers.nix
+++ b/flake/wrappers.nix
@@ -20,8 +20,14 @@
       nixvim = import ../wrappers/nixos.nix self;
       default = self.nixosModules.nixvim;
     };
-    # For backward compatibility
-    homeManagerModules = self.homeModules;
+    # Alias for backward compatibility
+    # Added 2025-05-25 in https://github.com/nix-community/nixvim/pull/3387
+    homeManagerModules =
+      let
+        cond = lib.trivial.oldestSupportedReleaseIsAtLeast 2505;
+        msg = "nixvim: flake output `homeManagerModules` has been renamed to `homeModules`.";
+      in
+      lib.warnIf cond msg self.homeModules;
     homeModules = {
       nixvim = import ../wrappers/hm.nix self;
       default = self.homeModules.nixvim;


flake/wrappers.nix
```diff
@@ -20,9 +20,11 @@
       nixvim = import ../wrappers/nixos.nix self;
       default = self.nixosModules.nixvim;
     };
-    homeManagerModules = {
+    # For backward compatibility
+    homeManagerModules = self.homeModules;
```
MattSturgeon, line 24: Let's add a warning so users know they're using an alias.

```suggestion
    # Alias for backward compatibility
    # Added 2025-05-25 in https://github.com/nix-community/nixvim/pull/3387
    homeManagerModules = builtins.warn "nixvim: flake output `homeManagerModupes` has been renamed to `homeModules`." self.homeModules;
```

We'll probably backport the alias _without_ the warning though.

We _could_ do:

```nix
# Alias for backward compatibility
# Added 2025-05-25 in https://github.com/nix-community/nixvim/pull/3387
homeManagerModules = lib.warnIf (lib.trivial.oldestSupportedReleaseIsAtLeast 2505) "nixvim: flake output `homeManagerModupes` has been renamed to `homeModules`." self.homeModules;
```

, to have the warning show up on branches new enough that 25.05 is the "oldest supported version".

`oldestSupportedReleaseIsAtLeast 2505` will be true on `main` in a few weeks, when 24.11 transitions from "deprecated" to "unmaintained". It will _always_ be `false` on the nixos-25.05 branch though, as `oldestSupportedRelease` is only bumped on nixpkgs `master`, not on the stable channels.

But it's also fine to have a static warning that we manually remove in the backport.

MattSturgeon (approved):
Thanks for picking this up!

I didn't realize we were still using `homeManagerModules` even after I personally pushed to standardize on using `homeModules` upstream :sweat_smile:

Diff LGTM, but it'd be nice to add a warning as part of this PR if you're able.

Aehmlo:
> Thanks for picking this up!

Happy to help! Glad the project seems amenable to this change.

> I didn't realize we were still using `homeManagerModules` even after I personally pushed to standardize on using `homeModules` upstream 😅

Thank you for helping make this standardization process happen! :)

> Diff LGTM, but it'd be nice to add a warning as part of this PR if you're able.

Done in a follow-up commit. I refrained from squashing/rebasing, since I know GitHub doesn't play nicely with that, but happy to clean up the history a little bit if desired before merging.

MattSturgeon:
@mergifyio backport nixos-25.05

MattSturgeon:
@mergifyio queue

PR #3394, opened 2025-05-25
plugins/mini-trailspace: init

HeitorAugustoLN:
This pull request introduces a new Neovim plugin module for the `mini.trailspace`, along with corresponding test cases.

### DIFF
--- a/plugins/by-name/mini-trailspace/default.nix
+++ b/plugins/by-name/mini-trailspace/default.nix
@@ -1,0 +1,12 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "mini-trailspace";
+  moduleName = "mini.trailspace";
+  packPathName = "mini.trailspace";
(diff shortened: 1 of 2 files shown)


plugins/by-name/mini-trailspace/default.nix
```diff
@@ -0,0 +1,22 @@
...
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  settingsOptions = {
+    only_in_normal_buffers = defaultNullOpts.mkBool true ''
+      Highlight only in normal buffers (ones with empty 'buftype').
+      This is useful to not show trailing whitespace where it usually doesn't matter.
+    '';
+  };
+
```
GaetanLepage: We now avoid to declare any setting option explicitly. This reduces the maintenance burden.
Just keep the `settingsExample` statement.

khaneliman (approved):

PR #3407, opened 2025-05-29
ci: check repo before running publish jobs

MattSturgeon:
- **ci: check repo before running publish jobs**
- **ci/flakehub-publish-rolling: use context for repo name**
- **ci/flakestry-publish-rolling: add workflow name**

### DIFF
--- a/.github/workflows/flakehub-publish-rolling.yml
+++ b/.github/workflows/flakehub-publish-rolling.yml
@@ -5,6 +5,7 @@ on:
       - "main"
 jobs:
   flakehub-publish:
+    if: github.repository == 'nix-community/nixvim'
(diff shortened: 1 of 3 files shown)


.github/workflows/flakehub-publish-rolling.yml
```diff
@@ -5,6 +5,7 @@ on:
       - "main"
 jobs:
   flakehub-publish:
+    if: github.repository == 'nix-community/nixvim'
```
MattSturgeon, line 8: It would be better practice to check necessary variables are present instead of checking the repo or owner. That approach would allow deliberate testing of forks while still preventing accidental runs.

This PR is still good to merge if we don't have the energy to find a better solution right now.

Issue #3238, opened 2025-04-28
[RFC] Stripping down the test set (i.e. make the CI go brrrrr)


GaetanLepage:
## Motivation

Our testing strategy is quite exhaustive, but it is large and takes time to evaluate and run (especially from scratch).
Fortunately, there is a lot of redundancy, and we surely can lighten it meaningfully.

## General idea

- Establish, design and implement a criteria for marking a test as _platform-agnostic_. Such tests would not need to run on all four platforms.
- Refactor our test set to maximize the number of such tests:
    - Remove the _platform-specificy_ from most tests,
    - Create extra (but way fewer) tests to account for what has been removed.

### Which tests are targetted by this RFC

Let us decompose or our huge test set $T$ in two categories:
- **$T_\text{custom}$ ("top-level" tests):** Everything that is not automatically generated from `test-sources/`.
- **$T_\text{config}$ ("config" tests):** Everything under `test-sources/` that is automatically processed as a configuration test (i.e., we run `nvim` with this config)

Let's leave $T_\text{custom}$ aside. There aren't too many tests in this group and it makes sense to have them run on all platforms.

Let's instead focus on $T_\text{config}$.

### _Platform-agnostic_ vs. _Platform-specific_ tests
A nixvim configuration is said to be **platform agnostic (PA)** if it has no platform-specific behavior. I.e.:

---

**Definition:** A configuration where `extraPackages`, `extraPlugins`, `extraPythonPackages`... are empty.
(Note: this definition is probably incomplete)
Conversely, tests that are not PA, will be refered to as **platform specific (PS)**.

---

Of course such tests still rely on neovim, which is platform-specific, but we can make the assumption that neovim is working on all platforms (it is extensively tested elsewhere).
Also, nothing prevents the tests from having "platform-specific" lua code in the `extraConfigLua` option, which is also technically PS.
I will assume that this probably does not exist or at least has a very low risk.
These purely platform-agnostic tests can be limited to running exclusively on our primary platform (most likely `x86_64-linux`).

### Factorization of the test set

Currently, most of our tests are technically **PS**.
However, the main idea of this proposal relies on two properties:
1. Each test $t \in T_c$ can be split in two sub-tests: $t_{PS}$ (a platform specific component) and $t_{PA}$ (a platform agnostic component).
2. We can easily build _few_ macro tests that **factorize** all $t_{PS}$ components of the tests in $T_c$.

I will illustrate this with the example of plugin tests.
[...]

PR #3410, opened 2025-05-29
flake: add `ci.buildbot` output

MattSturgeon:
### Background

I've long complained that buildbot is too tightly coupled to our flake's `checks` output; the same one used by `nix flake check`. Being unable to customise what buildbot builds has been a pain point for a while.

Turns out around seven months ago buildbot added a per-repo config file in https://github.com/nix-community/buildbot-nix/pull/318.

The available options are documented on its readme, [here](https://github.com/nix-community/buildbot-nix#per-repository-configuration). Currently only `lock_file` and `attribute`, but the latter enables this PR!

### Core change

This PR adds an explicit `ci.buildbot` output to the flake, which we configure buildbot to use instead of the default `checks` output via a `buildbot-nix.toml` file.

This allows us to specify a different set of tests for `nix flake check` and buildbot.

### Additional changes

This PR is **not** focused on _fully_ taking advantage of the new separation, however it does make some minor tweaks:

- ~~`packages` are now only built on x86_64-linux, but are built by `nix flake check` on all systems~~ :rocket:
  - _Dropped for now due to [feedback](https://github.com/nix-community/nixvim/pull/3410#discussion_r2116552870). Will address docs packages specifically in future work._
- Many tests imported by `tests/default.nix` are also only built on x86_64-linux.
  - "Platform" tests and "module" tests (generated by `fetch-tests`) are still built on all platforms.

Without this PR buildbot had 288 builds, with just tests it has 205, with tests and packages it has 214, with tests (mostly single system) and packages (all systems), it has 235.

### Future work:

- Some builds may be better off done by a GitHub Actions workflow instead of buildbot. This will further reduce the load on buildbot while also improving the UX of how other tests get displayed in PR status checks.
  - The `ci.buildbot` option/output has been designed to accommodate `ci.*` sibling options, such as [GitHub Actions job matrices](https://github.com/nix-community/nixvim/blob/1b25700ae36df1da162020abf3a39416235815ef/.github/workflows/check.yml#L26-L31).
- We can further iterate on deciding _which_ tests should/shouldn't be built by CI and/or `nix flake check`
  - E.g., do we really need `packages` to be built by `nix flake check`?

### Related:

- https://github.com/nix-community/nixvim/issues/3238

### DIFF
--- a/flake/dev/package-tests.nix
+++ b/flake/dev/package-tests.nix
@@ -4,5 +4,9 @@
     {
       # Test that all packages build fine when running `nix flake check`.
       checks = config.packages;
+
+      # Test that all packages build when running buildbot
+      # TODO: only test the docs on x86_64-linux
+      ci.buildbot = config.packages;
     };
 }
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -5,6 +5,7 @@
   linkFarm,
   self, # The flake instance
   system ? pkgs.stdenv.hostPlatform.system,
+  allSystems ? true,
 }:
 let
   autoArgs = pkgs // {
@@ -31,8 +32,15 @@ let
   callTests = lib.callPackagesWith autoArgs;
 
   selfPackages = self.packages.${system};
+
+  # For tests that CI should only build on one system,
+  # This is true when on that system.
+  #
+  # TODO: consider refactoring tests/default.nix so that some tests are
+  # defined by it, while others are defined elsewhere...
+  buildForThisSystem = allSystems || system == "x86_64-linux";
 in
-{
+lib.optionalAttrs buildForThisSystem {
   extra-args-tests = callTest ./extra-args.nix { };
   extend = callTest ./extend.nix { };
   extra-files = callTest ./extra-files.nix { };
@@ -47,10 +55,12 @@ in
   lsp-all-servers = callTest ./lsp-servers.nix { };
 }
 # Expose some tests from the docs as flake-checks too
-// lib.optionalAttrs (selfPackages ? docs) {
+// lib.optionalAttrs (selfPackages ? docs && buildForThisSystem) {
   # Individual tests can be run using: nix build .#docs.user-configs.tests.<test>
   docs-user-configs = linkFarm "user-configs-tests" selfPackages.docs.user-configs.tests;
 }
+
+# These are always built on all systems, even when `allSystems = false`
 // callTests ./platforms { }
 # Tests generated from ./test-sources
 # Grouped as a number of link-farms in the form { test-1, test-2, ... test-N }
--- a/buildbot-nix.toml
+++ b/buildbot-nix.toml
@@ -1,0 +1,1 @@
+attribute = "ci.buildbot"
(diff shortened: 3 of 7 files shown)


MattSturgeon:
I forget what `error: writing to file: Broken pipe` means. Does that mean that buildbot ran out of memory trying to eval the attribute? I'm able to eval/build these attributes locally, but it _is_ very memory intensive.

Once I get this PR cleaned up and start replacing "dynamic" attrs with "literal"/"explicit" attrs then the eval should be much cheaper again.

tests/default.nix
```diff
@@ -31,8 +32,15 @@ let
   callTests = lib.callPackagesWith autoArgs;
 
   selfPackages = self.packages.${system};
+
+  # For tests that CI should only build on one system,
+  # This is true when on that system.
+  #
+  # TODO: consider refactoring tests/default.nix so that some tests are
+  # defined by it, while others are defined elsewhere...
+  buildForThisSystem = allSystems || system == "x86_64-linux";
```
khaneliman, line 41: Might be nice to make the fallback checking against a list if we want to set it up to support a `whitelist` of systems we always run on. Could help if the CI situation improves for `aarch64-darwin` down the road.

flake/dev/package-tests.nix
```diff
@@ -1,8 +1,12 @@
...
 {
   perSystem =
     { config, ... }:
     {
       # Test that all packages build fine when running `nix flake check`.
       checks = config.packages;
     };
+
+  # Test that all packages build on x86_64-linux when running buildbot
+  flake.ci.buildbot = { inherit (self.packages) x86_64-linux; };
```
khaneliman, line 11: Same idea here, map over the `whitelist` and include those.

MattSturgeon: Do we really need to build our internal packages on multiple systems in CI?

The only _problematic_ package is `docs` (and anything that depends on it); the others could be built on all systems.

But given they're all internal, I'm not currently motivated to come up with a less "quick and dirty" approach.

tests/default.nix
khaneliman: Ah, if these are just platform agnostic then that's fine.

flake/dev/package-tests.nix
khaneliman: I think it makes more sense to validate our internal tooling works for all contributors than the test cases that are platform agnostic.

MattSturgeon: Ok. I'll revert this PR back including all internal packages on all systems, then put some effort into restricting only the `docs` packages to one system in a follow up.

MattSturgeon: Done

khaneliman (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
> Backport failed for `nixos-24.11`, because it was unable to cherry-pick the commit(s).
> 
> Please cherry-pick the changes locally and resolve any conflicts.

Looked into it, there's too many differences between nixvim 24.11 and 25.05 to be worth backporting when we'll be dropping 24.11 support soon anyway.

MattSturgeon:
> Git push to origin failed for nixos-25.05 with exitcode 1

This seems to be caused by the backport workflow running twice. We should be using the same conditions as nixpkgs, but maybe something is wrong?

PR #3414, opened 2025-05-31
plugins/yaml-schema-detect: init

FKouhai:


plugins/by-name/yaml-schema-detect/default.nix
```diff
@@ -0,0 +1,29 @@
...
+  description = ''
+    A Neovim plugin that automatically detects and applies YAML schemas for your YAML files using yaml-language-server (yamlls)
+  '';
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.yaml-schema-detect" [
+      {
+        when = !config.plugins.which-key.enable;
+        message = ''
+          This plugin requires `plugins.which-key` to be enabled
```
MattSturgeon: I don't see this requirement mentioned on the [plugin's README](https://github.com/cwrau/yaml-schema-detect.nvim/).

FKouhai: they do not appear to be there, but when I tried running the tests without it it would complain that it needs which-key

MattSturgeon: It could be an upstream bug; maybe they have which-key integration that forgets to check it is present before importing it?

FKouhai: it's in their [init.lua](https://github.com/cwrau/yaml-schema-detect.nvim/blob/debcb2835685a06d5b22666bb30d2d930e23d68a/lua/yaml-schema-detect/init.lua#L428)

FKouhai: do you think we could merge it? or should I raise an issue upstream about it?(personally speaking im fine having that warning for the users)

MattSturgeon: I've opened https://github.com/cwrau/yaml-schema-detect.nvim/issues/6, so let's give upstream a little time to respond.

I don't mind merging if we add references to that issue, but ideally we'd merge once we know what the correct behaviour is.

FKouhai: thank you Matt <3

FKouhai: I see upstream closed the PR with the additions you mentioned, I guess we can wait for the next round of vimPlugins updates after that I'll rebase and fix the tests, and again thanks Matt :)

FKouhai: bumping this, updated the test removing which-key and removed the warning that mentioned that the plugin requires it, with that I was able to successfully run the tests I'll wait for #3562 to be merged and then I'll rebase and push the commit again to be ready for a review

khaneliman: Want to try rebasing and removing warning?

plugins/by-name/yaml-schema-detect/default.nix
```diff
@@ -0,0 +1,17 @@
...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "yaml-schema-detect";
+  packPathName = "yaml-schema-detect.nvim";
+  package = "yaml-schema-detect-nvim";
+
+  description = ''
+    A Neovim plugin that automatically detects and applies YAML schemas for your YAML files using yaml-language-server (yamlls)
+  '';
+
```
khaneliman, line 17: would be nice to have a `settingsExample`

tests/test-sources/plugins/by-name/yaml-schema-detect/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.yaml-schema-detect.enable = true;
+  };
```
khaneliman, line 4: Add a `defaults` test case, please. 
https://github.com/cwrau/yaml-schema-detect.nvim/blob/main/lua/yaml-schema-detect/init.lua#L427-L434

plugins/by-name/yaml-schema-detect/default.nix
```diff
@@ -0,0 +1,25 @@
...
+  settingsExample = {
+    keymap = {
+      refresh = "<leader>yr";
+      cleanup = "<leader>yc";
+      info = "<leader>yi";
+    };
+  };
+
+  maintainers = [ lib.maintainers.FKouhai ];
```
khaneliman: Don't know if we really have a rule on this, but it's easier to find above the description.

khaneliman (approved):
Thanks, looks better

plugins/by-name/yaml-schema-detect/default.nix
FKouhai: oh sure, I'm always used to have it at the bottom I could move it up no biggie

PR #3418, opened 2025-06-02
ci/update: cleanup & fix re-apply logic now that we use `nixvim-ci[bot]`

MattSturgeon:
#3412 demonstrated that the "re-apply extra commits" logic is broken now that we commit updates as `nixvim-ci[bot]`. The logic is still looking for a base commit authored by `github-actions[bot]`.

Additionally, include some cleanup refactoring for how we get the committer info. This is the same logic used in my personal config and also recently added to home-manager's CI.


khaneliman (approved):

MattSturgeon:
@mergifyio queue

MattSturgeon:
> Git push to origin failed for nixos-25.05 with exitcode 1

The actual error is:

> ``! [remote rejected] backport/nixos-24.11/3418 -> backport/nixos-24.11/3418 (refusing to allow a GitHub App to create or update workflow `.github/workflows/update.yml` without `workflows` permission)``

I.e. we can't backport workflow changes because `nixvim-ci[bot]` doesn't have `workflows: write` permission.

I'm unsure if we want to change this, and if so whether it needs explicit permissions in the `actions/create-github-app-token` step? See [Create a token with specific permissions](https://github.com/actions/create-github-app-token#create-a-token-with-specific-permissions) and [`permission-<permission name>`](https://github.com/actions/create-github-app-token#permission-permission-name).

We probably also need to ask an org owner to add the permission to our app installation? Unless @GaetanLepage is able? This would be a two step process; first [edit the nixvim-ci app](https://github.com/settings/apps/nixvim-ci) itself to "require" the permission, then re-authenticate the [repo installation](https://github.com/settings/apps/nixvim-ci/installations) to "accept" the new permissions.

MattSturgeon:
> I.e. we can't backport workflow changes because `nixvim-ci[bot]` doesn't have `workflows: write` permission.

Hm. I'm not sure why https://github.com/nix-community/nixvim/pull/3423 was successful when this wasn't... Maybe because this adds a "step" with a different id (`user-id` -> `user-info`)?

PR #3422, opened 2025-06-02
plugins/treesitter: don't exclude nvim-treesitter from combining by default

stasjok:
Originally nvim-treesitter was added to standalonePlugins as workaround for https://github.com/NixOS/nixpkgs/pull/321550.
Nvim-treesitter from nixpkgs don't include queries from upstream parsers anymore since https://github.com/NixOS/nixpkgs/pull/344849.

If the user has some other plugin containing treesitter queries, this
change can cause a build failure due to collisions. But since it is
easier to add the plugin to standalonePlugins compared to removing it, I
think this should be the default.


MattSturgeon:
@mergifyio queue

PR #3435, opened 2025-06-04
plugins/treesitter: prevent mkRaw over-injection by removing injection.combined

HeitorAugustoLN:
Fixes the issue I was having with wrong commenting/uncommenting with nixvim injections


khaneliman (approved):
LGTM, tested it fixes the issue reported

PR #3445, opened 2025-06-08
output: add extraPackagesAfter option

sportshead:
Allow users to add packages to the end of `PATH` in the neovim wrapper. This is useful for LSP versions that might need to be overriden based on the environment, e.g. `haskell-language-server` versions provided by a project's devshell.

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -58,7 +58,14 @@ in
     extraPackages = mkOption {
       type = with types; listOf (nullOr package);
       default = [ ];
-      description = "Extra packages to be made available to neovim";
+      description = "Extra packages to be made available to neovim, added to the start of `PATH`";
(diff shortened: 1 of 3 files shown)


MattSturgeon (approved):
Awesome, thank you!

This will definitely be useful.

Maybe we should add some kinda "fallback" boolean-type option for `lsp.servers.*` options in a followup PR that controls whether the package is installed to `--prefix` or `--suffix`?

sportshead:
> Maybe we should add some kinda "fallback" boolean-type option for `lsp.servers.*` options in a followup PR that controls whether the package is installed to `--prefix` or `--suffix`?

That's a great idea, much better than having to set `.package = null` and adding it to `extraPackagesAfter`! I'll see what I can do.

PR #3450, opened 2025-06-09
plugins/lsp: add packageFallback option

sportshead:
> Maybe we should add some kinda "fallback" boolean-type option for lsp.servers.* options in a followup PR that controls whether the package is installed to --prefix or --suffix?

https://github.com/nix-community/nixvim/pull/3445#pullrequestreview-2908615597

Add a `packageFallback` option to allow overriding LSP versions from local PATH.

### DIFF
--- a/plugins/lsp/language-servers/_mk-lsp.nix
+++ b/plugins/lsp/language-servers/_mk-lsp.nix
@@ -55,6 +55,16 @@ in
         lib.nixvim.mkMaybeUnpackagedOption "plugins.lsp.servers.${name}.package" pkgs name
           package;
 
+      packageFallback = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          When enabled, the language server package will be added to the end of the `PATH` _(suffix)_ instead of the beginning _(prefix)_.
+
+          This can be useful if you want local versions of the language server (e.g. from a devshell) to override the nixvim version.
+        '';
+      };
+
       cmd = mkOption {
         type = with types; nullOr (listOf str);
         default =
@@ -148,7 +158,8 @@ in
   };
 
   config = lib.mkIf enabled {
-    extraPackages = [ cfg.package ];
+    extraPackages = lib.optional (!cfg.packageFallback) cfg.package;
+    extraPackagesAfter = lib.optional cfg.packageFallback cfg.package;
 
     plugins.lsp.enabledServers = [
       {
--- a/tests/test-sources/plugins/lsp/_lsp.nix
+++ b/tests/test-sources/plugins/lsp/_lsp.nix
@@ -265,4 +265,58 @@
           }
         ];
     };
+
+  package-fallback =
+    { config, ... }:
+    {
+      test.buildNixvim = false;
+
+      plugins.lsp = {
+        enable = true;
+        servers = {
+          nil_ls = {
+            enable = true;
+
+            packageFallback = true;
+          };
+          rust_analyzer = {
+            enable = true;
+
+            installCargo = true;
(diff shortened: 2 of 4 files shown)


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -55,6 +55,15 @@ in
         lib.nixvim.mkMaybeUnpackagedOption "plugins.lsp.servers.${name}.package" pkgs name
           package;
 
+      packageFallback = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          When enabled, the language server package will be added to the end of `PATH` (--suffix) instead of the beginning (--prefix).
+          This can be useful if you want local versions of the language server (e.g. from a devshell) to override the nixvim version.
```
MattSturgeon, line 64:
```suggestion

          This can be useful if you want local versions of the language server (e.g. from a devshell) to override the nixvim version.
```

Need a fully blank line (`\n\n`) to separate markdown paragraphs, otherwise markdown will join the lines into a single line.

tests/test-sources/plugins/lsp/_lsp.nix
```diff
@@ -265,4 +265,56 @@
           }
         ];
     };
+
+  package-fallback =
```
MattSturgeon, line 269: I think we can set `test.buildNixvim` or `test.runNvim` to false here, unless building/running is actually part of the test? At first glance, it looks like we only care about the assertions, which will be checked even if `test.buildNixvim = false`.

MattSturgeon (commented):
Concept looks good to me. I wonder how it will scale to other places where we install to the PATH, but I think this is _mostly_ useful for LSPs anyway, so I'm not going to think too hard about that yet...

Miner comments on some details below

tests/test-sources/plugins/lsp/_lsp.nix
sportshead: Yep, it only needs the assertions

sportshead:
> Concept looks good to me. I wonder how it will scale to other places where we install to the PATH, but I think this is _mostly_ useful for LSPs anyway, so I'm not going to think too hard about other parts of nixvim yet...

I can't really think about much other use cases for this, apart from maybe if you wanted a different ripgrep binary (why you would want that I'm not sure). In that case it's probably best to just let the user set `.package = null` and manually add it to `extraPackagesAfter` IMO

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -55,6 +55,16 @@ in
         lib.nixvim.mkMaybeUnpackagedOption "plugins.lsp.servers.${name}.package" pkgs name
           package;
 
+      packageFallback = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          When enabled, the language server package will be added to the end of `PATH` (--suffix) instead of the beginning (--prefix).
```
MattSturgeon: Sorry, I should've spotted this earlier.

Not anything major, but if you're including CLI options in markdown they should be marked as code using `` `backticks` ``, like you've already done with `` `PATH` ``.

```suggestion
          When enabled, the language server package will be added to the end of the `PATH` (`--suffix`) instead of the beginning (`--prefix`).
```

However, in this case we don't really need to quote the CLI options. So it's probably better to do this instead:

```suggestion
          When enabled, the language server package will be added to the end of the `PATH` _(suffix)_ instead of the beginning _(prefix)_.
```

GaetanLepage (approved):

PR #3451, opened 2025-06-09
plugins/telescope-zf-native-nvim: init

Ladas552:
Add support for [telescope-zf-native.nvim](https://github.com/natecraddock/telescope-zf-native.nvim), providing native telescope bindings to zf for sorting results.

### DIFF
--- a/plugins/by-name/telescope/extensions/zf-native.nix
+++ b/plugins/by-name/telescope/extensions/zf-native.nix
@@ -1,0 +1,72 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+  mkExtension = import ./_mk-extension.nix;
+in
+mkExtension {
+  name = "zf-native";
+  extensionName = "zf-native";
+  package = "telescope-zf-native-nvim";
+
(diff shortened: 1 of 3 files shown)


khaneliman (approved):
LGTM overall, just not familiar with the `combinePlugins` enough to say about that config

Ladas552:
> LGTM overall, just not familiar with the `combinePlugins` enough to say about that config

I added it because it gives an error without the binaries linked in combinePLugins. idk if there is a better way for this particular plugin

plugins/by-name/telescope/extensions/zf-native.nix
```diff
@@ -0,0 +1,72 @@
...
+      initial_sort = null;
+      smart_case = true;
+    };
+  };
+
+  dependencies = [ "zf" ];
+
+  extraConfig = cfg: {
+    # zf-native shared binaries
+    performance.combinePlugins.pathsToLink = [ "/lib" ];
```
MattSturgeon, line 70: This might need to go into `pathsToLink`'s central/default value, if other plugins also have `lib` directories.

Cc @stasjok

PR #3452, opened 2025-06-09
plugins/lzn-auto-require: init

axelkar:
Includes end-to-end test

https://github.com/horriblename/lzn-auto-require

### DIFF
--- a/plugins/by-name/lzn-auto-require/default.nix
+++ b/plugins/by-name/lzn-auto-require/default.nix
@@ -1,0 +1,16 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lzn-auto-require";
+  description = ''
+    Auto load optional plugins via lua modules with `lz.n`.
+
+    This plugin overrides the `require` function to also search for plugins in
+    `{packpath}/*/opt` and load them using `lz.n`.
+  '';
+
+  hasSettings = false;
+  setup = ".enable";
+  configLocation = lib.mkOrder 5000 "extraConfigLuaPost"; # Register after everything to catch mistates
+
+  maintainers = [ lib.maintainers.axka ];
+}
--- a/tests/test-sources/plugins/by-name/lzn-auto-require/default.nix
+++ b/tests/test-sources/plugins/by-name/lzn-auto-require/default.nix
@@ -1,0 +1,28 @@
+{
+  e2e =
+    { lib, ... }:
+    {
+      plugins.lz-n.enable = true;
+      plugins.lzn-auto-require.enable = true;
+
+      plugins.smart-splits = {
+        enable = true;
+        lazyLoad.settings.lazy = true; # manual lazy-loading
+      };
+
+      extraConfigLuaPost = lib.mkMerge [
+        (lib.mkOrder 4999 ''
+          local success, _ = pcall(require, 'smart-splits')
+          if success then
+            print("require should not succeed")
+          end
+        '')
+        (lib.mkOrder 5001 ''
+          local success, _ = pcall(require, 'smart-splits')
+          if not success then
+            print("require should succeed")
+          end
+        '')
+      ];
+    };
+}


axelkar:
Test's init.lua:
```lua
-- Ignore the user lua configuration
vim.opt.runtimepath:remove(vim.fn.stdpath('config'))              -- ~/.config/nvim
vim.opt.runtimepath:remove(vim.fn.stdpath('config') .. "/after")  -- ~/.config/nvim/after

-- Nixvim's internal module table
-- Can be used to share code throughout init.lua
local _M = {}

require('lz.n').load( { { "smart-splits.nvim", after = function()
 require('smart-splits').setup({ })
 
end, lazy = true } })

local success, _ = pcall(require, 'smart-splits')
if success then
  print("require should not succeed")
end

require('lzn-auto-require').enable({ })

local success, _ = pcall(require, 'smart-splits')
if not success then
  print("require should succeed")
end
```

plugins/by-name/lzn-auto-require/default.nix
```diff
@@ -0,0 +1,9 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lzn-auto-require";
+
```
khaneliman, line 10:
```suggestion

  hasSettings = false;
```

khaneliman (commented):
LGTM overall. Just think we should restrict the `settings` from being used since this plugin doesn't have any configuration.

plugins/by-name/lzn-auto-require/default.nix
```diff
@@ -0,0 +1,10 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lzn-auto-require";
```
MattSturgeon, line 3: This would benefit from having a `description`. That'd be a good place to mention what this plugin is, why you might use it, and (maybe) how it can be used.

plugins/by-name/lzn-auto-require/default.nix
```diff
@@ -0,0 +1,10 @@
...
+
+  hasSettings = false;
+  setup = ".enable";
+  configLocation = lib.mkAfter "extraConfigLuaPost"; # Register after everything to catch mistates
```
MattSturgeon: If you really want to be after _everything_, do you also want to be after `mkAfter` with a custom `mkOrder`?

khaneliman: In my module I use `mkOrder 5000` https://github.com/khaneliman/khanelivim/blob/c94560d1cd12dbada5d1fcb963aeeebd729dd272/modules/nixvim/plugins/lz-n/default.nix#L10-L14

plugins/by-name/lzn-auto-require/default.nix
axelkar: Is this good? `require` lazy-loader for `lz.n`

plugins/by-name/lzn-auto-require/default.nix
axelkar: There's only a handful of plugins that define `configLocation`, but now using `mkOrder 5000`.

MattSturgeon: > There's only a handful of plugins that define `configLocation`

I wasn't sure if you were intending to be after _user_ definitions.

I see `mkAfter` as a user-facing function, and tend to prefer avoiding it in "internal" definitions, as an attempt at avoiding accidental conflicts with user definitions.

plugins/by-name/lzn-auto-require/default.nix
MattSturgeon: Honestly, it doesn't really mean much to me. I'd still have to go read the upstream documentation to know what this plugin is for.

MattSturgeon: The upstream description is:

> Auto load optional plugins via lua modules with lz.n 

and the start of their [README](https://github.com/horriblename/lzn-auto-require) is:

> Overrides `require` to also search for plugins in `{packpath}/*/opt` and load them with `lz.n`

I'd probably suggest something like:

```nix
description = ''
  Auto load optional plugins via lua modules with lz.n.

  This plugin overrides the `require` function to also search for plugins in
  `{packpath}/*/opt` and load them using `lz.n`.
''
```

plugins/by-name/lzn-auto-require/default.nix
khaneliman: I just did it that far down because the docs state they want it to be one of the last things in your config to load. 

https://github.com/horriblename/lzn-auto-require?tab=readme-ov-file#usage
>Place this code somewhere in your config, and call it as late as possible (e.g. at the bottom of your init.lua).
>
>require('lzn-auto-require').enable()

axelkar: > > There's only a handful of plugins that define `configLocation`
> 
> I wasn't sure if you were intending to be after _user_ definitions.
> 
> I see `mkAfter` as a user-facing function, and tend to prefer avoiding it in "internal" definitions, as an attempt at avoiding accidental conflicts with user definitions.

Agree. Enabling lzn-auto-require after user definitions is what we're after.

khaneliman (approved):

PR #3456, opened 2025-06-11
lib/builders: pass text as file

MattSturgeon:
This should fix potential "argument list too long" errors from bash when writing large files.

It looks like we have minimal test coverage for lib/builders currently, likely because we can't really test derivation content in our "lib-test". There is some coverage in the performance module tests, but no unit tests for the functions themselves.

We can test them using `pkgs.testers.testEqualContents` and `pkgs.testers.testBuildFailure`, but I don't have time to do that right now.

Without unit tests, I may have missed some edge-cases. I simply went through the `lib/builders.nix` file and added `passAsFile = ["text"]` to any instances of `inheret text`.

cc @khaneliman: can you still reproduce your issue?

### DIFF
--- a/lib/builders.nix
+++ b/lib/builders.nix
@@ -34,16 +34,21 @@ lib.fix (builders: {
   */
   writeLuaWith =
     pkgs: name: text:
-    pkgs.runCommand name { inherit text; } ''
-      echo -n "$text" > "$out"
-
-      ${lib.getExe pkgs.stylua} \
-        --no-editorconfig \
-        --line-endings Unix \
-        --indent-type Spaces \
-        --indent-width 4 \
-        "$out"
-    '';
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Buildbot is being _very_ slow atm, but there are a few failures showing up: https://buildbot.nix-community.org/#/builders/26/builds/5554

Seems related to luajit:

E.g.:

> ``luajit: ...ttr-09lahi25s7zcjgf5kbw95824psh4k4mlx84p8vygdcf1cd1679i5:1: unexpected symbol near '/'``

EDIT: one of the builders was already expecting a filepath, not contents.

khaneliman (approved):
Resolves my flake not building. Thanks for the quick fix Matt!

PR #3457, opened 2025-06-11
ci/mergify: drop

MattSturgeon:
We no longer use mergify, so drop the config file. This can be reverted if we ever change our minds.


GaetanLepage (approved):

MattSturgeon:
@GaetanLepage did you want to disable/uninstall the Mergify GH App on the repo?

Not strictly necessary, but it would prevent Mergify running against PRs unnecessarily and adding unrelated status checks.

If we change our minds and want it back, we may need to ask an org owner, but that shouldn't be an issue and seems unlikely at this point anyway.

GaetanLepage:
> @GaetanLepage did you want to disable/uninstall the Mergify GH App on the repo?

Yes sure, but I don't have enough rights to remove it. We need to ask an org owner.

GaetanLepage:
> We need to ask an org owner.

Done. The app was removed, along with the `nix-infra-bot` _collaborator_:
https://matrix.to/#/%23nix-community%3Anixos.org/%24QsXfIrTiqkDrglhSY-jgTGyn0osOm_zIUnKw7hR0jzA?via=blad.is&via=matrix.org&via=envs.net

Issue #3434, opened 2025-06-04
ci: move logic into an in-repo directory


MattSturgeon:
While it's nice for certain workflows not to need to checkout the repo, this makes things hard to develop and test.

We have some stuff in `update-scripts` already, but perhaps this being specifically focused on “updates” has given us tunnel vision...

Nixpkgs typically do a "sparse checkout" of the `ci/` directory; "sparse" means other parts of the repo are not downloaded. This allows keeping most of the workflow's logic in nix & shell scripts that can be used or tested independently.

We could do something similar.

This may also make the recently added "supported versions" info in `version-info.toml` somewhat redundant, because we could have a script that downloads `channels.nix` from NixOS/infra and extracts the relevant info for the particular workflow. Such a script could be used by both docs and update-other, without https://github.com/nix-community/nixvim/issues/3431 needing to reference `version-info.toml` specifically.[^1]

As another example, the logic in the docs workflow or the update workflow could be extracted into in-repo scripts that can be run/tested locally, without needing to actually run GitHub Actions.

[^1]: yes I feel a little silly realising this now, having only just merged that feature into `version-info.toml`...

PR #3462, opened 2025-06-11
ci/update-other: generate branch list from version-info

### LINKED ISSUE #3432: ci: generate 'update-other' matrix from `version-info.toml`

MattSturgeon:
Uses a [sparse-checkout](https://github.com/actions/checkout#scenarios) to download only the `version-info.toml` file, then uses the [`yq-go`](https://github.com/mikefarah/yq) action to extract the "other" branches as a JSON array.

This means we no longer need to manually update the workflow file when adding/removing stable branches. 🎉

Fixes #3432

---

This can be seen running on my fork in this workflow run: https://github.com/MattSturgeon/nixvim/actions/runs/15597236191

Notice that the branch matrix was created correctly, and both updates attempted. Each update failed because my fork doesn't have the relevant branches.

I created one of the branches on my fork and ran it again: https://github.com/MattSturgeon/nixvim/actions/runs/15597329168

This time notice that the `nixos-25.05` branch's `update.yml` was triggered successfully, and ran here: https://github.com/MattSturgeon/nixvim/actions/runs/15597337865

---

This doubles as a proof-of-concept for a similar change planned in the docs workflow: https://github.com/nix-community/nixvim/issues/3433

---

If it turns out we only use the "other versions" info in CI, then this could be refactored to run a script that directly parses NixOS/infra's `channels.nix` without needing to write that info to `version-info.toml`.

I believe the only other place we'll need this info will be building the multi-version docs, so perhaps a CI script is better?

Let's see as time goes on; this idea is already documented as part of #3434.

### DIFF
--- a/.github/workflows/update-other.yml
+++ b/.github/workflows/update-other.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   update:
     name: Trigger updates for nixvim's stable branches
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     if: github.repository == 'nix-community/nixvim'
     strategy:
       matrix:
--- a/.github/workflows/update-other.yml
+++ b/.github/workflows/update-other.yml
@@ -6,6 +6,7 @@ on:
 
 # Allow running workflows
 permissions:
+  contents: read
   actions: write
 
 jobs:
--- a/.github/workflows/update-other.yml
+++ b/.github/workflows/update-other.yml
@@ -15,17 +15,21 @@ jobs:
     runs-on: ubuntu-24.04-arm
     if: github.repository == 'nix-community/nixvim'
     outputs:
-      branches: ${{ steps.list-branches.outputs.branches }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-other.yml
```diff
@@ -11,7 +11,7 @@ permissions:
 jobs:
   update:
     name: Trigger updates for nixvim's stable branches
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
GaetanLepage: Why running this on ARM?

MattSturgeon: Anecdotally, nixpkgs seem to have found most GitHub workflow run faster and/or more efficiently on ARM.

They've migrated many of their workflow jobs to ARM over recent months.

Since we aren't doing anything that requires x86 here, I figured why not.

GaetanLepage: TIL!

MattSturgeon:
Thanks for your reviews!

PR #3466, opened 2025-06-13
ci/docs: extract action & refactor workflow

### LINKED ISSUE #3465: docs: build versions using a job matrix
We can speed up and simplify the docs build by building each version of the docs website using matrix jobs.

This will mean each job needs to upload its result as a workflow artifact, then a final job will download all the artifacts and deploy them to the github pages site.

As a graph, this looks like:

```
   compute-matrix
          |
      /   |   \
  main  25.05  24.11
      \   |   /
     deploy-docs
```

The `compute-matrix` job reads `version-info.toml` and outputs a json matrix.

The matrix jobs build the docs on the specified branch, upload the result to the workflow's artifacts with a short lifespan, and write the upload URL to the job outputs.

The `deploy-docs` downloads the artifact from each build job, combines them into the correct directory tree, then uploads it to GitHub Pages.

MattSturgeon:
Fixes #3465 and is preparation for #3433.

The high-level concept is to build each version of the docs in a separate parallel job, which should be faster and simpler.

1. A `build-docs` __action__ has been extracted; this encapsulates how we build & upload an individual version of the docs.
2. The `website.yml` **workflow** is responsible for _using_ `build-docs` on each branch, combining the builds, and deploying to github-pages.

This should have a few benefits:
- We can reuse the `build-docs` action to test PRs (not implemented in this PR)
- We can generate a build matrix from `version-info.toml` (not implemented in this PR)
- Deploying docs _should_ be faster, as each version is built in parallel, in separate containers
- The workflow+action should each be easier to reason about

These runs successfully built docs for all branches on my fork, and deployed to my fork's github-pages :tada:
- https://github.com/MattSturgeon/nixvim/actions/runs/15659242425
- https://github.com/MattSturgeon/nixvim/actions/runs/15666127205
- https://github.com/MattSturgeon/nixvim/actions/runs/15666205431

image

To avoid doing too much in one PR, the job matrix is hard-coded for now. In a future PR, I plan to generate the job matrix with a preparatory job, like we do in the update-other workflow.

One thing to note about using a local in-repo Action: this requires checking out the repo (or sparse-checking-out `.github`) before using the Action. This means the action file comes from whatever branch is checked out, _not_ from the ref in the workflow's context. This is in contrast to how `workflow_call` works, and means we need to ensure all supported branches have the new `build-docs` Action.


.github/actions/build-docs/action.yml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Move to sub-path
+      if: inputs.sub-path
+      shell: bash
+      env:
+        path: ${{ inputs.sub-path }}
+      run: |
+        mv result _result
+        mkdir -p "result/$path"
+        rmdir "result/$path"
+        mv _result "result/$path"
```
GaetanLepage: Isn't there a way to provide the name of the output symlink when running a Nix command?
That would allow you to get rid of this gymnastics.

GaetanLepage (approved):
The diff looks good! Thanks for testing the change on your repo.

.github/actions/build-docs/action.yml
GaetanLepage: https://releases.nixos.org/nix/nix-2.13.5/manual/command-ref/new-cli/nix3-build.html#options
> `--out-link` / `-o path`
> Use path as prefix for the symlinks to the build results. It defaults to result.

MattSturgeon: Much better, thanks:

```diff
diff --git a/.github/actions/build-docs/action.yml b/.github/actions/build-docs/action.yml
index fc2df4ce..078be042 100644
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -32,10 +32,12 @@ runs:
     - name: nix-build
       shell: bash
       env:
+        subPath: ${{ inputs.sub-path }}
         baseHref: ${{ inputs.base-href }}
         versions: ${{ inputs.versions }}
       run: |
         nix-build \
+            --out-link "docs/$subPath" \
             --argstr baseHref "$baseHref" \
             --argstr versionsJson "$versions" \
             --expr '
@@ -55,21 +57,10 @@ runs:
               }
             '
 
-    - name: Move to sub-path
-      if: inputs.sub-path
-      shell: bash
-      env:
-        path: ${{ inputs.sub-path }}
-      run: |
-        mv result _result
-        mkdir -p "result/$path"
-        rmdir "result/$path"
-        mv _result "result/$path"
-
     - name: Upload artifact
       id: upload
       uses: actions/upload-pages-artifact@v3
       with:
-        path: result
+        path: docs
         name: ${{ inputs.artifact-name }}
         retention-days: ${{ inputs.retention-days }}
```

It even creates missing directories when out-link is nested :rocket: 

```sh
$ rm -rf foo

$ nix build nixpkgs#hello --out-link foo/bar/baz

$ ls -R foo
foo:
bar/

foo/bar:
baz@
```

.github/workflows/website.yml
```diff
@@ -0,0 +1,133 @@
+name: Documentation
+
+on:
+  # TODO: Consider running on schedule instead of on push
```
MattSturgeon: I'll drop this TODO in the next push. We can discuss this in a separate issue rather than committing the idea into the repo.

MattSturgeon (commented):
I've done another pass of cleanup inspired by @GaetanLepage's suggestions.

I'll leave this open briefly in case anyone has further feedback. If not, I'll merge later.

MattSturgeon:
Do we want to do something like this, so that the per-version artifacts are immediately deleted once we've downloaded them?

```patch
commit cb789d66bebf1512339378e61016bdba94331fcf
Author: Matt Sturgeon <matt@sturgeon.me.uk>
Date:   Sun Jun 15 20:19:15 2025 +0100

    ci/docs: remove redundant artifacts
    
    Although these artifacts already have a short retention of one day, we
    can immediately clean up the per-version docs artifacts once we've
    downloaded them in the "combine" job.

diff --git a/.github/workflows/website.yml b/.github/workflows/website.yml
index 674aac8d..258a08fb 100644
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -82,6 +82,9 @@ jobs:
     runs-on: ubuntu-24.04-arm
     needs: build
 
+    permissions:
+      actions: write # to delete artifacts
+
     env:
       in: artifacts
       out: combined
@@ -109,6 +112,12 @@ jobs:
         with:
           path: ${{ env.out }}
 
+      - name: Delete Artifacts
+        uses: geekyeggo/delete-artifact@v5
+        with:
+          name: "*-docs"
+          useGlob: true
+
   deploy:
     name: Deploy
     runs-on: ubuntu-24.04-arm
```

They don't do any real harm, other than cluttering the workflow run's summary page. After 24h all artifacts will be deleted anyway, so their storage cost is negligible.

Issue #3052, opened 2025-03-07
[DOCS] Add descriptions to plugins page


osbm:
It would be super helpful to have a one sentence summary of the plugin, just like the [search.nixos.org](https://search.nixos.org).

![Image](https://github.com/user-attachments/assets/1308792b-09f3-4452-b1a6-8b6b3297e030)
I would place the description between the plugin name and the URL.

osbm:
Sorry, i just saw that this already exists. https://nix-community.github.io/nixvim/plugins/earthly.html

Maybe i can contribute PRs for adding descriptions to other plugins as well

PR #3487, opened 2025-06-17
treewide: add plugin descriptions

osbm:
Related to the https://github.com/nix-community/nixvim/issues/3052

I tried to make these explanations make as much sense as possible for the uninitiated.

But there are some plugins that are defined as nix modules, and their documentation pages doesn't show any meta fields just options. So i couldn't do much about them. But these should cover  every plugin inside the `by-name` folder.

Any change request is welcome.

### DIFF
--- a/plugins/by-name/lsp-format/default.nix
+++ b/plugins/by-name/lsp-format/default.nix
@@ -28,6 +28,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
   ];
 
   description = ''
+    A wrapper around Neovims native LSP formatting.
+
+    ---
+
     ## Configuring a Language
 
     `lsp-format` uses a table defining which lsp servers to use for each language.
--- a/plugins/by-name/mini/default.nix
+++ b/plugins/by-name/mini/default.nix
@@ -7,6 +7,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   packPathName = "mini.nvim";
   # Just want it before most other plugins for the icons provider.
   configLocation = lib.mkOrder 800 "extraConfigLua";
+  description = "A collection of Neovim plugins that are small, focused, and modular. They can be used individually or together to enhance your Neovim experience.";
 
   maintainers = [ lib.maintainers.khaneliman ];
 
--- a/plugins/cmp/default.nix
+++ b/plugins/cmp/default.nix
@@ -15,6 +15,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
   maintainers = [ lib.maintainers.GaetanLepage ];
 
   description = ''
+    A completion engine for Neovim written in Lua, designed to be fast and extensible.
+
+    ---
+
     ### Completion Source Installation
 
     If `plugins.cmp.autoEnableSources` is `true` Nixivm will automatically enable the corresponding source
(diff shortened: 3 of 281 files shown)


GaetanLepage (commented):
Thank you very much for this contribution!

I wonder if we should follow [nixpkgs' guideline for descriptions](https://nixos.org/manual/nixpkgs/unstable/#var-meta-description), i.e. no "A" at the beginning, no "." at the end, etc.

osbm:
I think setting consistent guidelines for the first line of the description makes a lot of sense(even if only for this PR).

If yall decide to add such guidelines, i can easily fix all of the descriptions.

MattSturgeon:
> I wonder if we should follow [nixpkgs' guideline for descriptions](https://nixos.org/manual/nixpkgs/unstable/#var-meta-description), i.e. no "A" at the beginning, no "." at the end, etc.

Unlike nixpkgs, we don't have a distinct `description` and `longDescription`.

I think in practice our descriptions are more analogous to `longDescription`; they're also kinda like _option_ descriptions... Therefore, full sentences and grammar make sense IMO. I believe this is what we have usually done, up 'til now.

plugins/cmp/default.nix
```diff
@@ -15,6 +15,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
   maintainers = [ lib.maintainers.GaetanLepage ];
 
   description = ''
+    A completion engine for Neovim written in Lua, designed to be fast and extensible.
+
+    ---
```
MattSturgeon, line 20: Is this following an established convention elsewhere in nixvim? I'm not convinced a `<hr>` is really necessary immediately before a `<h2>`

plugins/by-name/mini/default.nix
```diff
@@ -7,6 +7,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   packPathName = "mini.nvim";
   # Just want it before most other plugins for the icons provider.
   configLocation = lib.mkOrder 800 "extraConfigLua";
+  description = "A collection of Neovim plugins that are small, focused, and modular. They can be used individually or together to enhance your Neovim experience.";
```
MattSturgeon, line 10: Note for long lines:

These descriptions are rendered as common mark (markdown), so single linebreaks will be replaced with ordinary spaces when rendered.

This means you can do something like this, if preferred, and it will render the same.

```nix
description = ''
  A collection of Neovim plugins that are small, focused, and modular.
  They can be used individually or together to enhance your Neovim experience.
'';
```

Technically, that's equivialent to the nix string:

```nix
description = "A collection of Neovim plugins that are small, focused, and modular.\nThey can be used individually or together to enhance your Neovim experience.\n";
```

(Note, a central `\n` and a trailing `\n` at the end)

However since we're passing these strings to markdown, that difference doesn't matter.

plugins/by-name/lsp-format/default.nix
```diff
@@ -28,6 +28,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
   ];
 
   description = ''
+    A wrapper around Neovims native LSP formatting.
+
+    ---
```
MattSturgeon, line 33: Again, is this a common pattern in nixvim? I see you've added 22 of these :grin:

MattSturgeon (approved):
I've not read each individual description, but most seem reasonable. I assume they're based on upstream descriptions?

Thanks for doing this work!

A couple questions below, but no objections from me.

khaneliman (approved):

PR #3491, opened 2025-06-19
plugins/otter: make `autoActivate` in `lsp` module

HeitorAugustoLN:


plugins/by-name/otter/default.nix
```diff
@@ -113,6 +113,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
     plugins.lsp.onAttach = lib.mkIf cfg.autoActivate ''
       require('otter').activate()
     '';
+
+    lsp.onAttach = lib.mkIf cfg.autoActivate ''
+      require('otter').activate()
+    '';
```
MattSturgeon: Should we do this _instead_ of `plugins.lsp.onAttach`, rather than in _addition to_ it?

IIRC, `lsp.onAttach` is always used (if non-empty), while `plugins.lsp.onAttah` is only used if `plugins.lsp.enable`.

In effect, this means that if you have a line in both `onAttach` options, it'll run once when `plugins.lsp.enable = false` and **twice** when `plugins.lsp.enable = true`.

Unless I'm forgetting something?

khaneliman: I agree, it makes sense to just migrate it to the `lsp` module. This isn't dependent on anything in `plugins.lsp` I believe.

HeitorAugustoLN: Migrated to the new lsp module instead of using both

PR #3495, opened 2025-06-22
Add gruvbox-material colorscheme

f4z3r:
Gruvbox-material is a colorscheme in pure lua: https://github.com/f4z3r/gruvbox-material.nvim. It is highly customizable and supports a wide variety of plugins.

### DIFF
--- a/plugins/colorschemes/gruvbox-material.nix
+++ b/plugins/colorschemes/gruvbox-material.nix
@@ -1,0 +1,42 @@
+{
+  lib,
+  config,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gruvbox-material";
+  isColorscheme = true;
+  packPathName = "gruvbox-material.nvim";
+  package = "gruvbox-material-nvim";
+
+  maintainers = [ lib.maintainers.f4z3r ];
+
+  settingsExample = {
+    italics = true;
+    contrast = "medium";
+    comments = {
+      italics = true;
+    };
+    background = {
+      transparent = false;
+    };
+    float = {
+      force_background = false;
+    };
+    signs = {
+      highlight = true;
+    };
+    customize = lib.nixvim.nestedLiteralLua ''
+      function(g, o)
+        local colors = require("gruvbox-material.colors").get(vim.o.background, "medium")
+        if g == "CursorLineNr" then
+          o.link = nil            -- wipe a potential link, which would take precedence over other
+                                  -- attributes
+          o.fg = colors.orange    -- or use any color in "#rrggbb" hex format
+          o.bold = true
+        end
+        return o
+      end
+    '';
+  };
+}
--- a/tests/test-sources/plugins/colorschemes/gruvbox-material.nix
+++ b/tests/test-sources/plugins/colorschemes/gruvbox-material.nix
@@ -1,0 +1,66 @@
+{ lib, ... }:
+{
+  empty = {
+    colorschemes.gruvbox-material.enable = true;
+  };
+
+  defaults = {
+    colorschemes.gruvbox-material = {
+      enable = true;
+
+      settings = {
+        italics = true;
+        contrast = "medium";
+        comments = {
+          italics = true;
+        };
+        background = {
+          transparent = false;
+        };
+        float = {
+          force_background = false;
+          background_color = null;
+        };
+        signs = {
+          highlight = true;
+        };
+        customize = null;
+      };
+    };
+  };
+
+  example = {
+    colorschemes.gruvbox-material = {
+      enable = true;
+
+      settings = {
+        italics = true;
+        contrast = "medium";
+        comments = {
+          italics = true;
+        };
+        background = {
+          transparent = false;
+        };
+        float = {
+          force_background = false;
+        };
+        signs = {
+          highlight = true;
+        };
+        customize = lib.nixvim.mkRaw ''
+          function(g, o)
+            local colors = require("gruvbox-material.colors").get(vim.o.background, "medium")
+            if g == "CursorLineNr" then
+              o.link = nil            -- wipe a potential link, which would take precedence over other
+                                      -- attributes
+              o.fg = colors.orange    -- or use any color in "#rrggbb" hex format
+              o.bold = true
+            end
+            return o
+          end
+        '';
+      };
+    };
+  };
+}
--- a/plugins/default.nix
+++ b/plugins/default.nix
@@ -12,6 +12,7 @@
     ./colorschemes/everforest.nix
     ./colorschemes/github-theme.nix
     ./colorschemes/gruvbox.nix
+    ./colorschemes/gruvbox-material.nix
     ./colorschemes/kanagawa.nix
     ./colorschemes/kanagawa-paper.nix
     ./colorschemes/melange.nix


lib/maintainers.nix
```diff
@@ -102,6 +102,13 @@
     githubId = 50964650;
     name = "gwen";
   };
+  f4z3r = {
+    email = "f4z3r-github@pm.me";
+    name = "Jakob Beckmann";
+    github = "f4z3r";
+    githubId = 32326425;
+    keys = [ { fingerprint = "358A 6251 E2ED EDC1 9717  14A7 96A8 BA6E C871 2183"; } ];
+  };
```
khaneliman: Separate commit, please.

plugins/colorschemes/gruvbox-material.nix
```diff
@@ -0,0 +1,81 @@
...
+        local colors = require("gruvbox-material.colors").get(vim.o.background, "medium")
+        if g == "CursorLineNr" then
+          o.link = nil            -- wipe a potential link, which would take precedence over other
+                                  -- attributes
+          o.fg = colors.orange    -- or use any color in "#rrggbb" hex format
+          o.bold = true
+        end
+        return o
+      end
+      ```
```
khaneliman: This would be better suited for the `settingsExample`

GaetanLepage (commented):
Thank you for this contribution.
We now tend to avoid declaring new options explicitly (i.e. not having `settingsOptions` at all).
Of course, `colorschemes.gruvbox-material.settings` remains a free-form option which allows users to provide arbitrary values.

Would you mind getting rid of the option declarations?
If you absolutely want to keep them, this is fine too. But bare in mind that you (as the module maintainer) will be entitled to keep them up to date.

f4z3r:
> Thank you for this contribution. We now tend to avoid declaring new options explicitly (i.e. not having `settingsOptions` at all). Of course, `colorschemes.gruvbox-material.settings` remains a free-form option which allows users to provide arbitrary values.
> 
> Would you mind getting rid of the option declarations? If you absolutely want to keep them, this is fine too. But bare in mind that you (as the module maintainer) will be entitled to keep them up to date.

No worries. I removed the `settingsOptions` and made the example a bit longer to showcase more options. I hope this is fine.

GaetanLepage:
> No worries. I removed the `settingsOptions` and made the example a bit longer to showcase more options. I hope this is fine.

Thanks!
You also need to add a test case in `tests/test-sources/plugins/colorschemes/gruvbox-material.nix` (look at other files to see how it's done).
Also, squash your commits so as to leave only two:
- `maintainers: add f4z3r`
- `colorschemes/gruvbox-material: init`

lib/maintainers.nix
MattSturgeon: Thanks for splitting this out.

_Really_ sorry to nit-pick this, but the maintainers commit should be before the plugin commit.

If someone was to bisect the history, the gruvbox commit would show up as a build failure because it uses a maintainer that is only added in the _next_ commit.

EDIT: looking more closely, I see you handled that by not adding a maintainer to the plugin until the second commit. We'd still prefer adding the maintainer in a earlier commit and _then_ using it in a later commit.

MattSturgeon: Actually, looking at the CI failure: https://buildbot.nix-community.org/#/builders/559/builds/5431/steps/1/logs/stdio

You're already a nixpkgs maintainer, so there's no need to add yourself here. You can drop the commit.

plugins/colorschemes/gruvbox-material.nix
```diff
@@ -0,0 +1,42 @@
...
+    background = {
+      transparent = false;
+    };
+    float = {
+      force_background = false;
+    };
+    signs = {
+      highlight = true;
+    };
+    customize = ''
```
GaetanLepage:
```suggestion
    customize = lib.nixvim.nestedLiteralLua ''
```

tests/test-sources/plugins/colorschemes/gruvbox-material.nix
```diff
@@ -0,0 +1,66 @@
...
+        background = {
+          transparent = false;
+        };
+        float = {
+          force_background = false;
+        };
+        signs = {
+          highlight = true;
+        };
+        customize = ''
```
GaetanLepage:
```suggestion
        customize = lib.nixvim.mkRaw ''
```

GaetanLepage (commented):
Looks very good! You only need to properly mark lua strings as such.

f4z3r:
@GaetanLepage added the functions to mark the strings.

@MattSturgeon if I understand it correctly, if I am a nixpkgs maintainer I don't need to add myself to the maintainers list. I removed myself from there. I left the maintainer reference in the plugin reference. Should that also be removed?

Edit: quickly looking at the code in more detail, it does not seem to make sense to reference a maintainer that is not listed in the maintainer list in this repository. So I removed the reference as well.

plugins/colorschemes/gruvbox-material.nix
```diff
@@ -0,0 +1,40 @@
+{
+  lib,
+  config,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gruvbox-material";
+  isColorscheme = true;
+  packPathName = "gruvbox-material.nvim";
+  package = "gruvbox-material-nvim";
```
MattSturgeon, line 10: > quickly looking at the code in more detail, it does not seem to make sense to reference a maintainer that is not listed in the maintainer list in this repository. So I removed the reference as well.

`lib` comes from nixpkgs. You are a nixpkgs maintainer. This will work fine and is the same as many of our other maintainer entries.

```suggestion
  package = "gruvbox-material-nvim";

  maintainers = [ lib.maintainers.f4z3r ];
```

In fact, we have CI that enforces _not_ adding to our in-tree maintainers list if you're already in nixpkgs' list. [That's how](https://buildbot.nix-community.org/#/builders/559/builds/5431/steps/1/logs/stdio) I worked out you were a nixpkgs maintainer.

Without this, we get:
```
error: function 'anonymous lambda' called without required argument 'maintainers'
```

f4z3r: My bad. I will re-add the maintainer field then. Will be back at the computer in 15 minutes to fix it 👍

chelsea6502:
Hi f4z3r and others,

While I appreciate the effort to bring gruvbox-material to nixvim, I would have rather have had the 'gruvbox-material' colorscheme name reserved for the earlier and more widely used repository: https://github.com/sainnhe/gruvbox-material

I was hoping to bring sainnhe's version to the Nixvim repo, but it seems like the colour scheme is already taken. Is there a way we can have both somehow, or are we forced to choose one or the other?

f4z3r:
We can rename the lua version to `gruvbox-material-nvim`. Would be fine for me. That is how the package is named anyways in nixpkgs, etc.

MattSturgeon:
Renaming `colorscheme.gruvbox-material` to  `colorscheme.gruvbox-material-nvim` wouldn't immediately free up the old name for adding a new plugin, because we'd ideally want the old name to be a "rename alias" to avoid breaking users' configs.

However, because this has only been merged for a few weeks and hasn't made it into a "stable" version, I think it's fine for such a rename alias to be short-lived and replaced by a plugin using https://github.com/sainnhe/gruvbox-material, which is [already packaged](https://search.nixos.org/packages?channel=unstable&show=vimPlugins.gruvbox-material) in nixpkgs.

Issue #3417, opened 2025-06-02
Build docs on GitHub instead of buildbot


MattSturgeon:
Currently we configure buildbot to build everything in our `packages` flake output on all systems. For the docs, this means x86_64-linux and aarch64-linux.

We don't really care that much if the docs can build on aarch64, so this is a waste of resources. Further, buildbot often struggles building docs on this platform; it is often the longest running build by far.

Since #3410, we are able to configure what buildbot builds explicitly without affecting `nix flake check`. I'd propose we set up a GitHub Actions workflow to build docs on pull_request or pull_request_target events and remove docs from buildbot.

Eventually, this workflow could also publish the PR's docs somewhere for easy viewing, however that should be out of scope for the initial PR. This would likely mean opening a dedicated "nixvim PR docs" repo for hosting another github-pages instance, or using an alternative to github pages.

TODO:

- [ ] Remove docs packages from buildbot
- [ ] Build docs website on github
- [ ] Build other docs packages on github
- [ ] Publish docs websites for PRs

PR #3503, opened 2025-06-24
treewide: re-enable darwin sandbox disablements

khaneliman:
This pull request introduces several changes to improve configuration handling, simplify code, and remove platform-specific restrictions. The most notable updates include restructuring language tools in `ci/efmls-configs.nix`, simplifying expressions in `docs/default.nix`, and removing Darwin-specific exceptions in plugin configurations.

### Efmls case sensitive support
* [`ci/efmls-configs.nix`](diffhunk://#diff-b23aa4e05c041b2fa0219fdee605957fc8307cfadf5595de4416fab692885459L16-R43): Refactored the handling of language tools by grouping languages by lowercase names and merging their tools. This ensures canonical naming and avoids case-sensitive duplication. Also, merged linters and formatters across case variations for better consistency.
  - [x] Backwards compatible rename of options

### Removal of Platform-Specific Restrictions:
* [`docs/default.nix`](diffhunk://#diff-eb8d49802027f9ca035129d9b6dca2489d8de2643b4114108f4966567c638806L103-R103): Removed darwin limitation to allow building and serving docs on darwin.
  - [ ] ~Limit docs to be tested only on linux~ We will handle later

* [`tests/test-sources/plugins/by-name/claude-code/default.nix`](diffhunk://#diff-2608ef3b819fc7f91a164f06d900bf224ce0ef779461684b51ccd61e5ad85161L1-R1): Removed the conditional disabling of the plugin on Darwin platforms, enabling it universally.
* [`tests/test-sources/plugins/by-name/none-ls/default.nix`](diffhunk://#diff-56b3b1b4d4e6f552d914c6da0a7dbded642a7e37717b2d23d1dd81d3e219bd72L103-R103): Removed restrictions preventing the plugin from being enabled on Darwin, making it available on all platforms.
* ~[`tests/test-sources/plugins/by-name/vimtex/default.nix`](diffhunk://#diff-f4d2100ed33860d170daab1da30d8f75dbc4c74fc3620306a537b4701ef01377L1-R9): Eliminated the workaround for disabling `texlivePackage` on Darwin, allowing the plugin to function without platform-specific exceptions. [[1]](diffhunk://#diff-f4d2100ed33860d170daab1da30d8f75dbc4c74fc3620306a537b4701ef01377L1-R9) [[2]](diffhunk://#diff-f4d2100ed33860d170daab1da30d8f75dbc4c74fc3620306a537b4701ef01377L74-R66)~ Has a different issue that requires the workaround still.

### DIFF
--- a/tests/test-sources/plugins/by-name/none-ls/default.nix
+++ b/tests/test-sources/plugins/by-name/none-ls/default.nix
@@ -100,8 +100,7 @@
     in
     {
       plugins.none-ls = {
-        # sandbox-exec: pattern serialization length 159032 exceeds maximum (65535)
-        enable = !hostPlatform.isDarwin;
+        enable = true;
 
         sources =
           let
@@ -125,6 +124,21 @@
               ++ lib.optionals hostPlatform.isDarwin [
                 # TODO 2025-04-20 build failure
                 "ansiblelint"
+                # TODO 2025-06-24 build failure
+                "elm_format"
+                # TODO 2025-06-24 marked broken / unsupported platform
+                "clazy"
+                "haml_lint"
+                "racket_fixw"
+                "raco_fmt"
+                "rubyfmt"
+              ]
+              ++ lib.optionals (hostPlatform.isDarwin && hostPlatform.isx86_64) [
+                # TODO: 2025-06-24 build failure
+                "gleam_format"
+                "ptop"
+                # NOTE: No hash for x86 darwin
+                "verible_verilog_format"
               ];
           in
           # Enable every none-ls source that has an option
--- a/tests/test-sources/plugins/by-name/vimtex/default.nix
+++ b/tests/test-sources/plugins/by-name/vimtex/default.nix
@@ -1,7 +1,6 @@
 { pkgs, ... }:
 let
-  # CI failure: When trying to build `pkgs.texlive.combined.scheme-medium` on darwin, it fails with
-  #  > sandbox-exec: pattern serialization length 75279 exceeds maximum (65535)
+  # TODO: Added 2025-06-24 texlive dependency broken
   disableTexlivePackageOnDarwin = pkgs.lib.optionalAttrs pkgs.stdenv.isDarwin {
     texlivePackage = null;
   };
--- a/ci/efmls-configs.nix
+++ b/ci/efmls-configs.nix
@@ -13,27 +13,34 @@ let
   miscFormatters = toLangTools' "misc" "formatters";
 
   sources =
-    (lib.listToAttrs (
-      lib.map (
-        lang:
+    let
+      # Group languages by lowercase name and merge their tools
+      groupedLanguages = builtins.groupBy (lang: lib.toLower lang) languages;
+      mergedLanguages = lib.mapAttrs (
+        lowerLang: langList:
         let
-          toLangTools = toLangTools' lang;
+          # Use the first language as the canonical name but prefer lowercase if available
+          canonicalLang = if lib.elem lowerLang langList then lowerLang else lib.head langList;
+          # Merge all tools from all case variations
+          allLinters = lib.unique (lib.concatMap (lang: toLangTools' lang "linters") langList);
+          allFormatters = lib.unique (lib.concatMap (lang: toLangTools' lang "formatters") langList);
         in
         {
-          name = lang;
+          name = canonicalLang;
           value = {
             linter = {
-              inherit lang;
-              possible = (toLangTools "linters") ++ miscLinters;
+              lang = canonicalLang;
+              possible = allLinters ++ miscLinters;
             };
             formatter = {
-              inherit lang;
-              possible = (toLangTools "formatters") ++ miscFormatters;
+              lang = canonicalLang;
+              possible = allFormatters ++ miscFormatters;
             };
           };
         }
-      ) languages
-    ))
+      ) groupedLanguages;
+    in
+    (lib.listToAttrs (lib.attrValues mergedLanguages))
     // {
       all = {
         linter = {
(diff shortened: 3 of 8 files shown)


MattSturgeon:
Nice!

> Limit docs to be tested only on linux

This would be solved by https://github.com/nix-community/nixvim/issues/3417, although that isn't the only solution. We could also add conditions to the docs packages in the buildbot flake output.

I don't see this necessarily as a blocker though.

khaneliman:
> Nice!
> 
> > Limit docs to be tested only on linux
> 
> This would be solved by #3417, although that isn't the only solution. We could also add conditions to the docs packages in the buildbot flake output.
> 
> I don't see this necessarily as a blocker though.

Sounds good, gonna clean this up a bit later today and try to get it ready, then.

MattSturgeon:
> I don't see this necessarily as a blocker though.

@GaetanLepage may have a different opinion.

I guess it depends on how intensive darwin builds of the docs are for buildbot.

If they're really slow, we should hold of until we have disabled them for buildbot. Otherwise it's a minor issue.

GaetanLepage:
> > I don't see this necessarily as a blocker though.
> 
> @GaetanLepage may have a different opinion.
> 
> I guess it depends on how intensive darwin builds of the docs are for buildbot.
> 
> If they're really slow, we should hold of until we have disabled them for buildbot. Otherwise it's a minor issue.

Ideally, I would keep the docs building enabled on darwin until we can migrate to GHA.
Is there an advantage of disabling them right now? (besides lightening the buildbot load)

MattSturgeon:
> Is there an advantage of disabling them right now? (besides lightening the buildbot load)

Buildbot load is the only factor to consider.

khaneliman:
> Backwards compatible rename of options

Wasn't sure how to tackle this with the `efmls` options.

MattSturgeon:
> > Backwards compatible rename of options
> 
> Wasn't sure how to tackle this with the `efmls` options.

Looking at the diff, it is only `HTML` and `JSON` that have been removed. I'd suggest we just find/create an appropriate module somewhere to put rename modules - unless I'm missing something?

---

Ah, I see. This is complicated by `plugins.efmls-configs.setup` being a submodule, so we will run into https://github.com/NixOS/nixpkgs/issues/96006.

In this case, both the old/new options are within the same submodule, so [this](https://github.com/NixOS/nixpkgs/issues/96006#issuecomment-2935101404) applies:

> Just to be clear, you _can_ use `mkRenamedOptionModule` (and its other `doRename`-based siblings) within a submodule.
> 
> However, for it to create the expected warnings you need to add a `warnings` option to the submodule and use it to propagate warnings to the "main" `warnings` option in the outer configuration. For example, we do this in nixvim [here](https://github.com/nix-community/nixvim/blob/9328f4437d5f788d1c066b274a0aea492dc5fde2/modules/lsp/servers/server-renames.nix) and [here](https://github.com/nix-community/nixvim/blob/9328f4437d5f788d1c066b274a0aea492dc5fde2/modules/lsp/servers/default.nix#L160-L161).

(This kinda thing is what https://github.com/NixOS/nixpkgs/pull/391544 aims to solve)

---

While it's clearly possible to provide a rename, I wonder if it is worth the effort? I don't mind putting something together and pushing it up if you like, but how many users will actually be using the capitalized `HTML` or `JSON` options in their config?

MattSturgeon:
I've pushed 35e37228e85c3bdfe272f7e13ebb078d3f46d728 to implement this. Feel free to drop it if you'd rather _not_ provide aliases.

khaneliman:
> I've pushed [35e3722](https://github.com/nix-community/nixvim/commit/35e37228e85c3bdfe272f7e13ebb078d3f46d728) to implement this. Feel free to drop it if you'd rather _not_ provide aliases.

LGTM, thanks

tests/test-sources/plugins/by-name/vimtex/default.nix
```diff
@@ -1,7 +1,6 @@
 { pkgs, ... }:
 let
-  # CI failure: When trying to build `pkgs.texlive.combined.scheme-medium` on darwin, it fails with
-  #  > sandbox-exec: pattern serialization length 75279 exceeds maximum (65535)
+  # TODO: Added 2025-06-24 texlive dependency broken
```
MattSturgeon, line 3: The commit message says the test was re-enabled, but the diff only shows the comment being updated

tests/test-sources/plugins/by-name/none-ls/default.nix
```diff
@@ -125,6 +124,21 @@
...
                 # TODO 2025-04-20 build failure
                 "ansiblelint"
+                # TODO 2025-06-24 build failure
+                "elm_format"
+                # TODO 2025-06-24 marked broken / unsupported platform
+                "clazy"
+                "haml_lint"
+                "racket_fixw"
+                "raco_fmt"
+                "rubyfmt"
```
MattSturgeon, line 134: Can any of these be detected with `meta.broken` or `lib.meta.availableOn`?

tests/test-sources/plugins/by-name/vimtex/default.nix
khaneliman: Ah, yeah. Forgot to update message after discovering a dependency error.

tests/test-sources/plugins/by-name/none-ls/default.nix
khaneliman: Not a bad idea for these package sets.

khaneliman: ~have a change I'm testing real quick~ meh, not cooperating. can handle in a follow up

PR #2255, opened 2024-09-15
[WIP]: new-plugin CI workflow

MattSturgeon:
Initial conceptual stub for a CI workflow that can announce new plugins.

Pushing this up WIP and unfinished as I find GitHub workflows particularly draining 😅

Hopefully I've stubbed enough out that someone could continue this if I eventually abandon it. 👀

PR #3510, opened 2025-06-26
flake: add diff-plugins command

MattSturgeon:
Prints the plugins added or removed when compared to a particular nixvim revision.

This is intended as a starting point for automatically generating the matrix announcements, as attempted in #2255 and #2593.

I may consider moving the python script into the `ci/` directory to make sparse-checkout easier if it ends up being used by CI workflows. Although, saying that the full checkout would be needed to compare its plugins to the other anyway...

A CI workflow can use this script by supplying `github.event.before`, which for `on: push` events comes from the [`push` webhook event payload](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push). Alternatively we might use `github.event.pull_request.merge_commit_sha` or `github.event.pull_request.head.sha` from `pull_request_target`, since that also gives us access to things like `github.event.pull_request.number` ([`pull_request: closed` webhook event](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=closed#pull_request)).

This script can be tested locally by running `diff-plugins <some_sha>` to compare the current local nixvim to some arbitrary "before" state.

### DIFF
--- a/flake/dev/diff-plugins.py
+++ b/flake/dev/diff-plugins.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+
+import json
+import re
+import subprocess
+from argparse import ArgumentParser, ArgumentTypeError
+
+
+def main():
+    """
+    Main function to compare nixvim plugins with another revision.
+    """
+
+    parser = ArgumentParser(description="Compare nixvim plugins with another revision")
+    parser.add_argument(
+        "flakeref",
+        metavar="old",
+        help="the commit or flakeref to compare against",
+        type=flakeref,
+    )
+    parser.add_argument(
+        "--compact",
+        "-c",
+        help="produce compact json instead of prettifying",
+        action="store_true",
+    )
+    args = parser.parse_args()
+
+    after_plugins = list_plugins(".")
+    before_plugins = list_plugins(args.flakeref)
+    print(
+        json.dumps(
+            diff(before_plugins, after_plugins),
+            separators=((",", ":") if args.compact else None),
+            indent=(None if args.compact else 4),
+            sort_keys=(not args.compact),
+            default=list,
+        )
+    )
+
+
+def flakeref(arg):
+    """
+    An argparse type that represents a flakeref, or a partial flakeref that we can
+    normalise using sane defaults.
+    """
+    default_protocol = "github:"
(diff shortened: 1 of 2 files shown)


flake/dev/diff-plugins.py
```diff
@@ -0,0 +1,108 @@
...
+        "nix",
+        "eval",
+        f"{flake}#nixvimConfiguration.options",
+        "--apply",
+        expr,
+        "--json",
+    ]
+    out = subprocess.check_output(cmd)
+    # Parse as json, converting the lists to sets
+    return {k: set(v) for k, v in json.loads(out).items()}
```
khaneliman, line 104: Does it make sense to try just leveraging our `list-plugins` package/script?

MattSturgeon: I thought about it briefly, but I think it's simpler to keep them separate?

The list-plugins script is meant to print a user-readable summary of the plugins, right? While this one is meant to print a JSON diff.

Unless you have ideas for how we could do it without adding too much complexity to the list-plugins script?

khaneliman: Yeah, this looks to be a pretty simple implementation. We can always refactor later since it might require some more rework for `list-plugins`

flake/dev/diff-plugins.py
```diff
@@ -0,0 +1,108 @@
...
+    parser.add_argument(
+        "--compact",
+        "-c",
+        help="produce compact json instead of prettifying",
+        action="store_true",
+    )
+    args = parser.parse_args()
+
+    after_plugins = list_plugins(".")
+    before_plugins = list_plugins(args.flakeref)
```
khaneliman, line 30: Could make it more robust and allow passing in 2 flakerefs if we wanted to compare between different locations, even. Then default to `.` when its not provided? Just a thought for future possibility

MattSturgeon: We had this in the previous attempts. I left it out of this one to keep things as simple as possible.

Follow-up PRs welcome :wink:

MattSturgeon: Ended up actually _needing_ this in https://github.com/nix-community/nixvim/pull/3513 :see_no_evil:

PR #3512, opened 2025-06-27
plugins/kitty-scrollback: init

nim65s:
Add support for [kitty-scrollback.nvim](https://github.com/mikesmithgh/kitty-scrollback.nvim), to open kitty scrollback in nvim

### DIFF
--- a/plugins/by-name/kitty-scrollback/default.nix
+++ b/plugins/by-name/kitty-scrollback/default.nix
@@ -4,4 +4,13 @@ lib.nixvim.plugins.mkNeovimPlugin {
   package = "kitty-scrollback-nvim";
 
   maintainers = [ lib.maintainers.nim65s ];
+
+  settingsExample = lib.literalExpression ''
+    {
+      # create your config
+      myconfig.kitty_get_text.ansi = false;
+      # or extend a builtin one
+      ksb_builtin_last_cmd_output.ansi = false;
+    }
+  '';
 }


plugins/by-name/kitty-scrollback/default.nix
```diff
@@ -0,0 +1,7 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "kitty-scrollback";
+  package = "kitty-scrollback-nvim";
+
+  maintainers = [ lib.maintainers.nim65s ];
```
MattSturgeon, line 6: Should we have a `settingsExample` inspired by something in https://github.com/mikesmithgh/kitty-scrollback.nvim#plugin-configuration ?

plugins/by-name/kitty-scrollback/default.nix
```diff
@@ -0,0 +1,7 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "kitty-scrollback";
```
MattSturgeon, line 3: Do we need to set `packPathName` to `"kitty-scrollback.nvim"`?

(See https://nix-community.github.io/nixvim/CONTRIBUTING.html#mkneovimplugin and [`:h 'packpath'`](https://neovim.io/doc/user/options.html#'packpath'))

Aside: we really should add some way of asserting the correct packpath name is defined :face_with_diagonal_mouth:

MattSturgeon (commented):
Thanks! Looking good

nim65s:
woops, I forgot this PR, sorry about that.

plugins/by-name/kitty-scrollback/default.nix
nim65s: I have been dogfooding this for a while, and it work without.
But I'm not sure if this would be useful anyway.

khaneliman: `packPathName` was only relevant for lazy loading. But, it's no longer required after a recent PR that handles it automatically for you.

plugins/by-name/kitty-scrollback/default.nix
```diff
@@ -0,0 +1,14 @@
...
+  package = "kitty-scrollback-nvim";
+
+  maintainers = [ lib.maintainers.nim65s ];
+
+  settingsExample = {
+    # create your config
+    myconfig.kitty_get_text.ansi = false;
+    # or extend a builtin one
+    ksb_builtin_last_cmd_output.ansi = false;
+  };
```
MattSturgeon: If these comments are intended to be shown in documentation, then you need:

```suggestion
  settingsExample = lib.literalExpression ''
    {
      # create your config
      myconfig.kitty_get_text.ansi = false;
      # or extend a builtin one
      ksb_builtin_last_cmd_output.ansi = false;
    }
  '';
```

PR #3513, opened 2025-06-27
ci: add WIP pr-merged workflow

MattSturgeon:
Adds a workflow that runs when a PR is merged. The plan is to use this to do things like sending the matrix announcement messages, however for now it is just printing things to the markdown workflow run "summary".

My plan is to let this run whenever a PR is closed and come back later to look at what was printed. I'm especially interested in how merge commits and base commits are calculated when the merge queue is used, and when multiple PRs are merged as one queue group.

I can't easily test queues, because they can only be enabled on _org_ repos, not on personal repos or forks.

I've tested as much as I can on my fork, here's an example run: https://github.com/MattSturgeon/nixvim/actions/runs/15928141828?pr=18

### DIFF
--- a/.github/workflows/pr-merged.yml
+++ b/.github/workflows/pr-merged.yml
@@ -1,0 +1,166 @@
+name: PR Merged
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  contents: read
+
+jobs:
+  merged:
+    name: Show info
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-24.04-arm
+    steps:
+      # github.event.pull_request.base may not refer to the _actual_ parent commit of the PR,
+      # instead it is the commit that the PR was merged/rebased onto.
+      #
+      # Since we want to diff the PR's commits, query the first commit's first parent.
+      #
+      # Alternatively, we could compare
+      # github.event.pull_request.merge_commit_sha to github.event.pull_request.base.sha,
+      # however I'm unsure how that'll interact with grouped Merge Queue 🤔
+      - name: Get base commit
+        id: base
+        env:
+          GH_TOKEN: ${{ github.token }}
+          owner: ${{ github.repository_owner }}
+          repo: ${{ github.event.repository.name }}
+          pr: ${{ github.event.pull_request.number }}
+        run: |
+          out=$(mktemp)
+          gh api graphql -F owner="$owner" -F repo="$repo" -F pr="$pr" -f query='
+              query($owner: String!, $repo: String!, $pr: Int!) {
+                repository(owner: $owner, name: $repo) {
+                  pullRequest(number: $pr) {
+                    commits(first: 1) {
+                      edges {
+                        node {
+                          commit {
+                            parents(first: 1) {
+                              edges {
+                                node {
+                                  oid
+                                }
+                              }
+                            }
+                          }
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+              ' --jq '
+                .data
+                .repository
+                .pullRequest
+                .commits
+                .edges[0]
+                .node
+                .commit
+                .parents
+                .edges[0]
+                .node
+                .oid
+              ' > "$out"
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-merged.yml
```diff
@@ -0,0 +1,177 @@
...
+        # TODO: sparse checkout
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: |
+            accept-flake-config = true
+
+      # NOTE: This is here for debugging and gathering data
+      # TODO: Remove
```
GaetanLepage, line 87: What is the time scope for this `TODO: Remove`?

MattSturgeon: Once we understand how PRs merged in grouped batches (by the queue) show up.

Likely it will be removed the next time we touch this workflow, hopefully to integrate it with a "generate message" script.

GaetanLepage: Ok!

GaetanLepage (approved):
Looks good to me.
I fully trust you on the GHA stuff ;)

flake/dev/diff-plugins.py
```diff
@@ -13,11 +13,20 @@ def main():
...
     )
+    parser.add_argument(
+        "head_flakeref",
+        metavar="after",
+        help="the commit or flakeref to compare to (default .)",
+        type=flakeref,
+        nargs="?",
+        default=".",
+        const=".",
+    )
```
khaneliman, line 30: tee hee :P

.github/workflows/pr-merged.yml
```diff
@@ -0,0 +1,177 @@
...
+          head: ${{ github.event.pull_request.merge_commit_sha }}
+        run: |
+          out=$(mktemp)
+
+          # FIXME: first run will print to stdout: "pre-commit installed at .git/hooks/pre-commit"
+          nix develop --command echo first run
+
+          # FIXME: evaluating & building the devshell is expensive, we should
+          # provide an alternative way for CI to run diff-plugins
+          nix develop --command diff-plugins --compact "$base" "$head" > "$out"
```
khaneliman: If we want to avoid building the entire dev shell, we can just make the `diff-plugins` use a `nix-shell` shebang and call it directly.

khaneliman: created https://github.com/nix-community/nixvim/pull/3516 to show you what i mean

MattSturgeon:
With #3516 merged, I've updated to do a sparse checkout of only the script and run it directly with no devshell.

Here's an updated run to show it working: https://github.com/MattSturgeon/nixvim/actions/runs/15947974206?pr=20

I had to add `nixpkgs` to the `nix_path` in the install-nix action, else the `nix-shell` shebangs don't know where to get packages from.

.github/workflows/pr-merged.yml
```diff
@@ -0,0 +1,166 @@
...
+        id: checkout
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: flake/dev/diff-plugins.py
+          sparse-checkout-cone-mode: false
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:nixpkgs-unstable
```
khaneliman, line 82: This is something I recently was migrating away from in Home Manager because it leads to an unstable CI. So maybe we'll want to tackle the `flake.lock` wrapper / shim sooner than later.

khaneliman (approved):
Test runs look good, thanks Matt

.github/workflows/pr-merged.yml
MattSturgeon: Good point. Shouldn't be too complicated, depending on how closely we want to mirror how `call-flake.nix` evaluates a flake.lock. i.e. how many edge cases we want to handle (follows, `_2` inputs, etc).

I believe using flake-compat would lead to all inputs being downloaded, so we probably wouldn't want to do that in this case. But we could do something similar to how we _import_ flake-compat, i.e. read directly from the lockfile JSON.

PR #3516, opened 2025-06-28
flake/dev: use nix-shell shebangs for scripts

khaneliman:
Helps with https://github.com/nix-community/nixvim/pull/3513#discussion_r2173435209

image

### DIFF
--- a/flake/dev/devshell.nix
+++ b/flake/dev/devshell.nix
@@ -107,17 +107,17 @@
             }
             {
               name = "locate-lsp-packages";
-              command = ''${pkgs.python3.interpreter} ${./locate-lsp-packages.py}'';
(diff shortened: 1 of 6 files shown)


khaneliman:
> Is it a problem that we're getting these dependencies from the user's nix path, instead of our flake lock?

I think the only way to get it to lock to our inputs would be using `nix develop` but then it evaluates the whole devshell. Or a fetch tarball that matches the hash from the flake.lock ?

GaetanLepage (approved):

MattSturgeon:
I think it'd be possible to write a wrapper, similar to our `default.nix` that loads flake-compat. Such a wrapper could import nixpkgs from the lock file...

But I don't think it's necessarily worth it in this instance 🤔

PR #3517, opened 2025-06-28
flake/dev/diff-plugins: make a package output; ci: use diff-plugins from package output

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/3513#discussion_r2173522304 to see if we can just provide it as a package so it uses our flake lock without the dev shell.


MattSturgeon:
This won't work with a sparse checkout, since we'd need to checkout the flake files (`flake.nix`, `flake.lock`, `flake/`) as well as any file that the flake-parts module eval tries to import...

Non-issue if we don't mind doing a full checkout, though...

khaneliman:
> This won't work with a sparse checkout, since we'd need to checkout the flake files (`flake.nix`, `flake.lock`, `flake/`) as well as any file that the flake-parts module eval tries to import...
> 
> Non-issue if we don't mind doing a full checkout, though...

Sorry, completely forgot it got changed to do that.

PR #3523, opened 2025-07-02
ci: use pinned nixpkgs in CI & shell

MattSturgeon:
Adds a ci script to fetch the pinned nixpkgs from nixvim's `flake.lock`, similar to how we fetch the pinned flake-compat in `default.nix`.

While we _could_ directly use `.nodes.nixpkgs.locked`, it wasn't much effort to do it "properly" and use the input node specified by the root node:

<details><summary>Nix impl</summary>
<p>

```nix
lock = importJSON ../flake.lock;
root = lock.nodes.${lock.root};
nixpkgs = lock.nodes.${root.inputs.nixpkgs};
```

</p>
</details> 

```jq
.nodes[.nodes[.root].inputs.nixpkgs].locked
| "https://github.com/\(.owner)/\(.repo)/archive/\(.rev).tar.gz"
```

This dance ensures we use the _correct_ nixpkgs node, in case we ever accidentally ended up with multiple nixpkgs nodes in our lockfile; e.g. we forget to "follow" another input's nixpkgs input.

In the devshell, we already have `NIXPKGS_PATH`, which is a custom variable defined by numtide/devshell. We can also prefix `NIX_PATH`, so that `<nixpkgs>` points to nixvim's nixpkgs while in the devshell.

In CI, we can instantiate `ci/fetch-nixpkgs.nix` to automatically fetch the pinned nixpkgs and manually add this to `NIX_PATH`.

I have verified that this resolves to the same in-store path that our flake uses:

```
nix-repl> :lf .
Added 31 variables.

nix-repl> inputs.nixpkgs.outPath
"/nix/store/0mq640xwzmwc519d1xs1x8nx3q0v55gy-source"

$ nix-instantiate --eval ./ci/fetch-nixpkgs.nix
"/nix/store/0mq640xwzmwc519d1xs1x8nx3q0v55gy-source"
```

This addresses @khaneliman's concern about potential issues using impure channels in CI.

Tested on my fork here: https://github.com/MattSturgeon/nixvim/actions/runs/16022461295/job/45202450697?pr=23

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -82,7 +82,6 @@ jobs:
       - name: Install Nix
         uses: cachix/install-nix-action@v31
         with:
-          nix_path: nixpkgs=channel:nixos-unstable
           github_access_token: ${{ steps.app-token.outputs.token }}
 
       - name: Create update branch


GaetanLepage (approved):

.github/workflows/pr-merged.yml
```diff
@@ -73,16 +73,24 @@ jobs:
...
         with:
-          nix_path: nixpkgs=channel:nixpkgs-unstable
           extra_nix_config: |
             accept-flake-config = true
 
+      - name: Fetch nixpkgs
+        run: |
+          nixpkgs=$(mktemp)
+          nix-instantiate --raw --eval ./ci/fetch-nixpkgs.nix > "$nixpkgs"
+          echo "NIX_PATH=nixpkgs=$(cat "$nixpkgs")" >> "$GITHUB_ENV"
```
MattSturgeon: Looking at what @khaneliman just added to home-manager, we may be able to get rid of `fetch-nixpkgs.nix` entirely:

https://github.com/nix-community/home-manager/blob/7c455533409411b4053bb6d7db71eb35e0544254/.github/workflows/update-maintainers.yml#L48

(I was under the impression that NIX_PATH needed to refer to local paths, but it seems it can also be a URL)

khaneliman: Yeah, saw this pop up and was wondering if you liked that approach, or not.

MattSturgeon: I like its simplicity. One less file to maintain/checkout.

I'm not sure what all the implications are (pros/cons), e.g. does not specifying the `narHash` have any major downsides for us in CI?

The jq query also doesn't "properly" follow the node input paths, as outlined in this PR's description. However that's a minor issue in practice, since we _should_ never have multiple nixpkgs nodes in our lockfile :crossed_fingers:

khaneliman: > I'm not sure what all the implications are (pros/cons), e.g. does not specifying the `narHash` have any major downsides for us in CI?

Hmm.. that's a good observation... I'm not sure if there's anything that could happen with a MITM in a github action runner or cause any issues that would give a false sense of security. It's not going to be caching anything like a hydra build or cachix build or anything.

MattSturgeon: I've gone with your approach for now.

Tested again here: https://github.com/MattSturgeon/nixvim/actions/runs/16035928139/job/45247572019?pr=24

khaneliman (approved):

PR #3525, opened 2025-07-03
all-maintainers.nix: init; flake/dev/generate-all-maintainers: init 

khaneliman:
Copying implementation from https://github.com/nix-community/home-manager/pull/7350 with tweaks to work in nixvim. https://github.com/nix-community/infra/pull/1837


flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,411 @@
...
+    def find_nix_files(self) -> List[Path]:
+        """Find all .nix files in the modules and plugins directories."""
+        module_files = list(self.modules_dir.rglob("*.nix"))
+        plugin_files = list(self.plugins_dir.rglob("*.nix"))
+        nix_files = module_files + plugin_files
+        print(f"📁 Found {len(module_files)} .nix files in modules, {len(plugin_files)} in plugins (total: {len(nix_files)})")
+        return nix_files
+
+    def extract_maintainer_lines(self, file_path: Path) -> List[str]:
+        """Extract lines containing maintainer references from a file."""
```
MattSturgeon: This is an interesting approach.

I think it'd be much nicer to evaluate an empty `nixvimConfiguration` and use the `meta.maintainers` option value as our source of truth.

khaneliman: Yeah, this was to go as fast as possible without eval slowing it down. It could be refactored to use eval if you can get it go faster. 

EDIT: I'm trying again to see if it was just a way I messed up the implementation before.. I dont think I tried a empty config and was trying to eval each file or something.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,245 @@
...
+
+        except subprocess.TimeoutExpired:
+            print("❌ Timeout while extracting maintainers")
+            sys.exit(1)
+        except Exception as e:
+            print(f"❌ Error extracting maintainers: {e}")
+            sys.exit(1)
+
+    def format_maintainer_entry(self, name: str, info: Dict, source: str) -> str:
+        """Format a single maintainer entry with nix fmt compatible formatting."""
```
MattSturgeon: It's frustrating that we need to output as nix. Ideally the rfc39 impl would also support importing JSON. cc @zowoq 

If you want to avoid all this boilerplate, you could use `lib.generators.toPretty`, although that's not guaranteed to be nixfmt compliant, it _should_ be.

Still, JSON is the "correct" way to store such static/generated data...

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,245 @@
...
+
+        print(f"🏠 Nixvim maintainers: {len(nixvim_maintainers)}")
+        print(f"📦 Nixpkgs maintainers: {len(nixpkgs_maintainers)}")
+
+        with open(self.output_file, "w") as f:
+            f.write("""# Nixvim all maintainers list.
+#
+# This file combines maintainers from:
+# - Nixvim specific maintainers (lib/maintainers.nix)
+# - Nixpkgs maintainers referenced in Nixvim modules
```
MattSturgeon: Would it be simpler to say we list all referenced maintainers, from either nixpkgs or nixvim?

E.g. if there was hypothetically a nixvim maintainer entry that is not maintaining any module, there's no reason to include them in all-maintainers.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,57 @@
+# Extract maintainers from nixvim configuration using meta.maintainers
+# This script evaluates an empty nixvimConfiguration and extracts the merged maintainer information
+let
+  nixvim = import ../../..;
+  pkgs = import <nixpkgs> { };
```
MattSturgeon: If we're importing nixvim above (using flake-compat) we can access `nixvim.inputs.nixpkgs`.

Either way, if we don't need an actual pkgs instance, we can either import `<nixpkgs/lib>` or use `nixvim.inputs.nixpkgs.lib` to get `lib` without evaluating pkgs.

Alternatively, we can get nixvim's extended lib from its empty configuration, since we'll be evaluating that anyway:

```
inherit (nixvim.nixvimConfigurations.${builtins.currentSystem}.default._module.specialArgs) lib;
```

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,57 @@
...
+
+  emptyConfig = lib.nixvim.evalNixvim {
+    modules = [ ];
+    extraSpecialArgs = { };
+    system = "x86_64-linux";
+  };
+
+  inherit (emptyConfig.config.meta) maintainers;
+
+  extractMaintainerObjects = maintainerData: lib.unique (lib.flatten (lib.attrValues maintainerData));
```
MattSturgeon: Does this actually need `flatten` or is `concatLists` ok? Is this nicer as a pipe?

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,57 @@
...
+  allMaintainerObjects = extractMaintainerObjects maintainers;
+
+  getMaintainerName =
+    maintainer:
+    if lib.hasAttr "github" maintainer then
+      maintainer.github
+    else if lib.hasAttr "name" maintainer then
+      maintainer.name
+    else
+      null;
```
MattSturgeon:
```suggestion
    maintainer.github or maintainer.name or null;
```

A) `hasAttr` and `getAttr` are almost always better as `?` and `.`

B) in this case we can just use `or`

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,57 @@
...
+
+  allMaintainerNames = lib.filter (name: name != null) (map getMaintainerName allMaintainerObjects);
+
+  maintainerDetails = lib.listToAttrs (
+    map (obj: {
+      name = getMaintainerName obj;
+      value = obj;
+    }) (lib.filter (obj: getMaintainerName obj != null) allMaintainerObjects)
+  );
```
MattSturgeon: Personal preference: I think a pipe is clearer here

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,57 @@
...
+  nixvimMaintainers = import ../../../lib/maintainers.nix;
+  nixvimMaintainerNames = lib.attrNames nixvimMaintainers;
+  categorizedMaintainers = {
+    nixvim = lib.filterAttrs (name: _: lib.elem name nixvimMaintainerNames) maintainerDetails;
+    nixpkgs = lib.filterAttrs (name: _: !(lib.elem name nixvimMaintainerNames)) maintainerDetails;
```
MattSturgeon: This may also be possible with a `partition` or a `groupBy`. However, that'd require using `attrsToList` and `listToAttrs`.

If we cared about performance here, that'd mean only having a single pass checking the predicate. Here that shouldn't really matter, realistically.

MattSturgeon (commented):
Initial review from my phone, so I may have missed some wider context.

Looks good!

flake/dev/generate-all-maintainers/generate-all-maintainers.py
khaneliman: `lib.generators.toPretty` cleaned up a lot of boilerplate, thanks Matt!

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,93 @@
...
+
+  formatMaintainer =
+    name: info: source:
+    let
+      # Handle identifiers that start with numbers or contain invalid characters
+      quotedName =
+        if lib.match "[0-9].*" name != null || lib.match "[^a-zA-Z0-9_-].*" name != null then
+          ''"${name}"''
+        else
+          name;
```
MattSturgeon: We can use `lib.strings.escapeNixIdentifier` for escaping nix identifiers.

Alternatively, couldn't we let `toPretty` handle this by passing it `{ ${name} = filtered info; }`?

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,93 @@
+# Extract maintainers from nixvim configuration using meta.maintainers
+# This script evaluates an empty nixvimConfiguration and extracts the merged maintainer information
+let
+  nixvim = import ../../..;
+  lib = nixvim.inputs.nixpkgs.lib.extend nixvim.lib.overlay;
+
+  emptyConfig = lib.nixvim.evalNixvim {
+    modules = [ ];
+    extraSpecialArgs = { };
+    system = "x86_64-linux";
```
MattSturgeon: I nearly asked whether this is problematic when running this on other platforms. However I don't think we ever need to evaluate `pkgs` for `meta.maintainers`.

We could alternatively handle this like this:

```nix
modules = [
  # We don't need pkgs, set it to null so we notice if we use it accidentally 
  { _module.args.pkgs = lib.mkForce null; }
  # ...
]
```

or
```nix
# We don't need pkgs, set it to null so we notice if we use it accidentally 
extraSpecialArgs.pkgs = null;

```


That way we ensure we don't accidentally evaluate anything that depends on the pkgs module arg.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,188 @@
...
+        help="Output file path (default: <root>/all-maintainers.nix)",
+    )
+
+    args = parser.parse_args()
+
+    if args.root:
+        nixvim_root = args.root
+    else:
+        script_dir = Path(__file__).parent
+        nixvim_root = script_dir.parent.parent.parent
```
MattSturgeon: If run via flake tooling, this will use the in-store flake, not the local checked out repo.

It is possible to ask `git` where the current worktree root is:
```sh
git rev-parse --show-toplevel
```

Perhaps we could start with the CLI arg, attempt to ask git, fallback to using `__file__` (with a warning) if that fails?

Or is it fine/good for the `nixvim_root` to be the in-store path when using flakes?

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,188 @@
...
+# This file is automatically generated using meta.maintainers from nixvim evaluation
+# DO NOT EDIT MANUALLY
+#
+# To regenerate: nix run .#generate-all-maintainers
+#
+{
+""")
+
+            # Use the formatted maintainers from Nix evaluation
+            print("✨ Adding formatted maintainers using lib.generators.toPretty...")
```
MattSturgeon:
```suggestion
            print("✨ Serialising maintainers using lib.generators.toPretty...")
```

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,93 @@
...
+
+      # Filter out internal fields
+      filteredInfo = lib.filterAttrs (k: v: !lib.hasPrefix "_" k) info;
+    in
+    "  # ${source}\n  ${quotedName} = ${
+        lib.generators.toPretty {
+          multiline = true;
+          indent = "    ";
+        } filteredInfo
+      };";
```
MattSturgeon: By including the `name` in the serialised value, we don't have to escape it:

```suggestion
    "  # ${source}\n"
    + lib.generators.toPretty
        {
          multiline = true;
          indent = "  ";
        }
        { ${name} = filteredInfo; }
    + ";";
```

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,93 @@
...
+    maintainerData:
+    lib.pipe maintainerData [
+      lib.attrValues
+      lib.concatLists
+      lib.unique
+    ];
+
+  allMaintainerObjects = extractMaintainerObjects maintainers;
+
+  getMaintainerName = maintainer: maintainer.github or maintainer.name or null;
```
MattSturgeon: Since this "name" is actually insignificant and ignored, it only really needs to be any unique key.

We could theoretically just hash the entire value and use that as a name, although in practice we only really care about maintainers that have a `github` and if _that_ isn't unique we have bigger issues.

all-maintainers.nix
```diff
@@ -0,0 +1,399 @@
...
+      name = "Nathan Felber";
+    };
+  # nixpkgs
+  nwjsmith = {
+      email = "nate@theinternate.com";
+      github = "nwjsmith";
+      githubId = 1348;
+      name = "Nate Smith";
+    };
+  # nixpkgs
```
MattSturgeon: Q: would it be problematic to include a `source` key in the maintainer attrs?

I assume the rfc39 impl would just ignore any attrs it doesn't care about, rather than validating them.

I think it'd be nicer (and simpler) to have `source = "nixpkgs"` instead of a prefix comment.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
MattSturgeon: In fact, if we ditched the prefix comment, we could massively simplify by using `toPretty` to serialise the entire file all in one go, instead of one-at-a-time.

all-maintainers.nix
MattSturgeon, line 1: Can we move this into `lib`, so it sits next to `maintainers.nix`?

Alternatively, can we put it in `generated/`?

I don't think there's a strong reason to have this at the top-level, especially if we won't be using it internally and it is only used by external tooling.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,147 @@
...
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'nix-community' || github.event_name == 'workflow_dispatch'
+    steps:
+      - name: Create GitHub App token
+        uses: actions/create-github-app-token@v2
+        if: vars.CI_APP_ID
+        id: app-token
+        with:
+          app-id: ${{ vars.CI_APP_ID }}
+          private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
```
MattSturgeon, line 41: I still need to get around to doing this in the other workflows, but ideally we would be explicit about the permissions we need here.

Also the `github.token` permissions too (which I assume will be the same). _That_ permissions block is a good place to put a comment explaining why we support both `github.token` and the app token.

I assume here we need `contents: write` and `pull_requests: write`?

.github/workflows/update-maintainers.yml
MattSturgeon, line 1: Style nit: in nixvim the workflows usually have more blank lines to separate ideas; e.g. between top-level objects (`name`, `on`, `jobs`), and usually between individual `steps`

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,147 @@
...
+            --body "$pr_body" \
+            --label "dependencies" \
+            --label "maintainers"
+      - name: Summary
+        env:
+          has_changes: ${{ steps.check-changes.outputs.has_changes }}
+          changes: ${{ steps.check-changes.outputs.changes_summary }}
+        run: |
+          if [[ "$has_changes" == "true" ]]; then
+            echo "✅ Successfully created PR with maintainer updates"
```
MattSturgeon: It would be awesome if this linked to the new PR :wink:

I _think_ `gh pr create` will usually print the URL, however you can also get info for the current branch's PR using `gh pr view --json`

You can specify the specific fields you wish to include in your json object, and even use `--jq` to format as GITHUB_OUTPUT:

```sh
$ gh pr view \
  --json url \
  --json number \
  --jq 'to_entries[] | "\(.key)=\(.value)"'
number=1234
url=https://github.com/nix-community/nixvim/pull/1234
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,147 @@
...
+            echo "has_changes=true" >> "$GITHUB_OUTPUT"
+            added=$(git diff --numstat all-maintainers.nix | cut -f1)
+            removed=$(git diff --numstat all-maintainers.nix | cut -f2)
+            echo "changes_summary=+$added -$removed lines" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Validate generated file
+        if: steps.check-changes.outputs.has_changes == 'true'
+        run: |
+          echo "🔍 Validating generated all-maintainers.nix..."
+          if nix eval --file ./all-maintainers.nix --json > /dev/null; then
```
MattSturgeon: Do we want to prefer using `nix-instantiate --eval` over `nix eval`? Both do the same thing, most of the time. Using the "classic" command makes it explicit that we aren't doing anything flake related, so we don't need to copy anything to store.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,147 @@
...
+      - name: Setup Git
+        run: |
+          git config user.name "${{ steps.user-info.outputs.name }}"
+          git config user.email "${{ steps.user-info.outputs.email }}"
+      - name: Generate updated maintainers list
+        run: |
+          echo "📋 Generating updated all-maintainers.nix..."
+          nix run .#generate-all-maintainers -- --root . --output ./all-maintainers.nix
+          echo "🎨 Formatting with nixfmt..."
+          nix fmt all-maintainers.nix
```
MattSturgeon: Do we want to ["group"](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines) these?

```sh
echo "::group::📋 Generating updated all-maintainers.nix..."
nix run .#generate-all-maintainers -- --root . --output ./all-maintainers.nix
echo "::endgroup::"
echo "::group::🎨 Formatting with nixfmt..."
nix fmt all-maintainers.nix
echo "::endgroup::"
```

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,188 @@
...
+            nixvim_root
+            / "flake"
+            / "dev"
+            / "generate-all-maintainers"
+            / "extract-maintainers-meta.nix"
+        )
+
+    def extract_maintainers_from_meta(self) -> Dict:
+        """Extract maintainer information using meta.maintainers."""
+        print("🔍 Extracting maintainers using meta.maintainers...")
```
MattSturgeon, line 61: Are we "extracting" maintainers, or "evaluating" them?

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,188 @@
...
+by evaluating an empty nixvimConfiguration, which is much simpler and more
+reliable than parsing files with regex.
+"""
+
+import argparse
+import json
+import subprocess
+import sys
+from pathlib import Path
+from typing import Dict
```
GaetanLepage:
```suggestion
```
You can use `dict` instead (doesn't require importing anything).

.github/workflows/update-maintainers.yml
khaneliman: I can add them back... just usually prefer to let my lsp format it.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
khaneliman: Couldn't get either of these to work. They both cause the eval to fail 

```bash
❌ Failed to extract maintainers: error:
       … while evaluating attribute 'categorized'
         at /Users/khaneliman/github/nixvim/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix:76:3:
           75|   details = maintainerDetails;
           76|   categorized = categorizedMaintainers;
             |   ^
           77|   formatted = formatAllMaintainers;

       … while evaluating attribute 'nixpkgs'
         at /Users/khaneliman/github/nixvim/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix:45:5:
           44|     nixvim = lib.listToAttrs partitionedMaintainers.right;
           45|     nixpkgs = lib.listToAttrs partitionedMaintainers.wrong;
             |     ^
           46|   };

       … while evaluating definitions from `/nix/store/6s1w74lcgfp6ird1pw8p841adnpw30ar-source/modules/top-level/files'
```

zowoq:
I've renamed `nixvim` team to `nixvim-committers` and created `nixvim-maintainers` with triage permissions.

https://github.com/orgs/nix-community/teams/nixvim-committers
https://github.com/orgs/nix-community/teams/nixvim-maintainers

https://github.com/nix-community/infra/pull/1886

Also added "Organization permissions" -> "Members" -> "Read-only" to the `nixvim-ci` app.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
khaneliman: Part of extracting them is evaluating the files, but we aren't' just running an eval. We're returning data after we have run the eval.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
khaneliman: This is just a fallback value when an argument isn't provided to speed up quick testing when calling the file itself, CI provides the `root` and `out` paths to the package.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
khaneliman: Image

This worked to get around eval issues..

GaetanLepage: Isn't such a patch overkill? (well done in writing it though haha)

.github/workflows/update-maintainers.yml
MattSturgeon: > just usually prefer to let my lsp format it.

If there's a decent way to format workflows, I'm keen to have that in our treefmt config.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
MattSturgeon: Yes I know it's a fallback.

But does the current fallback value (usually an in-store copy of the flake) actually make sense?

---

EDIT: looks like it doesn't:
```
$ nix run .#generate-all-maintainers
Error: Could not find maintainers.nix in /nix
Please specify --root or run from Nixvim directory
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+            added=$(git diff --numstat all-maintainers.nix | cut -f1)
+            removed=$(git diff --numstat all-maintainers.nix | cut -f2)
+            echo "changes_summary=+$added -$removed lines" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Validate generated file
+        if: steps.check-changes.outputs.has_changes == 'true'
+        run: |
+          echo "🔍 Validating generated all-maintainers.nix..."
+          if nix-instantiate --eval --file ./all-maintainers.nix --json > /dev/null; then
```
MattSturgeon:
```suggestion
          if nix-instantiate --eval generated/all-maintainers.nix --json > /dev/null; then
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+          nix run .#generate-all-maintainers -- --root . --output ./all-maintainers.nix
+          echo "::endgroup::"
+          echo "::group::🎨 Formatting with nixfmt..."
+          nix fmt all-maintainers.nix
+          echo "::endgroup::"
+
+      - name: Check for changes
+        id: check-changes
+        run: |
+          if git diff --quiet all-maintainers.nix; then
```
MattSturgeon:
```suggestion
          if git diff --quiet generated/all-maintainers.nix; then
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+
+      - name: Setup Git
+        run: |
+          git config user.name "${{ steps.user-info.outputs.name }}"
+          git config user.email "${{ steps.user-info.outputs.email }}"
+
+      - name: Generate updated maintainers list
+        run: |
+          echo "::group::📋 Generating updated all-maintainers.nix..."
+          nix run .#generate-all-maintainers -- --root . --output ./all-maintainers.nix
```
MattSturgeon:
```suggestion
          nix run .#generate-all-maintainers -- --root . --output generated/all-maintainers.nix
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+            - Nixvim specific maintainers from lib/maintainers.nix
+            - Nixpkgs maintainers referenced in Nixvim modules
+
+            Changes: ${{ steps.check-changes.outputs.changes_summary || 'No content changes' }}
+
+            Generated by: flake/dev/generate-all-maintainers/generate-all-maintainers.py
+          pr_body: |
+            ## 📋 Summary
+
+            This PR updates the master maintainers list (`all-maintainers.nix`) which combines:
```
MattSturgeon:
```suggestion
            This PR updates the master maintainers list (`generated/all-maintainers.nix`) which combines:
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+
+            - **Generated by:** `flake/dev/generate-all-maintainers/generate-all-maintainers.py`
+            - **Trigger:** ${{ github.event_name == 'schedule' && 'Scheduled weekly update' || 'Manual workflow dispatch' }}
+            - **Validation:** File syntax verified with `nix-instantiate --eval`
+
+            ## 📚 Usage
+
+            This file can be imported and used for maintainer lookups:
+            ```nix
+            let allMaintainers = import ./all-maintainers.nix; in
```
MattSturgeon:
```suggestion
            let allMaintainers = import ./generated/all-maintainers.nix; in
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+            # Access any maintainer by name: allMaintainers.username
+            ```
+
+            ---
+            🤖 *This PR was automatically created by the [update-maintainers workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})*
+        run: |
+          branch_name="maintainers/update-$(date +%Y%m%d%H%M%S)"
+          git checkout -b "$branch_name"
+
+          git add all-maintainers.nix
```
MattSturgeon:
```suggestion
          git add generated/all-maintainers.nix
```

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+            else
+              echo "❌ Failed to create pull request."
+              echo "### ❌ PR Creation Failed" >> $GITHUB_STEP_SUMMARY
+              echo "A pull request was intended but the URL was not captured. Please check the logs." >> $GITHUB_STEP_SUMMARY
+            fi
+          else
+            echo "ℹ️ No changes detected - maintainers list is up to date."
+            echo "### ℹ️ No Changes" >> $GITHUB_STEP_SUMMARY
+            echo "The maintainers list is up-to-date. No PR was created." >> $GITHUB_STEP_SUMMARY
+          fi
```
MattSturgeon: Almost all the variables referenced in this run script are not defined.

E.g. `SHOULD_CREATE_PR`, `PR_URL`, `CHANGES_SUMMARY`, etc.

You do set `has_changes` and `changes`, but these are not used.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+
+          git push origin "$branch_name"
+
+          pr_url=$(gh pr create \
+            --title "$title" \
+            --body "$pr_body" \
+            --label "dependencies" \
+            --label "maintainers")
+
+          echo "pr_url=$pr_url" >> "$GITHUB_OUTPUT"
```
MattSturgeon: This step doesn't have an `id`, so you won't be able to reference its outputs.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
MattSturgeon: The issue was coming from option-checking, where the module system checks all config definitions correspond to actually declared options. Some of our options/configs depend on things like `pkgs.runCommand`.

Disabling checking works:

```diff
diff --git a/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix b/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
index f44ae3f9..8ac79ff8 100644
--- a/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
+++ b/flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
@@ -5,9 +5,12 @@ let
   lib = nixvim.inputs.nixpkgs.lib.extend nixvim.lib.overlay;
 
   emptyConfig = lib.nixvim.evalNixvim {
-    modules = [ ];
-    extraSpecialArgs = { };
-    system = "x86_64-linux";
+    modules = [
+      { _module.check = false; }
+    ];
+    extraSpecialArgs = {
+      pkgs = null;
+    };
   };
 
   inherit (emptyConfig.config.meta) maintainers;
```

And should give a tiny performance boost too.

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,67 @@
...
+  nixvimMaintainerNames = lib.attrNames nixvimMaintainers;
+  partitionedMaintainers = lib.partition (nameValue: lib.elem nameValue.name nixvimMaintainerNames) (
+    lib.attrsToList maintainerDetails
+  );
+
+  maintainerDetails = lib.pipe allMaintainerObjects [
+    (map (obj: {
+      name = obj.github;
+      value = obj // {
+        source = if lib.elem obj.github nixvimMaintainerNames then "nixvim" else "nixpkgs";
```
MattSturgeon: How about re-using the already evaluated partition? Doing attribute name lookups should be faster than `elem`.

```suggestion
        source =
          if categorizedMaintainers.nixvim ? ${obj.github} then
            "nixvim"
          else if categorizedMaintainers.nixpkgs ? ${obj.github} then
            "nixpkgs"
          else
            throw "${obj.github} is neither a nixvim or nixpkgs maintainer";
```

.github/workflows/update-maintainers.yml
khaneliman: `yamlfmt` has an annoying bug of infinite inserting blank lines on save, though... so it would suck to be auto enforced. https://github.com/google/yamlfmt/issues/169 
image

flake/dev/generate-all-maintainers/extract-maintainers-meta.nix
```diff
@@ -0,0 +1,67 @@
...
+  ];
+
+  categorizedMaintainers = {
+    nixvim = lib.listToAttrs partitionedMaintainers.right;
+    nixpkgs = lib.listToAttrs partitionedMaintainers.wrong;
+  };
+
+  formattedMaintainers = lib.generators.toPretty {
+    multiline = true;
+    indent = "  ";
```
MattSturgeon: Now you are formatting the whole attrset, I believe we want the default indentation `""`.


Currently it is generating with 4 spaces, then 2:

```nix
{
    "347Online" = {
      email = "katiejanzen@347online.me";
      github = "347Online";
      githubId = 38228372;
      name = "Katie Janzen";
      source = "nixpkgs";
    };
```

```suggestion
```

.github/workflows/update-maintainers.yml
MattSturgeon: Also:

Testing locally, I believe we also need `--strict` when using nix-instantiate.

It seems `--json` implies strict for `nix eval`, but not for `nix-instantiate`.

When using `--json` with `nix-instantiate` without `--strict` it tries to serialise thunks as json, which fails.

---

For the record, I wasn't asking you to switch to/from `nix eval` or `nix-instantiate`, I was just bringing up their similarities and differences and asking which we preferred for this use case.

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,210 @@
...
+""")
+
+        self.validate_generated_file()
+        self.print_statistics(maintainer_data)
+
+    def validate_generated_file(self) -> bool:
+        """Validate the generated Nix file syntax."""
+        try:
+            result = subprocess.run(
+                ["nix-instantiate", "--eval", str(self.output_file)],
```
MattSturgeon: Again, I believe we also need `--strict` here, or it will lazily evaluate to a set of thunk values.

Do we also want to redirect the output to /dev/null ?

khaneliman:
Test run https://github.com/khaneliman/nixvim/actions/runs/16123566518/job/45495054506 I don't have git config setup for PR on fork atm But, should work when the `user-info` step returns the app information, like in HM. Something still isn't right with generating the file in this refactoring here, though.. taking a look.

khaneliman:
https://github.com/khaneliman/nixvim/actions/runs/16123952334/job/45496299935 alright, have to capture the output from validation.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+            ---
+            🤖 *This PR was automatically created by the [update-maintainers workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})*
+        run: |
+          branch_name="maintainers/update-$(date +%Y%m%d%H%M%S)"
+          git checkout -b "$branch_name"
+
+          git add generated/all-maintainers.nix
+          git commit -m "$title" -m "$commit_body"
+
+          git push origin "$branch_name"
```
MattSturgeon: Do we need `--set-upstream` here, for `gh pr create` to detect this has been pushed already?

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+          private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+
+      - name: Get GitHub App user info
+        id: user-info
+        if: vars.CI_APP_ID
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          slug: ${{ steps.app-token.outputs.app-slug }}
```
MattSturgeon, line 51: Here, we could do something like:

```yaml
- name: Get GitHub App user info
  id: user-info
  env:
    GH_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
    slug: ${{ steps.app-token.outputs.app-slug || 'github-actions' }}
    if: ${{ vars.CI_APP_ID && '' || '41898282' }}
  run: |
    name="$slug[bot]"
    if [ -z "$id" ]; then
      id=$(gh api "/users/$name" --jq .id)
    fi

    # ...
```

MattSturgeon: I'm guessing at some point we'll want to extract this "gen token and setup git" logic into a composite action?

The upside is that we can maintain the logic in a central place. The downside is we need to checkout the repo (or at least the `.github/actions` directory) before using a local composite action.

MattSturgeon: > Here, we could do something like:

Or it might be simpler to handle that in the `setup-git` step's `env` inputs

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,179 @@
...
+        with:
+          token: ${{ steps.app-token.outputs.token || github.token }}
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+
+      - name: Setup Git
+        run: |
+          git config user.name "${{ steps.user-info.outputs.name }}"
+          git config user.email "${{ steps.user-info.outputs.email }}"
```
MattSturgeon:
```suggestion
      - name: Setup Git
        env:
          name: ${{ steps.user-info.outputs.name || 'github-actions[bot]' }}
          email: ${{ steps.user-info.outputs.email || '41898282+github-actions[bot]@users.noreply.github.com' }}
        run: |
          git config user.name "$name"
          git config user.email "$email"
```

1. Always use env to inject values, rather than directly embedding `${{ }}` templates into bash
2. This should be a better fix than my [other suggestion](https://github.com/nix-community/nixvim/pull/3525#discussion_r2190728564)

.github/workflows/update-maintainers.yml
khaneliman: > I'm guessing at some point we'll want to extract this "gen token and setup git" logic into a composite action?

Was thinking about that too while bouncing around all these workflows.

.github/workflows/update-maintainers.yml
MattSturgeon: Looks like this was reverted last push?

khaneliman: Sorry, testing and cherry-picking stuff from my main branch in my fork.

khaneliman:
Alright, after doing some testing https://github.com/khaneliman/nixvim/actions/runs/16126857467 I made it work better with existing PRs

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+                repository(owner: $owner, name: $repo) {
+                  pullRequests(first: 1, states: OPEN, headRefName: $branch) {
+                    nodes {
+                      number
+                      url
+                    }
+                  }
+                }
+              }
+            ' | jq --raw-output '
```
MattSturgeon, line 138: FYI `gh api` also has a `--jq` flag:
```suggestion
            ' --jq '
```

Sometimes it's still worth it to pipe to jq though, e.g. if you want to use `--exit-status` to trigger a failure if the jq query doesn't match the input JSON.


---


We use a similar graphql query elsewhere, but one issue I'm now spotting is that `headRefName` could also match PRs on forks that use the same branch name... I don't see an obvious quick fix in the graphql explorer, and the issue applies to both places, so this is fine for now.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+          changes: ${{ steps.check-changes.outputs.changes_summary }}
+          pr_url: ${{ steps.create-pr.outputs.pr_url}}
+          pr_num: ${{ steps.open_pr_info.outputs.number }}
+        run: |
+          if [[ "$has_changes" == "true" ]]; then
+              if [[ -n "$pr_num" ]]; then
+                echo "✅ Successfully updated PR with new changes."
+                echo "$changes"
+                echo "🔗 PR URL: $pr_url"
+                echo "### ✅ PR Updated: [$pr_url]($pr_url)" >> $GITHUB_STEP_SUMMARY
```
MattSturgeon:
```suggestion
                echo "### ✅ PR Updated" >> $GITHUB_STEP_SUMMARY
                echo "[$pr_url]($pr_url)" >> $GITHUB_STEP_SUMMARY
```

?

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+              elif [[ -n "$pr_url" ]]; then
+                echo "✅ Successfully created PR with maintainer updates."
+                echo "$changes"
+                echo "🔗 PR URL: $pr_url"
+                echo "### ✅ PR Created: [$pr_url]($pr_url)" >> $GITHUB_STEP_SUMMARY
```
MattSturgeon: (same)

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+          git commit -m "$title" -m "$commit_body"
+
+          echo "Pushing to remote branch $pr_branch"
+          git push --force --set-upstream origin "$pr_branch"
+
+          if [ -z "$pr_num" ]; then
+            echo "Creating new pull request."
+            PR_URL=$(gh pr create \
+              --title "$title" \
+              --body "$pr_body")
```
MattSturgeon: Personally, I find subshells, command substitution, etc clearer like this:


```suggestion
            PR_URL=$(
              gh pr create \
                --title "$title" \
                --body "$pr_body"
            )
```

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,208 @@
...
+#
+""")
+
+            # Use the formatted maintainers from Nix evaluation
+            print(
+                "✨ Serializing formatted maintainers using lib.generators.toPretty..."
+            )
+            f.write(formatted_maintainers)
+            f.write("""
+""")
```
MattSturgeon:
```suggestion
            f.write("\n")
```

flake/dev/generate-all-maintainers/generate-all-maintainers.py
```diff
@@ -0,0 +1,208 @@
...
+            f.write("""# Nixvim all maintainers list.
+#
+# This file lists all referenced maintainers in Nixvim.
+#
+# This file is automatically generated using meta.maintainers from nixvim evaluation
+# DO NOT EDIT MANUALLY
+#
+# To regenerate: nix run .#generate-all-maintainers
+#
+""")
```
MattSturgeon: We can use [`inspect.cleandoc()`](https://docs.python.org/3/library/inspect.html#inspect.cleandoc) or [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) to allow us to format this nicer:


```suggestion
            f.write(inspect.cleandoc("""
              # Nixvim all maintainers list.
              #
              # This file lists all referenced maintainers in Nixvim.
              #
              # This file is automatically generated using meta.maintainers from nixvim evaluation
              # DO NOT EDIT MANUALLY
              #
              # To regenerate: nix run .#generate-all-maintainers
              #
            """))
```

See https://stackoverflow.com/a/48112903 for the subtle differences between `cleandoc` and `dedent`.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+      create_pr:
+        description: "Create PR even if no changes"
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  update-maintainers:
+    runs-on: ubuntu-latest
+    # Creating PRs to update the generated maintainers file
```
MattSturgeon: Should this comment be at the head of the file? Immediately before the job? It feels out of place here before the `permissions` block.

This _would_ be a good location to mention that we explicitly support `github.token` for easier testing on forks, though.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+    env:
+      pr_branch: update/maintainers-${{ github.ref_name }}
+    if: github.repository_owner == 'nix-community' || github.event_name == 'workflow_dispatch'
```
MattSturgeon: Personally, I try to keep the conditional close to the start, usually immediately after the job name and/or `runs-on`.

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,245 @@
+name: Update maintainers list
+
+on:
+  push:
+    branches: ["main"]
+    paths: ["lib/maintainers.nix"]
```
MattSturgeon: I'd personally prefer:


```suggestion
    branches: main
    paths: lib/maintainers.nix
```

or


```suggestion
    branches:
      - main
    paths:
      - lib/maintainers.nix
```

MattSturgeon (approved):
No issues with the diff from me. And I trust your testing process.

Minor comments below, all either trivial or subjective. They can all be ignored if preferred.

.github/workflows/update-maintainers.yml
khaneliman: It was to explain the permissions

.github/workflows/update-maintainers.yml
```diff
@@ -0,0 +1,254 @@
...
+        description: "Create PR even if no changes"
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  update-maintainers:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'nix-community' || github.event_name == 'workflow_dispatch'
+    # Permissions required for workflow
```
MattSturgeon, line 24: Nit: this makes it sound like this is the canonical place where the permissions are defined. However we actually have them in two places, and this is the less important of the two since it is only used on forks that don't have an APP_ID.

MattSturgeon (approved):

.github/workflows/update-maintainers.yml
khaneliman: Not sure why this would be less important? These are required for a token to get generated with permissions required, the others are just to limit the scope of permissions granted to an app.

MattSturgeon: These dictate the permissions of `github.token` and are entirely unrelated to the app token.

I said they're less important because `github.token` is only used when there's no app token.

khaneliman: >These dictate the permissions of github.token and are entirely unrelated to the app token.
>> the others are just to limit the scope of permissions granted to an app.

Understood, as part of previous response.

Guess its differing opinion on what's **important**. I think declaring the app token permissions as less because they are optional for functionality. But, the app token permissions are just a couple lines lower with the exact same naming convention and didn't think it warranted copy/paste for same explanation.

Issue #3518, opened 2025-06-29
Performance improvement


gepbird:
## Problem

I'm a new NixVim user, previously I was symlinking my `~/.config/nvim` with NixOS/home-manager. After changing a lua file and restarting neovim, it was immediately applied.

With NixVim, it takes 31s on my machine to build the (standalone) package after changing some configuration, when experimenting with new plugins, this gets rather frustrating very quickly.

Out of that 31s, ~15s were spent on evaluating Nix code, 14s spent on building the wrapped Neovim package, and the rest ~2s was spent on some minor things.

## Improvement: skip fixup phase

We can get rid of most of the time spent on building the wrapped Neovim package by skipping the fixup phase, for me this improves the overall build time from 31s to 18s (-42% !)

For just testing the changes, I didn't see any downsides with that, AFAIK it skips shrinking, stripping, running autoPatchELF, patching shebangs, gzipping man pages, auditing tmpdir, separating debug info and updating some autootols config.

After some reverse engineering I got to this implementation: https://github.com/gepbird/nvim/commit/6e7420fbb4dbf404f5f4b33ced45fadb1edaaa2a. It overrides `wrapNeovimUnstable` to skip the fixup phase in the wrapper derivation, and passes this patched `pkgs` to NixVim.

## Improvement: removing extra derivations

Adding a simple `enableMan = false;` NixVim config improves the eval time from 13.5s to 6.6s. Removing `printInitPackage` decreases it to 5.5s (if you have other packages installed from nixvim like a language server, this change may not help and much: instead of 6.8s -> 5.5s improvement, you will get 8.2s -> 8.15s improvement). With all of these changes, the overall build time is 11.5s.

## Alternative solution

With [impureRtp](https://nix-community.github.io/nixvim/NeovimOptions/index.html?highlight=impure#impurertp), I could do the experimenting part in `~/.config/nvim` and when I'm done rewrite it in NixVim but I'm not a big fan of this workflow.

saygo-png:
I think everything mentioned here has been implemented.

gepbird:
> I think everything mentioned here has been implemented.

Yes, thanks for working on them :)

Also if anyone else comes here, I found https://github.com/nix-community/nixvim/pull/3664#issuecomment-3263229934 pretty useful for quickly iterating on the config without having to wait some seconds for the Nix eval and build.

PR #3529, opened 2025-07-05
top-level/output: add enablePrintInit option

gepbird:
This PR works towards https://github.com/nix-community/nixvim/issues/3518

When not including `printInitPackage`, evaluation time and build time gets a small boost

I ran a small benchmark of changing some Nixvim config then building it, 5 times for each scenario. On my "dev" nixvim package (which includes the tweaks in the aforementioned PR) this improved the time from 12.97s to 12.30s (-5.16%) and my "prod" nixvim package (which is likely what most people use) it goes down from 33.91s to 33.06s (-2.5%)

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -81,6 +81,17 @@ in
       '';
     };
 
+    enablePrintInit = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description = ''
+        Install a tool that shows the content of the generated `init.lua` file.
+
+        Run it using `${config.build.printInitPackage.meta.mainProgram}`.
+      '';
+    };
+
     build = {
       # TODO: `standalonePackage`; i.e. package + printInitPackage + man-docs bundled together
 
@@ -287,10 +298,12 @@ in
             with config.build;
             [
               nvimPackage
-              printInitPackage
             ]
             ++ lib.optionals config.enableMan [
               manDocsPackage
+            ]
+            ++ lib.optionals config.enablePrintInit [
+              printInitPackage
             ];
           meta.mainProgram = "nvim";
         };


modules/top-level/output.nix
```diff
@@ -81,6 +81,15 @@ in
       '';
     };
 
+    enablePrintInit = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Install a tool that shows the content of the generated `init.lua` file.
+        Run it using `${config.build.printInitPackage.meta.mainProgram}`.
```
MattSturgeon, line 91: Generally, option docs should not depend on anything from the _actual_ `config`.

I'd recommend explicitly referencing the default, so that we aren't using a potentially dynamic value:

```suggestion
        Run it using `${options.build.printInitPackage.default.meta.mainProgram}`.
```

modules/top-level/output.nix
```diff
@@ -81,6 +81,15 @@ in
       '';
     };
 
+    enablePrintInit = mkOption {
+      type = types.bool;
+      default = true;
```
MattSturgeon, line 86:
```suggestion
      default = true;
      example = false;
```

modules/top-level/output.nix
```diff
@@ -81,6 +81,15 @@ in
...
+      description = ''
+        Install a tool that shows the content of the generated `init.lua` file.
```
MattSturgeon, line 89: Should we split this into two paragraphs?


```suggestion
        Install a tool that shows the content of the generated `init.lua` file.

```

MattSturgeon (commented):
Great addition!

modules/top-level/output.nix
gepbird: Thanks for the quick review!

I mostly copied this from:

 https://github.com/nix-community/nixvim/blob/239d331bb48673dfd00d7187654892471cd60d44/modules/top-level/output.nix#L150-L158

I don't quite understand how should that work, can you expand on it?

I get this error when trying to apply your change:
```console
       error: attribute 'default' missing
       at /nix/store/fdjb71pknf5wg4l3v60cy12cjsdawx6q-source/modules/top-level/output.nix:92:56:
           91|
           92|         Run it using `${options.build.printInitPackage.default.meta.mainProgram}`.
             |                                                        ^
           93|       '';
```

MattSturgeon: > I mostly copied this from:
> 
>  https://github.com/nix-community/nixvim/blob/239d331bb48673dfd00d7187654892471cd60d44/modules/top-level/output.nix#L150-L158

If we already have the same thing a few lines down in the same file, then this is fine. We can fix both in one go in another PR another time.

> I don't quite understand how should that work, can you expand on it?
> 
> I get this error when trying to apply your change:
> ```console
>        error: attribute 'default' missing
>        at /nix/store/fdjb71pknf5wg4l3v60cy12cjsdawx6q-source/modules/top-level/output.nix:92:56:
>            91|
>            92|         Run it using `${options.build.printInitPackage.default.meta.mainProgram}`.
>              |                                                        ^
>            93|       '';
> ```

Looks like we don't actually set the option's `default` while declaring the `printInitPackage` option, so my suggestion won't work as there is no `default` attr in this case.

MattSturgeon (approved):

Issue #2418, opened 2024-10-15
[BUG] FTPlugin for markdown (tab size 2)


anotherhadi:
| Field        | Description |
|--------------|-------------|
| Nixpkgs      | `unstable` | 
| Home Manager | `unstable` | 

- [ x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

For my markdowns files I want to set tabs size to 2 spaces.
I tried to set it using ftplugins but it looks like it has no effect on my configuration. (The file is created in the ftplugin directory)

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
    extraFiles = {
      "ftplugin/markdown.lua".text = ''
        vim.opt.tabstop = 2
        vim.opt.shiftwidth = 2
        vim.opt.expandtab = true 
      '';
    };
}
```

stasjok:
This is not specific to NixVim. First of all, for buffer-local options, you should use `vim.bo` or `vim.opt_local` (or it'll be remembered and used as default values for all new buffers). Secondly, to extend the default ftplugin, use the "after" directory. For more information, see `:h ftplugin`. Try this:
```nix
programs.nixvim = {
    extraFiles = {
      "after/ftplugin/markdown.lua".text = ''
        vim.bo.tabstop = 2
        vim.bo.shiftwidth = 2
        vim.bo.expandtab = true 
      '';
    };
}
```

anotherhadi:
> This is not specific to NixVim. First of all, for buffer-local options, you should use `vim.bo` or `vim.opt_local` (or it'll be remembered and used as default values for all new buffers). Secondly, to extend the default ftplugin, use the "after" directory. For more information, see `:h ftplugin`. Try this:
> 
> ```nix
> programs.nixvim = {
>     extraFiles = {
>       "after/ftplugin/markdown.lua".text = ''
>         vim.bo.tabstop = 2
>         vim.bo.shiftwidth = 2
>         vim.bo.expandtab = true 
>       '';
>     };
> }
> ```

Now working, awesome thanks!

PR #3535, opened 2025-07-07
Fix example for option `extraFiles` on file `ftplugin/nix.lua`

n0099:
Following https://github.com/nix-community/nixvim/issues/2418#issuecomment-2413714276
https://vi.stackexchange.com/questions/26169/difference-between-adding-vim-after-ftplugin-vs-vim-ftplugin/26170#26170

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -102,10 +102,10 @@ in
       default = { };
       example = lib.literalExpression ''
         {
-          "ftplugin/nix.lua".text = '''
-            vim.opt.tabstop = 2
-            vim.opt.shiftwidth = 2
-            vim.opt.expandtab = true
+          "after/ftplugin/nix.lua".text = '''
+            vim.opt_local.tabstop = 2
+            vim.opt_local.shiftwidth = 2
+            vim.opt_local.expandtab = true
           ''';
         }
       '';


modules/files.nix
```diff
@@ -102,10 +102,10 @@ in
       default = { };
       example = lib.literalExpression ''
         {
-          "ftplugin/nix.lua".text = '''
-            vim.opt.tabstop = 2
-            vim.opt.shiftwidth = 2
-            vim.opt.expandtab = true
+          "after/ftplugin/nix.lua".text = '''
+            vim.bo.tabstop = 2
```
MattSturgeon: While `vim.bo` is the buffer local version of `vim.o`, I believe `vim.opt_local` is the buffer local version of `vim.opt`.

I don't think it makes a difference for primitive values, but I think we normally use `vim.opt` and its siblings in examples because it has better support for lua tables as values.

modules/files.nix
```diff
@@ -102,10 +102,10 @@ in
       default = { };
       example = lib.literalExpression ''
         {
-          "ftplugin/nix.lua".text = '''
-            vim.opt.tabstop = 2
-            vim.opt.shiftwidth = 2
-            vim.opt.expandtab = true
+          "after/ftplugin/nix.lua".text = '''
```
MattSturgeon, line 105: I think the original intent of this example was to show a simple file being added. Moving the file to `after` makes the example more complex, IMO.

I assume the reason you've done this is because this specific example of setting options without using `after` was causing you issues in some specific scenario? E.g. a scenario where you set the same options in your main nixvim config?

I wonder it it'd be better to instead come up with a different example, or to have an example that writes both `after` and non-`after` files?

MattSturgeon (commented):
I need a bit more context to be convinced this is actually better; the idea was to have a somewhat realistic, but importantly __simple__ example of an extra file. Maybe the example of setting options in a ftplugin file was just a bad example.

Additionally, we need to fix the commit message. In nixvim we use path-based commit scoping, e.g.:

```
modules/files: improve `extraFiles` example
```

modules/files.nix
n0099: https://vi.stackexchange.com/questions/26169/difference-between-adding-vim-after-ftplugin-vs-vim-ftplugin/26170#26170

modules/files.nix
n0099: https://old.reddit.com/r/neovim/comments/stxqk2/buffer_options/hx7t8ea/

modules/files.nix
MattSturgeon: Yes, I'm aware of what `after` is.

MattSturgeon (approved):
My preference would be to come up with another simpler example, however this is fine for now. Thanks!

PR #3537, opened 2025-07-07
update adapters and opts, cleanup prompt example

auipga:
I found some missing adapters [here](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua) and an obsolete `opts` part, that doesn't exist [here](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua#L533-L543).

The maintainer should look at it. Maybe the differences are intentional.

### DIFF
--- a/plugins/by-name/codecompanion/default.nix
+++ b/plugins/by-name/codecompanion/default.nix
@@ -22,16 +22,25 @@ lib.nixvim.plugins.mkNeovimPlugin {
         anthropic = "anthropic";
         azure_openai = "azure_openai";
         copilot = "copilot";
+        deepseek = "deepseek";
         gemini = "gemini";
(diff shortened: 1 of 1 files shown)


MattSturgeon (approved):

MattSturgeon:
Could you reword the commit message to use our usual scoped format:

```
plugins/codecompanion: update defaults and examples
```

> I found some missing adapters

Note that this PR is only changing the documented "plugin default", not adding any actual settings options.

However, that's fine because `settings` should be a freeform option. Unless we declare a restrictive sub-option, anything not matching a declared option can still be defined.

MattSturgeon:
> Could you reword the commit message to use our usual scoped format:

I've pushed this.

PR #3542, opened 2025-07-08
ci: add tag-maintainers workflow

khaneliman:
Used to parse files changed and determine who should be notified of the changes.

Tested briefly, just throwing it out here to get some reviewing started.

### DIFF
--- a/.github/workflows/tag-maintainers.yml
+++ b/.github/workflows/tag-maintainers.yml
@@ -67,52 +67,9 @@ jobs:
           PR_AUTHOR: "${{ github.event.pull_request.user.login }}"
           CHANGED_FILES: '${{ steps.changed-files.outputs.changed_files }}'
         run: |
-          if [[ -z "$CHANGED_FILES" ]]; then
-            echo "No plugin files changed. No maintainers to tag."
-            echo "maintainers=" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          echo "Evaluating nixvim meta system for all maintainers..."
-          # Get a JSON object mapping plugin paths to maintainer data.
-          ALL_MAINTAINERS=$(nix eval --impure --expr "
-            let
-              nixvim = import ./.;
-              lib = nixvim.inputs.nixpkgs.lib.extend nixvim.lib.overlay;
-              emptyConfig = lib.nixvim.evalNixvim {
-                modules = [ { _module.check = false; } ];
-                extraSpecialArgs = { pkgs = null; };
-              };
-              inherit (emptyConfig.config.meta) maintainers;
-            in
-              maintainers
-          " --json 2>/dev/null || echo "{}")
-
-          echo "Finding maintainers for changed files..."
-          FOUND_MAINTAINERS=$(
-            echo "$CHANGED_FILES" | while IFS= read -r FILE; do
-              [[ -z "$FILE" ]] && continue
-              PLUGIN_DIR=$(dirname "$FILE")
-
-              # Use jq to find maintainers whose path key ends with the plugin directory.
-              echo "$ALL_MAINTAINERS" | jq -r --arg plugindir "$PLUGIN_DIR" '
-                to_entries[] |
-                select(.key | endswith($plugindir)) |
-                .value[] |
-                select(has("github")) |
-                .github
-              ' 2>/dev/null
-            done
-          )
-
-          # De-duplicate the list, remove the PR author, and format as a space-separated string.
-          MAINTAINERS_LIST=$(echo "$FOUND_MAINTAINERS" | grep -v -w -F "$PR_AUTHOR" | sort -u | tr '\n' ' ' | sed 's/ *$//')
-
-          if [[ -z "$MAINTAINERS_LIST" ]]; then
-            echo "No maintainers found for changed files (or only the PR author is a maintainer)."
-          else
-            echo "Found maintainers to notify: $MAINTAINERS_LIST"
-          fi
+          MAINTAINERS_LIST=$(./ci/tag-maintainers/extract-maintainers.py \
+            --changed-files "$CHANGED_FILES" \
+            --pr-author "$PR_AUTHOR")
 
           echo "maintainers=$MAINTAINERS_LIST" >> "$GITHUB_OUTPUT"
 
--- a/ci/tag-maintainers/extract-maintainers.py
+++ b/ci/tag-maintainers/extract-maintainers.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+"""
+Extract maintainers from changed plugin files.
+
+This script extracts the maintainer extraction logic from the tag-maintainers workflow
+for easier testing and validation.
+"""
+
+import argparse
+import json
+import subprocess
+import sys
+from typing import List
+
+
+def run_nix_eval(expr: str) -> str:
+    """Run a Nix evaluation expression and return the result."""
+    try:
+        result = subprocess.run(
+            ["nix", "eval", "--impure", "--json", "--expr", expr],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+        return result.stdout.strip()
+    except subprocess.CalledProcessError as e:
+        print(f"Error running Nix evaluation: {e}", file=sys.stderr)
+        print(f"Stderr: {e.stderr}", file=sys.stderr)
+        return "[]"
+
+
+def extract_maintainers(changed_files: List[str], pr_author: str) -> List[str]:
+    """Extract maintainers from changed plugin files."""
+    if not changed_files:
+        print("No plugin files changed. No maintainers to tag.", file=sys.stderr)
+        return []
+
+    print("Finding maintainers for changed files...", file=sys.stderr)
+
+    changed_files_nix = '[ ' + ' '.join(f'"{f}"' for f in changed_files) + ' ]'
+
+    nix_expr = f"""
+    let
+      nixvim = import ./.;
+      lib = nixvim.inputs.nixpkgs.lib;
+      emptyConfig = nixvim.lib.nixvim.evalNixvim {{
+        modules = [ {{ _module.check = false; }} ];
+        extraSpecialArgs.pkgs = null;
+      }};
+      inherit (emptyConfig.config.meta) maintainers;
+
+      changedFiles = {changed_files_nix};
+
+      # Find maintainers for files that match changed plugin directories
+      relevantMaintainers = lib.concatLists (
+        lib.mapAttrsToList (path: maintainerList:
+          let
+            matchingFiles = lib.filter (file:
+              lib.hasSuffix (dirOf file) path
+            ) changedFiles;
+          in
+            if matchingFiles != [] then maintainerList else []
+        ) maintainers
+      );
+
+      # Extract GitHub usernames
+      githubUsers = lib.concatMap (maintainer:
+        if maintainer ? github then [ maintainer.github ] else []
+      ) relevantMaintainers;
+
+    in
+      lib.unique githubUsers
+    """
+
+    result = run_nix_eval(nix_expr)
+
+    try:
+        maintainers = json.loads(result)
+    except json.JSONDecodeError:
+        print(f"Error parsing Nix evaluation result: {result}", file=sys.stderr)
+        return []
+
+    filtered_maintainers = [m for m in maintainers if m != pr_author]
+
+    if not filtered_maintainers:
+        print("No maintainers found for changed files (or only the PR author is a maintainer).", file=sys.stderr)
+        return []
+    else:
+        print(f"Found maintainers to notify: {' '.join(filtered_maintainers)}", file=sys.stderr)
+        return filtered_maintainers
+
+
+def main() -> None:
+    """Main function to handle command line arguments and run the extraction."""
+    parser = argparse.ArgumentParser(
+        description="Extract maintainers from changed plugin files"
+    )
+    parser.add_argument(
+        "--changed-files",
+        help="Newline-separated list of changed files",
+        default="",
+    )
+    parser.add_argument(
+        "--pr-author",
+        required=True,
+        help="GitHub username of the PR author",
+    )
+
+    args = parser.parse_args()
+    changed_files = [f.strip() for f in args.changed_files.split('\n') if f.strip()]
+    maintainers = extract_maintainers(changed_files, args.pr_author)
+    print(' '.join(maintainers))
+
+
+if __name__ == "__main__":
+    main()


.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
+name: Sync Plugin Maintainer Reviewers
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, reopened, synchronize]
+
+# Concurrency settings to ensure that only one instance of this workflow runs per PR.
+# If a new commit is pushed, it cancels the previous run.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.run_id }}
```
MattSturgeon: I don't think it's possible for the the event name to be `pull_request`; it should always be `pull_request_target`, because it is the only event this workflow runs on.

I believe it may be better to use `github.event.pull_request.number` though, because `head_ref` isn't guaranteed to be unique; multiple PRs could have the same branch names if they're from different forks.


How about:
```
group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
?

---

Sometimes it's a good idea to also have `on: pull_request` with `paths: .GitHub/workflow/this_workflow.yml`, so that we get a sanity check if a PR modifies the workflow. But then we need conditions throughout the workflow to handle things that would fail without the permissions we get from `pull_request_target`.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+  cancel-in-progress: true
```
MattSturgeon, line 11: I wonder what the pros and cons are of queuing vs cancelling?

If we let the older runs finish, and the later runs are relatively idempotent, then that may end up being simpler and more efficient.

The alternative is that we cancel the older runs, which may or may not be nearly finished, wasting any work that it _had_ done.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+
+permissions:
+  contents: read       # To checkout code
+  pull-requests: write # To add/remove reviewers and comment
+
+jobs:
+  tag-maintainers:
+    runs-on: ubuntu-latest
+    if: |
+      github.repository_owner == 'nix-community' &&
```
MattSturgeon: If we have this check in the condition, there is no reason to have `permissions` above.

`permissions: {}` will explicitly set no permissions for `github.token`.

If the intention of the permissions block was to allow testing on forks that don't have an App installed, then we need to drop the `owner` condition.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+
+      # Step 2: Checkout the code from the base branch.
+      # This is a security measure for `pull_request_target` to run trusted code.
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
```
MattSturgeon: Probably better to get the latest code from the base branch, the PRs base commit is pretty arbitrary from my observations.

We can do this using `github.base_ref`.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+
+      # Step 3: Get the specific Nixpkgs revision from the lock file.
+      - name: Get Nixpkgs revision from flake.lock
+        id: get-nixpkgs
+        run: |
+          echo "rev=$(jq -r '.nodes.nixpkgs.locked.rev' flake.lock)" >> "$GITHUB_OUTPUT"
```
MattSturgeon: Can we match the implementation from the pr-merge workflow?

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+
+      # Step 5: Identify which plugin files have changed in the PR.
+      - name: Get changed plugin files
+        id: changed-files
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+        run: |
+          CHANGED_FILES=$(gh pr diff ${{ github.event.pull_request.number }} --name-only || true)
```
MattSturgeon: While the PR number should be safe, I'd highly recommend always passing template values into scripts via `env`.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          echo "Changed files:"
+          echo "$CHANGED_FILES"
+          echo "changed_files<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGED_FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
MattSturgeon:
```suggestion
          {
            echo "changed_files<<EOF"
            echo "$CHANGED_FILES"
            echo EOF
          } >> "$GITHUB_OUTPUT"
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          # De-duplicate the list, remove the PR author, and format as a space-separated string.
+          MAINTAINERS_LIST=$(echo "$FOUND_MAINTAINERS" | grep -v -w -F "$PR_AUTHOR" | sort -u | tr '\n' ' ' | sed 's/ *$//')
+
+          if [[ -z "$MAINTAINERS_LIST" ]]; then
+            echo "No maintainers found for changed files (or only the PR author is a maintainer)."
+          else
+            echo "Found maintainers to notify: $MAINTAINERS_LIST"
+          fi
+
+          echo "maintainers=$MAINTAINERS_LIST" >> "$GITHUB_OUTPUT"
```
MattSturgeon: Since we've done a checkout of trusted code, we could extract this into a `ci/` script so that it can be maintained/tested/developed separately?

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+              # Use jq to find maintainers whose path key ends with the plugin directory.
+              echo "$ALL_MAINTAINERS" | jq -r --arg plugindir "$PLUGIN_DIR" '
+                to_entries[] |
+                select(.key | endswith($plugindir)) |
+                .value[] |
+                select(has("github")) |
+                .github
+              ' 2>/dev/null
+            done
+          )
```
MattSturgeon: We evaluated all maintainers, then query the full set with jq.

I think it would be simpler to directly evaluate the intersection of files and maintainers directly in nix?

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          echo "maintainers=$MAINTAINERS_LIST" >> "$GITHUB_OUTPUT"
+
+      # Step 7: Get lists of existing reviewers to avoid duplicates.
+      - name: Get current reviewers
+        id: current-reviewers
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+        run: |
+          PENDING_REVIEWERS=$(gh pr view ${{ github.event.pull_request.number }} --json reviewRequests --jq '.reviewRequests[].login')
+          PAST_REVIEWERS=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews" --jq '.[].user.login')
```
MattSturgeon: Again, let's use `env` instead of injection templates directly into bash

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          USERS_TO_EXCLUDE=$(printf "%s\n%s" "$PENDING_REVIEWERS" "$PAST_REVIEWERS" | sort -u)
+
+          echo "pending_reviewers<<EOF" >> $GITHUB_OUTPUT
+          echo "$PENDING_REVIEWERS" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+          echo "users_to_exclude<<EOF" >> $GITHUB_OUTPUT
+          echo "$USERS_TO_EXCLUDE" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
MattSturgeon:
```suggestion
          {
            echo "pending_reviewers<<EOF"
            echo "$PENDING_REVIEWERS"
            echo EOF

            echo "users_to_exclude<<EOF"
            echo "$USERS_TO_EXCLUDE"
            echo EOF
          } >> $GITHUB_OUTPUT
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+
+          echo "Current pending reviewers: $PENDING_REVIEWERS"
+          echo "Complete list of users to exclude: $USERS_TO_EXCLUDE"
+
+      # Step 8: Filter the maintainer list to only include repository collaborators.
+      # You can only request reviews from users with write access/collaborators.
```
MattSturgeon:
```suggestion
      # You can only request reviews from users with at least triage permissions.
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+          MAINTAINERS: ${{ steps.extract-maintainers.outputs.maintainers }}
+          PENDING_REVIEWERS: ${{ steps.current-reviewers.outputs.pending_reviewers }}
+        run: |
+          remove_reviewers() {
+            local reviewers_to_remove="$1"
+            local reason="$2"
+            for REVIEWER in $reviewers_to_remove; do
+              echo "Removing review request from $REVIEWER ($reason)"
+              gh api --method DELETE "/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/requested_reviewers" \
```
MattSturgeon: Again, templates vs vars

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+                --input - <<< "{\"reviewers\": [\"$REVIEWER\"]}"
```
MattSturgeon: My gut doesn't like the idea of constructing JSON with a bash string... I'm guessing GitHub usernames can't contain things like `"` so it's probably fine. But I'd be more comfortable using `jq` to marshall the JSON.

```
jq -c '{ reviewers: [.] }'
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,235 @@
...
+      # Step 1: Generate a GitHub App token if configured.
+      - name: Create GitHub App token
+        uses: actions/create-github-app-token@v2
+        if: vars.CI_APP_ID
+        id: app-token
+        with:
+          app-id: ${{ vars.CI_APP_ID }}
+          private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
```
MattSturgeon, line 33: Do we also need a permission to view org team members?

.github/workflows/tag-maintainers.yml
khaneliman: We only want the current state of the PR diff to be used for decisions. Someone could have a version they push that pings 100 people because of not being based correctly. They could rebase and push before the tags happen and we prevent needlessly pinging people. We dont really need a queue of requesting, removing, requesting, reviewing from all the pushes someone makes.

.github/workflows/tag-maintainers.yml
khaneliman: Yes, good catch.

.github/workflows/tag-maintainers.yml
khaneliman: Yeah, that is half baked.. I don't really feel like this benefits from fork support the same as some other flows. But, I will remove this condition since I already implemented the github token fallback stuff.

.github/workflows/tag-maintainers.yml
khaneliman: Sorry, missed during moving stuff around.

.github/workflows/tag-maintainers.yml
khaneliman: Yeah, good idea.

.github/workflows/tag-maintainers.yml
khaneliman: That's not true.

.github/workflows/tag-maintainers.yml
khaneliman: Yes, do you remember what its called again?

MattSturgeon: I had to look it up again... `permission-members: read`

https://github.com/actions/create-github-app-token/blob/93c1f04d6f14ea5b416e7a8dfd80446101c6adef/action.yml#L58-L59

.github/workflows/tag-maintainers.yml
khaneliman: >group: ${{ github.workflow }}-${{ github.event.pull_request.number }}

Updated the group identifier. 

>Sometimes it's a good idea to also have on: pull_request with paths: .GitHub/workflow/this_workflow.yml, so that we get a sanity check if a PR modifies the workflow. But then we need conditions throughout the workflow to handle things that would fail without the permissions we get from pull_request_target.

But, wasn't sure what you meant on this part.

.github/workflows/tag-maintainers.yml
khaneliman: Sweet, thanks

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+
+      # Step 5: Identify which plugin files have changed in the PR.
+      - name: Get changed plugin files
+        id: changed-files
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+          PR_NUM: ${{ github.event.pull_request.number }}
+        run: |
```
MattSturgeon, line 53: Just for my own understanding: `gh pr diff` will show the pr diff against its base (as it would appear on the website), **not** the diff against the local worktree (`git diff HEAD..FETCH_HEAD`)?

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+
+      # Step 6: Evaluate Nix code to find maintainers for the changed files.
+      - name: Extract maintainers from changed files
+        id: extract-maintainers
+        env:
+          PR_AUTHOR: "${{ github.event.pull_request.user.login }}"
+          CHANGED_FILES: '${{ steps.changed-files.outputs.changed_files }}'
+        run: |
+          if [[ -z "$CHANGED_FILES" ]]; then
+            echo "No plugin files changed. No maintainers to tag."
```
MattSturgeon: IIUC, there's no need to explicitly set an empty value for GitHub outputs.

Reading an undefined value will usually evaluate to an empty string in GHA.

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+            echo "maintainers=" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          echo "Evaluating nixvim meta system for all maintainers..."
+          # Get a JSON object mapping plugin paths to maintainer data.
+          ALL_MAINTAINERS=$(nix eval --impure --expr "
+            let
+              nixvim = import ./.;
+              lib = nixvim.inputs.nixpkgs.lib.extend nixvim.lib.overlay;
```
MattSturgeon: IIRC, we expose a `lib` flake output that already contains evalNixvim?


```suggestion
              emptyConfig = nixvim.lib.nixvim.evalNixvim {
```

I might be misremembering the exact attrpath

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+              emptyConfig = lib.nixvim.evalNixvim {
+                modules = [ { _module.check = false; } ];
+                extraSpecialArgs = { pkgs = null; };
+              };
+              inherit (emptyConfig.config.meta) maintainers;
+            in
```
MattSturgeon: Is this better or worse?


```suggestion
            in
            emptyConfig.config.meta.maintainers
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+          fi
+
+          echo "Evaluating nixvim meta system for all maintainers..."
+          # Get a JSON object mapping plugin paths to maintainer data.
+          ALL_MAINTAINERS=$(nix eval --impure --expr "
+            let
+              nixvim = import ./.;
+              lib = nixvim.inputs.nixpkgs.lib.extend nixvim.lib.overlay;
+              emptyConfig = lib.nixvim.evalNixvim {
+                modules = [ { _module.check = false; } ];
```
MattSturgeon:
```suggestion
                extraSpecialArgs.pkgs = null;
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+          CHANGED_FILES: '${{ steps.changed-files.outputs.changed_files }}'
+        run: |
+          if [[ -z "$CHANGED_FILES" ]]; then
+            echo "No plugin files changed. No maintainers to tag."
+            echo "maintainers=" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          echo "Evaluating nixvim meta system for all maintainers..."
+          # Get a JSON object mapping plugin paths to maintainer data.
```
MattSturgeon: I suspect this is possible with a pure eval.

Something like:

```sh
nix eval .#lib.evalNixvim --apply '
  evalNixvim:
  let
    # ...
  in
  maintainers
'
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+            .nodes[.nodes[.root].inputs.nixpkgs].locked
+            | "https://github.com/\(.owner)/\(.repo)/archive/\(.rev).tar.gz"
+            | "url=\(.)"
+          ' flake.lock >> "$GITHUB_OUTPUT"
+
+      # Step 4: Install Nix with the correct configuration.
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/${{ steps.get-nixpkgs.outputs.rev }}.tar.gz
```
MattSturgeon:
```suggestion
          nix_path: nixpkgs=${{ steps.get-nixpkgs.outputs.url }}
```

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,248 @@
...
+          extra_nix_config: |
+            experimental-features = nix-command flakes
```
MattSturgeon: I believe the cachix install nix action enables flakes by default; or at least we use them in other workfloww without explicitly enabling them, iirc

.github/workflows/tag-maintainers.yml
MattSturgeon: Thinking about it, with the current implementation do we actually need NIX_PATH at all? If we're using `nix eval` and importing the whole flake, then we should never lookup `<nixpkgs>`.

Maybe I'm missing something, or you have plans to use lookup paths in a refactor?

.github/workflows/tag-maintainers.yml
MattSturgeon: Maybe not if we refactor though... E.g. if we need to pass in the `files` array, we need to eval an expression to be able to use `--argstr`, `--arg-json`, etc. You can't mix flake --apply with --arg.

MattSturgeon (commented):
Note: suggestions such as refactoring jq stuff into nix expressions or extracting logic into separate files are **not** blocking. If it's proven working, it can be merged with the current design and iterated on later.

Having seen it working on your fork and in home-manager gives me additional confidence.

.github/workflows/tag-maintainers.yml
khaneliman: Oh, yeah it's probably default now. I've just had these in my actions since I got started with nix.

.github/workflows/tag-maintainers.yml
khaneliman: This causes eval errors, thats why we did the module.check in the last pr

.github/workflows/tag-maintainers.yml
khaneliman: i'm gonna test the changes in my fork again and see if its even needed anymore, it doesn't look like it. i think we had it with the others for `nix-instantiate` ?

.github/workflows/tag-maintainers.yml
khaneliman:
```bash
./ci/tag-maintainers/extract-maintainers.py --pr-author khaneliman --changed-files plugins/by-name/dashboard/default.nix
Finding maintainers for changed files...
Error running Nix evaluation: Command '['nix', 'eval', '--json', '--expr', '\n    let\n      nixvim = import ./.;\n      lib = nixvim.inputs.nixpkgs.lib;\n      emptyConfig = nixvim.lib.nixvim.evalNixvim {\n        modules = [ { _module.check = false; } ];\n        extraSpecialArgs.pkgs = null;\n      };\n      inherit (emptyConfig.config.meta) maintainers;\n\n      changedFiles = [ "plugins/by-name/dashboard/default.nix" ];\n\n      # Find maintainers for files that match changed plugin directories\n      relevantMaintainers = lib.concatLists (\n        lib.mapAttrsToList (path: maintainerList:\n          let\n            matchingFiles = lib.filter (file:\n              lib.hasSuffix (dirOf file) path\n            ) changedFiles;\n          in\n            if matchingFiles != [] then maintainerList else []\n        ) maintainers\n      );\n\n      # Extract GitHub usernames\n      githubUsers = lib.concatMap (maintainer:\n        if maintainer ? github then [ maintainer.github ] else []\n      ) relevantMaintainers;\n\n    in\n      lib.unique githubUsers\n    ']' returned non-zero exit status 1.
Stderr: error:
       … while calling the 'import' builtin
         at «string»:3:16:
            2|     let
            3|       nixvim = import ./.;
             |                ^
            4|       lib = nixvim.inputs.nixpkgs.lib;

       error: access to absolute path '/home' is forbidden in pure evaluation mode (use '--impure' to override)

No maintainers found for changed files (or only the PR author is a maintainer).

File: .github/workflows/tag-maintainers.yml (hunk 16 above)
Comment (khaneliman) [reply .github/workflows/tag-maintainers.yml:53]: I'm not 100% certain of the differences, I was just using `gh` cli for the other commands and continued the trend. 

EDIT: https://github.com/cli/cli/pull/962 and the issue do make it sound like that's the purpose of letting them handle what to diff against, in case of differing bases.

File: .github/workflows/tag-maintainers.yml (hunk 22 above)
Comment (khaneliman) [reply .github/workflows/tag-maintainers.yml]: https://github.com/khaneliman/nixvim/pull/8 looks it works without it fine

File: .github/workflows/tag-maintainers.yml (hunk 21 above)
Comment (MattSturgeon) [reply .github/workflows/tag-maintainers.yml]: Yes, my suggestion was to use `--apply` instead of `--expr`.

This allows applying a function expression to a flake output without needing to import it impurely via flake-compat.

It is only useful if you don't need to supply multiple arguments; i.e. the flake output is the only thing applied to the expression.

File: .github/workflows/tag-maintainers.yml (hunk 20 above)
Comment (MattSturgeon) [reply .github/workflows/tag-maintainers.yml]: Sorry the comment ended up on the wrong line. I think you pushed while I was still reviewing.

The suggestion was to replace:

```diff
- extraSpecialArgs = {
-   pkgs = null;
- };
+ extraSpecialArgs.pkgs = null;
```

ci/tag-maintainers/extract-maintainers.py
```diff
@@ -0,0 +1,116 @@
...
+
+def extract_maintainers(changed_files: List[str], pr_author: str) -> List[str]:
+    """Extract maintainers from changed plugin files."""
+    if not changed_files:
+        print("No plugin files changed. No maintainers to tag.", file=sys.stderr)
+        return []
+
+    print("Finding maintainers for changed files...", file=sys.stderr)
+
+    changed_files_nix = '[ ' + ' '.join(f'"{f}"' for f in changed_files) + ' ]'
```
MattSturgeon, line 40: `nix eval` can insert raw, string, and file arguments. And nix can deserialise JSON using `builtins.fromJSON`. Therefore, we should never need to manually marshal to nix.


```
· --arg name expr

  Pass the value expr as the argument name to Nix functions.

· --arg-from-file name path

  Pass the contents of file path as the argument name to Nix functions.

· --arg-from-stdin name

  Pass the contents of stdin as the argument name to Nix functions.

· --argstr name string

  Pass the string string as the argument name to Nix functions.
```

ci/tag-maintainers/extract-maintainers.py
```diff
@@ -0,0 +1,116 @@
...
+
+    nix_expr = f"""
```
MattSturgeon, line 42: I'd be tempted to put this in a loose nix file so that we benefit from formatting, lsp, etc, and so we can manually evaluate the file without needing to use the full script.

ci/tag-maintainers/extract-maintainers.py
```diff
@@ -0,0 +1,116 @@
...
+    let
+      nixvim = import ./.;
```
MattSturgeon, line 44: I'd consider making the "path to Nixvim" overridable via CLI args.

That way external code could specify explicitly where the root is, a default value could use `git rev-parse --show-toplevel`, a wrapped script could use the in-store `self` path, etc.

This would look something like:

```nix
{
  nixvimPath ? ./.,
  nixvim ? import nixvimPath,
  changedFilesJson ? throw "either supply changedFiles or changedFilesJson",
  changedFiles ? builtins.fromJSON changedFilesJson,
}:
```

.github/workflows/tag-maintainers.yml
MattSturgeon, line 1: My instinct is that additional logic could be extracted into separate files.

MattSturgeon (approved):
LGTM. Are we ready to try this out for real?

.github/workflows/tag-maintainers.yml
MattSturgeon: > But, wasn't sure what you meant on this part.

In nixpkgs, most workflows have 

```
on:
  pull_request:
    paths: .github/workflows/foo.yml
```

So that PRs modifying the workflow will run the workflow. 

But any part of the workflow that uses `secrets`, or write permissions on `github.token`, must then check that `event_name != pull_request` before doing something that can't be done on a pull_request run.

MattSturgeon:
One thing we'll need to monitor is whether _removing_ review requests causes friction; AFAIK there's no way to differentiate requests from CI vs manual requests

khaneliman:
> One thing we'll need to monitor is whether _removing_ review requests causes friction; AFAIK there's no way to differentiate requests from CI vs manual requests

Hmm good point. I didn't think about that flow of a manual request and then changes pushed. Although, we won't remove a user that has an active review, so it would be the time between a manual review request and them providing some feedback that gets them reviewed from intermediary pushes.

MattSturgeon:
It's already showing up...

image

PR #3547, opened 2025-07-11
user-configs: add @c4patino's config"

c4patino:


GaetanLepage (approved):
Thanks for sharing!

c4patino:
Thanks! I'm trying to get more involved with the contributing back both here and to nixpkgs, however I'm still trying to find where help is needed. I'm finding it hard from the outside to figure out exactly where to start.

Issue #3546, opened 2025-07-10
Remove review requests that no longer match a PR


MattSturgeon:
When initially implemented in #3542, the tag-maintainers workflow removed review requests that no longer match the maintainers list.

This was removed in #3544 because it often false-positives manual review requests.

In https://github.com/nix-community/nixvim/pull/3544#issuecomment-3057500549, @khaneliman suggested querying the PR's timeline events to see who requested reviews. If the reviewer was only requested by @nixvim-ci[bot] then we can safely remove the review request.

PR #3543, opened 2025-07-09
ci/tag-maintainers: split nix into separate file + minor tweaks

MattSturgeon:
- **ci: run treefmt**
- **ci/tag-maintainers: split nix into separate file**
- **ci/tag-maintainers: pass changed file to nix as json**
- **ci/tag-maintainers: minor cleanup**

Two issues I've spotted but not resolved:

The `hasSuffix` check assumes that the "changed file" has a `/default.nix` component on the end that is not present in the `maintainers` file key; which is usually true but isn't guaranteed. Maybe it'd be best to always import `default.nix` files explicitly? That way docs links would also correctly link to the actual file...

If the number of changed files gets too large, the _changed files_ JSON might exceed the shell arg length limit. It might be worth writing the changed files list to a file and passing that in instead.

PR #3548, opened 2025-07-11
ci/tag-maintainers: refactor managing reviewers

khaneliman:
Closes https://github.com/nix-community/nixvim/issues/3546

Move to separate script that looks at history of requests to determine who needs to be removed. We will not remove reviews from those who were manually requested.

Won't work until `extract-maintainers` is fixed, though. Tested in home-manager https://github.com/khaneliman/home-manager/pull/591 Was testing in my nixvim fork but realized the extraction of maintainers was broken in https://github.com/nix-community/nixvim/pull/3543

### DIFF
--- a/.github/workflows/tag-maintainers.yml
+++ b/.github/workflows/tag-maintainers.yml
@@ -73,79 +73,23 @@ jobs:
 
           echo "maintainers=$MAINTAINERS_LIST" >> "$GITHUB_OUTPUT"
 
-      # Get lists of existing reviewers to avoid duplicates.
-      - name: Get current reviewers
-        id: current-reviewers
+      # Manage reviewers
+      - name: Manage reviewers
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
-          PR_NUM: ${{ github.event.pull_request.number }}
-          REPO: ${{ github.repository }}
-        run: |
-          PENDING_REVIEWERS=$(gh pr view "$PR_NUM" --json reviewRequests --jq '.reviewRequests[].login')
-          PAST_REVIEWERS=$(gh api "repos/$REPO/pulls/$PR_NUM/reviews" --jq '.[].user.login')
-          USERS_TO_EXCLUDE=$(printf "%s\n%s" "$PENDING_REVIEWERS" "$PAST_REVIEWERS" | sort -u)
-
-          {
-            echo "pending_reviewers<<EOF"
-            echo "$PENDING_REVIEWERS"
-            echo EOF
-            echo "users_to_exclude<<EOF"
-            echo "$USERS_TO_EXCLUDE"
-            echo EOF
-          } >> $GITHUB_OUTPUT
-
-          echo "Current pending reviewers: $PENDING_REVIEWERS"
-          echo "Complete list of users to exclude: $USERS_TO_EXCLUDE"
-
-      # Filter the maintainer list to only include repository collaborators.
-      # You can only request reviews from users with at least triage permissions.
-      - name: Check maintainer collaborator status
-        id: check-collaborators
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
           MAINTAINERS: ${{ steps.extract-maintainers.outputs.maintainers }}
-          USERS_TO_EXCLUDE: ${{ steps.current-reviewers.outputs.users_to_exclude }}
-          REPO: "${{ github.repository }}"
+          CHANGED_FILES: ${{ steps.changed-files.outputs.changed_files }}
+          BOT_NAME: ${{ steps.app-token.outputs.app-slug || 'github-actions' }}
         run: |
-          NEW_REVIEWERS=()
-
-          # If there are no maintainers, exit early.
-          if [[ -z "$MAINTAINERS" ]]; then
-            echo "No maintainers to check."
-            echo "new_reviewers=" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          for MAINTAINER in $MAINTAINERS; do
-            if echo "$USERS_TO_EXCLUDE" | grep -q -w "$MAINTAINER"; then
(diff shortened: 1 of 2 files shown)


ci/tag-maintainers/manage-reviewers.py
```diff
@@ -0,0 +1,234 @@
...
+    level=logging.INFO,
+    format="%(message)s",
+    stream=sys.stderr,
+)
+
+MANUAL_REVIEW_REQUEST_QUERY: Final[str] = """
+query($owner: String!, $repo: String!, $prNumber: Int!) {
+  repository(owner: $owner, name: $repo) {
+    pullRequest(number: $prNumber) {
+      timelineItems(first: 100, itemTypes: [REVIEW_REQUESTED_EVENT]) {
```
MattSturgeon: I wonder if it's more relevant to have the first or last 100 timeline items?

ci/tag-maintainers/manage-reviewers.py
```diff
@@ -0,0 +1,234 @@
...
+            "-F", f"repo={repo}",
+            "-F", f"prNumber={pr_number}",
+        ])
+        data = json.loads(result)
+        nodes = data.get("data", {}).get("repository", {}).get("pullRequest", {}).get("timelineItems", {}).get("nodes", [])
+
+        manually_requested = {
+            node["requestedReviewer"]["login"]
+            for node in nodes
+            if node and node.get("actor", {}).get("__typename") == "User" and node.get("requestedReviewer")
```
MattSturgeon: What are we asserting here? The `__typename`? Is that always `"User"` for normal users and something else for bots/apps?

I believe we should be more specific and check that the user isn't `nixvim-ci[bot]` (or better yet use the bot's user id).

It is conceivable that some other bot may request reviews for a PR, and we shouldn't then try to "manage" those requests with this bot.

It should be possible to pass in the "current user id" as a CLI arg, since that is what we actually want to look for, right?

ci/tag-maintainers/manage-reviewers.py
```diff
@@ -0,0 +1,234 @@
...
+
+def is_collaborator(owner: str, repo: str, username: str) -> bool:
+    """Checks if a user is a collaborator on the repository."""
+    try:
+        # The `--silent` flag makes `gh` exit with 0 on success (2xx) and 1 on error (e.g., 404)
+        run_gh_command(["api", f"repos/{owner}/{repo}/collaborators/{username}", "--silent"])
+        return True
+    except GHError:
+        # A non-zero exit code (GHError) implies the user is not a collaborator (404) or another issue.
+        return False
```
MattSturgeon: I'd prefer if we handled `404` and "another issue" separately. Other issues should probably raise/propagate an error.

MattSturgeon (commented):
Looks good conceptually. I haven't fully understood the impl yet, but I trust you to have implemented it well 😁

ci/tag-maintainers/manage-reviewers.py
khaneliman: I started checking by name and had to do dumb string checking hacks between info returned from the app info step (we suffix with `[bot]`) and what's reported in the graphql query. If you want to be that specific, we can go back to that method. But, I found the type name much simpler and less error prone. The only concern we should have is whether someone chose to request a review manually and any bot that can request reviews would be controlled by us, I believe.  

```json
{
  "data": {
    "repository": {
      "pullRequest": {
        "timelineItems": {
          "nodes": [
            {
              "actor": {
                "__typename": "Bot",
                "login": "khanelibot"
              },
              "requestedReviewer": {
                "login": "GaetanLepage"
              }
            },
            {
              "actor": {
                "__typename": "User",
                "login": "khaneliman"
              },
              "requestedReviewer": {
                "login": "MattSturgeon"
              }
            },
            {
              "actor": {
                "__typename": "Bot",
                "login": "khanelibot"
              },
              "requestedReviewer": {
                "login": "GaetanLepage"
              }
            }
          ]
        }
      }
    }
  }
}
```

ci/tag-maintainers/manage-reviewers.py
khaneliman: Yeah, I am also curious if there is any harm to setting this to a higher number where it doesn't matter. I find it hard to imagine us ever hitting the 100 review request events, though.

ci/tag-maintainers/manage-reviewers.py
khaneliman: Yeah, that's true.. I was just adding this to silence the error logging of expected errors like that and went too simple.

ci/tag-maintainers/manage-reviewers.py
khaneliman: Bumped to 250 (max allowed by their graphql endpoint) and switched to last. But, it would be nice to eventually refactor this to be smarter about looking at the current state of the reviewers and query for their specific review history maybe? To see if the last review event for that user was from a user or bot.

ci/tag-maintainers/manage-reviewers.py
khaneliman: While typing this out, I think it made me realize that I was on the right path before but ignored we were the ones returning `[bot]` and i just had to strip that from the property i was passing into the script...

khaneliman:
https://github.com/khaneliman/nixvim/pull/11 tested matching just our bot user.

ci/tag-maintainers/manage-reviewers.py
MattSturgeon: > But, it would be nice to eventually refactor this to be smarter about looking at the current state of the reviewers and query for their specific review history maybe? 

While that may be more correct, I think the current implementation better optimises for making less API requests. I think that's more important, given how unlikely it is we'll have over 250 review request events.

.github/workflows/tag-maintainers.yml
```diff
@@ -73,79 +73,22 @@ jobs:
...
-            echo "No new reviewers to add."
-          fi
+          ./ci/tag-maintainers/manage-reviewers.py \
+            --owner "$OWNER" \
+            --repo "$REPO" \
+            --pr-number "$PR_NUMBER" \
+            --pr-author "$PR_AUTHOR" \
+            --current-maintainers "$MAINTAINERS" \
+            --changed-files "$CHANGED_FILES" \
+            --bot-user-name "${{ steps.app-token.outputs.app-slug }}"
```
MattSturgeon: _Please_ use an env var 😭

MattSturgeon: Also, when defining the var, we'll need `|| 'github-actions'` in case there is no app setup in the fork.

MattSturgeon (approved):
Minor nit, LGTM otherwise

PR #3551, opened 2025-07-12
ci/tag-maintainers: fix shallow history + flake-compat issues

MattSturgeon:
This should fix the issues raised by @khaneliman in #3546.

So far, only tested locally.


khaneliman (approved):
LGTM, thanks for fixing it. https://github.com/khaneliman/nixvim/pull/10 tested this with https://github.com/nix-community/nixvim/pull/3548 on top of it.

PR #3553, opened 2025-07-13
plugins/no-neck-pain: init

chelsea6502:
Requesting to add one of my favourite Neovim plugins: No Neck Pain. It centers the buffer to the middle of the screen.

This is my first time contributing to Nixvim, so let me know if there's any conventions I've missed. I assume I will have to add myself as a maintainer?

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -25,6 +25,12 @@
     githubId = 60232782;
     name = "Boney Patel";
   };
+  ChelseaWilkinson = {
+    email = "mail@chelseawilkinson.me";
+    github = "chelsea6502";
+    githubId = 122856043;
+    name = "Chelsea Wilkinson";
+  };
   DanielLaing = {
     email = "daniel@daniellaing.com";
     matrix = "@bodleum:matrix.org";
--- a/plugins/by-name/no-neck-pain/default.nix
+++ b/plugins/by-name/no-neck-pain/default.nix
@@ -1,0 +1,30 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "no-neck-pain";
+  maintainers = [ lib.maintainers.ChelseaWilkinson ];
+  packPathName = "no-neck-pain.nvim";
+  package = "no-neck-pain-nvim";
+  description = "☕ Dead simple yet super extensible zen mode plugin to protect your neck.";
+
+  settingsExample = {
+    width = 80;
+    autocmds = {
+      enableOnVimEnter = true;
+      skipEnteringNoNeckPainBuffer = true;
+    };
+    buffers = {
+      colors = {
+        background = "catppuccin-frappe";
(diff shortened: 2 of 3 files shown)


plugins/by-name/no-neck-pain/default.nix
```diff
@@ -0,0 +1,31 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "no-neck-pain";
+  maintainers = [ lib.maintainers.ChelseaWilkinson ];
+  packPathName = "no-neck-pain.nvim";
+  package = "no-neck-pain-nvim";
+  url = "https://github.com/shortcuts/no-neck-pain.nvim";
```
khaneliman:
```suggestion
```

Should be fetched from meta

plugins/by-name/no-neck-pain/default.nix
```diff
@@ -0,0 +1,31 @@
...
+  description = "☕ Dead simple yet super extensible zen mode plugin to protect your neck.";
+
+  moduleName = "no-neck-pain";
```
khaneliman:
```suggestion
```

Already defaults to `name`

tests/test-sources/plugins/by-name/no-neck-pain/default.nix
```diff
@@ -0,0 +1,11 @@
+{
+  empty = {
+    plugins.no-neck-pain.enable = true;
+  };
+
+  defaults = {
+    plugins.no-neck-pain = {
+      enable = true;
```
khaneliman, line 8: Should add at least some of the defaults from https://github.com/shortcuts/no-neck-pain.nvim/blob/main/lua/no-neck-pain/config.lua#L148-L338 for testing the plugin.

lib/maintainers.nix
```diff
@@ -25,6 +25,12 @@
     githubId = 60232782;
     name = "Boney Patel";
   };
+  ChelseaWilkinson = {
+    email = "mail@chelseawilkinson.me";
+    github = "chelsea6502";
+    githubId = 122856043;
+    name = "Chelsea Wilkinson";
+  };
```
khaneliman, line 33: Can you split this into a separate commit before the addition of the module.

chelsea6502: Resolved

tests/test-sources/plugins/by-name/no-neck-pain/default.nix
chelsea6502: Ah, now I understand nixvim tests! I've added some essential default settings for the test

chelsea6502:
@khaneliman Thanks for the help. I've squashed your suggested changes into the commit. 

I realised the example config was failing, so I've fixed those, and added a test for the example config.

shortcuts (commented):
Thanks for doing this ❤️

khaneliman (approved):

PR #3563, opened 2025-07-16
plugins/lspkind: migrate to mkNeovimPlugin

saygo-png:


plugins/by-name/lspkind/default.nix
```diff
@@ -51,30 +38,20 @@ in
...
-          else
-            { }
-        )
-        // cfg.extraOptions;
+        cfg.settings
+        // (lib.optionalAttrs doCmp {
+          maxwidth = cfg.cmp.maxWidth;
+          ellipsis_char = cfg.cmp.ellipsisChar;
+          inherit (cfg.cmp) menu;
+        });
```
MattSturgeon: This doesn't feel right for a freeform settings option.

I believe the idiomatic approach would be to have 

```nox
plugins.lspkind.settings = mkIf doCmp {
  # ...
};
```

As an actual config definition.

plugins/by-name/lspkind/default.nix
```diff
@@ -51,30 +38,20 @@ in
...
-            }
-          else
-            { }
-        )
-        // cfg.extraOptions;
+        cfg.settings
+        // (lib.optionalAttrs doCmp {
+          maxwidth = cfg.cmp.maxWidth;
+          ellipsis_char = cfg.cmp.ellipsisChar;
+          inherit (cfg.cmp) menu;
```
MattSturgeon: I'm not convinced these should even still exist as extra-options, since users can define them directly in settings themselves.

MattSturgeon (commented):
This is a difficult plugin to modernize, as it has several bespoke options and seems to be "set up" in different ways depending on the use case.

I think more of the hard-coded logic in nixvim can be dropped or deprecated, in favour of users directly interacting with freeform settings.

I'm fine with still having extra options that controls which method of setting up is used.

Alternatively, we could have different freeform settings options; one per setup method?

saygo-png:
Sorry, this diff got quite noisy and i forgot to update the test so there are two pushes.

- I removed the extra options you mentioned before, since they can just be defined in settings.
- Added an assertion to make sure the user only has the cmp integration enabled when cmp is also enabled, since thats the only case where enabling this option makes sense
- Thanks to the assertion doCmp is no longer needed and i just directly use cfg.cmp.enable
- I added another optional string so that no config gets added to cmp unless the integration is explicitly enabled

> Alternatively, we could have different freeform settings options; one person setup method?

I don't understand what you mean by this.

saygo-png:
The test is failing now because the "empty" attribute set doesn't enable the cmp plugin.
I don't really understand what the different attribute sets in the test files are meant to be, so i'm not sure if i should add the cmp plugin to the "empty" one or disable the lspkind cmp integration instead.

These test failures make me doubt the addition of this assertion. It will break existing configs, since cmp integration is enabled by default. Should i just remove the assert, or maybe switch cmp integration to false by default?

saygo-png:
Rebase

khaneliman:
> These test failures make me doubt the addition of this assertion. It will break existing configs, since cmp integration is enabled by default. Should i just remove the assert, or maybe switch cmp integration to false by default?

Personally, I like having the assertion to make sure that someone is aware that the configuration won't work, as expected. It is troublesome, obviously, that it's the default behavior to enable the integration.. because, yes, it will break anyone's config that might have the module enabled but isn't using `nvim-cmp`. However, I feel like this is one of those cases, that it makes sense to make users aware they have a configuration that doesn't make sense.

plugins/by-name/lspkind/default.nix
```diff
@@ -2,93 +2,64 @@
...
+
+                return (${cfg.cmp.after})(entry, vim_item, kind)
+              end
+            ''
+          else
+            ''
+              require('lspkind').cmp_format(${lib.nixvim.toLuaObject cfg.settings})
+            '';
+      in
+      lib.optionalString cfg.cmp.enable format';
```
MattSturgeon: Can we use `mkIf` instead of `optionalString`?

The subtle difference is that using `optionalString` means we _always_ define `plugins.cmp.settings.formatting.format`, sometimes as an empty string, other times as a non-empty.

IMO it's better to not define it at all when `cmp.enable` is false, as that avoids conflicting with other (user) definitions.

```suggestion
      lib.mkIf cfg.cmp.enable format';
```

I'm also not opposed to inlining the value:


```nix
lib.mkIf cfg.cmp.enable (
  if cfg.cmp.after != null then
    ''
      function(entry, vim_item)
        local kind = require('lspkind').cmp_format(${lib.nixvim.toLuaObject cfg.settings})(entry, vim_item)
        return (${cfg.cmp.after})(entry, vim_item, kind)
      end
    ''
  else
    ''
      require('lspkind').cmp_format(${lib.nixvim.toLuaObject cfg.settings})
    ''
)
```

saygo-png: I didn't realize the difference between `mkIf` and `optionalString`. Switched to `mkIf`, inline looks cleaner so I did that too.

plugins/by-name/lspkind/default.nix
```diff
@@ -2,93 +2,61 @@
...
-              return (${cfg.cmp.after})(entry, vim_item, kind)
-            end
-          ''
-        else
-          ''
-            require('lspkind').cmp_format(${lib.nixvim.toLuaObject options})
-          '';
     };
+
+    plugins.lspkind.luaConfig.content = lib.optionalString (!cfg.cmp.enable) ''
```
MattSturgeon: I should've spotted this last review, but ideally this'd also be `mkIf`

```suggestion
    plugins.lspkind.luaConfig.content = lib.mkIf (!cfg.cmp.enable) ''
```

Less important here, though.

MattSturgeon (approved):
Awesome!

plugins/by-name/lspkind/default.nix
saygo-png: It's good because i accidentally messed up the commits last force push (fixed up plugin change into the test commit, fixed now).

Issue #3469, opened 2025-06-13
[PLUGIN REQUEST] img-clip plugin.


niksingh710:
| Field    | Description |
|----------|-------------|
| Plugin   | `img-clip`  | 
| Homepage | `https://github.com/hakonharnes/img-clip.nvim`     | 
| Nixpkgs  | [vimPlugins.img-clip-nvim](https://search.nixos.org/packages?channel=25.05&show=vimPlugins.img-clip-nvim&from=0&size=50&sort=relevance&type=packages&query=img-clip) | 

## Extra Information

Pre Include packages based on system.

PR #3564, opened 2025-07-16
plugins/img-clip: init

saygo-png:
Closes https://github.com/nix-community/nixvim/issues/3469

I'm not sure if i wrote the test properly.

There is room for improvement here, as [upstream offers snippets for integrating](https://github.com/HakonHarnes/img-clip.nvim/?tab=readme-ov-file#-integrations) with telescope, oil.nvim and snacks.
We could provide these integrations as options. I'm planning to add this in the future.


MattSturgeon (approved):
Thanks 👍

khaneliman (approved):

Issue #2638, opened 2024-12-11
MkNeovimPlugin Tracker


khaneliman:
Just creating a tracker to keep track of what plugins need to be migrated to modern mkNeovimPlugin.

| Typ | Sty | DW | path
|-----|-----|----|--------------------------------------------------------
|    | ❌  |    | plugins/by-name/alpha/default.nix
|    | ❌  |    | plugins/by-name/autoclose/default.nix
|    | ❌  | ⚠  | plugins/by-name/clangd-extensions/default.nix
|    | ❌  |    | plugins/by-name/clipboard-image/default.nix
|    | ❌  |    | plugins/by-name/copilot-lua/default.nix
|    | ❌  |    | plugins/by-name/coverage/default.nix
|    | ❌  |    | plugins/by-name/cursorline/default.nix
|    | ❌  |    | plugins/by-name/dap/dap-go.nix
|    | ❌  |    | plugins/by-name/dap/dap-python.nix
|    | ❌  |    | plugins/by-name/dap/dap-ui.nix
|    | ❌  |    | plugins/by-name/dap/dap-virtual-text.nix
|    | ❌  |    | plugins/by-name/diffview/default.nix
|    | ❌  | ⚠  | plugins/by-name/fidget/default.nix
|    | ❌  |    | plugins/by-name/git-worktree/default.nix
|    | ❌  |    | plugins/by-name/gitlinker/default.nix
|    | ❌  |    | plugins/by-name/harpoon/default.nix
|    | ❌  |    | plugins/by-name/illuminate/default.nix
|    | ❌  |    | plugins/by-name/image/default.nix
|    | ❌  |    | plugins/by-name/inc-rename/default.nix
|    | ❌  |    | plugins/by-name/lastplace/default.nix
|    | ❌  |    | plugins/by-name/lean/default.nix
|    | ❌  | ⚠  | plugins/by-name/lspsaga/default.nix
|    | ❌  | ⚠  | plugins/by-name/luasnip/default.nix
|    | ❌  |    | plugins/by-name/mark-radar/default.nix
|    | ❌  |    | plugins/by-name/marks/default.nix
|    | ❌  |    | plugins/by-name/mkdnflow/default.nix
|    | ❌  |    | plugins/by-name/multicursors/default.nix
|    | ❌  |    | plugins/by-name/navbuddy/default.nix
|    | ❌  | ⚠  | plugins/by-name/neo-tree/default.nix
[...]

PR #3573, opened 2025-07-21
plugins/nvim-tree: migrate to mkNeovimPlugin

saygo-png:
Related tracking issue: https://github.com/nix-community/nixvim/issues/2638

Sorry, the diff is huge because the original was around 1300 lines.
This a reduction of 1000 lines, making this MUCH easier to maintain.
I respect Lepage for having the will to originally write this many options.

Some extraOptions remain because they are not setup settings.

There is a deprecation that is supposed to be removed after 24.11.
I kept it because the deprecation discussion issue is still open.


saygo-png:
Rebase

plugins/by-name/nvim-tree/deprecations.nix
```diff
@@ -0,0 +1,182 @@
+lib: {
+  imports = [
+    (lib.mkRemovedOptionModule [
+      "plugins"
+      "nvim-tree"
+      "view"
+      "hideRootFolder"
+    ] "Set `plugins.nvim-tree.renderer.rootFolderLabel` to `false` to hide the root folder.")
```
MattSturgeon: This could be done using `mkChangedOptionModule`, but if you don't think it's worth the effort, I'm fine with that.

plugins/by-name/nvim-tree/deprecations.nix
```diff
@@ -0,0 +1,182 @@
...
+    })
+
+  ];
+
+  # Deprecations after mkNeovimPlugin migration
+  deprecateExtraOptions = true;
+
+  optionsRenamedToSettings =
+    let
+      mkOptionPaths = map (lib.splitString ".");
```
MattSturgeon: Clever way to improve how the option paths are formatted!

Personally, I'd inline it to remove the let..in block, though:

```nix
optionsRenamedToSettings = map (lib.splitString ".") [
  "disableNetrw"
  "hijackNetrw"
  "autoReloadOnWrite"
  #...
```

plugins/by-name/nvim-tree/deprecations.nix
```diff
@@ -0,0 +1,182 @@
...
+
+    # TODO: added 2025-04-07, remove after 25.05
+    (lib.nixvim.mkRemovedPackageOptionModule {
+      plugin = "nvim-tree";
+      packageName = "git";
+    })
+
+  ];
+
+  # Deprecations after mkNeovimPlugin migration
```
MattSturgeon: Ideally, this would have a datestamp for posterity (for quick reference without needing a `git blame`)

MattSturgeon (approved):
I've not reviewed each removed option, but I trust the overall quality of the PR.

Thanks for tackling this!

plugins/by-name/nvim-tree/deprecations.nix
saygo-png: This had to be changed because the path in the warning message becomes deprecated with this PR. Sorry for missing that.

saygo-png:
Big diff again just due to the inlining of the partially applied map.

plugins/by-name/nvim-tree/deprecations.nix
```diff
@@ -0,0 +1,188 @@
...
+          "hideRootFolder"
+        ];
+      in
+      lib.mkChangedOptionModule oldOptPath [
+        "plugins"
+        "nvim-tree"
+        "settings"
+        "renderer"
+        "root_folder_label"
+      ] (config: !lib.getAttrFromPath oldOptPath config)
```
MattSturgeon, line 18: This matches the documented example: https://github.com/NixOS/nixpkgs/blob/aa5367d94d4710035ce6c58a19ba5ecfd9866e8d/lib/modules.nix#L1668

However IMO it's simpler to hard-code it, without using `getAttrFromPath`:

```suggestion
      ] (config: !config.plugins.nvim-tree.view.hideRootFolder)
```

I'm guessing the example uses that because other rename/deprecation functions take a generic option-path and can't hard-code the attrpath?

MattSturgeon: Although, I guess you'd need to declare `oldOptPath` anyway, for the "from" argument. So maybe my suggestion doesn't make sense after all.

MattSturgeon (approved):

GaetanLepage:
Thanks @saygo-png !!!

PR #3574, opened 2025-07-21
plugins/rainbow: init

saygo-png:
https://github.com/luochen1990/rainbow/

Something easier to review from me this time.

This plugin adds rainbow parentheses. I'm aware that we already have rainbow-delimiters, but that uses treesitter. Not everyone likes that project. Even though I do use treesitter I find that this non TS implementation works much faster and does the same thing.

I enable the active setting by default because that is described as part of "setup" in the upstream documentation.

### DIFF
--- a/plugins/by-name/rainbow/default.nix
+++ b/plugins/by-name/rainbow/default.nix
@@ -1,0 +1,38 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "rainbow";
+  package = "rainbow";
+  description = "Rainbow parentheses improved — shorter code, no level limit, smooth and fast with powerful configuration.";
+  maintainers = [ lib.maintainers.saygo-png ];
+  globalPrefix = "rainbow_";
+
+  settingsExample = {
+    active = 1;
+    conf = {
+      guifgs = [
+        "#7d8618"
+        "darkorange3"
+        "seagreen3"
+        "firebrick"
+      ];
+      operators = "_,_";
+      parentheses = [
+        "start=/(/ end=/)/ fold"
+        "start=/\\[/ end=/\\]/ fold"
+      ];
+      separately = {
+        "*" = { };
+        markdown = {
+          parentheses_options = "containedin=markdownCode contained";
+        };
+        haskell = {
+          parentheses = [
+            "start=/\\[/ end=/\\]/ fold"
+            "start=/\v\{\ze[^-]/ end=/}/ fold"
+          ];
+        };
+        css = 0;
+      };
+    };
+  };
+}
--- a/tests/test-sources/plugins/by-name/rainbow/default.nix
+++ b/tests/test-sources/plugins/by-name/rainbow/default.nix
@@ -1,0 +1,112 @@
+{
+  empty = {
+    plugins.rainbow.enable = true;
+  };
+
+  defaults = {
+    plugins.rainbow = {
+      enable = true;
+      settings = {
+        conf = {
+          guifgs = [
+            "royalblue3"
+            "darkorange3"
+            "seagreen3"
+            "firebrick"
+          ];
+          ctermfgs = [
+            "lightblue"
+            "lightyellow"
+            "lightcyan"
+            "lightmagenta"
+          ];
+          guis = [ "" ];
+          cterms = [ "" ];
+          operators = "_,_";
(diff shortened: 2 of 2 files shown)


GaetanLepage:
When you create such a PR, can you include a link to the plugin's homepage?

> I enable the active setting by default because that is described as part of "setup" in the upstream documentation.

I don't think that this is a reason for us to enable it by default.
If the default value is `0`, then we should leave that.

Also, can you add `settingsExample` and an `example` test case in the test file please?

saygo-png:
> When you create such a PR, can you include a link to the plugin's homepage?

Sure thing.

> I don't think that this is a reason for us to enable it by default. If the default value is `0`, then we should leave that.

Removed it.

> Also, can you add `settingsExample` and an `example` test case in the test file please?

Added an extensive example taken mostly from upstream, i only changed one of the colors to a hex notation to show that you can also define colors that way.

plugins/by-name/rainbow/default.nix
```diff
@@ -0,0 +1,66 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "rainbow";
+  package = "rainbow";
+  description = "Rainbow parentheses improved — shorter code, no level limit, smooth and fast with powerful configuration.";
+  maintainers = [ lib.maintainers.saygo-png ];
+  globalPrefix = "rainbow_";
+
+  settingsExample = {
```
GaetanLepage, line 9: You can leave the `example` test case as is, but `settingsExample` here is used to generate the documentation. Hence, it would be better to shorten it to 5-10 options, otherwise it becomes quite unreadable.

tests/test-sources/plugins/by-name/rainbow/default.nix
```diff
@@ -0,0 +1,68 @@
+{
+  empty = {
+    plugins.rainbow.enable = true;
+  };
+
```
GaetanLepage, line 5: Could you please add a `defaults` test case that explicitly sets each option to its default value, as documented upstream?

saygo-png: I've added what i could, but sadly i just don't have the regex skills to properly escape these complex expressions in nix, in such a way that they also get escaped in lua while preserving their original meaning.
You can see what is missing from the defaults and the regex horror here: <https://github.com/luochen1990/rainbow/blob/76ca1a20aa42edb5c65c19029968aad4625790dc/autoload/rainbow_main.vim>

plugins/by-name/rainbow/default.nix
saygo-png: The defaults set by the plugin are almost identical to the example i took from the readme. So the default and example test cases almost test the same thing. Should i remove the example test case because of that?

khaneliman: Preferred to have `settingsExample` and `example` be more real world configuration examples.

tests/test-sources/plugins/by-name/rainbow/default.nix
```diff
@@ -0,0 +1,145 @@
...
+  default = {
```
khaneliman:
```suggestion
  defaults = {
```

Minor nit, we use this naming convention.

tests/test-sources/plugins/by-name/rainbow/default.nix
```diff
@@ -0,0 +1,145 @@
...
+            css = 0;
+            sh = 0;
+            vimwiki = 0;
+          };
+        };
+      };
+    };
+  };
+
+  example = {
```
khaneliman, line 83: Would pare this down to only fields that are changed from `defaults`

plugins/by-name/rainbow/default.nix
saygo-png: > Preferred to have `settingsExample` and `example` be more real world configuration examples.

I can't really come up with anything more practical here. The issue is that the configuration for this plugin is mostly just colors, the rest is defined by default and works well. Majority of the fancy options are for fixing compatibility with other plugins but I've never had to use them.

Current example does show how you can disable filetypes, basic regex escaping from nix, and configure the colors which id wager is what most people will ever do with this plugin's configuration.

GaetanLepage (approved):
Thanks! Sorry for the review delay.

PR #3575, opened 2025-07-21
plugins/faster: init

saygo-png:
https://github.com/pteroctopus/faster.nvim/

This plugin is very useful when working with large files.

### DIFF
--- a/tests/test-sources/plugins/by-name/faster/default.nix
+++ b/tests/test-sources/plugins/by-name/faster/default.nix
@@ -1,0 +1,113 @@
+{
+  empty = {
+    plugins.faster.enable = true;
+  };
+
+  defaults = {
+    plugins.faster = {
+      enable = true;
+      settings = {
+        behaviours = {
+          bigfile = {
+            on = true;
+            features_disabled = [
+              "illuminate"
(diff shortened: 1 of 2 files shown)


saygo-png:
I added a config example and a corresponding test case.

saygo-png:
Added a "default" test case using explicit plugin defaults. Shortened the example a bit for nicer docs.

tests/test-sources/plugins/by-name/faster/default.nix
```diff
@@ -0,0 +1,113 @@
+{
+  empty = {
+    plugins.faster.enable = true;
+  };
+
+  default = {
```
GaetanLepage:
```suggestion
  defaults = {
```
nit (let's try to keep this consistent with other tests).

GaetanLepage (commented):
LGTM overall!

tests/test-sources/plugins/by-name/faster/default.nix
saygo-png: Sorry, missed that.

GaetanLepage (approved):
LGTM, thanks!

PR #3580, opened 2025-07-22
plugins/dashboard: add an option for nvim-dashboard

Fovir-GitHub:
[nvim-dashboard](https://github.com/nvimdev/dashboard-nvim?tab=readme-ov-file#hyper) had supported the option of enabling `mru` list. And this PR add the option `plugins.dashboard.settings.config.mru.enable` to achieve the function.


plugins/by-name/dashboard/default.nix
```diff
@@ -397,6 +397,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
               ${requiresTheme "hyper"}
             ''
             {
+              enable = helpers.defaultNullOpts.mkBool true ''
```
khaneliman, line 400: @MattSturgeon  do we want to continue limiting the `settings` with these sub options? Or remove them and let it be more freeform?

khaneliman (commented):
Not as familiar with `mkCompositeOption`, but looks like we do need to declare suboptions for them to be used from what I can see in lib.

plugins/by-name/dashboard/default.nix
MattSturgeon: `mkCompositeOption` is basically a wrapper around making a submodule option.

It automatically adds a freeformType to the submodule.

Therefore, we shouldn't be restricted to the explicitly declared options; it should already be freeform.

Although having explicitly declared options does of course add the additional type-checking and custom type-merging that's defined by those options.

FovirDev: Thanks for explanation!

At first, I noticed that the `project` has a suboption `enable` while `mru` does not, which made me confused. That is the reason why I opened this pull request. In my opinion, declaring this option explicitly could make the configuration more convenient and clear.

Despite this, I am happy to close this PR if it is not truly necessary.

MattSturgeon (approved):
It'd probably be better (long term) to remove the explicitly declared options, rather than add new ones.

However, until then, this should reduce confusion surrounding our vs their `enable` option.

PR #3581, opened 2025-07-25
plugins/lsp: use new lsp module to implement plugins.lsp.inlayHints (two-way alias)

GaetanLepage:


plugins/lsp/default.nix
```diff
@@ -168,15 +171,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
               options = {
                 inherit (cfg.keymaps) silent;
                 desc = "${descPrefix} ${actionStr}";
-              } // actionProps;
+              }
+              // actionProps;
```
MattSturgeon: Unrelated formatting change.

Did this get missed in the update, or does this need rebasing now the update is merged?

PR #3588, opened 2025-07-28
license: various fixes 

awwpotato:
including "the \<project\> contributors" or similar in the copyright notice is best practice, also updating the year is too obviously

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,8 +1,19 @@
 Copyright 2020-2025 NixVim contributors
 
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Thanks

LICENSE
```diff
@@ -1,8 +1,19 @@
-Copyright 2020 Pedro Alves <pta2002@pta2002.com>
+Copyright 2020-2025 Pedro Alves <pta2002@pta2002.com> and the NixVim contributors
```
MattSturgeon: Thanks. I think copyright is usually the last year the licence was updated? I'm not a lawyer, but having it as a range could wrongly imply that the copyright expires at 2025.

As for the copyright holder, if @pta2002 is ok with the change, I'd prefer the simple wording:


```suggestion
Copyright 2025 Nixvim contributors
```

LICENSE
```diff
@@ -1,8 +1,19 @@
...
 
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
```
MattSturgeon, line 8: My preference would be to use the same line wrapping style as the original/upstream licence.

Unlike Stylix, Nixvim doesn't have a strict 80 char line limit.

Generally, we use semantic line wrapping, but we also make exceptions for when it makes sense to have long unwrapped lines.

EDIT: I'm not blocking on this, just expressing my preference.

LICENSE
MattSturgeon: Alternatively, it is also common to have multiple copyright lines. This typically happens when projects are forked.

```suggestion
Copyright 2020 Pedro Alves <pta2002@pta2002.com>
Copyright 2025 Nixvim contributors
```

LICENSE
0xda157: > My preference would be to use the same line wrapping style as the original/upstream licence.

MIT license doesn't hard wrap on 80 chars on their website because it expected that the browser will wrap the text, this isn't the case for a lot of text editors. Most projects use the version of the MIT license wrapped to 80 chars and it's what GitHub uses when creating a new project. Also going to https://mit-license.org/license.txt gives the version wrapped to 80 chars so it seems like it's indented to be wrapped when displayed as plaintext.

> Unlike Stylix, Nixvim doesn't have a strict 80 char line limit.
> 
> Generally, we use semantic line wrapping, but we also make exceptions for when it makes sense to have long unwrapped lines.

I'm aware of not having a char line limit, I'm just wrapping here because the current formatting is unreadable with line wrapping turn off in editor (or a very wide editor window)

LICENSE
0xda157: > Thanks. I think copyright is usually the last year the licence was updated? I'm not a lawyer, but having it as a range could wrongly imply that the copyright expires at 2025.

It's what every major project uses for their copyright headers so I assume it's legally sound. Removing the original copyright date is probably a bad idea. Changing the copyright holder probably isn't a big deal because @pta2002 would fall under the NixVim contributors, but it's not generally what I've seen done by other projects.

pta2002 (approved):
As the original copyright owner I am fully OK with this change (in fact, it makes a lot more sense to me to just be "nixvim contributors").

So take this as my OK to bundle me in as part of that group, if you'd like. If I'd known the size of the project when I added that license file this is how I'd do it from the get-go :P

LICENSE
MattSturgeon: Ok, based on discussion so far, I think we can go for:


```suggestion
Copyright 2020-2025 Nixvim contributors
```

?

PR #3592, opened 2025-07-29
ISSUE_TEMPLATES: convert to yml

0xda157:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+      required: true
+  - type: dropdown
+    attributes:
+      label: Home Manager Release
+      multiple: false
+      options:
+        - "unstable"
+        - "25.05"
+        - "24.11"
+        - "other"
```
MattSturgeon, line 30: Should have a "none" option?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+  - type: checkboxes
+    attributes:
+      label: I have read the FAQ
+      options:
+        - label: >-
+            I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html)
+            and my bug is not listed there.
```
MattSturgeon, line 37: Does this support reference links?


```suggestion
            I have read the [FAQ] and my bug is not listed there.

            [FAQ]: https://nix-community.github.io/nixvim/user-guide/faq.html
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+          required: true
+  - type: textarea
+    attributes:
+      label: Description
+      description: See https://stackoverflow.com/questions/29416909
```
MattSturgeon: Do these descriptions support markdown or html? It'd be nice to have a human readable link instead, e.g.:


```suggestion
      description: See [How can I write good bug reports?](https://stackoverflow.com/questions/29416909)
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+  - type: textarea
+    attributes:
+      label: Minimal, Reproducible Example (MRE)
+      description: See https://stackoverflow.com/help/minimal-reproducible-example
```
MattSturgeon: Same. Although, at least this link has a title-based path that makes sense to a human.

.github/ISSUE_TEMPLATE/plugin_request.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: Plugin Request
+description: Request support for a new plugin
+title: "[PLUGIN REQUEST] <PLUGIN>"
+labels: ["enhancement"]
```
MattSturgeon, line 5: Adding a bespoke "plugin request" label here would help us simplify our Plugin Requests project board...

Currently it adds issues with matching titles, but that's kinda hanky.

This matches the existing behaviour though, so I'm just noting this for a future PR.

.github/ISSUE_TEMPLATE/plugin_request.yml
```diff
@@ -0,0 +1,25 @@
...
+      description: Plugin NixVim should support
+  - type: input
+    attributes:
+      label: Homepage
+      description: URL to the plugin repository
+  - type: input
+    attributes:
+      label: Nixpkgs
+      description: Nixpkgs URL or `false`
+      placeholder: "`false`"
```
MattSturgeon: This matches the existing form, but I'm not convinced we really need the "or false" bit, or the placeholder?

.github/ISSUE_TEMPLATE/plugin_request.yml
```diff
@@ -0,0 +1,25 @@
...
+  - type: textarea
+    attributes:
+      label: Extra Information
+      description: >-
+        Provide extra information about the plugin, and any ideas regarding
+        NixVim-specific options.
```
MattSturgeon: Nit: if we're wrapping, I'd personally prefer to do it on the comma, rather than at an arbitrary line length.


Also, we are rather inconsistent on nixvim vs Nixvim vs NixVim. I think Nixvim is the most canonical style?

.github/ISSUE_TEMPLATE/bug_report.yml
0xda157: it's not required so it defaults to None
image

.github/ISSUE_TEMPLATE/bug_report.yml
0xda157: it doesn't from my testing

.github/ISSUE_TEMPLATE/plugin_request.yml
0xda157: I think the best way to do this would just make it not required instead of telling them to put `false`.

.github/ISSUE_TEMPLATE/plugin_request.yml
0xda157: > Also, we are rather inconsistent on nixvim vs Nixvim vs NixVim. I think Nixvim is the most canonical style?

I'm just copying over what was in the markdown version here. It seems like the README uses NixVim so I assumed that was the preferred style.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,51 @@
...
+          required: true
+  - type: textarea
+    attributes:
+      label: Description
+      description: See [How can I write good bug reports?](https://stackoverflow.com/questions/29416909)
+  - type: textarea
+    attributes:
+      label: Minimal, Reproducible Example (MRE)
+      description: See [How to create a Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)
+      render: nix
```
0xda157, line 47: either we can set the render to nix or provide a default value with the markdown code block, using the render option is cleaner but might be confusing/end up in poorly formatted issues.

.github/ISSUE_TEMPLATE/plugin_request.yml
MattSturgeon: Maybe we can find a way to enforce a consistent name style using treefmt 🤔

.github/ISSUE_TEMPLATE/bug_report.yml
MattSturgeon: Agreed. Here, we know that a MRE should usually be nix code. Committers can always edit issue descriptions if needed. I think it is best to mark as rendered as nix.

MattSturgeon (approved):
Thanks, again!

PR #3595, opened 2025-08-01
colorschemes/gruvbox-baby: init

saygo-png:
https://github.com/luisiacc/gruvbox-baby

This theme is a variation on gruvbox. I know we already have 2 gruvboxes, with a third coming soon.
Feel free to reject this if u feel as though its redundant. I'm not sure if nixvim is trying to be selective about inclusions or not, based on what is already packaged.

### DIFF
--- a/plugins/colorschemes/gruvbox-baby.nix
+++ b/plugins/colorschemes/gruvbox-baby.nix
@@ -1,0 +1,27 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gruvbox-baby";
+  maintainers = [ lib.maintainers.saygo-png ];
+  isColorscheme = true;
+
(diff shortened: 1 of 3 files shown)


plugins/colorschemes/gruvbox-baby.nix
```diff
@@ -0,0 +1,27 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gruvbox-baby";
+  maintainers = [ lib.maintainers.saygo-png ];
+  isColorscheme = true;
+
+  # Despite being a lua plugin, it's configured via globals without a setup function.
+  callSetup = false;
```
khaneliman, line 8:
```suggestion
  callSetup = false;
  hasLuaConfig = false;
```

PR #3608, opened 2025-08-07
colorschemes/gruvbox-material: init

saygo-png:
https://github.com/sainnhe/gruvbox-material

My favourite Neovim gruvbox.

Removes fresh deprecation in accordance with https://github.com/nix-community/nixvim/pull/3495#issuecomment-3071377763

There is a missing `defaults` test option `colors_override` which is by default [set to an empty vimscript dictionary](https://github.com/sainnhe/gruvbox-material/blob/d080710b58e605b45db0b0e4288205ca6f7587f0/autoload/gruvbox_material.vim#L35)
I excluded it because I could not figure out how to make it empty in nix without errors. neither `__raw = "{}"` or `__empty = null` worked.
I tested the option with actual settings and it works fine, i just can't figure out how to make it empty for the test.


plugins/deprecation.nix
```diff
@@ -23,12 +23,6 @@ let
       It is recommended to use `plugins.pckr` or `plugins.lazy` instead.
     '';
   };
-  renamed.colorschemes = {
-    # Added 2025-08-06
```
MattSturgeon, line 27: I'd like to wait a few weeks before merging, to give users chance to update and see this deprecation warning.

Some time in September should be ok?

MattSturgeon (commented):
LGTM

plugins/deprecation.nix
saygo-png: Sure.

saygo-png: @MattSturgeon Could this be merged now?

MattSturgeon (approved):
SGTM, thanks!

GaetanLepage (approved):
Thanks!

Issue #3561, opened 2025-07-15
CI: Check formatting using GHA


MattSturgeon:
Rather than having buildbot check treefmt, we could do this with a GHA workflow.

The advantage of doing this is that it is immediately obvious from the status check name that the failure is formatting related.

Additionally, we avoid running the check multiple times (on all platforms), although we can also solve that without having to resort to GHA.

PR #3611, opened 2025-08-09
ci: added treefmt workflow to gha and removed from buildbot

c4patino:
Seemed like low hanging fruit in [#3561](https://github.com/nix-community/nixvim/issues/3561)

- Migrates the treefmt foramtting check from buildbot to a dedicated Github Actions workflow.
- Removes the buildbot treefmt check from flake/dev/fmt.nix to avoid duplicate runs.


.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,29 @@
...
+  treefmt:
+    name: Check formatting with treefmt
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+      - name: Run treefmt check (nix flake check)
+        run: nix flake check --show-trace
```
MattSturgeon: This should build the specific derivation, rather than all checks.

E.g. `nix build .#checks.x86_64-linix.treefmt`

(It may not be named `treefmt`, I'm on my phone and can't verify)

.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,29 @@
...
+      - "LICENSE"
+      - "flake.lock"
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "**.md"
+      - "**.svg"
+      - "LICENSE"
+      - "flake.lock"
```
MattSturgeon: Ignoring files should be done by the treefmt configuration, not by GHA.

Although, if this is a true subset of our treefmt excludes then it is okay as a performance optimisation... Assuming we remember to keep it up to date if the treefmt config changes.

.github/workflows/treefmt.yml
c4patino: Oh! I didn't even know you can interact with checks like that. You learn something new everyday I guess. Fixed (it is called treefmt btw; https://github.com/numtide/treefmt-nix/blob/1298185c05a56bff66383a20be0b41a307f52228/flake-module.nix#L73-L75).

.github/workflows/treefmt.yml
c4patino: It is indeed a subset. This was actually intended to ignore changes to the filetypes which are most frequently changed. Minimum here is flake.lock imo since that changes so much.

c4patino:
Hi @MattSturgeon is there anything else that you need on this?

.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "**.md"
+      - "**.svg"
+      - ".gitignore"
+      - "LICENSE"
+      - "flake.lock"
```
MattSturgeon: We don't really need to do this on `push`, since we require all changes go through a PR. But we should do it on `merge_group` so that it runs on the merge queue.

Additionally, we'll need to add this or a meta-job to our “required status checks”; cc @GaetanLepage 

I'll try and look into adding a meta-job we can use for this, similar to the `no pr failures` job in nixpkgs.

.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,31 @@
+name: Formatting
+
+on:
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - "**.svg"
+      - ".gitignore"
+      - "LICENSE"
+      - "flake.lock"
```
MattSturgeon: GHA recently added support for yaml anchors and references, so we can avoid needing to write this list twice.

The first instance should declare the anchor with `&ignored`, then the second instance can be replaced with a reference to it with `*ignored`


```suggestion
    paths-ignore: &ignored
      - "**.md"
      - "**.svg"
      - ".gitignore"
      - "LICENSE"
      - "flake.lock"
```

Not relevant if we're removing the `on: push` event, though.

---

It'd also be nice to see a comment on this list, cross-referencing it with the treefmt-nix configuration so we remember to keep both in sync.

.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  treefmt:
+    name: Check formatting with treefmt
```
MattSturgeon: This name shows up in the status check list as `Workflow name / Job name`; in this case `Formatting / Check formatting with treefmt`.

Bike-shedding the names **doesn't** block this PR, because we can change them at any time in other PRs. But my personal preference would be to be less verbose and avoid redundancy between workflow & job names.

How about `Formatting / treefmt` or `Lint / formatting` or `Check / Formatting`?

.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+      - name: Run treefmt check
+        run: nix build .#checks.x86_64-linux.treefmt
```
MattSturgeon: Running with `--accept-flake-config` will cause the flake config to be used. That's useful here because it configures the nix-community cachix.

Alternatively, we could use a cachix GHA step, like we do in some other workflows. That way we can also _push_ builds to cachix.

.github/workflows/treefmt.yml
MattSturgeon: I'll handle merge queues and required checks in #3628

For now it is fine to just remove the `on: push` event.

```suggestion
```

.github/workflows/treefmt.yml
wolfgangwalther: > GHA recently added support for yaml anchors and references, so we can avoid needing to write this list twice.

whoot? TIL. That's great to hear. Although from looking into it, it seems to be pretty limited :/.

MattSturgeon: Support is tracked in https://github.com/actions/runner/issues/1182

I assume most limitations come from workflows not supporting arbitrary unknown keys?

wolfgangwalther: Arbitrary keys + merge keys were what I used a lot for GitLab. Also GitLab can reference "templates" across files (different syntax, but same concepts). This would be so much better than re-usable workflows or custom actions for code reuse...

.github/workflows/treefmt.yml
c4patino: switching to `lint / treefmt`. forgot that those show up as really long. i've had problems with that before when using reusable workflows.

.github/workflows/treefmt.yml
c4patino: yep, push has been removed.

.github/workflows/treefmt.yml
c4patino: Wow! Didn't even know that was possible without an explicit cachix step. I don't think its really necessary for this one to add a cachix step since it shouldn't (at least not for any reason i can think of) need to push to cachix.

.github/workflows/treefmt.yml
c4patino: so are we going to have gaetan take it from here? I'm afraid I've never used meta-jobs or merge queues before so I'll be rather unhelpful here :disappointed:. looking into `merge_group` though. never used that tag before.

MattSturgeon: I realise I linked to this PR (#3611), I had meant to link to #3628 (fixed).

c4patino: Ah I see. I was wondering if we going to extend this PR to essentially do what you are describing in #3628. For some reason the backlink from the mention of this from the other PR isn't showing up for me.

.github/workflows/treefmt.yml
MattSturgeon: FYI you renamed the job id to `lint`, not the workflow name. The workflow `name: Formatting` is at the top of the file 🙃 (and should probably be somewhat in-sync with the actual filename, currently `treefmt.yml`, too).

Also, if the job name and id are the same, there's no need to set its name explicitly.

c4patino: thats embarassing. fixed. it should just be `lint / treefmt` now.

PR #3622, opened 2025-08-12
plugins/blink-cmp: add "exact" fuzzy sort

shimeoki:
"exact" value from [the documentation](https://cmp.saghen.dev/configuration/fuzzy.html#built-in-sorts) was missing

### DIFF
--- a/plugins/by-name/blink-cmp/settings-options.nix
+++ b/plugins/by-name/blink-cmp/settings-options.nix
@@ -443,6 +443,7 @@ in
     sorts =
       defaultNullOpts.mkListOf
         (types.enum [
+          "exact"
           "label"
           "sort_text"
           "kind"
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
Could benefit from being more `freeform` with a slight tweak to support any string, and use the enum as a guide.

budimanjojo:
Can this be backported to `25.05` branch?

shimeoki:
did i understand correctly that backporting in this case means "make the same change in the nixos-25.05 branch"? pretty new in nix ecosystem, so asking just to be sure.

as i saw, the bot does these changes automatically. is that ok if i just open a PR manually?

budimanjojo:
I think this workflow (https://github.com/nix-community/nixvim/actions/workflows/backport.yml) should automatically backport this to the 25.05 branch. But the last run didn't do it, not sure why.

PR #3624, opened 2025-08-13
fix(copilot-chat): fix default model to one supported

stuart-warren:
it seems that gpt-4 is no longer supported

https://docs.github.com/en/copilot/reference/ai-models/supported-models

using gpt-4 throws 400 Bad Request errors

fixes nix-community#3623

### DIFF
--- a/plugins/by-name/copilot-chat/default.nix
+++ b/plugins/by-name/copilot-chat/default.nix
@@ -31,8 +31,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
       System prompt to use.
     '';
 
-    model = defaultNullOpts.mkStr "gpt-4" ''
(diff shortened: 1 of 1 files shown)


plugins/by-name/copilot-chat/default.nix
```diff
@@ -31,8 +31,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
       System prompt to use.
     '';
 
-    model = defaultNullOpts.mkStr "gpt-4" ''
-      GPT model to use, 'gpt-3.5-turbo' or 'gpt-4'.
+    model = defaultNullOpts.mkStr "gpt-4.1" ''
+      GPT model to use, 'gpt-3.5-turbo' or 'gpt-4.1'.
```
khaneliman, line 35: Should probably be updated to mention other possible models, like claude.

PR #3627, opened 2025-08-18
plugins/dap: add pipe type adapter

FredeHoey:
These are for pipe based debuggers, that don't use stdout but rather a unix pipe. The cmake debugger is an example of this and would be configured something like this

```nix
dap = {
  adapters = {
    pipes = {
      cmake = {
        pipe = "$\{pipe}";
        executable = {
            command = "cmake";
            args = [ "--debugger" "--debugger-pipe" "$\{pipe}" "build" ];
        };
      };
    };
  };
};
```

### DIFF
--- a/plugins/by-name/dap/dapHelpers.nix
+++ b/plugins/by-name/dap/dapHelpers.nix
@@ -77,6 +77,25 @@ rec {
     '';
   };
 
+  pipeAdapterOption = mkAdapterType {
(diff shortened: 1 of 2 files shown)


FredeHoey:
@khaneliman done :)

khaneliman:
> @khaneliman done :)

I pasted the message to use, we use a different commit style. It also wouldn't qualify as a fix for conventional commits since it's a new feature.

FredeHoey:
@khaneliman Totally missed that. Sorry

PR #3628, opened 2025-08-18
ci: init `pr.yml` and `merge-queue.yml` workflows

MattSturgeon:
These workflows will serve as top-level entry points for running GHA jobs against PRs and merge groups in the merge queue.

This allows us to have a meta-job "no PR failures" that "needs" all merge requirements. This makes it easier to maintain the "required status checks" ruleset.

Additionally, by having separate workflows for `pull_request` and `merge_group`, we can have different "needs" for each.

This is related to #3611, which is adding the first GHA-based check that will need to be "required".

This is based on @wolfgangwalther's excellent work in nixpkgs: [merge-group.yml](https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/merge-group.yml) and [pr.yml](https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/pr.yml)

The PR workflow ran Successfully here: https://github.com/nix-community/nixvim/actions/runs/17108993910/job/48525322498?pr=3628

We won't see the Merge Queue workflow run until this PR enters the queue.

@GaetanLepage you will need to add `PR checks successful` to the required status checks list in our ruleset. EDIT: the actual status check name may change throughout this PR.

image

### DIFF
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -1,0 +1,39 @@
+name: Merge Queue
+
+on:
+  merge_group:
+
+permissions: {}
+
+jobs:
+  lint:
+    uses: ./.github/workflows/lint.yml
+    with:
+      ref: ${{ github.event.merge_group.head_sha }}
+
+  # This job posts the "Required Status Checks" to satisfy our ruleset.
+  required-checks:
+    # It "needs" all the jobs that should block the Merge Queue.
+    # Modify this list to add or remove jobs from required status checks.
+    needs:
+      - lint
+
+    name: Required checks
+    runs-on: ubuntu-24.04-arm
+    permissions:
+      statuses: write
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const { serverUrl, repo, runId, payload } = context
+            await github.rest.repos.createCommitStatus({
+              ...repo,
+              sha: payload.merge_group.head_sha,
+              target_url: `${serverUrl}/${repo.owner}/${repo.repo}/actions/runs/${runId}`,
+              // WARNING:
+              // Do NOT change the context name or it will not match the ruleset.
+              // This would prevent all PRs from merging.
+              context: 'PR checks successful',
+              state: 'success',
+            })
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,43 @@
+name: PR
+
+on:
+  pull_request_target:
+
+concurrency:
+  group: pr-${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  lint:
+    uses: ./.github/workflows/lint.yml
+    with:
+      ref: ${{ github.event.pull_request.head.sha }}
+
+  # This job posts the "Required Status Checks" to satisfy our ruleset.
+  required-checks:
+    # It "needs" all the jobs that should block merging a PR.
+    # Modify this list to add or remove jobs from required status checks.
+    needs:
+      - lint
+
+    name: Required checks
+    runs-on: ubuntu-24.04-arm
+    permissions:
+      statuses: write
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const { serverUrl, repo, runId, payload } = context
+            await github.rest.repos.createCommitStatus({
+              ...repo,
+              sha: payload.pull_request.head.sha,
+              target_url: `${serverUrl}/${repo.owner}/${repo.repo}/actions/runs/${runId}`,
+              // WARNING:
+              // Do NOT change the context name or it will not match the ruleset.
+              // This would prevent all PRs from merging.
+              context: 'PR checks successful',
+              state: 'success',
+            })
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,13 +1,11 @@
 name: lint
 
 on:
-  pull_request:
-    paths-ignore:
-      - "**.md"
-      - "**.svg"
-      - ".gitignore"
-      - "LICENSE"
-      - "flake.lock"
+  workflow_call:
+    inputs:
+      ref:
+        required: true
+        type: string
 
 permissions:
   contents: read
@@ -18,6 +16,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+
       - name: Install Nix
         uses: cachix/install-nix-action@v31
         with:


.github/workflows/pr.yml
```diff
@@ -0,0 +1,30 @@
+name: PR
+
+on:
+  pull_request:
```
wolfgangwalther: You don't do any of the `pull_request` vs `pull_request_target` dance in this repo, right? Because that leads to problems for us in Nixpkgs (currently working towards a fix), where one of the two triggers finishes early and makes `no-pr-failures` succeed. Then the auto-merge goes through, without even considering the other workflow - which might fail.

As long as you only have a single trigger here, this should be fine, though.

khaneliman: We only use `pull_request_target` for tagging maintainers, iirc.

MattSturgeon: If this workflow needs access to `secrets` or additional permissions, then it'd need to use `pull_request_target`. Currently, I assume it'd be running jobs that don't need these permissions, so `pull_request` is fine.

---

I see your point about running the `no PR failures` job on both events though.

Maybe the way to deal with that is to have separate workflow files for each event, and not list the job at all in the `pull_request` workflow? Similar to how we deal with `merge_group` and `pull_request_target` events/workflows.

Or alternatively, have each child workflow declare its own `on: pull_request: paths` so that it runs independently without the `no PR failures` job?

wolfgangwalther: > Maybe the way to deal with that is to have separate workflow files for each event, and not list the job at all in the `pull_request` workflow? Similar to how we deal with `merge_group` and `pull_request_target` events/workflows.

Yes, that's my goal for Nixpkgs. I just need to reduce duplication for that scenario first.

MattSturgeon:
Rebased and added `lint.yml` to the required jobs list.

This does mean that `lint.yml` now runs unconditionally, since there's no `paths-ignore` equivalent for jobs or `workflow_call`.

It may be possible to use something like https://github.com/dorny/paths-filter in a `prepare` job and make the `lint` job conditional on its output, but I'd rather avoid the complexity for now.

EDIT: ran successfully here: https://github.com/nix-community/nixvim/actions/runs/17108993910/job/48525322498?pr=3628

khaneliman:
> Rebased and added `lint.yml` to the required jobs list.
> 
> This does mean that `lint.yml` now runs unconditionally, since there's no `paths-ignore` equivalent for jobs or `workflow_call`.
> 
> It may be possible to use something like https://github.com/dorny/paths-filter in a `prepare` job and make the `lint` job conditional on its output, but I'd rather avoid the complexity for now.
> 
> EDIT: ran successfully here: https://github.com/nix-community/nixvim/actions/runs/17108993910/job/48525322498?pr=3628

I did add `paths-filter` to home manager and it works well.

MattSturgeon:
> I did add `paths-filter` to home manager and it works well.

Good to know 👍🏻 

I'll still hold off for now.

One of the main arguments for not linting when specific paths are the only ones modified was that we update `flake.lock` frequently. However, even in that scenario it is possible that the update is bringing in changes to the formatters themselves, fixes from treefmt-nix, etc.

MattSturgeon:
I've [just pushed](https://github.com/nix-community/nixvim/compare/f0b0954bfa068f67bf2b0b2d75ec90f8748695d1..2eab9f26fb23b3baba5223d661db8065777730c8) a refactor based on https://github.com/NixOS/nixpkgs/pull/435929, with a few tweaks to wording and formatting.

This now uses @wolfgangwalther's new proposal of manually posting the required status check when the jobs are successful. This should prove simpler than of skipping it when successful and manually failing otherwise.

Being able to manually post status checks requires `permissions: statuses: write`. IIUC _any_ write permissions require using `pull_request_target` instead of `pull_request`, which means taking a bit more care over which revisions are checked out and run.

It also makes it more difficult to test changes to the workflow itself, since PRs will use the base branch's workflow instead of the PR's. That means we _won't_ see it running on this PR, unless we _also_ trigger it via `pull_request`.

.github/workflows/merge-queue.yml
MattSturgeon, line 1: Optionally, we could combine `pr.yml` and `merge-queue.yml` into a single workflow.

We'd need to do things like `${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha }}` and it'd be more difficult to specify _different_ checks for PRs vs the Queue.

But it would remove a lot of duplication.

Thoughts?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,43 @@
...
+      statuses: write
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const { serverUrl, repo, runId, payload } = context
+            await github.rest.repos.createCommitStatus({
+              ...repo,
+              sha: payload.pull_request.head.sha,
+              target_url: `${serverUrl}/${repo.owner}/${repo.repo}/actions/runs/${runId}`,
```
wolfgangwalther, line 37: Did you leave out the `?pr=` suffix out here on purpose? This is a bit annoying when clicking the link, because the "back" link from the summary page won't go to the PR, but to the actions tab.

.github/workflows/merge-queue.yml
wolfgangwalther: I think it makes a lot of sense to re-use the same file when you can. It looks like this should be easily possible here.

You could even change the [`run-name`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?s=09#run-name) so you don't get "PR" for both.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,43 @@
+name: PR
+
+on:
+  pull_request_target:
```
wolfgangwalther, line 4: One downside of going with the `pull_request_target` event: Changes to the workflow file now won't be tested in the respective PR anymore. Now you need to start looking into whether to add the `pull_request` trigger as well.

Overall, it *might* be simpler to use the non-javascript-status approach here with the old "skipped vs failed job" approach.

.github/workflows/pr.yml
MattSturgeon: Good catch. I copied this from your `merge-group.yml` without spotting that `pr.yml` had a different URL.

.github/workflows/pr.yml
MattSturgeon: > One downside of going with the `pull_request_target` event: Changes to the workflow file now won't be tested in the respective PR anymore. Now you need to start looking into whether to add the `pull_request` trigger as well.

Yeah, this is frustrating. Maybe I should consider adding a "testing" `on: pull_request` trigger to this PR so that it's _done_.

> Overall, it *might* be simpler to use the non-javascript-status approach here with the old "skipped vs failed job" approach.


I'm of two minds here:

1. On the one hand `pull_request` is much simpler.
2. On the other hand, if we want to do _anything_ that needs secrets or write permissions then we'll eventually need to switch.
   If so, we may as well do it now and get access to the better "required check" implementation.

I like the _idea_ of the better impl because we can choose a saner name for the required check from the start. It's also less confusing for contributors if we have a green check, rather than a grey one.

MattSturgeon:
I noticed in #3657 that we have configured the repo to require manual approval before running `pull_request` workflows on untrusted forks:

image

image


Switching to `pull_request_target` would also resolve this.

MattSturgeon:
We still need to mark `lint.yml`'s jobs as required status checks, but this PR is stale and I haven't had time to work on it.

I still think a meta-job is easier to maintain long term, but maybe it'll be simpler initially to ensure lint's job names are distinctive enough to be directly required?

PR #3631, opened 2025-08-20
plugins/contextfiles-nvim: init

### LINKED ISSUE #3331: [PLUGIN REQUEST] contextfiles.nvim
| Field    | Description |
|----------|-------------|
| Plugin   | `contextfiles.nvim`  |
| Homepage | [contextfiles.nvim](https://github.com/banjo/contextfiles.nvim)     |
| Nixpkgs  | `false` |

## Extra Information

Plugin that scans for *project files* e.g. `.cursor/rules`, basically files that better tell your LLM what to do and/or give it more context. This plugins is a very good extension to the already existing `codecompanion.nvim` plugin. A nixvim specifc option to integrate these two automatically according to [these](https://github.com/banjo/contextfiles.nvim?tab=readme-ov-file#codecompanion) instruction would probably be nice.

c4patino:
add support for [contextfiles.nvim](https://github.com/banjo/contextfiles.nvim)

Resolves #3331

### DIFF
--- a/plugins/by-name/contextfiles/default.nix
+++ b/plugins/by-name/contextfiles/default.nix
@@ -1,0 +1,14 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  moduleName = "contextfiles";
+  name = "contextfiles";
+  package = "contextfiles-nvim";
+
+  callSetup = false;
+  hasLuaConfig = false;
+  hasSettings = false;
+
+  maintainers = with lib.maintainers; [
+    c4patino
+  ];
+}
--- a/tests/test-sources/plugins/by-name/contextfiles/default.nix
+++ b/tests/test-sources/plugins/by-name/contextfiles/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.contextfiles.enable = true;
+  };
+}


GaetanLepage:
Hi!
Thanks for the contribution!

Please, rename the file (and the plugin) to just `contextfiles`.
Also add a `defaults` and `example` test case.
I encourage you to look at other recently packaged plugins for inspiration.

Feel free to ask your questions on matrix or in this thread :)

PS: Sorry for the delay...

c4patino:
Hi @GaetanLepage,

You're good, np! I was very confused by the test file actually. I'm afraid I'm a bit lost. That particular component is the one that is causing the failure, even though the package seems to build on my end. I'll take a look at the other plugins to see if I can figure it out.

Do you know why the builder is failing by any chance? I was trying to debug but I haven't figured out how the test cases get unpacked yet. it seems like the contextfiles-nvim option seems to not be exposed which is strange...

```
error: Cannot build '/nix/store/xjmar32fl7hpzb3lzf27vp3ghgms6xrk-empty.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/lj9h1hfa65nyvfmgxf86wjmrbs0a1skd-empty
       Last 9 log lines:
       > ERROR: Error detected while processing /nix/store/v3py62kbc06m809f9bbzz4md3bhhvz33-init.lua:
       > E5113: Error while calling lua chunk: /nix/store/v3py62kbc06m809f9bbzz4md3bhhvz33-init.lua:9: module 'contextfiles-nvim' not found:
       >      no field package.preload['contextfiles-nvim']
       >   no file '/nix/store/h5c0yi6zkdhdkhs67ph8x5s0c4sqjg5m-luajit-2.1.1741730670-env/share/lua/5.1/contextfiles-nvim.lua'
       >     no file '/nix/store/h5c0yi6zkdhdkhs67ph8x5s0c4sqjg5m-luajit-2.1.1741730670-env/share/lua/5.1/contextfiles-nvim/init.lua'
       >        no file '/nix/store/h5c0yi6zkdhdkhs67ph8x5s0c4sqjg5m-luajit-2.1.1741730670-env/lib/lua/5.1/contextfiles-nvim.so'
       > stack traceback:
       >       [C]: in function 'require'
       >      /nix/store/v3py62kbc06m809f9bbzz4md3bhhvz33-init.lua:9: in main chunk
       For full logs, run:
         nix log /nix/store/xjmar32fl7hpzb3lzf27vp3ghgms6xrk-empty.drv
error: Cannot build '/nix/store/xfz1ikkyl28qwrzd4m8w1yv427hhmayd-plugins-by-name-contextfiles-nvim.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/cgpy3bb769my1ivc8i54i0i8b9mjqafm-plugins-by-name-contextfiles-nvim
```

EDIT: nvm i see how the modules get mapped to the require function now.

c4patino:
Okay. Resolved all of the build issues and stuff. It turns out I misread how their documentation explains things. This is really supposed to be a module which is supposed to have `callSetup = false`. Reason why I misread it is because their README shows default options, but those are for functions seemingly and not for a setup function. I was looking through the actual repo and wasn't able to find a `M.setup` function exposed.

MattSturgeon:
> I was looking through the actual repo and wasn't able to find a `M.setup` function exposed.

If it has settings that are applied to a `setup`-style function, you can also specify a different name.

`mkNeovimPlugin` defaults its `setup` attribute to `".setup"`, but you can override it with anything.

If there are multiple functions that can receive structured config, then we may need to create multiple `settings`-like options, too.

Or if none of them are really a good fit for `settings`, then we may need `hasSettings = false`.

c4patino:
There aren't any settings. This package gets used through the exposed functions only. I can set `hasSettings = false`, didn't know that option existed. I was just settings `callSetup = false`.

MattSturgeon:
Might be worth skimming over
 - https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md
- https://github.com/nix-community/nixvim/blob/main/lib/plugins/mk-neovim-plugin.nix

to check which attributes are available and their documentation.

OTTOMH, we may also want `hasLuaConfig = false`? Although maybe it makes sense for users to put their calls to its functions in the plugins `luaConfig` option?

c4patino:
> OTTOMH, we may also want `hasLuaConfig = false`? Although maybe it makes sense for users to put their calls to its functions in the plugins `luaConfig` option?

added the hasLuaConfig false. calls are expected to be something more like this:

```lua
local context = require("contextfiles")

local current_file_path = vim.api.nvim_buf_get_name(0)
local files = context.get_context_files(current_file_path, { context_dir = ".cursor/rules" })
```
which doesn't make sense to have as a `luaConfig` option.

MattSturgeon:
SGTM. Thanks!

Issue #3181, opened 2025-04-12
[dev] Deprecation policy


GaetanLepage:
I want to try to reach a consensus on the deprecation (`lib.mkRe[nam,mov]edOptionModule`) removal policy.

If my understanding is correct, we already all agree on the following:
1. We should guarantee a seamless experience for stable channel users. This means that warnings tagged with `# TODO: added 2025-04-12 remove after 25.05` must be kept until (at least) the `25.05` branch-off.
2. If a (sizeable) performance hit was to be measured, it would be legitimate to remove all the "old" warnings.

Having said that, we need to decide on what to do with "old" warnings in the absence of a performance hit.

What are you each thinking about this?
- A) We should keep warnings as long as possible, except if we manage to measure a performance difference.
- B) It is fine to remove the "old" warnings, to keep the code base clean and avoid unnecessary clutter
- C) We should actively enforce a strict policy. This means actively removing any warning that is considered "old".

Let us have this discussion once and for all to avoid bringing it back every month ;)

khaneliman:
I'm more of a B/C person. I don't like having too much tech debt crowding the source code. Some of it was a pain to develop and implement, but it was done out of kindness to users to cause the least disruptive changes to get to our end state. I don't want to keep it around forever, though. I'm hesitant to say "strict policy", though. Personally, I'd say anything that's been around for 2 releases is easy to just gut, if it's only been a single release we can be more opinionated about whether it can be removed or should stick around another release.

MattSturgeon:
Here's my proposal for a strict policy:

Any deprecation notice, alias, removal, etc that has had a warning/notice in place for at least **two** releases can be removed without significant discussion.

(Perhaps we would make an effort to remove **all** deprecation older than two or three releases? But this shouldn't be enshrined as policy imo)

If anything was changed with a silent alias one release, with a warning notice introduced **later**, we would start counting from when the notice was introduced.

Anything that has had a warning notice for **one*" release can be removed only if there is a clear reason to do so, after appropriate discussion. This should be an exceptional circumstance, e.g. a warning somehow conflicting with a new feature.

Any notice that has *"not yet*" made it into a stable release should only be removed if it is invalid. Usually this would mean replacing the notice with something more accurate.

MattSturgeon:
Exactly.

> Do you think we should enact it now, then?

I'm fine with that, assuming no objections from @khaneliman or @traxys. 

[...]

PR #3634, opened 2025-08-20
plugins/lspsaga: migrate to mkNeovimPlugin

saygo-png:
Tracking issue: https://github.com/nix-community/nixvim/issues/2638

Things worth paying attention to:
- I changed the boolean conditions for deprecations since the options they used no longer are a boolean by default.
- I added a `mkChangedModule` deprecation for `finder.filter` but I haven't tested it myself, because I don't understand the purpose of it in the plugin's upstream docs. It should work as before because it's just following the old behavior.
- The warning in `default.nix` mentioned `plugins.ltex-extra` which looked like a mistake to me, so I changed it to `plugins.lspsaga`

The code in `deprecations.nix` is ugly but it's hard to avoid with this many different deprecations.

Sorry for the big diff!


plugins/by-name/lspsaga/default.nix
```diff
@@ -1,641 +1,36 @@
 {
   lib,
   helpers,
```
MattSturgeon: FYI `helpers` is a deprecated alias for `lib.nixvim`

```suggestion
```

plugins/by-name/lspsaga/deprecations.nix
```diff
@@ -0,0 +1,205 @@
...
+                  config.plugins.mini.enable
+                  && config.plugins.mini.modules ? icons
+                  && config.plugins.mini.mockDevIcons
+                )
+              )
+              {
+                enable = lib.mkOverride 1490 true;
+              };
+        }
+      )
```
MattSturgeon: I believe it was a mistake that this isn't conditional on `cfg.enable`.

If so, this could be moved to `extraConfig` (if preferred), or `cfg.enable &&` could be added here.

Alternatively, we've probably had this in enough release versions that it could be dropped now (although our deprecation lifespan policy [discussion](https://github.com/nix-community/nixvim/issues/3181) is still unresolved)

plugins/by-name/lspsaga/deprecations.nix
```diff
@@ -0,0 +1,205 @@
...
+      [
+        "signs"
+        "headers"
+        "maxDialogWidth"
+        "icons"
+        "maxFinderPreviewLines"
+        "keys"
+        "borderStyle"
+        "renamePromptPrefix"
+      ]
```
MattSturgeon: I know our deprecation policy [discussion](https://github.com/nix-community/nixvim/issues/3181) is still open, however this was added August 2023 and has been in 23.11, 24.05, 24.11, and 25.05.

I don't think any maintainers would argue for keeping it at this point 🙂

plugins/by-name/lspsaga/deprecations.nix
MattSturgeon: Actually, it must've been conditional before as this is now causing CI to fail. ([e.g.](https://buildbot.nix-community.org/#/builders/2908/builds/5301))

MattSturgeon (commented):
I couldn't bring myself to review each settings option that is now a deprecated rename, congrats on having the motivation to tackle those yourself!

This seems good once CI is green again.

plugins/by-name/lspsaga/deprecations.nix
saygo-png: Keeping this deprecation since it's more recent until the discussion gets resolved.

PR #3635, opened 2025-08-21
user-configs: remove HeitorAugustoLN's configuration

HeitorAugustoLN:
Removes my configuration from configuration examples, since it doesn't use nixvim anymore

### DIFF
--- a/docs/user-configs/list.toml
+++ b/docs/user-configs/list.toml
@@ -57,11 +57,6 @@ owner = "gwg313"
 repo = "nvim-nix"
 
 [[config]]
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Thanks for updating here!

mnw looks like an interesting project. I've had plans to design something _similar_ for a while.

FYI you're still mentioning nixvim in some places, such as your repo's description.

HeitorAugustoLN:
> FYI you're still mentioning nixvim in some places, such as your repo's description.

Oh, didn't even notice it, thanks.

PR #3643, opened 2025-08-23
ci/update-maintainers: drop `issues:write` permission

MattSturgeon:
The [update-maintainers](https://github.com/nix-community/nixvim/actions/workflows/update-maintainers.yml) workflow creates a `nixvim-ci` app token with permissions: https://github.com/nix-community/nixvim/blob/ecc7880e00a2a735074243d8a664a931d73beace/.github/workflows/update-maintainers.yml#L42-L44

However this is throwing errors that some of these permissions are not present on the app installation:

> The permissions requested are not granted to this installation. - https://docs.github.com/rest/reference/apps#create-an-installation-access-token-for-an-app

E.g. on this run: https://github.com/nix-community/nixvim/actions/runs/17036431911

I'm guessing it's `issues: write` we're missing, as I believe the other perms are used successfully in other workflows?

We requested this permission to be able to create new labels, when labelling the PR. However that is not necessary here, because a) we don't apply any labels and b) if we did we could create them manually. https://github.com/nix-community/nixvim/blob/ecc7880e00a2a735074243d8a664a931d73beace/.github/workflows/update-maintainers.yml#L213-L215

@zowoq: am I correct that it is `issues:write` we're missing, and the others are present?

### DIFF
--- a/.github/workflows/update-maintainers.yml
+++ b/.github/workflows/update-maintainers.yml
@@ -24,11 +24,9 @@ jobs:
     # Permissions required for workflow
     # `contents`: to update maintainers file
     # `pull-requests`: to create pr
-    # `issues`: to label pr
     permissions:
       contents: write
       pull-requests: write
-      issues: write
     env:
       pr_branch: update/maintainers-${{ github.ref_name }}
     steps:
@@ -41,7 +39,6 @@ jobs:
           private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
           permission-contents: write
           permission-pull-requests: write
-          permission-issues: write
 
       - name: Get GitHub App user info
         id: user-info


zowoq:
> @zowoq: am I correct that it is issues:write we're missing, and the others are present?

Yes. Is this a permission that you wanted on the app that I overlooked?

MattSturgeon:
> Yes.

Thanks for checking!

> Is this a permission that you wanted on the app that I overlooked?

I'll wait and see how this PR is reviewed.

AFAIK we shouldn't need that permission currently, but we may need it in the future for other things. Or maybe there's something I'm missing that means we can't drop it from this workflow.

khaneliman:
Would like if we added the permission so we could leverage labeling. But, I don't think it's actually required in this workflow anymore. I think I originally had labels I was going to do, but removed them from this one.

khaneliman (approved):

zowoq:
> Would like if we added the permission so we could leverage labeling.

I've added `issues:write` to the app.

PR #3645, opened 2025-08-24
plugins/comfy-line-numbers: init

### LINKED ISSUE #3545: [PLUGIN REQUEST] Comfy Line Numbers
| Field    | Description |
|----------|-------------|
| Plugin   | `Comfy Line Numbers`  | 
| Homepage | `https://github.com/mluders/comfy-line-numbers.nvim`     | 
| Nixpkgs  | `false` | 

## Extra Information

This is a simple plugin making relative line changes easier by allowing you to use your left hand only for the numbers while your right hand handles the j and k key.

c4patino:
add support for [comfy-line-numbers.nvim](https://github.com/mluders/comfy-line-numbers.nvim)

Resolve #3545

### DIFF
--- a/plugins/by-name/comfy-line-numbers/default.nix
+++ b/plugins/by-name/comfy-line-numbers/default.nix
@@ -1,0 +1,43 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "comfy-line-numbers";
+  package = "comfy-line-numbers-nvim";
+
+  description = "A Neovim plugin that makes vertical motions more comfortable.";
+
+  maintainers = with lib.maintainers; [
+    c4patino
+  ];
+
+  settingsExample = lib.literalExpression ''
+    {
+      labels = [
+        "1" "2" "3" "4" "5"
+        "11" "12" "13" "14" "15"
+        "21" "22" "23" "24" "25"
+        "31" "32" "33" "34" "35"
+        "41" "42" "43" "44" "45"
+        "51" "52" "53" "54" "55"
+        "111" "112" "113" "114" "115"
+        "121" "122" "123" "124" "125"
+        "131" "132" "133" "134" "135"
+        "141" "142" "143" "144" "145"
+        "151" "152" "153" "154" "155"
+        "211" "212" "213" "214" "215"
+        "221" "222" "223" "224" "225"
+        "231" "232" "233" "234" "235"
+        "241" "242" "243" "244" "245"
+        "251" "252" "253" "254" "255"
+      ];
+
+      up_key = "k";
+      down_key = "j";
+
+      hidden_file_types = [ "undotree" ];
+      hidden_buffer_types = [
+        "terminal"
+        "nofile"
+      ];
+    }
+  '';
+}
--- a/tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
+++ b/tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
@@ -1,0 +1,47 @@
+{ lib, ... }:
+{
+  empty = {
+    plugins.comfy-line-numbers.enable = true;
+  };
+
+  defaults = {
+    plugins.comfy-line-numbers = {
+      enable = true;
+      settings = {
+        labels = lib.strings.splitString "," (
+          lib.concatStringsSep "," [
+            "1,2,3,4,5"
+            "11,12,13,14,15"
+            "21,22,22,22,25"
+            "31,32,33,34,35"
+            "41,42,43,44,45"
+            "51,52,53,54,55"
+            "111,112,113,114,115"
+            "121,122,123,124,125"
+            "131,132,133,134,135"
+            "141,142,143,144,145"
+            "151,152,153,154,155"
+            "211,212,213,214,215"
+            "221,222,223,224,225"
+            "231,232,233,234,235"
+            "241,242,243,244,245"
+            "251,252,253,254,255"
+          ]
+        );
+
+        up_key = "k";
+        down_key = "j";
+
+        hidden_file_types = [ "undotree" ];
+        hidden_buffer_types = [
+          "terminal"
+          "nofile"
+        ];
+
+      };
+
+    };
+
+  };
+
+}


plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,102 @@
...
+      "242"
+      "243"
+      "244"
+      "245"
+      "251"
+      "252"
+      "253"
+      "254"
+      "255"
+    ];
```
MattSturgeon: Is this a useful example?

If so, should we replace this `settingsExample` with one that uses some functions to generate the list? Otherwise I worry that the example will take up too much space and be too verbose.

We could do something like:

```nix
  {
    labels = builtins.genList (i: toString (i + 1)) 254;

    up_key = "k";
    down_key = "j";

    hidden_file_types = [ "undotree" ];
    hidden_buffer_types = [ "terminal" "nofile" ];
  }
'';
```

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,107 @@
...
+          "242"
+          "243"
+          "244"
+          "245"
+          "251"
+          "252"
+          "253"
+          "254"
+          "255"
+        ];
```
MattSturgeon: We can also use `genList` to ~~simplify the test~~ _make the test more readable_.

```suggestion
        labels = builtins.genList (x: toString (x + 1)) 254;
```

plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,102 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "comfy-line-numbers";
+  package = "comfy-line-numbers-nvim";
+
+  description = "A Neovim plugin that makes vertical motions more comfortable.";
+
+  maintainers = with lib.maintainers; [ c4patino ];
```
MattSturgeon: Personal preference nitpick (feel free to ignore and resolve with no action):

I try to format such lists so that future changes can be done with minimal diffs.

In this case, formatting it as:

```suggestion
  maintainers = with lib.maintainers; [
    c4patino
  ];
```

...would allow adding/removing/changing maintainers without touching the actual `maintainers =` line.

plugins/by-name/comfy-line-numbers/default.nix
c4patino: This is a useful example straight from their readme. I was trying to use genlist but its essentially all numbers in order that can be composed of 1,2,3,4,5, so it wasn't quite as simple as I would've liked. If it'll take too much space we can shorten it though.

MattSturgeon: Ah, I think the formatting threw me off. I didn't spot that it is only numbers 1-5 in each multiple of 10.

I think this may make sense to manually format (as a literal expression), such that each multiple of ten gets its own row?

```nix
[
  "1" "2" "3" "4" "5"
  "11" "12" "13" "14" "15"
  ...
]
```

---

Also, is it intended that they are strings, or would numbers also work?

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
MattSturgeon: Please disregard the above suggestion.

Instead, if we think there's value in working around nixfmt here, we could do something like:

- Use `splitString` to split up single-lines by some deliminator.
- Use `concatLists`, `flatten`, or `concatMap` to join multiple/nested lists.

plugins/by-name/comfy-line-numbers/default.nix
c4patino: Yeah fair enough. This was also the general consensus in nixpkgs i believe.

plugins/by-name/comfy-line-numbers/default.nix
c4patino: They have to be strings. The lua plugin throws an error if they are not. I tried it before. I actually had it formatted this way originally, but I'm not sure how to get around nixfmt for this.

MattSturgeon: > but I'm not sure how to get around nixfmt for this.

You can't. That's why I'm suggesting a `literalExpression` which is itself a string-representation of the example.

The content of a literal expression can be formatted however we like, or include example function calls, etc. Because it isn't _real_ nix code.

c4patino: Take a look at what I switched it to. I think that this is rather readable and does something similar to this, while still allowing us to do keep it as native nix?

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
c4patino: I think I've done this in response to your above comment, sorry didn't see this until I had already tried the above :).

plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,43 @@
...
+        "131,132,133,134,135"
+        "141,142,143,144,145"
+        "151,152,153,154,155"
+        "211,212,213,214,215"
+        "221,222,223,224,225"
+        "231,232,233,234,235"
+        "241,242,243,244,245"
+        "251,252,253,254,255"
+      ]
+    );
```
MattSturgeon: This kinda thing is useful in the test, where we need real code.

However, the example only exists to be rendered in the docs. Here we should use a literalExpression:

```nix
  {
    lines = [
      "1" "2" "3" "4" "5"
      # the literal expression string can contain anything,
      # and it'll be rendered as-is in the docs.
    ];
  }
'';
```

Without a literalExpression, this would still be rendered in the docs as the final evaluated value:
```nix
[
  "1"
  "2"
  "3"
  ...
]
```

plugins/by-name/comfy-line-numbers/default.nix
MattSturgeon: I've commented on the latest code, but just for the continuity of this thread:

In the example we don't really care too much about it being "real" code.

What'll happen here is the docs code will evaluate your example, and render the final value as-per nixfmt style.

I.e. in the docs it'll look exactly as it did before, one item per line.

The only way to supply custom formatting to docs values like `example` is a literalExpression.

plugins/by-name/comfy-line-numbers/default.nix
c4patino: Ah okay. That makes sense. Changed. Thanks for the help! I didn't know that we could do that with `literalExpression`, I've always avoided the function since it never seemed useful.

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,47 @@
+{ lib, ... }:
+{
+  empty = {
+    plugins.comfy-line-numbers.enable = true;
+  };
+
+  defaults = {
```
MattSturgeon, line 7: Q: is this test actually testing defaults, or is it testing an example?

plugins/by-name/comfy-line-numbers/default.nix
c4patino: Ah. Okay. Good to know. I was under the impression that the raw code itself was being rendered, since from what I've seen its always matched what was in the source file. So I figured if you create that list programmatically the code itself is what is rendered in the docs.

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
c4patino: Defaults. Checked in their init.lua for it just to make sure.

plugins/by-name/comfy-line-numbers/default.nix
```diff
@@ -0,0 +1,43 @@
...
+
+      up_key = "k";
+      down_key = "j";
+
+      hidden_file_types = [ "undotree" ];
+      hidden_buffer_types = [
+        "terminal"
+        "nofile"
+      ];
+    };
```
MattSturgeon:
```suggestion
    }
```

tests/test-sources/plugins/by-name/comfy-line-numbers/default.nix
c4patino: I suppose in a way that makes the test redundant since in theory `empty` should be performing the same tests, but from the other test files it seemed like this was the way it was done generally and would trigger a failure if they ever change their settings layout so that we can update the example as well.

MattSturgeon (approved):
Otherwise LGTM.

Sorry my reviews have been a bit piecemeal on this PR. Thanks for your persistence!

c4patino:
> Otherwise LGTM.
> 
> Sorry my reviews have been a bit piecemeal on this PR. Thanks for your persistence!

No you're all good! Appreciate the feedback. I've written my own derivation before but am only recently getting more involved, so I'm still learning 😁.

PR #3648, opened 2025-08-25
plugins/tardis: add tardis-nvim

FredeHoey:


lib/maintainers.nix
```diff
@@ -39,6 +39,12 @@
     name = "Daniel Laing";
     keys = [ { fingerprint = "0821 8B96 DC73 85E5 BB7C  A535 D264 3BD2 13BC 0FA8"; } ];
   };
+  FredeHoey = {
+    email = "frederikbraendstrup@gmail.com";
+    github = "FredeHoey";
+    githubId = 7551358;
+    name = "Frede Braendstrup";
+  };
```
MattSturgeon: Could you do this in a separate commit _before_ the commit that adds the new plugin?

You'll end up with:

- `maintainers: add FredeHoey`
- `plugins/tardis: init`

MattSturgeon: Actually, you can remove this entry altogether, because you're already a nixpkgs maintainer.

```suggestion
```

plugins/by-name/tardis/default.nix
```diff
@@ -0,0 +1,28 @@
...
+  moduleName = "tardis-nvim";
+  packPathName = "tardis.nvim";
+  package = "tardis-nvim";
+  description = ''
+    Timetravel for neovim
+  '';
+
+  maintainers = [ lib.maintainers.fredeb ];
+
+  settingsOptions = {
```
MattSturgeon: Generally, in recent code, we've not been adding new `settings` options.

That's not to say we have a hard rule against them, but usually we find they're more trouble than they're worth and make the plugin module harder to maintain.

Instead, we usually prefer having a `settingsExample` that gives an example of how to define `settings`. This could be copied from the `settings` value in your `example` test, for example.

tests/test-sources/plugins/by-name/tardis/default.nix
```diff
@@ -0,0 +1,35 @@
...
+
+  defaults = {
+    plugins.tardis = {
+      enable = true;
+
+      settings = {
+      };
+    };
+  };
+
```
MattSturgeon: This is identical to your `empty` test. I don't think it's needed.

Historically, we would use the `defaults` test to explicitly define the "plugin defaults" that are documented by the `settingsOptions`. If you're keeping your explicitly declared settings options, then you'd copy the defaults you define there to this test. The idea being that this test validates those defaults are valid values for this plugin.

However, if we aren't declaring any `settings` options, there is nothing to validate and this test isn't needed.

lib/maintainers.nix
FredeHoey: Ah, yes that makes sense

plugins/by-name/tardis/default.nix
FredeHoey: Okay, so just to make sure I understand, when a plugin has a `settings` field set in a users config, it that will be tranformed to a lua object that will be passed to the setup function for the plugin?

tests/test-sources/plugins/by-name/tardis/default.nix
FredeHoey: Makes sense. I copied this from a different plugin and didn't read it thoroughly enough. Sorry!

plugins/by-name/tardis/default.nix
MattSturgeon: Yes. `settings` gets serialized to lua, and passed to the plugin's `setup` function.

FredeHoey: Great!

plugins/by-name/tardis/default.nix
```diff
@@ -0,0 +1,30 @@
...
+        revision_message = "<C-m>";
+        commit = "<C-g>";
+      };
+      settings = {
+        initial_revisions = 10;
+        max_revisions = 256;
+        show_commit_index = false;
+      };
+    };
+  };
```
MattSturgeon: I'm a little confused. We seem to have three levels of `settings` here; there's a `settingsExample.settings.settings` attr which contains `initial_revisions`, `max_revisions`, etc.

`settingsExample` should represent what will be assigned to `settings`, not contain an example of _assigning_ `settings`.

The second, nested, `settings` looks correct though: that matches the upstream plugin option named `settings`, which should be passed as `setup({ settings = {}, })`.

```suggestion
  settingsExample = {
    keymap = {
      next = "<C-j>";
      prev = "<C-k>";
      quit = "q";
      revision_message = "<C-m>";
      commit = "<C-g>";
    };
    settings = {
      initial_revisions = 10;
      max_revisions = 256;
      show_commit_index = false;
    };
  };
```

FredeHoey: Totally correct. Don't know how I messed that up

FredeHoey:
@MattSturgeon sorry to ping, but I'm not sure this build failure is something I caused, can you verify?

MattSturgeon:
> I'm not sure this build failure is something I caused, can you verify?

Sorry for the delay. [Looks like](https://buildbot.nix-community.org/#/builders/4017/builds/837) the docs can't find the plugin's homepage:

> unable to get URL for `plugins.tardis`.

Most packages in nixpkgs specify `meta.homepage`, so [by default we use that](https://github.com/nix-community/nixvim/blob/b7e96214e8e7244eceae73c606dcd243f6d180a3/lib/plugins/utils.nix#L99-L103). But if not, we need to manually supply a [`url` argument](https://github.com/nix-community/nixvim/blob/b7e96214e8e7244eceae73c606dcd243f6d180a3/lib/plugins/mk-neovim-plugin.nix#L8) to `mkNeovimPlugin`. In this case, I think you want the following:

```nix
url = "https://github.com/FredeHoey/tardis.nvim";
```

EDIT: it looks like `meta.homepage` is [now present](https://github.com/NixOS/nixpkgs/blame/d7600c775f877cd87b4f5a831c28aa94137377aa/pkgs/applications/editors/vim/plugins/generated.nix#L14540) on the nixpkgs package. Rebasing after we bump the lockfile should resolve the issue.

FredeHoey:
@MattSturgeon Right, that makes sense. I just managed to push the change before your message, so please disregard the build pipeline

MattSturgeon:
Hm. On further investigation, I think something weird is broken on nixpkgs.

Your PR's branch has nixpkgs rev [`f937f8ecd1c70efd7e9f90ba13dfb400cf559de4`](https://github.com/NixOS/nixpkgs/blob/f937f8ecd1c70efd7e9f90ba13dfb400cf559de4/pkgs/applications/editors/vim/plugins/generated.nix#L14466-14477) pinned in its lockfile. However that revision _does_ define `meta.homepage` on the package.

Despite it being defined in `generated.nix`, it is _not_ present when actually evaluating that revision of nixpkgs:

```nix
$ gh pr checkout 3648

$ nix repl

nix-repl> :lf .
Added 31 variables.

nix-repl> inputs.nixpkgs.sourceInfo
{
  lastModified = 1755829505;
  lastModifiedDate = "20250822022505";
  narHash = "sha256-4/Jd+LkQ2ssw8luQVkqVs9spDBVE6h/u/hC/tzngsPo=";
  outPath = "/nix/store/bs1sz392mx4ig6l17a6baj8qv1w9pr4s-source";
  rev = "f937f8ecd1c70efd7e9f90ba13dfb400cf559de4";
  shortRev = "f937f8e";
}

nix-repl> inputs.nixpkgs.legacyPackages.x86_64-linux.vimPlugins.tardis-nvim.meta ? homepage
false

nix-repl> inputs.nixpkgs.legacyPackages.x86_64-linux.vimPlugins.tardis-nvim.meta
{
  available = true;
  broken = false;
  insecure = false;
  maintainers = [ ... ];
  maintainersPosition = { ... };
  name = "vimplugin-tardis.nvim-2025-07-28";
  outputsToInstall = [ ... ];
  position = "/nix/store/bs1sz392mx4ig6l17a6baj8qv1w9pr4s-source/pkgs/applications/editors/vim/plugins/generated.nix:14468";
  unfree = false;
  unsupported = false;
}
```

<details><summary>I also reproduced the issue on latest nixpkgs master.</summary>
<p>

```
$ git rev-parse HEAD
8fa7c7ad4c13dd85474498bf1b557ff7b4e0b93e

$ nix-instantiate --eval --strict --attr vimPlugins.tardis-nvim.meta
{ available = true; broken = false; insecure = false; maintainers = [ { email = "frederikbraendstrup@gmail.com"; github = "FredeHoey"; githubId = 7551358; name = "Frede Emil"; } ]; maintainersPosition = { column = 5; file = "/home/matt/nixpkgs/master/pkgs/applications/editors/vim/plugins/overrides.nix"; line = 3405; }; name = "vimplugin-tardis.nvim-2025-07-28"; outputsToInstall = [ "out" ]; position = "/home/matt/nixpkgs/master/pkgs/applications/editors/vim/plugins/generated.nix:14533"; unfree = false; unsupported = false; }

$ nix-instantiate --eval --strict --attr vimPlugins.tardis-nvim.meta.homepage
error: attribute 'homepage' in selection path 'vimPlugins.tardis-nvim.meta.homepage' not found
```

</p>
</details> 

I'm guessing there's some weird `//` shadowing going on somewhere... Any insight @GaetanLepage or @khaneliman?

---

To unblock this PR, it is fine to add an explicit `url`, as you pushed earlier. However, I'd prefer it be done with a `FIXME` or `TODO` comment noting that something is broken in nixpkgs.

MattSturgeon:
> Rebasing after we bump the lockfile should resolve the issue.

Sorry, I didn't mean to suggest you include commits from the update in your PR. My suggestion was rebasing _after that PR is merged_ may help.

However, as per https://github.com/nix-community/nixvim/pull/3648#issuecomment-3239438693 that is not the case. A lockfile update **will not help** until we fix whatever shadowing issue is present in nixpkgs.

To be clear, the issue is still present on nixpkgs `master`. I haven't bisected to see when the issue was introduced, yet.

---

For now, the only solution is to either wait for a fix in nixpkgs or explicitly define `url`.

khaneliman:
I see the issue. Creating a PR upstream shortly.

FredeHoey:
> > Rebasing after we bump the lockfile should resolve the issue.
> 
> Sorry, I didn't mean to suggest you include commits from the update in your PR. My suggestion was rebasing _after that PR is merged_ may help.
> 
> However, as per [#3648 (comment)](https://github.com/nix-community/nixvim/pull/3648#issuecomment-3239438693) that is not the case. A lockfile update **will not help** until we fix whatever shadowing issue is present in nixpkgs.
> 
> To be clear, the issue is still present on nixpkgs `master`. I haven't bisected to see when the issue was introduced, yet.
> 
> For now, the only solution is to either wait for a fix in nixpkgs or explicitly define `url`.

Sorry, just some shoddy git fat fingering. Should be fixed now

plugins/by-name/tardis/default.nix
```diff
@@ -0,0 +1,31 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "tardis";
+  moduleName = "tardis-nvim";
+  packPathName = "tardis.nvim";
+  package = "tardis-nvim";
+  description = ''
+    Timetravel for neovim
+  '';
+  # FIXME: This shouldn't be necessary as meta.homepage is defined in nixpkgs.
+  # see https://github.com/nix-community/nixvim/pull/3648#issuecomment-3239438693
```
MattSturgeon: Now we have a PR upstream, we can reference that instead.

```suggestion
  # TODO: Remove after https://github.com/NixOS/nixpkgs/pull/438707
```

MattSturgeon (approved):
Thanks for your efforts!

FredeHoey:
@MattSturgeon Thanks for your time and thorough review! Really appreciate it!

PR #3653, opened 2025-08-26
plugins/perfanno: add perfanno-nvim plugin

FredeHoey:
Adds https://github.com/t-troebst/perfanno.nvim

perfanno-nvim was just recently added to nixpkgs, so cannot be merged until flake.lock is updated

see https://github.com/NixOS/nixpkgs/pull/436798

### DIFF
--- a/plugins/by-name/perfanno/default.nix
+++ b/plugins/by-name/perfanno/default.nix
@@ -1,0 +1,41 @@
+{ lib, ... }:
+
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "perfanno";
+  packPathName = "perfanno.nvim";
+  package = "perfanno-nvim";
+  description = "NeoVim lua plugin that annotates source code with profiling information from perf, LuaJIT, or other profilers.";
+
+  maintainers = [ lib.maintainers.fredeb ];
+
+  settingsExample = {
+    line_highlights = lib.nixvim.nestedLiteralLua "require('perfanno.util').make_bg_highlights(nil, '#CC3300', 10)";
+    vt_highlight = lib.nixvim.nestedLiteralLua "require('perfanno.util').make_fg_highlight('#CC3300')";
+    formats = [
+      {
+        percent = true;
+        format = "%.2f%%";
+        minimum = 0.5;
+      }
+      {
+        percent = false;
+        format = "%d";
+        minimum = 1;
+      }
+    ];
+
+    annotate_after_load = true;
+    annotate_on_open = true;
+    telescope = {
+      enabled = lib.nixvim.nestedLiteralLua "pcall(require, 'telescope')";
+      annotate = true;
+    };
+
+    ts_function_patterns = {
+      default = [
+        "function"
+        "method"
+      ];
+    };
+  };
+}
--- a/tests/test-sources/plugins/by-name/perfanno/default.nix
+++ b/tests/test-sources/plugins/by-name/perfanno/default.nix
@@ -1,0 +1,49 @@
+{
+  empty =
+    { pkgs, ... }:
+    {
+      plugins.perfanno.enable = !pkgs.stdenv.isDarwin;
+    };
+
+  example =
+    { pkgs, ... }:
+    {
+      plugins.perfanno = {
+        enable = !pkgs.stdenv.isDarwin;
+
+        settings = {
+          line_highlights.__raw = ''require("perfanno.util").make_bg_highlights(nil, "#CC3300", 10)'';
+          vt_highlight.__raw = ''require("perfanno.util").make_fg_highlight("#CC3300")'';
+          formats = [
+            {
+              percent = true;
+              format = "%.2f%%";
+              minimum = 0.5;
+            }
+            {
+              percent = false;
+              format = "%d";
+              minimum = 1;
+            }
+          ];
+
+          annotate_after_load = true;
+          annotate_on_open = true;
+          telescope = {
+            enabled.__raw = ''pcall(require, "telescope")'';
+            annotate = true;
+          };
+
+          ts_function_patterns = {
+            default = [
+              "function"
+              "method"
+            ];
+            weirdlang = [
+              "weirdfunc"
+            ];
+          };
+        };
+      };
+    };
+}


MattSturgeon:
> ` error: vimPlugins.perfanno-nvim cannot be found in pkgs `

Looks like the package isn't in the currently pinned nixpkgs revision. I _can_ see it on [nixos search](https://search.nixos.org/packages?channel=unstable&show=vimPlugins.perfanno-nvim), so you may need to wait for us to update the lockfile.

FredeHoey:
@MattSturgeon Yes I upstreamed the package :) I just put this here for when upstream is updated. When do you reckon that will happen?

MattSturgeon:
> When do you reckon that will happen?

https://github.com/nix-community/nixvim/pull/3647 currently has several issues that need fixing or suppressing before we can update, so "when" is currently unclear 😢

FredeHoey:
@MattSturgeon Totally fine :) LMK if there's something I can do

MattSturgeon:
> LMK if there's something I can do

We "just" need someone to go through each failure and disable the relevant failing tests. Or if disabling our tests isn't appropriate, we may need to fix the issues uncovered by our CI upstream in nixpkgs.

This isn't always fun or easy, especially when using buildbot and needing to navigate from _cached_ build failures back to the original build that has the error logs...

Sometimes local building/evaluating/debugging is also needed.

I'm sure we'll get to it eventually. You can see @GaetanLepage has already made a start with [`plugins/project-nvim: fix moduleName`](https://github.com/nix-community/nixvim/pull/3647/commits/ec99be9bb18f2c73c45ed659438dd12e4f746e9f).

GaetanLepage:
I rebased on `main`.

tests/test-sources/plugins/by-name/perfanno/default.nix
```diff
@@ -0,0 +1,49 @@
...
+
+  example =
+    { pkgs, ... }:
+    {
+      plugins.perfanno = {
+        enable = !pkgs.stdenv.isDarwin;
+
+        settings = {
+          line_highlights.__raw = ''util.make_bg_highlights(nil, "#CC3300", 10)'';
+          vt_highlight.__raw = ''util.make_fg_highlight("#CC3300")'';
```
MattSturgeon: > example> ERROR: Error detected while processing /nix/store/v6ydz5x55nvr6ipplapbcwq4gil7kjy0-init.lua:
> example> E5113: Error while calling lua chunk: /nix/store/v6ydz5x55nvr6ipplapbcwq4gil7kjy0-init.lua:13: attempt to index global 'util' (a nil value)
> example> stack traceback:
> example>        /nix/store/v6ydz5x55nvr6ipplapbcwq4gil7kjy0-init.lua:13: in main chunk
> -- https://buildbot.nix-community.org/#/builders/4425/builds/442

Upstream has `local util = require("perfanno.util")` higher up in their example, so here we want:

```suggestion
          line_highlights.__raw = ''require("perfanno.util").make_bg_highlights(nil, "#CC3300", 10)'';
          vt_highlight.__raw = ''require("perfanno.util").make_fg_highlight("#CC3300")'';
```

---

Or alternatively, outside the test in the main plugin, we could add something like:

```nix
extraConfig = {
  # TODO: should we also `lib.mkBefore` or `lib.mkOrder` this definition?
  plugins.perfanno.luaConfig.pre = ''
    local util = require("perfanno.util")
  '';
};
```

This would add `util` to the scope for anyone adding code to the plugin's `luaConfig`, which could be helpful _or_ surprising 😅

plugins/by-name/perfanno/default.nix
```diff
@@ -0,0 +1,43 @@
+{ lib, ... }:
+
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "perfanno";
+  packPathName = "perfanno.nvim";
+  package = "perfanno-nvim";
+  description = "NeoVim lua plugin that annotates source code with profiling information from perf, LuaJIT, or other profilers.";
+  # TODO: Remove after https://github.com/NixOS/nixpkgs/pull/438707
+  url = "https://github.com/t-troebst/perfanno.nvim";
```
MattSturgeon: https://github.com/NixOS/nixpkgs/pull/438707 is in all unstable channels now, and _should_ be part of the revision we've updated to.

It should be safe to remove the explicit URL now:

```suggestion
```

_Note to self: should we throw when `url` is explicitly defined the same as the package's `meta.homepage`?_

plugins/by-name/perfanno/default.nix
```diff
@@ -0,0 +1,43 @@
...
+
+  maintainers = [ lib.maintainers.fredeb ];
+
+  settingsExample = {
+    line_highlights.__raw = ''util.make_bg_highlights(nil, "#CC3300", 10)'';
+    vt_highlight.__raw = ''util.make_fg_highlight("#CC3300")'';
```
MattSturgeon: As per the test code, this example may confuse users. Either we want to add `util` to scope so the example makes sense, or we want to show explicit usage of `require` in the example instead.

---

Also: when we pass an actual attrset to `settingsExample` (or any option example), the docs will render it using `lib.generators.toPretty`. This means things like `foo.__raw = ...;` will be converted to:

```
foo = {
  __raw = ...;
}
```

If that isn't what we want, the main workaround is to assign a `literalExpression` to the example instead. In this case, though, we also have some more specific helpers for embedding "raw lua" in an option example: [`lib.nixvim.literalLua`](https://github.com/search?q=repo%3Anix-community%2Fnixvim%20literalLua&type=code).

Note this only applies to the example used in the docs, not the example we _run_ in the tests.

tests/test-sources/plugins/by-name/perfanno/default.nix
FredeHoey: Hmm, preloading seems a little too "black-magicky" to me. I would.expect to be required to import any module I need myself.
I'll have a look at an elegant way to do this later this week

plugins/by-name/perfanno/default.nix
FredeHoey: I think that an assertion would be great. I almost feel like a build should assert if both define a url, regardless of the value, because and then assume the upstream is always correct

plugins/by-name/perfanno/default.nix
FredeHoey: Thanks for the heads up! I'll check this later this week too

FredeHoey:
@MattSturgeon I added the explicit require. Do you still think I should use `literalLua` for the example?

MattSturgeon:
Testing locally with `nix run .#docs`:

(or without cloning: `nix run github:nix-community/nixvim?ref=pull/3653/merge#docs`)

image

Personally, I'd prefer using `literalExpression` or `nestedLiteralLua`, but this isn't a blocker.

FredeHoey:
Oh wow, yes that looks awful 😅

I'll fix it up

plugins/by-name/perfanno/default.nix
```diff
@@ -0,0 +1,41 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "perfanno";
+  packPathName = "perfanno.nvim";
+  package = "perfanno-nvim";
+  description = "NeoVim lua plugin that annotates source code with profiling information from perf, LuaJIT, or other profilers.";
+
+  maintainers = [ lib.maintainers.fredeb ];
+
+  settingsExample = {
+    line_highlights = lib.nixvim.literalLua ''require("perfanno.util").make_bg_highlights(nil, "#CC3300", 10)'';
```
MattSturgeon: Sadly, using the wrong one doesn't cause a CI failure, but I believe we need `nestedLiteralLua` here instead of `literalLua`.

You may also wish to switch to using single quotes in your lua, as I don't think it is smart enough to render a `''` string instead of `"`, unless the string includes newlines.

Or if you want _full_ control, you can render the entire example yourself manually using a `literalExpression`.

You can confirm how it renders locally, using `nix run .#docs` (which may be a _long_ build on the initial run).

FredeHoey: Fair. I did kick off the docs build, but it's still going, so I'm still waiting for the result. I'm at 7650/8389 now, so I'll wait for it to finish and validate

MattSturgeon: Yes, it's a shame our docs take so long. I've put effort in the past into re-implementing the docs more efficiently, but sadly none of those efforts ever made it over the line.

FredeHoey:
@MattSturgeon take 92, how we doin? 😅

PR #3657, opened 2025-08-31
Added nicklundin08/dots user config

nicklundin08:
Hello! 

Thank you for taking the time to put maintain this project. Ive finally cleaned up my dotfiles enough to feel comfortable sharing them and would love to be a part of the example configuration


MattSturgeon (approved):
Usually we'd want the commit message formatted as `scope: change`, e.g. `user-configs: add nicklundin08`.

PR #3660, opened 2025-09-03
treewide: Disable fixup phase for builds

saygo-png:
This implements the change suggested in https://github.com/nix-community/nixvim/issues/3518

It skips the fixup phase on our nixvim deriviations which almost halves build times.
I've been running this setup myself for a week or so, and did not spot any issues related to it.

This could also be implemented as an overlay instead of overriding on each call.

### DIFF
--- a/ci/nvim-lspconfig/default.nix
+++ b/ci/nvim-lspconfig/default.nix
@@ -19,7 +19,10 @@ let
     ];
   };
 
-  nvim = wrapNeovimUnstable neovim-unwrapped nvimConfig;
+  nvim = (wrapNeovimUnstable neovim-unwrapped nvimConfig).overrideAttrs {
+    dontFixup = true;
+  };
+
 in
 runCommand "lspconfig-servers"
   {
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -220,7 +220,13 @@ in
         }
       );
 
-      wrappedNeovim = pkgs.wrapNeovimUnstable package neovimConfig;
+      # TODO: 2025-01-06
+      # Wait for user feedback on disabling the fixup phase.
+      # Ideally this will be upstreamed to nixpkgs.
+      # See https://github.com/nix-community/nixvim/pull/3660#discussion_r2326250439
+      wrappedNeovim = (pkgs.wrapNeovimUnstable package neovimConfig).overrideAttrs {
+        dontFixup = true;
+      };
 
       customRC = helpers.concatNonEmptyLines [
         (helpers.wrapVimscriptForLua wrappedNeovim.initRc)


ci/nvim-lspconfig/default.nix
```diff
@@ -19,7 +19,10 @@ let
     ];
   };
 
-  nvim = wrapNeovimUnstable neovim-unwrapped nvimConfig;
+  nvim = (wrapNeovimUnstable neovim-unwrapped nvimConfig).overrideAttrs (_: {
+    dontFixup = true;
+  });
```
MattSturgeon: `overrideAttrs` can receive _either_ a function _or_ a set. No need to pass a function and ignore its input.

```suggestion
  nvim = (wrapNeovimUnstable neovim-unwrapped nvimConfig).overrideAttrs {
    dontFixup = true;
  };
```

modules/top-level/output.nix
```diff
@@ -220,7 +220,9 @@ in
         }
       );
 
-      wrappedNeovim = pkgs.wrapNeovimUnstable package neovimConfig;
+      wrappedNeovim = (pkgs.wrapNeovimUnstable package neovimConfig).overrideAttrs (_: {
+        dontFixup = true;
+      });
```
MattSturgeon: This will affect users, not just our CI.

Is there a scenario where the fixup phase may be needed on this wrapper?

Of course, users could also override the package when they install it to set `dontFixup` back to false, but would it make sense to have a nixvim option for whether we will disable/enable the fixup phase in the wrapper drv?

If the fixup phase is never useful for this wrapper, it begs the question why don't we disable it in nixpkgs (or at least, disable it by default there).

MattSturgeon (commented):
Thanks for taking the time to implement this!

My preference would be to make this change upstream in nixpkgs, instead. However we are in urgent need of improving CI build times here in nixvim, so we could merge this as an interim solution.

Or maybe there's some reason I don't know of why this doesn't make sense to upstream into nixpkgs? cc @teto

modules/top-level/output.nix
khaneliman: I agree, it's a little weird to crate overrides in here unconditionally if it's really not useful for the wrapper.

saygo-png: > This will affect users, not just our CI.

That was my intention.

> Is there a scenario where the fixup phase may be needed on this wrapper?

I don't think so. Especially considering that this change passed all of our tests. This is the manual entry on this phase. I don't think any of the points are applicable to a wrapper like this. <https://nixos.org/manual/nixpkgs/stable/#ssec-fixup-phase>

Because of this I think this should stay, we can revert this commit once we upstream the change to nixpkgs (if this phase really is unnecessary). I'll remove this part if you disagree though. After all there might be breakage that our tests don't catch and my sample of 1 testing didn't either.

khaneliman: I guess, maybe it does make sense to pilot test it in nixvim's smaller user base than all of nixpkgs neovim users?

saygo-png:
Looking at mnw, they also skip the fixup phase. https://github.com/Gerg-L/mnw/blob/ed6cc3e48557ba18266e598a5ebb6602499ada16/wrapper.nix#L301

This is the wrapper used by [nvf](https://github.com/NotAShelf/nvf/blob/main/flake.nix)

modules/top-level/output.nix
MattSturgeon: SGTM.

In that case, should we have a `TODO: assess and upstream (or revert)` style comment, maybe with a link to this discussion?

E.g.:
```nix
# TODO: 2025-01-06
# Wait for user feedback on disabling the fixup phase.
# Ideally this will be upstreamed to nixpkgs.
# See https://github.com/nix-community/nixvim/pull/3660#discussion_r2326250439
```

MattSturgeon:
> Looking at mnw

Off topic: I'd be open to a PR that added an experimental option for choosing between the nixpkgs wrapper and mnw.

...Assuming we could find a way to do that without the code being _too_ messy, repetitive, with too many special-cases, etc.

We'd also need to think about whether we wanted mnw as a flake input. Or if we should fetch it some other way (which would likely mean IFD).

Edit: tracking as https://github.com/nix-community/nixvim/issues/3663

ci/nvim-lspconfig/default.nix
```diff
@@ -19,7 +19,14 @@ let
     ];
   };
 
-  nvim = wrapNeovimUnstable neovim-unwrapped nvimConfig;
+  # TODO: 2025-01-06
+  # Wait for user feedback on disabling the fixup phase.
+  # Ideally this will be upstreamed to nixpkgs.
+  # See https://github.com/nix-community/nixvim/pull/3660#discussion_r2326250439
```
MattSturgeon:
```suggestion
```


Sorry I wasn't clear earlier. Because this is in the `ci/` folder, we shouldn't need to consider end users here. These are scripts to be run internally, usually by GHA.

MattSturgeon (approved):

MattSturgeon:
Thanks!

PR #3661, opened 2025-09-04
plugins/rhubarb: init

santoshxshrestha:
Add rhubarb plugin which provides GitHub extension for fugitive.vim.

### DIFF
--- a/plugins/by-name/rhubarb/default.nix
+++ b/plugins/by-name/rhubarb/default.nix
@@ -1,0 +1,19 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "rhubarb";
+  packPathName = "vim-rhubarb";
+  package = "vim-fugitive";
+  description = "Rhubarb is a GitHub extension for fugitive.vim.";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  dependencies = [ "git" ];
+
+  imports = [
+    (lib.nixvim.mkRemovedPackageOptionModule {
+      plugin = "rhubarb";
+      packageName = "git";
+    })
+  ];
+
+}


plugins/by-name/rhubarb/default.nix
```diff
@@ -0,0 +1,22 @@
...
+    "fugitive"
+  ];
+
+  imports = [
+    (lib.nixvim.mkRemovedPackageOptionModule {
+      plugin = "rhubarb";
+      packageName = "git";
+    })
+  ];
+
```
MattSturgeon:
```suggestion
```

This isn't needed here, in a new plugin. That's for deprecating old options that we used to have on some old plugins.

plugins/by-name/rhubarb/default.nix
```diff
@@ -0,0 +1,22 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "rhubarb";
+  packPathName = "vim-rhubarb";
+  package = "vim-rhubarb";
+  description = "Rhubarb is a GitHub extension for fugitive.vim.";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
```
MattSturgeon: Would you be willing to become the plugin's maintainer?

This mostly just means you'd get pinged in PRs touching the plugin.

plugins/by-name/rhubarb/default.nix
```diff
@@ -0,0 +1,22 @@
...
+  ];
+
+  imports = [
+    (lib.nixvim.mkRemovedPackageOptionModule {
+      plugin = "rhubarb";
+      packageName = "git";
+    })
+  ];
+
+}
```
MattSturgeon: We need to create a test file in `tests/test-sources`.

There are hundreds of similar plugins whose tests you can look at for inspiration, but we only really need an "empty" test and maybe an "example" test.

MattSturgeon (changes requested):
Thanks for your PR!

With a few small changes I think we can get this merged. Feel free to reach out here or on matrix if you need more detailed help. There may also be examples in recently merged PRs or in our contributing docs.

Once you've made changes, could you squash commits into atomic changes? For example, if you add yourself to the maintainers file that'd be one commit, and adding the plugin would be another, but we wouldn't expect separate commits fixing things in commits that are part of the same PR.

Also a tip: if you fetch and rebase onto the latest main branch, CI will run much faster, as we've recently made some improvements.

plugins/by-name/rhubarb/default.nix
vimlinuz: yeah sure I would love to .
 so i just have to put the name there is it .

MattSturgeon: First, in a separate commit, you add yourself to https://github.com/nix-community/nixvim/blob/main/lib/maintainers.nix

Then, in the "plugins" commit, you reference the maintainer attribute you added instead of Gaétan's.

plugins/by-name/rhubarb/default.nix
vimlinuz: I did similar to that one of the fugitive

plugins/by-name/rhubarb/default.nix
```diff
@@ -0,0 +1,14 @@
...
+  name = "rhubarb";
+  packPathName = "vim-rhubarb";
+  package = "vim-rhubarb";
+  description = "Rhubarb is a GitHub extension for fugitive.vim.";
+
+  maintainers = [ lib.maintainers.santoshxshrestha ];
+
+  dependencies = [
+    "git"
+    "fugitive"
```
MattSturgeon: > error: The option `dependencies.fugitive` does not exist

If we need this dependency we need to add it to the [`__depPackages`](https://github.com/nix-community/nixvim/blob/4933397031f2add4ecbb05eaa19accf9750b0e5f/modules/dependencies.nix#L91) definition. However, the dependencies option is really for binary/exe dependencies, to install onto the PATH; IIUC fugitive is a vim plugin, so it doesn't make sense here.

@nix-community/nixvim-committers how do we usually handle dependencies between plugins now? Do we add assertions that the other plugin is enabled? Do we `mkDefault` enabling the other plugin? Do we mention it in this plugin's `description`?

IIRC, what we want is to add an assertion to your plugin's `extraConfig` section:

```nix
extraConfig = {
  assertions = lib.nixvim.mkAssertions "plugins.rhubarb" [
    {
      assertion = config.plugins.fugitive.enable;
      message = "You must enable `plugins.fugitive` when using `rhubarb`.";
    }
  ];
};
```

tests/test-sources/plugins/by-name/rhubarb/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  empty = {
+    plugins.rhubarb.enable = true;
```
MattSturgeon: If you add the assertion mentioned above, you'll also need to satisfy it here by enabling `plugins.fugitive`.

lib/maintainers.nix
```diff
@@ -138,6 +138,11 @@
     githubId = 131687037;
     name = "Saygo";
   };
+  santoshxshrestha = {
+    github = "santoshxshrestha";
+    githubId = 182977126;
+    name = "Santosh Shrestha";
+  };
```
MattSturgeon: Awesome, thanks for adding your maintainers entry.

I can confirm the values are correct:

```
$ gh api /user/182977126 --jq .login
santoshxshrestha
$ gh api /user/182977126 --jq .name
Santosh Shrestha
```

---

When you get chance, could you use an "interactive rebase" to squash and re-order your 4 commits into 2?

Currently you have:
```
aea5419f: plugins/rhubarb: address review feedback
f32c4da1: maintainers: add santoshxshrestha
484999c7: plugins/rhubarb: fix package name and add fugitive dependency
766eccef: plugins/rhubarb: init
```

And we want to pick and squash as follows:
```
$ git fetch upstream
$ git rebase -i upstream/HEAD
```

```
pick f32c4da1 # maintainers: add santoshxshrestha
pick 766eccef # plugins/rhubarb: init
fixup 484999c7 # plugins/rhubarb: fix package name and add fugitive dependency
fixup aea5419f # plugins/rhubarb: address review feedback
```

Going forward, you can then use `git commit --amend` to re-create your most recent commit with new amendments. Or re-run `git rebase -i` to do squashes and fixups interactively.

After rebasing or amending, you'll need to "force" push to overwrite history. Typically, I'd recommend using:
```
git push --force-with-lease
```

See: [7.6 Git Tools - Rewriting History](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

plugins/by-name/rhubarb/default.nix
MattSturgeon: I see you've added the assertion, which is instead of having a "dependency". IIUC, you should remove this "fugitive" dependency and drop vim-fugitive from the __depPackages set.

vimlinuz: so just assertion will do the work ?

MattSturgeon: The dependencies system doesn't support creating a dependency on a vim plugin. It only supports creating dependencies on binaries that can be run from the CLI, like `git`.

Therefore, the assertion is the only option; asking users to also enable the plugin module.

Maybe "only option" is a lie: your plugin could also enable the other plugin, but _usually_ we avoid this in favour of explicit user definitions.

vimlinuz: got it, so here in this case git stays

modules/dependencies.nix
```diff
@@ -149,6 +149,7 @@ in
       util-linux.default = "util-linux";
       websocat.default = "websocat";
       wezterm.default = "wezterm";
+      fugitive.default = "vimPlugins";
```
MattSturgeon:
```suggestion
```

MattSturgeon:
I've unresolved https://github.com/nix-community/nixvim/pull/3661#discussion_r2337008916, because we still need to clean up the git history.

Are you familiar with using `git rebase -i`, force pushing, etc?

vimlinuz:
tried re basing but it made the thing worse

vimlinuz:
Done from my side

PR #3664, opened 2025-09-06
modules/output: rewrite nixvim-print-init

saygo-png:
Following https://github.com/nix-community/nixvim/issues/3518 i've tried to optimize our `nixvim-print-init` package for faster builds.

The problem with the current version is that it depends on the config, which means that it rebuilds on every change to the config. This PR finds the `init.lua` config during runtime which makes the package cache well in the nix store.

However there are 2 problems:
- The code is more error prone and ugly (though this could be improved a bit)
- I could not measure a performance difference...

I benchmarked it using `(time nix build) 2> >(tail -n1 >> timings.txt)` incrementing a tabstop value in the config each time while having `enablePrintInit` set to `True`
(`tail` is used here to remove the dirty tree warning)
```
# This PR nixvim-print-init
nix build  5,23s user 0,78s system 57% cpu 10,481 total
nix build  5,15s user 0,75s system 60% cpu 9,830 total
nix build  5,89s user 0,89s system 60% cpu 11,255 total
nix build  5,34s user 0,77s system 79% cpu 7,651 total
nix build  5,30s user 0,99s system 60% cpu 10,409 total

# Normal nixvim-print-init
nix build  5,22s user 0,71s system 58% cpu 10,124 total
nix build  5,24s user 0,75s system 79% cpu 7,495 total
nix build  5,36s user 0,66s system 59% cpu 10,065 total
nix build  5,46s user 0,76s system 60% cpu 10,338 total
nix build  5,32s user 0,74s system 59% cpu 10,200 total

# Disabled nixvim-print-init
nix build  5,26s user 0,72s system 60% cpu 9,880 total
nix build  5,11s user 0,76s system 58% cpu 9,953 total
nix build  5,30s user 0,82s system 59% cpu 10,296 total
nix build  5,35s user 0,73s system 60% cpu 10,118 total
nix build  6,00s user 0,87s system 60% cpu 11,397 total
```

However in the issue I mentioned earlier the author measures a 1s difference between building print-init and not building it, which I cannot measure. This makes me think that this optimization could still be worthwhile, I just can't see it on my hardware.

If someone could benchmark this PR, i'd appreciate it.
Especially @gepbird since you measured the 1s difference earlier.


saygo-png:
I don't understand those 3 failing tests, why don't they have the wrapped config in their PATH?

MattSturgeon:
> The problem with the current version is that it depends on the config, which means that it rebuilds on every change to the config.

Thanks for the proposal, but I don't understand why this is a problem...

The init printer is essentially a wrapper of the config, so the config itself is only built if it isn't already present in the nix store.

By the time you get to _running_ the printer, everything is already built and probably installed on your system.

If you want to print another config, maybe one you're working on that isn't installed to your system, then of course you need to build that config once, before it can be printed.

(Frankly, I don't think the init printer should be something we keep forever. At a certain point it may make more sense for users to directly inspect the in-store files themselves, which they could find by evaluating the `initSource` option. This is mostly a documentation issue. Ideally we'd also make it easier to find other files generated by nixvim, such as `after/` or `ftplugin/` files.)

saygo-png:
> Thanks for the proposal, but I don't understand why this is a problem...

Based on the issue I mentioned in the beginning: building the init printer takes around a second. If we could prevent that build on each config change, this would speed up builds by around a second, which is significant when trying to quickly adjust your config back and forth.

The issue isn't building the config, its building the printer, which has to be rebuilt each time (and checked with shellcheck).

That was my assumption but so far based on my benchmarks — and I managed to make another person benchmark this on their machine — this actually yields no visible performance gain. I think this PR is a dud so i'm tossing it.

I'm only confused as to why I don't see any gains from disabling printInit — this PR aside. Since gepbird seems have an entire second of difference.

khaneliman:
> (Frankly, I don't think the init printer should be something we keep forever. At a certain point it may make more sense for users to directly inspect the in-store files themselves, which they could find by evaluating the initSource option. This is mostly a documentation issue. Ideally we'd also make it easier to find other files generated by nixvim, such as after/ or ftplugin/ files.)

Personally, I use it all the time for quickly throwing it's output into a `init.lua` to tweak the config / debug stuff instead of just rebuilding over and over, when I need to trial and error things.

gepbird:
I ran the benchmarks as you did on my config, here are the results (`enableMan` is not disabled, as it is default, fixupPhase is disabled from your other PR 3 days ago, btw thanks for that!): 

```console
# This PR nixvim-print-init
nb dev  13.42s user 1.46s system 67% cpu 21.989 total
nb dev  13.49s user 1.59s system 67% cpu 22.388 total
nb dev  13.27s user 1.54s system 67% cpu 21.919 total

# Normal nixvim-print-init
nb dev  13.53s user 1.50s system 68% cpu 21.997 total
nb dev  13.21s user 1.42s system 67% cpu 21.606 total
nb dev  13.27s user 1.45s system 68% cpu 21.631 total

# Disabled nixvim-print-init
nb dev  13.43s user 1.48s system 68% cpu 21.770 total
nb dev  13.25s user 1.46s system 67% cpu 21.751 total
nb dev  12.99s user 1.69s system 67% cpu 21.621 total
```

IIRC when I tested it and got to the conclusion that it takes a second longer, I already disabled the man pages and the evaluation stole the time (probably because needing to evaluate an extra `bat` package). With the latest version of nixvim and the optimizations I mentioned, I cannot reproduce this 1 second delay, I assume some changes got rid of it, I'll investigate.

MattSturgeon:
> The issue isn't building the config, its building the printer, which has to be rebuilt each time (and checked with shellcheck).

Ah I see.

In that case the wrapper approach is still correct, but we _can_ consider how the wrapper itself can be built faster.

E.g. don't use `writeShellApplication` with its shellcheck phase... Since this is a one-line wrapper we don't need linting.

That said, we're probably talking micro seconds here?

gepbird:
On an old nixvim (ef0fa015a8236241bdcc27f32e6a4aa537d96cf8) I can reproduce it, this time only looking at the eval time. There's no difference on latest nixvim, however the eval time has increased:
```console
# enablePrintInit = false, old nixvim (ef0fa015a8236241bdcc27f32e6a4aa537d96cf8)
nix eval .#dev  4.21s user 0.57s system 81% cpu 5.889 total
nix eval .#dev  4.20s user 0.52s system 84% cpu 5.568 total
nix eval .#dev  4.21s user 0.53s system 84% cpu 5.607 total

# enablePrintInit = true, old nixvim (ef0fa015a8236241bdcc27f32e6a4aa537d96cf8)
nix eval .#dev  4.89s user 0.71s system 83% cpu 6.698 total
nix eval .#dev  4.99s user 0.70s system 83% cpu 6.801 total
nix eval .#dev  4.81s user 0.70s system 84% cpu 6.555 total


# enablePrintInit = false, latest nixvim
nix eval .#dev  5.85s user 0.85s system 82% cpu 8.172 total
nix eval .#dev  5.78s user 0.85s system 82% cpu 8.084 total
nix eval .#dev  5.85s user 0.82s system 81% cpu 8.141 total

# enablePrintInit = true, latest nixvim
nix eval .#dev  5.91s user 0.84s system 81% cpu 8.260 total
nix eval .#dev  5.84s user 0.83s system 82% cpu 8.114 total
nix eval .#dev  5.92s user 0.82s system 82% cpu 8.218 total
```

gepbird:
Apologies for rushing my previous comments: the main change between that 5.5s -> 6.5s and constant 8.1s eval was me enabling `plugins.lsp.servers.elixirls` in my config, which probably introduced the first more complex package that evaled `stdenv.mkDerivation`. Then it doesn't matter if I eval `bat` or not, it won't add much to the eval time, as Nix can reuse most of the things from evaling the `elixirls` package, hence the "constant" 8.1s.

By eval time I don't think this change is necessary, it only helps, if you have no LSPs or other "complex" packages. By build time I didn't notice a big difference, maybe 50-100ms (including differences in eval time which is probably different for you), I think the current implementation is fine for now, especially that you have plans to revamp/remove it in the future.

saygo-png:
Thank you so much for all the testing @gepbird!

Sadly looks like this was just a red herring. I don't think there is any point in changing this package, sorry for leading you guys astray with me.

PR #3665, opened 2025-09-06
tests: add all-package-defaults

GaetanLepage:
Motivated by a future removal of the generated tests on non-`x86_64-linux` platforms, let's add a test that ensures all of our package dependencies correctly build on all platforms.

- **tests: move pkgsForTests definition to tests/default.nix**
- **tests: add all-package-defaults**
- **tests/none-ls: typo**

### DIFF
--- a/tests/all-package-defaults.nix
+++ b/tests/all-package-defaults.nix
@@ -1,0 +1,62 @@
+{
+  lib,
+  stdenv,
+  pkgsForTest,
+  linkFarmFromDrvs,
+}:
+let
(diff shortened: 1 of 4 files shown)


tests/all-package-defaults.nix
```diff
@@ -0,0 +1,62 @@
+{
+  lib,
+  stdenv,
+  pkgsForTest,
+  linkFarmFromDrvs,
+}:
+let
```
MattSturgeon, line 7: Do we need a top-level comment just explaining why we have this test or what it is testing?

PR #3666, opened 2025-09-07
ci: build all-packages-defaults on all platforms

MattSturgeon:
This was missed in #3665, let's see if it actually builds on all platforms currently...


MattSturgeon:
@GaetanLepage I've resolved the eval failures, but we also have some build failures in transitive inputs. I'm struggling to identify the direct dependencies we need to disable.

PR #3667, opened 2025-09-07
ci: only build "grouped" tests on x86_64-linux

MattSturgeon:
Similar to 939a4428334e880eec7608c452a01b8c8b499aae by @zowoq, but with #3666 we will be more confident only running these tests on a single platform; as outlined by #3238.

As part of reducing these test cases to a single platform, I've also gone through and removed the now-redundant disablements.

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -93,5 +93,5 @@ in
 
   # TODO: consider whether all these tests are needed to be built by buildbot
   buildbot =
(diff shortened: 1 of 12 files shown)


GaetanLepage:
Looks great! Any reason for not marking it as ready?

MattSturgeon:
> Any reason for not marking it as ready?

My train arrived, so I shut my laptop 😅

GaetanLepage (approved):
LGTM!

Issue #3659, opened 2025-09-03
[BUG] :Telescope keymaps throws with faster-nvim and bytecompilation


saygo-png:
### Plugin

Telescope, faster-nvim

### Nixpkgs Release

unstable

### Home Manager Release

None

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

This might end up being duplicate of https://github.com/nix-community/nixvim/issues/3658. Depends on if this is a faster/conform-nvim issue or telescope issue.

Error:
```
Error executing Lua callback: ...ges/start/telescope.nvim/lua/telescope/actions/utils.lua:0: attempt to index a nil value
stack traceback:
        ...ges/start/telescope.nvim/lua/telescope/actions/utils.lua: in function ''
        ...ckages/start/telescope.nvim/lua/telescope/make_entry.lua: in function <...ckages/start/telescope.nvim/lua/telescope/make_entry.lua:0>
        ...scope.nvim/lua/telescope/finders/async_static_finder.lua: in function 'new_table'
        ...tart/telescope.nvim/lua/telescope/builtin/__internal.lua: in function ''
        ...tart/telescope.nvim/lua/telescope/builtin/__internal.lua: in function ''
        ...ages/start/telescope.nvim/lua/telescope/builtin/init.lua: in function <...ages/start/telescope.nvim/lua/telescope/builtin/init.lua:0>
        ...mPackages/start/telescope.nvim/lua/telescope/command.lua: in function ''
        ...mPackages/start/telescope.nvim/lua/telescope/command.lua: in function 'load_command'
        ...telescope.nvim-scm-1-unstable-scm-1/plugin/telescope.lua: in function <...telescope.nvim-scm-1-unstable-scm-1/plugin/telescope.lua:0>
```

Generated lua:
```lua
-- Ignore the user lua configuration
vim.opt.runtimepath:remove(vim.fn.stdpath("config")) -- ~/.config/nvim
vim.opt.runtimepath:remove(vim.fn.stdpath("config") .. "/after") -- ~/.config/nvim/after

-- Nixvim's internal module table
-- Can be used to share code throughout init.lua
local _M = {}

require("nvim-web-devicons").setup({})

require("telescope").setup({})

local __telescopeExtensions = {}
for i, extension in ipairs(__telescopeExtensions) do
    require("telescope").load_extension(extension)
end

require("faster").setup({})
```

[...]

PR #3670, opened 2025-09-08
modules/performance: add uncompiledPlugins option to plugin byte compilation

saygo-png:
This option is a sister option to `standalonePlugins`in `combinePlugins`. It allows the user to skip some plugins from byte compilation.

This is a needed change because some plugins (I know of at least faster and conform) break when byte compiled. With this option users don't have to choose between having byte compilation, having broken functionality or not using certain plugins. It allows for a compromise where only the problematic plugins are skipped, meanwhile the rest is still compiled.

I've tested that adding`faster.nvim` to the exception list (either as a package or name) fixes https://github.com/nix-community/nixvim/issues/3659

Things of note:
1. This breaks all existing configurations that set `performance.byteCompileLua.plugins` to a boolean.

     I could not figure out a way to deprecate this without an error, because this PR changes the old option to a set.
     The [error thrown by nixpkgs](https://github.com/NixOS/nixpkgs/blob/e8381f9fd66682d6618e6827207745e31100426c/lib/modules.nix#L775) is fairly good.
   * `mkRenamedOptionModule` throws infinite recursion
   * `mkRemovedOptionModule` throws because `.plugins` is not removed
   * `mkChangedOptionModule` throws because `.plugins` is not removed
   * an assertion in `assertions` is overridden by the nixpkgs assertion of the module system. It detects that the attrset is set to a boolean and throws a "type" error.

2. This PR does not include a test.

     The lua tests are broken due to a build failure for luajit. I couldn't feasibly write a new and correct test without the ability to run it and check its behavior.

### DIFF
--- a/modules/performance.nix
+++ b/modules/performance.nix
@@ -60,6 +60,17 @@ in
       luaLib = lib.mkEnableOption "luaLib" // {
         description = "Whether to byte compile lua library.";
       };
+      excludedPlugins = lib.mkOption {
+        type = with types; listOf (either str package);
+        default = [ ];
+        example = lib.literalExpression ''
+          [
+            "faster.nvim"
+             pkgs.vimPlugins.conform-nvim
+          ];
+        '';
+        description = "List of plugins (names or packages) to exclude from byte compilation.";
+      };
     };
 
     combinePlugins = {
--- a/modules/top-level/plugins/byte-compile-plugins.nix
+++ b/modules/top-level/plugins/byte-compile-plugins.nix
@@ -4,10 +4,20 @@
   Inputs: List of normalized plugins
   Outputs: List of normalized (compiled) plugins
 */
-{ lib, pkgs }:
+{
+  lib,
+  pkgs,
+  excludedPlugins,
+}:
+normalizedPlugins:
 let
   builders = lib.nixvim.builders.withPkgs pkgs;
   inherit (import ./utils.nix lib) mapNormalizedPlugins;
+
+  excludedNames = map (p: if builtins.isString p then p else lib.getName p) excludedPlugins;
+  isExcluded = p: builtins.elem (lib.getName p.plugin) excludedNames;
+
+  partitionedPlugins = builtins.partition isExcluded normalizedPlugins;
 
   byteCompile =
     p:
@@ -15,4 +25,4 @@ let
       prev: lib.optionalAttrs (prev ? dependencies) { dependencies = map byteCompile prev.dependencies; }
     );
 in
-mapNormalizedPlugins byteCompile
+(mapNormalizedPlugins byteCompile partitionedPlugins.wrong) ++ partitionedPlugins.right
--- a/modules/top-level/plugins/default.nix
+++ b/modules/top-level/plugins/default.nix
@@ -32,7 +32,10 @@ in
     build.plugins =
       let
         shouldCompilePlugins = byteCompileCfg.enable && byteCompileCfg.plugins;
-        byteCompilePlugins = pkgs.callPackage ./byte-compile-plugins.nix { inherit lib; };
+        byteCompilePlugins = pkgs.callPackage ./byte-compile-plugins.nix {
+          inherit lib;
+          inherit (config.performance.byteCompileLua) excludedPlugins;
+        };
 
         shouldCompileLuaLib = byteCompileCfg.enable && byteCompileCfg.luaLib;
         inherit (pkgs.callPackage ./byte-compile-lua-lib.nix { inherit lib; }) byteCompilePluginDeps;


MattSturgeon:
> 1. I could not figure out a way to deprecate this without an error, because this PR changes the old option to a set.


One way we could convert a boolean option into a set of options would be to actually make the old boolean a union of (`bool`, `submodule`).

Using a `submodule` for this is a little over-kill, but it's currently the easiest way to encapsulate a set of options within an option's type.

The union could be `oneOf`, `either`, `coercedTo`, or nixvim's custom fork of `coercedTo`: [`transitionType`](https://github.com/nix-community/nixvim/blob/d38eb947272dd4e6d138648b1b49360301db6859/lib/deprecation.nix#L91-L124) (yes, I still need to upstream something like it to nixpkgs lib).

In this case it'd look something like:

```nix
plugins = lib.mkOption {
  type =
    let
      oldType = lib.types.bool;
      newType = lib.submodule {
        options = {
          enable = lib.mkEnableOption "plugins" // {
            description = "Whether to byte compile lua plugins.";
          };
          uncompiledPlugins = lib.mkOption {
            type = with types; listOf (either str package);
            default = [ ];
            example = [ "faster.nvim" ];
            description = "List of plugins (names or packages) to exclude from byte compilation.";
          };
        };
      };
    in
    lib.nixvim.transitionType oldType (enable: { inherit enable; }) newType;
  # ...
  # NOTE: don't declare a default here, to avoid it being documented...
  # Instead, define `plugins = lib.mkOptionDefault {}` in an unconditional config section.
}
```

At some point we'd want to simplify the submodule to a plain set of options, which is technically a breaking change. It's breaking because people can assign any kind of module to a submodule-type option, including a function, file, compact, or verbose style module. Including imports, additional option declarations, etc.

We _could_ disallow that by replacing the type's `check` function with `lib.isAttrs`:

```
newType' = newType // {
  description = "attribute set";
  check = builtins.isAttrs;
};
```
...but the complexity is adding up just to cater to really niche deprecation scenarios. And it'd still allow `{ imports = []; }` and `{ options = {}; ` and `{ config = {}; }` (etc) style modules anyway; it'd only block files and functions.

Sadly, there isn't an easy alternative to submodules for encapsulating plain sets of options in a _type_, though. The closest proposal I'm aware of is [`lib.types.record`](https://github.com/NixOS/nixpkgs/pull/334680), but even that is a bit fancier than just a set of options...

modules/performance.nix
```diff
@@ -51,8 +51,16 @@ in
         default = true;
         example = false;
       };
-      plugins = lib.mkEnableOption "plugins" // {
-        description = "Whether to byte compile lua plugins.";
+      plugins = {
+        enable = lib.mkEnableOption "plugins" // {
+          description = "Whether to byte compile lua plugins.";
+        };
+        uncompiledPlugins = lib.mkOption {
```
MattSturgeon: Nit: `byteCompileLua.plugins.uncompiledPlugins` feels redundant; "plugins" is in the option path twice.

Could we choose a name that leverages the context of its option path? E.g. `byteCompileLua.plugins.exclude` or something?

modules/top-level/plugins/byte-compile-plugins.nix
```diff
@@ -4,15 +4,25 @@
...
+  lib,
+  pkgs,
+  uncompiledPlugins,
+}:
+normalizedPlugins:
 let
   builders = lib.nixvim.builders.withPkgs pkgs;
   inherit (import ./utils.nix lib) mapNormalizedPlugins;
 
+  uncompiledNames = map (p: if builtins.isString p then p else lib.getName p) uncompiledPlugins;
```
MattSturgeon: In this context "uncompiled" is ambiguous: it could mean plugins that have not (yet) been compiled, but the intended meaning is plugins that _should_ not be compiled.

Similar to the option name, I think "excluded plugins" could work, what do you think?

modules/top-level/plugins/byte-compile-plugins.nix
```diff
@@ -4,15 +4,25 @@
...
+  isUncompiled = p: builtins.elem (lib.getName p.plugin) uncompiledNames;
```
MattSturgeon: Again, same nitpick. This isn't about whether it _is_ uncompiled, but whether it _should_ be.

I'm also cautious of negative boolean/predicate names. Things usually flow better and make more sense with positive names, though there's no hard-and-fast rule here.

Also, one entire unrelated nitpick: it may be necessary in some cases to compare by equality instead of name. E.g. we could take the plugin's `outPath` (its in-store path) or even directly use `==` on plugins.

How about:

```nix
shouldCompile =
  plugin:
  let
    name = lib.getName plugin;
  in
  lib.none (
    excluded:
    if lib.isString excluded then
      name == excluded
    else
      plugin == excluded
      # or: plugin.outPath == excluded.outPath
  ) excludedPlugins;
```

?

MattSturgeon (commented):
Interesting concept. Thanks again for another great contribution!

I haven't considered what use-cases this has (maybe @stasjok will have an opinion?).

In the meantime, here's some initial feedback on the implementation:

stasjok:
> I haven't considered what use-cases this has (maybe @stasjok will have an opinion?).

I don't have objections to the ability to exclude plugins from compilation. Though in this particular case with telescope I think this is better to be fixed upstream. But who knows, this option might be useful for some niche cases for someone.

modules/top-level/plugins/byte-compile-plugins.nix
saygo-png: >  It may be necessary in some cases to compare by equality instead of name.

I don't want to change this because the `standalonePlugins` option does it this way. Changing would cause us to have 2 options which do the same thing yet have slightly different behavior

https://github.com/nix-community/nixvim/blob/d38eb947272dd4e6d138648b1b49360301db6859/modules/top-level/plugins/combine-plugins.nix#L35-L37

stasjok:
Maybe in order to preserve backward compatibility to leave `plugins` option as is and add another one like `excludedPlugins` or `pluginExceptions`? So it is defined like this:
```nix
      performance.byteCompileLua = {
        enable = true;
        plugins = true;
        excludedPlugins = [ ... ];
      };
```
I don't think this option will be used often. AFAIK byte compilation is pretty harmful, except when you need a plain text files or `debug.*` module.

There is also a `-g` flag to luajit that preserves debug info (if it's enabled, backtraces would point to original files and with correct line numbers). I didn't enable it in order to produce minimal files. But I though about leaving an option (it would be awkward because byte compilation helpers are in lib).

modules/top-level/plugins/byte-compile-plugins.nix
MattSturgeon: Good shout. Maybe we should consider improving both in another PR then.

stasjok: > Maybe we should consider improving both in another PR then.

I want to point out that I changed it to `getName` in #3421. Literal comparison should work in this PR (if you want) because byte-compilation is the first step, but it won't work in combinePlugins (because byte-compilation is changing derivations). After I changed it to name comparison there was no point to compare it also with `==`, because if it's the same derivation, name would be equal anyway.

MattSturgeon: Ah, I see. In that case #3421 shouldn't be reverted unless someone want's to explore one of the alternative solutions discussed there.

As for here, I could be convinced either way: on the one hand, having the ability to compare with `==` would be more powerful in very niche situations; on the other hand, having consistency would be more intuitive and less confusing.

saygo-png:
I've used the submodule approach, but i can't figure out a way of hiding the old option from docs. I have to include a description otherwise a warning is shown for a lack of a description. We have warnings to errors enabled on this so the build fails.

```
> error: option performance.byteCompileLua.plugins has no description
> Treating warnings as errors. Set documentation.nixos.options.warningsAreErrors to false to ignore these warnings.
```

Setting `internal = true` on `plugins` also hides the suboptions from docs. Setting `internal = false` on the suboptions does not help.

Is there a way around this or should I just use the approach suggested by Stasjok?

modules/performance.nix
```diff
@@ -51,9 +51,29 @@ in
         default = true;
         example = false;
       };
-      plugins = lib.mkEnableOption "plugins" // {
-        description = "Whether to byte compile lua plugins.";
+
+      plugins = lib.mkOption {
+        description = "TODO: This shows up in docs. it should not.";
```
MattSturgeon: Sadly this is a limitation of the current docs tooling.

The `visible` attribute can currently be `true`, `false`, or `"shallow"`. When set to shallow, sub-options are not documented.

I might be able to send a PR upstream to add a `visible = "transparent"` or something? That way the option itself could be ignored but its sub-options included.

modules/performance.nix
```diff
@@ -51,9 +51,29 @@ in
...
+                exclude = lib.mkOption {
+                  type = with types; listOf (either str package);
+                  default = [ ];
+                  example = [ "faster.nvim" ];
+                  description = "List of plugins (names or packages) to exclude from byte compilation.";
+                };
+              };
+            };
+          in
+          lib.nixvim.transitionType oldType (enable: { inherit enable; }) newType;
```
MattSturgeon: It's awesome to see this implemented (well done!), but I do wonder if it is worth its complexity? @stasjok 's suggestion of keeping the old `plugins` option as-is and adding a sibling `excludedPlugins` option is _much_ simpler to understand and maintain.

khaneliman: Yeah, I expected this to be a blacklist option added that we just filter things out from being compiled.

modules/performance.nix
MattSturgeon: Looking at the implementation details of `optionAttrSetToDocList`, we can use `internal = true` as a stand-in for a `visible = "transparent"`.

This is kinda hacky and relies on implementation details; but currently if an option is internal then it gets filtered out of the docs.

However only `visible` is checked when deciding whether to include sub-options, so an `internal = true; visible = true;` option will be hidden, but its sub-options will not be.

If we use this behaviour, we should add an explanatory comment, since the docs team could change `optionAttrSetToDocList` at any point to do additional filtering of visible/internal options; there's even talk of deprecating `internal` entirely, although that seems unlikely at this point.

modules/performance.nix
saygo-png: simpler it is then. using `excludePlugins` instead of suggested `excludedPlugins` for consistency with other options.

modules/performance.nix
```diff
@@ -60,6 +60,12 @@ in
       luaLib = lib.mkEnableOption "luaLib" // {
         description = "Whether to byte compile lua library.";
       };
+      excludePlugins = lib.mkOption {
```
MattSturgeon: Nit: "exclude plugins" sounds like a boolean option; whether to exclude plugins. As this is a list of plugins to be excluded, I think "excluded plugins" is more natural.

modules/performance.nix
```diff
@@ -60,6 +60,12 @@ in
...
+        type = with types; listOf (either str package);
+        default = [ ];
+        example = [ "faster.nvim" ];
```
MattSturgeon: Would a `literalExpression` example that demonstrates using a `pkgs.vimPlugins.*` package, alongside the string, be useful?

MattSturgeon (approved):
SGTM

PR #3673, opened 2025-09-09
Add option to allow disabling wrapping wl-clipboard

SuperSandro2000:
So that I do not have a wayland on my headless server.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -29,6 +29,13 @@ in
       '';
     };
 
+    waylandSupport = mkOption {
+      type = types.bool;
+      default = pkgs.stdenv.hostPlatform.isLinux;
+      defaultText = lib.literalExpression "pkgs.stdenv.hostPlatform.isLinux";
+      description = "Enable additional wayland support, such as wl-clipboard for wayland clipboard syncing.";
+    };
+
     withRuby = mkOption {
       type = types.bool;
       default = true;
@@ -199,6 +206,7 @@ in
             extraPython3Packages
             viAlias
             vimAlias
+            waylandSupport
             withRuby
             withNodeJs
             withPerl


modules/top-level/output.nix
```diff
@@ -29,6 +29,12 @@ in
       '';
     };
 
+    waylandSupport = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Whether to wrap wl-clipboard for wayland clipboard syncing support.";
```
MattSturgeon: The name `waylandSupport` sounds more general than clipboard-related stuff. I'd be worried that this description could easily get out of sync with what the option _actually_ does, e.g. if upstream nixpkgs ever adds additional functionality to its `waylandSupport` option.

We could be really vague and just say "enable additional wayland support", we could link to upstream documentation, or we could give wl-clipboard as an _example of_ something that gets additional wayland support when this option is used.

MattSturgeon (commented):
Thanks for syncing our options with new ones available in the underlying wrapper!

MattSturgeon:
Looks like this is also highlighting some issues in CI...

Some may be caused by having the option always default to `true` (instead of, say `hostPlatform.isLinux`).

Others may be caused by the [new `all-package-defaults` test](https://github.com/nix-community/nixvim/pull/3665) not filtering out option defaults from output options like [`build.package`](https://nix-community.github.io/nixvim/NeovimOptions/build/index.html#buildpackage); maybe filtering out readOnly and internal options would be enough for that?

modules/top-level/output.nix
SuperSandro2000: > we could link to upstream documentation

I don't think there is any but I'll add the rest.

MattSturgeon (approved):
Sorry, I forgot to check your commit message in my last review. Could you change it to match our usual scoped style (similar to nixpkgs)?

Looking at the [file's history](https://github.com/nix-community/nixvim/commits/af72663d4ffdcd6c15766346e6fa373ce6c1649e/modules/top-level/output.nix), it'd usually be `modules:` or `modules/output:`

How about ``modules/output: add `waylandSupport` option``?

Otherwise LGTM, thanks!

SuperSandro2000:
No problemo :)

n0099:
`wl-clipboard` is introduced by https://github.com/NixOS/nixpkgs/pull/439451

https://github.com/NixOS/nixpkgs/issues/498672
> * [ ] I dont see why we special case wayland with `lib.optionals finalAttrs.waylandSupport [ wl-clipboard ]` which adds more wrapping. Why `wl-clipboard` vs other solutions ? what if you are using X ? I would rather remove it as too opiniated.

PR #3676, opened 2025-09-10
plugins/harpoon: fix pack path name

khaneliman:
Lazy loading needs the correct name for finding the plugin in `opt`. With change to `harpoon2` as the default, we need to fix the `packPathName` to match.

### DIFF
--- a/plugins/by-name/harpoon/default.nix
+++ b/plugins/by-name/harpoon/default.nix
@@ -9,6 +9,7 @@ in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "harpoon";
   package = "harpoon2";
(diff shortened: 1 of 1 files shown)


MattSturgeon:
> Lazy loading needs the correct name for finding the plugin in opt

Kinda off topic, but is there a way we can cover lazy loading in our test suite? Or at least a way to verify names are correct?

khaneliman:
> Kinda off topic, but is there a way we can cover lazy loading in our test suite? Or at least a way to verify names are correct?

Would be nice to have something setup to verify. I think last time I talked about it, we could default `packPathName` to the default package `pname` since that should be the `packPathName` that is created from the wrapper.

PR #3681, opened 2025-09-12
docs/mdbook: general cleanup and renames

MattSturgeon:
Clean up the `docs/mdbook/default.nix` expression in an attempt to make it easier to understand.

I'm trying to understand how this expression works, in particular how it decides which options should go on a page vs a new sub-page. As part of my effort to understand the code, I've made a few renames and cleanup.

This PR should be close to zero rebuilds, with the exception of some of the "page summary" derivations now having a more descriptive drv name. These "summary" derivations encapsulate the "plugin description", maintainer list, URL, etc.

Eventually, I'd love to completely delete this file and replace it with something based on `optionAttrSetToDocList` (and using `groupBy` to group options into pages). But if we want to have redirects to prevent breaking existing links,[^1] then I'll need to understand the existing code well enough to produce an index of where specific options _used to be_.

[^1]: If we are happy to break the old URLs, then a rewrite would suddenly get **much** simpler...

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -84,7 +84,7 @@ let
       ) opt.declarations;
     };
 
-  evaledModules = helpers.modules.evalNixvim {
+  configuration = helpers.modules.evalNixvim {
     modules = [
       {
         isDocs = true;
@@ -95,7 +95,7 @@ let
 
   options-json =
     (pkgs.nixosOptionsDoc {
-      inherit (evaledModules) options;
+      inherit (configuration) options;
       inherit transformOptions;
     }).optionsJSON;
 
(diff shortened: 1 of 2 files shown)


MattSturgeon:
> This PR should be close to zero rebuilds, with the exception of some of the "page summary" derivations now having a more descriptive drv name.

Looking at the [build logs](https://buildbot.nix-community.org/#/builders/1017/builds/6161/steps/1/logs/stdio), this holds true. The only derivations being rebuilt are the `doc-summary-*` derivations.

GaetanLepage (approved):

Issue #1141, opened 2024-02-21
[BUG] Conform plugin does not install enabled formatters


alisonjenkins:
| Field        | Description |
|--------------|-------------|
| Plugin       | `conform-nvim`  | 
| Nixpkgs      | `pkgs.vimPlugins.conform-nvim` | 
| Home Manager | `<VERSION>` | 

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description

When a formatter is enabled in conform-nvim it is possible that the formatter is not installed on the system and then not work. It would be beneficial to make it so that if a formatter is enabled the formatter is installed from nixpkgs where possible.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  conform-nvim = {
    enable = true;
    formattersByFt = {
      "*" = [ "codespell" ];
      "_" = [ "trim_whitespace" ];
      go = [ "goimports" "golines" "gofmt" "gofumpt" ];
      javascript = [ [ "prettierd" "prettier" ] ];
      json = [ "jq" ];
      lua = [ "stylua" ];
      nix = [ "alejandra" ];
      python = [ "isort" "black" ];
      rust = [ "rustfmt" ];
      sh = [ "shfmt" ];
      terraform = [ "terraform_fmt" ];
    };
  };
}
```
Unless installed by some other means any one of these formatters could be missing and as a result not function.

Bodleum:
The same thing happens with linters.

charludo:
Same here. As a workaround, you can just add the required packages to `nixvim.extraPackages`, for example:

```nix
  programs.nixvim.plugins.lint.lintersByFt.python = [ "pylint" ];
  programs.nixvim.plugins.conform-nvim.formattersByFt.python = [ "black" ];
  programs.nixvim.extraPackages = with pkgs; [ black pylint ];
```

This way the packages will be available inside `nixvim`, but not pollute the `PATH` outside of it.

MattSturgeon:
We should probably make it so that the `formattersByFt` list values can be either string or package

charludo:
> We should probably make it so that the `formattersByFt` list values can be either string or package

The issue is that the string that Conform expects is not always the same as a packages' name. IMO it would be easier to handle it just like (for example) the LSP plugin is handled, where enabling a server (formatter/linter here) also installs the package (via an internal mapping similar to that in the LSP plugin)

[...]

Issue #2412, opened 2024-10-12
generated: unpackaged options should display a better error message


MattSturgeon:
Currently we rely on a no-default package option when there is no known package in nixpkgs for a none-ls, efmls, or LSP server.

This results in a fairly generic `option was accessed but not defined` error.

Instead we should handle these cases with our own custom assertion so that we can provide a more useful error message.

PR #3682, opened 2025-09-13
plugins/conform-nvim: automatically install formatters

saygo-png:
Closes https://github.com/nix-community/nixvim/issues/1141

This PR tries to automatically install the needed packages for Conform formatters.
It extracts formatters names from the `settings.formatters_by_ft`  option. Then for each name
it searches for a corresponding package in 4 attribute sets, in order of priority

1. User overrides
2. Manually written special cases
3. Auto generated defaults
4. Custom formatters (all set to null)

A user can choose what package a formatter should install, especially useful for versioned formatters like for python 3.12 or 3.13. Since this value can be null, they can also make a formatter not install anything.

The automatic package attrset right now is created with `pkgs.${conform-name}`.  More advanced logic could be added, such as replacing all `_` with `-` and checking if that exists in nixpkgs. But i choose against this since it would only cut down the manual list by a few entries but would add a lot of hard to predict behavior. The current simple defaulting mechanism cuts the manual list in about half.

The test right now is very simple. Upstream has each formatter in its own file with a name identical to the Conform name. So I just list that directory of the nix derivation for the plugin, remove `.nix` and generate a Nix list of strings.
To prevent IFD this is generated with CI. Then I just create a Nixvim config with every supported formatter in Conform.

Limitations of this approach:
 1.  There could be a Conform formatter "foo" that matches nixpkgs `pkgs.foo`. But the formatter expects executable `foofmt` meanwhile the nixpkg does not contain it. This will not be caught as an error in the test.
 2.  It could also be the case that Conform formatter "foo" matches `pkgs.foo` AND contains an executable with the same name, but it actually is a totally different executable and package, just with incidentally matching names.
 
 I think one could write a more extensive test that covers 1 using Lua but i'm not good enough with Lua to do that.
 
 There is no test for the override functionality or one that checks if custom formatters' packages are getting set to `null`
 This is because I have no clue how to write that. I'd have to eval those options first, read them, and assert that some predicates hold. For example for the nulling i'd imagine I would read the `customFormatters` attrset, get the values with `lib.attrValues` and assert that everything is null. If someone could point me to how to achieve that I could write those tests.


saygo-png:
Almost forgot about this:
The `example` for the `overrides` option does not render properly in the docs. The issue is with the python package. How do I add that properly? I know i can brute force it with just a plain string but is there no way to make this package render nicely without turning the entire example into a string?

saygo-png:
Just found issue https://github.com/nix-community/nixvim/issues/2412
This pr is not as ready as I thought. I'm gonna change the behavior for unpackaged formatters. I'll send a warning if a user declares a formatter that isn't packaged.

ci/conform-nvim.nix
```diff
@@ -0,0 +1,19 @@
...
+  supportedFormatters =
+    let
+      luaFiles = builtins.filter (lib.hasSuffix ".lua") formatterFiles;
+    in
+    map (lib.removeSuffix ".lua") luaFiles;
+in
+writeText "conform-formatters.nix" (
+  "# WARNING: DO NOT EDIT\n"
+  + "# This file is generated with packages.<system>.conform-formatters, which is run automatically by CI\n"
+  + (lib.generators.toPretty { } supportedFormatters)
```
MattSturgeon: I know we've done this in the past, but I'd prefer to avoid generating nix code. Generating an actual data format such as CSV, txt lines, JSON, TOML, YAML, etc is much more appropriate.

plugins/by-name/conform-nvim/default.nix
```diff
@@ -43,6 +48,26 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      enable = lib.mkEnableOption "
+        automatic installation of formatters listed in `settings.formatters_by_ft` and `settings.formatters`
+      ";
+      overrides = lib.mkOption {
+        type = with types; attrsOf (nullOr package);
+        default = { };
+        example = {
+          "treefmt" = null;
+          "pyproject-fmt" = lib.literalExpression "pkgs.python312Packages.pyproject-parser";
+        };
```
MattSturgeon: You can't nest a literalExpression within a value. It needs to be assigned directly to the docs value (in this case, to `example`).

```suggestion
        example = lib.literalExpression ''
          {
            "treefmt" = null;
            "pyproject-fmt" = pkgs.python312Packages.pyproject-parser;
          }
        '';
```

plugins/by-name/conform-nvim/default.nix
```diff
@@ -234,4 +259,18 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+    let
+      inherit (import ./auto-install.nix { inherit pkgs lib; }) getPkgFromConformName collectFormatters;
+      getPkgFromConformName' = getPkgFromConformName {
+        customFormatters = mapAttrs (_: _: null) cfg.settings.formatters;
+        inherit (cfg.autoInstall) overrides;
+      };
+      names = collectFormatters (attrValues cfg.settings.formatters_by_ft);
+    in
+    {
+      extraPackages = lib.optionals cfg.autoInstall.enable (map getPkgFromConformName' names);
```
MattSturgeon: `lib.mkIf` and `mkMerge` are almost always the better approach when creating module-system option definitions.

The difference is that creating an "`optional`" definition will _always_ create a definition, either of `[]` or the actual value (`map ... names`), while using `mkIf` will only create the definition if the condition is true.

This comes into play when deciding which "override" priority to use, and when merging multiple definitions. E.g. a `mkIf false (mkForce "")` definition will not actually create the `mkForce ""` definition, so the highest override priority will likely remain lower than "force".

tests/test-sources/plugins/by-name/conform-nvim/default.nix
```diff
@@ -4,6 +4,18 @@
     plugins.conform-nvim.enable = true;
   };
 
+  all-formatters =
+    let
+      allFormatters = import ../../../../../generated/conform-formatters.nix;
```
MattSturgeon: Does the test enforce that all formatters listed here have an attribute in `plugins/by-name/conform-nvim/formatter-packages.nix`? I can't see any assertions for missing attributes but I'm probably mis-understanding how the test works.

tests/test-sources/plugins/by-name/conform-nvim/default.nix
```diff
@@ -4,6 +4,18 @@
...
+    in
```
MattSturgeon, line 10: Looks like this test will need to be able to "disable" specific formatters. Or maybe it's better to handle that in the base option?

Either way, CI is currently failing to build `gci`, which is a dependency of the `all-packages` test: https://buildbot.nix-community.org/#/builders/4446/builds/516/steps/1/logs/stdio

saygo-png: This PR already has that (in `formatter-packages.nix`), I just didn't disable `gci` earlier because it wasn't failing... which makes me very confused as to why is is now after updating the example. The inputs didn't change so it shouldn't be different...

saygo-png:
I've added warnings for packages that do not get installed, they look like this:

```
evaluation warning: Nixvim (conform-nvim): You have enabled the 'blue' formatter that relies on a package marked 'unpackaged'.
                    Because of that it will not be installed. To disable this warning, explicitly disable installing the package
                    by setting the 'conform-nvim.autoInstall.overrides.blue' option to 'null'. You can also disable
                    all warnings related to packages not installed by autoInstall with 'conform-nvim.autoInstall.enableWarnings'.
```

I did a small refactor of the `generate.nix` script to make it easier to add more JSON generated files.

I think the code for this looks like a bit of a disaster but it does work and I couldn't come up with a cleaner implementation. It would look better in Haskell with algebraic data types and pattern matching :).

tests/test-sources/plugins/by-name/conform-nvim/default.nix
saygo-png: Yes with a caveat. Each formatter from that list goes through the `or` chain. If it's not found in any of the attrsets Nix throws an evaluation error. The caveat is that they are not all in `plugins/by-name/conform-nvim/formatter-packages.nix` because of edge cases (look at the `or` chain) and defaulting. If a conform name matches a package exactly in nixpkgs, then it is automatically picked. This decreases the size of the explicit list by around half.

MattSturgeon:
> It would look better in Haskell with algebraic data types and pattern matching :).

Just for the record, the generate scripts can technically be in any language. We commonly use bash either at buildtime (`runCommand`) or at runtime (`writeShellApplication`, `writeShellScriptBin`), sometimes we also use python scripts. However, the language chosen should be one that @nix-community/nixvim-committers are comfortable with; we're probably not ready for haskell just yet, but I'd personally love to learn at some point™ 😁

ci/conform-nvim.nix
```diff
@@ -0,0 +1,15 @@
...
+  writeText,
+}:
+let
+  formattersDir = "${vimPlugins.conform-nvim}/lua/conform/formatters";
+  formatterFiles = builtins.attrNames (builtins.readDir formattersDir);
+  supportedFormatters =
+    let
+      luaFiles = builtins.filter (lib.hasSuffix ".lua") formatterFiles;
+    in
+    map (lib.removeSuffix ".lua") luaFiles;
```
MattSturgeon: Tip: I often find this pattern of "bind each step in the process to an intermediate let variable" is better represented as a `pipe`.

Let bindings are usually more useful when you either: a) want to cache the value, b) use the value multiple times, c) need to name the value for clarity

Here, a pipe could be:

```nix
lib.pipe "${vimPlugins.conform-nvim}/lua/conform/formatters" [
  builtins.readDir
  builtins.attrNames
  (builtins.filter (lib.hasSuffix ".lua"))
  (map (lib.removeSuffix ".lua"))
]
```

Or using the [experimental pipe operator](https://github.com/NixOS/rfcs/pull/148):

```nix
"${vimPlugins.conform-nvim}/lua/conform/formatters"
  |> builtins.readDir
  |> builtins.attrNames
  |> builtins.filter (lib.hasSuffix ".lua")
  |> map (lib.removeSuffix ".lua")
```

(Currently, we only use the experimental operator [in a script](https://github.com/nix-community/nixvim/blob/7f45eae65baa38d77d09e0fcf5bfeab6c0f733c0/ci/version-info/supported-versions.nix#L17-L22) that is itself evaluated by another script that [can specify nix config](https://github.com/nix-community/nixvim/blob/7f45eae65baa38d77d09e0fcf5bfeab6c0f733c0/ci/version-info/default.nix#L16-L18), so I wouldn't recommend it here)

---

Optionally, you could also tack `builtins.toJSON` onto the end of the pipe. In fact, you could even add `(writeText "conform-formatters")` after that in the pipe and the whole suddenly becomes a single pipe!

```nix
lib.pipe "${vimPlugins.conform-nvim}/lua/conform/formatters" [
  builtins.readDir
  builtins.attrNames
  (builtins.filter (lib.hasSuffix ".lua"))
  (map (lib.removeSuffix ".lua"))
  builtins.toJSON
  (writeText "conform-formatters")
]
```

ci/conform-nvim.nix
```diff
@@ -0,0 +1,15 @@
+{
+  vimPlugins,
+  lib,
+  writeText,
+}:
+let
+  formattersDir = "${vimPlugins.conform-nvim}/lua/conform/formatters";
```
MattSturgeon: Note, doing a `readDir` on this is an example of [IFD](https://nixos.org/manual/nix/stable/language/import-from-derivation). Not too important here in a generation script where we don't care about eval performance, but just so you are aware when writing other expressions.

The main issue with IFD is that eval must pause while it builds or fetches the derivation to be imported from, which can be much slower than evaluating once and then building everything after; hopefully in parallel.

If you wanted to avoid IFD here, you'd do it by moving all of this logic into the build script. E.g. using `runCommand`, you'd pass it `vimPlugins.conform-nvim` or `"${vimPlugins.conform-nvim}/lua/conform/formatters"` as a derivation attr, which creates a build-time env var. Then you would do scan the directory in the build script, e.g. using `ls`, `find`, or a shell glob.

As I say: not important here. This is an FYI not a request for changes.

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+        conformName = cn;
+        mark = eitherPkg;
+      }
+    else
+      eitherPkg;
+
+  mkWarnsFromEitherPkgList =
+    let
+      mkWarn =
+        x:
```
MattSturgeon: Nit: It's probably more meaningful to do:

```suggestion
        {
          conformName,
          mark,
          ...
        }:
```

or

```suggestion
        {
          conformName,
          mark,
        }:
```

(as appropriate)

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+        lib.nixvim.mkWarnings "conform-nvim" [
+          {
+            when = true;
+            message = ''
+              You have enabled the '${x.conformName}' formatter that relies on a package marked '${x.mark}'.
+              Because of that it will not be installed. To disable this warning, explicitly disable installing the package
+              by setting the 'conform-nvim.autoInstall.overrides.${x.conformName}' option to 'null'. You can also disable
```
MattSturgeon: Tip: you can string-interpolate an option (from the `options` set) to get it's full option-path, including any prefix like `programs.nixvim`:

```suggestion
              by setting the '${opts.autoInstall.overrides}.${x.conformName}' option to 'null'. You can also disable
```

Note: it must be an actual option, not a set of options, since it is only options that have the `__toString` attr.

Also: it can be difficult to get access to sub-options, so sometimes you will want to interpolate the top-level option and then manually type the sub-option path.

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+              all warnings related to packages not installed by autoInstall with 'conform-nvim.autoInstall.enableWarnings'.
+            '';
+          }
+        ];
+    in
+    l: concatMap mkWarn (filter isSTypeAttrSet l);
```
MattSturgeon: I wonder if a pipe is clearer?

```suggestion
    flip pipe [
      (filter isSTypeAttrSet)
      (concatMap mkWarn)
    ];
```

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+
+  collectFormatters =
+    a:
+    let
+      filteredAttrs = filter (x: isString x || isAttrs x) (lib.flatten a);
+      partitioned = partition isString filteredAttrs;
```
MattSturgeon: Again, maybe a pipe?

```suggestion
      partitioned = lib.pipe a [
        lib.flatten
        (filter (x: isString x || isAttrs x))
        (partition isString)
      ];
```

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+              by setting the 'conform-nvim.autoInstall.overrides.${x.conformName}' option to 'null'. You can also disable
+              all warnings related to packages not installed by autoInstall with 'conform-nvim.autoInstall.enableWarnings'.
+            '';
+          }
+        ];
+    in
+    l: concatMap mkWarn (filter isSTypeAttrSet l);
+
+  collectFormatters =
+    a:
```
MattSturgeon: Should `a` be called `formatters`?

plugins/by-name/conform-nvim/default.nix
```diff
@@ -43,6 +48,36 @@ lib.nixvim.plugins.mkNeovimPlugin {
         "formatOnSave"
       ];
 
+  extraOptions = {
+    autoInstall = {
+      enable = lib.mkEnableOption "
+        automatic installation of formatters listed in `settings.formatters_by_ft` and `settings.formatters`
+      ";
```
MattSturgeon: Does `mkEnableOption` trim the string? I'm worried the trailing `\n` may result in the final `.` being on its own line:

```md
Whether to enable automatic installation of formatters listed in `settings.formatters_by_ft` and `settings.formatters`
.
```

Note: you can test how the docs render fairly easily using `nix run .#docs` or `serve-docs` in the devshell.

plugins/by-name/conform-nvim/default.nix
```diff
@@ -43,6 +48,36 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        default = { };
+        example = lib.literalExpression ''
+          {
+            "treefmt" = null;
+            "pyproject-fmt" = pkgs.python312Packages.pyproject-parser;
+          };
+        '';
+        description = ''
+          Attribute set of formatter names to nix packages.
+          Can be set to null to disable auto-installing of a specific formatter.
```
MattSturgeon:
```suggestion
          Individual formatters can be set to null to disable auto-installing of a specific formatter.
```

Or maybe

```suggestion
          You can set a formatter to `null` to disable auto-installing its package.
```

plugins/by-name/conform-nvim/default.nix
```diff
@@ -234,4 +269,27 @@ lib.nixvim.plugins.mkNeovimPlugin {
       }
     ```
   '';
+
+  extraConfig =
+    cfg:
```
MattSturgeon: Tip: if you want to be able to use options elsewhere, e.g. in `auto-install` assertions, you can access the plugin's option set here as:

```suggestion
    cfg: opts:
```

Or the top-level nixvim `options` set at the top of the file as a normal module arg.

plugins/by-name/conform-nvim/default.nix
```diff
@@ -234,4 +269,27 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      inherit (import ./auto-install.nix { inherit pkgs lib; })
+        getEitherPkgFromName
+        collectFormatters
+        cleanEitherPkgList
+        mkWarnsFromEitherPkgList
+        ;
+      getEitherPkgFromName' = getEitherPkgFromName {
+        customFormatters = mapAttrs (_: _: null) cfg.settings.formatters;
+        inherit (cfg.autoInstall) overrides;
+      };
```
MattSturgeon: I don't quite understand this. Should it be clarified?

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+  inherit (import ./formatter-packages.nix { inherit pkgs lib; }) sType formatter-packages;
+  sTypeList = lib.mapAttrsToList (_: v: v) sType;
+  isSTypeAttrSet = x: lib.elem (x.mark or "") sTypeList;
+in
+rec {
+  cleanEitherPkgList = filter (x: !isSTypeAttrSet x);
+
+  getEitherPkgFromName =
+    { customFormatters, overrides }:
+    cn:
```
MattSturgeon: What is `cn`? Can we name it more verbosely so that it becomes self-documenting?

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+    filter
+    isString
+    isAttrs
+    attrValues
+    concatMap
+    partition
+    ;
+
+  inherit (import ./formatter-packages.nix { inherit pkgs lib; }) sType formatter-packages;
+  sTypeList = lib.mapAttrsToList (_: v: v) sType;
```
MattSturgeon:
```suggestion
  sTypeList = attrValues sType;
```

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,65 @@
...
+  isSTypeAttrSet = x: lib.elem (x.mark or "") sTypeList;
```
MattSturgeon: Probably more explicit to say either:

```suggestion
  isSTypeAttrSet = x: x ? mark && lib.elem x.mark sTypeList;
```

Which avoids calling `elem` at all, but evaluates `x` twice.

Or

```suggestion
  isSTypeAttrSet = x: lib.elem (x.mark or null) sTypeList;
```

Which is clearer that `null` is a nothing-value.

Also: I'd probably prefer a more descriptive name than `x`. What is an _`x`_ here? If the function is for use with _any_ attrset maybe call it `set` or `attrs`?

saygo-png:
> Just for the record, the generate scripts can technically be in any language.

Sorry I was unclear; I meant the Nix code for the plugin which obviously is staying Nix.

ci/conform-nvim.nix
saygo-png: I'm aware of IFDs. I'm using it intentionally here as IFDs are allowed in generation scripts.

plugins/by-name/conform-nvim/default.nix
saygo-png: image

Looks fine

plugins/by-name/conform-nvim/default.nix
saygo-png: I'm partially applying the function and get a function back.

`customFormatters` is a mapping of all user defined formatters (`cfg.settings.formatters`) to `null`.

This is done so the `or` chain finds a defined value for them and doesn't throw an error about a missing attribute. Since these formatters are custom, it's impossible to know what package they want i just null them so they don't install anything. Users can still map these formatters to a package using `overrides`.

Not sure how i'd clarify this in code.

plugins/by-name/conform-nvim/auto-install.nix
saygo-png: I don't understand this change

plugins/by-name/conform-nvim/auto-install.nix
saygo-png: Keeping `x` here as this function is a predicate that works on anything. The abstract name shows the intention of a general predicate.

plugins/by-name/conform-nvim/auto-install.nix
saygo-png: I don't think it is in this case because you are flipping the arguments. The lambda is simpler.

plugins/by-name/conform-nvim/default.nix
MattSturgeon: > I'm partially applying the function and get a function back.

That part was clear 😅 

I guess it's the function name itself that's throwing me off...

Aside: typically, a function that has a configurable version and a per-configured version will use the "With" suffix, e.g. `lib.strings.trim` and `lib.strings.trimWith`.

But that aside, I think the sentence "get either package from name" doesn't really make sense to me. The "either" part makes me expect the rest of the sentence to tell me what the options are, the "from name" part implies that the name is where we will find the package.

Based on the `or` chain you describe, it sounds like we're "selecting" a package from a list of possible providers; those providers including `overrides`, nixpkgs, and `customFormatters` as the fallback? How about simply `selectPackage` (and `selectPackageWith`)? Or if you want to explicitly highlight that it uses the name, `getPackageByName` or `selectPackageByName`?

> customFormatters

Would it make sense to pass this as a list instead of a set of nulls, or are you using something like `intersectAttrs`?

Or maybe it'd be simpler to pass the `settings.formatters` set as is, and ignoring its values would be an implementation detail of the underlying function; rather than something the caller has to handle.

Would "enabledFormatters" be a better name? Or "formatterNames"? Or "configuredFormatters"? Or if the intention is that this is a set of fallback values, maybe that should be reflected in its name: "defaultPackages" or "fallbackFormatters"?

> Since these formatters are custom

Do we know they're (all) ~~custom~~ _unknown_ at this stage? Isn't this the full list of all configured formatters, known or not?

plugins/by-name/conform-nvim/auto-install.nix
MattSturgeon: TBH I only flipped the argument in that suggestion to avoid having to name `l`; that's one advantage of a pipe is that you _can_ represent it as a partially applied function, something that _can't_ be done with the nested function application.

`flip pipe` is such a common pattern that I think most people understand what it does. But I agree, it's technically a layer of abstraction that needs to be understood; i.e. a complexity. I can't wait for the pipe operator to become stable! 😁 

However, I'd still personally prefer:

```suggestion
    somBetterNameForL: pipe somBetterNameForL [
      (filter isSTypeAttrSet)
      (concatMap mkWarn)
    ];
```

Aside: what does `l` even stand for here? `list`?

plugins/by-name/conform-nvim/default.nix
MattSturgeon: It _has_ added whitespace. If you were to inspect the actual nix string in the repl, you'd see it is a newline; this newline then gets rendered as a `" "` space when rendering the markdown into HTML.

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -0,0 +1,71 @@
...
+rec {
+  cleanEitherPkgList = filter (x: !isSTypeAttrSet x);
+
+  getEitherPkgFromName =
+    { customFormatters, overrides }:
+    conformName:
+    let
+      eitherPkg =
+        overrides.${conformName} or formatter-packages.${conformName} or pkgs.${conformName}
+          or customFormatters.${conformName};
```
MattSturgeon: I think this is the cause of my confusion in https://github.com/nix-community/nixvim/pull/3682#discussion_r2356333124

To me, "either" implies one of two. It _can_ be used with more than two, but that's less common and usually only happens when all options are explicitly listed in the sentence.

E.g., this example has three:

> Either I'm stupid, tired, or both!

It's also confusing to me to use "either" as a noun (or part of a compound noun); once you have the package it's now... just a package, it no longer represents the union implied by "either".

How about we try to drop the "either" terminology throughout and instead refer to selecting, choosing, or getting, a package from one of the possible sources?

Here, I'd suggest:

```suggestion
      package =
        overrides.${name} or formatter-packages.${name} or pkgs.${name}
          or customFormatters.${name};
```

Since `customFormatters` is always a set of nulls, couldn't we replace `or customFormatters.${name}` with `or null`?

If not, perhaps is could be renamed to `settingsFormatters`, `configuredFormatters`, `userFormatters`, etc? Or if it should be more general, perhaps `permittedFallbacks`, `fallbackFormatters`, `defaultValues`, etc?

saygo-png: > To me, "either" implies one of two.

In my head i was thinking of it that way too, as in `either package or sType` where sType is a sum type of 3 constructors
```haskell
data SType = Unpackaged | DarwinOnly | Broken
data EitherPkg = Pkg | Null | SType
```
Here i called it eitherPkg because it could either be a package or this sum/set type or null.

> Since customFormatters is always a set of nulls, couldn't we replace or customFormatters.${name} with or null?

Sorry I think i'm doing a poor job of explaining this but bare with me:
No, because we want to throw an error on undefined formatters. If we always default to null then there will be no errors and the "all-formatters" test will be pointless. The point of mapping nulls here is to *only* not throw on custom formatters. This way we can maintain a mapping of all available conform names to packages or nulls.

I'll rename to configuredFormatters in a sec

MattSturgeon: > > To me, "either" implies one of two.
> 
> In my head i was thinking of it that way too, as in `either package or sType` where sType is a sum type of 3 constructors
> 
> ```haskell
> data SType = Unpackaged | DarwinOnly | Broken
> data EitherPkg = Pkg | SType
> ```

Maybe my confusion is coming from you thinking about this from the perspective of the types in question, while I'm thinking from the perspective of the actual variable (and its value)?

I'm still not convinced `getEitherPkgFromName` clearly gets across that the "either" here refers to the fact its result may either be a package, an sType, or null.

> Here i called it eitherPkg because it could either be a package or `null`

I think I'm starting to understand.

Howabout `maybePkg` or `maybePackage`? Or just call it `package` and clarify if necessary that it is nullable?

> No, because we want to throw an error on undefined formatters. If we always default to null then there will be no errors and the "all-formatters" test will be pointless. The point of mapping nulls here is to _only_ not throw on custom formatters. 

Makes sense. I wonder if it'd be more explicit, albeit less elegant, to instead pass the "permitted names" as a list and throw if name isn't in that list?

```
or (throwIfNot (elem name permittedNames) "..." null)
```

saygo-png: I like the `throwIfNot` approach, I've implemented that. The error message asks if the name is correct because a user could make a simple typo and get that the "package not found error" which could be confusing.

I also went with `maybePackage` since that seems to be the most fitting name.

MattSturgeon: > The error message asks if the name is correct because a user could make a simple typo and get that the "package not found error" which could be confusing.

Good point. One advantage of your original implementation is that nix's missing attribute error will try to suggest possible typo candidates if there are similarly named attributes that _do_ exist.

However, if this is a potentially user-facing error it is probably best to craft a bespoke error message. 👍

ci/generate.nix
```diff
@@ -4,6 +4,7 @@
   efmls-configs-sources,
   none-ls-builtins,
   lspconfig-servers,
+  conform-formatters,
   nixfmt-rfc-style,
```
khaneliman:
```suggestion
  nixfmt,
```

nit: unrelated to your change, but this should point to stable now

plugins/by-name/conform-nvim/formatter-packages.nix
```diff
@@ -0,0 +1,146 @@
...
+    jsonnetfmt = jsonnet;
+    inherit (texlive.pkgs) latexindent;
+    lua-format = luaformatter;
+    mago_format = mago;
+    mago_lint = mago;
+    markdownlint = markdownlint-cli;
+    mix = beamMinimal28Packages.elixir;
+    nginxfmt = nginx-config-formatter;
+    nimpretty = nim;
+    nixpkgs_fmt = nixpkgs-fmt;
```
khaneliman, line 115:
```suggestion
    inherit nixfmt;
    nixpkgs_fmt = nixpkgs-fmt;
```
? I see it in the list at https://github.com/nix-community/nixvim/pull/3682/files#diff-e933535c9f5bd07146c4b67a8f94982c4d4f24cb3e3f70db608b8877b65be9ecR136 but don't see it here.

would be nice to have a test that everything from our generated list is accounted for, i believe we do something like that in lsp servers?

saygo-png: The conform name is identical to the nix package so the package is found without an explicit mapping.

The test i wrote accounts for every formatter from the generated list, there just is this defaulting behaviour.

khaneliman: Sorry, just briefly skimmed and must have missed how it was being tested to account for that.

khaneliman (approved):
Thanks for doing this, I meant to do it a long time ago!

Seems to work fine for most formatters. For some reason, I get a lua error trying  to use `ConformInfo` when letting `csharpier` get auto installed. But, if I leave my manual configuration of the pkg mapping in `formatters.*.command` it is fine.

MattSturgeon:
> But, if I leave my manual configuration of the pkg mapping in `formatters.*.command` it is fine.

Would that make for a suitable example (in a followup PR)?

GaetanLepage:
Well done @saygo-png!

saygo-png:
@khaneliman The csharpier issue is an upstream issue with conform-nvim https://github.com/stevearc/conform.nvim/issues/699 not this PR

khaneliman:
> @khaneliman The csharpier issue is an upstream issue with conform-nvim [stevearc/conform.nvim#699](https://github.com/stevearc/conform.nvim/issues/699) not this PR

Looks like someone created a PR for it, too. Thanks for identifying it. https://github.com/stevearc/conform.nvim/pull/772

PR #3691, opened 2025-09-16
patterns: add module

khaneliman:
A simple pattern viewer for Neovim with support for Lua patterns & regex.
https://github.com/OXY2DEV/patterns.nvim

Oxy has insane customization options available.. limited to just a few in the defaults case.

### DIFF
--- a/plugins/by-name/patterns/default.nix
+++ b/plugins/by-name/patterns/default.nix
@@ -1,0 +1,47 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "patterns";
+  packPathName = "patterns.nvim";
+  package = "patterns-nvim";
+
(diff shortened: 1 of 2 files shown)


plugins/by-name/patterns/default.nix
```diff
@@ -0,0 +1,47 @@
...
+      };
+      hover = {
+        "e" = {
+          callback = "edit";
+        };
+      };
+    };
+
+    windows = {
+      hover.__raw = ''
```
MattSturgeon: Nit: it may be better to use `nestedLiteralLua` here, or to wrap the whole example as a `literalExpression` to preserve your `foo.__raw` formatting.

Otherwise, `toPretty` will produce:
```
foo = {
  __raw = "";
}
```

khaneliman: Didn't even know we had that function, updated :P

PR #3694, opened 2025-09-16
opencode: add module

khaneliman:
Seamlessly integrate the opencode AI assistant with Neovim — convenient and editor-aware research, reviews, and requests.
https://github.com/NickvanDyke/opencode.nvim

### DIFF
--- a/plugins/by-name/opencode/default.nix
+++ b/plugins/by-name/opencode/default.nix
@@ -1,0 +1,36 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "opencode";
+  packPathName = "opencode.nvim";
+  package = "opencode-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  description = ''
+    OpenCode.nvim provides seamless integration with Claude Code for AI-assisted development.
+
+    > [!NOTE]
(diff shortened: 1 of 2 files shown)


plugins/by-name/opencode/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  description = ''
+    OpenCode.nvim provides seamless integration with Claude Code for AI-assisted development.
+
+    > [!NOTE]
+    > Recommended: `snacks.enable` with `settings.input.enabled = true` for better prompt input
+    > Required: `snacks.enable` to use opencode.nvim's embedded terminal
+
+    > [!TIP]
+    > Set `opts.autoread = true` if using the `auto_reload` option.
```
MattSturgeon, line 17: The note above has reasons _why_ `snacks` is recommended/required, but this tip doesn't.

The user may wonder why `autoread` is recommended.

MattSturgeon:
@khaneliman anything blocking this? I don't think https://github.com/nix-community/nixvim/pull/3694#discussion_r2356364848 _necessarily_ needs addressing as it was mostly just an observation.

khaneliman:
Sorry, forgot about it. Can merge, was just trying to look into it a bit more on the technical aspects to see if there was a better note other than just copying upstream's note (and saw a squirrel).

PR #3705, opened 2025-09-23
plugins/chadtree: fix misplaced ignore settings

adfaure:
I tried to modify the ignore settings for chadtree but it has been misplaced in `options.ignore`. The proper place is outside of `options` (https://github.com/ms-jpq/chadtree/blob/chad/docs/CONFIGURATION.md#chadtree_settingsignore)

To check that it is actually broken, you can add the following to your `nixvim` configuration. That should build, but chadtree throws an error when we start neovim.

```nix
  plugins.chadtree = {
    enable = true;
    ignore = {
      nameExact = [ "__pycache__" ".git" ];
    };
  };
```

### DIFF
--- a/plugins/by-name/chadtree/default.nix
+++ b/plugins/by-name/chadtree/default.nix
@@ -70,31 +70,29 @@ in
       versionControl = helpers.defaultNullOpts.mkBool true ''
         Enable version control. This can also be toggled. But unlike `show_hidden`, does not have a default keybind.
       '';
-
-      ignore = {
-        nameExact =
-          mkListStr
-            [
-              ".DS_Store"
-              ".directory"
-              "thumbs.db"
-              ".git"
-            ]
-            ''
-              Files whose name match these exactly will be ignored.
-            '';
-
-        nameGlob = mkListStr [ ] ''
-          Files whose name match these glob patterns will be ignored.
-          ie. `*.py` will match all python files
-        '';
-
-        pathGlob = mkListStr [ ] ''
-          Files whose full path match these glob patterns will be ignored.
-        '';
-      };
     };
+    ignore = {
+      nameExact =
+        mkListStr
+          [
+            ".DS_Store"
+            ".directory"
+            "thumbs.db"
+            ".git"
+          ]
+          ''
+            Files whose name match these exactly will be ignored.
+          '';
 
+      nameGlob = mkListStr [ ] ''
+        Files whose name match these glob patterns will be ignored.
+        ie. `*.py` will match all python files
+      '';
+
+      pathGlob = mkListStr [ ] ''
+        Files whose full path match these glob patterns will be ignored.
+      '';
+    };
     view = {
       openDirection =
         helpers.defaultNullOpts.mkEnum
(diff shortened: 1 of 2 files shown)


adfaure:
I will fix the tests asap.

GaetanLepage:
Thanks for the contribution!
However, we would need to add some deprecation notices for such changes.
I propose to first [migrate the plugin to the new `mkNeovimPlugin` builder](https://github.com/nix-community/nixvim/pull/3708). It will then be easier to make this change.

adfaure:
> Thanks for the contribution! However, we would need to add some deprecation notices for such changes. I propose to first [migrate the plugin to the new `mkNeovimPlugin` builder](https://github.com/nix-community/nixvim/pull/3708). It will then be easier to make this change.

Hi ! 
Sorry, I don't know much about the internals of nixvim so I don't know about the migration. I just found a bug that affected me and that was "easy" to solve, so I did it.
Currently, I am using my fork since the current state is broken. In my opinion, the sooner it is merged, the better.

I trust your judgment on this; please let me know when I can rebase, or you can cherry-pick my commit. It's up to you.

GaetanLepage:
> > Thanks for the contribution! However, we would need to add some deprecation notices for such changes. I propose to first [migrate the plugin to the new `mkNeovimPlugin` builder](https://github.com/nix-community/nixvim/pull/3708). It will then be easier to make this change.
> 
> Hi ! Sorry, I don't know much about the internals of nixvim so I don't know about the migration. I just found a bug that affected me and that was "easy" to solve, so I did it. Currently, I am using my fork since the current state is broken. In my opinion, the sooner it is merged, the better.
> 
> I trust your judgment on this; please let me know when I can rebase, or you can cherry-pick my commit. It's up to you.

I added a commit on top of #3708 that fixes the rename warnings for the `ignore.*` options.
Hence, once it will be merged, a user having the following config:
```nix
plugins.chadtree = {
  enable = true;
  options.ignore.nameExact = [ "foo" ];
};
```
will see the following warning:
```
The option `plugins.chadtree.options.ignore.nameExact' defined in `/nix/store/yar58vxxpv7mdxfcnclk753l9pbr4pvj-source/tests/test-sources/plugins/by-name/chadtree/default.nix' has been renamed to `plugins.chadtree.settings.ignore.name_exact'.
```
Nixvim will automatically rewrite his configuration to:
```nix
plugins.chadtree = {
  enable = true;
  settings.ignore.name_exact = [ "foo" ];
};
```

adfaure:
I don't understand what really happened here, but I am glad that it is fixed !
Thank you for your reactivity.

Maybe we can close this MR ?

[edit] Ok, I got it. The new builder handles the renaming of options and generates the appropriate information for the users. Interesting.

GaetanLepage:
> [edit] Ok, I got it. The new builder handles the renaming of options and generates the appropriate information for the users. Interesting.

Yes, that's basically it. This new system is much better, both for us and for the users.
Now, you will be able to arbitrarily set any option under `plugins.chadtree.settings` without requiring us to maintain option declarations in nixvim.

MattSturgeon:
> The new builder handles the renaming of options and generates the appropriate information for the users.

That comes from nixpkgs' `lib.mkRenamedOptionModule` function, which creates an alias at the old name that warns when it is used.

Our `mkNeovimPlugin` function simply provides a central place where things we do across all plugins can be implemented. This improves internal consistency within nixvim.

`lib.mkRenamedOptionModule` is used internally by `mkNeovimPlugin` to create deprecation aliases for the plugin's old options.

PR #3708, opened 2025-09-24
 plugins/chadtree: migrate to mkNeovimPlugin

GaetanLepage:
- **plugins/chadtree: remove web-devicons automatic activation**
- **plugins/chadtree: migrate to mkNeovimPlugin**
- **plugins/chadtree: account for initially wrong declaration for the 'ignore.*' options**


plugins/by-name/chadtree/renamed-options.nix
```diff
@@ -0,0 +1,322 @@
...
+  [
+    "keymap"
+    "toggles"
+    "toggleFollow"
+  ]
+  [
+    "keymap"
+    "toggles"
+    "toggleVersionControl"
+  ]
```
MattSturgeon, line 344: When importing this file, we _could_ use `map (splitString ".")` as some other plugins have done recently. If so, it'd allow a flatter format in this file.

Of course, that only works when no options actually use `.` in their attr names.

tests/test-sources/plugins/by-name/chadtree/default.nix
```diff
@@ -6,130 +6,150 @@
...
-          };
-          rerooting = {
-            changeDir = [ "b" ];
-            changeFocus = [ "c" ];
-            changeFocusUp = [ "C" ];
+      };
+    };
+  };
+
+  bigExample = {
```
MattSturgeon, line 24: 😂

plugins/by-name/chadtree/default.nix
```diff
@@ -1,506 +1,29 @@
...
-                winfixwidth = true;
-                wrap = false;
-            }
-          ''
-          ''
-            Set of window local options to for CHADTree windows.
-          '';
-    };
+  callSetup = false;
+  settingsDescription = "Options provided to `vim.api.nvim_set_var('chadtree_settings', <USER_CONFIG>)`.";
```
MattSturgeon: How about

```suggestion
  settingsDescription = "Options provided as `vim.api.nvim_set_var('chadtree_settings', settings)`.";
```

?

plugins/by-name/chadtree/renamed-options.nix
GaetanLepage: This is very smart. Now that I'm done writing it, I would rather leave it like this ^^

tests/test-sources/plugins/by-name/chadtree/default.nix
GaetanLepage: ;)

plugins/by-name/chadtree/default.nix
GaetanLepage: Sure, done!

MattSturgeon (approved):

GaetanLepage:
Thanks for the review @MattSturgeon!

PR #3709, opened 2025-09-24
rhubarb: init

santoshxshrestha:
Rhubarb extends fugitive.vim with GitHub integration, allowing you to open GitHub files or URLs directly from Vim using `:GBrowse`.


MattSturgeon (approved):
Thanks for your contribution, and for cleaning up the git history!

PR #3712, opened 2025-09-25
plugins/lsp/packages: make ansiblels unpackaged

will-lol:
[ansible-language-server](https://github.com/NixOS/nixpkgs/pull/445884) was removed upstream.
Despite not using this language server, the absence of the package upstream was preventing my configuration from building.
I imagine most configurations would be broken by this once it hits nixpkgs-unstable.
I have moved this into the class of unpackaged LSPs. This allows users to package this server themselves if they wish.
For those wanting to keep their configurations with ansiblels working for the time being, I was able to build but not test (I don't use ansible) using the following:

```
          lsp.servers.ansiblels = {
            enable = true;
            package = pkgs.buildNpmPackage rec {

              pname = "ansible-language-server";
              version = "1.2.1";

              src = pkgs.fetchFromGitHub {
                owner = "ansible";
                repo = "ansible-language-server";
                tag = "v${version}";
                hash = "sha256-e6cOWoryOxWnl8q62rlGmSgwLVnoxLMwNFoGlUZw2bQ=";
              };

              npmDepsHash = "sha256-Lzwj0/2fxa44DJBsgDPa43AbRxggqh881X/DFnlNLig=";
              npmBuildScript = "compile";

              postPatch = ''
                sed -i '/"prepare"/d' package.json
                sed -i '/"prepack"/d' package.json
              '';

              npmPackFlags = [ "--ignore-scripts" ];
            };
          };
```

### DIFF
--- a/plugins/lsp/lsp-packages.nix
+++ b/plugins/lsp/lsp-packages.nix
@@ -5,6 +5,7 @@
     "alloy_ls"
     "anakin_language_server"
     "antlersls"
+    "ansiblels"
     "apex_ls"
     "autohotkey_lsp"
     "awk_ls"
@@ -186,7 +187,6 @@
     aiken = "aiken";
     air = "air-formatter";
     angularls = "angular-language-server";
-    ansiblels = "ansible-language-server";
     arduino_language_server = "arduino-language-server";
     asm_lsp = "asm-lsp";
     ast_grep = "ast-grep";


c2fc2f:
Same problem, I don't use `ansiblels` and so I can no longer rebuild my system *(I want to kill myself)*

culxttes (approved):

wra-bradshaw:
`nix run .#docs` showed that ansiblels would be removed from `lsp.servers` in docs (consistent with other unpackaged servers e.g. intelephense). 
I tried to run `nix flake check --all-systems` but I don't think it can be run on my poor Macbook Air (29.06GB memory usage!)
Screenshot 2025-09-25 at 15 51 04

MattSturgeon:
> I tried to run `nix flake check --all-systems` but I don't think it can be run on my poor Macbook Air (29.06GB memory usage!)

I wouldn't ever try and run `nix flake check` in nixvim; we have way too many things to evaluate in our `checks` output and it's not optimized for nix garbage collection. In CI we can break this up into smaller chunks of things to eval & build separately.

Instead, use our devshell's `tests` command with `-i` to interactively select the relevant test you actually need to run. Or use `nix run .#docs` to test the docs.

PR #3721, opened 2025-09-27
treewide: infer `packPathName`

saygo-png:
`packPathName` is the result of calling `lib.getName` on a (vim plugin) derivation.

https://github.com/NixOS/nixpkgs/blob/eeb12bd734acc270f419bd91936ebec41689026a/pkgs/applications/editors/vim/plugins/utils/vim-utils.nix#L171-L179

There is no reason why this should be manually entered or even an argument to mk*Plugin

I'll squash the commits after review  — easier to review the semantic changes when they are in a separate commit.


khaneliman:
Appreciate someone having the energy to do this work... been wanting this change for a while just haven't had the energy for open source for a bit with work.

MattSturgeon:
Thanks for working on this!

My initial concern is that we need to ensure evaluating packages is still lazy; a user's configuration shouldn't evaluate any packages it isn't installing. We can verify this with some temporary traces, warnings, throws, etc.

Also: If this is used in user configs we shouldn't be evaluating the package default. We should be evaluating the actually defined package, which we can get from `cfg.package`.

We can't do that if the value is used in the docs though, because the docs shouldn't rely on config values, nor any derivations.

Incidentally, if we wanted the default, we could get it from `opts.package.default` without needing to resolve the "package" argument supplied to `mkNeovimPlugin`.

lib/plugins/mk-neovim-plugin.nix
```diff
@@ -114,15 +114,15 @@ let
...
+                    message = "You have enabled lazy loading for ${name} but have not provided any configuration.";
                   }
                 ];
 
                 plugins.lz-n = {
                   plugins = [
                     (
                       {
-                        __unkeyed-1 = packPathName;
+                        __unkeyed-1 = getPackPathName cfg.package;
```
MattSturgeon: How about inlining and adding a clarifying comment?

```suggestion
                        # The packpath name is always the derivation name
                        __unkeyed-1 = lib.getName cfg.package;
```

saygo-png:
I was clearly tired yesterday looking at this code today...

This is a clean implementation but it comes with a drawback(?). The docs now use the `name` of the plugin instead of the `packPathName`. I don't think this really matters since they are mostly the same. It could bring clarity too since it will now match with our option names.

This should be lazy because the only `lib.getName` call is inside a mkIf that depends on the plugin being lazyloaded.

lib/plugins/mk-neovim-plugin.nix
saygo-png: Comments get out of date and bitrot, i'd rather have this be an alias that clarifies the usage.

saygo-png:
Should i squash the commits or is it better like this?

MattSturgeon:
> The docs now use the `name` of the plugin instead of the `packPathName`. I don't think this really matters since they are mostly the same. It could bring clarity too since it will now match with our option names.

Yeah, I noticed this too.

For context: historically, `packPathName` used to be called "originalName" and it was usually derived from upstream repo names instead of derivation names. When we added lazy-loading we realised that `originalName` was (in practice) almost always the pack-path name, so we renamed it and used it as such.

That said, I don't see much value in listing two names per plugin in the documentation. Even if we do want this, in hindsight we should handle it separately from the packpath name; especially if we can reliably get the packpath name from the derivation name.

If we wanted to maintain the concept of a plugin's "original name" in documentation, then I'd propose we derive this from the plugin's homepage URL where possible and require it be explicitly defined where not possible. E.g. github URLs could be parsed into original names, but other plugins would need to define it explicitly.

TLDR: I personally have no objection to using "name" for all documentation, but if anyone does have an objection: see above.

MattSturgeon:
> Should i squash the commits or is it better like this?

I think it's fine either way. It's certainly easier to review with the treewide and implementation changes separate, although I assume the first commit would fail to eval without the second?

lib/plugins/mk-neovim-plugin.nix
MattSturgeon: > Comments get out of date and bitrot, i'd rather have this be an alias that clarifies the usage.

I don't see how comments are any different to names in this regard, which can also get out of date and bitrot.

The indention of both the name and the comment are to clarify the intent behind the code, and in this case I personally think a comment does a better job.

My worry with having the `getPackPathName` function, especially with a wide scope, is that it may encourage use elsewhere (e.g. in docs, or somewhere less lazy).

khaneliman (approved):
LGTM and my generated init.lua looks identical so the plugins i was lazy loading has the correct names. This should alleviate some maintenance and contribution burden. Thanks

PR #3726, opened 2025-09-27
docs/contributing: change maintainers to a required field

saygo-png:
Current docs are wrong — maintainers is a required argument that does not default to `[]`. 

https://github.com/nix-community/nixvim/blob/e0f1e4ae4bb8762b7c51c3a514ca19664fad9c3b/lib/plugins/mk-neovim-plugin.nix#L5-L9


MattSturgeon (approved):
Thanks!

PR #3729, opened 2025-09-28
docs: improve consistency

saygo-png:
Mostly sets "NixVim" as the standard spelling.  Fixes capitalization for "Neovim". Few other minor tweaks.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
-# Contributing (and hacking onto) nixvim
+# Contributing (and hacking onto) Nixvim
 
-This document is mainly for contributors to nixvim, but it can also be useful for extending nixvim.
+This document is mainly for contributors to Nixvim, but it can also be useful for extending Nixvim.
 
 ## Submitting a change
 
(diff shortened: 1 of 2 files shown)


saygo-png:
This is non exhaustive — there are many more cases of "Nixvim" spelling but I don't want this to turn into another 50+ files changed PR. Especially since some cases are not trivial (For example tests that match on strings containing "Nixvim")

MattSturgeon:
> Mostly sets "NixVim"

Is is just me that finds that capitalization extremely ugly? 😅 

My personal preference is "Nixvim" in prose and "nixvim" in code. This feels more consistent with Neovim; which is never NeoVim 😜 

But thanks a lot for taking the time to improve consistency -- that's certainly a win!

saygo-png:
I don't like it either lol. But this was the name used in the biggest header in the readme. We could change it to "Nixvim". I just didn't mention it since this is such a personal prefrence thing.

saygo-png:
Drafting since i'll change it to "Nixvim" instead

MattSturgeon (approved):
Awesome, thanks!

I notice this only covers `CONTRIBUTING` and `README`. I assume there are plenty of other cases we could look at elsewhere in option descriptions and other docs? Do you think this is worth a tracking issue and/or a script-aided follow-up PR?

Thanks again for another great contribution!

khaneliman:
> > Mostly sets "NixVim"
> 
> Is is just me that finds that capitalization extremely ugly? 😅
> 
> My personal preference is "Nixvim" in prose and "nixvim" in code. This feels more consistent with Neovim; which is never NeoVim 😜
> 
> But thanks a lot for taking the time to improve consistency -- that's certainly a win!

that's why i never want to capitalize. not because i'm too lazy to hit shift.. but because i think it's uglier. :P

saygo-png:
> Do you think this is worth a tracking issue and/or a script-aided follow-up PR?

I think a tracking issue is a bit much but I plan to make a follow-up PR (probably sometime this month) to capitalize it properly in other places. No script is necessary you can just use :cdo with some form of preview to get context.

Issue #2813, opened 2025-01-10
[PLUGIN REQUEST] haskell-tools


MrQubo:
| Field    | Description |
|----------|-------------|
| Plugin   | `haskell-tools` | 
| Homepage | `https://github.com/MrcJkb/haskell-tools.nvim` | 
| Nixpkgs  | `https://github.com/NixOS/nixpkgs/blob/bffc22eb12172e6db3c5dde9e3e5628f8e3e7912/pkgs/development/lua-modules/generated-packages.nix#L646-L667` | 

## Extra Information

Plugin says that haskell-language-server lsp must not be configured manually which conflicts with `plugins.lsp.servers.hls` nixvim module. Nixvim should probably warn/error if `plugins.lsp.servers.hls.enable` is set to `true` when `haskell-tools` plugin is also enabled.

Additional options for enabling optional dependencies, probably most of them should be enabled by default in nixvim.

Might also consider an option for automatically generating hoogle database.

xecarlox94:
Hi guys, I have never contributed to any nix project. do you think this would be a easy contribution since it already is packaged in nixpkgs? 

I would like to have this because I am working on a personal medium size haskell project and it would be really useful to have these tools enable

GaetanLepage:
> Hi guys, I have never contributed to any nix project. do you think this would be a easy contribution since it already is packaged in nixpkgs?
> 
> I would like to have this because I am working on a personal medium size haskell project and it would be really useful to have these tools enable

Sure! Go and give it a try!
I encourage you to have a look at the `CONTRIBUTING.md` and at the recently merged plugin addition PRs :)
Feel free to ask your questions on the [matrix dev channel](https://matrix.to/#/#nixvim-dev:matrix.org)!

PR #3732, opened 2025-09-28
plugins/haskell-tools: init

saygo-png:
https://github.com/MrcJkb/haskell-tools.nvim

closes https://github.com/nix-community/nixvim/issues/2813

This plugin is complicated — it hooks up to a lot of different tools. This PR does not add integrations for them beyond Telescope. I'd certainly want to add one for `iron-nvim` too. I think it will be easier to do that incrementally though. Later i'll play around with local Hoogle installations and see if there is anything we can add on the Nixvim side to make the setup easier.

Lazyloading with Telescope integration is broken, but that's a problem with our telescope module not accounting for lazyloading and loading every extension at one point. This plugin is lazy by design so there is no real need to lazyload it manually.

### DIFF
--- a/plugins/by-name/haskell-tools/default.nix
+++ b/plugins/by-name/haskell-tools/default.nix
@@ -1,0 +1,64 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "haskell-tools";
+  package = "haskell-tools-nvim";
+  maintainers = [ lib.maintainers.saygo-png ];
+
+  # This is a filetype plugin that doesn't use a setup function.
+  # Configuration is passed to a global table.
+  callSetup = false;
+
+  extraOptions = {
+    enableTelescope = lib.mkEnableOption "telescope integration";
+    hlsPackage = lib.mkPackageOption pkgs "haskell-language-server" {
+      nullable = true;
+    };
+  };
+
+  settingsExample = {
+    hls = {
+      default_settings = {
+        haskell = {
+          formattingProvider = "ormolu";
+          plugin = {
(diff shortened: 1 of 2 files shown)


plugins/by-name/haskell-tools/default.nix
```diff
@@ -0,0 +1,64 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin rec {
+  name = "haskell-tools";
```
MattSturgeon: Can we avoid the `rec` if possible? They look neat and elegant at first, until you try to read existing code that uses `with` or `rec` (without an LSP).

I don't mind them when applied to a small scope/value, but across an entire file they aren't ideal.

```suggestion
let
  name = "haskell-tools";
in
lib.nixvim.plugins.mkNeovimPlugin {
  inherit name;
```

You could use `lib.fix` here, but that feels over-engineered for this case:

```nix
lib.nixvim.plugins.mkNeovimPlugin (lib.fix (self: {
```

khaneliman:
> Lazyloading with Telescope integration is broken, but that's a problem with our telescope module not accounting for lazyloading and loading every extension at one point. This plugin is lazy by design so there is no real need to lazyload it manually.

Yeah... in a past life I meant to break that all up for lazy loading purposes...

MattSturgeon (changes requested):
Very minor, but I'd like to avoid introducing new cases of antipatterns like overuse of `rec`

plugins/by-name/haskell-tools/default.nix
saygo-png: No problem. I like recursion but don't mind changing it.

PR #3733, opened 2025-09-28
plugins.lsp: refactor to use the new lsp module

MattSturgeon:
- **modules/lsp: port `packageFallback` option from plugins.lsp**
  From #3730

- **plugins/lsp: use the new lsp module under the hood**
  Meat of the PR

- **plugins/lsp: simplify `extraSettings` deprecation**
  Minor cleanup


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -51,9 +51,41 @@ in
...
+                if enabled then
+                  pkg
+                else if self.isDefined then
+                  v
+                else
+                  opt.default or v;
+            }
+            // lib.optionalAttrs (opt ? example) { inherit (opt) example; }
+            // lib.optionalAttrs (opt ? defaultText) { inherit (opt) defaultText; }
+          )
```
MattSturgeon, line 81: Not having a default here was intentional, as it simplifies the other side of the alias (below in `config`).

However, it seems to be causing issues with other parts of the codebase that assume they can get a default from package options. E.g. when we look for the LSP server's homepage URL in the docs, or in tests like `all-package-defaults`.

modules/lsp/servers/default.nix
```diff
@@ -156,13 +157,37 @@ in
             let
               luaName = toLuaObject server.name;
               luaSettings = toLuaObject server.settings;
+              wrap = server.__settingsWrapper or lib.id;
```
GaetanLepage, line 160: The `server.__settingsWrapper` defaults to `lib.id`, so why do you need to `or lib.id` here?

MattSturgeon: The `__settingsWrapper` option isn't declared for the special `"*"` server-config

MattSturgeon: Another way we could handle this would be to remove the default and declare the option externally from `plugins/lsp/servers/_mk-lsp.nix`, since that defines a module that gets merged in. That may be neater, as the option would only exist when using the old API. I'll experiment locally and see which makes the most sense.

MattSturgeon: [Done](https://github.com/nix-community/nixvim/compare/41835f5407a958cb941e998133959c8fa6dd41f4..abb6f0d4c410a1645961a97e47c46de8653de235?w=1)

GaetanLepage (approved):

PR #3735, opened 2025-09-29
flake: add nixf-diagnose to treefmt config

MattSturgeon:
Checks nixd's diagnostic lints using libnixf.

We already have statix, but nixd (libnixf) has some additional linting rules that are worth checking too.

Inspired by https://github.com/NixOS/nixpkgs/pull/438559

### DIFF
--- a/plugins/by-name/preview/default.nix
+++ b/plugins/by-name/preview/default.nix
@@ -1,9 +1,4 @@
-{
-  lib,
-  helpers,
-  ...
-}:
-with lib;
+{ lib, ... }:
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "preview";
   package = "Preview-nvim";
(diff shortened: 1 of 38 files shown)


plugins/by-name/preview/default.nix
```diff
@@ -1,15 +1,10 @@
-{
-  lib,
-  helpers,
-  ...
-}:
-with lib;
```
347Online, line 6: Out of curiosity, was there any particular reason to remove some `with lib;`s but not others? Or did the removed ones just become unused?

FKouhai: lib is being imported directly, I assume that allows us to get rid of `with lib;`

MattSturgeon: > Out of curiosity, was there any particular reason to remove some `with lib;`s but not others? Or did the removed ones just become unused?

As I was going through the files to fix nixf linting issues, I occasionally spotted `with lib`s that were trivial to remove, i.e. only had a single usage. If I noticed a `with` that could be easily removed, I did it. If it would require a larger patch, I skipped it.

Most (all?) of the `with lib`s that I removed were only used for the maintainers entry:

```diff
- maintainers.foo
+ lib.maintainers.foo
```

khaneliman (approved):

PR #3736, opened 2025-09-29
plugins/marks: migrate to mkNeovimPlugin

saygo-png:
Intentionally not deprecating the "bookmarks" option because unless I'm mistaken it never worked.
The assertion for it requires names of the attrset to be integers, but names in Nix can only be strings so it always fails.

### DIFF
--- a/plugins/by-name/marks/default.nix
+++ b/plugins/by-name/marks/default.nix
@@ -1,188 +1,24 @@
-{
-  lib,
-  helpers,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
-let
-  cfg = config.plugins.marks;
-in
-{
-  meta.maintainers = [ maintainers.GaetanLepage ];
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "marks";
+  package = "marks-nvim";
+  description = "A better user experience for viewing and interacting with Neovim marks.";
+  maintainers = [ ];
 
-  options.plugins.marks = lib.nixvim.plugins.neovim.extraOptionsOptions // {
-    enable = mkEnableOption "marks.nvim";
-
-    package = lib.mkPackageOption pkgs "marks.nvim" {
-      default = [
-        "vimPlugins"
-        "marks-nvim"
-      ];
+  settingsExample = {
+    cyclic = true;
+    refreshInterval = 150;
+    mappings = {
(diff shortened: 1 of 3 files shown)


GaetanLepage:
> Intentionally not deprecating the "bookmarks" option because unless I'm mistaken it never worked.

Sure, but you then have to add:
```nix
imports = [
  (lib.mkRemovedOptionModule [ "plugins" "mark" "bookmarks" ] ''
    <INSERT REASON/HELP FOR THIS REMOVAL>
  '')
];
```

plugins/by-name/marks/default.nix
```diff
@@ -1,188 +1,24 @@
...
-      Custom mappings.
-      Set a mapping to `false` to disable it.
-    '';
   };
 
-  config = mkIf cfg.enable {
-    extraPlugins = [ cfg.package ];
-
-    assertions = lib.nixvim.mkAssertions "plugins.mark" {
-      assertion = all (n: elem n (range 0 9)) (attrNames cfg.bookmarks);
```
MattSturgeon, line 156: I guess this was _supposed_ to be `(n: elem n (map toString (range 0 9)))`. Good spot that this was never working.

plugins/by-name/marks/deprecations.nix
```diff
@@ -0,0 +1,22 @@
...
+    "refreshInterval"
+    "signPriority"
+    "excludedFiletypes"
+    "excludedBuftypes"
+    "mappings"
+  ];
+
+  imports = [
+    (lib.mkRemovedOptionModule [ "plugins" "marks" "bookmarks" ] ''
+      This option always caused a failed assertion, it has never worked.
```
MattSturgeon, line 19: Should this error hint at users that they can define `settings.bookmark_<num>` (e.g. `settings.bookmark_1`)?

It's good to point out the option never worked, but it's even better to explain _how_ to make it work with the new system 😁

saygo-png: I'll add it though i'm not sure who will benefit from that. No one uses this option because it always throws eval. I don't think anyone will see this message unless im missing some case.

MattSturgeon: Yeah, that makes sense. You're right: by definition no one must be using this option.

MattSturgeon (approved):

PR #3737, opened 2025-09-29
modules/lsp: use a for loop to create LSP keymaps

GaetanLepage:
This makes the generated `init.lua` shorter.

### DIFF
--- a/modules/lsp/keymaps.nix
+++ b/modules/lsp/keymaps.nix
@@ -104,16 +104,16 @@ in
         group = "nixvim_lsp_binds";
         callback = lib.nixvim.mkRaw ''
           function(args)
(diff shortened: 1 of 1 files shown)


MattSturgeon:
GitHub mobile is having a meltdown posting review comments, but I think you need `toLuaObject` when serialising the keymaps.

Slightly concerning that the test suite doesn't pick up this issue

GaetanLepage:
> GitHub mobile is having a meltdown posting review comments, but I think you need `toLuaObject` when serialising the keymaps.
> 
> Slightly concerning that the test suite doesn't pick up this issue

Good catch. Done.

MattSturgeon (approved):
Thanks

PR #3739, opened 2025-09-30
modules/lsp: test lsp keymaps & cleanup lua table

MattSturgeon:
- **modules/test: add `extraInputs` option**
- **tests/modules/lsp: test lsp keymaps**
- **modules/lsp: print keymaps table multiline**
- **modules/lsp: select relevant fields in keymaps table**

Follow up to #3737, this PR adds a test to assert the expected autoCmd is created for LSP keymaps. Follow-up commits then tweak the output to be more readable by printing a multiline table and omitting irrelevant fields.

My nvim knowledge is too weak to come up with a way to verify the lua code at runtime, which would be nice for catching syntax errors. The issue is the lua code only runs when the `LspAttach` autocmd event is triggered, and I don't know how to do that in our headless test environment. For now, I just set `test.runNvim = false`, but any ideas for a better solution would be great. EDIT: [done](https://github.com/nix-community/nixvim/compare/e2e2b3f1825600bc4acf41200ebbbcd7a8d8cc78..68b63ac2694d2fd91cc763d5062abea473c0613a).

### DIFF
--- a/tests/test-sources/modules/lsp.nix
+++ b/tests/test-sources/modules/lsp.nix
@@ -29,6 +29,123 @@
     };
   };
 
+  keymaps =
+    {
+      lib,
+      pkgs,
+      config,
+      ...
+    }:
+    let
+      autoCmds = config.autoCmd;
+      autoCmd = builtins.head autoCmds;
+
+      print = lib.generators.toPretty { };
+      expect = name: expected: actual: {
+        assertion = expected == actual;
+        message = "Expected ${name} to be ${print expected}, but found ${print actual}";
+      };
+    in
+    {
+      lsp.keymaps = [
+        {
+          key = "gd";
+          lspBufAction = "definition";
+        }
+        {
+          key = "K";
+          lspBufAction = "hover";
+        }
+        {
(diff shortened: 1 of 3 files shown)


GaetanLepage (approved):
Looks good. Thanks for tackling this.

tests/test-sources/modules/lsp.nix
```diff
@@ -29,6 +29,104 @@
...
+                )
+                vim.keymap.set(keymap.mode, keymap.key, keymap.action, options)
+              end
+            end
+          '';
+        })
+      ];
+
+      # TODO: how to test at runtime when the keymaps are not registered until LspAttach?
+      test.runNvim = false;
```
MattSturgeon: I've come up with something locally, pushing now: [diff](https://github.com/nix-community/nixvim/compare/e2e2b3f1825600bc4acf41200ebbbcd7a8d8cc78..68b63ac2694d2fd91cc763d5062abea473c0613a)

GaetanLepage (approved):
Even better.

PR #3740, opened 2025-09-30
plugins/nvim-notify: add more render styles

auipga:
one render style was still missing


MattSturgeon (commented):
Is there a `wrapped-minimal` and/or `wrapped-simple`?

khaneliman (approved):
LGTM https://github.com/rcarriga/nvim-notify/blob/8701bece920b38ea289b457f902e2ad184131a5d/lua/notify/render/init.lua#L9-L15 looks like it was added in july last year https://github.com/rcarriga/nvim-notify/pull/286

Issue #3742, opened 2025-09-30
[PLUGIN REQUEST] sidekick


storopoli:
### Plugin

`sidekick.nvim`

### Homepage

https://github.com/folke/sidekick.nvim

### Extra Information

New plugin by folke to be state of the art AI-assisted coding in neovim.
https://github.com/NixOS/nixpkgs/pull/446983 was merged today but still has not yet hit `nixpkgs` tracking at <https://nixpk.gs/pr-tracker.html?pr=446983>

khaneliman:
I was actually working on this right now.

PR #3743, opened 2025-09-30
plugins/sidekick: init module

khaneliman:
Adding support for new folke plugin https://github.com/folke/sidekick.nvim
Some of the latest functionality needs a bump, but this is a freeform settings module with cli dependencies that come with it for using functionality. 

Closes https://github.com/nix-community/nixvim/issues/3742
Created https://github.com/NixOS/nixpkgs/pull/447571 for bumping to latest functionality but depends on https://github.com/NixOS/nixpkgs/pull/446983 making it to our flake.lock.

### DIFF
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -1,0 +1,35 @@
+{
+  lib,
+  config,
+  options,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "sidekick";
+  package = "sidekick-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  dependencies = [
+    "claude-code"
+    "copilot"
+    "gemini"
+    "opencode"
+  ];
+
+  extraConfig = {
+    assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
+      assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
+      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
+    };
+  };
+
+  settingsExample = {
+    cli = {
+      mux = {
+        backend = "zellij";
+        enabled = true;
+      };
+    };
+  };
+}
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -100,6 +100,7 @@ in
       bat.default = "bat";
       claude-code.default = "claude-code";
       codeium.default = "codeium";
+      copilot.default = "github-copilot-cli";
       coreutils = {
         default = "coreutils";
         example = "uutils-coreutils";
--- a/tests/test-sources/plugins/by-name/sidekick/default.nix
+++ b/tests/test-sources/plugins/by-name/sidekick/default.nix
@@ -1,0 +1,169 @@
+{
+  empty = {
+    plugins.copilot-lua.enable = true;
+    plugins.sidekick.enable = true;
+  };
+
+  defaults = {
+    plugins.copilot-lua.enable = true;
+    plugins.sidekick = {
+      enable = true;
+      settings = {
+        jump = {
+          jumplist = true;
+        };
+        signs = {
+          enabled = true;
+          icon = " ";
+        };
+        nes = {
+          enabled.__raw = ''
+            function(buf)
+              return vim.g.sidekick_nes ~= false and vim.b.sidekick_nes ~= false
+            end'';
+          debounce = 100;
+          trigger = {
+            events = [
+              "InsertLeave"
+              "TextChanged"
+              "User SidekickNesDone"
+            ];
+          };
+          clear = {
+            events = [
+              "TextChangedI"
+              "BufWritePre"
+              "InsertEnter"
+            ];
+            esc = true;
+          };
+          diff = {
+            inline = "words";
+          };
+        };
+        cli = {
+          watch = true;
+          win = {
+            wo = { };
+            bo = { };
+            width = 80;
+            height = 20;
+            layout = "vertical";
+            position = "right";
+            keys = {
+              stopinsert = [
+                "<esc><esc>"
+                "stopinsert"
+                { mode = "t"; }
+              ];
+              hide_n = [
+                "q"
+                "hide"
+                { mode = "n"; }
+              ];
+              hide_t = [
+                "<c-q>"
+                "hide"
+              ];
+              win_p = [
+                "<c-w>p"
+                "blur"
+              ];
+              blur = [
+                "<c-o>"
+                "blur"
+              ];
+              prompt = [
+                "<c-p>"
+                "prompt"
+              ];
+            };
+          };
+        };
+        mux = {
+          backend = "zellij";
+          enabled = false;
+        };
+        tools = {
+          aider = {
+            cmd = [ "aider" ];
+            url = "https://github.com/Aider-AI/aider";
+          };
+          amazon_q = {
+            cmd = [ "q" ];
+            url = "https://github.com/aws/amazon-q-developer-cli";
+          };
+          claude = {
+            cmd = [ "claude" ];
+            url = "https://github.com/anthropics/claude-code";
+          };
+          codex = {
+            cmd = [
+              "codex"
+              "--search"
+            ];
+            url = "https://github.com/openai/codex";
+          };
+          copilot = {
+            cmd = [
+              "copilot"
+              "--banner"
+            ];
+            url = "https://github.com/github/copilot-cli";
+          };
+          cursor = {
+            cmd = [ "cursor-agent" ];
+            url = "https://cursor.com/cli";
+          };
+          gemini = {
+            cmd = [ "gemini" ];
+            url = "https://github.com/google-gemini/gemini-cli";
+          };
+          grok = {
+            cmd = [ "grok" ];
+            url = "https://github.com/superagent-ai/grok-cli";
+          };
+          opencode = {
+            cmd = [ "opencode" ];
+            url = "https://github.com/sst/opencode";
+          };
+          qwen = {
+            cmd = [ "qwen" ];
+            url = "https://github.com/QwenLM/qwen-code";
+          };
+        };
+        prompts = {
+          explain = "Explain this code";
+          diagnostics = {
+            msg = "What do the diagnostics in this file mean?";
+            diagnostics = true;
+          };
+          diagnostics_all = {
+            msg = "Can you help me fix these issues?";
+            diagnostics = {
+              all = true;
+            };
+          };
+          fix = {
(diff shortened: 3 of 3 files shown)


plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,36 @@
...
+  lib,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "sidekick";
+  package = "sidekick-nvim";
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  dependencies = [
```
khaneliman, line 13: Unfortunately, don't see `copilot-cli` from github packaged in nixpkgs.

khaneliman:
Marking draft until we can merge.

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,53 @@
...
+          Please enable it by setting `plugins.copilot-lua.enable = true`.
+        '';
+      }
+    ];
+
+    # TODO: Add copilot-language-server support to main module.
+    lsp.servers.copilot = {
+      enable = true;
+
+      settings.settings = {
```
storopoli: Should this be duplicated? `settings.settings`?

khaneliman: Yes, see https://github.com/nix-community/nixvim/issues/3745

storopoli: I see, it was very awkward. Solving this.

plugins/by-name/sidekick/default.nix
jolars: It's on its way: https://github.com/NixOS/nixpkgs/pull/446946

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,53 @@
...
+        message = ''
+          Sidekick requires `copilot-lua` to be enabled.
+          Please enable it by setting `plugins.copilot-lua.enable = true`.
+        '';
+      }
+    ];
+
+    # TODO: Add copilot-language-server support to main module.
+    lsp.servers.copilot = {
+      enable = true;
```
jolars: I don't think you're supposed to add both copilot.lua and the lsp server.

And we may want to check if copilot.lua or copilot.vim are already enabled, because they both bundle their own versions of the copilot LSP. If someone has copilot.lua enabled already, for instance, I think you would get two versions of the LSP running simultaneously.

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,53 @@
...
+
+      settings.settings = {
+        "copilot" = {
```
jolars: Hm, maybe I'm mistaken, but shouldn't this set `root_markers` too? Maybe it defaults to something reasonable, like `.git`. The lspconfig version also has a custom command setup to authenticate copilot.

Here's my config, which I took directly from lspconfig:

```nix
    lsp = {
      servers = {
        copilot = {
          enable = true;

          package = pkgs.copilot-language-server;

          settings = {
            cmd = [
              "${pkgs.copilot-language-server}/bin/copilot-language-server"
              "--stdio"
            ];

            root_markers = [ ".git" ];

            on_attach.__raw = ''
              function(client, bufnr)
                vim.api.nvim_buf_create_user_command(bufnr, 'LspCopilotSignIn', function()
                  sign_in(bufnr, client)
                end, { desc = 'Sign in Copilot with GitHub' })
                vim.api.nvim_buf_create_user_command(bufnr, 'LspCopilotSignOut', function()
                  sign_out(bufnr, client)
                end, { desc = 'Sign out Copilot with GitHub' })
              end
            '';

            settings = {
              telemetry = {
                telemetryLevel = "all";
              };
            };
          };
        };
      };
    };

```

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,53 @@
...
+          cmd = [
+            "${lib.getExe pkgs.copilot-language-server}"
```
jolars: I needed to use `pkgs.copilot-language-server-fhs` to escape this problem: https://github.com/NixOS/nixpkgs/issues/391730. Maybe we shouldn't hardcode the package here?

jolars: Nevermind, this has been renamed to be copilot-language-server now.

plugins/by-name/sidekick/default.nix
khaneliman: Check health failed without configuring the lsp explicitly.

plugins/by-name/sidekick/default.nix
khaneliman: We already configure root_markers for all servers in nixvim.

plugins/by-name/sidekick/default.nix
khaneliman: So after reading up a little more... I think you're right.  It seems that a lot of the stuff that got it working for me is coming from `lspconfig`.. I don't want to turn this module into a duplication of `lspconfig`, necessarily...

plugins/by-name/sidekick/default.nix
jolars: But is copilot included in nixvim as an lsp server? I have been trying to run the CI scripts to update this list, but it doesn't seem to identify copilot for some reason. I posted about this in the Development channel on Matrix.

plugins/by-name/sidekick/default.nix
jolars: I think this was a bug in sidekick and I saw some commits related to this in the repo. I also saw that error before, but now that I updated to `52a6ed40d312726a45ffc191fdc81791c4d928f5` it says

> `- ✅ OK Using `copilot.lua`'s bundled LSP server`

khaneliman: Oh, let me try disabling this lsp manual configuration and just enabling copilot.lua again. 

>Sidekick Copilot LSP ~
- ❌ ERROR No Copilot LSP server is enabled with `vim.lsp.enable(...)`


Still seem to get it on my end using https://github.com/NixOS/nixpkgs/pull/447993/files#diff-54bef7a6c9c5497e77086f7935fbfe5c45e285f36f12f15b7044f42430e8e3ebR13708-R13716

plugins/by-name/sidekick/default.nix
khaneliman: > But is copilot included in nixvim as an lsp server? I have been trying to run the CI scripts to update this list, but it doesn't seem to identify copilot for some reason. I posted about this in the Development channel on Matrix.

Not currently but, Matt has a PR that would pick it up https://github.com/nix-community/nixvim/pull/3748

We use a `*` wildcard, so it applies to all servers, though.

plugins/by-name/sidekick/default.nix
jolars: Ah, yes, I think it just changed today. I am using:

```nix
  programs.nixvim = {
    extraPlugins = [
      (pkgs.vimUtils.buildVimPlugin {
        name = "sidekick-nvim";
        src = pkgs.fetchFromGitHub {
          owner = "folke";
          repo = "sidekick.nvim";
          rev = "52a6ed40d312726a45ffc191fdc81791c4d928f5";
          hash = "sha256-+uChBKOYqxhrk3NghcX/PsHXYRgI1lJaLbEFOW9mgVg=";
        };
        nvimSkipModules = [
          "sidekick.docs"
        ];
      })
    ];
  };
```

khaneliman: I'm running an overlay of my `vimPlugins` set bump... but I'm thinking its an issue with my `copilot-lua` configuration, then. 

If I try enabling `nes` on `copilot-lua` I see this, even though I've configured it. 
```bash
[Copilot.lua] copilot-lsp is not available, disabling nes.
```

I had to fix it by adding `pkgs.vimPlugins.copilot-lsp` when using `copilot-lua`. Then, sidekick showed a successful health check for me. This seems like a miss in our `copilot-lua` module, imo. I'm going to add that as a package we install in there for allowing the next edit suggestion functionality. 

But, then it seems like we can either choose to configure the `lsp` explicitly or encourage using `copilot-lua` with the package included there.

jolars: But are you sure the copilot-lsp plugin is actually necessary for NES via sidekick? The health heck is passing for me without it and the docs do not mention that it's necessary.

I think a good default would be to default to enabling the copilot lsp unless copilot.lua is enabled.

khaneliman: Wasn't able to get it working locally without it. I tried a few different variations

- copilot-lua disabled - copilot-lsp missing - lsp.servers.copilot disabled - doesn't pass healthcheck
- copilot-lua enabled - no copilot-lsp - doesn't pass healthcheck
- copilot-lua enabled - with copilot-lsp - passes healthcheck
- copilot-lua disabled -  lsp.servers.copilot enabled - passes healthcheck

khaneliman: I finally realized what was causing the issue. I had `server` configured in my copilot configuration, which actually broke it... 

```nix
        server = {
          type = "binary";
          custom_server_filepath = lib.getExe pkgs.copilot-language-server;
        };
```

After removing that, I could see the: 

```bash 
Sidekick Copilot LSP ~
- ✅ OK Using `copilot.lua`'s bundled LSP server
- ✅ OK Sidekick is handling Copilot LSP status notifications for client: 5
```

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  dependencies = [
+    "claude-code"
+    "copilot"
+    "gemini"
+    "opencode"
+  ];
+
+  extraConfig = {
+    assertions = [
```
MattSturgeon: Currently, we're using `mkAssertions` for consistency.

tbh we should probably simplify by adding an `assertions` attr directly to mkNeovimPlugin, at some point.

plugins/by-name/sidekick/default.nix
```diff
@@ -0,0 +1,36 @@
...
+      {
+        assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
+        message = "sidekick requires either copilot-lua (plugins.copilot-lua.enable) or copilot LSP (lsp.servers.copilot.enable) to be enabled";
```
MattSturgeon: Nit:

I'm starting to prefer interpolating options so users get the full path (e.g. including `programs.nixvim.` if appropriate):


```suggestion
        message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
```

MattSturgeon (commented):
SGTM

plugins/by-name/sidekick/default.nix
khaneliman: Sorry, forgot that was introduced.. been a while.

MattSturgeon (approved):
Thanks!

PR #3749, opened 2025-10-01
plugins/rainbow-delimiters: migrate to mkNeovimPlugin

saygo-png:

Things of note:

I'm deprecating "default" for `""` as i think that was a hack for us not converting "" to proper lua.
Sadly i'm not sure how I can shove this deprecation into `deprecations.nix`

Keeping the old strategies option overall because I think it's nice to make users not write Lua for that.

Users can define both `strategies` and `settings.strategies` which leads to mkMerge errors. I think that's okay and is clear enough to the user.

### DIFF
--- a/plugins/by-name/rainbow-delimiters/default.nix
+++ b/plugins/by-name/rainbow-delimiters/default.nix
@@ -1,24 +1,30 @@
 {
   lib,
-  helpers,
   config,
-  pkgs,
+  options,
   ...
 }:
-with lib;
-{
-  options.plugins.rainbow-delimiters = lib.nixvim.plugins.neovim.extraOptionsOptions // {
-    enable = mkEnableOption "rainbow-delimiters.nvim";
+let
+  opts = options.plugins.rainbow-delimiters;
+  inherit (builtins) any isNull;
+  inherit (lib) mapAttrs' nameValuePair isString;
+  inherit (lib.nixvim) mkRaw toLuaObject nestedLiteralLua;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "rainbow-delimiters";
+  package = "rainbow-delimiters-nvim";
+  description = "Rainbow delimiters for Neovim with Tree-sitter";
+  maintainers = [ ];
 
-    package = lib.mkPackageOption pkgs "rainbow-delimiters.nvim" {
-      default = [
-        "vimPlugins"
-        "rainbow-delimiters-nvim"
-      ];
-    };
+  # This plugin uses globals for configuration.
+  callSetup = false;
 
-    strategy = helpers.defaultNullOpts.mkAttrsOf' {
-      type = types.enum [
+  # TODO: introduced 2025-10-1: remove after 26.05
+  inherit (import ./deprecations.nix lib) deprecateExtraOptions optionsRenamedToSettings imports;
+
+  extraOptions = {
+    strategy = lib.nixvim.defaultNullOpts.mkAttrsOf' {
+      type = lib.types.enum [
         "global"
         "local"
         "noop"
@@ -29,12 +35,13 @@ with lib;
       description = ''
         Attrs mapping Tree-sitter language names to strategies.
         See `|rb-delimiters-strategy|` for more information about strategies.
+        This option adds definitions to `${opts.settings}.strategy` wrapped with required Lua code.
       '';
-      example = literalMD ''
+      example = lib.literalMD ''
         ```nix
         {
           # Use global strategy by default
-          default = "global";
+          "" = "global";
 
           # Use local for HTML
           html = "local";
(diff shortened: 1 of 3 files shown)


plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
+      }
 
+      # TODO: introduced 2025-10-1: remove after 26.05
+      {
+        when = elem "default" (lib.attrNames (if cfg.strategy == null then { } else cfg.strategy));
         message = ''
-          Both `rainbow-delimiters.whitelist` and `rainbow-delimiters.blacklist` should not be
-          set simultaneously.
-          Please remove one of them.
+          Setting the keys to `default` in `strategy` is deprecated.
```
MattSturgeon: Nit:

```suggestion
          Setting `${opts.strategy}` keys to the string `"default"` is deprecated.
```

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
-        }
-        // cfg.extraOptions;
+    assertions = lib.nixvim.mkAssertions "plugins.rainbow-delimiters" {
+      assertion = any isNull [
+        (cfg.settings.whitelist or null)
+        (cfg.settings.blacklist or null)
+      ];
+      message = ''
+        Both `rainbow-delimiters.settings.whitelist` and `rainbow-delimiters.settings.blacklist` should not be
+        set simultaneously.
```
MattSturgeon: No need for line wrapping here, as this will print to stderr as-is.

Also, we can take advantage of option-string-interpolation here:

```suggestion
        Both `${opts.settings}.whitelist` and `${opts.settings}.blacklist` should not be set simultaneously.
```

Also, something feels off about the phrasing. Not critical, and I can't quite pinpoint what it is anyway.

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
+    globals.rainbow_delimiters = lib.mkMerge [
+      cfg.settings
+      (lib.mkIf (cfg.strategy != null) {
+        strategy = mapAttrs' (
+          n: v:
+          nameValuePair (if n == "default" then "" else n) (
+            if isString v then mkRaw "require 'rainbow-delimiters'.strategy['${v}']" else v
+          )
+        ) cfg.strategy;
+      })
```
MattSturgeon, line 125: > Users can define both `strategies` and `settings.strategies` which leads to mkMerge errors. I think that's okay and is clear enough to the user.

This should merge fine, IIUC. You should only see merge conflicts if you try to define the same key in both `strategy` and `settings.strategy` with different values.


Maybe it's worth having a test case for defining `strategy.foo` and `settings.strategy.bar`, just to be sure? Such a test could set `test.buildNixvim = false`.

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
+      '';
     };
+
+    globals.rainbow_delimiters = lib.mkMerge [
+      cfg.settings
+      (lib.mkIf (cfg.strategy != null) {
+        strategy = mapAttrs' (
+          n: v:
+          nameValuePair (if n == "default" then "" else n) (
+            if isString v then mkRaw "require 'rainbow-delimiters'.strategy['${v}']" else v
```
MattSturgeon: Nit1: lua noobs like me usually find this style more readable:

```suggestion
            if isString v then mkRaw "require('rainbow-delimiters').strategy['${v}']" else v
```

Nit2: we should use `toLuaObject` for serialising values, even strings, to avoid accidental string escape issues. In this case that means use `toLuaObject instead of wrapping with `'` manually:

```suggestion
            if isString v then mkRaw "require 'rainbow-delimiters'.strategy[${toLuaObject v}]" else v
```

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
-      level = helpers.defaultNullOpts.mkLogLevel "warn" ''
-        Only messages equal to or above this value will be logged.
-        The default is to log warnings or above.
-        See `|log_levels|` for possible values.
-      '';
+  settingsExample = {
+    blacklist = [ "json" ];
+    strategy = {
+      "".__raw = "require 'rainbow-delimiters'.strategy['global']";
+      "nix".__raw = "require 'rainbow-delimiters'.strategy['local']";
```
MattSturgeon: You may wish to have a `lib.literalExpression` example here, to avoid awkward rendering in the docs.

Or alternatively use our `lib.nixvim.nestedLiteralLua` function instead of `__raw`.

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,60 @@ with lib;
...
+    ];
   };
 
-  config =
-    let
-      cfg = config.plugins.rainbow-delimiters;
-    in
-    mkIf cfg.enable {
-      warnings = lib.nixvim.mkWarnings "plugins.rainbow-delimiters" {
+  extraConfig = cfg: {
```
MattSturgeon, line 81: If following my suggestions to use option string-interpolation in warnings:
```suggestion
  extraConfig = cfg: opts: {
```

plugins/by-name/rainbow-delimiters/deprecations.nix
```diff
@@ -0,0 +1,40 @@
...
+        (basePathAnd [
+          "settings"
+          "query"
+        ])
+        (
+          config:
+          let
+            old = lib.getAttrFromPath oldOptPath config;
+          in
+          ifNonNull' old (mapAttrs' (n: nameValuePair (if n == "default" then "" else n)) old)
```
MattSturgeon, line 36: Should this have a warning, similar to the `cfg.strategy` one?

I guess that's hard to implement, because I assume _reading_ the old option's value in the warning will trigger _this_ option's deprecation warning?

EDIT: this is probably a bug in `mkMergedOptionModule`, but it [_doesn't_](https://github.com/NixOS/nixpkgs/blob/deff17b7b414e449d138598b616643f5ef072b3d/lib/modules.nix#L1697-L1701) add the `trace` on `apply` warning that `doRename` [does](https://github.com/NixOS/nixpkgs/blob/deff17b7b414e449d138598b616643f5ef072b3d/lib/modules.nix#L1912) _(actually [here](https://github.com/NixOS/nixpkgs/blob/deff17b7b414e449d138598b616643f5ef072b3d/lib/modules.nix#L1621))_... Therefore you _can_ read the old option value without triggering any warnings...

MattSturgeon (commented):
Looks good. I can see lots of thought and effort in the deprecation, which is appreciated.

Sorry it's taken a while to review; I've looked at it a few times but until now I haven't had time to fully digest the plugin and the changes.

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,59 @@ with lib;
...
         when = !config.plugins.treesitter.enable;
         message = "This plugin needs treesitter to function as intended.";
-      };
-      assertions = lib.nixvim.mkAssertions "plugins.rainbow-delimiters" {
-        assertion = (cfg.whitelist == null) || (cfg.blacklist == null);
+      }
 
+      # TODO: introduced 2025-10-1: remove after 26.05
+      {
+        when = elem "default" (lib.attrNames (if cfg.strategy == null then { } else cfg.strategy));
```
MattSturgeon: `elem "x" (attrNames foo)` is just `foo ? x`:

```suggestion
        when = cfg.strategy ? default;
```

I'm pretty sure `?` supports anything on the lhs (even `null`), but if not you could also do:

```suggestion
        when = cfg ? strategy.default;
```

plugins/by-name/rainbow-delimiters/default.nix
saygo-png: Sorry it's me being unclear again. Yes it merges fine if the keys are different. I don't think a test is necessary since this is standard behavior just my wording was unclear. I do still think that there might be a potential for confusion here? Since there are 2 ways to set strategies — not sure where or if I should clarify the difference.

addendum:
Since I started contributing it's been surprisingly difficult to explain myself clearly. Part of that is surely my lack of people skills but also inexperience in both coding and open source (I've only really started coding this year and this is my first project i contribute to regularly). Often I'm torn between making my messages wordy but accurate or brief but inaccurate.

MattSturgeon: Don't worry, you're doing fine. Communication is a universal challenge. There will always be points of clarification and misunderstandings.

I'm usually guilty of elaborating too much, which just results in people skim reading my message; therefore typing it out was a waste of time 😅

> Since there are 2 ways to set strategies — not sure where or if I should clarify the difference.

Perhaps the abstraction option's description should mention that it will add definitions to `settings.strategy`.

plugins/by-name/rainbow-delimiters/deprecations.nix
saygo-png: If this is a bug this might be relying on it, but it will be caught by our tests if it changes.

plugins/by-name/rainbow-delimiters/default.nix
```diff
@@ -56,102 +60,66 @@ with lib;
...
+          You can set them to an empty string `""` instead.
         '';
-      };
-
-      extraPlugins = [ cfg.package ];
+      }
+      {
+        when = (cfg.query ? default) && (cfg.query.default != "_mkMergedOptionModule");
+        message = ''
+          Setting `${opts.strategy}` keys to the string `"default"` is deprecated.
```
MattSturgeon:
```suggestion
          Setting `${opts.query}` keys to the string `"default"` is deprecated.
```

MattSturgeon (approved):
Thanks again!

PR #3748, opened 2025-09-30
ci/nvim-lspconfig: migrate to the new API

MattSturgeon:
Migrates the nvim-lspconfig "update script" to use the new `lsp/` directory-based system. Previously, the update script looked for server configs in the `lua/lspconfig/configs` module, however upstream have frozen and deprecated that directory. New LSP servers are only added to the new directory.

As a result, the new update script detects **26 new LSP servers** that we were previously missing out on.

<details><summary><strong>Newly added servers</strong></summary>
<p>

- atopile
- cir_lsp_server
- clarinet
- codebook
- copilot
- cspell_ls
- docker_language_server
- emmylua_ls
- expert
- gnls
- herb_ls
- kotlin_lsp
- phptools
- pli
- pyrefly
- roslyn_ls
- sqruff
- stylua
- termux_language_server
- tofu_ls
- tombi
- tsgo
- ty
- vespa_ls
- vue_ls
- zuban

</p>
</details> 

As part of the refactor, I've done a cleanroom rewrite of the update script, reusing only minimal parts, such as the pandoc "description" filter.

The generated file is now a dictionary (server names -> lspconfig description) instead of a list.

TODO:

- [x] Check how tightly-coupled the lspconfig descriptions are to nvim-lspconfig
  - If too tightly-coupled, we probably don't _want_ the descriptions in the new module
  - Conclusion: yes, they're tightly-coupled. We **don't** want them in the new module.
- [x] Port the `clean-desc` pandoc filter to the new update script
- [x] Add a new file that lists the disjoint between the new & old generated files
  - This can be used to `mkRemoveOptionModule` any unsupported lsp servers
  - Done in #3750 
- [x] Adapt usage of the generated file to match it's new scheme (in tests, `lsp`, `plugins.lsp`)
- [ ] Use the new `errors` attribute
  - The new update script captures errors printed by lspconfig when evaluating the server config
  - This can be used in a future PR to add warnings when users enable a deprecated server
  - and/or avoid creating explicit options for deprecated servers in the new freeform `lsp.servers` option
  - E.g. to handle issues like #3600 
  - NOTE: this feature was dropped when rebasing on #3761 
- [x] Update the `lsp-packages` list to match the updated file
  - The new file contains 26 servers we weren't picking up before 
- [ ] Move the `lsp-packages` file out of the `plugins/lsp` namespace
  - It is used by both `lsp` and `plugins.lsp`

PR #3750, opened 2025-10-01
plugins.lsp: automatically remove unsupported servers

MattSturgeon:
In #3733 we refactored `plugins.lsp` to use the new `lsp` module under the hood. This means `plugins.lsp` switched from using nvim-lspconfig's deprecated `require('lspconfig')` API to neovim's new API.

Unfortunately, not _all_ server configs have been migrated to the new API. We can detect this by reading files in the respective _new_ and _old_ directories. If a server config module exists in the old directory but not the new directory, then it only supports the old API. The reverse is also true for servers only in the new directory, but that's a separate issue we will solve in #3748.

A new update script will check which "old" files do not have an equivalent "new" file, then the plugins.lsp module will create a removal assertion for any servers that are listed in the generated file.

### DIFF
--- a/generated/unsupported-lspconfig-servers.json
+++ b/generated/unsupported-lspconfig-servers.json
@@ -1,0 +1,16 @@
+[
+  "bitbake_ls",
+  "bqnlsp",
+  "cadence",
+  "clarity_lsp",
+  "codeqlls",
+  "delphi_ls",
+  "drools_lsp",
+  "haxe_language_server",
+  "openedge_ls",
+  "pkgbuild_language_server",
+  "relay_lsp",
+  "sourcery",
+  "vdmj",
+  "vuels"
+]
--- a/ci/generate.nix
+++ b/ci/generate.nix
@@ -51,6 +51,7 @@ writeShellApplication {
 
     generate_json "${conform-formatters}" "conform-formatters"
     generate_json "${lspconfig-servers}" "lspconfig-servers"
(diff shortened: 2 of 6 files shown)


GaetanLepage (approved):

generated/unsupported-lspconfig-servers.json
```diff
@@ -0,0 +1,18 @@
...
+  "apex_ls",
+  "bitbake_ls",
+  "bqnlsp",
+  "cadence",
+  "clarity_lsp",
+  "codeqlls",
+  "delphi_ls",
+  "drools_lsp",
+  "haxe_language_server",
+  "leanls",
```
MattSturgeon:
```suggestion
```

tests/test-sources/plugins/by-name/lean/default.nix
```diff
@@ -6,6 +6,14 @@
...
   # in the `lean` plugin configuration.
   lspDisabled = {
+    # TODO: 2025-10-01: nvim-lspconfig currently doesn't support leanls via the new API
+    test = {
+      buildNixvim = false;
+      assertions = expect: [
+        (expect "count" 1)
+      ];
+    };
+
```
MattSturgeon:
```suggestion
```

MattSturgeon (commented):
leanls will be supported again after #3723 is merged.

MattSturgeon:
> leanls will be supported again after #3723 is merged.

Rebased and re-ran the update script

GaetanLepage:
Thanks @MattSturgeon

PR #3752, opened 2025-10-02
plugins/copilot-lua: add copilot-lsp

khaneliman:
Enables next edit suggestion support. Enabling configuration without plugin will throw an error. Also, will allow sidekick to use copilot-lsp for it's own support. Discovered in https://github.com/nix-community/nixvim/pull/3743#discussion_r2399203655

### DIFF
--- a/plugins/by-name/copilot-lua/default.nix
+++ b/plugins/by-name/copilot-lua/default.nix
@@ -1,4 +1,9 @@
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 let
   inherit (lib) types;
   inherit (lib.nixvim) defaultNullOpts;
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Would it be worth it to add it as a dependency of `vimPlugins.copilot-lua` in nixpkgs?

khaneliman:
> Would it be worth it to add it as a dependency of `vimPlugins.copilot-lua` in nixpkgs?

Not really a hard dependency, though. Just allows you to enable next edit suggestions. Was thinking of adding a module for this and making it something we can toggle the relationship that way, too. Since it has its own configuration we can do.

GaetanLepage:
> > Would it be worth it to add it as a dependency of `vimPlugins.copilot-lua` in nixpkgs?
> 
> Not really a hard dependency, though. Just allows you to enable next edit suggestions. Was thinking of adding a module for this and making it something we can toggle the relationship that way, too. Since it has its own configuration we can do.

I'm fine with merging this as is. Feel free to push the button :)

Issue #3745, opened 2025-09-30
[RFC] Designing the new `lsp` module API


GaetanLepage:
_Authored by @MattSturgeon and @GaetanLepage_

---

We recently introduced the new top-level `lsp` module that aims to replace the existing `plugins.lsp` module (#3203).
It is not obvious how we should represent the `lsp.servers.<name>` nix options that represent the neovim lua [`vim.lsp.Config`](https://neovim.io/doc/user/lsp.html#vim.lsp.Config) options, which get passed to the [`vim.lsp.config()`](https://neovim.io/doc/user/lsp.html#vim.lsp.config()) function.

Old API:
```nix
plugins.lsp = {
  enable = true;
  servers.lua_ls = {
    enable = true;
    package = pkgs.lua-ls;
    
    rootMarkers = [ ".git" ];
    settings.diagnostics.globals = [ "vim" ];
  };
};
```

New API:
```nix
lsp = {
  servers = {
    lua_ls = {
      enable = true;
      package = pkgs.lua-ls;
      
      # How to structure these...
      root_markers = [ ".git" ];
      settings.diagnostics.globals = [ "vim" ];
      #...?
  };
};
```

- [`vim.lsp.config()` function](https://neovim.io/doc/user/lsp.html#vim.lsp.config())
- [`vim.lsp.Config` type](https://neovim.io/doc/user/lsp.html#vim.lsp.Config)

How do we expose LSP 'config' options (`cmd`, `root_markers`, ...)?

### Solution A: _flat_ (directly under `lsp.servers.<name>`)

E.g., include the options directly within our server option:
```nix
lsp.servers = {
  lua_ls = {
    enable = true;
    package = pkgs.lua-ls;
    
    rootMarkers = [ ".git" ];
    settings.diagnostics.globals = [ "vim" ];
};
```

**Pros:**
- Closer to [the current `plugins.lsp.servers.<name>` API](https://nix-community.github.io/nixvim/plugins/lsp/servers/atlas/index.html#atlas). Thus, easier to migrate to.
- Shorter and arguably _more intuitive_ attribute paths

**Cons:**
- Not [RFC-42](https://github.com/NixOS/rfcs/blob/45b76f20add8d7e0e1d0bc0eed9357b067c899a6/rfcs/0042-config-option.md) compliant, with all the drawbacks it implies (less future proof, mixed with Nixvim-specific options like `package` and `enable`)\
[...]

PR #3755, opened 2025-10-03
modules/lsp: rename server `settings` -> `cfg`

MattSturgeon:
This is one possible solution to [the RFC](https://github.com/nix-community/nixvim/issues/3745), other solutions include [renaming back to `config`](https://github.com/nix-community/nixvim/pull/3762), doing nothing, or removing the freeform structured config option and having top-level explicit options instead.

**Please keep discussion on this PR limited to code review, discussion about which solution to choose should happen on the RFC issue.**

The commit message is written as-if we've chosen this solution; once it's merged we have. Until this or an alternative is merged, the RFC is still up for debate.

> While `settings` is the idiomatic name for an RFC42-style freeform structured config option, it is very confusing when `settings` itself has a `settings` sub-option.
>
> `config` is also a poor choice, as it conflicts with module system terminology and requires using `shorthandOnlyDefinesConfig`.
>
> Neovim's `:h vim.lsp.config()` documentation names the parameter `{cfg}`, so we've chosen that at the least-bad option.

Closes #3762

### DIFF
--- a/modules/lsp/servers/default.nix
+++ b/modules/lsp/servers/default.nix
@@ -98,7 +98,7 @@ in
       '';
       default = { };
       example = {
-        "*".settings = {
+        "*".cfg = {
           root_markers = [ ".git" ];
           capabilities.textDocument.semanticTokens = {
             multilineTokenSupport = true;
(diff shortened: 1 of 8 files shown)


FKouhai (approved):
diff LGTM

PR #3760, opened 2025-10-03
plugins/lsp/lsp-packages: add defaults for new LSPs

MattSturgeon:
26 new LSPs were added when we updated the update script in https://github.com/nix-community/nixvim/pull/3748

Add a package association for all the new LSPs I was able to find a package for.

This is a draft until #3748 and #3759 are merged.

### DIFF
--- a/plugins/lsp/lsp-packages.nix
+++ b/plugins/lsp/lsp-packages.nix
@@ -7,7 +7,6 @@
     "antlersls"
     "ansiblels"
     "apex_ls"
-    "atopile"
     "autohotkey_lsp"
     "awk_ls"
     "azure_pipelines_ls"
@@ -28,10 +27,8 @@
     "circom-lsp"
     "clarinet"
     "cobol_ls"
-    "codebook"
     "coffeesense"
     "contextive"
-    "copilot"
     # coqPackages.coq-lsp is unavailable since the bump to coq 9.0: https://github.com/NixOS/nixpkgs/pull/389454
     "coq_lsp"
     "cspell_ls"
@@ -44,12 +41,10 @@
     "dcmls"
     "debputy"
     "djlsp"
-    "docker_language_server"
     "ds_pinyin_lsp"
     "dts_lsp"
     "ecsact"
     "ember"
-    "emmylua_ls"
     "esbonio"
     "expert"
     "facility_language_server"
@@ -124,7 +119,6 @@
     "puppet"
     "purescriptls"
     "pyre"
-    "pyrefly"
     "r_language_server"
     "racket_langserver"
     "raku_navigator"
@@ -135,7 +129,6 @@
     "robotframework_ls"
     "roc_ls"
     "rome"
-    "roslyn_ls"
     "rpmspec"
     "ruff_lsp" # deprecated and removed from nixpkgs
     "salt_ls"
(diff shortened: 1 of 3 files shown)


khaneliman (approved):

plugins/lsp/lsp-packages.nix
```diff
@@ -358,6 +352,7 @@
     rescriptls = "rescript-language-server";
     # This is not entirely true, but the server is deprecated
     rls = "rustup";
+    roslyn_ls = "roslyn-ls";
```
khaneliman, line 355: Need to disable testing this package. It's been broken for a while. 

EDIT: Well, a dotnet vmr dependency has transient errors and is super slow to build so its usually not cached.

MattSturgeon:
I've rebased this now that the other PRs are merged, but I haven't dealt with the CI failures yet.

MattSturgeon:
On darwin we get:
```
Package ‘kicad-base-9.0.4’ in /nix/store/mfcjzs42j5pdamz58q6hhpc23jvs581r-source/pkgs/applications/science/electronics/kicad/base.nix:205 is marked as broken, refusing to evaluate.
```

However `kicad-base` must be a transitive dependency of one of the packages, as it's not in this PR's diff.

plugins/lsp/lsp-packages.nix
MattSturgeon: That's the same issue as this, right?

https://github.com/nix-community/nixvim/blob/b9c5a75cc63a351b5d92a8af5e72841fba0ee383/tests/test-sources/plugins/lsp/omnisharp.nix#L6-L10

MattSturgeon: Hm, actually we're not seeing the same "marked insecure" issue here, we actually attempt to build, but fail while building a transitive:

```
dotnet-vmr-10.0.0-rc.1.drv' failed with exit code 139;
./prep-source-build.sh: line 363:  6041 Segmentation fault         (core dumped)
```

khaneliman: No this has an issue on nixpkgs specifically for it already. https://github.com/NixOS/nixpkgs/issues/431811

Ironically, I think we only have the issue on linux... it builds fine on darwin.

khaneliman (approved):

mathjiajia:
Previously, I was using
```nix
programs.nixvim.lsp.servers.copilot = {
    enable = true;
    package = pkgs.copilot-language-server;
    ...
};
```
and it worked fine.

However, after this commit, I now get the following error:
```
error: Package ‘copilot-language-server-1.377.0’ in /nix/store/w9pfi29d0j9kagbcl1vb7wrlki48bfmb-source/pkgs/by-name/co/copilot-language-server/package.nix:47 has an unfree license (‘GitHub Copilot License’), refusing to evaluate.
```

MattSturgeon:
> Previously, I was using
> ```nix
> programs.nixvim.lsp.servers.copilot = {
>     enable = true;
>     package = pkgs.copilot-language-server;
>     ...
> };
> ```
> and it worked fine.
> 
> However, after this commit, I now get the following error:
> ```
> error: Package ‘copilot-language-server-1.377.0’ in /nix/store/w9pfi29d0j9kagbcl1vb7wrlki48bfmb-source/pkgs/by-name/co/copilot-language-server/package.nix:47 has an unfree license (‘GitHub Copilot License’), refusing to evaluate.
> ```


I assume you are saying you removed your explicit package definition so that you can use our default?

The first thing to be aware of is that there are multiple nixpkgs instances. You have a `pkgs` argument in your home-manager configuration, which is supplied to home-manager modules. And there is another `pkgs` that gets supplied to nixvim modules in your `programs.nixvim` submodule.


```nix
{ pkgs, ... }:
{
  # The pkgs here is configured by home-manager

  programs.nixvim =
    { pkgs, ... }:
    {
      # The pkgs here is configured by nixvim
    };
}
```

The default for all our packages options comes from nixvim's `pkgs`, not home-manager's.

You probably have `config.allowUnfree` somewhere in your home-manager configuration, or wherever the pkgs instance it is using is configured. You probably want to do this for nixvim's nixpkgs too.

```nix
programs.nixvim = {
  nixpkgs.config.allowUnfree = true;
}
```

mathjiajia:
Thank you.
You are right and it works now.

PR #3762, opened 2025-10-03
modules/lsp: rename server `settings` → `config`

MattSturgeon:
This is one possible solution to [the RFC](https://github.com/nix-community/nixvim/issues/3745), other solutions include [renaming to `cfg` instead](https://github.com/nix-community/nixvim/pull/3755), doing nothing, or removing the freeform structured config option and having top-level explicit options instead.

**Please keep discussion on this PR limited to code review, discussion about which solution to choose should happen on the RFC issue.**

This effectively reverts https://github.com/nix-community/nixvim/pull/3254

Closes #3755

### DIFF
--- a/modules/lsp/servers/default.nix
+++ b/modules/lsp/servers/default.nix
@@ -19,7 +19,14 @@ let
   # Create a submodule type from `server.nix`
   # Used as the type for both the freeform `lsp.servers.<name>`
   # and the explicitly declared `lsp.servers.*` options
-  mkServerType = args: types.submodule (mkServerModule args);
+  mkServerType =
+    args:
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
My preferred option.

MattSturgeon:
Since renaming back to `config` means we're reliant on `shorthandOnlyDefinesConfig`, I've pushed a second commit that _explicitly_ sets it via `submoduleWith`.

This is a no-op change, since it is already the default for `submodule`.

MattSturgeon:
This has been agreed on as the best solution to [the RFC](https://github.com/nix-community/nixvim/issues/3745), merging.

PR #3763, opened 2025-10-04
plugins/copilot-lsp: add module

khaneliman:
https://github.com/copilotlsp-nvim/copilot-lsp

### DIFF
--- a/plugins/by-name/copilot-lsp/default.nix
+++ b/plugins/by-name/copilot-lsp/default.nix
@@ -1,0 +1,24 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "copilot-lsp";
+  package = "copilot-lsp";
+
+  description = ''
+    A lightweight and extensible Neovim plugin for integrating GitHub Copilot's AI-powered
+    code suggestions via Language Server Protocol (LSP).
+
+    > [!NOTE]
+    > This plugin requires the `copilot-language-server`.
+    > Either enable through `copilot-lua` plugin or install manually.
+  '';
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  settingsExample = {
+    nes = {
+      move_count_threshold = 5;
+      distance_threshold = 100;
+      reset_on_approaching = false;
+    };
+  };
+}
--- a/tests/test-sources/plugins/by-name/copilot-lsp/default.nix
+++ b/tests/test-sources/plugins/by-name/copilot-lsp/default.nix
@@ -1,0 +1,18 @@
+{
+  empty = {
+    plugins.copilot-lsp.enable = true;
+  };
+
+  defaults = {
+    plugins.copilot-lsp = {
+      enable = true;
+      settings = {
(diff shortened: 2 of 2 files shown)


plugins/by-name/copilot-lsp/default.nix
```diff
@@ -0,0 +1,24 @@
...
+  name = "copilot-lsp";
+  package = "copilot-lsp";
+
+  description = ''
+    A lightweight and extensible Neovim plugin for integrating GitHub Copilot's AI-powered
+    code suggestions via Language Server Protocol (LSP).
+
+    > [!NOTE]
+    > This plugin requires the `copilot-language-server`.
+    > Either enable through `copilot-lua` plugin or install manually.
```
MattSturgeon, line 12: This implies `lsp.servers.copilot` and `plugins.lsp.servers.copilot` are not suitable ways to install the LSP. Is that correct?

MattSturgeon (approved):

plugins/by-name/copilot-lsp/default.nix
khaneliman: It requires it to be on `PATH` and those ways just set the `cmd` for vim.lsp, unfortunately. I was testing for a bit before I realized that was the issue.

MattSturgeon: The LSP module just adds the package to the PATH, so something else may have been going on with your testing.

Specifically, our lsp modules don't touch `cmd` (except for a _very_ small handful of exceptions in the old `plugins.lsp`).

Could your testing have been thrown off by this being one of the LSPs that only got support added a few hours ago?

MattSturgeon: It's possible that this plugin is taking responsibility for setting up and enabling the LSP via `vim.lsp.config()` and `vim.lsp.enable()`, so maybe our modules doing that is redundant. Although I'm not sure if that would cause issues.

khaneliman: Well, wtf... I couldn't get a single NES to work when using `lsp.servers.copilot` without `copilot-lua`...I'll play around again for a min.

MattSturgeon: At risk of stating the obvious, another thing to keep in mind is that the new `lsp` module doesn't install nvim-lspconfig. You'll want to ensure you're testing with `plugins.lspconfig`.

Alternatively, it's also possible that testing with lspconfig is causing its own issues, if both this plugin and lspconfig are configuring the LSP they may conflict with each other.

khaneliman: Yeah, it's been a nightmare testing all the permutations with / without lspconfig, with / without copilot-lua, with / without copilot-lsp for all these different combinations...

MattSturgeon: That does sound painful! To be clear, I've approved the PR because I'm happy with it. I'm not expecting the initial wording to be perfect and cover all possibilities.

Issue #3764, opened 2025-10-04
[BUG] `erlang-ls` no longer available in nixpkgs (removed in PR #448119)


culxttes:
### Plugin

LSP

### Nixpkgs Release

other

### Home Manager Release

other

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

Hello,

While updating my NixVim configuration, I encountered the following error:

```
error: 'erlang-ls' has been removed as it has been archived upstream. Consider using 'erlang-language-platform' instead
```

This issue occurs because `erlang-ls` was removed from **nixpkgs** in https://github.com/NixOS/nixpkgs/pull/448119, following its archival upstream.

As a result, any NixVim module or LSP configuration that references `erlang-ls` is now broken.
The recommended replacement package is `erlang-language-platform`, which provides similar language server capabilities for Erlang.

**Proposed solution:**

* Update NixVim’s Erlang LSP configuration to use `erlang-language-platform` instead of `erlang-ls`.

### Minimal, Reproducible Example (MRE)

```nix
{ pkgs, ... }: 

{
  programs.nixvim = {
    enable = true;
    lsp.enable = true;
  };
}
```

MattSturgeon:
I assume you are specifying a different nixpkgs revision, overriding nixvim's lock. I assume this because we don't see this issue in our test suite, yet.

Removing any `inputs.nixpkgs.follows` from your nixvim flake input should help.

PR #3765, opened 2025-10-04
plugins/lsp: remove erlang-ls

culxttes:
This PR removes all references to `erlang-ls` from the NixVim configuration.
The package has been removed from **nixpkgs** in https://github.com/NixOS/nixpkgs/pull/448119 after being archived upstream.

It fixes https://github.com/nix-community/nixvim/issues/3764, which reported build errors due to the missing `erlang-ls` package.

**Details**

* `erlang-ls` is no longer available in nixpkgs.
* The upstream project has been archived and is not maintained.
* Users are advised to switch to https://github.com/WhatsApp/erlang-language-platform, which provides LSP functionality for Erlang.

**Changes**

* Removed `erlang-ls` from the list of supported LSP servers.


generated/lspconfig-servers.json
```diff
@@ -94,7 +94,6 @@
...
   "erg_language_server": "https://github.com/erg-lang/erg#flags ELS ELS (erg-language-server) is a\nlanguage server for the Erg programming language. erg-language-server\ncan be installed via `cargo` and used as follows:\n\n``` sh\ncargo install erg --features els\nerg --language-server\n```",
-  "erlangls": "https://erlang-ls.github.io Language Server for Erlang. Clone\n[erlang_ls](https://github.com/erlang-ls/erlang_ls) Compile the project\nwith `make` and copy resulting binaries somewhere in your \\$PATH eg.\n`cp _build/*/bin/* ~/local/bin` Installation instruction can be found\n[here](https://github.com/erlang-ls/erlang_ls). Installation\nrequirements: - [Erlang OTP 21+](https://github.com/erlang/otp) -\n[rebar3 3.9.1+](https://github.com/erlang/rebar3)",
```
MattSturgeon, line 97: This file is auto-generated from nvim-lspconfig. You can't just make changes to it directly.

This deprecation should be handled in nvim-lspconfig.

plugins/lsp/lsp-packages.nix
```diff
@@ -243,7 +243,6 @@
     emmet_ls = "emmet-ls";
     emmylua_ls = "emmylua-ls";
     erg_language_server = "erg";
-    erlangls = "erlang-ls";
```
MattSturgeon, line 246: It is fine to just move the lsp to the "unpackaged" section when we bump our flake.lock.

MattSturgeon (changes requested):
Thanks for your contribution, but this change should happen in nvim-lspconfig, not nixvim.

Nixvim will need to improve how it handles deprecating LSPs at some point, e.g. we could have a separate `deprecated-lsp-servers.nix` file. But currently we simply list all servers that nvim-lspconfig supports.

c2fc2f:
Sorry, I had a little problem with my git, which was entirely my fault but it's good now

MattSturgeon (approved):
Thanks

PR #3766, opened 2025-10-04
tests/all-package-defaults: recursively collect package options into submodules

GaetanLepage:


tests/all-package-defaults.nix
```diff
@@ -126,14 +127,24 @@ let
...
-  NOTE: This doesn't currently recurse into submodule option sets.
-*/
+  # Collect all derivation-type option defaults in the top-level option set.
+  # Recurse into submodule option sets too.
+  collectAllOptions =
+    options:
+    lib.concatMap (
+      opt:
+      let
+        visible = opt.visible or true;
```
MattSturgeon, line 137:
```suggestion
        visible = opt.visible or true;
        optVisible = if lib.isBool visible then visible else visible == "shallow";
```

tests/all-package-defaults.nix
```diff
@@ -126,14 +127,24 @@ let
...
+        subOptionsVisible = if lib.isBool visible then visible else visible == "transparent";
+        subOptionSet = opt.type.getSubOptions opt.loc;
+        subOptions = lib.optionals (subOptionSet != { }) (collectAllOptions subOptionSet);
+      in
+      [ opt ] ++ lib.optionals subOptionsVisible subOptions
```
MattSturgeon:
```suggestion
      lib.optional optVisible opt
      ++ lib.optionals subOptionsVisible subOptions
```

tests/all-package-defaults.nix
```diff
@@ -126,14 +127,24 @@ let
...
   isEnabled = p: !(builtins.elem (lib.getName p) disabledTests);
   isAvailable = lib.meta.availableOn hostPlatform;
-in
-/*
-  Collect all derivation-type option defaults in the top-level option set.
 
-  NOTE: This doesn't currently recurse into submodule option sets.
-*/
+  # Collect all derivation-type option defaults in the top-level option set.
+  # Recurse into submodule option sets too.
```
MattSturgeon: I think this comment is now in the wrong place: most of this comment relates to the pipe below, not this function.

If you want a comment on this function it can simply be "collects all visible options, including sub options".

If you add such a comment, then we probably don't need the note about recursion in the other comment.

MattSturgeon (commented):
Nice!

I think we probably want to restrict this to visible options only? See code suggestions.

tests/all-package-defaults.nix
GaetanLepage: Sure, good point.

tests/all-package-defaults.nix
```diff
@@ -126,14 +127,27 @@ let
...
+        subOptionsVisible = if lib.isBool visible then visible else visible == "transparent";
+        subOptionSet = opt.type.getSubOptions opt.loc;
+        subOptions = lib.optionals (subOptionSet != { }) (collectAllOptions subOptionSet);
+      in
+      lib.optional optVisible opt ++ lib.optionals subOptionsVisible subOptions
+    ) (lib.collect lib.isOption options);
+
+in
+# Collect all derivation-type option defaults in the top-level option set.
+# Recurse into submodule option sets too.
```
MattSturgeon: This is what I meant by:

> If you add such a comment, then we probably don't need the note about recursion in the other comment.

```suggestion
# Collect all derivation-type option defaults in Nixvim.
```

Issue #3600, opened 2025-08-04
lspconfig: `volar` has been deprecated and renamed to `vue_ls`


wvffle:
According to https://github.com/neovim/nvim-lspconfig/pull/3843, `volar` has been deprecated and renamed to `vue_ls`

When launching nvim, I get:

```
volar is deprecated, use vue_ls instead.                                                                                                                                              
Feature will be removed in lspconfig 3.0.0                                                                                                                                            
Press ENTER or type command to continue  
```

I enable volar with:
```nix
{
  lsp.servers.volar.enable = true;
}
```

wvffle:
In the meantime I've migrated over to `vtsls` as it is required for vue language server now.

When I open up a `.vue` file, a new error is shown:

```
Error executing vim.schedule lua callback: ...ack/myNeovimPackages/start/nvim-lspconfig/lsp/vue_ls.lua:44: attempt to index local 'r' (a nil value)
stack traceback:
	...ack/myNeovimPackages/start/nvim-lspconfig/lsp/vue_ls.lua:44: in function 'handler'
	...wrapped-0.11.3/share/nvim/runtime/lua/vim/lsp/client.lua:682: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>

```

This seems to be already fixed in the lspconfig: https://github.com/neovim/nvim-lspconfig/commit/177be29e7b9c084c9200ee4033eb66e04ac536ec

hyshka:
> In the meantime I've migrated over to `vtsls` as it is required for vue language server now.
> 
> When I open up a `.vue` file, a new error is shown:
> 
> ```
> Error executing vim.schedule lua callback: ...ack/myNeovimPackages/start/nvim-lspconfig/lsp/vue_ls.lua:44: attempt to index local 'r' (a nil value)
> stack traceback:
> 	...ack/myNeovimPackages/start/nvim-lspconfig/lsp/vue_ls.lua:44: in function 'handler'
> 	...wrapped-0.11.3/share/nvim/runtime/lua/vim/lsp/client.lua:682: in function ''
> 	vim/_editor.lua: in function <vim/_editor.lua:0>
> ```
> 
> This seems to be already fixed in the lspconfig: [neovim/nvim-lspconfig@177be29](https://github.com/neovim/nvim-lspconfig/commit/177be29e7b9c084c9200ee4033eb66e04ac536ec)

[...]

PR #3771, opened 2025-10-04
plugins/lsp/vue_ls: add a `tslsIntegration` option

jfly:
This new `plugins.lsp.servers.vue_ls.tslsIntegration` is a shameless copy of of pre-existing `plugins.lsp.servers.volar.tslsIntegration` option.

Full disclosure: I have not tested this as I use the newer `lsp` module.

I'm not sure if this fixes
https://github.com/nix-community/nixvim/issues/3600, but it's a least part of it.

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -119,7 +119,7 @@ let
     };
     ts_ls = {
       # NOTE: Provide the plugin default filetypes so that
-      # `plugins.lsp.servers.volar.tslsIntegration` doesn't wipe out the default filetypes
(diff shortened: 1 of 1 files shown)


khaneliman (approved):

luis-agm:
@jfly out of curiosity, what do you mean by "new lsp module"?

MattSturgeon:
> out of curiosity, what do you mean by "new lsp module"? 

We currently have an "old" `plugins.lsp` module and a "new"/"experimental" `lsp` module.

We're working behind the scenes to bring the new module up to feature parity (where appropriate) and plan for how & when to deprecate the old module.

PR #3775, opened 2025-10-04
plugins/blame-nvim: init

axelkar:
https://github.com/FabijanZulj/blame.nvim

### DIFF
--- a/plugins/by-name/blame-nvim/default.nix
+++ b/plugins/by-name/blame-nvim/default.nix
@@ -1,0 +1,19 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "blame-nvim";
+  moduleName = "blame";
+  description = "fugitive.vim-style `git blame` visualizer for Neovim";
+
+  maintainers = [ lib.maintainers.axka ];
+
+  settingsExample = {
+    date_format = "%Y-%m-%d";
+    views.default = lib.nixvim.nestedLiteralLua "require('blame.views.virtual_view')";
+    format_fn = lib.nixvim.nestedLiteralLua "require('blame.formats.default_formats').date_message";
+    colors = [
+      "Pink"
+      "Aqua"
+      "#ffffff"
+    ];
+  };
+}
--- a/tests/test-sources/plugins/by-name/blame-nvim/default.nix
+++ b/tests/test-sources/plugins/by-name/blame-nvim/default.nix
@@ -1,0 +1,54 @@
+{
+  empty = {
+    plugins.blame-nvim.enable = true;
+  };
+
+  defaults = {
+    plugins.blame-nvim = {
+      enable = true;
+      settings = {
+        date_format = "%d.%m.%Y";
+        relative_date_if_recent = true;
+        virtual_style = "right_align";
+        views = {
+          window.__raw = "require('blame.views.window_view')";
+          virtual.__raw = "require('blame.views.virtual_view')";
+          default.__raw = "require('blame.views.window_view')";
+        };
+        focus_blame = true;
+        merge_consecutive = false;
+        max_summary_width = 30;
+        colors = null; # won't output literal nil
+        blame_options = null; # won't output literal nil
+        format_fn.__raw = "require('blame.formats.default_formats').commit_date_author_fn";
+        commit_detail_view = "vsplit";
+        mappings = {
+          commit_info = "i";
+          stack_push = "<TAB>";
+          stack_pop = "<BS>";
+          show_commit = "<CR>";
+          close = [
+            "<esc>"
+            "q"
+          ];
+        };
+      };
+    };
+  };
+
+  example = {
+    plugins.blame-nvim = {
+      enable = true;
+      settings = {
+        date_format = "%Y-%m-%d";
+        views.default.__raw = "require('blame.views.virtual_view')";
+        format_fn.__raw = "require('blame.formats.default_formats').date_message";
+        colors = [
+          "Pink"
+          "Aqua"
+          "#ffffff"
+        ];
+      };
+    };
+  };
+}


plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,145 @@
+{ lib, ... }:
+let
+  inherit (lib.nixvim) defaultNullOpts;
+in
```
GaetanLepage:
```suggestion
```

plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,145 @@
...
+            '';
+
+            close = mkMappingOption [ "<esc>" "q" ] ''
+              Mapping to close the window view.
+            '';
+          };
+        };
+      };
+  };
+
```
GaetanLepage:
```suggestion
```

GaetanLepage (commented):
Thank you for your contribution.
We try to avoid explicitly defining `settingsOptions` at all cost and I thus recommend you to drop the `settingsOptions` definitions.
This will help reducing the maintenance burden in the future.
As `settings` is a free-form option, users will be able to provide any value.

Now, if you really want to keep those definitions in, you are free to do so. In this case, I hope that you will be ready to closely monitor this plugin's evolution and possibly add deprecation warnings & co when necessary.

axelkar:
> We try to avoid explicitly defining `settingsOptions` at all cost

I thought that was a requirement here. Sounds great!

GaetanLepage:
> > We try to avoid explicitly defining `settingsOptions` at all cost
> 
> I thought that was a requirement here. Sounds great!

No they are not. Maybe we would need to improve our documentation...
Sorry for the wasted time on writing the options :/

plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,21 @@
...
+      date_format = "%Y-%m-%d";
+      views.default = lib.nixvim.mkRaw '''
+        require("blame.views.virtual_view")
+      ''';
+      format_fn = lib.nixvim.mkRaw '''
+        require("blame.formats.default_formats").date_message
+      ''';
+      colors = [ "Pink" "Aqua" "#ffffff" ];
+    }
+  '';
```
GaetanLepage: I would stick to the usual way of declaring `settingsExample` (not using `literalExpression` at the outer level) while we take a global decision on the matter.

axelkar: Seems that `nestedLiteralLua` semantically does the same without wrapping everything in a string, but it doesn't work with e.g. `plugins.foo.settings = options.foo.settings.example` either.

tests/test-sources/plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,58 @@
...
+
+  defaults = {
+    plugins.blame-nvim = {
+      enable = true;
+      settings = {
+        date_format = "%d.%m.%Y";
+        relative_date_if_recent = true;
+        virtual_style = "right_align";
+        views = {
+          window.__raw = ''require("blame.views.window_view")'';
```
GaetanLepage:
```suggestion
          window.__raw = "require('blame.views.window_view')";
```
FYI, you can switch to simple quotes in the lua string. Then you are able to use the regular double quotes for the nix string (not a big deal obviously).

tests/test-sources/plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  example = {
+    plugins.blame-nvim = {
+      enable = true;
+      settings = {
+        date_format = "%Y-%m-%d";
+        views.default.__raw = ''
+          require("blame.views.virtual_view")
+        '';
```
GaetanLepage:
```suggestion
        views.default.__raw = "require('blame.views.virtual_view')";
```
Same here, you can compact it.

plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,23 @@
...
+  maintainers = [ lib.maintainers.axka ];
+
+  settingsExample = {
+    date_format = "%Y-%m-%d";
+    views.default.__raw = ''
+      require("blame.views.virtual_view")
+    '';
+    format_fn.__raw = ''
+      require("blame.formats.default_formats").date_message
+    '';
```
GaetanLepage:
```suggestion
    views.default = lib.nixvim.nestedLiteralLua "require('blame.views.virtual_view')";
    format_fn = lib.nixvim.nestedLiteralLua "require('blame.formats.default_formats').date_message";
```
If I'm not wrong, this will make the documentation look decent.
@MattSturgeon is more familiar with those helpers than I am.

axelkar: FYI, I posted this in the other thread just a moment ago:

> Seems that `nestedLiteralLua` semantically does the same [as `literalExpression`] without wrapping everything in a string, but it doesn't work with e.g. `plugins.foo.settings = options.foo.settings.example` either.

plugins/by-name/blame-nvim/default.nix
MattSturgeon: > while we take a global decision on the matter.

@GaetanLepage There's no need for global decisions here. `literalExpression` and `literalMD` are just tools in the toolbox to produce specific kinds of example (or defaultText) values.

Any time you want any of the following:
- A specific formatting not produced by `lib.generators.toPretty`
- A non-literal nix expression (e.g. referencing `config` or applying some function)
- Visible comments

...you should use a `literalExpression`.

Any time you want to include non-code text in your docs-value, you should use `literalMD`.

This is the same as nixpkgs, so we don't really need to document it ourselves, however we do explicitly mention it in [CONTRIBUTING -> Writing option examples](https://github.com/nix-community/nixvim/blob/main/CONTRIBUTING.md#writing-option-examples):

> Most of the time you will set `example` to a nix value and the module system will render it using `lib.generators.toPretty`. However, sometimes the example you wish to show would benefit from a specific formatting, may include function application or use of nix operators, or may benefit from some non-code explanation. In that scenario, you should use `lib.literalExpression` or `lib.literalMD`.

plugins/by-name/blame-nvim/default.nix
MattSturgeon: > but it doesn't work with e.g. `plugins.foo.settings = options.foo.settings.example` either.

Correct, docs values (`example` and `defaultText`) are not intended to be _used_ in actual option definitions.

The fact they can be declared via a real nix value is purely a convenience and a sanity check, since typing nix values in a string has the risk of syntax errors.

If the only reason your example was a literalExpression was to use `mkRaw`, then yes, `nestedLiteralLua` is usually the preferred approach. If you had other reasons to use a literalExpression, then that's fine too.

tests/test-sources/plugins/by-name/blame-nvim/default.nix
```diff
@@ -0,0 +1,54 @@
...
+        views = {
+          window.__raw = "require('blame.views.window_view')";
+          virtual.__raw = "require('blame.views.virtual_view')";
+          default.__raw = "require('blame.views.window_view')";
+        };
+        focus_blame = true;
+        merge_consecutive = false;
+        max_summary_width = 30;
+        colors = null; # won't output literal nil
+        blame_options = null; # won't output literal nil
```
MattSturgeon, line 22: Do you want a literal `nil`? If so, you can do:

```suggestion
        colors.__raw = "nil";
        blame_options.__raw = "nil";
```

MattSturgeon (approved):
SGTM

tests/test-sources/plugins/by-name/blame-nvim/default.nix
axelkar: After removing `settingsOptions` that should work!

GaetanLepage:
Sorry, why was the module named `blame-nvim` and not `blame`? cc @MattSturgeon

axelkar:
> Sorry, why was the module named `blame-nvim` and not `blame`? cc @MattSturgeon 

It should've been `blame`. 🤦‍♂️ I'll probably also make the default settings check use literal Lua `nil` once I get to fixing that.

GaetanLepage:
> > Sorry, why was the module named `blame-nvim` and not `blame`? cc @MattSturgeon
> 
> It should've been `blame`. 🤦‍♂️ I'll probably also make the default settings check use literal Lua `nil` once I get to fixing that.

Don't worry. I opened https://github.com/nix-community/nixvim/pull/3798.

PR #3076, opened 2025-03-16
plugins/aw-watcher: init

axelkar:
Depends on https://github.com/NixOS/nixpkgs/pull/390520

PR #3776, opened 2025-10-04
treewide(docs): use literalLua instead of __raw in option examples

axelkar:
See https://github.com/nix-community/nixvim/pull/3076#discussion_r2404140875

### DIFF
--- a/plugins/by-name/nvim-snippets/default.nix
+++ b/plugins/by-name/nvim-snippets/default.nix
@@ -41,7 +41,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
     '';
 
     search_paths =
-      defaultNullOpts.mkListOf types.str [ { __raw = "vim.fn.stdpath('config') .. '/snippets'"; } ]
+      defaultNullOpts.mkListOf types.str
+        (lib.nixvim.literalLua "{ vim.fn.stdpath('config') .. '/snippets' }")
         ''
           Paths to search for snippets.
         '';
--- a/plugins/by-name/auto-session/default.nix
+++ b/plugins/by-name/auto-session/default.nix
@@ -18,7 +18,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
       Enables/disables auto creating, saving and restoring.
     '';
 
-    root_dir = defaultNullOpts.mkStr { __raw = "vim.fn.stdpath 'data' .. '/sessions/'"; } ''
+    root_dir = defaultNullOpts.mkStr (lib.nixvim.literalLua "vim.fn.stdpath 'data' .. '/sessions/'") ''
       Root directory for session files.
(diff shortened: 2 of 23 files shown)


plugins/by-name/nvim-snippets/default.nix
```diff
@@ -41,7 +41,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
     '';
 
     search_paths =
-      defaultNullOpts.mkListOf types.str [ { __raw = "vim.fn.stdpath('config') .. '/snippets'"; } ]
+      defaultNullOpts.mkListOf types.str
+        (lib.nixvim.literalLua "{ vim.fn.stdpath('config') .. '/snippets' }")
```
GaetanLepage, line 45:
```suggestion
      defaultNullOpts.mkListOf types.str
        [ (lib.nixvim.literalLua "vim.fn.stdpath('config') .. '/snippets'" ])
```
`mkListOf types.str` options allow `__raw` elements by default.

axelkar: There's no `__raw` there anymore, that argument is only for documentation. The whole goal of this PR is to get rid of `__raw` because it doesn't render well. Another way to do this would be to pipe `__raw` to `literalLua` in `pluginDefault`.

Actually, if we were to start making the default config checks automatically based on `pluginDefault`s, they would be broken by this PR. I'll turn this into a draft PR.

MattSturgeon: > Actually, if we were to start making the default config checks automatically based on pluginDefaults, they would be broken by this PR. I'll turn this into a draft PR.

There's no need to block this PR on tangential or unrelated refactoring.

`pluginDefault` is purely for documentation. There are no plans to use it for generating tests.

Feel free to undraft.

MattSturgeon: @GaetanLepage 's suggestion of `[ (lib.nixvim.literalLua "vim.fn.stdpath('config') .. '/snippets'" ])` won't work, because it's embedding a literalExpression within a value.

You could do any of:

```nix
# I think this is what Gaétan meant:
[ (nestedLiteralLua "...") ]
```

```nix
# This is the least "magic":
literalExpression ''[ (lib.nixvim.mkRaw "...") ]''
```

```nix
# This has a disadvantage of not demonstrating how list items can be raw values:
literalLua "{ ... }"
```

axelkar: > There's no need to block this PR on tangential or unrelated refactoring.
> 
> `pluginDefault` is purely for documentation. There are no plans to use it for generating tests.
> 
> Feel free to undraft.

Great! I thought that was the reason @GaetanLepage suggested a list at the top level.

khaneliman (approved):

GaetanLepage:
Can you solve the merge conflict please?

PR #3778, opened 2025-10-05
plugins/roslyn: init module

khaneliman:
https://github.com/seblyng/roslyn.nvim

### DIFF
--- a/plugins/by-name/roslyn/default.nix
+++ b/plugins/by-name/roslyn/default.nix
@@ -1,0 +1,28 @@
+{ config, lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "roslyn";
+  package = "roslyn-nvim";
+
+  dependencies = [
+    "roslyn_ls"
+  ];
+
+  maintainers = [ lib.maintainers.khaneliman ];
+
+  extraConfig = {
+    warnings = lib.nixvim.mkWarnings "plugins.roslyn" {
+      when = config.lsp.servers.roslyn_ls.enable;
+
+      message = ''
+        You have enabled both `lsp.servers.roslyn_ls` and `plugins.roslyn`.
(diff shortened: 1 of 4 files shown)


plugins/by-name/roslyn/default.nix
```diff
@@ -0,0 +1,21 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "roslyn";
+  package = "roslyn-nvim";
+
+  description = ''
+    Roslyn LSP plugin for neovim.
+
+    > [!NOTE]
+    > This plugin requires `roslyn-ls`.
+    > Either enable through `lsp.servers.roslyn_ls` or install manually.
```
khaneliman: Just curious how we want to handle these sort of plugins. Seems to be a few recently that could have the dependency provided just through our `lsp.servers.*.package` mechanism, or we could add `dependencies` to it for the lsp. Otherwise, just been adding a note, for now... but have this one, the previous copilot-lsp, and next i have a razor lsp plugin..

MattSturgeon: I guess it depends. If it just depends on the package, then that means it does its own configuration and enabling of the server.

If users should configure and enable the server themselves, then they should use `lsp.servers`.

We should also consider whether nvim-lspconfig conflicts. If so, does the plugin use a different "name" when calling `vim.lsp.enable()` so that it doesn't load nvim-lspconfig's server config? Anecdotally, I've seen some nvim-lspconfig descriptions that say things like "don't configure with lspconfig if you're using xyz plugin, as that plugin does its own configuration".

khaneliman: One is loaded as `roslyn` and the other `roslyn_ls` but it's still doing basically the same thing. I added a warning and just created a dependency to load it for the user.

MattSturgeon (approved):

Issue #3773, opened 2025-10-04
Where do LSP "quirks" belong in the new lsp module API?


jfly:
In the older `plugins.lsp` API, `plugins/lsp/language-servers/default.nix` has a huge collection of very useful options and sane defaults. For example, I recently learned about enabling `volar` [automagically enables `ts_ls` and reconfigures it with the appropriate plugin](https://github.com/nix-community/nixvim/blob/3fa0e487260af16dde609940e49c3ddc6c31c6ed/plugins/lsp/language-servers/default.nix#L155-L185). Really awesome, useful stuff.

As far as I can tell, we don't have a place for stuff like this yet in the newer `lsp` module API. To accomplish this, I had to put a version of that config in my personal nixvim config (https://github.com/jfly/snow/commit/f4b37e5f5ec6326a8bf7121a73cef25e15c94d14).

Is there a home for this and I just haven't found it yet? Or is this all still being figured out, and that's what https://github.com/nix-community/nixvim/issues/3745 is tracking?

MattSturgeon:
> Is there a home for this and I just haven't found it yet? Or is this all still being figured out, and that's what #3745 is tracking?

This is something that's still undetermined, but it's a separate issue to #3745.

That issue is about how we should represent the config passed to `vim.lsp.config()`.

Firstly, there's a fundamental difference between `lsp` and `plugins.lsp`: the former just represents neovim's builtin lsp API, while the latter represents nvim-lspconfig; a repository of configs for servers.

With the new system, that is split into `plugins.lspconfig`. That plugin simply adds nvim-lspconfig to the rtp, and is meant to be used in tandem with `lsp`.

While `lsp` is first-and-foremost supposed to represent neovim's API as nix options, we do add _some_ additional sugar, such as the `lsp.servers.<name>.package` option. This option is unrelated to the neovim API and simply adds lsp packages to the nvim wrapper's PATH.

For me, it makes sense that the `plugins.lsp` API had some minor tweaks and corrections to nvim-lspconfig's configs. But with `lsp` and `plugins.lspconfig` split, I'm not sure where such corrections would belong.

Keep in mind that the `lsp` module can be used without nvim-lspconfig if someone wants to configure the servers from scratch without the pre-defined configs.

[...]

PR #3783, opened 2025-10-06
modules/lsp: auto-import per-server custom modules

MattSturgeon:
Adding a module in the following places will import the module into that specific server's submodule.

This allows creating server-specific options.

Added an assertion to ensure all custom modules correspond to an actual server option to avoid accidental dead code:

> The following custom modules do not correspond to an LSP server option:
> - modules/lsp/servers/custom/foo.nix

This is a step towards porting things like `rust_analyzer.cargoPackage` over to the new module. I.e. any server-specific feature that makes sense with or without nvim-lspconfig.

Server-specific features that only make sense with nvim-lspconfig should instead live within `plugins.lspconfig`. We'll have to consider what that API should look like too, but in another PR.

See #3773


modules/lsp/servers/default.nix
```diff
@@ -191,5 +191,33 @@ in
...
+          in
+          {
+            assertion = modules == [ ];
+            message = ''
+              The following custom modules do not correspond to an LSP server option:
+              ${lib.concatMapStringsSep "\n" (module: "- ${module}") modules}
+            '';
+          }
+        )
+      ];
```
MattSturgeon: Maybe this assertion should be a dedicated test outside of the nixvim's public modules, or part of `test-sources/modules/lsp.nix`?

This should never trigger on user's nixvim configurations, so it's a waste of their resources to evaluate the assertion.

GaetanLepage: I agree. It would be better. (not crucial though).

PR #3787, opened 2025-10-06
plugins/rust-tools: remove module

khaneliman:
Not migrating an archived plugin. Just removing it and encouraging using rustaceanvim.


GaetanLepage (approved):

MattSturgeon (commented):
🎉

PR #3791, opened 2025-10-07
plugins/kiwi: init

### LINKED ISSUE #3615: [PLUGIN REQUEST] kiwi.nvim
### Plugin

kiwi.nvim

### Homepage

https://github.com/serenevoid/kiwi.nvim

### Nixpkgs URL

_No response_

### Extra Information

Similar to vimwiki.

GaetanLepage:
Add [kiwi.nvim](https://github.com/serenevoid/kiwi.nvim), a stripped down VimWiki for Neovim.

Closes #3615

### DIFF
--- a/plugins/by-name/kiwi/default.nix
+++ b/plugins/by-name/kiwi/default.nix
@@ -1,0 +1,18 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "kiwi";
+  package = "kiwi-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  settingsExample = {
+    __unkeyed-1 = {
+      name = "work";
+      path = "work-wiki";
+    };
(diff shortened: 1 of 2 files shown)


khaneliman (approved):

plugins/by-name/kiwi/default.nix
```diff
@@ -0,0 +1,18 @@
...
+  settingsExample = {
+    __unkeyed-1 = {
+      name = "work";
+      path = "work-wiki";
+    };
+    __unkeyed-2 = {
+      name = "personal";
+      path = "personal-wiki";
+    };
+  };
```
MattSturgeon, line 17: It'd be great for plugins like this one to be able to drop the submodule type entirely. It is unnecessary when we aren't declaring sub-options and it prevents us from using a more inclusive type for `settings`.

In this case we probably want `either listOf attrsOf`.

nomisreual:
Thanks a lot!

PR #3792, opened 2025-10-07
plugins/langmapper: init module

ein-shved:
Add [langmapper](https://github.com/Wansmer/langmapper.nvim) plugin configuration


plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+  settingsOptions = {
+    map_all_ctrl = mkBool true "Add mapping for every CTRL+ binding or not";
+    ctrl_map_modes = mkKeyModes [
+      "n"
+      "o"
+      "i"
+      "c"
+      "t"
+      "v"
+    ] "Modes to `map_all_ctrl`";
```
MattSturgeon: This might format nicer if we use [`lib.stringToCharacters`](https://noogle.dev/f/lib/stringToCharacters):

```suggestion
    ctrl_map_modes = mkKeyModes (lib.stringToCharacters "noictv") ''
      Modes to `map_all_ctrl`
    '';
```

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+
+  mkKeyModes = mkListOf lib.nixvim.keymaps.modeType;
+  cfg = config.plugins.langmapper;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "langmapper";
+  package = "langmapper-nvim";
+  description = "A plugin that makes Neovim more friendly to non-English input methods";
+  maintainers = [ lib.maintainers.shved ];
+  settingsOptions = {
```
MattSturgeon: Is it necessary to have these settings sub-options explicitly declared? And are you willing to keep them up-to-date if upstream changes?

Generally, in newer modules we tend to rely more on the freeform nature of `settings` and avoid explicitly declaring settings sub-options.

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+        type = types.str;
+        default = ''`qwertyuiop[]asdfghjkl;'zxcvbnm~QWERTYUIOP{}ASDFGHJKL:"ZXCVBNM<>'';
+      };
+      local_layouts = mkOption {
+        description = "List of localized layouts to perform mappings from";
+        type = with types; listOf str;
+        default = [ ];
+        apply =
+          res:
+          optional cfg.langmap.cyrillic ''ёйцукенгшщзхъфывапролджэячсмитьËЙЦУКЕНГШЩЗХЪФЫВАПРОЛДЖЭЯЧСМИТЬБЮ''
```
MattSturgeon: Doing this in `apply` seems unnecessary and hard to override.

If it needs to be done at all, can it be done via merging option definitions?

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+          ++ res;
+      };
+      cyrillic = mkEnableOption "cyrillic layout";
+      # TODO: add more layouts;
```
MattSturgeon: Is there a lower-maintaince way to do this? Or is having layouts defined in nix code worth the effort at all? I'm skeptical we want to maintain this.

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+    };
+  };
+  extraConfig = mkIf cfg.enable {
```
MattSturgeon: extraConfig is only applied when the plugin is enabled anyway:

```suggestion
  extraConfig = {
```

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+    extraConfigLuaPre = mkIf cfg.langmap.enable ''
```
MattSturgeon: I assume this should be grouped into the plugin's segment of lua config?

```suggestion
    plugins.langmapper.luaConfig.pre = mkIf cfg.langmap.enable ''
```

Note: I'd usually prefer defining `luaConfig.content` as `pre` and `post` are mainly intended for end-users.

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,104 @@
...
+        vim.fn.join({${
+          let
+            makeMap = local: "escape([[${local}]]) .. ';' .. default_layout";
+          in
+          concatStringsSep "," (map makeMap cfg.langmap.local_layouts)
+        }}, ',')
+      end
+      _configure_langmap()
+    '';
+    extraConfigLuaPost = mkIf cfg.automapping.enable ''
```
MattSturgeon, line 48:
```suggestion
    plugins.langmapper.luaConfig.post = mkIf cfg.automapping.enable ''
```

plugins/by-name/langmapper/default.nix
ein-shved: Well, I did not read `caution` in contributing guide carefully. Sorry for that, will remove

plugins/by-name/langmapper/default.nix
ein-shved: I do not think that `vim.opt.langmap` is using in another way and it is much more handy to have something like

```nix
{
     i_am_using_cyrillic_layout = true;
     i_am_using_japanise_layout = true;
     i_am_using_korean_layout = false;
     i_am_using_sun_type_7_keyboard = true;
}
```

instead of long lua code with unclear lines of symbols.

But you are right - nixvim is not very suitable place for this, we are better to implement this in extra plugin.

ein-shved:
@MattSturgeon to summarize: is it better to remove `extraOptions` and `settingsOption` at all for this patch?

GaetanLepage:
> @MattSturgeon to summarize: is it better to remove `extraOptions` and `settingsOption` at all for this patch?

Yes.

plugins/by-name/langmapper/default.nix
ein-shved: Removed

plugins/by-name/langmapper/default.nix
ein-shved: Removed

plugins/by-name/langmapper/default.nix
ein-shved: Removed

plugins/by-name/langmapper/default.nix
ein-shved: Removed

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,11 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "langmapper";
+  package = "langmapper-nvim";
+  description = "A plugin that makes Neovim more friendly to non-English input methods";
+
+  maintainers = [ lib.maintainers.shved ];
```
GaetanLepage, line 20:
```suggestion
  maintainers = [ lib.maintainers.shved ];
  
  settingsExample = {
    automapping_modes = [ "i" "n" "v" "x" "s" ];
    map_add_ctrl = false;
  };
```

tests/test-sources/plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,31 @@
+{
+  empty = {
+    plugins.langmapper.enable = true;
+  };
+
```
GaetanLepage, line 6:
```suggestion

  example = {
    plugins.langmapper = {
      enable = true;
      
      settings = {
        automapping_modes = [ "i" "n" "v" "x" "s" ];
        map_add_ctrl = false;
      };
    };
  };

```

MattSturgeon:
> to summarize: is it better to remove `extraOptions` and `settingsOption` at all for this patch?

Removing `settingsOption` is generally best. `extraOptions` is a more personal decision.

Sometimes having a few "extra" options is nice, to provide extra functionality or more nix-flavoured abstractions. But you have to decide whether the option in question is worth the complexity and/or maintenance burden.

In this case, the plugin has a `automapping()` function to call  separately from `setup()`, so having an option to represent automappings in nix sounds useful. But it needs to be designed in a somewhat "freeform" way; i.e. make use of `attrsOf anything` and `toLuaObject`.

I'm not sure I really understand the point of the `langmap` option though. It is setting `vim.opt.langmap` directly in lua (note that we have an `opts` option for setting things in nix, but that won't support dynamically prepending to the existing lua value). But the main issue I have with the way that option was designed is that we will need to maintain a set of mappings for various languages (keyboard layouts?).

If I understood _why_ it is useful for this specific plugin _and_ we could design it in a more low-maintenance way, _then_ I'd consider it.

ein-shved:
> But the main issue I have with the way that option was designed is that we will need to maintain a set of mappings for various languages (keyboard layouts?).

Yes, the perfect way is to design and maintain options for different combinations of keyboard layouts (generic, mac, sun, custom, etc ...) and languages.

The way I've implemented it first was not perfect and the nixvim is not the right place to design and implement this in perfect way. This is functionality for extra plugin. Maybe someday I'll do it.

ein-shved:
> In this case, the plugin has a `automapping()` function to call separately from `setup()`, so having an option to represent automappings in nix sounds useful. But it needs to be designed in a somewhat "freeform" way; i.e. make use of `attrsOf anything` and `toLuaObject`.

The plugin documentation says that I have to 
> call the `langmapper.automapping({ buffer = false })` function at the very end of your `init.lua`

But I better not to use `extraConfigLuaPost` in plugin module. How to get out of the situation?

plugins/by-name/langmapper/default.nix
ein-shved: Thank you!

tests/test-sources/plugins/by-name/langmapper/default.nix
ein-shved: Thank you!

MattSturgeon:
> The plugin documentation says that I have to
> 
> > call the `langmapper.automapping({ buffer = false })` function at the very end of your `init.lua`
> 
> But I better not to use `extraConfigLuaPost` in plugin module. How to get out of the situation?

If that documentation is justified (and not over-simplified) then you can of course use `extraConfigLuaPost` or even `extraConfigLuaPost = lib.mkAfter ""`. Usually we prefer to group plugin config together in that plugin's `luaConfig`, so that it is easier to manage and so related lua lines end up together in init.lua. However if there's a good reason to do something else, then that's fine.

Usually explanatory comments are helpful in such exceptional circumstances.

ein-shved:
@MattSturgeon Thanks! Done! =)

plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,57 @@
...
+      "n"
+      "v"
+      "x"
+      "s"
+    ];
+    map_add_ctrl = false;
+  };
+  extraOptions = {
+    automapping = {
+      enable = mkEnableOption "automapping calling at the end of init.lua" // {
```
MattSturgeon:
```suggestion
      enable = mkEnableOption "calling `automapping` at the end of `init.lua`" // {
```

or

```suggestion
      enable = mkEnableOption "calling `automapping`" // {
```

or

```suggestion
      enable = mkEnableOption "setting up automappings" // {
```

tests/test-sources/plugins/by-name/langmapper/default.nix
```diff
@@ -0,0 +1,66 @@
...
+      enable = true;
+
+      settings = {
+        automapping_modes = [
+          "i"
+          "n"
+          "v"
+          "x"
+          "s"
+        ];
```
MattSturgeon: Tip: if this formatting is annoying, you can do something like:

```suggestion
        automapping_modes = lib.stringToCharacters "invxs";
```

ein-shved: Not formatting, but a bunch of `"` =)

plugins/by-name/langmapper/default.nix
ein-shved: Yeah that is better

PR #3799, opened 2025-10-08
plugins/cybu: init

Fovir-GitHub:
Add [cybu.nvim](https://github.com/ghillb/cybu.nvim).


FovirDev:
Since the `vimPlugins.cybu-nvim` is added into `nixpkgs` for a short time, and I update the `flake`, which is unstable. So this PR don't need to be reviewed until the `nixpkgs` becomes relatively stable.

FovirDev:
In addition, shall I update the flake or not? If I should not, I will revert it.

GaetanLepage:
No, you should not update the flake in this PR. We are already tackling this in https://github.com/nix-community/nixvim/pull/3797.
Let's just wait for it to be merged before rebasing this PR.

FovirDev:
I have dropped the flake update.

plugins/by-name/cybu/default.nix
```diff
@@ -0,0 +1,48 @@
...
+          switch = "on_close";
+          view = "paging";
+          update_on = "buf_enter";
+        };
+        auto = {
+          view = "rolling";
+        };
+      };
+    };
+    display_time = 750;
```
GaetanLepage: Not a deal breaker, but I would encourage you to shorten the `settingsExample`.
Indeed, this will appear in the documentation and it will benefit from being shorter.

You can leave a more comprehensive `example` test case in the test file.

plugins/by-name/cybu/default.nix
```diff
@@ -0,0 +1,48 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "cybu";
+  package = "cybu-nvim";
+
+  maintainers = [ lib.maintainers.Fovir ];
+
+  extraPlugins = with pkgs; [
+    vimPlugins.plenary-nvim
+    vimPlugins.nvim-web-devicons
+  ];
```
GaetanLepage: @MattSturgeon what is our policy for icon management already?
As those two dependencies are optional, and only needed when icon support is enabled, I don't think we should add them by default.

FovirDev: But the plugin itself enables the icon by default [Cybu Configuration](https://github.com/ghillb/cybu.nvim/blob/main/lua/cybu/config.lua#L23). Should we follow the default plugin?

FovirDev:
> Minor remarks. Otherwise looking good!
> 
> Also, can you squash all the commits into a single one `plugins/cybu: init` please?

I have done it already.

plugins/by-name/cybu/default.nix
FovirDev: I have removed unnecessary settings from the example, and left those I think are related to the core function of the plugin.

plugins/by-name/cybu/default.nix
FovirDev: I think you are right. So I disabled the icon by default and removed the two dependencies.

MattSturgeon: We normally add an assertion predicated on one of the compatible icon plugins being enabled:
https://github.com/nix-community/nixvim/blob/a3792ed5fd8e3e89f06879f5d59f85202ab84f2c/plugins/by-name/tiny-devicons-auto-colors/default.nix#L93-L105

In this case it could also check this plugin's relevant setting.

sometimes we enable the plugins by default:
https://github.com/nix-community/nixvim/blob/a3792ed5fd8e3e89f06879f5d59f85202ab84f2c/plugins/by-name/alpha/default.nix#L163-L179

The way we handle it is rather fragmented treewide, partly because some plugins depend on icons more strongly than others, but mostly because we haven't centralized our logic. This has led to some assertions getting outdated, e.g. we now have two ways to enable mini-icons, but most assertions only check the older way.

PR #3803, opened 2025-10-09
plugins/neo-tree: migrate to mkNeovimPlugin

GaetanLepage:
This is a chunky one!

- **plugins/neo-tree: remove old warnings**
- **plugins/neo-tree: remove deprecated automatic web-devicons integration**
- **plugins/neo-tree: migrate to mkNeovimPlugin**

Tracking: #2638


plugins/by-name/neo-tree/deprecations.nix
```diff
@@ -0,0 +1,173 @@
...
+          message = ''
+            Use `plugins.neo-tree.settings.${newSubPath}`.
+            WARNINGS:
+              - sub-option names are the upstream ones, and use snake_case:
+                `displayName` -> `display_name`
+              - Lua string are not automatically converted to lua:
+                `foo = "vim.fn.stdpath('data')"` -> `foo.__raw = "vim.fn.stdpath('data')"`
+          '';
+        in
+        lib.mkRemovedOptionModule oldPath message;
```
khaneliman, line 24: This is to avoid the difficult graceful transition for those existing mutations we were doing, i'm assuming?

GaetanLepage: Yes, I would prefer to simply get rid of the special handling logic that we had for those options.

PR #3804, opened 2025-10-09
plugins/navbuddy: migrate to mkNeovimPlugin

saygo-png:
Not deprecating "keymapsSilent" as it was a noop. I presume it was meant to be used in the `mappings` option but never was. Not fixing it because I feel that if the user wants to customize the binds like that, they should just be using our toplevel `keymaps` option or using `extraConfigLua*`.

defaults: https://github.com/hasansujon786/nvim-navbuddy/blob/18802efb91bcfafc54aed513ef97da1dfeae4059/lua/nvim-navbuddy/init.lua#L137-L273

I removed the `fuzzy_finder` bind from the defaults test as it throws without telescope and with the integration set to `nil`. (I'm assuming there is some logic that removes that bind if telescope/snacks isn't found but i can't find it)

One major issue is that the `mappings` option is merged with `settings` using `//` so it does not use the nixpkgs merging system. Not sure how I could add that in though as this value gets transpiled to Lua, so i can't just use `lib.mkMerge`. This is problematic as keymaps can both be set using the plugin's toplevel `mappings` option and the `settings` one.

### DIFF
--- a/plugins/by-name/navbuddy/default.nix
+++ b/plugins/by-name/navbuddy/default.nix
@@ -1,273 +1,32 @@
-{
-  lib,
-  helpers,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
+{ lib, ... }:
 let
-  cfg = config.plugins.navbuddy;
-
-  percentageType = types.ints.between 0 100;
-  mkPercentageOpt = default: helpers.defaultNullOpts.mkNullable percentageType (toString default);
+  inherit (lib.nixvim) nestedLiteralLua;
 in
-{
-  options.plugins.navbuddy = lib.nixvim.plugins.neovim.extraOptionsOptions // {
-    enable = mkEnableOption "nvim-navbuddy";
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "navbuddy";
+  package = "nvim-navbuddy";
+  moduleName = "nvim-navbuddy";
+  description = "A simple popup display that provides a breadcrumbs feature using an LSP server.";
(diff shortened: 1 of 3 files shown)


plugins/by-name/navbuddy/default.nix
```diff
@@ -1,159 +1,37 @@
...
+    use_default_mapping = true;
+    icons = {
+      Array = "> ";
+      Boolean = "> ";
+      Class = "> ";
     };
+  };
 
+  extraOptions = {
     mappings =
```
GaetanLepage: I will open a follow-up PR to remove this option and tell the users to use `settings.mappings` instead.

plugins/by-name/navbuddy/default.nix
```diff
@@ -196,78 +74,27 @@ in
...
     in
-    mkIf cfg.enable {
-      extraPlugins = [ cfg.package ];
-
-      extraConfigLua = ''
-        local actions = require("nvim-navbuddy.actions")
-        require('nvim-navbuddy').setup(${lib.nixvim.toLuaObject setupOptions})
+    {
+      plugins.navbuddy.luaConfig.content = ''
+        local actions = require('nvim-navbuddy.actions')
```
GaetanLepage: In the (close) future, I would prefer getting rid of this and letting users explicitly write `require('nvim-navbuddy.actions').append_name()` in their mappings RHSs.

saygo-png:
I'll just get rid of `mappings` and the action binding in this PR.

PR #3805, opened 2025-10-09
modules/lsp/hls: port `ghcPackage` option from `plugins.lsp`

MattSturgeon:
WIP PR to begin porting `plugins.lsp` server-specific options to the new `lsp` module. Depends on #3783.

Open questions:
- Do we still want `installXyz` and `xyzPackage` options, or just nullable package options?
- Which options should be ported to `lsp` and which to `plugins.lspconfig`?

### DIFF
--- a/modules/lsp/servers/custom/hls.nix
+++ b/modules/lsp/servers/custom/hls.nix
@@ -1,0 +1,38 @@
+{
+  lib,
+  pkgs,
+  config,
+  options,
+  ...
+}:
+{
+  options = {
+    installGhc = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
+      description = "Whether to install `ghc`.";
+    };
+
+    ghcPackage = lib.mkPackageOption pkgs "ghc" { };
+  };
+
+  config = lib.mkIf config.enable {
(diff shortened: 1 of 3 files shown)


sportshead:
IIRC from my previous testing installing hls will cause *a version* (not sure if its the default nixpkgs one) of ghc to be in the closure, since hls has a dependency on ghc. Therefore I don't think `installGhc` actually does anything apart from possibly wasting another ~3gb of disk space, since the extra ghc will only be available in nvim (where it isn't needed) anyway.

sportshead:
Had a quick test and hls works with `installGhc = false` (using `plugins.lsp`). I do think `packageFallback` is much more useful for haskell – hls loves to complain whenever the bundled ghc version is not the exact same as the workspace ghc version.

saygo-png:
>  hls works with installGhc = false

I've tested it and it does not. I used `nix-shell` with `--pure` to get a clean environment. You most likely have ghc in your path from somewhere else, which is why it worked for you.
image

That said I much prefer a nullable package here instead of  an install option. I think it's more idiomatic and consistent with the rest of Nixvim and other Nix projects like home-manager.

PR #3807, opened 2025-10-10
docs/contributing: change `url` to an optional field

saygo-png:
`url` defaults to the home page defined in the plugin package.


MattSturgeon:
This is an odd one, because it _is_ required, even though it has a default. The oddity comes when the default isn't available; i.e. when the package has no `meta.homepage`.

saygo-png:
I see. It is weird in that it's required but I don't think we have a single plugin that explicitly sets `url`.

MattSturgeon:
> I don't think we have a single plugin that explicitly sets `url`.

Interesting, that's good to know. I guess nixpkgs must've improved how many vim plugin packages have a `meta.homepage`, as we certainly used to have `url`s explicitly declared.

PR #3810, opened 2025-10-11
docs/contributing: minor tweaks

saygo-png:
The most semantic change here is in the tables in `settingsOptions`. I changed it from optional to not recommended. 

This is motivated by a lot of new contributors being under the impression that they should declare `settings` options. I think "optional" implies "not needed, but desired". We do explain it in the warning but you have to click to get there. Hopefully this will prevent wasting time on writing options that then get deleted during review.

The rest is just me reading through and fixing mistakes/inconsistencies. For example the tables had the same fields but with different descriptions. I just picked one of descriptions and applied it to both.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ In order to submit a change you must be careful of several points:
 - The tests must pass. This can be done through `nix flake check --all-systems` (this also checks formatting).
 - The change should try to avoid breaking existing configurations.
 - If the change introduces a new feature it should add tests for it (see the architecture section for details).
-- The commit title should be consistent with our style. This usually looks like "plugins/<name>: fixed some bug",
+- The commit title should be consistent with our style. This usually looks like "plugins/\<name\>: fixed some bug",
   you can browse the commit history of the files you're editing to see previous commit messages.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -84,11 +84,11 @@ A template plugin can be found in (plugins/TEMPLATE.nix)[https://github.com/nix-
...
 | **moduleName**               | The Lua name for the plugin.                                                                                                                                                                                                                         | No       | `name` parameter                                                                                |
-| **optionsRenamedToSettings** | Options that have been renamed and move to the `settings` attribute.                                                                                                                                                                                 | No       | `[]`                                                                                            |
+| **optionsRenamedToSettings** | Options that have been renamed and moved to the `settings` attribute.                                                                                                                                                                                | No       | `[]`                                                                                            |
```
khaneliman: We should probably add a note about the transformation that gets applied.

saygo-png:
As a side note — I'm gonna try to add a fast docs build for development tomorrow. It's very painful to work on docs right now because each change takes around 3 minutes to build on my system, the majority of which is the mdbook search index.

PR #3812, opened 2025-10-12
plugins/gitlinker: migrate to mkNeovimPlugin

GaetanLepage:
Tracking: https://github.com/nix-community/nixvim/issues/2638

### DIFF
--- a/plugins/by-name/gitlinker/deprecations.nix
+++ b/plugins/by-name/gitlinker/deprecations.nix
@@ -1,0 +1,105 @@
+lib: {
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "gitlinker"
+      ];
+    in
+    [
+      (
+        let
+          oldPath = basePluginPath ++ [ "actionCallback" ];
+        in
+        lib.mkChangedOptionModule oldPath
+          (
+            basePluginPath
+            ++ [
+              "settings"
+              "opts"
+              "action_callback"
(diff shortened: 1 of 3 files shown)


plugins/by-name/gitlinker/deprecations.nix
```diff
@@ -0,0 +1,93 @@
...
+          )
+          (
+            config:
+            let
+              oldValue = lib.getAttrFromPath oldPath config;
+            in
+            lib.mapAttrs (
+              source: callback:
+              if lib.isString callback then
+                lib.nixvim.mkRaw "require('gitlinker.hosts').${callback}"
```
MattSturgeon: This should trigger a warning, in addition to the rename warning, so that users know the new option doesn't assume all strings are lua code.

This can either be `lib.warn` directly here in this conversion, or it could be a separate `warnings` option definition that reads the old value and checks none of the attrValues are strings.

E.g. doing it here in the conversion:

```nix
lib.warnIf (any isString (attrValues oldValue))
  "blah blah blah, defining `${showOption oldPath}` values as strings is not supported by its replacement `${showOption newPath}`. Please use `__raw` strings instead."
  (
    mapAttrs (source: callback:
    # ...
```

GaetanLepage: I added warnings. Please tell me if you think of a better phrasing.

PR #3813, opened 2025-10-12
plugins/startup: migrate to mkNeovimPlugin

saygo-png:

Removed assertions since they are hard (impossible?) to implement with freeform settings as there is no separation between sections and other settings.

Docs rendering is a bit ugly for the `__empty` part but as far as i know there is no `nestedLiteralLua` function for empty tables.

`settingsExample` gets tested by the `builtin-theme` test.

### DIFF
--- a/plugins/by-name/startup/default.nix
+++ b/plugins/by-name/startup/default.nix
@@ -1,257 +1,19 @@
-{
-  lib,
-  helpers,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
-let
-  cfg = config.plugins.startup;
-in
-{
-  meta.maintainers = [ maintainers.GaetanLepage ];
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "startup";
+  package = "startup-nvim";
+  description = "A highly configurable Neovim startup screen.";
+  maintainers = [ ];
 
-  options.plugins.startup = lib.nixvim.plugins.neovim.extraOptionsOptions // {
-    enable = mkEnableOption "startup.nvim";
+  # Plugin uses 2 functions to setup...
+  callSetup = false;
 
-    package = lib.mkPackageOption pkgs "startup.nvim" {
-      default = [
-        "vimPlugins"
-        "startup-nvim"
-      ];
-    };
+  # TODO: introduced 2025-10-17: remove after 26.05
+  inherit (import ./deprecations.nix lib) deprecateExtraOptions optionsRenamedToSettings imports;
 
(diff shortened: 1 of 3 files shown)


plugins/by-name/startup/default.nix
```diff
@@ -261,84 +23,35 @@ in
...
-          // cfg.extraOptions;
+  settingsExample = {
+    theme = "dashboard";
+    options = {
+      mapping_keys = true;
+      cursor_column = 0.5;
+      after = null;
+      empty_lines_between_mappings = true;
+      disable_statuslines = true;
+      paddings.__empty = { };
```
MattSturgeon: > Docs rendering is a bit ugly for the `__empty` part but as far as i know there is no `nestedLiteralLua` function for empty tables.

It's ok to assign a `literalExpression` to the settings example if you want to format something in a particular way.

Or if you prefer to do things the way `nestedLiteralLua` does internally, that's using the `__pretty` feature that `lib.generators.toPretty` supports.

plugins/by-name/startup/default.nix
```diff
@@ -261,84 +23,35 @@ in
     };
   };
 
-  config = mkIf cfg.enable {
-    assertions =
-      let
-        sectionNames = attrNames cfg.sections;
-        numSections = length sectionNames;
-      in
-      lib.nixvim.mkAssertions "plugins.startup" [
```
MattSturgeon, line 270: > Removed assertions since they are hard (impossible?) to implement with freeform settings as there is no separation between sections and other settings.

I don't see anything impossible here. Once moved to the new attr under `settings`, we'd just have to account for its freeform nature. I.e. we'd need to allow for the attr being missing (e.g with `?` or `or`), and for it being an unexpected type (e.g. raw lua or anything else we're not expecting).

We'd want to consider whether such edge cases should satisfy the assertions or show up as their own assertion failures.

So it's possible. The question is more about whether we _want_ to maintain such assertions here in nixvim.

saygo-png: The issue is that the `sections` is a freeform option in the plugin itself too. As in there are no defined ones, so we have to find them since there no longer is Nixvim's `sections`.  The only way I can think of doing this would be to code in all of the options of the plugin, and then check for any attributes that do not match the option names.

This is nearly equal in cost to just defining `settingsOptions` so I don't want to do this. This would also cause non existing plugin options to be interpreted as new sections which could lead to misleading errors for the user.

plugins/by-name/startup/default.nix
saygo-png: I prefer to avoid writing Nix in strings because it's error prone so I used `__pretty`

image

PR #3814, opened 2025-10-12
plugins/obsidian: drop old warnings, drop settings options declarations

GaetanLepage:
- **plugins/obsidian: drop old deprecation warnings**
- **plugins/obsidian: drop settings options declarations**


Ev357:
Thank you so much

PR #3815, opened 2025-10-13
plugins/lsp/vue_ls: add vtslsIntegration option

sportshead:
Add a vue_ls integration by default for vtsls in addition to tsls which is already implemented (see #3771).

Haven't added it to Volar since it is deprecated anyway (#3600)

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -9,6 +9,15 @@ let
 
   inherit (lib) mkOption types;
 
+  typescriptFiletypes = [
+    "javascript"
+    "javascriptreact"
+    "javascript.jsx"
+    "typescript"
+    "typescriptreact"
+    "typescript.tsx"
+  ];
+
   lspExtraArgs = {
     dartls = {
       settings = cfg: { dart = cfg; };
@@ -57,14 +66,7 @@ let
       # `plugins.lsp.servers.volar.tslsIntegration` and `plugins.lsp.servers.vue_ls.tslsIntegration` don't wipe out the default filetypes
       extraConfig = {
         plugins.lsp.servers.ts_ls = {
-          filetypes = [
-            "javascript"
-            "javascriptreact"
-            "javascript.jsx"
-            "typescript"
-            "typescriptreact"
-            "typescript.tsx"
-          ];
+          filetypes = typescriptFiletypes;
         };
       };
     };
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/default.nix
```diff
@@ -146,27 +148,52 @@ let
...
+            location = "${lib.getBin cfg.package}/lib/language-tools/packages/language-server";
+            languages = [ "vue" ];
+          };
+        in
+        {
+          assertions = lib.nixvim.mkAssertions "plugins.lsp.servers.vue_ls" {
+            assertion = (cfg.tslsIntegration or cfg.vtslsIntegration) -> (cfg.package != null);
+            message = "When `${opts.tslsIntegration}` or `${opts.vtslsIntegration}` is enabled, `${opts.package}` must not be null.";
+          };
+          plugins.lsp.servers.ts_ls = lib.mkIf (cfg.enable && cfg.tslsIntegration) {
```
GaetanLepage:
```suggestion
          plugins.lsp.servers.ts_ls = lib.mkIf cfg.tslsIntegration {
```

Everything in `extraConfig` is already conditioned on `cfg.enable`:
https://github.com/nix-community/nixvim/blob/a1d3ca973234f8013fb99903e915b700e6ddbbb5/lib/plugins/mk-neovim-plugin.nix#L93-L107

MattSturgeon:
I'm hesitant to add new features to `plugins.lsp` that we'll need to consider how to migrate to the new `lsp` and `plugins.lspconfig` replacements.

But I also don't want to block a PR on us figuring out how to approach that.

sportshead:
This doesn't really add a new feature per se, it only adds the existing feature (automatic vue_ls compatibility) to vtsls in addition to tsls which already had it. Although I do get what you mean, this does need to be migrated to the new `lsp` module eventually.

plugins/lsp/language-servers/default.nix
```diff
@@ -78,27 +80,52 @@ let
...
+        let
+          plugin = {
+            name = "@vue/typescript-plugin";
+            location = "${lib.getBin cfg.package}/lib/language-tools/packages/language-server";
+            languages = [ "vue" ];
+          };
+        in
+        {
+          assertions = lib.nixvim.mkAssertions "plugins.lsp.servers.vue_ls" {
+            assertion = (cfg.tslsIntegration or cfg.vtslsIntegration) -> (cfg.package != null);
```
MattSturgeon:
```suggestion
            assertion = (cfg.tslsIntegration || cfg.vtslsIntegration) -> (cfg.package != null);
```

This is the wrong "or" operator; logical or (`||`) vs attribute selection fallback (`or`)

https://nix.dev/manual/nix/2.32/language/operators.html

plugins/lsp/language-servers/default.nix
```diff
@@ -78,27 +80,52 @@ let
...
+          };
+          plugins.lsp.servers.ts_ls = lib.mkIf (cfg.enable && cfg.tslsIntegration) {
+            filetypes = [ "vue" ];
+            extraOptions = {
+              init_options = {
+                plugins = lib.mkIf (cfg.package != null) [ plugin ];
+              };
+            };
+          };
+          plugins.lsp.servers.vtsls = lib.mkIf (cfg.enable && cfg.vtslsIntegration) {
```
MattSturgeon: https://github.com/nix-community/nixvim/pull/3815/changes#r2529224980 also applies here:


```suggestion
          plugins.lsp.servers.vtsls = lib.mkIf cfg.vtslsIntegration {
```

plugins/lsp/language-servers/default.nix
MattSturgeon: This is unresolved. Perhaps it was reverted in a rebase?

sportshead: Yes, good catch, must have dropped it by accident

MattSturgeon (approved):

PR #3817, opened 2025-10-13
plugins/wilder: migrate to mkNeovimPlugin

GaetanLepage:
This was a tricky one.
This plugin has both a classic `.setup(opts)` configuration pattern and another imperative method to configure other options:
```lua
wilder.set_option('renderer', wilder.popupmenu_renderer({
  highlighter = {
    wilder.lua_pcre2_highlighter(),
    wilder.lua_fzy_highlighter(),
  },
  highlights = {
    accent = wilder.make_hl('WilderAccent', 'Pmenu', {{a = 1}, {a = 1}, {foreground = '#f4468f'}}),
  },
}))
```

In the end, I used the conventional free-form `settings` option and added an extra option called `options` (`attrsOf anything`) that will be mapped to `wilder.set_option` calls.

- **plugins/wilder: remove rename warning (2yo)**
- **plugins/wilder: migrate to mkNeovimPlugin**

Tracking: #2638

### DIFF
--- a/plugins/by-name/wilder/default.nix
+++ b/plugins/by-name/wilder/default.nix
@@ -1,258 +1,56 @@
-{
-  lib,
-  helpers,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
-let
-  cfg = config.plugins.wilder;
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "wilder";
+  package = "wilder-nvim";
+  maintainers = [ lib.maintainers.GaetanLepage ];
 
-  mkKeyOption =
-    default: desc:
-    helpers.defaultNullOpts.mkNullable
-      (
-        with types;
-        either str (submodule {
-          options = {
-            key = mkOption {
-              type = str;
-              description = desc;
-            };
+  # TODO: introduced 2025-10-13: remove after 26.05
+  inherit (import ./deprecations.nix lib) deprecateExtraOptions optionsRenamedToSettings imports;
 
-            fallback = mkOption {
-              type = either str (enum [ false ]);
-              description = ''
-                For a no-op, set <fallback> to "".
-
-                The fallback mappings can be disabled by setting `fallback` to 0.
-
-                Disabling the fallback mapping allows a `|<Cmd>|` mapping to be used which improves
-                rendering performance slightly as the mappings to be called outside the sandbox
-                (see `|:map-expression|`).
-              '';
-            };
-          };
-        })
-      )
-      default
-      ''
-        ${desc}
-
-        NOTE:
-        A string or an attrs (with keys `key` and `fallback`) representing the mapping to bind to
-        `|wilder#next()|`.
-        If a string is provided, it is automatically converted to `{key = <KEY>; fallback = <KEY>;}`.
-
-        - `mapping` is the `|cmap|` used to bind to `|wilder#next()|`.
-        - `fallback` is the mapping used if `|wilder#in_context()|` is false.
-      '';
-in
-{
-  options.plugins.wilder = lib.nixvim.plugins.neovim.extraOptionsOptions // {
-    enable = mkEnableOption "wilder-nvim";
-
-    package = lib.mkPackageOption pkgs "wilder-nvim" {
-      default = [
-        "vimPlugins"
-        "wilder-nvim"
-      ];
-    };
-
-    ### Setup options ###
-    enableCmdlineEnter = helpers.defaultNullOpts.mkBool true ''
-      If true calls `wilder#enable_cmdline_enter()`.
-      Creates a new `|CmdlineEnter|` autocmd to which will start wilder when the cmdline is
-      entered.
-    '';
-
-    modes =
-      helpers.defaultNullOpts.mkListOf
-        (types.enum [
-          "/"
-          "?"
-          ":"
-        ])
-        [
-          "/"
-          "?"
-        ]
-        "List of modes that wilder will be active in.";
-
-    wildcharm =
-      helpers.defaultNullOpts.mkNullable (with types; either str (enum [ false ])) "&wildchar"
-        ''
-          Key to set the 'wildcharm' option to.
-          Can be set to v:false to skip the setting.
-        '';
-
-    nextKey = mkKeyOption "<Tab>" ''
-      A key to map to `wilder#next()` providing next suggestion.
-    '';
-
-    prevKey = mkKeyOption "<S-Tab>" ''
-      A key to map to `wilder#prev()` providing previous suggestion.
-    '';
-
-    acceptKey = mkKeyOption "<Down>" ''
-      Mapping to bind to `wilder#accept_completion()`.
-    '';
-
-    rejectKey = mkKeyOption "<Up>" ''
-      Mapping to bind to `wilder#reject_completion()`.
-    '';
-
-    acceptCompletionAutoSelect = helpers.defaultNullOpts.mkBool true ''
-      The `auto_select` option passed to `wilder#accept_completion()`, if mapped.
-    '';
-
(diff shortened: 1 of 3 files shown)


MattSturgeon:
> This plugin has both a classic `.setup(opts)` configuration pattern and another imperative method to configure other options:

Q: is `set_option` used to override opts passed to `setup` or is it for configuring unrelated options that can't be configured via `setup`?

If it's the former, then I don't think we need to support `set_option`. If it's the latter, then your design sounds fine.

plugins/by-name/wilder/default.nix
```diff
@@ -1,271 +1,53 @@
...
-    package = lib.mkPackageOption pkgs "wilder-nvim" {
-      default = [
-        "vimPlugins"
-        "wilder-nvim"
-      ];
     };
+  };
+  extraConfig = cfg: {
+    plugins.wilder.luaConfig.post = lib.mkIf (cfg.options != { }) ''
+      wilder = require("wilder")
```
MattSturgeon: Should this be a `local`?

```suggestion
      local wilder = require("wilder")
```

plugins/by-name/wilder/default.nix
```diff
@@ -1,271 +1,53 @@
...
-
-    package = lib.mkPackageOption pkgs "wilder-nvim" {
-      default = [
-        "vimPlugins"
-        "wilder-nvim"
-      ];
     };
+  };
+  extraConfig = cfg: {
+    plugins.wilder.luaConfig.post = lib.mkIf (cfg.options != { }) ''
```
MattSturgeon: Generally, I prefer leaving `luaConfig.pre` and `luaConfig.post` empty for end-users. Would it be better to define `luaConfig.content` with a custom `mkOrder` merge prio?

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,112 @@
...
+              v
+            else
+              let
+                # v.fallback can be either a string or a bool.
+                prettyFallback = if lib.isString v.fallback then ''"${v.fallback}"'' else "${v.fallback}";
+              in
+              lib.warn
+                ''
+                  WARNING: the `plugins.wilder.settings.${newOptName}` does not automatically process the provided input.
+                  You will need to provide the value as a list `[ "${v.key}" ${prettyFallback} ]`.
```
MattSturgeon:
```suggestion
                  You will need to provide the value as a list `${lib.generators.toPretty { multiline = false; } [ v.key v.fallback ]}`.
```

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,112 @@
...
+      optionsPath = basePluginPath ++ [ "options" ];
+
+      mkKeymapRename =
+        oldOptName: newOptName:
+        let
+          oldPath = basePluginPath ++ [ oldOptName ];
+          newPath = settingsPath ++ [ newOptName ];
+          convert =
+            v:
+            if (v == null) || (lib.isString v) then
```
MattSturgeon: Should we flip this to `if isAttrs v`, since that is the case we actually care about handling (and all other cases use `v` as-is)

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,112 @@
...
+              newPath = optionsPath ++ [ "pipeline" ];
+            in
+            lib.mkChangedOptionModule oldPath newPath (
+              config:
+              let
+                oldValue = lib.getAttrFromPath oldPath config;
+              in
+              if lib.isList oldValue then
+                lib.warn ''
+                  WARNING: `plugins.wilder.options.pipeline` will not convert the list elements raw lua strings.
```
MattSturgeon:
```suggestion
                  WARNING: `plugins.wilder.options.pipeline` will not convert list elements to raw lua strings.
```

plugins/by-name/wilder/deprecations.nix
GaetanLepage: Good idea. Done.

plugins/by-name/wilder/deprecations.nix
GaetanLepage: Neat.

plugins/by-name/wilder/default.nix
GaetanLepage: I ended up setting `callSetup` to `false` and explicitly writing all the lua config in `luaConfig.content`.

plugins/by-name/wilder/default.nix
GaetanLepage: Yes it should. Fixed.

GaetanLepage:
> Q: is `set_option` used to override opts passed to `setup` or is it for configuring unrelated options that can't be configured via `setup`?
> 
> If it's the former, then I don't think we need to support `set_option`. If it's the latter, then your design sounds fine.

Unfortunately, it seems to be the latter. Otherwise I would have gone with a `settings`-only approach.

[Their README.md has several examples that use the two entry points simultaneously.](https://github.com/gelguy/wilder.nvim/?tab=readme-ov-file#customising-the-pipeline)

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,111 @@
...
+            v:
+            if lib.isAttrs v then
+              let
+                newValue = [
+                  v.key
+                  v.fallback
+                ];
+              in
+              lib.warn ''
+                WARNING: the `plugins.wilder.settings.${newOptName}` does not automatically process the provided input.
```
MattSturgeon:
```suggestion
                WARNING: `${lib.showOption newPath}` will not automatically process the provided input.
```

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,111 @@
...
+          newPath = optionsPath ++ [ newOptName ];
+        in
+        lib.mkChangedOptionModule oldPath newPath (
+          config:
+          let
+            oldValue = lib.getAttrFromPath oldPath config;
+          in
+          if lib.isString oldValue then
+            lib.warn ''
+              WARNING: `plugins.wilder.options.${newOptName}` will not convert the value to a raw lua string.
```
MattSturgeon:
```suggestion
              WARNING: `${lib.showOption newPath}` will not convert the value to a raw lua string.
```

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,111 @@
...
+            '' lib.nixvim.mkRaw oldValue
+          else
+            oldValue
```
MattSturgeon: A slightly more functional way to write this is:

```nix
(
  if lib.isString oldValue then
    lib.warn "" lib.nixvim.mkRaw
  else
    lib.id
)
oldValue
```

plugins/by-name/wilder/deprecations.nix
```diff
@@ -0,0 +1,111 @@
...
+              newPath = optionsPath ++ [ "pipeline" ];
+            in
+            lib.mkChangedOptionModule oldPath newPath (
+              config:
+              let
+                oldValue = lib.getAttrFromPath oldPath config;
+              in
+              if lib.isList oldValue then
+                lib.warn ''
+                  WARNING: `plugins.wilder.options.pipeline` will not convert list elements to raw lua strings.
```
MattSturgeon:
```suggestion
                  WARNING: `${lib.showOption newPath}` will not convert list elements to raw lua strings.
```

plugins/by-name/wilder/deprecations.nix
GaetanLepage: Neat! Done.

GaetanLepage:
Thanks for the thorough review @MattSturgeon!

PR #3820, opened 2025-10-14
plugins/nvim-osc52: remove module

saygo-png:
The plugin is obsolete as the feature provided by it has been added to Neovim itself.

I don't think a deprecation is needed because the plugin has had a warning for 9 months about it's coming demise.
https://github.com/nix-community/nixvim/blob/69d72bb82d7ba1904df47cf092ce1d4c88448001/plugins/by-name/nvim-osc52/default.nix#L78-L84

related https://github.com/nix-community/nixvim/issues/2638 as this plugin is on the list currently


khaneliman:
Should add it to removed though https://github.com/nix-community/nixvim/pull/3787

MattSturgeon:
> I don't think a deprecation is needed because the plugin has had a warning for 9 months about it's coming demise.

I agree it is not _as_ critical when we've had a warning in place, but I'd still prefer to see a removal entry in `plugins/deprecation.nix`.

plugins/deprecation.nix
```diff
@@ -23,6 +23,12 @@ let
       The `rust-tools` project has been abandoned.
       It is recommended to use `rustaceanvim` instead.
     '';
+
+    # Added 2025-10-14
+    nvim-osc52 = ''
+      The `nvim-osc52` plugin is obsolete.
+      The functionality it provided is natively supported since Neovim 0.10.0.
+    '';
```
MattSturgeon, line 32: Should we match the original warning?

```suggestion
      See `:h clipboard-osc52` for more details:
      https://neovim.io/doc/user/provider.html#clipboard-osc52
    '';
```

PR #3823, opened 2025-10-17
plugins/eyeliner: init

axelkar:
> Move faster with unique f/F indicators.

https://github.com/jinh0/eyeliner.nvim

### DIFF
--- a/plugins/by-name/eyeliner/default.nix
+++ b/plugins/by-name/eyeliner/default.nix
@@ -1,0 +1,17 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "eyeliner";
+  package = "eyeliner-nvim";
+  description = "Move faster with unique f/F indicators";
+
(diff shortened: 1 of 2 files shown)


plugins/by-name/eyeliner/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  maintainers = [ lib.maintainers.axka ];
+
+  settingsExample = lib.literalExpression ''
+    {
+      -- show highlights only after keypress
+      highlight_on_key = true,
+      -- dim all other characters if set to true
+      dim = true,
+    }
+  '';
```
axelkar, line 16: I just noticed. Should this be Lua or Nix code?

MattSturgeon: It should be nix code:
```suggestion
  settingsExample = lib.literalExpression ''
    {
      # Show highlights only after keypress
      highlight_on_key = true;
      # Dim all other characters if set to true
      dim = true;
    }
  '';
```

PR #3825, opened 2025-10-17
plugins/treesitter-textobjects: migrate to mkNeovimPlugin

GaetanLepage:
Tracking: #2638

### DIFF
--- a/plugins/by-name/treesitter-textobjects/deprecations.nix
+++ b/plugins/by-name/treesitter-textobjects/deprecations.nix
@@ -1,0 +1,102 @@
+lib: {
+  imports =
+    let
+      basePluginPath = [
+        "plugins"
+        "treesitter-textobjects"
+      ];
+      settingsPath = basePluginPath ++ [ "settings" ];
+
+      mkChangedKeymapOption =
+        oldSubPath: newSubPath:
+        let
+          oldPath = basePluginPath ++ oldSubPath;
+          newPath = settingsPath ++ newSubPath;
+
+          processKeymaps = lib.mapAttrs (
+            key: mapping:
+            if lib.isString mapping then
+              mapping
+            else
+              lib.warn
+                ''
+                  WARNING: `${lib.showOption newPath}` will expect `query_group` instead of `queryGroup`.
+                ''
+                {
+                  inherit (mapping) query;
+                  query_group = mapping.queryGroup;
+                  inherit (mapping) desc;
+                }
+          );
+        in
+        lib.mkChangedOptionModule oldPath newPath (
+          config:
+          let
+            oldValue = lib.getAttrFromPath oldPath config;
+          in
+          (if oldValue == null then lib.id else processKeymaps) oldValue
+        );
+    in
+    [
+      (
+        let
+          oldPath = basePluginPath ++ [
+            "select"
+            "includeSurroundingWhitespace"
+          ];
+          newPath = settingsPath ++ [
+            "select"
+            "include_surrounding_whitespace"
+          ];
+        in
+        lib.mkChangedOptionModule oldPath newPath (
+          config:
+          let
+            oldValue = lib.getAttrFromPath oldPath config;
+          in
+          (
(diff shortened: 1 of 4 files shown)


tests/test-sources/plugins/by-name/treesitter-textobjects/default.nix
```diff
@@ -12,71 +12,73 @@
...
-            af = "@function.outer";
-            "if" = "@function.inner";
-            ac = "@class.outer";
-            ic = {
-              query = "@class.inner";
-              desc = "Select inner part of a class region";
+        settings = {
+          select = {
+            enable = true;
+            disable = [ ];
```
saygo-png, line 18: Shouldn't we be using actual empty tables? As i understand the original is equal to not having this line.
```suggestion
            disable.__empty = { };
```

GaetanLepage: Technically, you are right. And you could rightfully make the same claim for `null`.
The reality is that this error is common across the entire test suite.

Maybe we should go and do a tree-wide change to use `__empty` and `__raw "nil"`.

saygo-png: > this error is common across the entire test suite.

I worded it as a question because it is so common that I wasn't sure if I was correct.

> Maybe we should go and do a tree-wide change to use __empty and __raw "nil".

This seems pretty fun to me. I'll see if i can modify my WIP Nix linter to do this automatically.

plugins/by-name/treesitter-textobjects/deprecations.nix
```diff
@@ -0,0 +1,105 @@
...
+          config:
+          let
+            oldValue = lib.getAttrFromPath oldPath config;
+          in
+          (
+            if oldValue == null then
+              lib.id
+            else
+              lib.warn ''
+                WARNING: `${lib.showOption newPath}` will not convert the sub-option names from camelCase to snake_case.
```
khaneliman: Not sure I understand the warning here. Aren't the new options being converted manually to snake_case in what you pass in below?

GaetanLepage: Not the sub-options. Each keymap value can be either a string or an attrs.
In the latter, we used to map `queryGroup` to `query_group`. Users should not blindly copy-pase
```nix
select.keymaps.foo = {
  query  = "X";
  queryGroup = "Y";
  desc = "Z";
};
```
to
```nix
settings.select.keymaps.foo = {
  query  = "X";
  queryGroup = "Y";
  desc = "Z";
};
```
but to
```nix
settings.select.keymaps.foo = {
  query  = "X";
  query_group = "Y";
  desc = "Z";
};
```

MattSturgeon: Should this warning be predicated on `isAttrs` instead of `!= null`?

GaetanLepage: I moved them so that they would trigger only when the keymap value is actually an attrs.

plugins/by-name/treesitter-textobjects/deprecations.nix
```diff
@@ -0,0 +1,102 @@
...
+          newPath = settingsPath ++ newSubPath;
+
+          processKeymaps = lib.mapAttrs (
+            key: mapping:
+            if lib.isString mapping then
+              mapping
+            else
+              lib.warn
+                ''
+                  WARNING: `${lib.showOption newPath}` will not convert the sub-option names from camelCase to snake_case.
```
MattSturgeon: Seeing as this only applies to `query_group`, perhaps we could be more specific in this warning?


```suggestion
                  WARNING: `${lib.showOption newPath}` will expect `query_group` instead of `queryGroup`.
```

GaetanLepage: Done!

PR #3829, opened 2025-10-18
plugins/treesitter: fix folding option

saygo-png:
This snippet in main is out of date and does not work.
https://neovim.io/doc/user/treesitter.html#vim.treesitter.foldexpr()


MattSturgeon (approved):
LGTM, and matches the linked docs. But I don't understand what `v:` is or where it's documented. Looks like some way to tell `:set` that this is a lua value?

saygo-png:
> But I don't understand what v: is

I don't either. Lua confuses me greatly. I just know this snippet works and the old one doesn't.

PR #3832, opened 2025-10-18
treewide: generate JSON instead of Nix

saygo-png:
https://github.com/nix-community/nixvim/pull/3682#discussion_r2347275971

It also cleans up the generate script a bit so we no longer have to repeat the name and the files always have the .json extension even in the nix derivation.

Didn't touch the maintainers script because I can't nor want to read Python.


ci/conform-nvim.nix
```diff
@@ -9,5 +9,5 @@ lib.pipe "${vimPlugins.conform-nvim}/lua/conform/formatters" [
   (builtins.filter (lib.hasSuffix ".lua"))
   (map (lib.removeSuffix ".lua"))
   builtins.toJSON
-  (writeText "conform-formatters")
+  (writeText "conform-formatters.json")
```
MattSturgeon: Should we consider [`writers.writeJSON`](https://github.com/NixOS/nixpkgs/blob/e067934d171a7d69737714fde41c9a9f87b4fa6e/pkgs/build-support/writers/data.nix#L67-L76) instead of `toJSON` + `writeText`?

It's an alias of `(pkgs.formats.json { }).generate`.

ci/efmls-configs.nix
```diff
@@ -54,8 +54,4 @@ let
       };
     };
 in
-writeText "efmls-configs-sources.nix" (
-  "# WARNING: DO NOT EDIT\n"
-  + "# This file is generated with packages.<system>.efmls-configs-sources, which is run automatically by CI\n"
-  + (lib.generators.toPretty { } sources)
-)
+writeText "efmls-configs-sources.json" (builtins.toJSON sources)
```
MattSturgeon: Same: `writeJSON`

ci/generate.nix
```diff
@@ -33,25 +33,21 @@ writeShellApplication {
...
-      nixfmt "$generated_dir/$2.nix"
-    }
-
     generate_json() {
-      echo "$2"
-      prettier --parser=json "$1" >"$generated_dir/$2.json"
+      # Get file part from filepath and remove hash prefix
+      output_name=$(basename "$1" | cut -d "-" -f "2-")
+      basename "$1" ".json"
+      prettier "$1" > "$generated_dir/$output_name"
```
MattSturgeon, line 40: Aside: directly calling `prettier` in the update-script probably isn't the best design. Really, we should be using our `treefmt` configuration, so that we also get any settings (etc). But doing that without having to evaluate the rest of nixvim's flake may be non-trivial.

ci/none-ls.nix
```diff
@@ -3,12 +3,9 @@
...
-writeText "none-ls-sources.nix" (
-  "# WARNING: DO NOT EDIT\n"
-  + "# This file is generated with packages.<system>.none-ls-builtins, which is run automatically by CI\n"
-  + (lib.generators.toPretty { } builtinSourceNames)
-)
+lib.pipe "${vimPlugins.none-ls-nvim.src}/doc/builtins.json" [
+  lib.importJSON
+  (lib.mapAttrs (_: lib.attrNames))
+  builtins.toJSON
+  (writeText "none-ls-sources.json")
```
MattSturgeon: Same: `writeJSON`

ci/rust-analyzer/default.nix
```diff
@@ -227,8 +227,4 @@ let
     }
   ) rustAnalyzerProperties;
 in
-writeText "rust-analyzer-options.nix" (
-  "# WARNING: DO NOT EDIT\n"
-  + "# This file is generated with packages.<system>.rust-analyzer-options, which is run automatically by CI\n"
-  + (lib.generators.toPretty { } (lib.mergeAttrsList rustAnalyzerOptions))
-)
+writeText "rust-analyzer-options.json" (builtins.toJSON (lib.mergeAttrsList rustAnalyzerOptions))
```
MattSturgeon: Same: `writeJSON`

ci/generate.nix
saygo-png: I agree but out of scope for this PR. Though in practice our treefmt prettier is the same as this one for JSON files but of course this would help keep it in sync in the future.

ci/conform-nvim.nix
```diff
@@ -1,13 +1,11 @@
...
-  writeText,
 }:
 lib.pipe "${vimPlugins.conform-nvim}/lua/conform/formatters" [
   builtins.readDir
   builtins.attrNames
   (builtins.filter (lib.hasSuffix ".lua"))
   (map (lib.removeSuffix ".lua"))
-  builtins.toJSON
-  (writeText "conform-formatters")
+  (lib.writers.writeJSON "conform-formatters.json")
```
MattSturgeon: `writers` is part of `pkgs`, not `lib`. We should get it from the callPackage args above.

```suggestion
  (writers.writeJSON "conform-formatters.json")
```

ci/conform-nvim.nix
```diff
@@ -1,13 +1,12 @@
 {
   vimPlugins,
+  pkgs,
```
MattSturgeon:
```suggestion
  writers,
```

ci/conform-nvim.nix
saygo-png: Sorry I forgot CI doesn't run this. For some reason this writer creates slightly differently formatted JSON than writeText + toJSON and prettier is input dependant so it doesnt normalize it.

ci/conform-nvim.nix
```diff
@@ -1,13 +1,12 @@
...
-  writeText,
 }:
 lib.pipe "${vimPlugins.conform-nvim}/lua/conform/formatters" [
   builtins.readDir
   builtins.attrNames
   (builtins.filter (lib.hasSuffix ".lua"))
   (map (lib.removeSuffix ".lua"))
-  builtins.toJSON
-  (writeText "conform-formatters")
+  (pkgs.writers.writeJSON "conform-formatters.json")
```
MattSturgeon:
```suggestion
  (writers.writeJSON "conform-formatters.json")
```

MattSturgeon (commented):
We should avoid taking `pkgs` as an argument in callPackage functions.

ci/conform-nvim.nix
MattSturgeon: > For some reason this writer creates slightly differently formatted JSON than writeText + toJSON

Makes sense. It is using `jq` to normalise the JSON: https://github.com/NixOS/nixpkgs/blob/852841c072353c343ba2fdc70d02469da3be886e/pkgs/pkgs-lib/formats.nix#L151-L164

> prettier is input dependant

That is cursed... Are you saying that `prettier` and `treefmt` are now producing different output?

I'm fine with the formatting being different from before, so long as `treefmt` remains idempotent.

saygo-png: > Are you saying that prettier and treefmt are now producing different output?

No! `writers.writeJSON` and `writeText (toJSON)` produce differently formatted JSON. So the diff on the generation changed just now. Prettier and treefmt produce the same output. I was just mentioning that despite both being formatted the formatting changed, because prettier output depends on the initial formatting.

MattSturgeon (approved):

MattSturgeon:
> cancelled (Master Shutdown)
> -- https://buildbot.nix-community.org/#/builders/26/builds/7198

Sounds like buildbot is rebooting

zowoq:
> Sounds like buildbot is rebooting

Yes, sorry. I'll re-queue this PR.

PR #3835, opened 2025-10-19
plugins/lazy: migrate to mkNeovimPlugin

GaetanLepage:
In this PR, I make sure to keep the exact same API as the previous implementation.
This change does not affect users.

I plan to open a follow-up PR to investigate introducing a `settings` option.

- **plugins/lazy: remove old deprecation warning**
- **plugins/lazy: migrate to mkNeovimPlugin**

Tracking: #2638

### DIFF
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -1,161 +1,134 @@
 {
   lib,
-  helpers,
-  config,
   pkgs,
   ...
 }:
-with lib;
-let
-  cfg = config.plugins.lazy;
-  lazyPlugins = cfg.plugins;
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lazy";
+  package = "lazy-nvim";
+  maintainers = [ ];
 
-  processPlugin =
-    plugin:
+  hasSettings = false;
+  callSetup = false;
+
+  dependencies = [
+    "git"
+  ];
+
+  extraOptions =
     let
-      mkEntryFromDrv =
-        p:
-        if lib.isDerivation p then
-          {
-            name = "${lib.getName p}";
-            path = p;
-          }
-        else
-          {
-            name = "${lib.getName p.pkg}";
-            path = p.pkg;
-          };
-      processDependencies =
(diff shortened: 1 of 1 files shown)


khaneliman:
Just curious what we gain from the current migration here. Seems like it doesn't provide the freeform settings config or expanded extra options etc. How does it differ from what was started in https://github.com/nix-community/nixvim/pull/2082 or would it be better to take over the effort on a branch that is further along in that refactoring like that work?

GaetanLepage:
> Just curious what we gain from the current migration here. Seems like it doesn't provide the freeform settings config or expanded extra options etc. How does it differ from what was started in #2082 or would it be better to take over the effort on a branch that is further along in that refactoring like that work?

TBH, I started this without realizing that there was an ongoing effort.
Also, I wanted to keep this PR straightforward and to then open a follow-up to introduce `plugins.lazy.settings`.

MattSturgeon:
> I plan to open a follow-up PR to investigate introducing a `settings` option.

Good plan. The prior migration attempt stalled because doing it all at once proved too complex.

MattSturgeon:
> Just curious what we gain from the current migration here.

My initial reaction was that it'll make the diff easier to review if one PR does the big restructure and another PR does the `settings` migration.

plugins/pluginmanagers/lazy.nix
```diff
@@ -1,169 +1,134 @@
...
-            path = p.pkg;
-          };
-      processDependencies =
-        if plugin ? dependencies && plugin.dependencies != null then
-          builtins.concatMap processPlugin plugin.dependencies
-        else
-          [ ];
-    in
-    [ (mkEntryFromDrv plugin) ] ++ processDependencies;
+  hasSettings = false;
```
MattSturgeon, line 11: We could signal our intent by flagging `hasSettings` with a FIXME or TODO.

MattSturgeon (approved):
Diff LGTM on first pass

plugins/pluginmanagers/lazy.nix
GaetanLepage: Already opened https://github.com/nix-community/nixvim/pull/3844 ;)

PR #3839, opened 2025-10-21
plugins/tailwind-tools: throw warning when enabling

DerHalbGrieche:
This pull request adds a warning to the `tailwind-tools` plugin configuration to inform users about the plugin's upstream status and potential deprecation issues.

Configuration improvements:

* Added an `extraConfig` block to display a warning when the `tailwind-tools` plugin is enabled, notifying users that the upstream GitHub project is archived and that enabling the plugin triggers an `lspconfig` deprecation warning. Users are advised to consider disabling the plugin or switching to an alternative.

### DIFF
--- a/plugins/by-name/tailwind-tools/default.nix
+++ b/plugins/by-name/tailwind-tools/default.nix
@@ -101,6 +101,12 @@ lib.nixvim.plugins.mkNeovimPlugin {
     };
   };
 
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.tailwind-tools" {
+      message = ''The upstream GitHub project for tailwind-tools is archived, and enabling this plugin triggers an lspconfig deprecation warning. Consider disabling it or switching to an alternative.'';
+    };
+  };
+
   settingsExample = {
     document_color = {
       conceal = {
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):
The warning was breaking the tests, so I [pushed](https://github.com/nix-community/nixvim/compare/aa402a25913a06e655a8dda37f297b16e63d5d44..30674eeaf82b02b2ece3695b566db66d6780c308) a fix for that. LTGM, thanks!

plugins/by-name/tailwind-tools/default.nix
```diff
@@ -101,6 +101,12 @@ lib.nixvim.plugins.mkNeovimPlugin {
     };
   };
 
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.tailwind-tools" {
+      message = ''The upstream GitHub project for tailwind-tools is archived, and enabling this plugin triggers an lspconfig deprecation warning. Consider disabling it or switching to an alternative.'';
+    };
+  };
+
```
khaneliman, line 109: Why not just use the built in deprecations module we created for this?

MattSturgeon: I forgot that supports deprecation; I thought it was only renames and removals.

PR #3846, opened 2025-10-23
plugins/coq-thirdparty: migrate to mkNeovimPlugin

GaetanLepage:
I suggest to keep the `sources` option around.

Tracking: #2638

### DIFF
--- a/flake/dev/list-plugins/list-plugins.py
+++ b/flake/dev/list-plugins/list-plugins.py
@@ -69,9 +69,7 @@ KNOWN_PATHS: dict[
         Kind,  # Vim / Neovim / misc
         bool,  # Has deprecation warnings
     ],
-] = {
-    "plugins/by-name/coq-thirdparty/default.nix": (State.OLD, Kind.NEOVIM, False),
-}
+] = {}
 for telescope_extension_name, has_depr_warnings in {
     "advanced-git-search": False,
     "file-browser": True,
--- a/plugins/by-name/coq-thirdparty/default.nix
+++ b/plugins/by-name/coq-thirdparty/default.nix
@@ -1,71 +1,33 @@
-{
-  lib,
-  config,
-  pkgs,
-  ...
-}:
-with lib;
-let
-  cfg = config.plugins.coq-thirdparty;
-in
-{
-  options.plugins.coq-thirdparty = {
-    enable = mkEnableOption "coq-thirdparty";
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "coq-thirdparty";
+  maintainers = [ lib.maintainers.GaetanLepage ];
 
-    package = lib.mkPackageOption pkgs "coq-thirdparty" {
-      default = [
-        "vimPlugins"
+  # The lua setup call is: `require('coq_3p')({...})`
+  moduleName = "coq_3p";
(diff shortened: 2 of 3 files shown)


plugins/by-name/coq-thirdparty/default.nix
```diff
@@ -1,71 +1,84 @@
...
-        { src = "demo"; }
-      ];
+          }
+        );
+        # Convert
+        #   [ { src = "foo"; } { src = "bar"; } ]
+        # to
+        #   { __unkeyed-1.src = "foo"; __unkeyed-2.src = "bar"; }
+        # so that this option value can directly be provided to `plugins.coq-thirdparty.settings`
+        apply = lib.nixvim.listToUnkeyedAttrs;
```
khaneliman: Didn't even realize we had this... I like this..

khaneliman (approved):

plugins/by-name/coq-thirdparty/default.nix
MattSturgeon: I'm not keen on having this transformation directly in the option's `apply`. I often find this confusing for end-users who usually expect to be able to read an option the same way they define it.

Personally, I'd do the transformation when defining `plugins.coq-thirdparty.settings`, instead.

MattSturgeon:
This is another plugin where we'd benefit from `settings` not needing to be a submodule, so that we can type it as `either listOf attrsOf`.

GaetanLepage:
> This is another plugin where we'd benefit from `settings` not needing to be a submodule, so that we can type it as `either listOf attrsOf`.

Yes!

plugins/by-name/coq-thirdparty/default.nix
GaetanLepage: I hesitated between both approaches. I'm fine with your suggestion. Updated.

GaetanLepage:
Much much better!

plugins/by-name/coq-thirdparty/default.nix
```diff
@@ -1,71 +1,32 @@
...
-        {
-          src = "nvimlua";
-          short_name = "nLUA";
-        }
-        {
-          src = "vimtex";
-          short_name = "vTEX";
-        }
-        { src = "demo"; }
+  imports =
```
MattSturgeon, line 11: Do you want a date comment on these renames?

GaetanLepage: Definitely. Done!

flake/dev/list-plugins/list-plugins.py
```diff
@@ -69,9 +69,7 @@ class State(Enum):
         Kind,  # Vim / Neovim / misc
         bool,  # Has deprecation warnings
     ],
-] = {
-    "plugins/by-name/coq-thirdparty/default.nix": (State.OLD, Kind.NEOVIM, False),
-}
+] = {}
```
MattSturgeon, line 72: `KNOWN_PATHS` is now empty. Should it be removed?

MattSturgeon (approved):

flake/dev/list-plugins/list-plugins.py
GaetanLepage: It is being filled with the telescope extensions below.

PR #3847, opened 2025-10-23
plugins/efmls-configs: cleanup

GaetanLepage:
- **plugins/efmls-configs: use the new lsp module**
- **plugins/efmls-configs: use dependencies for efm-langserver**

### DIFF
--- a/plugins/by-name/efmls-configs/default.nix
+++ b/plugins/by-name/efmls-configs/default.nix
@@ -20,10 +20,6 @@ in
       ];
     };
 
-    efmLangServerPackage = lib.mkPackageOption pkgs "efm-langserver" {
-      nullable = true;
-    };
-
     externallyManagedPackages = lib.mkOption {
       type = with lib.types; either (enum [ "all" ]) (listOf str);
       description = ''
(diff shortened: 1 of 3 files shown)


plugins/by-name/efmls-configs/default.nix
```diff
@@ -194,16 +190,25 @@ in
...
 
-      plugins.lsp.servers.efm = {
+      lsp.servers.efm = {
         enable = true;
-        extraOptions.settings.languages = setupOptions;
+        config.settings.languages = setupOptions;
       };
 
-      extraPackages = [ cfg.efmLangServerPackage ] ++ (map (v: cfg.toolPackages.${v}) nixvimPkgs.right);
+      extraPackages = map (v: cfg.toolPackages.${v}) nixvimPkgs.right;
```
MattSturgeon:
```suggestion
      extraPackages = map (name: cfg.toolPackages.${name}) nixvimPkgs.right;
```

plugins/by-name/efmls-configs/default.nix
```diff
@@ -20,10 +20,6 @@ in
       ];
     };
 
-    efmLangServerPackage = lib.mkPackageOption pkgs "efm-langserver" {
-      nullable = true;
-    };
```
MattSturgeon, line 25: I don't see where we handle _enabling_ `dependencies.efm-langserver`.

plugins/by-name/efmls-configs/default.nix
GaetanLepage: Done.

plugins/by-name/efmls-configs/default.nix
GaetanLepage: Oops, forgot it. Fixed!

PR #3849, opened 2025-10-24
plugins/efmls-configs: migrate to mkNeovimPlugin

GaetanLepage:
No user-facing changes.

Tracking: #2638

### DIFF
--- a/plugins/by-name/efmls-configs/default.nix
+++ b/plugins/by-name/efmls-configs/default.nix
@@ -1,114 +1,128 @@
 {
   lib,
-  helpers,
   config,
   pkgs,
   ...
 }:
-let
-  tools = lib.importJSON ../../../generated/efmls-configs-sources.json;
-  inherit (import ./packages.nix lib) packaged;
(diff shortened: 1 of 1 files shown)


plugins/by-name/efmls-configs/default.nix
```diff
@@ -1,114 +1,128 @@
...
+            formatter = ["clang-format" "uncrustify"];
+          };
+          go = {
+            linter = ["djlint" "golangci_lint"];
+          };
         }
-      ))
-    ];
+      */
+      setup = lib.mkOption {
```
MattSturgeon, line 80: Q: why are we still using a disconnected `setup` and `setupOptions` pattern instead of direct freeform `settings`?

GaetanLepage: I agree, it should be possible to replace that with a proper `settings` option. Maybe we could merge this PR as is and I'll open a follow-up (same as I did in the `lazy` case).

PR #3850, opened 2025-10-24
lib/options (mkSettingsOption): allow more types for settingsOption when no sub-options are explicitly declared

GaetanLepage:
Currently, all `settings` options are typed as `types.submodule`.
This allows to explicitly define sub-options, as explained in [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md).

However, we recently moved towards declaring as few sub-options as possible.
Besides, certain plugins' `setup` function allow for a non-attrs-like argument (e.g. list, string).

This PR types the `settings` option returned by `mkSettingsOption` as `lib.nixvim.lua-types.anything` if no sub-option was passed to the function. Further, it switches the submodule freeformType from `attrsOf anything` to `attrsOf lua-types.anything`.

This is the first time we're actually using `lua-types`, added back in #3213

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -329,14 +329,22 @@ rec {
       options ? { },
       description,
       example ? null,
+      # If no sub-options are explicitly declared, settings do not need to be a submodule.
+      submoduleType ? options != { },
     }:
     lib.mkOption {
       type =
-        with types;
-        submodule {
-          freeformType = attrsOf lib.nixvim.lua-types.anything;
-          inherit options;
-        };
+        let
+          anyLuaType = lib.nixvim.lua-types.anything;
+        in
+        if submoduleType then
+          types.submodule {
+            freeformType = types.attrsOf anyLuaType;
+            inherit options;
+          }
+        else
+          assert options == { };
+          anyLuaType;
       default = { };
       inherit description;
       example =
--- a/plugins/lsp/language-servers/_mk-lsp.nix
+++ b/plugins/lsp/language-servers/_mk-lsp.nix
@@ -179,6 +179,12 @@ in
       settings = lib.nixvim.mkSettingsOption {
         description = "The settings for this LSP.";
         options = settingsOptions;
+
+        # Some servers declare settings sub-options without using `settingsOptions`.
+        # This leads the `settings` option to not be typed as a submodule.
+        # Hence, we force all `plugins.lsp.<name>.settings` options to be types as submodules.
+        # FIXME This is not ideal, but `plugins.lsp` will be dropped entirely in a few months.
+        submoduleType = true;
       };
 
       extraOptions = mkOption {


lib/options.nix
```diff
@@ -333,10 +333,19 @@ rec {
...
-          freeformType = attrsOf anything;
-          inherit options;
-        };
+        # If no sub-options are explicitly declared, settings do not need to be an attrs.
+        if options == { } then
+          oneOf [
+            (attrsOf anything)
+            (listOf anything)
+            rawLua
+            str
```
MattSturgeon: Why str?

GaetanLepage: https://github.com/nix-community/nixvim/blob/af57bad10d71ca49dd5d81afcca17451cb732069/plugins/by-name/alpha/default.nix#L201-L210

:')

MattSturgeon:
>        error: The option `plugins.lsp.servers.pylsp.settings' in module `plugins/lsp/language-servers' would be a parent of the following options, but its type `(attribute set of anything) or list of anything or raw lua code or string' does not support nested options.
>         - option(s) with prefix `plugins.lsp.servers.pylsp.settings.configurationSources' in module `plugins/lsp/language-servers/pylsp.nix'
>         - option(s) with prefix `plugins.lsp.servers.pylsp.settings.plugins' in module `plugins/lsp/language-servers/pylsp.nix'
>         - option(s) with prefix `plugins.lsp.servers.pylsp.settings.rope' in module `plugins/lsp/language-servers/pylsp.nix'

Looks like option-declaration merging is going on here; where options can be defined outside of a submodule and the module system will merge the declarations into the submodule.

This may be us defining options in multiple locations, or this may be coming from `mkRenameOptionModule`s.

The fact it's exclusive to `pylsp` makes me think it is not related to `mkRenameOptionModule`.

MattSturgeon:
> Looks like option-decl merging is going on here; where options can be defined outside of a submodule and the module system will merge there declaration into the submodule.

Yep: https://github.com/nix-community/nixvim/blob/af57bad10d71ca49dd5d81afcca17451cb732069/plugins/lsp/language-servers/pylsp.nix#L18-L20

plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -179,6 +179,7 @@ in
       settings = lib.nixvim.mkSettingsOption {
         description = "The settings for this LSP.";
         options = settingsOptions;
+        submoduleType = true;
```
MattSturgeon, line 187: Can we do this in the specific LSP servers, rather than all of them?

lib/options.nix
MattSturgeon: That feels hyper-sepcific to that one plugin. Maybe we could be able to override or extend the type on a per-plugin basis, so that we don't have to allow `str` in all plugins' settings option.

Either that, or we go straight to using our `luaType` now, which already lists all possible primitive types.

MattSturgeon: > Either that, or we go straight to using our `luaType` now, which already lists all possible primitive types.

https://github.com/nix-community/nixvim/blob/af57bad10d71ca49dd5d81afcca17451cb732069/lib/lua-types.nix#L18-L21

GaetanLepage: > Either that, or we go straight to using our `luaType` now, which already lists all possible primitive types.

Much much cleaner indeed.

plugins/lsp/language-servers/_mk-lsp.nix
GaetanLepage: We could, but at the same time, I would argue that it makes sense to have `submodule` `settings` options for LSP servers.

MattSturgeon: > I would argue that it makes sense to have `submodule` `settings` options for LSP servers

I don't follow why this is the case for LSPs but not plugins.

GaetanLepage: I don't have a strong opinion on this, but my reasoning was the following:
- It would be slightly annoying to set `submoduleType = true` for each server that needs it.
- I cannot think of an example of a language server that is configured with something else than an attrs.

Whereas for plugins, there are exceptions.

But I don't mind, if you prefer declaring on the server-basis, that's perfectly fine by me.

lib/options.nix
```diff
@@ -329,14 +329,20 @@ rec {
...
     lib.mkOption {
       type =
         with types;
-        submodule {
-          freeformType = attrsOf anything;
-          inherit options;
-        };
+        if submoduleType then
+          submodule {
+            freeformType = attrsOf anything;
```
MattSturgeon: If this PR is changing some plugins to `lib.nixvim.lua-types.anything`, I think it may make sense to change all of the:


```suggestion
            freeformType = attrsOf lib.nixvim.lua-types.anything;
```

Arguably, this should happen in a prior commit:
1. use nixvim's lua-types for `settings`
2. avoid submodule for `settings` with no sub-options

plugins/lsp/language-servers/_mk-lsp.nix
MattSturgeon: I'm fine with it in this PR if the comment is framed as a TODO or FIXME. Especially as we plan to drop `plugins.lsp` entirely.

lib/options.nix
GaetanLepage: Good idea. Done.

plugins/lsp/language-servers/_mk-lsp.nix
GaetanLepage: I added a `FIXME`.

MattSturgeon (approved):
I could nitpick the commit messages (the summary lines are way too long), but the PR LGTM. Great to see `lua-types` finally being used in-tree! Thanks!

GaetanLepage:
> (the summary lines are way too long),

Oh yes, I'm so bad at this ^^

PR #3853, opened 2025-10-24
plugins/alpha: remove old deprecation warning

GaetanLepage:


plugins/by-name/alpha/default.nix
```diff
@@ -147,37 +139,8 @@ in
...
-                config.plugins.mini.enable
-                && config.plugins.mini.modules ? icons
-                && config.plugins.mini.mockDevIcons
-              )
-              || (config.plugins.mini-icons.enable && config.plugins.mini-icons.mockDevIcons)
-            )
-          )
-          {
-            enable = lib.mkOverride 1490 true;
-          };
```
MattSturgeon, line 179: What was this replaced with?

GaetanLepage: Like for other plugins, the user is responsible for enabling the right plugins. Isn't that our current approach to icon dependencies?

MattSturgeon: So was it replaced with a warning? An assertion? I think that's what we have in most other plugins.

GaetanLepage: For the past few months, users who enabled `iconsEnabled` had a warning. This PR removes:
- the warning
- the option
- the effect of this option

MattSturgeon: Sure. But when we introduced that warning, surely we replaced the deprecated option with something else? Or does this plugin just not need icons anymore?

GaetanLepage: Not really. I think (I might be wrong) that the intention was that the user manually enabled `devicons` (or `mini-icons`) by himself.

MattSturgeon: If so, what we usually do is assert that it is enabled. Idk how we usually handle that if it's an optional dependency, though.

GaetanLepage: The issue is that it is not easy to check if the use has enabled the support for file icons. And it is optional, which makes it non-mandatory to have an icon provider plugin.

If a user was relying on the option to auto-install an icon provider, he had a warning for the past few months to stop using the option. So at this point, he should have switched to manually enabling an icon-provider plugin.

Hence, I don't think that it is necessary to restart a warning period.

Issue #3728, opened 2025-09-28
[BUG] jupytext not found when running with jupytext-nvim enabled


taranarmo:
### Plugin

jupytext

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

Basically the #2367 happened again: having `jupytext` plugin enabled doesn't add jupytext package to PYTHONPATH.
Similar to above mentioned issue
- setting `programs.nixvim.extraPython3Packages = p: [p.jupytext];` doesn't work
- entering nix shell with `python3Packages.jupytext` does
- adding `python3Packages.jupytext` to `extraPackages` does work too

### Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
  enable = true;
  plugins.jupytext.enable = true;
}
```

PR #3855, opened 2025-10-25
plugins/jupytext: move `jupytext` to PATH

saygo-png:
closes https://github.com/nix-community/nixvim/issues/3728

`jupytext` needs to be in $PATH not the python path — adding it to python3dependencies wasn't doing anything.

This PR also cleans up the plugin.

### DIFF
--- a/plugins/by-name/jupytext/default.nix
+++ b/plugins/by-name/jupytext/default.nix
@@ -9,44 +9,6 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   maintainers = [ lib.maintainers.GaetanLepage ];
 
-  settingsOptions = {
-    style = lib.nixvim.defaultNullOpts.mkStr "hydrogen" ''
-      The jupytext style to use.
-    '';
-
-    output_extension = lib.nixvim.defaultNullOpts.mkStr "auto" ''
-      By default, the extension of the plain text file is automatically selected by jupytext.
-      This can be modified by changing the extension from auto to any other file extension supported
-      by Jupytext.
-      This is most useful to those using Quarto or Markdown.
-      Analogously, we can provide a default filetype that will be given to the new buffer by using
-      `force_ft`.
-      Again, this is only really useful to users of Quarto.
-    '';
-
-    force_ft = lib.nixvim.mkNullOrStr ''
-      Default filetype. Don't change unless you know what you are doing.
-    '';
-
-    custom_language_formatting = lib.nixvim.defaultNullOpts.mkAttrsOf lib.types.anything { } ''
-      By default we use the auto mode of jupytext.
-      This will create a script with the correct extension for each language.
-      However, this can be overridden in a per language basis if you want to.
-      For this you can set this option.
-
-      For example, to convert python files to quarto markdown:
-      ```nix
-        {
-          python = {
-            extension = "qmd";
-            style = "quarto";
-            force_ft = "quarto";
-          };
-        }
-      ```
-    '';
-  };
-
   settingsExample = {
     style = "light";
     output_extension = "auto";


plugins/by-name/jupytext/default.nix
```diff
@@ -62,22 +63,19 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-          jupytext
-        ]
-      '';
-      description = "Python packages to add to the `PYTHONPATH` of neovim.";
+    jupytextPackage = lib.mkPackageOption pkgs "jupytext" {
+      nullable = true;
+      default = [
+        "python313Packages"
+        "jupytext"
+      ];
```
GaetanLepage: We now prefer to use the `dependencies` module.
Add a new `jupytext` dependency in `modules/dependencies.nix`, then add `dependencies = [ "jupytext" ];` in the plugin module.

plugins/by-name/jupytext/deprecations.nix
```diff
@@ -0,0 +1,13 @@
...
+  imports = [
+    # TODO: introduced 2025-10-25: remove after 26.05
+    (lib.mkRenamedOptionModule
+      [
+        "plugins"
+        "jupytext"
+        "python3Dependencies"
+      ]
+      [ "extraPython3Packages" ]
+    )
```
MattSturgeon, line 11: Technically, the old option was gated behind the plugin's `enable` option before, while this alias will copy the option's definitions regardless of whether the plugin is enabled.

saygo-png: Good catch.

Sadly I don't know how to implement this — the `mkRenamedOptionModule` and its underlying `doRename` function don't allow me to customize the `to` value so I can't gate it. I thought `mkChangedOptionModule` would help here but to my knowledge there is no way to to override the warning it sends. The default one says that the type of the option has changed but this is misleading in this case. It's not the type but behaviour that is different.

I could inline `mkChangedOptionModule` with a custom warning but [the code for it is quite messy](https://github.com/NixOS/nixpkgs/blob/d792a6e0cd4ba35c90ea787b717d72410f56dc40/lib/modules.nix#L1690C3-L1690C23), I don't think we want that.

MattSturgeon: We've done this a few times, e.g. in `plugins.lsp` we do some conditional aliases.

We could also just manually forward that value (if defined) and warn if defined. This wouldn't add the additional warn-on-read, but that's not too important here.

However, in this case I think the option is niche enough that we can just live with the alias being slightly incorrect.

saygo-png: I think we just keep it as is — the other options just make the code a lot uglier for a niche option. I think it should be clear enough to the user that `extraPython3Packages` is not going to be gated behind this plugin's enable option as it is toplevel.

As an aside i could not find a conditional alias like that in `plugins.lsp`. We have `_renamed.nix` but those contain simple renames that are not conditional.

GaetanLepage (approved):
LGTM, thank you!
I'll let @MattSturgeon confirm that he's fine with this solution.

plugins/by-name/jupytext/deprecations.nix
MattSturgeon: > As an aside i could not find a conditional alias like that in plugins.lsp.

Almost all LSP options are now conditional aliases, e.g.:

https://github.com/nix-community/nixvim/blob/5da380555961da4f06fd88cccea218db3558def4/plugins/lsp/default.nix#L101-L108

https://github.com/nix-community/nixvim/blob/5da380555961da4f06fd88cccea218db3558def4/plugins/lsp/default.nix#L202

or:
https://github.com/nix-community/nixvim/blob/5da380555961da4f06fd88cccea218db3558def4/plugins/lsp/language-servers/_mk-lsp.nix#L89-L98

https://github.com/nix-community/nixvim/blob/5da380555961da4f06fd88cccea218db3558def4/plugins/lsp/language-servers/_mk-lsp.nix#L225

PR #3858, opened 2025-10-25
plugins/showkeys: init

HeitorAugustoLN:
Depends on NixOS/nixpkgs#455623

### DIFF
--- a/plugins/by-name/showkeys/default.nix
+++ b/plugins/by-name/showkeys/default.nix
@@ -1,0 +1,13 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "showkeys";
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  settingsExample = {
+    timeout = 5;
+    maxkeys = 5;
+    position = "bottom-center";
+    keyformat."<CR>" = "Enter";
+  };
+}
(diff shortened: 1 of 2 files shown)


plugins/by-name/showkeys/default.nix
```diff
@@ -0,0 +1,89 @@
...
+          "<Right>" = "󰁔";
+          "<PageUp>" = "Page 󰁝";
+          "<PageDown>" = "Page 󰁅";
+          "<M>" = "Alt";
+          "<C>" = "Ctrl";
+        }
+        ''
+          A dictionary to map key names to a custom string.
+        '';
+  };
```
GaetanLepage:
```suggestion
```

We now try to avoid declaring `settingsOptions` at all cost.
If you don't mind, I would encourage you to drop these option definitions.

tests/test-sources/plugins/by-name/showkeys/default.nix
```diff
@@ -0,0 +1,16 @@
+{
+  default = {
+    plugins.showkeys.enable = true;
+  };
+
```
GaetanLepage, line 6: Can you add a `defaults` test case with the plugin default values?

tests/test-sources/plugins/by-name/showkeys/default.nix
```diff
@@ -0,0 +1,59 @@
+{ lib, ... }:
+{
+  default = {
```
GaetanLepage:
```suggestion
  empty = {
```
Nit. Otherwise LGTM!

PR #3859, opened 2025-10-25
plugins/venv-selector: init

### LINKED ISSUE #3801: [PLUGIN REQUEST] venv-picker
### Plugin

venv-picker

### Homepage

https://github.com/linux-cultist/venv-selector.nvim

### Nixpkgs URL

_No response_

### Extra Information

A simple neovim plugin to let you choose what virtual environment to activate in neovim. 

Requires uv and FD in path.

GaetanLepage:
Add [venv-selector.nvim](https://github.com/linux-cultist/venv-selector.nvim), a plugin allowing selection of python virtual environment from within Neovim.

Closes #3801

### DIFF
--- a/plugins/by-name/venv-selector/default.nix
+++ b/plugins/by-name/venv-selector/default.nix
@@ -1,0 +1,40 @@
+{
+  lib,
+  config,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "venv-selector";
+  package = "venv-selector-nvim";
+
+  maintainers = [ lib.maintainers.GaetanLepage ];
+
+  dependencies = [
+    "fd"
+  ];
+
+  settingsExample = {
+    name = [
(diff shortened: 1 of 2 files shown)


plugins/by-name/venv-selector/default.nix
```diff
@@ -0,0 +1,40 @@
...
+          message = ''
+            You have to enable `plugins.${pickerName}` as `settings.picker` is set to `"${pickerName}"`.
+          '';
+        })
+        [
+          "telescope"
+          "fzf-lua"
+        ]
+    );
+  };
```
khaneliman, line 39: Kinda like this pattern, feel like we could do something for more plugins like that.

GaetanLepage: Yes, maybe introducing some helper for instance. I have no clear idea yet though.

PR #3866, opened 2025-10-27
docs/maintaining: add deprecation policy

saygo-png:
closes https://github.com/nix-community/nixvim/issues/3181

largely based on <https://github.com/nix-community/nixvim/issues/3181#issuecomment-2819516048>

I did switch to "should remove after n+2" from "can" as I think that reduces the gray zone when contributing.

I'm bad at writing so corrections very much welcome, just wanted to finally make this official.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -4,7 +4,24 @@ This file is intended as a reference for Nixvim's core maintainers, although it 
 
 ## Deprecation
 
-The deprecation policy discussion is happening in [#3181](https://github.com/nix-community/nixvim/issues/3181).
+We aim to guarantee a seamless experience for stable channel users by providing rename aliases, removal notices, and other warnings and assertions.
+
+Deprecations should include a comment in the format of:
+```nix
+# TODO: Added YYYY-MM-DD, remove after <n+2>
+```
+
+where:
+
+- `n` is the current release; 25.05 at the time of writing
+- `n+1` is the next release; 25.11 at the time of writing
+- `n+2` is the release following `n+1`; 26.05 at the time of writing
+
+If a warning is created on release `n`, then:
+
+- We **should not** remove it before `n+1` is released unless there is a bug or it is plain wrong
+- We **may** remove it after `n+1` is released, but only if it causes undue friction
+- We **should** remove it once `n+2` is released, as routine cleanup
 
 ## Releasing
 


MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
 
 ## Deprecation
 
-The deprecation policy discussion is happening in [#3181](https://github.com/nix-community/nixvim/issues/3181).
+We aim to guarantee a seamless experience for stable channel users.
+
+Deprecations should include a comment in the format of:
+```nix
+  # TODO: Added YYYY-MM-DD, Remove after <n+2>
```
MattSturgeon:
```suggestion
# TODO: Added YYYY-MM-DD, remove after <n+2>
```

Q: do we still want to use `TODO`?

Perhaps `NOTE` is more suitable, or simply omit the flag?

MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
 
 ## Deprecation
 
-The deprecation policy discussion is happening in [#3181](https://github.com/nix-community/nixvim/issues/3181).
+We aim to guarantee a seamless experience for stable channel users.
```
MattSturgeon:
```suggestion
We aim to guarantee a seamless experience for stable channel users by providing rename aliases, removal notices, and other warnings and assertions.
```

MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
...
+
+Deprecations should include a comment in the format of:
+```nix
+  # TODO: Added YYYY-MM-DD, Remove after <n+2>
+```
+
+where:
+
+- n is the current release (i.e. 2025.05 as of writing)
+- n+1 is the next release (i.e. 2025.11 as of writing)
```
MattSturgeon:
```suggestion
- `n` is the current release; 25.05 at the time of writing
- `n+1` is the next release; 25.11 at the time of writing
```

I'm not keen on having the actual current release written down; it's yet another thing to update when we cut a new release. Although we already have to grep the codebase for `s/25.05/25.11/` so it doesn't make things much worse.

We should use the actual naming convention though, `YY.MM` instead of `YYYY.MM`.

Would it make more sense to mention `n+2` (26.05), since that is the release we actually care about? Mentioning `n+1` here could get confusing 😵‍💫

MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
...
+
+If a warning is created on release n, then:
+
+- We **should not** (unless there is a bug, or it is plain wrong) remove it before n+1 is released
```
MattSturgeon:
```suggestion
- We **should not** remove it before n+1 is released, (unless there is a bug, or it is plain wrong)
```

MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
...
+- We **might** remove it if it annoys us starting from when n+1 is released
```
MattSturgeon:
```suggestion
- We **may** remove it after n+1 is released, but only if it cases undue friction
```

MAINTAINING.md
```diff
@@ -4,7 +4,23 @@ This file is intended as a reference for Nixvim's core maintainers, although it
...
+- We **should** remove it once n+2 is released
```
MattSturgeon:
```suggestion
- We **should** remove it once n+2 is released, as routine cleanup
```

MAINTAINING.md
saygo-png: I think `TODO` is good. This is not a note, it is something we should  act on later which fits the bill for `TODO`.

MAINTAINING.md
saygo-png: > it's yet another thing to update when we cut a new release

I don't think this needs to be updated. It's okay for this to be out of date as we clarify that this was at the time of writing. It just serves as a practical example for what we mean by these values. I think not mentioning the values might be a bit too abstract and difficult to understand.

MAINTAINING.md
```diff
@@ -4,7 +4,24 @@ This file is intended as a reference for Nixvim's core maintainers, although it
...
+where:
+
+- `n` is the current release; 25.05 at the time of writing
+- `n+1` is the next release; 25.11 at the time of writing
+- `n+2` is the release following `n+1`; 26.05 at the time of writing
+
+If a warning is created on release `n`, then:
+
+- We **should not** remove it before `n+1` is released unless there is a bug or it is plain wrong
+- We **may** remove it after `n+1` is released, but only if it cases undue friction
```
MattSturgeon: Sorry, I had a typo in my suggestion:

```suggestion
- We **may** remove it after `n+1` is released, but only if it causes undue friction
```

GaetanLepage:
Thanks @saygo-png!

PR #3872, opened 2025-10-29
colorschemes/vscode: drop settingsOptions, add settingsExample

loicreynier:


plugins/colorschemes/vscode.nix
GaetanLepage, line 23:
```suggestion
```
Let's drop all options.

GaetanLepage:
```suggestion
```

GaetanLepage:
```suggestion
      _vscode.setup(${lib.nixvim.toLuaObject cfg.settings})
```

GaetanLepage (commented):
Thanks for the contribution!

We now tend to simply get rid of all option definitions. This reduces the maintenance burder.
All `settings` options are `freeform` anyway and users can provide arbitrary `attrs` to them.

Could you please drop `settingsOptions` entirely? If you are willing to, can you also add `settingsExample` (see other plugins to see what it looks like)?

loicreynier:
Thanks for the update @GaetanLepage! I dropped the `settingsOptions` and added `settingsExample`.

PR #3876, opened 2025-10-31
plugins/ethersync: rename to teamtype

ddogfoodd:
ethersync was renamed to teamtype

Upstream Notice: https://github.com/teamtype/teamtype/releases/tag/v0.9.0
Nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/456770

Nixpkgs PR is merged but `nix flake check --all-systems` still returns:
```
error: vimPlugins.teamtype cannot be found in pkgs
```

### DIFF
--- a/plugins/by-name/dressing/default.nix
+++ b/plugins/by-name/dressing/default.nix
@@ -12,7 +12,13 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   settingsOptions =
     let
-      intOrRatio = with types; either ints.unsigned (numbers.between 0.0 1.0);
+      intOrRatio =
+        with types;
+        oneOf [
+          ints.unsigned
+          (numbers.between 0.0 1.0)
+          rawLua
+        ];
     in
     {
       input = {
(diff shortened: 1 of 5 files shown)


MattSturgeon:
> Nixpkgs PR is merged but `nix flake check --all-systems` still returns:

That PR was merged 4 days ago, so has likely only just landed in nixpkgs-unstable.

As standard practice, the PR added a transition alias. We don't _need_ to rush into renaming our usage, since users may have slightly outdated nixpkgs revisions (even once we bump ours). That said, due to laziness it should only affect users actually using this plugin either way, so updating promptly is also ok.

If we rename any options/plugins we should also do so in with deprecation aliases (e.g. using `plugins/deprecation.nix`). EDIT: I see you're already doing that, great!

As an aside: it looks like our update workflow has been broken for a couple days: https://github.com/nix-community/nixvim/actions/runs/19034563856/job/54355649974

GaetanLepage:
Thanks for the PR, we ended up handling this in #3912.

PR #3879, opened 2025-11-01
plugins: cleanup, remove most settings declarations (2)

saygo-png:


GaetanLepage (approved):
Thanks.
Just to be sure, @MattSturgeon are you fine with removing settings option declarations (in general)?

MattSturgeon:
> Just to be sure, @MattSturgeon are you fine with removing settings option declarations (in general)?

Yes.

However we (of course) need to be careful when removing "fancier" options that either didn't match the underlying lua exactly or implemented some behaviour.

The most obvious example would be the old-style "lua string" options (like `mkLuaFn`), that internally coerced string definitions into raw-lua definitions. Dropping any option that did any kind of coercion is a breaking change.

EDIT: at first glance, it looks like this PR has taken care to not remove such options yet.

saygo-png:
> at first glance, it looks like this PR has taken care to not remove such options yet.

I'm trying my best to only delete the non lua string ones but please do check that i'm not accidentally deleting some of them. With how many options i'm removing it's likely that I will miss something eventually.

PR #3908, opened 2025-11-02
Colorschemes by-name and new-plugin flag

HeitorAugustoLN:
- Adds a `--colorscheme` flag to new-plugin script, so contributors can easily create colorschemes.
- Moved colorschemes from `plugins/colorschemes` to `colorschemes/`

### DIFF
--- a/colorschemes/vscode/default.nix
+++ b/colorschemes/vscode/default.nix
@@ -1,0 +1,32 @@
+{
+  lib,
+  ...
+}:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "vscode";
+  isColorscheme = true;
+  package = "vscode-nvim";
+  colorscheme = null; # Color scheme is set by `require.("vscode").load()`
+  callSetup = false;
+
+  maintainers = [ lib.maintainers.loicreynier ];
+
+  settingsExample = {
+    transparent = true;
+    italic_comments = true;
+    italic_inlayhints = true;
+    underline_links = true;
+    terminal_colors = true;
+    color_overrides = {
+      vscLineNumber = "#FFFFFF";
+    };
+  };
+
+  extraConfig = cfg: {
+    colorschemes.vscode.luaConfig.content = ''
+      local _vscode = require("vscode")
+      _vscode.setup(${lib.nixvim.toLuaObject cfg.settings})
+      _vscode.load()
+    '';
+  };
+}
--- a/tests/test-sources/colorschemes/vscode/default.nix
+++ b/tests/test-sources/colorschemes/vscode/default.nix
@@ -1,0 +1,37 @@
+{
+  empty = {
+    colorschemes.vscode.enable = true;
+  };
+
+  defaults = {
+    colorschemes.vscode = {
+      enable = true;
+      settings = {
+        transparent = false;
+        italic_comments = false;
+        italic_inlayhints = false;
+        underline_links = false;
+        color_overrides = false;
+        group_overrides = { };
+        disable_nvimtree_bg = true;
+        terminal_colors = true;
+      };
+    };
+  };
+
+  example = {
+    colorschemes.vscode = {
+      enable = true;
+      settings = {
+        transparent = true;
+        italic_comments = true;
+        italic_inhayhints = true;
+        underline_links = true;
+        terminal_colors = true;
+        color_overrides = {
+          vscLineNumber = "#FFFFFF";
+        };
+      };
+    };
+  };
+}


modules/plugins.nix
```diff
@@ -3,14 +3,14 @@ let
...
 in
 {
-  imports = [
-    ../plugins
-  ]
-  ++ foldlAttrs (
-    prev: name: type:
-    prev ++ optional (type == "directory") (by-name + "/${name}")
-  ) [ ] (readDir by-name);
+  imports = [ ../plugins ] ++ mkByName ../plugins/by-name ++ mkByName ../colorschemes/by-name;
```
GaetanLepage:
```suggestion
  imports = [ ../plugins ] ++ (map mkByName [ ../plugins/by-name ../colorschemes/by-name ]);
```

colorschemes/by-name/modus/default.nix
```diff
@@ -103,11 +22,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
       functions = { };
       variables = { };
     };
-    on_colors = "function(colors) end";
-    on_highlights = "function(highlights, colors) end";
-  };
-
-  extraConfig = {
-    opts.termguicolors = lib.mkDefault true;
```
GaetanLepage, line 111: This is technically a non-trivial silent regression.

colorschemes/by-name/monokai-pro/default.nix
```diff
@@ -1,126 +1,23 @@
...
-    warnings = lib.nixvim.mkWarnings "colorschemes.monokai-pro" {
-      when = (cfg.settings.devicons == true) && (!config.plugins.web-devicons.enable);
-      message = ''
-        You have enabled `settings.devicons` but `plugins.web-devicons.enable` is `false`.
-        Consider enabling the plugin for proper devicons support.
-      '';
-    };
+    colorscheme = lib.mkIf (
+      cfg.settings ? filter && builtins.isString cfg.settings.filter
+    ) "monokai-pro-${cfg.settings.filter}";
```
GaetanLepage: I would not change the plugin's behavior in a `remove settingsOptions`-named commit.

colorschemes/by-name/palette/default.nix
```diff
@@ -109,10 +22,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
         (
           name: defaultPaletteNames:
           let
-            customPalettesNames = lib.attrNames cfg.settings.custom_palettes.${name};
+            customPalettesNames = lib.attrNames ((cfg.settings.custom_palettes or { }).${name} or { });
             allowedPaletteNames = customPalettesNames ++ defaultPaletteNames;
 
-            palette = cfg.settings.palettes.${name};
+            palette = (cfg.settings.palettes or { }).${name} or null;
```
GaetanLepage: Why not simply `cfg.settings.palettes.${name} or null`?
The `or` operator works even if a shallower attribute is missing.

HeitorAugustoLN:
I did this on a bit of a rush, so I accidentally deleted some things I shouldn't, I will do the removal of settingsOptions after in another PR, will just rebase things out

flake/dev/new-plugin.py
```diff
@@ -24,7 +24,28 @@
...
+  {maintainer_todo}maintainers = [ lib.maintainers.{maintainer} ];
+
+  # TODO provide an example for the `settings` option (or remove entirely if there is no useful example)
+  # NOTE you can use `lib.literalExpression` or `lib.literalMD` if needed
+  settingsExample = {{
+    foo = 42;
+    bar.__raw = "function() print('hello') end";
+  }};
+}}
+"""
```
MattSturgeon, line 46: So we now have plugin templates fragmented over _three_ places? Not a fan... But it's not this PR's fault and finding a better approach shouldn't be part of this PR anyway. I'm just complaining for the record, not requesting changes here.

modules/plugins.nix
```diff
@@ -2,15 +2,21 @@
...
     ../plugins
   ]
-  ++ foldlAttrs (
-    prev: name: type:
-    prev ++ optional (type == "directory") (by-name + "/${name}")
-  ) [ ] (readDir by-name);
+  ++ concatMap mkByName [
+    ../colorschemes/by-name
+    ../plugins/by-name
+  ];
```
MattSturgeon, line 21: Neat.

plugins/default.nix
```diff
@@ -2,39 +2,6 @@
...
-    ./colorschemes/nord.nix
-    ./colorschemes/one.nix
-    ./colorschemes/onedark.nix
-    ./colorschemes/oxocarbon.nix
-    ./colorschemes/palette.nix
-    ./colorschemes/poimandres.nix
-    ./colorschemes/rose-pine.nix
-    ./colorschemes/solarized-osaka.nix
-    ./colorschemes/tokyonight.nix
-    ./colorschemes/vscode.nix
```
MattSturgeon, line 36: If _all_ colourschemes are now by-name, it begs the question: do we need a `colorschemes/by-name` subdir or could we just have `colorschemes` itself be by-name?

I guess we also have the third option of `colorschemes-by-name`, although that's not my first choice.

colorschemes/ayu/default.nix
MattSturgeon, line 1: Seeing all these as "renamed without changes" gives me much more confidence reviewing. Thanks for postponing the other changes for future PRs!

saygo-png:
Sorry for the conflict!

MattSturgeon:
While GitHub will show a conflict, git should be able to resolve the conflicts automatically when rebasing rename-only diffs.

MattSturgeon:
Thanks for rebasing. I'd appreciate some discussion on https://github.com/nix-community/nixvim/pull/3908#discussion_r2488018910 before merging, but otherwise this SGTM.

plugins/default.nix
HeitorAugustoLN: I am fine with having just `colorschemes` instead of `colorschemes/by-name`. Thoughts @GaetanLepage?

GaetanLepage: > I am fine with having just `colorschemes` instead of `colorschemes/by-name`. Thoughts @GaetanLepage?

I agree, let's simplify this.

colorschemes/ayu/default.nix
MattSturgeon, line 1: I don't think we need to move `colorschemes` outside of `plugins`. I think `plugins/colorschemes` is fine.

The only renames I expected to see are `foo.nix` -> `foo/default.nix`.

HeitorAugustoLN: I think keeping colorschemes separate makes the structure clearer and helps distinguish them from general plugins.

saygo-png: The distinction is still there with `plugins/colorschemes`. I don't like moving it outside of `plugins` as a colorscheme is still just a plugin, it's just a certain kind of a plugin.

HeitorAugustoLN: My thinking is that the directory structure should reflect the option namespaces. The `plugins/` directory is for modules targeting the `plugins.*` namespace, and `colorschemes/` is for modules targeting the `colorschemes.*` namespace.

saygo-png: That's a good point. I agree now, I'd rather have it be outside of `plugins/` to reflect the options.

GaetanLepage: I am fine with both layouts. Separating `colorschemes` and `plugins` works fine.

PR #3910, opened 2025-11-03
plugins: cleanup, remove most settings declarations (3)

saygo-png:
cherrypicked from https://github.com/nix-community/nixvim/pull/3865


plugins/by-name/dap-go/default.nix
```diff
@@ -2,43 +2,13 @@
...
-  dapHelpers = import ../dap/dapHelpers.nix { inherit lib; };
-in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "dap-go";
   package = "nvim-dap-go";
   description = "An extension for nvim-dap providing configurations for launching go debugger.";
 
   maintainers = [ lib.maintainers.khaneliman ];
 
-  settingsOptions = {
```
GaetanLepage, line 18: Can you maybe add `settingsExample` in the meantime?

saygo-png: Sorry but I haven't used dap at all so I don't feel comfortable writing an example for an adapter. It's a complex project.

If you want to keep these settings for documentation then i could just fix the declaration type instead to allow for raw lua

MattSturgeon: cc @khaneliman: maybe you have an example?

Adding an example is out-of-scope for this PR, though IMO. It can be added and/or improved at any point in dedicated PRs.

plugins/by-name/ledger/default.nix
```diff
@@ -23,163 +20,6 @@ mkVimPlugin {
...
-      This is the number of lines of a horizontal quickfix window, or the number of columns of a
-      vertical quickfix window.
-    '';
-
-    qf_hide_file = helpers.defaultNullOpts.mkFlagInt 1 ''
-      Flag to show or hide filenames in the quickfix window:
-
-      Filenames in the quickfix window are hidden by default. Set this to 1 is
-      you want filenames to be visible.
-    '';
```
MattSturgeon, line 168: FYI `mkFlagInt` uses the `intFlag` type: https://github.com/nix-community/nixvim/blob/660fba984fe25ded6fa2e32016c05aebf4290273/lib/types.nix#L60-L67

This type does some warnings & coercion to support deprecated bool-type definitions.

We've probably had that long enough (~14mo) that we can replace coercion with a throw though. I think it _should_ be ok to drop these option defs at this point.

khaneliman:
> cc @khaneliman: maybe you have an example?
> 
> Adding an example is out-of-scope for this PR, though IMO. It can be added and/or improved at any point in dedicated PRs.

Yeah my config should have a lot of dap examples. But, I agree it can be done separately. Unless this is specific to `Go`, I haven't really programmed with that before.

PR #3911, opened 2025-11-03
plugins: fix types to allow for `rawLua` values

saygo-png:
This touches the types blocking https://github.com/nix-community/nixvim/pull/3865

I'm also including a fix for the `notebook-navigator` warning condition to account for `rawLua` `nil`

### DIFF
--- a/plugins/by-name/lazygit/default.nix
+++ b/plugins/by-name/lazygit/default.nix
@@ -63,8 +63,14 @@ lib.nixvim.plugins.mkVimPlugin {
       Config file path is evaluated if this value is `1`.
     '';
 
-    config_file_path = defaultNullOpts.mkNullable (with types; either str (listOf str)) [
-    ] "Custom config file path or list of custom config file paths.";
+    config_file_path = defaultNullOpts.mkNullable (
+      with types;
+      oneOf [
+        str
+        (listOf str)
+        rawLua
+      ]
+    ) [ ] "Custom config file path or list of custom config file paths.";
   };
 
   settingsExample = {


plugins/by-name/lazygit/default.nix
```diff
@@ -63,8 +63,19 @@ lib.nixvim.plugins.mkVimPlugin {
...
+        (
+          with types;
+          oneOf [
+            str
+            (listOf str)
+            rawLua
+          ]
+        )
+        [
+        ]
```
MattSturgeon: Does nixfmt allow: 

```suggestion
        [ ]
```
?

plugins/by-name/notebook-navigator/default.nix
```diff
@@ -105,7 +105,14 @@ lib.nixvim.plugins.mkNeovimPlugin {
   extraConfig = cfg: {
     warnings = lib.nixvim.mkWarnings "plugins.notebook-navigator" [
       {
-        when = (cfg.settings.activate_hydra_keys != null) && (!config.plugins.hydra.enable);
+        when =
+          (
+            !(builtins.elem cfg.settings.activate_hydra_keys [
```
MattSturgeon: Should this allow for a possible future where `activate_hydra_keys` is freeform?


```suggestion
            !(builtins.elem (cfg.settings.activate_hydra_keys or null) [
```
or
```suggestion
            !(cfg ? settings.activate_hydra_keys && builtins.elem cfg.settings.activate_hydra_keys [
```

MattSturgeon (approved):

plugins/by-name/lazygit/default.nix
saygo-png: Yes, but it's ugly. Changed it.

plugins/by-name/notebook-navigator/default.nix
saygo-png: I thought about it too but went against it since it currently isn't. But since you mentioned it too I added it.

MattSturgeon: I don't have an opinion, just raising the question.

You're right that if we have an explicit option, the presence check or fallback value is redundant.

PR #3914, opened 2025-11-04
lspkind: cmp.enable default to plugins.cmp.enable

repparw:
Fixes hitting assert when `lspkind.enable = true` and no config for cmp set

### DIFF
--- a/plugins/by-name/lspkind/default.nix
+++ b/plugins/by-name/lspkind/default.nix
@@ -23,7 +23,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
     cmp = {
       enable = lib.mkOption {
         type = types.bool;
-        default = true;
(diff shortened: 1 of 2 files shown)


plugins/by-name/lspkind/default.nix
```diff
@@ -23,7 +23,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
     cmp = {
       enable = lib.mkOption {
         type = types.bool;
-        default = true;
+        default = config.plugins.cmp.enable;
```
saygo-png, line 26:
```suggestion
        default = config.plugins.cmp.enable;
        defaultText = lib.literalExpression "config.plugins.cmp.enable";
```
We need to add this so that it renders nicely in the docs. You can preview the docs using `serve-docs` in the Nixvim shell.

PR #3916, opened 2025-11-06
dev/list-plugins: drop

GaetanLepage:
Now that #2638 is complete, we can remove the `state` component of the `list-plugins.py` script.


MattSturgeon:
Should we consider dropping the script entirely? Its main purpose was tracking which plugins were migrated.

Its secondary purpose of listing available plugins can be achieved a number of other ways:

- you can eval `attrNames options.plugins`
- you can read the docs

GaetanLepage:
Sure. I can do that. I will delete it in a second commit though, so that we can revert if we later want it back.

Issue #2746, opened 2024-12-26
[BUG] <BUG> Friendly Snippets do not work with blink-cmp when performance.combinePlugins is enabled


morgan-perry:
| Field        | Description          |
|--------------|----------------------|
| Plugin       | `friendly-snippets`  |
| Nixpkgs      | `unstable`           |
| Home Manager | `N/A`                |

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Description
When friendly snippets blink-cmp and performance.combinePlugins is enabled blink-cmp will not use friendly-snippets snippets, adding friendly-snippets to performance.combinePlugins.standalonePlugins will allow blink to use friendly-snippets. I don't believe I had this problem with cmp however I do not have a working config with it to test. I have also tested this with performance.byteCompileLua.enable and can confirm it works.
For further reference my standalone nixvim config can be found [here](https://github.com/morgan-perry/nixvim)

## Minimal, Reproducible Example (MRE)

```nix
  performance.combinePlugins = {
    enable = true;
    standalonePlugins = [
      "friendly-snippets" # needed for blink to access friendly-snippets
    ];
  };
  plugins.blink-cmp = {
    enable = true;
  };
  plugins.friendly-snippets.enable = true;
```

Alexis-Fouquet:
I had the same issue for Friendly Snippets with `plugins.cmp` and `plugins.luasnip` (and another snippet plugin, `nvim-snippets`).
This fix also worked for me. Thank you!
The [commit](https://github.com/Alexis-Fouquet/home/commit/63a53286d985c20f22e477ff049cce612c6d2388) where I fixed this in my config, if needed.

PR #3920, opened 2025-11-07
plugins/friendly-snippets: add warning when using `performance.combinePlugins`

saygo-png:
closes https://github.com/nix-community/nixvim/issues/2746

### DIFF
--- a/plugins/by-name/friendly-snippets/default.nix
+++ b/plugins/by-name/friendly-snippets/default.nix
@@ -1,5 +1,7 @@
 {
   lib,
+  config,
+  options,
   ...
 }:
 lib.nixvim.plugins.mkVimPlugin {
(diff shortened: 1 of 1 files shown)


plugins/by-name/friendly-snippets/default.nix
```diff
@@ -9,6 +11,32 @@ lib.nixvim.plugins.mkVimPlugin {
   maintainers = [ lib.maintainers.GaetanLepage ];
 
   extraConfig = {
+    warnings = lib.nixvim.mkWarnings "plugins.friendly-snippets" [
```
GaetanLepage: Nit, `mkWarnings` can be fed with a single warning instead of a list. Not a big deal.

```nix
warnings = lib.nixvim.mkWarnings "plugins.friendly-snippets" {
  when = true;
  message = "";
};

File: plugins/by-name/friendly-snippets/default.nix [hunk 2]
```diff
@@ -9,6 +11,32 @@ lib.nixvim.plugins.mkVimPlugin {
...
+      (
+        let
+          snippetConsumers = map (lib.splitString ".") [
+            "plugins.luasnip.enable"
+            "plugins.cmp.enable"
+            "plugins.blink-cmp.enable"
+            "plugins.nvim-snippets.enable"
+          ];
+          enabledConsumers = builtins.filter (path: lib.getAttrFromPath path config) snippetConsumers;
+          enabledConsumersPretty = lib.join ", " (
```
MattSturgeon: `join` is an alias for `concatStringsSep`

https://noogle.dev/f/lib/join

MattSturgeon: In this case, we can use `concatMapStringsSep` to inline the `map`

Issue #3313, opened 2025-05-11
[BUG] <BUG>


Mikilio:
| Field        | Description |
|--------------|-------------|
| Plugin       | `<PLUGIN>`  | 
| Nixpkgs      | `<VERSION>` | 
| Home Manager | `<VERSION>` | 

- [ ] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

## Minimal, Reproducible Example (MRE)

```nix
programs.nixvim = {
}
```

MattSturgeon:
markdown-preview is implemented using `mkVimPlugin`, because its settings do not use a lua `setup` function.

Currently, only `mkNeovimPlugin` supports lazy loading. Initially, the assumption was that lz-n only supports calling `setup` lazily, however it also supports lazily enabling plugins from the "optional" part of the runtimepath.

I've updated the issue title to reflect the underlying issue.

PR #3927, opened 2025-11-10
lib/plugins/mkVimPlugin: Support lazyloading

saygo-png:
Closes: https://github.com/nix-community/nixvim/issues/3313

This implementation uses a loop to set the globals. Since this is per plugin many of these loops might be created. I don't know if this will affect performance and I don't know how to benchmark this.

I tested this with `vimVisualMulti` and `rainbow`, both plugins can be successfully lazyloaded.

With this change all 400+ plugin modules we have will support lazyloading!

### DIFF
--- a/lib/plugins/mk-vim-plugin.nix
+++ b/lib/plugins/mk-vim-plugin.nix
@@ -59,6 +59,7 @@ let
       options = lib.setAttrByPath loc (
         {
           enable = lib.mkEnableOption name;
+          lazyLoad = lib.nixvim.mkLazyLoadOption name;
           autoLoad = lib.nixvim.mkAutoLoadOption cfg name;
         }
         // settingsOption
@@ -66,19 +67,67 @@ let
       );
 
       config = lib.mkIf cfg.enable (
+        let
+          globalsConfig = lib.nixvim.applyPrefixToAttrs globalPrefix (cfg.settings or { });
+        in
         lib.mkMerge [
-          {
-            inherit extraPackages extraPlugins;
-            globals = lib.nixvim.applyPrefixToAttrs globalPrefix (cfg.settings or { });
-          }
+          { inherit extraPackages extraPlugins; }
+
+          (lib.mkIf cfg.lazyLoad.enable {
+            assertions = [
+              {
+                assertion = (isColorscheme && colorscheme != null) || cfg.lazyLoad.settings != { };
+                message = "You have enabled lazy loading for ${name} but have not provided any configuration.";
+              }
+            ];
+
+            plugins.lz-n = {
+              plugins = [
+                (
+                  {
+                    # The packpath name is always the derivation name
+                    __unkeyed-1 = lib.getName cfg.package;
+                    # Use provided before, otherwise fallback to normal function wrapped globals config
+                    before =
+                      let
+                        luaGlobals = lib.nixvim.toLuaObject globalsConfig;
+                        before = cfg.lazyLoad.settings.before or null;
+                        default = lib.mkIf (luaGlobals != "{ }") ''
+                          function()
+                            local globals = ${luaGlobals}
+
+                            for k,v in pairs(globals) do
+                              vim.g[k] = v
+                            end
+                          end
+                        '';
+                      in
+                      if (lib.isString before || lib.types.rawLua.check before) then before else default;
+                    colorscheme = lib.mkIf isColorscheme (cfg.lazyLoad.settings.colorscheme or colorscheme);
+                  }
+                  // lib.removeAttrs cfg.lazyLoad.settings [
+                    "before"
+                    "colorscheme"
+                  ]
+                )
+              ];
+            };
+          })
+
+          (lib.mkIf (!cfg.lazyLoad.enable) {
+            globals = globalsConfig;
+          })
+
           (lib.optionalAttrs (isColorscheme && colorscheme != null) {
             colorscheme = lib.mkDefault colorscheme;
           })
+
           (lib.optionalAttrs (args ? extraConfig) (
             lib.nixvim.plugins.utils.applyExtraConfig {
               inherit extraConfig cfg opts;
             }
           ))
+
           (lib.nixvim.plugins.utils.enableDependencies dependencies)
         ]
       );


GaetanLepage (commented):
Does delaying setting global variables actually corresponds to "lazy loading"?
No (plugin) code is really executed when one runs `vim.g["foo_bar"] = true`, correct?

GaetanLepage:
This was a duplicate of https://github.com/NixOS/nixpkgs/pull/460183.
Also, it broke the package for `aarch64-linux`.

saygo-png:
> No (plugin) code is really executed when one runs vim.g["foo_bar"] = true, correct?

Yes

> Does delaying setting global variables actually corresponds to "lazy loading"?

As i understand lazyloading is not just loading the plugin lazily, it's also about loading it's config lazily. We do this for Lua plugins, and it's done in lz-n docs for vimscript ones. Look at `vim-startuptime` here https://github.com/lumen-oss/lz.n?tab=readme-ov-file#examples (Though as i'm writing this i realise they reccomend using `before` instead of `after` for vimscript plugins)

saygo-png:
Sorry, missed a big issue. Previous code always generated Lua even if `settings` was empty or set to things that filter out. The code I pushed makes it work properly but it looks horrible. Is there some better way to do this?

saygo-png:
Rebase and now i check if settings is empty using `toLua` which is much nicer and more reliable.

khaneliman (approved):

lib/plugins/mk-vim-plugin.nix
```diff
@@ -59,26 +59,74 @@ let
...
+              plugins = [
+                (
+                  {
+                    # The packpath name is always the derivation name
+                    __unkeyed-1 = lib.getName cfg.package;
+                    # Use provided before, otherwise fallback to normal function wrapped globals config
+                    before =
+                      let
+                        before = cfg.lazyLoad.settings.before or null;
+                        default = lib.mkIf (lib.nixvim.toLuaObject globalsConfig != "{ }") ''
```
MattSturgeon: IIUC eval cache correctly, doing `toLuaObject globalsConfig` twice may or may not eval twice. If we use a `let` binding for repeated values we not only DRY up the code a bit, we also make it easier for nix to re-use the value in both places.

lib/plugins/mk-vim-plugin.nix
```diff
@@ -59,26 +59,74 @@ let
...
+                          function()
+                            local globals = ${lib.nixvim.toLuaObject globalsConfig}
+
+                            for k,v in pairs(globals) do
+                              vim.g[k] = v
+                            end
+                          end
+                        '';
+                      in
+                      if (lib.isString before || lib.types.rawLua.check before) then before else default;
```
MattSturgeon, line 105: In nix, we don't need parentheses for `if` conditions

```suggestion
                      if lib.isString before || lib.types.rawLua.check before then before else default;
```

MattSturgeon: What's the actual intent here? Are you checking "is `before` a valid type" or are you checking "is `before` defined (non-null)"?

It may be better to use `!= null` if that matches the actual intent.

lib/plugins/mk-vim-plugin.nix
```diff
@@ -59,26 +59,74 @@ let
...
+                    colorscheme = lib.mkIf isColorscheme (cfg.lazyLoad.settings.colorscheme or colorscheme);
+                  }
+                  // lib.removeAttrs cfg.lazyLoad.settings [
+                    "before"
+                    "colorscheme"
+                  ]
```
MattSturgeon, line 111: We can use update order to avoid needing a `removeAttrs`.

```nix
{
  __unkeyed-1 = lib.getName cfg.package;
}
// cfg.lazyLoad.settings
// {
  before = ...
  colorscheme = ...
}
```

lib/plugins/mk-vim-plugin.nix
MattSturgeon: It feels like the correct approach here is to refactor things so that we can a non-null default for `lazyLoad.settings.before`, since that is _effectively_ what this is implementing, only without using the module system.

Having this be a proper option default would also make it self-documenting and user visible.

lib/plugins/mk-vim-plugin.nix
MattSturgeon: This also simplifies the `before` and `colorscheme` _condition_:

```nix
{
  __unkeyed-1 = lib.getName cfg.package;
}
// cfg.lazyLoad.settings
// optionalAttrs (cfg.lazyLoad.settings.before == null) {
  before = ...
}
// optionalAttrs isColorscheme {
  colorscheme = ...
}
```

lib/plugins/mk-vim-plugin.nix
saygo-png: The code here mirrors the one from mkNeovimPlugin. I've altered it to work with vimscript plugins, but otherwise tried keeping it the same to prevent introducing subtle differences (and bugs) between them. It used to be a simple null check, but was changed to this string/rawlua check in this commit https://github.com/nix-community/nixvim/commit/0997b371c784e9f623c8666ec006e8970ed76166.  As i understand the intent here was to add some type safety as suggested by you here https://github.com/nix-community/nixvim/pull/2602#discussion_r1878250990

I'll refactor to use the module system so the check won't have to be implemented, so the above is just for information. But since I'd like to keep these implementations close so if we do change how its done here i'll make a follow up pr that does the same changes to mkNeovimPlugin unless you disagree with that.

MattSturgeon: Thanks for reminding me of the history.

There's no "one right approach" here:
- Based on a freeform perspective, the attr can be anything so a stricter type check makes sense.
- If there's an actual option declared, though, it probably makes sense to make its effective default its actual default.

I agree, however we handle it, it should be consistent.

I'm fine for this to be merged as is, so we can consider refractors to both implementations in a dedicated PR.

saygo-png: > I'm fine for this to be merged as is, so we can consider refractors to both implementations in a dedicated PR.

Okay, i'll mark the remarks involving both as resolved and make a pr later that refactors both functions according to them

PR #3930, opened 2025-11-10
plugins/fileline: init

nim65s:
Hi,

This add support for https://github.com/lewis6991/fileline.nvim to allow opening files directly on the right line / column.

I find this useful for example when `ruff` says :
```
RUF059 Unpacked variable `c` is never used
   --> happypose/toolbox/datasets/augmentations.py:432:15
    |
430 |         assert obs.segmentation is not None
431 |         rgb = obs.rgb.copy()
432 |         h, w, c = rgb.shape
    |               ^
```

then I can start nvim with a copy-paste of `happypose/toolbox/datasets/augmentations.py:432:15`

(https://github.com/astral-sh/ruff/issues/14341 would be even better, of course, but it is not there yet)

### DIFF
--- a/tests/test-sources/plugins/by-name/fileline/default.nix
+++ b/tests/test-sources/plugins/by-name/fileline/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.fileline.enable = true;
+  };
+}
--- a/plugins/by-name/fileline/default.nix
+++ b/plugins/by-name/fileline/default.nix
@@ -1,0 +1,10 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "fileline";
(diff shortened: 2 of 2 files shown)


tests/test-sources/plugins/by-name/fileline/default.nix
```diff
@@ -0,0 +1,11 @@
+{
+  empty = {
+    plugins.fileline.enable = true;
+  };
+
+  defaults = {
+    plugins.fileline = {
+      enable = true;
+    };
+  };
```
GaetanLepage:
```suggestion
```
You can drop this ;)

GaetanLepage (commented):
A little squash and we should be good to go!

nim65s:
done, thanks !

PR #3932, opened 2025-11-10
plugins/gdscript-extended-lsp: init

HeitorAugustoLN:
Depends on https://github.com/NixOS/nixpkgs/pull/460433

### DIFF
--- a/plugins/by-name/zk/default.nix
+++ b/plugins/by-name/zk/default.nix
@@ -98,13 +98,13 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   };
 
-  extraConfig = cfg: {
+  extraConfig = cfg: opts: {
     warnings = lib.nixvim.mkWarnings "plugins.zk" (
       mapAttrsToList
         (picker: pluginName: {
           when = (cfg.settings.picker == picker) && !config.plugins.${pluginName}.enable;
           message = ''
-            You have set `plugins.zk.settings.picker = "${picker}"` but `plugins.${pluginName}` is not enabled in your config.
+            You have defined `${opts.settings}.picker = "${picker}"` but `plugins.${pluginName}` is not enabled.
           '';
         })
         {


plugins/by-name/gdscript-extended-lsp/default.nix
```diff
@@ -0,0 +1,11 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gdscript-extended-lsp";
+  package = "gdscript-extended-lsp-nvim";
+
+  maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  settingsExample = {
+    picker = "snacks";
```
GaetanLepage, line 9: Don't we want to add a warning if `picker` is set to `snacks` and `plugins.snacks` is not enabled?

HeitorAugustoLN: The plugin itself notifies the user, if the user executes the `require("gdscript-extended-lsp").pick()` and do not have the specificied plugin

traxys: It's better to raise nix warnings, as they can be caught much earlier (and for example they could make a nix flake check fail)

HeitorAugustoLN: Done

plugins/by-name/gdscript-extended-lsp/default.nix
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+  extraConfig = cfg: {
+    warnings = lib.nixvim.mkWarnings "plugins.gdscript-extended-lsp" (
+      lib.mapAttrsToList
+        (picker: pluginName: {
+          when =
+            (cfg.settings ? picker && cfg.settings.picker == picker) && !config.plugins.${pluginName}.enable;
```
MattSturgeon:
```suggestion
            cfg.settings.picker or null == picker && !config.plugins.${pluginName}.enable;
```

plugins/by-name/gdscript-extended-lsp/default.nix
```diff
@@ -0,0 +1,28 @@
...
+          message = ''
+            You have set `plugins.gdscript-extended-lsp.settings.picker = "${picker}"` but `plugins.${pluginName}` is not enabled in your config.
```
MattSturgeon:
```suggestion
            You have defined `${opts.settings}.picker = "${picker}"` but `plugins.${pluginName}` is not enabled.
```

HeitorAugustoLN:
Also improved the warning for other similar patterns I've taken inspiration from

PR #3936, opened 2025-11-13
plugins/abolish: init

Fovir-GitHub:
Add [abolish](https://github.com/tpope/vim-abolish).

### DIFF
--- a/tests/test-sources/plugins/by-name/abolish/default.nix
+++ b/tests/test-sources/plugins/by-name/abolish/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.abolish.enable = true;
+  };
+}
--- a/plugins/by-name/abolish/default.nix
+++ b/plugins/by-name/abolish/default.nix
@@ -1,0 +1,7 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "abolish";
+  package = "vim-abolish";
+
+  maintainers = [ lib.maintainers.Fovir ];
+}


tests/test-sources/plugins/by-name/abolish/default.nix
```diff
@@ -0,0 +1,19 @@
+_: {
```
GaetanLepage:
```suggestion
{
```
No need for this

tests/test-sources/plugins/by-name/abolish/default.nix
```diff
@@ -0,0 +1,19 @@
...
+    plugins.abolish = {
+      enable = true;
+    };
+  };
+
+  defaults = {
+    plugins.abolish = {
+      enable = true;
+    };
+  };
```
GaetanLepage: This is redundant in this case. Let's not increase the CI duration for nothing ;)

FovirDev:
@GaetanLepage Thanks for your suggestions. I have already fixed them.

tests/test-sources/plugins/by-name/abolish/default.nix
```diff
@@ -0,0 +1,7 @@
+{
+  empty = {
+    plugins.abolish = {
+      enable = true;
+    };
```
GaetanLepage:
```suggestion
    plugins.abolish.enable = true;
```
To be 100% consistent with the rest of the code base. Not a big deal.

GaetanLepage (approved):
Thanks!

tests/test-sources/plugins/by-name/abolish/default.nix
FovirDev: I have changed it.

PR #3935, opened 2025-11-11
plugins/gitportal: init

phinze:
👋 Hey there! This is my first contribution to nixvim - please let me know if I missed anything or if there are any improvements I should make!

---

### Summary
Add support for [gitportal.nvim](https://github.com/trevorhauter/gitportal.nvim), a plugin that bridges Neovim and Git hosting platforms.

### What does this plugin do?
GitPortal enables users to:
- Open the current file in their browser with optional line ranges
- Open shareable Git host permalinks directly in Neovim while respecting branch/commit context

### Changes in this PR
- Add plugin definition with full configuration support
- Include all 6 available settings options:
  - `always_include_current_line` - Always include the current line in URLs
  - `always_use_commit_hash_in_url` - Use commit hash instead of branch name
  - `browser_command` - Custom browser command (auto-detection if null)
  - `default_remote` - Default git remote to use (defaults to "origin")
  - `git_provider_map` - Custom mapping of git providers
  - `switch_branch_or_commit_upon_ingestion` - Behavior when ingesting permalinks ("always", "ask_first", or "never")
- Add test cases for both minimal and configured usage
- Add myself (phinze) to the maintainers list

### Example Configuration
```nix
{
  plugins.gitportal = {
    enable = true;
    settings = {
      always_include_current_line = true;
      default_remote = "upstream";
      switch_branch_or_commit_upon_ingestion = "ask_first";
    };
  };
}
```

### Testing
- Syntax validation passed
- Code formatted with \`nix fmt\`
- Plugin follows nixvim conventions (using \`mkNeovimPlugin\`)
- Test files created following the standard pattern

PR #3938, opened 2025-11-14
plugins/gitportal: init

phinze:
Adds [gitportal](https://github.com/trevorhauter/gitportal.nvim).

Recreated from #3935 where my rebasing over the update branch ended up doing more harm than good 😅

### DIFF
--- a/tests/test-sources/plugins/by-name/gitportal/default.nix
+++ b/tests/test-sources/plugins/by-name/gitportal/default.nix
@@ -1,0 +1,33 @@
+{ lib, ... }:
+{
+  empty = {
+    plugins.gitportal.enable = true;
+  };
+
+  defaults = {
+    plugins.gitportal = {
+      enable = true;
+
+      settings = {
+        always_include_current_line = false;
+        always_use_commit_hash_in_url = false;
+        browser_command = lib.nixvim.mkRaw "nil";
+        default_remote = "origin";
+        git_provider_map = lib.nixvim.mkRaw "nil";
+        switch_branch_or_commit_upon_ingestion = "always";
+      };
+    };
+  };
+
+  example = {
+    plugins.gitportal = {
+      enable = true;
+
+      settings = {
+        always_include_current_line = true;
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/by-name/gitportal/default.nix
```diff
@@ -0,0 +1,16 @@
+{
+  empty = {
+    plugins.gitportal.enable = true;
+  };
+
+  example = {
+    plugins.gitportal = {
+      enable = true;
+      settings = {
```
GaetanLepage, line 26:
```suggestion
      enable = true;

      settings = {
```

tests/test-sources/plugins/by-name/gitportal/default.nix
```diff
@@ -0,0 +1,16 @@
+{
+  empty = {
+    plugins.gitportal.enable = true;
+  };
+
```
GaetanLepage, line 6:
```suggestion

  defaults = {
    enable = true;
    
    settings = {
      always_include_current_line = false;
      always_use_commit_hash_in_url = false;
      browser_command = null;
      default_remote = "origin";
      git_provider_map = null;
      switch_branch_or_commit_upon_ingestion = "always";
    };
  };

```

saygo-png: We shouldn't use `null` here as it gets filtered. Use `lib.nixvim.mkRaw "nil"` to get a lua nil value or remove the lines to not set them.

phinze:
Wow you're fast! 😄 

```
error: The option `enable' does not exist. Definition values:
 - In `/nix/store/zl42mcncvvgjkfxgly8sykm4wrdwrl1p-source/tests/test-sources/plugins/by-name/gitportal/default.nix': true
```

^^ I had assumed the enable option was implicit. Any guidance on how to address this?

phinze:
I think I see it - needed to nest differently.

Issue #3951, opened 2025-11-18
[BUG] neo-tree fails to open log file when checking build.tests


0xda157:
### Plugin

neo-tree

### Nixpkgs Release

unstable

### Home Manager Release

None

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

running `nix flake check` on the below flake with neo-tree enabled results in 
```
ERROR: [Neo-tree WARN] Could not open log file: /build/.local/share/nvim/neo-tree.nvim.log /build/.local/share/nvim/neo-tree.nvim.log: No such file or directory
``` 

### Minimal, Reproducible Example (MRE)

```nix
{
  inputs = {
    nixpkgs.follows = "nixvim/nixpkgs";

    nixvim.url = "github:nix-community/nixvim";
  };

  outputs =
    {
      self,
      nixpkgs,
      nixvim,
    }:
    let
      inherit (nixpkgs) lib;
      forEachPkgs = f: lib.genAttrs lib.systems.flakeExposed (system: f nixpkgs.legacyPackages.${system});
    in
    {
      checks = forEachPkgs (pkgs: {
        default =
          (nixvim.lib.evalNixvim {
            inherit (pkgs) system;
            modules = lib.singleton {
              plugins.neo-tree.enable = true;
            };
          }).config.build.test;
      });
    };
}
```

GaetanLepage:
What happens here is that the test that you run starts a Neovim instance **in the nix sandbox**.
In this isolated environment, there is no home directory, so telescope will not be able to write its logfile to `~/.local/share/nvim/neo-tree.nvim.log`.

This is the very reason why we set `settings.log_to_file` to `false` in our test suite:
https://github.com/nix-community/nixvim/blob/7408b72662f2c6930448539638528b23c05ef17e/tests/test-sources/plugins/by-name/neo-tree/default.nix#L8-L11

0xda157:
> What happens here is that the test that you run starts a Neovim instance **in the nix sandbox**. In this isolated environment, there is no home directory, so telescope will not be able to write its logfile to `~/.local/share/nvim/neo-tree.nvim.log`.

[...]

PR #3952, opened 2025-11-19
lib: move `enableExceptInTests` impl to `build.test` option

MattSturgeon:
- **modules/test: fix passthru example**
- **lib: move `enableExceptInTests` impl to `build.test` option**

Motivated by https://github.com/nix-community/nixvim/issues/3951 and similar instances of `enableExceptInTests` not working well with our less-abstracted APIs.

Simplify the `enableExceptInTests` attribute, removing the `_nixvimTests` argument.

We now do a full re-eval of the nixvim configuration before building the test, giving us a central place to implement `enableExceptInTests` and its eventual replacement(s).

This extends support for `enableExceptInTests` to all methods of getting a nixvim test derivation. Previously, it only worked when using `mkTestDerivationFromNixvimModule` and it only worked for `helpers`, not `lib.nixvim`.

I'm a bit concerned about the re-eval increasing out buildbot load. But I can't see an alternative way to achieve a "we are in test mode now" without things getting messy.

Maybe we could have it so that the `build.test` option only uses `extendModules` when not already in "test mode"? That way our test suite could explicitly set test mode on the initial eval, with no re-eval needed...

### DIFF
--- a/modules/top-level/test.nix
+++ b/modules/top-level/test.nix
@@ -3,6 +3,8 @@
   config,
   options,
   lib,
+  helpers,
+  extendModules,
   ...
 }:
 let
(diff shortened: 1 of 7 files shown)


MattSturgeon:
> Maybe we could have it so that the `build.test` option only uses `extendModules` when not already in "test mode"? That way our test suite could explicitly set test mode on the initial eval, with no re-eval needed...

Done

modules/top-level/test.nix
```diff
@@ -3,6 +3,8 @@
   config,
   options,
   lib,
+  helpers,
```
GaetanLepage, line 6: Is this `helpers` meant to stay?

MattSturgeon: Yes. It's needed for propagating `helpers` into `testConfiguration` (the re-eval):

https://github.com/nix-community/nixvim/blob/a3efacb1ad28da27be5ded516f225ed04e930825/modules/top-level/test.nix#L287

GaetanLepage: Ok! But in the long run, we will get rid of it when `helpers` will get effectively dropped, right?

MattSturgeon: Of course

GaetanLepage (approved):
LGTM. Such performance + cleaning PRs are always welcome :)

PR #3954, opened 2025-11-19
lib/evalNixvim: deprecate `specialArgs.helpers`

MattSturgeon:
- **treewide: remove internal use of `helpers` module arg**
- **lib/evalNixvim: deprecate `specialArgs.helpers`**

`helpers` has been soft-deprecated for several releases. The first commit removes our internal usage, then the second commit introduces an eval warning that prints when `helpers` is used.

The treewide change was done with a combination of grep, quickfix list, and LSP rename.

### DIFF
--- a/tests/enable-except-in-tests.nix
+++ b/tests/enable-except-in-tests.nix
@@ -20,6 +20,7 @@ let
             message = "Expected lib.nixvim.enableExceptInTests to be true";
           }
           {
+            # NOTE: evaluating `helpers` here prints an eval warning
             assertion = !helpers.enableExceptInTests;
             message = "Expected helpers.enableExceptInTests to be false";
           }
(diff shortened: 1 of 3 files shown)


tests/enable-except-in-tests.nix
```diff
@@ -1,35 +1,69 @@
...
-    inherit pkgs module;
+    inherit pkgs;
+    module =
+      { lib, helpers, ... }:
+      {
+        assertions = [
+          {
+            # FIXME: should be false
+            assertion = lib.nixvim.enableExceptInTests;
+            message = "Expected lib.nixvim.enableExceptInTests to be true";
```
MattSturgeon, line 20: This is an existing bug in `mkTestDerivationFromNixvimModule` and `enableExceptInTests` uncovered by updating the test to read `lib.nixvim` instead of `helpers`.

This is fixed by #3952

PR #3958, opened 2025-11-20
ci: enable stricter Bash error checking

trueNAHO:
```
Co-authored-by: Matt Sturgeon <matt@sturgeon.me.uk>
```

This PR is entirely inspired by https://github.com/nix-community/stylix/pull/2001.

I suppose the following does not work because it is not a top-level `/.github/workflows/<WORKFLOW>.yml` file and because my LSP reports a `Property defaults is not allowed. [513]` error:

```diff
diff --git a/.github/actions/build-docs/action.yml b/.github/actions/build-docs/action.yml
index c745974c..b7907c5b 100644
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -26,6 +26,10 @@ outputs:
     description: The ID of the artifact that was uploaded.
     value: ${{ steps.upload.outputs.artifact-id }}

+defaults:
+  run:
+    shell: bash
+
 runs:
   using: "composite"
   steps:
```

Hopefully, `/.github/actions/build-docs/action.yml` is already covered by the following declarations:

- https://github.com/nix-community/nixvim/blob/695b0b80f8452bc584adf23eb58bdc9f599e35eb/.github/actions/build-docs/action.yml#L34
- https://github.com/nix-community/nixvim/blob/695b0b80f8452bc584adf23eb58bdc9f599e35eb/.github/actions/build-docs/action.yml#L43

If this breaks certain workflows upon merging, feel free to revert the affected workflows.

CC: @MattSturgeon

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -8,6 +8,10 @@ on:
       - closed
       - labeled
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   backport:
     name: Backport Pull Request
--- a/.github/workflows/flakehub-publish-rolling.yml
+++ b/.github/workflows/flakehub-publish-rolling.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - "main"
+defaults:
+  run:
+    shell: bash
 jobs:
   flakehub-publish:
     if: github.repository == 'nix-community/nixvim'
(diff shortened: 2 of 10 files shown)


MattSturgeon:
> Hopefully, `/.github/actions/build-docs/action.yml` is already covered by the following declarations:

Correct. For composite actions it is actually required to explicitly state the shell to run with in each step. For workflows, specifying the shell (and the default shell) is optional.

PR #3960, opened 2025-11-21
modules/keymaps: fix `lua` option deprecation warning

MattSturgeon:
The deprecation warning for the keymap-submocule `lua` option relied on `getSubOptions`, however this is fundamentally flawed because that function uses a different module eval from the one that merges submodule definitions.

Since definitions are not used by `getSubOptions`, `options.lua.isDefined` will never be true.

Instead, we have two choices:

1. Add a `luaIsDefined` option to the keymap submodule
2. Use the new v2 merge's `valueMeta` to access the actual module eval

In this PR I've implemented option 2, and added tests to ensure the warning actually works. Actually testing that warnings and deprecated options work as expected is important, as demonstrated by this one being broken so long without us noticing.

Since the warning will not have been shown to any users, I've pushed back the "will be removed after" epoch. (Sadly, I started working on this because I was hoping to _delete_ all the deprecated `lua` option code... Then I noticed it was fundamentally flawed).

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -54,32 +54,42 @@ in
     # TODO remove entirely in 25.05?
     warnings =
       let
-        # All keymap options that have historically supported the `lua` sub-option
-        keymapOptions = [
-          options.keymaps
-          options.keymapsOnEvents
-          # NOTE: lsp `diagnostic` and `lspBuf` don't use `mapOptionSubmodule` yet
-          # So we only need `lua` deprecation in lsp's `extra` option
-          options.plugins.lsp.keymaps.extra
-          # NOTE: tmux-navigator added `mapOptionSubmodule` support _after_ branching off 24.05
-          options.plugins.tmux-navigator.keymaps
-        ]
-        # NOTE: barbar added `mapOptionSubmodule` support shortly _before_ branching off 24.05
-        ++ builtins.attrValues (builtins.removeAttrs options.plugins.barbar.keymaps [ "silent" ]);
+        keymapsUsingLua =
+          lib.pipe
+            # All keymap options that have historically supported the `lua` sub-option
+            [
+              options.keymaps.valueMeta.list
+              (lib.concatMap (ev: ev.list) (lib.attrValues options.keymapsOnEvents.valueMeta.attrs))
+
+              # NOTE: lsp `diagnostic` and `lspBuf` don't use `mapOptionSubmodule` yet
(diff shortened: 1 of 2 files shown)


VvSsNPk:
`… while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/hr5pl7m833kkr0zv98iydy7mxyl10qld-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating definitions from `/nix/store/a4la088hi5id7sw77gy9zdvwn83dgczz-source/nixos/common.nix':

       … while evaluating the option `home-manager.users.royaleinstein.warnings':

       … while evaluating definitions from `/nix/store/ykn2jzkrv3k7zcafjkyddyj31bd6f9bx-source/wrappers/_shared.nix':

       … while evaluating the option `home-manager.users.royaleinstein.programs.nixvim.warnings':

       … while evaluating definitions from `/nix/store/ykn2jzkrv3k7zcafjkyddyj31bd6f9bx-source/modules/keymaps.nix':

       error: attribute 'valueMeta' missing
       at /nix/store/ykn2jzkrv3k7zcafjkyddyj31bd6f9bx-source/modules/keymaps.nix:61:15:
           60|             [
           61|               options.keymaps.valueMeta.list
             |               ^
           62|               (lib.concatMap (ev: ev.list) (lib.attrValues options.keymapsOnEvents.valueMeta.attrs))
~`


Getting this error when updating nixvim, feel like this error has something to do with this change

MattSturgeon:
Thanks for reporting!

Do you have a really old nixpkgs revision somewhere?

If `lib` is being constructed from a nixpkgs revision that doesn't have https://github.com/NixOS/nixpkgs/pull/391544, then this error is expected.

We _can_ handle that edge-case better, though 🤔

AleAndForCode:
> Thanks for reporting!
> 
> Do you have a really old nixpkgs revision somewhere?
> 
> If `lib` is being constructed from a nixpkgs revision that doesn't have [NixOS/nixpkgs#391544](https://github.com/NixOS/nixpkgs/pull/391544), then this error is expected.
> 
> We _can_ handle that edge-case better, though 🤔

I've just updated my flake inputs and got the same error for latest both stable and unstable nixpkgs

SJ-Fang23:
Same here, In my case, Nixvim is following NixOS-Unstable branch of the NixPkgs. Maybe the ValueMeta update hasn't made it way into downstream  NixOS-Unstable branch?

PR #3962, opened 2025-11-21
flake: remove lib.<system>.helpers

MattSturgeon:
Remove the deprecated `<flake>.lib.<system>.helpers` output, deprecated in https://github.com/nix-community/nixvim/pull/2727 during the 25.05 cycle (shortly _after_ 24.11 branch off).

Maybe we should hold off on this until after 25.11 is released and merge this during the 26.05 cycle?

### DIFF
--- a/flake/lib.nix
+++ b/flake/lib.nix
@@ -32,9 +32,6 @@
         check = pkgs.callPackage ../lib/tests.nix {
           inherit lib self system;
         };
-
-        # NOTE: no longer needs to be per-system
-        helpers = lib.warn "nixvim: `<nixvim>.lib.${system}.helpers` has been moved to `<nixvim>.lib.nixvim` and no longer depends on a specific system" self.lib.nixvim;
       }
     )
   );


khaneliman:
> Maybe we should hold off on this until after 25.11 is released and merge this during the 26.05 cycle?

I'm at that mindset with a lot of deprecations right now. Just waiting for release branch to clean a lot of stuff up in HM/Nixpkgs/etc

GaetanLepage:
I would slightly lean towards "sooner is better". But I can understand why holding it off is less scary.

+0.5 for dropping it now.

khaneliman:
Just figured we'd follow the deprecation policy we wrote up https://github.com/nix-community/nixvim/blob/e13ceae64fbdf1475ee71d4170dcd2869e17cfff/MAINTAINING.md?plain=1#L22 If it hasn't been deprecated in a release yet, doesn't make sense to rush to remove for stable users. (I personally don't live on stable, but have just been trying to adhere to that concept) Also is so close to another cutoff I don't see the benefit of doing it early.

MattSturgeon:
> If it hasn't been deprecated in a release yet

> deprecated [during] the 25.05 cycle

The warning was in the 25.05 release, so it has been in _a_ release. But I still consider "one" release (6mo) to be the extreme minimum, 12mo isn't so bad in ordinary circumstances.

My vote also goes to keeping the warning in the 25.11 release, then dropping it early in 26.06.

I still welcome early review and/or approval, but let's hold off a few weeks on actually merging.

khaneliman:
> > If it hasn't been deprecated in a release yet
> 
> > deprecated [during] the 25.05 cycle
> 
> The warning was in the 25.05 release, so it has been in _a_ release. But I still consider "one" release (6mo) to be the extreme minimum, 12mo isn't so bad in ordinary circumstances.
> 
> My vote also goes to keeping the warning in the 25.11 release, then dropping it early in 26.06.
> 
> I still welcome early review and/or approval, but let's hold off a few weeks on actually merging.

Sorry, I'm absolutely brain wrecked the last couple months.. I misread it as it wasn't in the last release. I'm fine with removing if it was already in a stable release. (Still think it's close enough to new release we can be nicer by just waiting a slight bit longer since we've already gone this far)

GaetanLepage:
Ok, let's hold it then :)

MattSturgeon:
> Maybe we should hold off on this until after 25.11 is released and merge this during the 26.05 cycle?

With 25.11 branched off, this should be good to go.

PR #3964, opened 2025-11-21
lib/modules: replace nixvim builders

MattSturgeon:
This PR adds two new builder functions, `buildNixvim` and `testNixvim`, which provide a simpler interface for building and testing Nixvim configurations. This is the first step toward deprecating the older builders:
- `makeNixvim`
- `makeNixvimWithModule`
- `mkTestDerivationFromNvim`
- `mkTestDerivationFromNixvimModule`

Both functions accept any of the following:
- raw args for `evalNixvim`
- an evaluated Nixvim configuration
- a package with a Nixvim `config` passthru

A new helper, `wrapEvalNixvim`, normalizes these inputs so callers do not need to distinguish between forms.

Built packages now expose `config` and `options` in passthru, which allows them to be reused as input to `testNixvim`. Tests ensure that the new builders mirror the argument structure of `evalNixvim`.

Doccomments are provided, but prominent user documentation is deferred until the new interfaces have time to settle.

Top-level lib or flake aliases are also not added; for now these functions are only available under the `lib.nixvim.modules` namespace.

The `extend` function (added by the legacy `makeNixvimWithModule` standalone wrapper) is also not implemented for packages built without the legacy builder. Currently I'm unsure if we _should_ replace that interface, or if we should suggest users use alternative methods like `extendModules` and/or importing common modules into different configurations.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -1,6 +1,7 @@
 {
   pkgs,
   config,
+  options,
   lib,
   ...
 }:
@@ -319,6 +320,9 @@ in
               printInitPackage
             ];
           meta.mainProgram = "nvim";
+          passthru = {
+            inherit config options;
+          };
         };
 
         printInitPackage = pkgs.writeShellApplication {


lib/modules.nix
```diff
@@ -14,6 +14,12 @@ let
     applyExtraConfig = "It has been moved to `lib.plugins.utils`";
     mkConfigAt = "It has been moved to `lib.plugins.utils`";
   };
+
+  wrapEvalNixvim =
```
khaneliman, line 18: Could do with a docstring as well explaining the purpose and what it does for us smooth brains. 

Seems like this is the part that can handle a raw configuration or an already evaluated one?

lib/modules.nix
```diff
@@ -57,6 +63,48 @@ in
       }
       // extraSpecialArgs;
     };
+
+  /**
+    Build a Nixvim package.
+
+    This is a thin wrapper around `<nixvim-configuration>.config.build.package`.
```
khaneliman, line 70: This line kinda confused me in the context of the rest of the docstring. We are building a package but it says its a thin wrapper around the output of a built package? Or am I reading it wrong?

khaneliman (approved):
LGTM overall just couple thoughts on living doc stuff

lib/modules.nix
MattSturgeon: This is definitely challenging to phrase clearly.

It is saying that it returns the `build.package` _option_ from a nixvim _configuration_.

In module system terminology, a "configuration" is the canonical name for the result of evaluating modules using `evalModules`. I.e. it has `_type = "configuration";`.

_Technically_, `buildNixvim` also accepts anything that has a `config` attr, including packages with `passthru.config`. But I don't want to distract from the key point.

I'm guessing there's simply not enough context that we're talking about the module system?

khaneliman: Ah I see. so `...around the ... option` ?

lib/modules.nix
MattSturgeon: I got lazy because this is private/internal and was extracted from functions that already had doc-comments. But you're right; this has enough logic and complexity to warrant some explanation.

lib/modules.nix
MattSturgeon: Maybe, but "option" alone doesn't do it justice, since it _extracts_ the option value _from_ a configuration (or any duck that looks like a configuration).

Perhaps something like "this function returns the `build.package` option's value, as defined in the supplied nixvim configuration"?

khaneliman: Yeah, that sounds good to me.

lib/modules.nix
khaneliman: You're fine, just figured I'd try to suggest something to aid anyone working with it who might not be as familiar with the nixpkgs lib functions involved to make it easier to understand what's happening here.

lib/modules.nix
```diff
@@ -57,6 +57,31 @@ in
...
+    : One of:
+      1. Arguments for `evalNixvim`.
+      2. A Nixvim configuration, as produced by `evalNixvim`.
+      3. A Nixvim package, as produced by `buildNixvim`.
+
+    # Output
+
+    An installable Nixvim package.
+  */
+  buildNixvim = lib.mirrorFunctionArgs self.modules.evalNixvim (
```
GaetanLepage: Is the `buildNixvim` mentionned in the docstring the same as this `buildNixvim`?
Otherwise, wouldn't you feed `buildNixvim` with _a Nixvim package, as produced by `buildNixvim`_?

lib/modules.nix
```diff
@@ -82,6 +82,31 @@ in
...
+      3. A Nixvim package, as produced by `buildNixvim`.
+
+    # Output
+
+    A buildable Nixvim test.
+  */
+  testNixvim = lib.mirrorFunctionArgs self.modules.evalNixvim (
+    input:
+    let
+      configuration = if input ? config.build.package then input else self.modules.evalNixvim input;
```
GaetanLepage: Maybe overkilled, but we could factorize this line in a separate function. Maybe that won't help with understanding though.

PR #3970, opened 2025-11-22
lib/modules: replace nixvim builders [v2]

MattSturgeon:
- **modules/output: include `config` and `options` in `build.package`**
- **lib/modules: init `buildNixvim` and `buildNixvimWith`**
- **lib/modules: init `testNixvim` and `testNixvimWith`**

This is an alternative to #3964 with a subtly different interface.

This version introduces a `With` variant alongside each function. The `With` variant expects the same args as `evalNixvim`, similar to `evalModules` or `submoduleWith`, while the non-`With` variant expects a plain module (similar to `submodule`). The plain module variant also accepts a derivation or configuration.

For now, I've kept it simple rather than trying to DRY-up the similar function implementations.

Closes #3964

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -98,6 +98,47 @@ in
   buildNixvimWith = lib.mirrorFunctionArgs self.modules.evalNixvim (
     args: (self.modules.evalNixvim args).config.build.package
   );
+
+  /**
+    Build a Nixvim test derivation.
+
+    # Inputs
+
+    `input`
+    : One of:
+      1. A Nixvim module or a list of modules.
+      2. A Nixvim configuration.
+      3. A Nixvim package.
+
+    # Output
+
+    A buildable Nixvim test.
+  */
+  testNixvim =
+    input:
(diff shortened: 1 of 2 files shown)


lib/modules.nix
```diff
@@ -57,6 +57,88 @@ in
...
+      lib.throwIfNot (input ? config.build.package)
+        "buildNixvim: received a derivation without the expected `config` attribute."
+        input.config.build.package
+    else if lib.isType "configuration" input then
+      lib.throwIfNot (input ? config.build.package)
+        "buildNixvim: received a configuration without the expected `build.package` option."
+        input.config.build.package
+    else
+      self.modules.buildNixvimWith {
+        modules = [ input ];
```
GaetanLepage: Suggestion: We could also accept a list of modules here:

```suggestion
        modules = lib.toList input;
```

khaneliman (approved):
I agree this is easier to understand.

MattSturgeon:
Thanks all for your feedback!

Issue #3966, opened 2025-11-22
[BUG] <BUG> Nixvim cannot build with attribute 'valueMeta' missing after recent update


SJ-Fang23:
### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

After I updated flake input version today (Nov 22, 2025), Nixvim cannot build properly, reporting that attribute valueMeta is missing. Here's the home-manager output:

```
error:
       … while evaluating a branch condition
         at «github:nixos/nixpkgs/c58bc7f5459328e4afac201c5c4feb7c818d604b?narHash=sha256-1jQnuyu82FpiSxowrF/iFK6Toh9BYprfDqfs4BB%2B19M%3D»/lib/lists.nix:142:18:
          141|       len = length list;
          142|       fold' = n: if n == len then nul else op (elemAt list n) (fold' (n + 1));
             |                  ^
          143|     in

       … while calling the 'length' builtin
         at «github:nixos/nixpkgs/c58bc7f5459328e4afac201c5c4feb7c818d604b?narHash=sha256-1jQnuyu82FpiSxowrF/iFK6Toh9BYprfDqfs4BB%2B19M%3D»/lib/lists.nix:141:13:
          140|     let
          141|       len = length list;
             |             ^
          142|       fold' = n: if n == len then nul else op (elemAt list n) (fold' (n + 1));

       … while evaluating the option `warnings':

       … while evaluating definitions from `/nix/store/ykn2jzkrv3k7zcafjkyddyj31bd6f9bx-source/wrappers/_shared.nix':

       … while evaluating the option `programs.nixvim.warnings':

       … while evaluating definitions from `/nix/store/ykn2jzkrv3k7zcafjkyddyj31bd6f9bx-source/modules/keymaps.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute 'valueMeta' missing
       at «github:nix-community/nixvim/e13ceae64fbdf1475ee71d4170dcd2869e17cfff?narHash=sha256-hTWn0Z/BM5wzrnzY3RVNyXbguJBFGksatRtQvjCOqM8%3D»/modules/keymaps.nix:61:15:
           60|             [
           61|               options.keymaps.valueMeta.list
             |               ^
[...]

PR #3975, opened 2025-11-23
tests/platforms: init `nixos-hm` test

MattSturgeon:
Test using home-manager module via a NixOS-managed home-manager configuration.

I was hoping this might reproduce #3966, but it does not. Maybe this is a useful test permutation regardless?

### DIFF
--- a/tests/platforms/default.nix
+++ b/tests/platforms/default.nix
@@ -16,6 +16,7 @@ in
 }
 // lib.optionalAttrs isLinux {
   nixos-module = (callTest ./nixos.nix { }).config.system.build.toplevel;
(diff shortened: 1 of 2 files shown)


khaneliman (approved):

MattSturgeon:
I'd like to consider whether this is actually a valuable addition to the test suite, or just a waste of CI minutes.

Seeing as it didn't catch anything this time, and the underlying issue proved unrelated, I'm minded to close without merging.

Thoughts?

GaetanLepage:
I like the idea. But indeed, if it doesn't catch anything relevant, I would prefer to avoid merging it in.

PR #3976, opened 2025-11-23
modules/keymaps: check for `valueMeta` before using it

MattSturgeon:
Some users report `error: attribute 'valueMeta' missing`.

This happens when `lib` is outdated; e.g. an older unstable revision or a 25.05 revision.

Check that `valueMeta` is actually present before using it. Warn if it is not.

This is a workaround for #3966

That issue seems to be caused by how our wrapper-modules bootstrap Nixvim's extended lib:
https://github.com/nix-community/nixvim/blob/2606dc719488065e735ff0fc622d743566b21ad4/wrappers/_shared.nix#L73-L77

Specifically, we take the host configuration's `lib` as the basis for our extended lib. It may be better to use our flake-locked Nixpkgs' `lib` instead. We can achieve that by using our `lib.nixvim` flake output, and/or our `inputs.nixpkgs.lib` flake input.

I'm not 100% convinced we should merge this. Fixing the underlying issue is probably better. Regardless, I'm pushing this up for posterity and discussion.

### DIFF
--- a/modules/keymaps.nix
+++ b/modules/keymaps.nix
@@ -54,6 +54,15 @@ in
     # TODO remove entirely in 25.05?
     warnings =
       let
+        missingValueMeta = lib.filter (opt: !(opt ? valueMeta)) (
+          [
+            options.keymaps
+            options.keymapsOnEvents
+            options.plugins.lsp.keymaps.extra
+            options.plugins.tmux-navigator.keymaps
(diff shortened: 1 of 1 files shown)


saygo-png:
Since the underlying issue of lib being out of sync has been fixed i'm closing this.

PR #3988, opened 2025-11-26
plugins/conform: refactor the autoInstall option implementation

saygo-png:
Now when setting a package as broken you need to provide the package name. This is for ease of maintaining. Currently setting a package as broken loses the name. When a package gets fixed, this forces the maintainers to either find the commit where it wasn't broken and get the name, or try looking it up again with `nix-locate` and such. With this change you just have to remove the `state.broken` part

some renames for clarity:
few `x` values get a proper name
`sType` -> `state`

I also removed an awful case where i had two functions checking if something was `sType`, but one was before being made into a list and the other after. Now there is only one check and i just create a partition attrset using a map and fold.

This is unrelated to the changes here, but should we set `autoInstall` to `true` by default now? I think this is the behaviour users expect and the initial bugs have been ironed out.

### DIFF
--- a/plugins/by-name/conform-nvim/auto-install.nix
+++ b/plugins/by-name/conform-nvim/auto-install.nix
@@ -8,26 +8,27 @@ let
   inherit (builtins)
     filter
     isString
+    isFunction
     isAttrs
     attrValues
     attrNames
     concatMap
     partition
     ;
-
-  inherit (import ./formatter-packages.nix { inherit pkgs lib; }) sType formatter-packages;
-  sTypeList = attrValues sType;
-  isSTypeAttrSet = x: lib.elem (x.mark or null) sTypeList;
+  inherit (import ./formatter-packages.nix { inherit pkgs lib; }) states formatter-packages;
 in
 rec {
-  cleanMaybePackageList = filter (x: !isSTypeAttrSet x);
-
-  getPackageByName =
+  getPackageOrStateByName =
     { configuredFormatters, overrides }:
     name:
     let
       permittedNames = lib.optionals (lib.isAttrs configuredFormatters) (attrNames configuredFormatters);
-      isSType = x: elem x sTypeList;
+      stateList = map (state: lib.fix (lib.toFunction state)) (attrValues states);
+      isState =
+        maybePackage:
+        lib.throwIf (isFunction maybePackage) "The '${name}' conform-nvim formatter package is a function" (
+          elem maybePackage stateList
+        );
       notFoundMsg = ''
         A package for the conform-nvim formatter '${name}' could not be found.
         It is not a user defined formatter. Is the formatter name correct?
@@ -36,15 +37,17 @@ rec {
         overrides.${name} or formatter-packages.${name} or pkgs.${name}
           or (throwIfNot (elem name permittedNames) notFoundMsg null);
     in
-    if isSType maybePackage then
+    if isState maybePackage then
       {
-        inherit name;
-        mark = maybePackage;
+        wrong = {
+          inherit name;
+          mark = maybePackage;
+        };
       }
     else
-      maybePackage;
+      { right = maybePackage; };
 
-  mkWarnsFromMaybePackageList =
+  mkWarnsFromStates =
     opts: list:
     let
       mkWarn =
@@ -61,7 +64,7 @@ rec {
           }
         ];
     in
-    concatMap mkWarn (filter isSTypeAttrSet list);
+    concatMap mkWarn list;
 
   collectFormatters =
     formatters:
--- a/plugins/by-name/conform-nvim/default.nix
+++ b/plugins/by-name/conform-nvim/default.nix
@@ -245,21 +245,21 @@ lib.nixvim.plugins.mkNeovimPlugin {
     let
       inherit (cfg.autoInstall) enable enableWarnings;
       inherit (import ./auto-install.nix { inherit pkgs lib; })
-        getPackageByName
+        getPackageOrStateByName
         collectFormatters
-        cleanMaybePackageList
-        mkWarnsFromMaybePackageList
+        mkWarnsFromStates
         ;
-      getPackageByNameWith = getPackageByName {
+      getPackageOrStateByNameWith = getPackageOrStateByName {
         configuredFormatters = cfg.settings.formatters;
         inherit (cfg.autoInstall) overrides;
       };
-      names = collectFormatters (attrValues cfg.settings.formatters_by_ft or { });
-      packageList = map getPackageByNameWith names;
-      warns = (mkWarnsFromMaybePackageList opts) packageList;
+      formatterNames = collectFormatters (attrValues (cfg.settings.formatters_by_ft or { }));
+      packagesAndStates = lib.foldAttrs (item: acc: [ item ] ++ acc) [ ] (
+        map getPackageOrStateByNameWith formatterNames
+      );
     in
     {
-      warnings = lib.mkIf (enable && warns != [ ] && enableWarnings) warns;
-      extraPackages = lib.mkIf enable (cleanMaybePackageList packageList);
+      warnings = lib.mkIf (enable && enableWarnings) (mkWarnsFromStates opts packagesAndStates.wrong);
+      extraPackages = lib.mkIf enable packagesAndStates.right;
     };
 }
--- a/plugins/by-name/conform-nvim/formatter-packages.nix
+++ b/plugins/by-name/conform-nvim/formatter-packages.nix
@@ -3,32 +3,34 @@
   ...
 }:
 with pkgs;
-rec {
-  sType = {
-    broken = "broken";
-    darwinOnly = "Darwin only";
+let
+  states = {
+    broken = _package: "broken";
+    darwinOnly = _package: "Darwin only";
     unpackaged = "unpackaged";
   };
-
+in
+{
+  inherit states;
   formatter-packages = {
     # 2025-11-15 dependency swift is broken
     # https://github.com/NixOS/nixpkgs/issues/461474
-    swift = sType.broken;
-    swift_format = sType.broken;
-    swiftformat = sType.broken;
-    swiftlint = sType.broken;
+    swift = states.broken swift;
+    swift_format = states.broken swift-format;
+    swiftformat = states.broken swiftformat;
+    swiftlint = states.broken swiftlint;
 
     # 2025-10-12 build failure on Darwin
-    smlfmt = if stdenv.isDarwin then sType.broken else smlfmt;
+    smlfmt = if stdenv.isDarwin then states.broken smlfmt else smlfmt;
 
     # 2025-11-25 build failure
-    roc = sType.broken;
+    roc = states.broken roc;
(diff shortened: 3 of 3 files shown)


khaneliman:
> This is unrelated to the changes here, but should we set autoInstall to true by default now? I think this is the behaviour users expect and the initial bugs have been ironed out.

I know we've got a couple that might have that behavior now and it does ease the configuration significantly to default it that way. This was one of the few plugins that was a PIA because of having to find every package yourself.

khaneliman (approved):

plugins/by-name/conform-nvim/default.nix
```diff
@@ -245,21 +245,21 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
         ;
-      getPackageByNameWith = getPackageByName {
+      getPackageOrStateByNameWith = getPackageOrStateByName {
         configuredFormatters = cfg.settings.formatters;
         inherit (cfg.autoInstall) overrides;
       };
-      names = collectFormatters (attrValues cfg.settings.formatters_by_ft or { });
-      packageList = map getPackageByNameWith names;
-      warns = (mkWarnsFromMaybePackageList opts) packageList;
+      formatterNames = collectFormatters (attrValues cfg.settings.formatters_by_ft or { });
```
MattSturgeon: I know this is correct, but using `or` unwrapped in a function application is confusing for many.

I'd suggest clarifying as:

```suggestion
      formatterNames = collectFormatters (attrValues (cfg.settings.formatters_by_ft or { }));
```
Like it was before.

plugins/by-name/conform-nvim/default.nix
```diff
@@ -245,21 +245,21 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      packagesAndStates = lib.foldAttrs (item: acc: [ item ] ++ acc) [ ] (
+        map getPackageOrStateByNameWith formatterNames
```
MattSturgeon, line 258: I think we can inline `formatterNames` if we switch from `foldAttrs` to `foldlAttrs`:

```suggestion
      packagesAndStates = lib.foldlAttrs (ft: fmt: acc: [ (getPackageOrStateByNameWith fmt) ] ++ acc) [ ] (
        cfg.settings.formatters_by_ft or { }
```

(Also, chaining fold and map feels wrong, let's inline that too.)

plugins/by-name/conform-nvim/default.nix
```diff
@@ -245,21 +245,21 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      );
     in
     {
-      warnings = lib.mkIf (enable && warns != [ ] && enableWarnings) warns;
-      extraPackages = lib.mkIf enable (cleanMaybePackageList packageList);
+      warnings = lib.mkIf (enable && enableWarnings) ((mkWarnsFromStates opts) packagesAndStates.wrong);
```
MattSturgeon: Nit, these parens aren't needed:


```suggestion
      warnings = lib.mkIf (enable && enableWarnings) (mkWarnsFromStates opts packagesAndStates.wrong);
```

plugins/by-name/conform-nvim/formatter-packages.nix
```diff
@@ -4,84 +4,84 @@
 }:
 with pkgs;
 rec {
-  sType = {
-    broken = "broken";
-    darwinOnly = "Darwin only";
+  states = {
+    broken = _package: "broken";
```
MattSturgeon, line 8: If preferred, this could be written as:

```suggestion
    broken = lib.toFunction "broken";
```

Which is essentially implemented as:
```nix
toFunction = x: if isFunction x then x else (_: x)
```

plugins/by-name/conform-nvim/formatter-packages.nix
```diff
@@ -4,84 +4,84 @@
 }:
 with pkgs;
 rec {
```
MattSturgeon: Off topic: I was never keen on having _both_ a `rec` and a `with`. It makes finding where a binding is coming from very difficult and also leads to surprising name shadowing.

Is the `rec` only needed for `states`? If so, maybe we could switch to

```
with pkgs;
let
  states = {...};
in
{
  inherit states;
  ...
}
```

to make the scopes clearer.

I think we want the `with` before the `let in`, so that our `state` binding shadows any potential `pkgs.states` attr.

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -8,26 +8,28 @@ let
...
-  cleanMaybePackageList = filter (x: !isSTypeAttrSet x);
-
-  getPackageByName =
+  getPackageOrStateByName =
     { configuredFormatters, overrides }:
     name:
     let
       permittedNames = lib.optionals (lib.isAttrs configuredFormatters) (attrNames configuredFormatters);
-      isSType = x: elem x sTypeList;
+      stateList = map (v: if isFunction v then lib.fix v else v) (attrValues states);
```
MattSturgeon:
```suggestion
      stateList = lib.pipe states [
        attrValues
        lib.toFunction
        lib.fix
      ];
```

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -8,26 +8,28 @@ let
...
+      isState =
+        maybePackage:
+        assert lib.assertMsg (
+          !isFunction maybePackage
+        ) "One of the conform formatter packages is a function";
+        elem maybePackage stateList;
```
MattSturgeon: I often find `throwIf` and `throwIfNot` end up clearer than `assertMsg`:


```suggestion
      isState =
        maybePackage:
        lib.throwIf (isFunction maybePackage)
          "One of the conform formatter packages is a function"
          (elem maybePackage stateList);
```

plugins/by-name/conform-nvim/default.nix
saygo-png: I don't think this usage of foldlAttrs is correct <https://noogle.dev/f/lib/foldlAttrs>

plugins/by-name/conform-nvim/formatter-packages.nix
saygo-png: I'd rather keep `_package` as it clarifies the usage. You can also achieve the same thing using `lib.const`

plugins/by-name/conform-nvim/auto-install.nix
saygo-png: This doesn't work as the entire list from `attrValues` gets passed to `lib.toFunction`. A map is needed which makes the pipe look ugly. I'll just do
```nix
map (state: lib.fix (lib.toFunction state)) (attrValues states)
```
Sadly Nix does not have a compose operator. If it did it could look like this
```nix
map (lib.fix . lib.toFunction) (attrValues states)
```

plugins/by-name/conform-nvim/default.nix
MattSturgeon: I typed it out on my phone; it was meant as an example not a canonical patch.

Looks like I mixed up the param order, specifically the accumulator is first

So it should be:

```suggestion
      packagesAndStates = lib.foldlAttrs (acc: _ft: fmt: [ (getPackageOrStateByNameWith fmt) ] ++ acc) [ ] (
        cfg.settings.formatters_by_ft or { }
```

plugins/by-name/conform-nvim/auto-install.nix
MattSturgeon: Ah good spot. My initial suggestion should've been

```suggestion
      stateList = lib.pipe states [
        attrValues
        (map lib.toFunction)
        (map lib.fix)
      ];
```

plugins/by-name/conform-nvim/default.nix
saygo-png: I still need to use `collectFormatters` here which creates a list, i cant just use `formatters_by_ft` directly. So I'd rather keep it the way it is right now.

saygo-png:
Rebased, would love if we could merge this sooner rather than later since it will conflict on nearly every lock update.

saygo-png:
Thanks!

khaneliman:
Haven't had time to investigate.. just did a flake update and get an eval error with this change. 

```bash
nix run
Using saved setting for 'allow-import-from-derivation = false' from ~/.local/share/nix/trusted-settings.json.
Using saved setting for 'extra-substituters = https://khanelivim.cachix.org https://nix-community.cachix.org' from ~/.local/share/nix/trusted-settings.json.
Using saved setting for 'extra-trusted-public-keys = khanelivim.cachix.org-1:Tb0jsMlhXSJDtI2ISiGPBrvL1XIzQrWap80AiJuBGI0= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs=' from ~/.local/share/nix/trusted-settings.json.
error:
         at «github:nixos/nixpkgs/cd5fedfc384cb98d9fd3827b55f4522f49efda42?narHash=sha256-VwPgoDgnd628GdE3KyLqTyPF1WWh0VwT5UoKygoi8sg%3D»/lib/modules.nix:1256:5:
         1255|
         1256|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |     ^
         1257|   };

         at «github:nixos/nixpkgs/cd5fedfc384cb98d9fd3827b55f4522f49efda42?narHash=sha256-VwPgoDgnd628GdE3KyLqTyPF1WWh0VwT5UoKygoi8sg%3D»/lib/modules.nix:1256:21:
         1255|
         1256|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                     ^
         1257|   };

       … while evaluating definitions from `/nix/store/r3i0h1671yjfr8xzldmvbyffyrmx2y0m-source/modules/transposition.nix':

       … while evaluating definitions from `/nix/store/23086yn8jr7l1lgd2q2g58qndd8aya5k-source/flake, via option perSystem':

       … while evaluating definitions from `/nix/store/j0v6jcqnxfxm77i6kb42rv8dd55ckqzs-source/flake/flake-modules/auto.nix, via option perSystem':

       … while evaluating the option `build.package':

       … while evaluating definitions from `/nix/store/j0v6jcqnxfxm77i6kb42rv8dd55ckqzs-source/plugins/by-name/conform-nvim':

       error: attribute 'wrong' missing
       at «github:nix-community/nixvim/96d3538c32816992f30d858095aa93c9ee32fd45?narHash=sha256-OW19qwRisjzD1vQknzGXbZ5yqtULI1FU1f6HZdKn5TA%3D»/plugins/by-name/conform-nvim/default.nix:262:78:
          261|     {
          262|       warnings = lib.mkIf (enable && enableWarnings) (mkWarnsFromStates opts packagesAndStates.wrong);
             |                                                                              ^
          263|       extraPackages = lib.mkIf enable packagesAndStates.right;
```

saygo-png:
> Haven't had time to investigate.. just did a flake update and get an eval error with this change.

I'm so sorry I have no idea how this slipped past me. I think when i was working on it i might have had a warning emitting causing this to eval properly.

PR #3996, opened 2025-11-29
plugins/conform: fix eval error

saygo-png:
Fixes a serious bug introduced by https://github.com/nix-community/nixvim/pull/3988 preventing eval.
The fold result doesn't always have a `wrong` or `right` attribute, as the first only exists when there are warnings and the latter when packages are defined.

So sorry for introducing this, I don't know how i missed this while testing it locally on my config.

### DIFF
--- a/plugins/by-name/conform-nvim/default.nix
+++ b/plugins/by-name/conform-nvim/default.nix
@@ -259,7 +259,9 @@ lib.nixvim.plugins.mkNeovimPlugin {
       );
     in
     {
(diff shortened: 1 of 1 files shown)


MattSturgeon:
> So sorry for introducing this, I don't know how i missed this while testing it locally on my config.

Don't worry, things happen. Is it worth adding test cases to cover the scenarios where these attrs are missing?

Thanks for the quick fix!

PR #4002, opened 2025-11-30
ci/update: refactor script and unify lockfile commits

MattSturgeon:
Refactors the flake update script to simplify and unify lockfile updates. Root and dev lockfile updates are now captured, normalized, and committed together in a single consolidated commit. This removes the previous need to track HEAD commits, rely on `--commit-lock-file`/`--amend`, or manage separate commits for each lockfile.

Also:
- Allows running from a subdir by `cd`ing to the flake root.
- Avoids polluting the `result` symlink when building `version-info`.
- Configures `nixConfig.commit-lock-file-summary` in `flake.nix` to match the new default used by the script.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,6 +18,7 @@
   };
 
   nixConfig = {
+    commit-lock-file-summary = "flake: Update";
     extra-substituters = [ "https://nix-community.cachix.org" ];
     extra-trusted-public-keys = [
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Initial run without issue: https://github.com/nix-community/nixvim/actions/runs/19805171638/job/56738213234

#3998 updated correctly

PR #4006, opened 2025-12-01
plugins/haskell-tools: add hlsPackageFallback option

saygo-png:
This is needed as hls and ghc need to be synced to the same ghc version (down to the commit hash). I enable it by default since hls requires an exact match, so in practice this will always mismatch unless some shell provides it. Making the shell take priority by default makes the experience smoother for users.

I tested this and it works as intended with true/false and `hlsPackage` set to null

### DIFF
--- a/plugins/by-name/haskell-tools/default.nix
+++ b/plugins/by-name/haskell-tools/default.nix
@@ -18,6 +18,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
     hlsPackage = lib.mkPackageOption pkgs "haskell-language-server" {
       nullable = true;
     };
+    hlsPackageFallback = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        When enabled, hls will be added to the end of the `PATH` _(suffix)_ instead of the beginning _(prefix)_.
+
+        This can be useful if you want local versions of hls (e.g. from a devshell) to override the Nixvim version.
+      '';
+    };
   };
 
   settingsExample = {
@@ -43,7 +52,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     globals.haskell_tools = cfg.settings;
-    extraPackages = [ cfg.hlsPackage ];
+    extraPackages = lib.optionals (!cfg.hlsPackageFallback) [ cfg.hlsPackage ];
+    extraPackagesAfter = lib.optionals cfg.hlsPackageFallback [ cfg.hlsPackage ];
 
     plugins.telescope.enabledExtensions = lib.mkIf cfg.enableTelescope [ "ht" ];
     assertions = lib.nixvim.mkAssertions "plugins.haskell-tools" {


plugins/by-name/haskell-tools/default.nix
```diff
@@ -43,7 +52,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   extraConfig = cfg: {
     globals.haskell_tools = cfg.settings;
-    extraPackages = [ cfg.hlsPackage ];
+    extraPackages = lib.optionals (!cfg.hlsPackageFallback) [ cfg.hlsPackage ];
+    extraPackagesAfter = lib.optionals cfg.hlsPackageFallback [ cfg.hlsPackage ];
```
MattSturgeon, line 56: Alternatively:


```suggestion
    ${if cfg.hlsPackageFallback then "extraPackagesAfter" else "extraPackages"} = [ cfg.hlsPackage ];
```

MattSturgeon (approved):
This does beg the question: why do we install these dependencies at all, if they are almost guaranteed to be the wrong version?

I think we've discussed that before, but I don't remember the arguments either way.

PR LGTM, thanks.

saygo-png:
> This does beg the question: why do we install these dependencies at all, if they are almost guaranteed to be the wrong version?

Its very convienient to have a version installed for smaller and simpler projects that will compile with a range of ghc versions. You dont need a ghc from the shell then and dont run into version mismatches.

edit: i reread my original message and i was unclear: the mismatch happens when a shell provides ghc. if both ghc and hls are from nixvim then they will work together, but might not work on the project itself depending on what ghc it needs. In practice most haskell projects will provide a ghc version in a shell which will mismatch

plugins/by-name/haskell-tools/default.nix
saygo-png: drafting since id like to add this. Im going to bed right now though.

saygo-png: This throws infinite recursion

```
warning: Git tree '/home/samsepi0l/builds/nixvim' is dirty
error:
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'plugins-by-name'
         whose name attribute is located at /nix/store/vfis5qdflk6m21gd1b7j9d31shlns4wn-source/pkgs/stdenv/generic/make-derivation.nix:541:13

       … while evaluating attribute 'buildCommand' of derivation 'plugins-by-name'
         at /nix/store/vfis5qdflk6m21gd1b7j9d31shlns4wn-source/pkgs/build-support/trivial-builders/default.nix:80:17:
           79|         enableParallelBuilding = true;
           80|         inherit buildCommand name;
             |                 ^
           81|         passAsFile = [ "buildCommand" ] ++ (derivationArgs.passAsFile or [ ]);

       … while evaluating the option `_module.freeformType':

       … while evaluating the module argument `config' in "/nix/store/wvvs5d4d6z4h54g05mk76lqbawhlgvps-source/plugins/by-name/haskell-tools:anon-1":

       … if you get an infinite recursion here, you probably reference `config` in `imports`. If you are trying to achieve a conditional import behavior dependent on `config`, consider importing unconditionally, and using `mkEnableOption` and `mkIf` to control its effect.

```

MattSturgeon: Ah, yeah. That makes sense.

PR #4007, opened 2025-12-01
treewide: 25.05 -> 25.11

GaetanLepage:
The last? step before publicly announcing the release.


MattSturgeon:
> The last? step before publicly announcing the release.


SGTM. We also need to migrate a few flake inputs to point to their respective stable branches (e.g. home-manager). Let's check which ones have a stable branch already before announcing.

PR #4016, opened 2025-12-03
modules/extraFiles: refactor to use symlinks and support directories

### LINKED ISSUE #3992: [BUG] `extraFiles` fails when `source` is a directory
### Plugin

_No response_

### Nixpkgs Release

unstable

### Home Manager Release

None

### I have read the FAQ

### Description

Just the title.

Fails with the following error:
```
error: Cannot build '/nix/store/g6zhk1l7bc5w6rmmj9l0fbh7w1i0j23z-nvim-config.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/pdlwp52il52kgjkcl36yf03rffbzh166-nvim-config
       Last 1 log lines:
       > cp: -r not specified; omitting directory '/nix/store/15b81403vyciiqw98ihz61dn9y73x978-lua'
       For full logs, run:
         nix log /nix/store/g6zhk1l7bc5w6rmmj9l0fbh7w1i0j23z-nvim-config.drv
```

```nix
programs.nixvim = {
    # Where lua is a directory, for example, of lua files to include.
    extraFiles."example".source = ./lua;
};
```

MattSturgeon:
### Summary

This PR refines the `extraFiles` option to support directory sources, ensure deterministic conflict detection, and expand the test suite. This makes `extraFiles` more predictable, more flexible, and less wasteful in the Nix store.

---

#### 1. Directory support

`extraFiles` entries may now specify a **directory** as their source. The directory is symlinked into the final output as-is, allowing structured layouts without requiring users to use `builtins.readDir`.

#### 2. Deterministic conflict detection

The build now validates the merged set of targets and reports conflicts early.

It detects:

- **Duplicate targets**
  Allowed only when all definitions resolve to the same store path or two files that satisfy `diff -q`. Differing duplicates produce an explicit build error.

- **Prefix conflicts**
  Any target that is a prefix of another (including file vs directory) triggers an assertion. Multiple conflicting pairs are reported together.

#### 3. Store-efficient output

Sources are now installed using **symlinks** rather than copying. This avoids unnecessary Nix store duplication and matches current behaviour in the NixOS `/etc` module.

#### 4. Expanded and consolidated tests

The test suite now includes coverage for:

- Basic file installation
- empty `extraFiles`
- Nested paths
- Directory sources
- Unicode filenames and content
- Empty and whitespace-only files
- Expected failures:
  - mismatching duplicates
  - target path prefix conflicts

Fixes #3992 cc @daniellaing

### DIFF
--- a/modules/top-level/files/default.nix
+++ b/modules/top-level/files/default.nix
@@ -56,6 +56,19 @@ in
   config =
     let
       extraFiles = lib.filter (file: file.enable) (lib.attrValues config.extraFiles);
+      targets = lib.pipe extraFiles [
+        (builtins.groupBy (entry: entry.target))
+        # Explicitly copy to store using "${}"
+        (lib.mapAttrs (_: map (entry: "${entry.finalSource}")))
+      ];
+      prefixConflicts = lib.optionals (targets != { }) (
+        let
+          names = lib.attrNames targets;
+          pairs = lib.zipLists names (lib.tail names);
+        in
+        lib.filter ({ fst, snd }: lib.hasPrefix "${fst}/" snd) pairs
+      );
+
       concatFilesOption = attr: lib.flatten (lib.mapAttrsToList (_: builtins.getAttr attr) config.files);
     in
     {
@@ -63,7 +76,15 @@ in
       extraPlugins = concatFilesOption "extraPlugins";
       extraPackages = concatFilesOption "extraPackages";
       warnings = concatFilesOption "warnings";
-      assertions = concatFilesOption "assertions";
+      assertions =
+        concatFilesOption "assertions"
+        ++ lib.nixvim.mkAssertions "extraFiles" {
+          assertion = prefixConflicts == [ ];
+          message = ''
+            Conflicting target prefixes:
+            ${lib.concatMapStringsSep "\n" ({ fst, snd }: "  - ${fst} ↔ ${snd}") prefixConflicts}
+          '';
+        };
 
       # Add files to extraFiles
       extraFiles = lib.mkDerivedConfig options.files (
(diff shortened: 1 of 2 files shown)


daniellaing:
Excellent stuff, well done!

modules/top-level/files/default.nix
```diff
@@ -76,28 +76,92 @@ in
       );
 
       # A directory with all the files in it
-      # Implementation based on NixOS's /etc module
-      build.extraFiles = pkgs.runCommandLocal "nvim-config" { passthru.vimPlugin = true; } ''
-        set -euo pipefail
+      # Implementation inspired by NixOS's /etc module
+      build.extraFiles =
```
khaneliman, line 101: Do we have any documentation that needs to be updated to reflect the new capabilities?

modules/top-level/files/default.nix
```diff
@@ -76,28 +76,92 @@ in
...
-          cp "$src" "$out/$target"
-        }
+            # Validate all targets
+            for ((i=0; i<numTargets; i++)); do
+              for ((j=i+1; j<numTargets; j++)); do
+                checkConflict \
+                  "''${targets[i]}" "''${sources[i]}" \
+                  "''${targets[j]}" "''${sources[j]}"
+              done
+            done
```
khaneliman: Would be nice if we could refactor to avoid O(n²) nested loops, or do we think `extraFiles` will always be small enough it doesn't matter? 

Would we benefit from doing the validation in nix before the bash installation?

daniellaing: I think checking conflicts is inherently O(n^2) because each file must be checked against all the others, granted order does not matter (checking file A against file B is equivalent to checking file B against file A), hence the inner loop starts at `j = i+1`.

It may not be possible to escape O(n^2) even if nested loops are refactored out?

MattSturgeon: Good point.

I don't think conflict detection is *inherently* `O(n²)`, though the current bash implementation certainly is.

Duplicate target detection can be expressed at eval time using `builtins.groupBy`, producing attrs keyed by target:

```nix
{
  "targetA" = [ srcA srcB ];
  "targetB" = [ srcC ];
}
```

This avoids pairwise comparisons and lets us deterministically reject mismatched duplicates, while still deferring source content inspection to build time (which is needed to inspect derivation outputs **without** IFD).

Target prefix conflicts (`foo` vs `foo/bar`) could also be detected at evaluation time. Since `builtins.attrNames` returns a sorted list of targets, we can scan adjacent entries to check prefixes (`O(n)`), rather than comparing every pair explicitly (`O(n²)`).

That said, I agree with the point that `extraFiles` is expected to stay relatively small in practice. A refactor here would mainly be about clarity and shifting more validation logic from bash into Nix; performance is a nice bonus, though.

modules/top-level/files/default.nix
MattSturgeon: The option description is:

> Extra files to add to the runtime path

And its example is:

```nix
{
  "after/ftplugin/nix.lua".text = ''
    vim.opt_local.tabstop = 2
    vim.opt_local.shiftwidth = 2
    vim.opt_local.expandtab = true
  '';
}
```

Which I don't think necessarily need to change, unless you have any specific suggestions.

There's no mention of `extraFiles` elsewhere in the docs.

modules/top-level/files/default.nix
MattSturgeon: I've pushed up [a refactor](https://github.com/nix-community/nixvim/compare/e501ff530a86ef2a82f6724677dd7a85d9c15ffc..95e368f5409718b42a627cf5682aa122f18262ff) and a small [followup](https://github.com/nix-community/nixvim/compare/95e368f5409718b42a627cf5682aa122f18262ff..1c2cdf6fb4e8a92b05faab356f42f7bc3b000946), which seems much simpler overall and should also be `O(n)`.

MattSturgeon:
Odd, `x86_64-linux.tests.group-13` failed in CI, but `nix build --keep-going .#ci.buildbot.x86_64-linux.tests.group-13` is successful for me locally.

mathjiajia:
It seems that, `lua` files in the symlinked directory will not be compiled
is this by design?

MattSturgeon:
Correct. If you symlink to a directory it is included as-is. We do not look at its content in any way.

PR #4019, opened 2025-12-03
modules/files: drop coercedFileType & simplify byteCompileLua condition

MattSturgeon:
#### modules/files: drop `coercedFileType`

Defining `extraFiles.*` as a string was deprecated in 24.05, and has printed a warning since 2024-07-07.

We can now drop support entirely and simplify the implementation.

#### modules/files: simplify `byteCompileLua` condition

Avoid needing a `specialArgs.topConfig` by binding `byteCompileLua` outside of the submodule scope.

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -6,13 +6,13 @@
 }:
 let
   builders = lib.nixvim.builders.withPkgs pkgs;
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Better!

PR #4020, opened 2025-12-03
wrappers: expose platform wrapper modules via `build.*Module` options

### LINKED ISSUE #4009: wrappers: support `extraSpecialArgs`
Hi.

I use `nixvim` both standalone and with `home-manager`. I have custom modules
written for standalone usage, but `home-manager` the options are namespaced under
`programs.nixvim`. Therefore, I import my modules like so:

```nix
{
  self, # flake reference
  ...
}:
{
  programs = {
    nixvim = {
      enable = true;
      imports = with self.nixvimModules; [
        module-name
      ];
    };
  };
}
```

However, `module-name` also takes `self` and other custom args, which fails with 
infinite recursion.

I solve this by reimplementing the `nixvim` `home-manager` module to add my args as
`extraSpecialArgs`, which is suboptimal.

Is there a better solution than this? I think the best way to enable this would be
the approach `home-manager` takes under `NixOS` by having
`home-manager.extraSpecialArgs` option. I can submit a PR with this feature, if you
would be interested.

Thanks in advance.

MattSturgeon:
This change exposes the platform wrapper modules (`nixosModule`, `homeModule`, and `nixDarwinModule`) as standard output options under `build.*Module` within a Nixvim configuration. This allows users to take an existing Nixvim configuration and directly embed it into NixOS, Home Manager, or nix-darwin without re-importing modules into `programs.nixvim` manually.

To construct these wrapper modules we need to evaluate a minimal Nixvim configuration that contains no `pkgs` or `nixpkgs.hostPlatform`. Such a configuration would normally fail to eval, but evaluating it with `_module.check = false` is lazy enough to read the wrapper options. This base module is marked with a unique key and is later disabled by `wrappers/_shared.nix` to prevent the no-check setting from leaking into user configurations.

Fixes #4009

### DIFF
--- a/flake/wrappers.nix
+++ b/flake/wrappers.nix
@@ -4,10 +4,26 @@
   ...
 }:
 let
-  inherit (lib.modules) importApply;
   # Added 2025-05-25; warning shown since 2025-08-01 (25.11)
   # NOTE: top-level binding of a fully resolved value, to avoid printing multiple times
   homeManagerModulesWarning = lib.warn "nixvim: flake output `homeManagerModules` has been renamed to `homeModules`." null;
+
+  # A base configuration used to evaluate the wrapper modules.
+  #
+  # While we don't define a `pkgs` or `hostPlatform` here, which would normally
+  # lead to eval errors, disabling option-declaration checking gives us enough
+  # laziness to evaluate the options we need.
+  #
+  # The `_module.check` module has a key, so we can disable it later in the
+  # platform wrapper modules.
+  configuration = self.lib.evalNixvim {
+    modules = [
+      {
+        key = "<internal:nixvim-nocheck-base-eval>";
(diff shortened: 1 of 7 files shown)


MattSturgeon:
The edge-case where a user configures nixvim with platform-a, and then uses that configuration's `nixosModule` on a NixOS configuration using platform-b, may need handling better.

That _probably_ means choosing a higher override priority when propagating the host's `system` to `programs.nixvim.nixpkgs.hostPlatform`:
https://github.com/nix-community/nixvim/blob/8d085d4094d84fced3b1230c890ddd27c97a8d20/wrappers/_shared.nix#L42-L44

Not necessarily in-scope for this PR though.

khaneliman (approved):

PR #4028, opened 2025-12-05
lib: reimplement `call`; drop `.override`

MattSturgeon:
Using `callPackageWith` adds `override` and `overrideDerivation` attributes to the result, which are not relevant to subsections of Nixvim's lib section.

Implement our own, simplified, `call` using `intersectAttrs` and `functionArgs`.

If users wish to modify parts of our lib section, they should extend the top-level fixpoint to ensure everything stays in sync. This can be done via the `lib.nixvim.extend` or `lib.extend` functions.

### DIFF
--- a/lib/top-level.nix
+++ b/lib/top-level.nix
@@ -22,13 +22,21 @@ lib.makeExtensible (
   self:
   let
     # Used when importing parts of our lib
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
LGTM

PR #4031, opened 2025-12-05
docs: discourage use of `follows`

saygo-png:
A lot of issues with Nixvim are caused by using `follows`. As I understand this made much more sense before Nixvim had it's own nixpkgs instance in non-standalone modes. Now follows should generally not be used even when Nixvim is a nixos/hm module.

This is also why i removed some text putting emphasis on having nixpkgs versions match, as to my understanding we aren't as sensitive to mismatches anymore even on module usage.

Also some minor changes.

### DIFF
--- a/docs/user-guide/faq.md
+++ b/docs/user-guide/faq.md
@@ -83,17 +83,13 @@ When using Nixvim, it is possible to encounter errors about something not being 
  error: <name> cannot be found in pkgs
 ```
 
-This usually means one of two things:
+This usually means one of few things:
+- You are using `follows` on `inputs.nixvim` causing Nixvim to have an unexpected version of Nixpkgs.
 - The nixpkgs version is not in line with Nixvim (for example nixpkgs nixos-25.11 is used with Nixvim master)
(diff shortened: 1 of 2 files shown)


GaetanLepage (approved):
Good idea. Diff LGTM

khaneliman (approved):

MattSturgeon (approved):
Awesome, this is much clearer for new users.

I will caveat that while I'm happy to _recommend_ avoiding `follows` because we have a test suite, I'd still rather Nixvim _generally_ work well with relatively any up-to-date Nixpkgs from the correct release version.

- Docs could be more nuanced discussing the pros/cons of `follows` (relevant here).
- Code should be as lazy as possible, so that issues caused by mismatches only show up when the affected features are actually enabled.
- New package attrs, or renames, should be used with caution — e.g. package rename aliases allow the old name to work across a range of Nixpkgs revisions, both before and after the rename.
  On the flip side, some people use `config.allowAliases = false` — but surely that adds undue friction?

We should probably mention these considerations in CONTRIBUTING.md or MAINTAINING.md so that it's a documented policy, too.

PR #4041, opened 2025-12-07
plugins/opencode: add lsof to extraPackages

simonwjackson:
## Summary
- Add lsof to extraPackages for opencode.nvim plugin

opencode.nvim uses lsof to detect running opencode instances in the current working directory. Without lsof available, the plugin spawns new terminal instances instead of reusing existing ones.

### DIFF
--- a/plugins/by-name/opencode/default.nix
+++ b/plugins/by-name/opencode/default.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ lib, pkgs, ... }:
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "opencode";
   package = "opencode-nvim";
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Looking at [health.lua](https://github.com/NickvanDyke/opencode.nvim/blob/main/lua/opencode/health.lua), it seems that `pgrep` and `lsof` are required on `win32`. However, it does need `curl` and `opencode`.

Such a plugin runtime dependency issue should be handled in nixpkgs directly.
I opened https://github.com/NixOS/nixpkgs/pull/468818.

MattSturgeon:
Thanks for highlighting this issue!

Closing in favour of https://github.com/NixOS/nixpkgs/pull/468818

PR #4042, opened 2025-12-08
plugins/sidekick/assertion: no copilot dependency when nes disabled

alisonjenkins:
When the Sidekick plugin has Next Edit Suggestions disabled we should not enforce the dependency on Copilot.

See:

https://github.com/folke/sidekick.nvim?tab=readme-ov-file#can-i-use-this-without-nes-just-for-cli-tools

### DIFF
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -17,10 +17,12 @@ lib.nixvim.plugins.mkNeovimPlugin {
     "opencode"
   ];
 
-  extraConfig = {
+  extraConfig = cfg: {
     assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
-      assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
-      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
+      assertion =
+        (cfg.settings.opts.nes.enabled or true)
+        -> (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable);
+      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled when NES is enabled";
     };
   };
 


plugins/by-name/sidekick/default.nix
```diff
@@ -17,12 +17,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-    assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
-      assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
-      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
+  extraConfig =
+    let
+      nesEnabled = config.plugins.sidekick.settings.opts.nes.enabled or true;
+    in
+    {
+      assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
+        assertion = if nesEnabled then (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable) else true;
```
MattSturgeon: The current assertion uses `if x then y else true`, which can be expressed more concisely as a logical implication (`->`):

```nix
        assertion = nesEnabled -> (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable);
````

This is equivalent to `!nesEnabled || (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable)`, but more readable.

File: plugins/by-name/sidekick/default.nix [hunk 2]
```diff
@@ -17,12 +17,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
     "opencode"
   ];
 
-  extraConfig = {
-    assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
-      assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
-      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
+  extraConfig =
+    let
+      nesEnabled = config.plugins.sidekick.settings.opts.nes.enabled or true;
```
MattSturgeon: We can use `cfg` here as a shorthand for `config.plugins.sidekick`, if we change `extraConfig` to a function taking `cfg` as a parameter.

plugins/by-name/sidekick/default.nix
```diff
@@ -17,12 +17,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
     "opencode"
   ];
 
-  extraConfig = {
-    assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
-      assertion = config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable;
-      message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled";
+  extraConfig =
+    let
```
MattSturgeon: We can avoid changing `extraConfig`'s indentation by moving this `let in` block down to the `assertion` attribute. Seeing as that is the only place where `nesEnabled` is used.

plugins/by-name/sidekick/default.nix
alisonjenkins: I tried this but it resulted in: 
```
warning: Git tree '/home/ali/git/neovim-nix-flake' is dirty
error:
         at /nix/store/55z53i8nflvx4y9662rc0b6xb9y43b66-source/lib/modules.nix:1253:5:
         1252|
         1253|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |     ^
         1254|   };

         at /nix/store/55z53i8nflvx4y9662rc0b6xb9y43b66-source/lib/modules.nix:1253:21:
         1252|
         1253|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                     ^
         1254|   };

       … while evaluating definitions from `/nix/store/r3i0h1671yjfr8xzldmvbyffyrmx2y0m-source/modules/transposition.nix':

       … while evaluating definitions from `/nix/store/jd20jxgy0cysf6pzdxs8l8airfkpfin5-source/flake.nix, via option perSystem':

       … while evaluating the option `_module.freeformType':

       … while evaluating the module argument `config' in "/nix/store/laqfxl6yn8hq8333hflk10liqvgdycpy-source/plugins/by-name/sidekick:anon-1":

       … if you get an infinite recursion here, you probably reference `config` in `imports`. If you are trying to achieve a conditional import behavior dependent on `config`, consider importing unconditionally, and using `mkEnableOption` and `mkIf` to control its effect.

```

plugins/by-name/sidekick/default.nix
alisonjenkins: Done, the implication operator is new to me. Seems very useful though I will have to see what other languages I use support it.

plugins/by-name/sidekick/default.nix
alisonjenkins: I just inlined the check.

plugins/by-name/sidekick/default.nix
MattSturgeon: Did you add `cfg` to the main module args? That'd produce that error.

Instead we get `cfg` as:

```nix
extraConfig = cfg: {
```

alisonjenkins: Refactored to use cfg as requested.

PR #4049, opened 2025-12-09
docs: clarify lazyLoad.after behavior and mention luaConfig.post

### LINKED ISSUE #4048: [Feature] Allow extending `lazyLoad.settings.after` for lz.n
**Updated**: sorry, I just realized that `luaContent.post` is what extract I am looking for. I have open a PR to improve the document. See PR for details.

## Summary

Right now, when lazyLoad.enable = true, Nixvim will automatically route the generated setup code (luaConfig.content) into the lazy loading provider (lz.n), and use it as the after function if `lazyLoad.settings.after` is not set.

But as soon as I set `lazyLoad.settings.after`, I can only override this behavior. There is no way to extend the default after (e.g. “keep what Nixvim would do, plus a bit more”).

## Example use case

I want to lazy load catppuccin and let Nixvim do its normal setup for it. After catppuccin has been loaded, configure bufferline integration. This order is required as documented in [catppuccin/nvim](https://github.com/catppuccin/nvim).

With the current design:

1. Nixvim will generate a lua function
2. If I now set `lazyLoad.settings.after`, I lose this default after and must completely reimplement it, just to add my bufferline color setup, for example:

```nix
plugins.bufferline = {
  enable = true;
  settings.highlights.__raw = ''require("catppuccin.special.bufferline").get_theme()'';
  lazyLoad.settings.lazy = true; # Lazy load manually
};

plugins.catppuccin = {
  enable = true;
  # `plugins.catppuccin.settings` is invalid, we have to reimplement it
  lazyLoad.settings.after.__raw = ''
    function()
      require("catppuccin").setup({
        -- plugins.catppuccin is invalid now
        flavour = "mocha";
        transparent_background = true;
      })
      require('lz.n').trigger_load("bufferline.nvim")
    end
  '';
};
```

The key point: I don’t want to change what Nixvim already generates, I just want to append some logic when the lazy-loaded plugin finishes its normal setup.

## Requested change

It would be very helpful if Nixvim provided a way to extend the default after instead of only overriding it. For example:

**Add an extra hook like `lazyLoad.settings.afterExtra`, which runs after the default after.**

This would allow configs like:

```nix
plugins.catppuccin = {
  enable = true;
  settings = {
    flavour = "mocha";
    transparent_background = true;
  };
  # only my extra part, keep default catppuccin setup
  lazyLoad.settings.afterExtra.__raw = ''
    require('lz.n').trigger_load("bufferline.nvim")
  '';
};
```

## Workaround

Completely override `lazyLoad.settings.after.__raw = ''function() ... end`, and manually re-call the plugin’s setup (duplicating what Nixvim would have generated), plus my extra code.

This works, but:

- It’s fragile (can get out of sync with Nixvim’s generated setup).
- It’s verbose for a very common pattern: “run the default setup, then do X”.

If there’s already an intended way to extend after without replacing it, I’d love to know; otherwise, I hope this feature can be considered.

yzx9:
This PR updates the *Lazy Loading* user guide to better explain how per-plugin
lazy loading works and how `lazyLoad.settings.after` interacts with the
auto-generated setup code.

This should make it clearer when to override `after` vs. when to use
`luaConfig.post` for post-setup configuration (e.g. configuring another plugin's
colors after a colorscheme is loaded).

close #4048

### DIFF
--- a/docs/user-guide/lazy-loading.md
+++ b/docs/user-guide/lazy-loading.md
@@ -31,11 +31,21 @@ provider's loading mechanism.
 If you are just wanting to store potential configuration without enabling it,
 you can explicitly disable it setting `lazyLoad.enable = false`.
 
-By default, we route the generated setup code to your lazy loading provider, but
-it can be overridden by using the `lazyLoad.settings` option.
+By default, Nixvim routes the generated setup code (defined in
+`plugins.<name>.luaConfig.content`) to your lazy loading provider. For the
+`lz.n` backend, this means Nixvim will automatically generate an `after`
+function for you, that will be called when a trigger condition is met. You can
+override this default by defining `lazyLoad.settings.after` yourself.
 
-For `Lz.n`, the `lazyLoad.settings.after` option contains what it will call when
-a trigger condition is met.
+> [!NOTE]
+> This is an **override**: once you set `lazyLoad.settings.after`, Nixvim will
+> use your function instead of the auto-generated one, it will not "extend"
+> or "wrap" the default behavior.
+>
+> If you only need to run some extra Lua **after** the plugin (or colorscheme) has
+> been configured, you usually do **not** need to override `lazyLoad.settings.after`.
+> Instead, you can use the module's `luaConfig.post` hook, see the plugin's /
+> colorscheme's documentation for details
 
 ## Configuring Triggers
 
@@ -124,5 +134,31 @@ colorscheme = "catppuccin";
 colorschemes.catppuccin = {
   enable = true;
   lazyLoad.enable = true;
+};
+```
+
+To configure special integrations after `catppuccin` has been set up (while
+still letting Nixvim manage lazy loading and the default `after`):
+
+```nix
+colorscheme = "catppuccin";
+colorschemes.catppuccin = {
+  enable = true;
+  lazyLoad.enable = true;
+
+  # This code runs after catppuccin is setup,
+  # regardless of whether it was lazy-loaded or not.
+  luaConfig.post = ''
+    -- At this point catppuccin is configured, so we can safely
+    -- derive bufferline highlights or similar settings from it.
+    require('lz.n').trigger_load("bufferline.nvim")
+  '';
+};
+
+# Configure bufferline to load after catppuccin
+plugins.bufferline = {
+  enable = true;
+  settings.highlights.__raw = "require('catppuccin.special.bufferline').get_theme()";
+  lazyLoad.settings.lazy = true; # Lazy load manually
 };
 ```


docs/user-guide/lazy-loading.md
```diff
@@ -126,3 +135,28 @@ colorschemes.catppuccin = {
...
+  luaConfig.post = ''
+    -- At this point catppuccin is configured, so we can safely
+    -- derive bufferline highlights or similar settings from it.
+    require('lz.n').trigger_load("bufferline.nvim")
+  '';
+};
+
+plugins.bufferline = {
+  enable = true;
+  settings.highlights.__raw = ''require("catppuccin.special.bufferline").get_theme()'';
```
GaetanLepage:
```suggestion
  settings.highlights.__raw = "require('catppuccin.special.bufferline').get_theme()";
```
(nit)

yzx9: fixed

docs/user-guide/lazy-loading.md
```diff
@@ -31,11 +31,20 @@ provider's loading mechanism.
 If you are just wanting to store potential configuration without enabling it,
 you can explicitly disable it setting `lazyLoad.enable = false`.
 
-By default, we route the generated setup code to your lazy loading provider, but
-it can be overridden by using the `lazyLoad.settings` option.
+By default, we route the generated setup code (what Nixvim would normally put in
+`luaConfig.content`) to your lazy loading provider. For the `lz.n` backend, this
```
MattSturgeon: "what Nixvim would normally put in `luaConfig.content`" is slightly misleading.

Nixvim _always_ defines `plugins.<name>.luaConfig.content`. The variable here is whether nixvim _also_ defines `plugins.lz-n.plugins.*.after`, in which case the `after` definition is derived from `luaConfig.content`'s value.

Saying "normally" implies that we don't define `luaConfig.content` when using lazyloading.


```suggestion
By default, Nixvim routes the generated setup code (defined in `plugins.<name>.luaConfig.content`)
to your lazy loading provider. For the `lz.n` backend, this
```

docs/user-guide/lazy-loading.md
```diff
@@ -31,11 +31,20 @@ provider's loading mechanism.
...
+means we automatically generate an `after` function for you, that will be called
+when a trigger condition is met.
 
-For `Lz.n`, the `lazyLoad.settings.after` option contains what it will call when
-a trigger condition is met.
+You can override this default by providing `lazyLoad.settings.after` yourself.
```
MattSturgeon: Terminology: in the module system the `config` block "defines" options:

```suggestion
You can override this default by defining `lazyLoad.settings.after` yourself.
```

Or, sometimes you "override" lower-priority definitions:

```suggestion
You can override this default by overriding `lazyLoad.settings.after`.
```

docs/user-guide/lazy-loading.md
```diff
@@ -31,11 +31,20 @@ provider's loading mechanism.
...
+Note that this is an **override**: once you set `lazyLoad.settings.after`, Nixvim
+will use your function instead of the auto-generated one, it will not "extend"
+or "wrap" the default behavior.
+
+If you only need to run some extra Lua **after** the plugin (or colorscheme) has
+been configured, you usually do **not** need to override `lazyLoad.settings.after`.
+Instead, you can use the module's `luaConfig.post` hook, see the plugin's /
+colorscheme's documentation for details
```
MattSturgeon: Would this section benefit from being a markdown admonishment?

```md
> [!NOTE]
> ...
```

(or maybe "caution" or "important"?)

> [!CAUTION]
> Very dangerous tips

docs/user-guide/lazy-loading.md
```diff
@@ -126,3 +135,28 @@ colorschemes.catppuccin = {
   lazyLoad.enable = true;
 };
 ```
+
+To configure special integrations after `catppuccin` has been set up (while
+still letting Nixvim manage lazy loading and the default `after`):
+
+```nix
+colorscheme = "catppuccin";
```
MattSturgeon, line 144: Isn't this implied by enabling the colorscheme module?

docs/user-guide/lazy-loading.md
```diff
@@ -126,3 +135,28 @@ colorschemes.catppuccin = {
...
+colorschemes.catppuccin = {
+  enable = true;
+  lazyLoad.enable = true;
+
+  # This code runs after catppuccin's setup has completed,
```
MattSturgeon:
```suggestion
  # This code runs after catppuccin is setup,
```

docs/user-guide/lazy-loading.md
```diff
@@ -126,3 +135,28 @@ colorschemes.catppuccin = {
...
+    -- derive bufferline highlights or similar settings from it.
+    require('lz.n').trigger_load("bufferline.nvim")
+  '';
+};
+
+plugins.bufferline = {
+  enable = true;
+  settings.highlights.__raw = "require('catppuccin.special.bufferline').get_theme()";
+  lazyLoad.settings.lazy = true; # Lazy load manually
+};
```
MattSturgeon, line 163: I don't see how this is related to the catpuccin example? Should this be in another example/codeblock?

MattSturgeon, line 163: To be valid nix, this example should be wrapped as an attrset:


```suggestion
{
  colorscheme = "catppuccin";
  colorschemes.catppuccin = {
    enable = true;
    lazyLoad.enable = true;

    # This code runs after catppuccin's setup has completed,
    # regardless of whether it was lazy-loaded or not.
    luaConfig.post = ''
      -- At this point catppuccin is configured, so we can safely
      -- derive bufferline highlights or similar settings from it.
      require('lz.n').trigger_load("bufferline.nvim")
    '';
  };

  plugins.bufferline = {
    enable = true;
    settings.highlights.__raw = "require('catppuccin.special.bufferline').get_theme()";
    lazyLoad.settings.lazy = true; # Lazy load manually
  };
};
```

yzx9: Maybe it’s a good idea, but I noticed that none of the examples on this page are valid Nix. Should I fix all of them as well?

yzx9: It's a valuable example here because:
1. It's a real-world case where `bufferline` must be set up after `catppuccin` is loaded.
2. It shows how to trigger a plugin to load manually by setting `lazyLoad.settings.lazy = true`.

docs/user-guide/lazy-loading.md
yzx9: I suspect it's different when setting up multiple colorschemes.

docs/user-guide/lazy-loading.md
MattSturgeon: Maybe a comment could introduce this part of the example, something like "configure bufferline to load after catppuccin"?

MattSturgeon: Ah, they probably should be. But this PR shouldn't be responsible for fixing the other examples.

Feel free to match the existing (invalid) style and dismiss this feedback. If you have the motivation, a follow-up PR to make the examples "valid" would be great!

docs/user-guide/lazy-loading.md
```diff
@@ -31,11 +31,21 @@ provider's loading mechanism.
 If you are just wanting to store potential configuration without enabling it,
 you can explicitly disable it setting `lazyLoad.enable = false`.
 
-By default, we route the generated setup code to your lazy loading provider, but
-it can be overridden by using the `lazyLoad.settings` option.
+By default, Nixvim route the generated setup code (defined in
+`plugins.<name>.luaConfig.content`) to your lazy loading provider. For the
+`lz.n` backend, this means Nixvim automatically generate an `after` function
+for you, that will be called when a trigger condition is met. You can override
+this default by defining `lazyLoad.settings.after` yourself.
```
khaneliman:
```suggestion
By default, Nixvim routes the generated setup code (defined in
`plugins.<name>.luaConfig.content`) to your lazy loading provider. For the
`lz.n` backend, this means Nixvim will automatically generate an `after` function
for you, that will be called when a trigger condition is met. You can override
this default by defining `lazyLoad.settings.after` yourself.
```

yzx9: Haha :) my bad English.

PR #4057, opened 2025-12-11
plugins/efmls-configs: rename setup to languages

GaetanLepage:
- **plugins/efmls-configs: rename setup to languages**
- **plugins/efmls-configs: inline efm language settings (cosmetic)**

### DIFF
--- a/plugins/by-name/efmls-configs/default.nix
+++ b/plugins/by-name/efmls-configs/default.nix
@@ -183,32 +183,6 @@ lib.nixvim.plugins.mkNeovimPlugin {
         map (
           tool: if lib.isString tool then lib.nixvim.mkRaw "require 'efmls-configs.${kind}.${tool}'" else tool
         ) (lib.toList opt);
-
-      setupOptions =
-        (lib.mapAttrs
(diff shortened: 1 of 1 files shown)


generated/all-maintainers.nix
MattSturgeon: Unrelated?

plugins/by-name/efmls-configs/default.nix
```diff
@@ -21,8 +22,18 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
     })
 
+    (
+      let
+        basePluginPath = [
+          "plugins"
+          "efmls-configs"
+        ];
+      in
+      lib.mkRenamedOptionModule (basePluginPath ++ [ "setup" ]) (basePluginPath ++ [ "languages" ])
```
MattSturgeon, line 32: We also need to handle this rename in `tests/test-sources/plugins/by-name/efmls-configs/default.nix` (see CI failure)

MattSturgeon (changes requested):
Otherwise LGTM

PR #4058, opened 2025-12-11
plugins/dap: allow lua for adapters

khaneliman:
Support `mkRaw` for adapter definitions to support more flexible configuration.

https://codeberg.org/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation

### DIFF
--- a/plugins/by-name/dap/dapHelpers.nix
+++ b/plugins/by-name/dap/dapHelpers.nix
@@ -98,7 +98,11 @@ rec {
 
   mkAdapterOption =
     name: type:
-    mkNullOrOption (with types; attrsOf (either str type)) ''
+    let
+      # TODO: Added 2025-12-11 (26.05)
+      strToRawLua = lib.nixvim.deprecation.transitionType types.str lib.nixvim.mkRaw types.rawLua;
+    in
+    mkNullOrOption (with types; attrsOf (either strToRawLua type)) ''
       Debug adapters of `${name}` type.
       The adapters can also be set to a function which takes three arguments:
 
@@ -150,8 +154,8 @@ rec {
     type: adapters:
     lib.mapAttrs (
       _: adapter:
-      if builtins.isString adapter then
-        lib.nixvim.mkRaw adapter
+      if lib.types.isRawType adapter then
+        adapter
       else
         lib.filterAttrs (n: _: n != "enrichConfig") (
           adapter
--- a/plugins/by-name/dap/default.nix
+++ b/plugins/by-name/dap/default.nix
@@ -22,10 +22,64 @@ lib.nixvim.plugins.mkNeovimPlugin {
   deprecateExtraOptions = true;
 
   extraOptions = {
-    adapters = lib.nixvim.mkCompositeOption "Dap adapters." {
-      executables = dapHelpers.mkAdapterOption "executable" dapHelpers.executableAdapterOption;
-      servers = dapHelpers.mkAdapterOption "server" dapHelpers.serverAdapterOption;
-      pipes = dapHelpers.mkAdapterOption "pipe" dapHelpers.pipeAdapterOption;
+    adapters = lib.nixvim.mkNullOrOption' {
+      type = types.submodule {
+        freeformType = types.attrsOf types.rawLua;
+        options = {
+          executables = dapHelpers.mkAdapterOption "executable" dapHelpers.executableAdapterOption;
+          servers = dapHelpers.mkAdapterOption "server" dapHelpers.serverAdapterOption;
+          pipes = dapHelpers.mkAdapterOption "pipe" dapHelpers.pipeAdapterOption;
+        };
+      };
+      description = ''
+        Debug Adapter Protocol adapters.
+
+        Adapters can be defined as dynamic functions or categorized structured configs.
+        See `Example` for usage patterns.
+      '';
+      example = lib.literalExpression ''
+        {
+          # Dynamic adapter using raw lua function
+          # Useful when adapter type is determined at runtime
+          python.__raw = '''
+            function(cb, config)
+              if config.request == 'attach' then
+                local port = (config.connect or config).port
+                local host = (config.connect or config).host or '127.0.0.1'
+                cb({
+                  type = 'server',
+                  port = assert(port, '`connect.port` is required for a godot `attach` configuration'),
+                  host = host,
+                  options = {
+                    source_filetype = 'gdscript',
+                  },
+                })
+              else
+                cb({
+                  type = 'executable',
+                  command = 'godot',
+                  args = { '--path', config.project_path or vim.fn.getcwd(), '--remote-debug', '127.0.0.1:6006' },
+                  options = {
+                    source_filetype = 'gdscript',
+                  },
+                })
+              end
+            end
+          ''';
+
+          # Categorized structured config for executable adapter
+          executables.python = {
+            command = "python";
+            args = [ "-m" "debugpy" ];
+          };
+
+          # Categorized structured config for server adapter
+          servers.netcoredbg = {
+            port = 4711;
+            executable.command = "netcoredbg";
+          };
+        }
+      '';
     };
 
     configurations =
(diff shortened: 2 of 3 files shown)


plugins/by-name/dap/dapHelpers.nix
```diff
@@ -98,7 +98,11 @@ rec {
 
   mkAdapterOption =
     name: type:
-    mkNullOrOption (with types; attrsOf (either str type)) ''
+    let
+      # TODO: Added 2025-12-11 (26.05)
+      strToRawLua = lib.nixvim.deprecation.transitionType types.str lib.nixvim.mkRaw types.rawLua;
```
khaneliman, line 103: I thought we had a helper for this kind of behavior but didn't see it, at first glance.

MattSturgeon: IIRC, we planned to implement something like this directly in the "strLua" types, so that it affected all instances treewide.

The only reason we haven't done that yet is because no one has gotten around to it.

I'm fine with you adding your own deprecation helper here, if you want to have it for this plugin without affecting the others yet. Or better yet, in `lib/deprecation.nix`.

plugins/by-name/dap/default.nix
```diff
@@ -22,10 +22,24 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          servers = dapHelpers.mkAdapterOption "server" dapHelpers.serverAdapterOption;
+          pipes = dapHelpers.mkAdapterOption "pipe" dapHelpers.pipeAdapterOption;
+        };
+      };
+      description = ''
+        Debug Adapter Protocol adapters.
+
+        You can define adapters in two ways:
+        1. Dynamic adapters using raw lua functions: `adapters.godot = { __raw = "function() ... end"; }`
+        2. Categorized structured configs: `adapters.executables.python = { command = "..."; }`
```
MattSturgeon: Would this be better placed in an option `example`? E.g. a commented literalExpression?

plugins/by-name/dap/default.nix
```diff
@@ -22,10 +22,24 @@ lib.nixvim.plugins.mkNeovimPlugin {
   deprecateExtraOptions = true;
 
   extraOptions = {
-    adapters = lib.nixvim.mkCompositeOption "Dap adapters." {
-      executables = dapHelpers.mkAdapterOption "executable" dapHelpers.executableAdapterOption;
-      servers = dapHelpers.mkAdapterOption "server" dapHelpers.serverAdapterOption;
-      pipes = dapHelpers.mkAdapterOption "pipe" dapHelpers.pipeAdapterOption;
+    adapters = lib.nixvim.mkNullOrOption' {
+      type = types.submodule {
+        freeformType = types.attrsOf types.rawLua;
```
MattSturgeon, line 27: I assume we don't need `adapters = mkRaw ""` support?

This freeformType will allow defining non-option attrs as rawLua, but it will not allow defining explicitly declared options as rawLua (unless the option's type supports it) and it will not allow defining rawLua instead of the root submodule.

That may be what you intended, I'm just double checking.

plugins/by-name/dap/default.nix
```diff
@@ -61,11 +75,21 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          "pipes"
+        ]
+      ) (cfg.adapters or { });
+
       options = {
         inherit (cfg) configurations;
 
         adapters =
-          (lib.optionalAttrs (cfg.adapters.executables != null) (
+          manualAdapters
```
MattSturgeon: What should the update order here be? E.g. should manualAdapters be added last to shadow the others?

Or should we be using `mkMerge` instead of `//` and allow the module system to handle conflicts?

plugins/by-name/dap/default.nix
khaneliman: Yeah, I was going back and forth whether to have a quick blurb in the description to make it more obvious, but it might make it easier to understand why you would choose either with a proper example.

plugins/by-name/dap/default.nix
khaneliman: > I assume we don't need adapters = mkRaw "" support?

It's `adapters.<attr> = mkRaw` with that `freeformType` isn't it? 

EDIT: I think I see what you're asking, whether we SHOULD support `rawLua` for the entire option itself... I didn't intend to add that with this. I'm not sure if that works with dap. 

>This freeformType will allow defining non-option attrs as rawLua, but it will not allow defining explicitly declared options as rawLua (unless the option's type supports it) and it will not allow defining rawLua instead of the root submodule.

The explicit options as `rawLua` is where that `transitionType` in the `dapHelpers` comes in play. This is just to allow an adapter to be a function instead of using the structured options only.

plugins/by-name/dap/default.nix
khaneliman: Yeah, that's smarter than a simple spread operator. I'll update it to use `mkMerge` to get a better merge strategy.

plugins/by-name/dap/default.nix
MattSturgeon: > > I assume we don't need adapters = mkRaw "" support?
> 
> EDIT: I think I see what you're asking, whether we SHOULD support `rawLua` for the entire option itself... I didn't intend to add that with this. I'm not sure if that works with dap. 

Yeah, that was what I meant. I wasn't sure whether there is any value in using `maybeRaw submodule` as the option type.

> The explicit options as `rawLua` is where that `transitionType` in the `dapHelpers` comes in play.

Gotcha. I hadn't cross-referenced whether the non-freeform options handled raw types correctly. Sounds like they do.

plugins/by-name/dap/dapHelpers.nix
khaneliman: Want me to leave it here for now and then migrate `mkStrLuaType` to use the transitionType in another PR and remove this internal definition there? 

```diff
diff --git a/lib/types.nix b/lib/types.nix
index 9a5e637b..dd026867 100644
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -10,19 +10,12 @@ let
 
   mkStrLuaType =
     description:
-    lib.mkOptionType {
+    # TODO: Added 2025-12-12 (26.05)
+    # Deprecate string usage in favor of rawLua
+    deprecation.transitionType types.str lib.nixvim.mkRaw types.rawLua
+    // {
       name = "strLua";
       inherit description;
-      descriptionClass = "noun";
-      check = v: lib.isString v || types.rawLua.check v;
-      merge =
-        loc: defs:
-        lib.pipe defs [
-          # Coerce strings to rawLua
-          # TODO: consider deprecating this behaviour
-          (lib.map (def: def // { value = lib.nixvim.mkRaw def.value; }))
-          (lib.options.mergeEqualOption loc)
-        ];
     };
 
   isRawType = v: lib.isString (v.__raw or null);
diff --git a/plugins/by-name/dap/dapHelpers.nix b/plugins/by-name/dap/dapHelpers.nix
index 1f3d7ad6..767c5786 100644
--- a/plugins/by-name/dap/dapHelpers.nix
+++ b/plugins/by-name/dap/dapHelpers.nix
@@ -98,11 +98,7 @@ rec {
 
   mkAdapterOption =
     name: type:
-    let
-      # TODO: Added 2025-12-11 (26.05)
-      strToRawLua = lib.nixvim.deprecation.transitionType types.str lib.nixvim.mkRaw types.rawLua;
-    in
-    mkNullOrOption (with types; attrsOf (either strToRawLua type)) ''
+    mkNullOrOption (with types; attrsOf (either strLua type)) ''
       Debug adapters of `${name}` type.
       The adapters can also be set to a function which takes three arguments:
 
```

MattSturgeon: SGTM

MattSturgeon (approved):
`manualAdapters` can be simplified, either to `removeAttrs` or `intersectAttrs` depending on the intent. You could also consider inlining it into the mkMerge and dropping the named binding.

Otherwise LGTM. Thanks!

khaneliman:
> `manualAdapters` can be simplified, either to `removeAttrs` or `intersectAttrs` depending on the intent. You could also consider inlining it into the mkMerge and dropping the named binding.
> 
> Otherwise LGTM. Thanks!

Inlined it with `removeAttrs`

PR #4047, opened 2025-12-08
plugins/treesitter: update injections to handle functions

khaneliman:
Handle highlighting strings that occur after a function condition. Only tested lua, but vim one is basically a copy paste. 

Before 

image

After 

image

PR #4061, opened 2025-12-12
plugins/treesitter: support let expressions with injections

khaneliman:
Follow up to https://github.com/nix-community/nixvim/pull/4047 with some more coverage.


khaneliman:
> I'm not keen on how repetitive the injections file is getting, but I don't know the (lisp?) language well enough to suggest refactors.

I agree the language feels incredibly repetitive, but from playing around with it and reading the docs I haven't found a nicer way... Maybe we'll learn something to consolidate it and make it more concise and can clean it up later. (Just getting into the tree-sitter queries and docs though since I'm wanting to get more familiar with it and it's easier to do when actually solving a problem I was facing)

PR #4066, opened 2025-12-12
plugins/lsp-format-modifications: init

ein-shved:


plugins/by-name/lsp-format-modifications/default.nix
```diff
@@ -0,0 +1,12 @@
+{
+  lib,
+  ...
+}:
```
GaetanLepage:
```suggestion
{ lib, ... }:
```
nit

plugins/by-name/lsp-format-modifications/default.nix
```diff
@@ -0,0 +1,12 @@
...
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "lsp-format-modifications";
+  package = "lsp-format-modifications-nvim";
+  callSetup = false;
```
GaetanLepage, line 5:
```suggestion

  This plugin has no settings.
  hasSettings = false;
  callSetup = false;
```

ein-shved: done

plugins/by-name/lsp-format-modifications/default.nix
ein-shved: done

PR #4075, opened 2025-12-13
plugins/lsp/language-servers/nixd: use dependencies.nixpkgs-fmt

GaetanLepage:
- **dependencies: add nixpkgs-fmt**
- **plugins/lsp/language-servers/nixd: use dependencies.nixpkgs-fmt**

### DIFF
--- a/plugins/lsp/language-servers/default.nix
+++ b/plugins/lsp/language-servers/default.nix
@@ -1,6 +1,5 @@
 {
   lib,
-  pkgs,
   ...
 }:
 let
@@ -38,9 +37,9 @@ let
     nixd = {
       settings = cfg: { nixd = cfg; };
       extraConfig = cfg: {
-        extraPackages = lib.optional (
+        dependencies.nixpkgs-fmt.enable = lib.mkIf (
           (cfg.settings.formatting.command or null) == [ "nixpkgs-fmt" ]
-        ) pkgs.nixpkgs-fmt;
+        ) (lib.mkDefault true);
       };
     };
     omnisharp = {


MattSturgeon:
I get that the "dependencies" module is supposed to keep things in sync, but honestly I feel like it's lacking the simple/obvious/intuitive API of local `fooPackage` options. I'm hesitant to start using the "dependencies" system for LSP packages too.

GaetanLepage:
> I get that the "dependencies" module is supposed to keep things in sync, but honestly I feel like it's lacking the simple/obvious/intuitive API of local `fooPackage` options. I'm hesitant to start using the "dependencies" system for LSP packages too.

I get what you mean, but in what way plugins are different than language servers?
We made the choice to introduce `dependencies`, so I think that we should now try to be as consistent as possible.

MattSturgeon:
Maybe I'm having second thoughts about `dependencies` being ready to replace all our package options treewide.

For me the main painpoints of `dependencies` are that it obfuscates what is installed and by which options. It is not obvious that `plugins.foo.enable` will install `dependencies.bar`. On the other hand, it is obvious that `plugins.abc.enable` will install `plugins.abc.xyzPackage`. IIRC we planned to address this when introducing the `dependencies` system, but have not done so yet.

Raising the consistency argument begs the question: why do we use `dependencies` for `plugins.<name>.fooPackage` options, but not the main `plugins.<name>.package` option? In practice, the answer is that the main `package` option is far less likely to relate to a package that is installed multiple times, but from a consistency perspective it makes no sense.

I wonder: should we be asking ourselves "does this dependency get installed by multiple places?" If "yes", use `dependencies`, if "no" use a one-off package option.

Or maybe there's a middleground where we return to having local package options, but have their defaults use the centralized `dependencies` system?

The reason I'm raising this now is that our `lsp` refactors feel like a good opportunity to reassess, and using the system outside of `plugins` highlights potential pain points we hadn't considered... Although I think most pain points also apply to our existing use.

khaneliman:
> Maybe I'm having second thoughts about `dependencies` being ready to replace all our package options treewide.
> 
> For me the main painpoints of `dependencies` are that it obfuscates what is installed and by which options. It is not obvious that `plugins.foo.enable` will install `dependencies.bar`. On the other hand, it is obvious that `plugins.abc.enable` will install `plugins.abc.xyzPackage`. IIRC we planned to address this when introducing the `dependencies` system, but have not done so yet.
> 

I was looking at introducing the `dependencies` paradigm in Home Manager recently and this is exactly something that came up as a concern while I was doing a proof of concept, as well. I love the idea of deduplicating package options for the same thing that create a mess of finding which options need to get updated to keep everything using your same preferred package. But, if it's not obvious which module's have the dependency, it's hard to make it a strict replacement. 

> Raising the consistency argument begs the question: why do we use `dependencies` for `plugins.<name>.fooPackage` options, but not the main `plugins.<name>.package` option? In practice, the answer is that the main `package` option is far less likely to relate to a package that is installed multiple times, but from a consistency perspective it makes no sense.
> 

I've always thought of the main package is obviously in the module as a source of truth for a configuration that is generated for that plugin and needs to be added to the filesystem in a specific location. It's not a random binary we need access to on PATH, which is the point of dependencies. 

> I wonder: should we be asking ourselves "does this dependency get installed by multiple places?" If "yes", use `dependencies`, if "no" use a one-off package option.
> 

I don't like creating this tech debt intentionally, we should have a set standard of using it or abandon it. The benefit of throwing these random binaries in `dependencies` is that someone doesn't need to scour the repo to make this decision every time they update a module. 

> Or maybe there's a middleground where we return to having local package options, but have their defaults use the centralized `dependencies` system?
> 

This might be a good approach to raise the visibility and documentation and create another level of flexibility. In Home Manager we have several modules that work like this for `systemd` and shell integrations. You get the centralized preference with a localized override available.

GaetanLepage:
> I've always thought of the main package is obviously in the module as a source of truth for a configuration that is generated for that plugin and needs to be added to the filesystem in a specific location. It's not a random binary we need access to on PATH, which is the point of dependencies.

Same.



> I don't like creating this tech debt intentionally, we should have a set standard of using it or abandon it. The benefit of throwing these random binaries in `dependencies` is that someone doesn't need to scour the repo to make this decision every time they update a module.

I agree.


> This might be a good approach to raise the visibility and documentation and create another level of flexibility. In Home Manager we have several modules that work like this for `systemd` and shell integrations. You get the centralized preference with a localized override available.

Yes why not!

---

@khaneliman what should we do in this case according to you? I am in favor to switching to `dependencies`. @MattSturgeon is not convinced.
Maybe you could help up settle on this decision?

khaneliman:
I'm partial to the concept of including all binaries needed on PATH as `dependencies` but finding a way  to show that a dependency is part of the module's requirements on the docs. Right now it's an invisible relationship and we need a way to represent that so people know it's happening. 

One proposal was above, providing an option for the package in the module's that need the dependency:

I just thought of a bad UX with providing the option itself with defaults in a module, for our case, where you might be a user scouring docs and see the option and update it... but not realize that it is only updating one instance of that dependency and that another module might use the default from `dependencies` resulting in it still being installed or duplicated with another version of that package on PATH. This also means multiple levels of maintaining package options, added complexity, etc etc. One benefit, if we have a package that is going to be applied to a specific scoped usage instead of just being bundled in PATH, it would make more sense as a package option that doesn't affect the global scope. 

Another option: 

Just make the docs properly represent the `dependencies` used with a module so it's more obvious about this relationship. If someone is viewing the `direnv` module in our documentation, they don't know we have a `dependencies = [ "direnv" ];` atm, but we could parse the `mk{Neo}VimPlugin` while evaluating to fetch and represent that information to the user and show that it's an overrideable option. We keep our current trajectory of migrating to `dependencies` but make the UX improved so that the central repository / collection of binaries to add to your closure are obvious and well documented.

PR #4080, opened 2025-12-15
plugins/lsp: drop `customCmd`

MattSturgeon:
All our custom `cmd` overrides have now been upstreamed into nvim-lspconfig (thanks @GaetanLepage). Going forward, we should continue that trend. If any future nix-specific `cmd`s aren't accepted in nvim-lspconfig, they can be patched in Nixpkgs if necessary.

This is a draft until NixOS/nixpkgs#470237 is more available. Currently it is in the latest nixpkgs-unstable (pinned here in nixvim) and nixos-unstable-small, but not yet in nixos-unstable.

### DIFF
--- a/modules/lsp/servers/default.nix
+++ b/modules/lsp/servers/default.nix
@@ -32,7 +32,7 @@ let
   # Create a server option
   # Used below for the `lsp.servers.*` options
   mkServerOption =
-    name: args:
+    { name, ... }@args:
     let
       homepage = lib.pipe options.lsp.servers [
         # Get suboptions of `lsp.servers`
@@ -59,11 +59,6 @@ let
       '';
       default = { };
     };
-
-  # Read package-option defaults from `packages.nix`
-  serverPackages = builtins.mapAttrs (name: package: {
-    inherit name package;
-  }) (import ./packages.nix).packages;
 in
 {
   options.lsp = {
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Oops, this is a duplicate of https://github.com/nix-community/nixvim/pull/4078

It's reassuring that we both implemented effectively identical patches.

I'm biased, but I think I _slightly_ prefer the commit message and comment reword on _this_ one. What do you think @GaetanLepage?

GaetanLepage:
Sure! I think it was fine to split the change into two commits, but I'm perfectly fine with your approach too.

MattSturgeon:
NixOS/nixpkgs#470237 is in nixos-unstable now

modules/lsp/servers/default.nix
```diff
@@ -60,19 +60,15 @@ let
...
     let
-      inherit (import ./packages.nix)
-        packages
-        customCmd
-        ;
+      inherit (import ./packages.nix) packages;
     in
     builtins.mapAttrs (name: v: {
       inherit name;
       package = v.package or v;
```
MattSturgeon: IIUC this can be simplified to:
```suggestion
      package = v;
```

Or rather:
```nix
builtins.mapAttrs (name: package: { inherit name package; }) (import ./packages.nix).packages
```

That's because the `v.package` attr only existed in `customCmd` declarations.

PR #4082, opened 2025-12-15
copilot-chat: update the config to the latest way mentioned in the do…

vimlinuz:
…cs of copilot-chat to generate the required lua config

the previous config that it was generating is like this 
```lua
require("CopilotChat").setup({
    answer_header = "## Ai ",
    auto_follow_cursor = true,
    highlight_selection = false,
    model = "gpt-4o",
    question_header = "## santosh ",
    show_help = false,
    window = { border = "none", height = 1, layout = "float", title = "─", width = 1 },
})
``` 
but the latest settings that we need to put there to make things like header work are mentioned there in the docs in the copilot chat so 
```lua

  headers = {
    user = '👤 You',
    assistant = '🤖 Copilot',
    tool = '🔧 Tool',
  },
``` 
so changed the following line which might fix it

### DIFF
--- a/plugins/by-name/copilot-chat/default.nix
+++ b/plugins/by-name/copilot-chat/default.nix
@@ -39,16 +39,16 @@ lib.nixvim.plugins.mkNeovimPlugin {
     '';
 
     headers = {
-      user_header = defaultNullOpts.mkStr "## User " ''
+      user = defaultNullOpts.mkStr "## User " ''
         Header to use for user questions.
       '';
 
-      assistant_header = defaultNullOpts.mkStr "## Copilot " ''
+      assistant = defaultNullOpts.mkStr "## Copilot " ''
         Header to use for AI answers.
       '';
 
-      error_header = defaultNullOpts.mkStr "## Error " ''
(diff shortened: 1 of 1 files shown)


vimlinuz:
should we have same name like that in the lua code or can it be different.

khaneliman:
These are basically just suggestions and documentation, you'd still have to make the change in your config. We've actually be migrating towards just removing the `settingsOptions` and updating the `settingsExample` to make it more clear and more maintainable

vimlinuz:
which part is to be modified to handle the actual options can I help with it ?

khaneliman:
Those are the actual options, but they default to `null` so by default it doesn't affect a user. A user can configure the plugin however they would like through `settings`. So, you can just update your config to match the new format and it'll accept it and then it'll work. 

I'm just saying that we've recently been removing those `settingsOptions` so that we aren't accidentally limiting users if we have incorrect / outdated types and just let the upstream documentation be the guide for how to configure the plugin. 

`settingsExample` is where you can show the recommended change for people as an example config.

PR #4084, opened 2025-12-15
plugins/treesitter: support new api 

khaneliman:
New treesitter branch will have a completely new API. update to support both legacy and new apis

Should add support for anyone using https://github.com/NixOS/nixpkgs/pull/470883

### DIFF
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -11,7 +11,6 @@ let
     literalExpression
     optionalString
     mkIf
-    mkDefault
     ;
   buildGrammarDeps = [
     "gcc"
@@ -312,6 +311,22 @@ lib.nixvim.plugins.mkNeovimPlugin {
       pkg: pkg.withPlugins (_: cfg.grammarPackages)
     );
 
+    # NOTE: This autoCmd is declared outside of Lua while the autogroup is created in luaConfig.content.
+    # This is fragile - if module generation order changes, the autocmd might be created before the
+    # autogroup exists (causing failure), or the autogroup's clear=true might clear this autocmd.
+    # The current order happens to work, but changes to nixvim's module system could break this.
+    autoCmd = lib.optional cfg.folding.enable {
+      event = "FileType";
+      group = "nixvim_treesitter";
+      pattern = "*";
+      callback.__raw = ''
+        function()
+          vim.wo[0][0].foldexpr = 'v:lua.vim.treesitter.foldexpr()'
+          vim.wo[0][0].foldmethod = 'expr'
+        end
+      '';
+    };
+
     warnings = lib.nixvim.mkWarnings "plugins.treesitter" (
       lib.map (packageName: {
         when = !cfg.nixGrammars && !config.dependencies.${packageName}.enable;
@@ -321,10 +336,5 @@ lib.nixvim.plugins.mkNeovimPlugin {
         '';
       }) buildGrammarDeps
     );
-
-    opts = mkIf cfg.folding.enable {
-      foldmethod = mkDefault "expr";
-      foldexpr = mkDefault "v:lua.vim.treesitter.foldexpr()";
-    };
   };
 }


GaetanLepage (approved):
I went through the diff and it seems great for starting the initial compat-implem. Good job!

plugins/by-name/treesitter/default.nix
```diff
@@ -189,136 +148,17 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-    };
+    install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
   };
 
   extraOptions = {
     folding = lib.mkEnableOption "tree-sitter based folding";
 
+    highlight = lib.mkEnableOption "tree-sitter based syntax highlighting";
+
+    indent = lib.mkEnableOption "tree-sitter based indentation";
```
HeitorAugustoLN: I would make them into folding.enable, highlight.enable and indent.enable. So it easier to add extra options in the future, in case its needed, for example, right now we could add an filetypes option for the pattern or something like that

plugins/by-name/treesitter/default.nix
```diff
@@ -189,136 +148,17 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-        node_decremental = "grm";
-        node_incremental = "grn";
-        scope_incremental = "grc";
-      };
-    };
-
-    indent = {
-      enable = true;
-    };
+    install_dir.__raw = "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
```
HeitorAugustoLN:
```suggestion
    install_dir = lib.nixvim.nestedLiteralLua "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
```

plugins/by-name/treesitter/default.nix
khaneliman: I did think about that and this might be the perfect time for that transition.

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +231,101 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+            `${packageName}` is required to build grammars as you are not using `${opt.nixGrammars}`.
+            You may want to set `${options.dependencies.${packageName}.enable}` to `true`.
+          '';
+        }) buildGrammarDeps
+      );
+
+      opts = mkIf cfg.folding.enable {
+        foldmethod = mkDefault "expr";
+        foldexpr = mkDefault "v:lua.vim.treesitter.foldexpr()";
+      };
```
HeitorAugustoLN: I would also migrate this to an auto command, as the README suggests

HeitorAugustoLN (approved):
LGTM

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,42 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      enable = true;
+    install_dir = lib.nixvim.nestedLiteralLua "vim.fs.joinpath(vim.fn.stdpath('data'), 'site')";
+  };
 
-      additional_vim_regex_highlighting = true;
-      disable = [ "rust" ];
-      custom_captures = { };
+  extraOptions = {
+    folding = mkOption {
+      type = types.either types.bool (
```
MattSturgeon: Can we use `transitionType` here?

That'd a) handle the deprecation for us and b) avoid documenting the `either` type.

We can also use `visible = "transparent"`, so that only the sub-options (`enable`) are documented. Although, I'm not sure nixvim's docs know how to handle `visible = "transparent"`...

khaneliman: The message was incredibly vague that's why I did this. Does the `visible` thing actually output the new submodule options? Because, it was basically along the lines of "changed from string to submodule" which isn't helpful. 

```bash
evaluation warning: Passing a boolean for `plugins.treesitter.folding' is deprecated, use submodule instead. Definitions:
                    - In `/nix/store/jfgsz3lpkq3m3c3zxffdnxajbn0gqanp-source/modules/nixvim/plugins/treesitter': 
```

khaneliman: >Although, I'm not sure nixvim's docs know how to handle visible = "transparent"...

Doesn't look like it 
```bash
serve-docs
=> Building Nixvim documentation...

error:
       … while evaluating the attribute 'value'
         at «github:NixOS/nixpkgs/09b8fda8959d761445f12b55f380d90375a1d6bb?narHash=sha256-aq%2BdQoaPONOSjtFIBnAXseDm9TUhIbe215TPmkfMYww%3D»/lib/modules.nix:1118:7:
         1117|     // {
         1118|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1119|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `flake.apps.x86_64-linux.docs.program':

       … while evaluating the attribute 'mergedValue'
         at «github:NixOS/nixpkgs/09b8fda8959d761445f12b55f380d90375a1d6bb?narHash=sha256-aq%2BdQoaPONOSjtFIBnAXseDm9TUhIbe215TPmkfMYww%3D»/lib/modules.nix:1192:5:
         1191|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1192|     mergedValue =
             |     ^
         1193|       if isDefined then

       … while evaluating definitions from `/nix/store/pqjppkj5kcs4001j5svlsz1595hcja0y-source/modules/transposition.nix':

       … while evaluating the option `perSystem.x86_64-linux.apps.docs.program':

       error: expected a Boolean but found a string: "transparent"
       at /home/khaneliman/Documents/github/nixvim/docs/mdbook/default.nix:67:18:
           66|         in
           67|         visible' && !internal;
             |                  ^
           68|     in
```

MattSturgeon: > Does the `visible` thing actually output the new submodule options?

Yes, when using NixOS docs tooling (`lib.options.optionAttrSetToDocsList`). But Nixvim re-implements that (tech debt), so we'd need to add support ourselves.

> The message was incredibly vague that's why I did this. […] Because, it was basically along the lines of "changed from string to submodule" which isn't helpful. 
> 
> ```bash
> evaluation warning: Passing a boolean for `plugins.treesitter.folding' is deprecated, use submodule instead. Definitions:
>                     - In `/nix/store/jfgsz3lpkq3m3c3zxffdnxajbn0gqanp-source/modules/nixvim/plugins/treesitter': 
> ```

Yep, that's definitely too vague.

You can keep your current solution, but we need to use the `compatType // { inherit (newType) description; }` pattern.

I'd also _prefer_ we use option-string-interpolation, so the warnings include the full path, complete with prefix.

Ideally, we should extend `transitionType` so we can have more control over its message.

khaneliman: > You can keep your current solution,

Do we have access to `options` in the `apply` function or how can we handle: 

>I'd also prefer we use option-string-interpolation, so the warnings include the full path, complete with prefix.

MattSturgeon: > Do we have access to `options` in the `apply` function or how can we handle:

Need to access it via the `options` module arg at the top of the file.

khaneliman: > evaluation warning: Passing a boolean to `plugins.treesitter.folding` in `/nix/store/jfgsz3lpkq3m3c3zxffdnxajbn0gqanp-source/modules/nixvim/plugins/treesitter' is deprecated, use `plugins.treesitter.folding.enable` instead.

This work?

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,48 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        default = {
+          enable = false;
+        };
+        description = "Tree-sitter based folding configuration.";
+        apply =
+          x:
+          if builtins.isBool x then
+            lib.warn
+              "Passing a boolean to `${lib.showOption options.plugins.treesitter.folding.loc}` is deprecated, use `${
+                lib.showOption (options.plugins.treesitter.folding.loc ++ [ "enable" ])
```
MattSturgeon: This can be:


```suggestion
              "Passing a boolean to `${options.plugins.treesitter.folding}` is deprecated, use `${options.plugins.treesitter.folding}.enable
```

As options have a `__toString` attr used in string interpolation.

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,48 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          };
+        };
+      in
+      mkOption {
+        type = (types.either types.bool foldingSubmodule) // {
+          inherit (foldingSubmodule) description;
+        };
+        default = {
+          enable = false;
+        };
```
MattSturgeon: Let's drop this default; if we get the sub-options properly documented then it isn't needed.

Even if it was needed, it'd be better as `defaultText`, because this is raising `enable`'s highestPrio from `mkOptionDefault` to `defaultOverridePriority`.

Instead, once https://github.com/nix-community/nixvim/pull/4092 is merged let's make the submodule option itself "transparent", so it is not itself documented:


```suggestion
        visible = "transparent";
        default = { };
```

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,48 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      let
+        foldingSubmodule = types.submodule {
+          options = {
+            enable = lib.mkEnableOption "tree-sitter based folding";
+          };
+        };
+      in
+      mkOption {
+        type = (types.either types.bool foldingSubmodule) // {
+          inherit (foldingSubmodule) description;
```
MattSturgeon: `either` doesn't implement `getSubOptions`, so if we want to document the `enable` option we need:
```suggestion
          inherit (foldingSubmodule) description getSubOptions;
```

plugins/by-name/treesitter/default.nix
MattSturgeon: > This work?

I'm fine with either `showFiles` (as above) or `showDefs` (as currently implemented). Your preference. I wonder if shifting the files list to the end makes the warning more readable, though?

khaneliman:
@MattSturgeon Alright, just got back and applied those changes. Looks like warning and documentation looks right.

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      autoCmd = lib.optional cfg.folding.enable {
+        event = "FileType";
+        group = "nixvim_treesitter";
+        pattern = "*";
+        callback.__raw = ''
+          function()
+            vim.wo[0][0].foldexpr = 'v:lua.vim.treesitter.foldexpr()'
+            vim.wo[0][0].foldmethod = 'expr'
+          end
         '';
```
HeitorAugustoLN: Just noticed this, why manually create auto commands for highlight and indent but not for folding?

khaneliman: We have runtime detection for master v main. Master handles that, main we need a custom autocmd. Folding has always been outside of nvim-treesitter

HeitorAugustoLN: makes sense

plugins/by-name/treesitter/default.nix
```diff
@@ -21,23 +20,42 @@ let
...
   description = ''
-    Provides an interface to [tree-sitter]
+    Provides an interface to [tree-sitter] for Neovim.
 
     > [!NOTE]
-    > This plugin defaults to all functionality disabled.
-    >
-    > Please explicitly enable the features you would like to use in `plugins.treesitter.settings`.
-    > For example, to enable syntax highlighting use the `plugins.treesitter.settings.highlight.enable` option.
+    > This module is designed for nvim-treesitter **main branch** (current standard).
```
MattSturgeon: Is this correct (merge ready)? I.e. is `main` branch the current standard that we get through nixpkgs? I assumed that was still "coming soon".

I guess this is fine as it avoids us coming back to update the wording when the statement _becomes_ true, and maybe we can argue that _outside of_ nixpkgs the `main` branch is the current standard.

Maybe this note should also more explicitly mention the `master` branch that the module _isn't_ designed for?

plugins/by-name/treesitter/default.nix
```diff
@@ -113,71 +116,26 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-        ''';
-      })
-    )
-    ```
-
-    Verify if the queries were picked up by running `:TSModuleInfo`.
+    > [!NOTE]
+    > This module is designed for nvim-treesitter **main branch** (current standard).
+    > For the legacy master branch, the module automatically detects it at runtime and
+    > passes your `settings` to `require('nvim-treesitter.configs').setup()`.
```
MattSturgeon: Oh, I see the note is kinda duplicated. This second one has the "mention `master`" idea I suggested above.

These two notes should probably be combined.

Also, I think the wording can be clarified. I think this is a step in the right direction, but still rather clunky:

> For backwards compatibility, the module will check for nvim-treesitter's legacy `master` branch at runtime.
> If found, the module passes your `settings` to `require('nvim-treesitter.configs').setup()`.

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,45 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-    sync_install = false;
+  extraOptions = {
+    folding =
+      let
+        foldingSubmodule = types.submodule {
+          options = {
+            enable = lib.mkEnableOption "tree-sitter based folding";
+          };
+        };
+      in
```
MattSturgeon: Nit: moving this let-block to the `type =` binding will allow the rest of the option declaration to be unindented (i.e. `folding = mkOption {`).

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,45 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          inherit (foldingSubmodule) description getSubOptions;
+        };
+        visible = "transparent";
+        default = { };
+        description = "Tree-sitter based folding configuration.";
+        apply =
+          x:
+          if builtins.isBool x then
+            lib.warn
+              "Passing a boolean to `${options.plugins.treesitter.folding}` is deprecated, use `${options.plugins.treesitter.folding}.enable. ${lib.options.showDefs options.plugins.treesitter.folding.definitionsWithLocations}"
```
MattSturgeon: Nit, missing closing quote:


```suggestion
              "Passing a boolean to `${options.plugins.treesitter.folding}` is deprecated, use `${options.plugins.treesitter.folding}.enable`. ${lib.options.showDefs options.plugins.treesitter.folding.definitionsWithLocations}"
```

Q: Maybe we should use nixpkgs-style `` ` ``-opening and `'`-closing quotes, to be consistent with other deprecation warnings?


```suggestion
              "Passing a boolean to `${options.plugins.treesitter.folding}' is deprecated, use `${options.plugins.treesitter.folding}.enable'. ${lib.options.showDefs options.plugins.treesitter.folding.definitionsWithLocations}"
```

Nit: commonly, showDefs is prefixed with `"Definitions:"`:

```suggestion
              "Passing a boolean to `${options.plugins.treesitter.folding}' is deprecated, use `${options.plugins.treesitter.folding}.enable'. Definitions:${lib.options.showDefs options.plugins.treesitter.folding.definitionsWithLocations}"
```

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,45 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        };
+      in
+      mkOption {
+        type = (types.either types.bool foldingSubmodule) // {
+          inherit (foldingSubmodule) description getSubOptions;
+        };
+        visible = "transparent";
+        default = { };
+        description = "Tree-sitter based folding configuration.";
+        apply =
```
MattSturgeon: I just realised, one disadvantage of using `apply` instead of `coercedTo` or `transitionType` is that the check happens _after_ definition-merging.

Therefore, if there are multiple definitions, some using the old type and others using the new type, option merging will fail and our deprecation notice will never print. Instead the user will see an incompatible types error, when `either` fails to merge `true` with `{ enable = true; }`.

If we handed the coersion during merging, this example would be handled fine as merging equal definitions.

Not a big deal, just something I realised and wanted to comment on.

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,45 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          x:
+          if builtins.isBool x then
+            lib.warn
```
MattSturgeon: This should have a "added" date, and a suggested "removal" release.

plugins/by-name/treesitter/default.nix
```diff
@@ -189,135 +147,45 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-
-            Using this option may slow down your editor, and you may see some duplicate highlights. \
-            Instead of true, it can also be a list of languages.
-          '';
-
-      enable = lib.nixvim.defaultNullOpts.mkBool false ''
-        Whether to enable treesitter highlighting.
-      '';
-
-      disable = lib.nixvim.defaultNullOpts.mkStrLuaFnOr (with lib.types; listOf (maybeRaw str)) null ''
```
MattSturgeon, line 214: These strLua types are harder to deprecate, because they permitted defining strings and coerced them into rawLua.

Just dropping the settings-suboption is a breaking change because the coercion is also dropped.

One way we could approach this is by checking `lib.isString (cfg.settings.highlight.disable or null)` and handling that by coercing the _actual_ value passed to `toLuaObject`. We could also warn in this scenario.

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+
+        if has_configs_module then
+          -- Master branch: Use the traditional configs API
+          require('nvim-treesitter.configs').setup(${lib.nixvim.toLuaObject cfg.settings})
+        else
+          -- Main branch: Pass install_dir to setup() and enable features via autocommands
+      ''
+      + (optionalString (mainBranchSettings != { }) ''
+        require'nvim-treesitter'.setup(${lib.nixvim.toLuaObject mainBranchSettings})
+      '')
```
MattSturgeon: This will render with weird formatting in `init.lua`. Something like:

```lua
if has_configs_module then
  -- Master branch: Use the traditional configs API
  require('nvim-treesitter.configs').setup({cfg.settings})
else
  -- Main branch: Pass install_dir to setup() and enable features via autocommands

require'nvim-treesitter'.setup({mainBranchSettings})

```

That's because a) `''` strings don's strip trailing newlines and b) the indentation is off. Although, the indentation is probably a non-issue, since we use stylua.

I'm also concerned that this flat structure of optionalString chains makes it hard to read the `if then else end`; specifically the `end` is kinda hidden on L298.

I think we want something like:
```nix
plugins.treesitter.luaConfig.content = ''
  -- Detect which branch of nvim-treesitter is installed
  local has_configs_module = pcall(require, 'nvim-treesitter.configs')

  if has_configs_module then
    -- Master branch: Use the traditional configs API
    require('nvim-treesitter.configs').setup(${lib.nixvim.toLuaObject cfg.settings})
  else
    -- Main branch: Pass install_dir to setup() and enable features via autocommands
    ${optionalString (mainBranchSettings != { }) ''
      require'nvim-treesitter'.setup(${lib.nixvim.toLuaObject mainBranchSettings})
    ''}

    ${optionalString highlightEnabled ''
      vim.api.nvim_create_autocmd('FileType', {
        pattern = '*',
        callback = function()
  	      pcall(vim.treesitter.start)
        end,
      })
    ''}
    ${optionalString indentEnabled ''
      vim.api.nvim_create_autocmd('FileType', {
        pattern = '*',
        callback = function()
  	      vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
        end,
      })
    ''}
  end
''
```

I think this makes the nix side more readable.

The lua side will still have weird indentation, since using `${}` to embed multiline strings will only have correct indentation on the initial line,[^1] but hopefully stylua fixes that.

[^1]: without hacks like `replaceString "\n" "\n  "`

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          end,
+        })
+      '')
+      + (optionalString indentEnabled ''
+        vim.api.nvim_create_autocmd('FileType', {
+          pattern = '*',
+          callback = function()
+            vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
+          end,
+        })
```
MattSturgeon: These autocmds are targeting the same event/pattern/etc. We could probably combine them?

```nix
optionalString (highlightEnabled || indentEnabled) ''
  -- Enable features via autocommands for main branch
    vim.api.nvim_create_autocmd('FileType', {
    pattern = '*',
    callback = function()
      ${optionalString highlightEnabled ''
        pcall(vim.treesitter.start)
      ''}
      ${optionalString indentEnabled ''
        vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
      ''}
    end,
  })
''
```

MattSturgeon: Also: I think best practice is to associate the autocmd with an autocmd group? This allows runtime introspection and disablement, IIUC.

For indent, you use `nixvim_treesitter` below. IDK if that should be shared for these autocmds, or if we should have separate groups like `nixvim_treesitter_indent`?

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
 
           for parser_name, ft in pairs(__parserFiletypeMappings) do
-            require('vim.treesitter.language').register(parser_name, ft)
+            vim.treesitter.language.register(parser_name, ft)
           end
         end
       '');
 
-    extraFiles = mkIf cfg.nixvimInjections { "queries/nix/injections.scm".source = ./injections.scm; };
+      extraFiles = mkIf cfg.nixvimInjections { "queries/nix/injections.scm".source = ./injections.scm; };
```
MattSturgeon: It pains me to see all of `extraConfig` indented due to adding a `let in` block. Would it make sense to move that let-block to the `plugins.treesitter.luaConfig.content =` binding where it is used?

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        end
       ''
       + (optionalString (cfg.languageRegister != { }) ''
+
         do
           local __parserFiletypeMappings = ${lib.nixvim.toLuaObject cfg.languageRegister}
 
           for parser_name, ft in pairs(__parserFiletypeMappings) do
-            require('vim.treesitter.language').register(parser_name, ft)
+            vim.treesitter.language.register(parser_name, ft)
```
MattSturgeon: Should this have legacy compat?

```lua
local language = if has_configs_module then require('vim.treesitter.language') else vim.treesitter.language end
```

Or is this already equivalent?

plugins/by-name/treesitter/default.nix
```diff
@@ -365,46 +233,110 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+      autoGroups.nixvim_treesitter.clear = true;
+
+      autoCmd = lib.optional cfg.folding.enable {
+        event = "FileType";
+        group = "nixvim_treesitter";
+        pattern = "*";
+        callback.__raw = ''
+          function()
+            vim.wo[0][0].foldexpr = 'v:lua.vim.treesitter.foldexpr()'
+            vim.wo[0][0].foldmethod = 'expr'
```
MattSturgeon: Do we want `wo` or `bo` here? Or maybe even `opt_local`, which IIUC is the more lua-like API?

IIUC `vim.bo` and `vim.opt_local` are both like `:setlocal` while `vim.wo` says:

> Like `:setlocal` if setting a `|global-local|` option or if `{bufnr}` is provided, like `:set` otherwise.

plugins/by-name/treesitter/default.nix
khaneliman: Yeah, I kept reorganizing the description and trying to figure out how to be represent the two concepts. I settled on the concept that "main is the first class citizen and master is just transitionally supported but not the focus". I don't want to keep the module in a state of supporting master branch for too long if upstream has already dropped support for it.

plugins/by-name/treesitter/default.nix
khaneliman: It's ready as soon as we say it's good. It works for existing configs and supports the new API for anyone who is following main already / wants to. I don't think we need to carefully time the documentation when it's already cutover upstream and we have a package option that people can use to follow main already.

plugins/by-name/treesitter/default.nix
khaneliman: Updated 
image

plugins/by-name/treesitter/default.nix
khaneliman: This is what upstream documents, so i'll keep it this way.

plugins/by-name/treesitter/default.nix
khaneliman: > Also: I think best practice is to associate the autocmd with an autocmd group? This allows runtime introspection and disablement, IIUC.

Did that but our ordering of the generated init.lua for augroups means they aren't available for the module code and I'd have to do a custom augroup in this module. If we care that much about the grouping, we can shift from using our nixvim module to doing it in lua.

plugins/by-name/treesitter/default.nix
khaneliman: It's correct now and documented upstream in master, as well, i'm honestly not sure if this would have worked before.

plugins/by-name/treesitter/default.nix
MattSturgeon: Up to you how important they are. I'm definitely not an autocmd/autogroup expert. IIRC, there's a shorthand for declaring and using an autogroup in lua (that we don't have access to when using nixvim's autogroups option).

plugins/by-name/treesitter/default.nix
MattSturgeon: Nice. The nix-side formatting still seems a bit off:

```diff
diff --git a/plugins/by-name/treesitter/default.nix b/plugins/by-name/treesitter/default.nix
index 3bd7ca91..757f055b 100644
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -260,33 +260,32 @@ lib.nixvim.plugins.mkNeovimPlugin {
       # Should be removed after transition period
       # Runtime API detection and conditional setup
       plugins.treesitter.luaConfig.content = ''
-        -- Detect nvim-treesitter API 
+        -- Detect nvim-treesitter API
         local has_configs_module = pcall(require, 'nvim-treesitter.configs')
 
         if has_configs_module then
           require('nvim-treesitter.configs').setup(${lib.nixvim.toLuaObject cfg.settings})
         else
-        ${optionalString (mainBranchSettings != { }) ''
-          require'nvim-treesitter'.setup(${lib.nixvim.toLuaObject mainBranchSettings})
-        ''}
-
-        -- Enable features via autocommands for modern nvim-treesitter
-        ${optionalString highlightEnabled ''
-          vim.api.nvim_create_autocmd('FileType', {
-            pattern = '*',
-            callback = function()
-              pcall(vim.treesitter.start)
-            end,
-          })
-        ''}${optionalString indentEnabled ''
-          vim.api.nvim_create_autocmd('FileType', {
-            pattern = '*',
-            callback = function()
-              vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
-            end,
-          })
-        ''}
-                end
+          ${optionalString (mainBranchSettings != { }) ''
+            require'nvim-treesitter'.setup(${lib.nixvim.toLuaObject mainBranchSettings})
+          ''}
+          -- Enable features via autocommands for modern nvim-treesitter
+          ${optionalString highlightEnabled ''
+            vim.api.nvim_create_autocmd('FileType', {
+              pattern = '*',
+              callback = function()
+                pcall(vim.treesitter.start)
+              end,
+            })
+          ''}${optionalString indentEnabled ''
+            vim.api.nvim_create_autocmd('FileType', {
+              pattern = '*',
+              callback = function()
+                vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
+              end,
+            })
+          ''}
+        end
         ${optionalString (cfg.languageRegister != { }) ''
 
           do
```

plugins/by-name/treesitter/default.nix
khaneliman: What were you thinking about warning about? We're going to be warning eventually about these settings being used, at all. I was doing a silent transition to main, initially, since we have runtime logic detection and haven't bumped the package in our repo.

MattSturgeon: If someone defines `settings.highlight.disable = "function() print('hello') end"`, that currently works and gets coerced into `settings.highlight.disable.__raw = "function() print('hello') end"`.

By dropping the option, we drop that coercion.

If we still support the legacy `master`-branch settings, then that's a breaking change.

You're right, that there's another (bigger) breaking change upcoming, when the package is updated to the `main` branch and these meetings _all_ become invalid.

Idk if/how you plan on dealing with this or that incompatibility, I'm just highlighting that removing this option is a breaking change (unlike removing the other settings options).

khaneliman: Yeah, I wasn't asking about the breaking change part just the mention about

> We could also warn in this scenario.

Didn't know if you meant the coercion or the setting going away

MattSturgeon: Sorry. Yeah, I meant to warn about the coercion.

Warning about old/legacy settings being defined would be a more general issue, and would only be relevant once we know at built-time that we're using the `main` branch.

But on reflection, I think silent coercion is more appropriate here; seeing as the setting will be dropped soon anyway.

HeitorAugustoLN:
I noticed the current autocommand broke my highlighting

khaneliman:
> I noticed the current autocommand broke my highlighting

I've been using this branch for a few days now. Did the recent auto group thing break it?

MattSturgeon:
> I noticed the current autocommand broke my highlighting

I suspect `clear` is the issue, because you create the autoGroup twice (once manually, and once via Nixvim's autoGroup option).

We either need to know which will run first and make that the one that `clear`s, _or_ use distinct groups.

What are the pros/cons of having one `nixvim_treesitter` group vs two or three?

In lua, we can also inline group creation:
```lua
    vim.api.nvim_create_autocmd('FileType', {
      group = vim.api.nvim_create_augroup('nixvim_treesitter_indent', { clear = true }),
      -- ...
    })
```

HeitorAugustoLN:
> I've been using this branch for a few days now. Did the recent auto group thing break it?

Yeah, I also suspect is clear = true

plugins/by-name/treesitter/default.nix
MattSturgeon: Something like:

```nix
legacySettings =
  cfg.settings
  # Before it was dropped, the `highlight.disable` option did str→raw coercion
  // lib.optionalAttrs (lib.isString (cfg.settings.highlight.disable or null)) {
    highlight = cfg.settings.highlight // {
      disable.__raw = cfg.settings.highlight.disable;
    };
  };
```

khaneliman:
That wasn't intentional I meant to remove the module based one.

HeitorAugustoLN:
I think we can have clear = true, with one augroup, if folding, highlight and indent are set together.

MattSturgeon:
> I think we can have clear = true, with one augroup, if folding, highlight and indent are set together.

IIUC, the `clear` is done by `nvim_create_augroup` not `nvim_create_autocmd`, so you can still reuse the autoGroup even if it has `clear = true`, you just can't recreate it?

(I think)

HeitorAugustoLN:
I just noticed we are creating the same augroup two times, that is what is making it not work

khaneliman:
Yes, it was related a requested change above wanting them all to have a group. I meant to replace the module with the inline lua. It's just leftover code from a refactor.

plugins/by-name/treesitter/default.nix
```diff
@@ -366,24 +232,80 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-      '')
-      + ''
-        require('nvim-treesitter.configs').setup(${lib.nixvim.toLuaObject cfg.settings})
+      let
+        # TODO: Added 2025-12-18 Check both legacy and new api options
+        # Add warning after transition period
+        # Map parser_install_dir to install_dir for main branch compatibility
+        # The main branch uses install_dir, while master branch uses parser_install_dir
+        mainBranchSettings =
+          lib.optionalAttrs ((cfg.settings.parser_install_dir or null) != null) {
```
MattSturgeon: Do we actually want to explicitly skip `null`, or is the intent:

```suggestion
          lib.optionalAttrs (cfg.settings ? parser_install_dir) {
```

...matching the `? install_dir` check below?

plugins/by-name/treesitter/default.nix
```diff
@@ -366,24 +232,80 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        # Add warning after transition period
+        # Top-level options (cfg.highlight.enable, cfg.indent.enable) are for main branch
+        # For master branch compatibility, users should use settings.highlight.enable directly
+        highlightEnabled = cfg.highlight.enable || (cfg.settings.highlight.enable or false);
+        indentEnabled = cfg.indent.enable || (cfg.settings.indent.enable or false);
+
+        # TODO: Added 2025-12-18 Coerce string disable functions to rawLua
+        # Before it was dropped, the `highlight.disable` option did str→raw coercion
+        # Remove after transition period
+        coercedSettings =
```
MattSturgeon: How about
```suggestion
        masterBranchSettings =
```
```suggestion
        legacySettings =
```
```suggestion
        legacyBranchSettings =
```
?

plugins/by-name/treesitter/default.nix
```diff
@@ -366,24 +232,80 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+        # Add warning after transition period
+        # Map parser_install_dir to install_dir for main branch compatibility
+        # The main branch uses install_dir, while master branch uses parser_install_dir
+        mainBranchSettings =
+          lib.optionalAttrs ((cfg.settings.parser_install_dir or null) != null) {
+            install_dir = cfg.settings.parser_install_dir;
+          }
+          // lib.optionalAttrs (cfg.settings ? install_dir) {
+            inherit (cfg.settings) install_dir;
+          };
```
MattSturgeon: So currently, the only setting passed to `require'nvim-treesitter'.setup` is `install_dir`.

It feels odd that other "freeform" settings are silently discarded, but I get why we're doing that due to the transition period.

I assume, at some point `settings` will return to being a "proper" freeform option and all its content will be passed to `setup`. However, at that point that change is a breaking change, because users may still have legacy/invalid settings.

What's the plan here? Should we have a FIXME or a NOTE highlighting this future issue?

plugins/by-name/treesitter/default.nix
```diff
@@ -366,24 +232,80 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              disable.__raw = cfg.settings.highlight.disable;
+            };
+          };
+      in
       ''
-      + (optionalString (cfg.languageRegister != { }) ''
-        do
-          local __parserFiletypeMappings = ${lib.nixvim.toLuaObject cfg.languageRegister}
+        -- Create autogroup for treesitter autocmds
+        local augroup = vim.api.nvim_create_augroup('nixvim_treesitter', { clear = true })
```
MattSturgeon, line 269: It feels a little fragile that this assumes `plugins.treesitter.luaContent` will be _before_ `autoGroups`'s lua.

Maybe we should have a comment NOTEing that?

khaneliman: I'm not sure I understand, this is the autoGroup lua. It's within treesitter luaContent. We put all plugin configs in `extraConfigLua` and the `autoCmd` stuff in `extraConfigLuaPost`. 

The current approach inlines everything into `plugins.treesitter.luaContent` except the one that will be added at bottom of file through the `autoCmd` option. 

I suppose I could see where we might change `autoCmd` module to write to `extraConfigLua` or `extraConfigLuaPre` and that messes with the order in the init.lua and breaks this assumption.

plugins/by-name/treesitter/default.nix
khaneliman: > It feels odd that other "freeform" settings are silently discarded, but I get why we're doing that due to the transition period.

I can include all freeform settings and just make sure we are migrating that value where it needs to be changed. Everything else will just get ignored by upstream, too but at least we don't have to think about maintenance of the possible changing upstream configuration.

plugins/by-name/treesitter/default.nix
MattSturgeon: The reason I say it is fragile is that we _also_ reference this autoGroup from the `autoCmd` option. If Nixvim gets an unrelated refactor in the future, assumptions about code order can break. I.e. the `autoCmd` option's lua may end up before the plugin's lua, causing it to be referencing an autogroup that doesn't exist yet.

khaneliman: > The reason I say it is fragile is that we _also_ reference this autoGroup from the `autoCmd` option. If Nixvim gets an unrelated refactor in the future, assumptions about code order can break. I.e. the `autoCmd` option's lua may end up before the plugin's lua, causing it to be referencing an autogroup that doesn't exist yet.

lol I was in the middle of updating my comment with the same thing :P Just felt like a concern in which that refactor would need to address / evaluate treewide effects of such a major change.

plugins/by-name/treesitter/default.nix
MattSturgeon: If upstream will ignore legacy settings, or implements its own warnings, then including freeform settings feels best. SGTM.

plugins/by-name/treesitter/default.nix
MattSturgeon: > Just felt like a concern in which that refactor would need to address / evaluate treewide effects of such a major change.

Maybe. But I'd rather avoid and/or highlight such fragile implementation details in advance, while we remember they exist, so that they become easier to spot later.

We've also done these kinda treewide changes _by accident_ before, without initially noticing that they broke things. E.g. when we introduced `plugins/by-name` and inadvertently changed the import order of almost all modules...

khaneliman:
alright, updated with suggested changes and added that additional note

plugins/by-name/treesitter/default.nix
```diff
@@ -366,24 +232,77 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          };
+
+        # TODO: Added 2025-12-18 Configure install_dir for main branch
+        # Map parser_install_dir to install_dir if install_dir not already set
+        # Remove after transition period
+        mainBranchSettings =
+          legacySettings
+          // lib.optionalAttrs (!(legacySettings ? install_dir) && legacySettings ? parser_install_dir) {
+            install_dir = legacySettings.parser_install_dir;
+          };
```
MattSturgeon: By flipping the `//` update order, we can simplify the condition:


```suggestion
        mainBranchSettings =
          lib.optionalAttrs (legacySettings ? parser_install_dir) {
            install_dir = legacySettings.parser_install_dir;
          }
          // legacySettings;
```

I assume we don't need to "clean up" the rename?


```nix
lib.optionalAttrs (legacySettings ? parser_install_dir) {
  install_dir = legacySettings.parser_install_dir;
}
// lib.removeAttrs legacySettings [ "parser_install_dir" ];
```

MattSturgeon (approved):

plugins/by-name/treesitter/default.nix
khaneliman: Nah, similar thought to the "just forward existing table". Plugin will just access the known properties and ignore anything else

PR #4089, opened 2025-12-16
modules/dependencies: fix enable option description

saygo-png:
Small documentation fix. Removing example since i think its pointless for a boolean

image
image


MattSturgeon:
> Removing example since i think its pointless for a boolean

I think we had it for consistency with `mkEnableOption`, which uses `{ default = false; example = true; }`.

EDIT: Oh, of course. We _were_ using `mkEnableOption`...

Note: a common pattern is `mkEnableOption null // { description = "Actual description."; }`

PR #4092, opened 2025-12-17
docs/mdbook: handle `visible = "transparent"`

MattSturgeon:
Adds parity with https://github.com/NixOS/nixpkgs/pull/441782 in our hacky manual docs logic.

See https://github.com/nix-community/nixvim/pull/4084#discussion_r2627667774

There are probably some edge-cases this doesn't cover. I don't really want to invest too much time finding them when the _right_ solution is to rip out our custom stuff and just use `lib.options.optionAttrSetToDocList` as intended.

One behaviour change is that `internal = true` is no longer considered when recursing into sub-options. This is intentional and matches how `lib.options.optionAttrSetToDocList` is implemented.

### DIFF
--- a/docs/mdbook/default.nix
+++ b/docs/mdbook/default.nix
@@ -66,6 +66,10 @@ let
           visible' = if lib.isBool visible then visible else visible != "transparent";
         in
         visible' && !internal;
+
+      # FIXME: isVisible is not a perfect check;
+      # it will false-positive on `visible = "transparent"`
+      hasVisible = opts: lib.any (v: lib.isAttrs v -> isVisible v) (lib.attrValues opts);
     in
     opts:
     if lib.isOption opts then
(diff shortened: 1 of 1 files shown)


khaneliman:
How is this supposed to be used? Adding to the example option, compiles the docs now, but it just hides the `folding` option entirely when i set it at the `folding = mkOption` level, if i put it in the `submodule` it doesn't do anything.

MattSturgeon:
> How is this supposed to be used? Adding to the example option, compiles the docs now, but it just hides the `folding` option entirely when i set it at the `folding = mkOption` level, if i put it in the `submodule` it doesn't do anything.

It is supposed to hide the option, but not its sub-options.

For `folding`, that'd mean `folding` is not documented but `folding.enable` is.

If it's not working, then that probably means some edge-cases with how `isVisible` is used on whole trees are coming back to bite us, which isn't too surprising.

khaneliman:
Hmm yeah, for some reason I don't see `enable` documented 
image

MattSturgeon:
> for some reason I don't see `enable` documented

Probably related to `either`, which doesn't implement `getSubOptions`.

Testing your PR merged in and this additional patch:

```diff
diff --git a/plugins/by-name/treesitter/default.nix b/plugins/by-name/treesitter/default.nix
index 1438e1fc..8bb3d655 100644
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -162,8 +162,9 @@ lib.nixvim.plugins.mkNeovimPlugin {
       in
       mkOption {
         type = (types.either types.bool foldingSubmodule) // {
-          inherit (foldingSubmodule) description;
+          inherit (foldingSubmodule) description getSubOptions;
         };
+        visible = "transparent";
         default = {
           enable = false;
         };

```

And it seems to be working:

image

---

Yes, it's annoying that `folding` gets its own dedicated sub-page. Unfortunately that's just how our docs impl works; whenever we cross a suboptions boundary we fork into subpages. We also do this whenever options are nested more than one attr-level deep.

I could probably come up with something that avoids this when the submodule option is not visible, but it's hard to fully understand our docs impl which makes it hard to work on.

PR #4104, opened 2025-12-21
plugins/language-servers/rust-analyzer: drop rust-analyer settings options

GaetanLepage:
I discussed this IRL with @traxys and he agreed to drop the workflow for automatically declaring the `rust-analyzer` settings options.

### DIFF
--- a/plugins/by-name/rustaceanvim/settings-options.nix
+++ b/plugins/by-name/rustaceanvim/settings-options.nix
@@ -248,26 +248,21 @@ in
       ```
     '';
 
-    default_settings =
-      lib.nixvim.mkNullOrStrLuaFnOr
-        (types.submodule {
-          freeformType = with types; attrsOf anything;
-        })
-        ''
-          Setting passed to rust-analyzer.
-          Defaults to a function that looks for a `rust-analyzer.json` file or returns an empty table.
-          See https://rust-analyzer.github.io/manual.html#configuration.
+    default_settings = lib.nixvim.mkNullOrStrLuaFnOr lib.nixvim.lua-types.anything ''
+      Setting passed to rust-analyzer.
+      Defaults to a function that looks for a `rust-analyzer.json` file or returns an empty table.
+      See https://rust-analyzer.github.io/manual.html#configuration.
 
-          This can also be the definition of a function:
-          `fun(project_root:string|nil, default_settings: table|nil):table`
+      This can also be the definition of a function:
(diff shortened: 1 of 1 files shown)


ci/default.nix
```diff
@@ -17,7 +17,6 @@ lib.fix (self: {
   # Derivations that build the generated files
   efmls-configs-sources = pkgs.callPackage ./efmls-configs.nix { };
   none-ls-builtins = pkgs.callPackage ./none-ls.nix { };
-  rust-analyzer-options = pkgs.callPackage ./rust-analyzer { };
```
MattSturgeon, line 20: We also need to update `ci/generate.nix` to remove reference to `rust-analyzer-options`.

plugins/lsp/language-servers/rust-analyzer-config.nix
MattSturgeon, line 1: `plugins/by-name/rustaceanvim/settings-options.nix` is also importing this file.

ci/default.nix
GaetanLepage: Thanks! I was indeed waiting for the CI to catch the last occurrences. You were even faster!

MattSturgeon: CI wouldn't have caught this until we attempted our next update 😉

MattSturgeon: (generally it's a good idea to grep for things you're removing, CI doesn't _always_ go down _every_ code path, so it can miss things)

plugins/by-name/rustaceanvim/settings-options.nix
```diff
@@ -248,26 +248,21 @@ in
...
-          `fun(project_root:string|nil, default_settings: table|nil):table`
-
-          Plugin default:
-          ```lua
-            function(project_root, default_settings)
-              return require('rustaceanvim.config.server').load_rust_analyzer_settings(project_root, { default_settings = default_settings })
-            end
-          ```
-        '';
+    default_settings = lib.nixvim.mkNullOrStrLuaFnOr lib.nixvim.lua-types.anything ''
```
MattSturgeon, line 251: I think `tableOf anything` would be more accurate here, and better match the old type and the documented type. But `anything` is more lenient. Up to you.

GaetanLepage: I chose this as it is now the type of all `settings` options that are not explicitly typed as `submodule`s.

PR #4115, opened 2025-12-24
plugins/otter: fix treesitter usage

wadsaek:
I've noticed that the otter plugin doesn't play well with the new treesitter api, this should fix it. I've grepped the nixvim tree for other occurrences of the same issue but didn't seem to find anything, so hopefully this is a one-off

### DIFF
--- a/tests/test-sources/plugins/by-name/otter/default.nix
+++ b/tests/test-sources/plugins/by-name/otter/default.nix
@@ -5,6 +5,16 @@
       # Avoid the warning
       treesitter = {
         enable = true;
+        highlight.enable = true;
+      };
+    };
+  };
+  emptyOldApi = {
+    plugins = {
+      otter.enable = true;
+      # Avoid the warning
+      treesitter = {
+        enable = true;
         settings.highlight.enable = true;
       };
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN (approved):

tests/test-sources/plugins/by-name/otter/default.nix
```diff
@@ -15,7 +25,7 @@
       # Avoid the warning
       treesitter = {
         enable = true;
-        settings.highlight.enable = true;
```
khaneliman, line 18: Why is this removed if it's an "old api" test case

HeitorAugustoLN: This is the defaults test case

khaneliman: Sorry, missed the github code collapse line...

wadsaek: it's not, the diff just looks this way when folded. that's another test suite

khaneliman (approved):

PR #4117, opened 2025-12-25
plugins/yuck: init

Eveeifyeve:
Adds yuck.vim as a plugin to nixvim.

### DIFF
--- a/plugins/by-name/yuck/default.nix
+++ b/plugins/by-name/yuck/default.nix
@@ -1,0 +1,8 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkVimPlugin {
+  name = "yuck";
+  package = "yuck-vim";
+  globalPrefix = "yuck_";
+
+  maintainers = [ lib.maintainers.eveeifyeve ];
+}
--- a/tests/test-sources/plugins/by-name/yuck/default.nix
+++ b/tests/test-sources/plugins/by-name/yuck/default.nix
@@ -1,0 +1,5 @@
+{
+  empty = {
+    plugins.yuck.enable = true;
+  };
+}


plugins/by-name/yuck/default.nix
```diff
@@ -0,0 +1,12 @@
+{
+  lib,
+  ...
+}:
```
GaetanLepage:
```suggestion
{ lib, ... }:
```
nit

plugins/by-name/yuck/default.nix
```diff
@@ -0,0 +1,12 @@
...
+lib.nixvim.plugins.mkVimPlugin {
+  name = "yuck";
+  packPathName = "yuck.vim";
```
GaetanLepage:
```suggestion
```
This shouldn't be needed.

plugins/by-name/yuck/default.nix
```diff
@@ -0,0 +1,12 @@
...
+  package = "yuck-vim";
+  globalPrefix = "yuck_";
```
GaetanLepage, line 5: Does this plugin needs `yuck` to be installed? If so, then:
```suggestion
  globalPrefix = "yuck_";
  
  dependencies = [ "yuck" ];
```
Also, add a commit before this one: `dependencies: add yuck`, adding `yuck.default = "yuck";` to `modules/dependencies.nix`.

tests/test-sources/plugins/by-name/yuck/default.nix
```diff
@@ -0,0 +1,12 @@
+{
+  empty = {
+    plugins.yuck.enable = true;
+  };
+
+  no-packages = {
+    plugins.yuck = {
+      enable = true;
+      yuckPackage = null;
+    };
```
GaetanLepage:
```suggestion
    plugins.yuck.enable = true;
    dependencies.yuck.enable = false;
```

GaetanLepage (commented):
The commit message (and PR title) should be `plugins/yuck: init` (missing `s`).

plugins/by-name/yuck/default.nix
GaetanLepage: You didn't answer. Is `pkgs.yuck` needed by this plugin?

eveeifyeve: Yes `vimPlugins.yuck-vim`

GaetanLepage: This is fine, it's already handled by the `package` option. I was talking about an extra dependency. But I think it's not the case, so just drop `dependencies = [ "yuck" ];`.

eveeifyeve: Fixed.

PR #4126, opened 2025-12-28
plugins/otter: fix highlighting warning

MattSturgeon:
Folllowup to #4115, which adapted the warning to the new `plugins.treesitter.highlight.enable` option, but failed to remove the leftover `null` check.
`highlight == null` would _never_ be true.

### DIFF
--- a/tests/test-sources/plugins/by-name/otter/default.nix
+++ b/tests/test-sources/plugins/by-name/otter/default.nix
@@ -67,6 +67,7 @@
     test.warnings = expect: [
       (expect "count" 1)
       (expect "any" "You have enabled otter, but treesitter syntax highlighting is not enabled.")
+      (expect "any" "Make sure `plugins.treesitter.highlight.enable` and `plugins.treesitter.enable` are enabled.")
     ];
     plugins = {
       otter.enable = true;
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/by-name/otter/default.nix
```diff
@@ -61,4 +61,30 @@
       };
     };
   };
+
+  no-highlight-warning = {
```
khaneliman:
```suggestion
  warning-no-highlight = {
```

tests/test-sources/plugins/by-name/otter/default.nix
```diff
@@ -61,4 +61,30 @@
...
+      (expect "any" "You have enabled otter, but treesitter syntax highlighting is not enabled.")
+      (expect "any" "Make sure `plugins.treesitter.highlight.enable` and `plugins.treesitter.enable` are enabled.")
+    ];
+    plugins = {
+      otter.enable = true;
+      treesitter.enable = true;
+    };
+  };
+
+  no-treesitter-warning = {
```
khaneliman:
```suggestion
  warning-no-treesitter = {
```

khaneliman (commented):
Might make it easier to understand, I was very confused at first with the name of the test case vs the test expectations. (Thought they were supposed to expect a count of 0)

PR #4138, opened 2026-01-01
feat(output): add vimdiffAlias option to provide nvim -d as vimdiff alias

takeokunn:
## Summary

relative issue: #1621

Add a `vimdiffAlias` option that creates a `vimdiff` wrapper script invoking `nvim -d`, similar to the existing `viAlias` and `vimAlias` options.

## Changes

- Add `vimdiffAlias` option (`bool`, default `false`) to `modules/top-level/output.nix`
- Add `build.vimdiffPackage` output that creates a shell script wrapper for `nvim -d`
- Include `vimdiffPackage` in the final package when `vimdiffAlias = true`
- Remove the Home Manager-specific shell alias implementation (bash/fish/zsh) in favor of a universal wrapper script approach
- Add tests for both enabled and disabled states

## Motivation

Previously, the `vimdiffAlias` option only worked via Home Manager shell aliases, which was:
1. Limited to Home Manager users
2. Shell-specific (required separate aliases for bash, fish, zsh)
3. Inconsistent with how `viAlias` and `vimAlias` work (which use symlinks in the package)

This change makes `vimdiffAlias` work universally by creating an executable wrapper script included in the package itself, making it available regardless of shell or installation method.

## Test Plan

- [x] `vimdiffAlias-enabled` test verifies wrapper script contains `nvim` and `-d`
- [x] `vimdiffAlias-enabled` test verifies `vimdiff` exists and is executable in final package
- [x] `vimdiffAlias-disabled` test verifies `vimdiff` does NOT exist in final package

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -29,6 +29,14 @@ in
       '';
     };
 
+    vimdiffAlias = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Create a `vimdiff` alias that runs `nvim -d`.
+      '';
+    };
+
     waylandSupport = mkOption {
       type = types.bool;
       default = lib.meta.availableOn pkgs.stdenv.hostPlatform pkgs.wayland;
@@ -170,6 +178,15 @@ in
         description = ''
           A tool to show the content of the generated `init.lua` file.
           Run using `${config.build.printInitPackage.meta.mainProgram}`.
+        '';
+        readOnly = true;
+        visible = false;
+      };
+
+      vimdiffPackage = mkOption {
+        type = types.package;
+        description = ''
+          A wrapper script that runs `nvim -d` as `vimdiff`.
         '';
         readOnly = true;
         visible = false;
@@ -318,6 +335,9 @@ in
             ]
             ++ lib.optionals config.enablePrintInit [
               printInitPackage
+            ]
+            ++ lib.optionals config.vimdiffAlias [
+              vimdiffPackage
             ];
           meta.mainProgram = "nvim";
           passthru = {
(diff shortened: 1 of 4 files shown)


modules/top-level/output.nix
```diff
@@ -318,6 +335,9 @@ in
             ]
             ++ lib.optionals config.enablePrintInit [
               printInitPackage
+            ]
+            ++ lib.optionals config.vimdiffAlias [
+              vimdiffPackage
```
khaneliman, line 340: issue (blocking): detect an existing `bin/vimdiff` instead of choosing one by path order

`symlinkJoin` keeps the earlier `nvimPackage` entry when a custom supported package already ships `bin/vimdiff`, so enabling this option can silently leave the package-provided command in place instead of the promised `nvim -d` wrapper. Build the normal paths first, then in `postBuild` fail with an actionable message if `$out/bin/vimdiff` exists (telling the user to set `vimdiffAlias = false` to keep it); otherwise link the generated wrapper. Please add a collision regression test.

khaneliman (changes requested):

modules/top-level/output.nix
```diff
@@ -336,6 +356,10 @@ in
           '';
         };
 
+        vimdiffPackage = pkgs.writeShellScriptBin "vimdiff" ''
+          exec "${config.build.nvimPackage}/bin/nvim" -d "$@"
```
MattSturgeon, line 360: Style:

1. using shell quotes `"` around a nix interpolation gives a false sense of security to readers; either the string is safe to interpolate (and doesn't need quotes) or it isn't safe and should use `lib.escapeShellArg` or similar
2. where possible, use `lib.getExe` or `lib.getExe'` instead of `/bin/foo`


```suggestion
          exec ${lib.getExe config.build.nvimPackage}" -d "$@"
```

----

**However**, this looks to me like we should be using `makeBinaryWrapper --add-flag -d` instead of writing our own wrapper using `writeShellScriptBin`.

PR #4140, opened 2026-01-03
user-configs: add @literally-sai's config

literally-sai:


GaetanLepage (approved):
Thanks for sharing :)

PR #4150, opened 2026-01-12
modules/lsp: add `inotify-tools` package when any lsp server is enabled

wvffle:
Before this change `:LspInfo` complains about `vim.lsp` using `libuv-watchdirs` for file watcher:
```
vim.lsp: File Watcher ~
- File watch backend: libuv-watchdirs
- ⚠️ WARNING libuv-watchdirs has known performance issues. Consider installing inotify-tools.
```

This PR adds `inotify-tools` as extra package when any of the lsp servers is enabled via `lsp.servers.*.enable` option, so the warning goes away.

### DIFF
--- a/modules/lsp/servers/default.nix
+++ b/modules/lsp/servers/default.nix
@@ -151,6 +151,10 @@ in
       extraPackages = lib.mkIf (packages.prefix or [ ] != [ ]) packages.prefix;
       extraPackagesAfter = lib.mkIf (packages.suffix or [ ] != [ ]) packages.suffix;
 
+      dependencies.inotify-tools = lib.mkIf (enabledServers != [ ] && pkgs.stdenv.hostPlatform.isLinux) {
+        enable = true;
+      };
+
       lsp.luaConfig.content =
         let
           mkServerConfig =
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -135,6 +135,7 @@ in
       grep.default = "gnugrep";
       gzip.default = "gzip";
       imagemagick.default = "imagemagick";
+      inotify-tools.default = "inotify-tools";
       jupytext.default = [
         "python313Packages"
         "jupytext"


HeitorAugustoLN:
Not sure about enabling it for any LSP, I don't have any LSP enabled that uses file watching. I think we can add `inotify-tools` in dependencies, and let each user enable it if they want

MattSturgeon:
> Not sure about enabling it for any LSP, I don't have any LSP enabled that uses file watching. I think we can add `inotify-tools` in dependencies, and let each user enable it if they want

My instinct is that LSPs that benefit from it should enable the dependency.

Maybe we can implement a test that greps through `Checkhealth` and/or `LspInfo` output, to validate each LSP gets the right dependencies?

I could be convinced that _any_ LSP being installed should enable it _if_ that is impractical to maintain that, though. Assuming installing it by default is relatively harmless & uncontroversial.

modules/lsp/servers/default.nix
```diff
@@ -151,6 +151,10 @@ in
       extraPackages = lib.mkIf (packages.prefix or [ ] != [ ]) packages.prefix;
       extraPackagesAfter = lib.mkIf (packages.suffix or [ ] != [ ]) packages.suffix;
 
+      dependencies.inotify-tools = lib.mkIf (enabledServers != [ ] && pkgs.stdenv.hostPlatform.isLinux) {
+        enable = true;
+      };
```
MattSturgeon, line 156: 1. As per https://github.com/nix-community/nixvim/pull/4075#issuecomment-3651092419 I'm unsure about continuing to spread usage of `dependencies` until we've ironed out how to advertise to users that a particular option/module enables a particular dependency.

2. As per https://github.com/nix-community/nixvim/pull/4150#issuecomment-3745175920 and https://github.com/nix-community/nixvim/pull/4150#issuecomment-3745433614, I'd ideally prefer to see a way of us maintaining this on a per-LSP-server basis, too.

I'm willing to overlook pt.1 if this is considered an important enough dependency that we need a solution now, before solving the wider `dependencies` questions.

Currently, I suspect pt.2 would be too high a maintenance burden. In the future if we had a way to assert specific `:LspInfo` output with nixvim's `test` module, this could become easier.

Perhaps we can address pt.2 by adding a TODO comment, along the lines of:

> Ideally this dependency should be enabled by the servers that depend on inotify-tools, instead of globally for all servers

Issue #1920, opened 2024-07-24
Improve config files and rtp handling


MattSturgeon:
@stasjok has an implementation here: [stasjok/nixvim@`optimize-rtp` (commits)](https://github.com/stasjok/nixvim/commits/optimize-rtp/) (last two commits), where the rtp is optimized as a "performance" option.

> In the codebase I need to check `wrapRc` and if `performance.optimizeRuntimePath` is enabled. So it kind of strange to allow adding additional paths to rtp, but it works only if optimization is enabled and doesn't work otherwise...
> 
> So my thoughts are, if there is an interest in this feature, it shouldn't be `performance` options. Nixvim should do it by default with additional options to allow adding extra paths to rtp and packpath. It gives several benefits:
> 
> 1. nixvim will be pure by default (only paths managed by nixvim will be in rtp by default), but user always can add additional paths with options. Some plugins, like treesitter, can add the paths there automatically when nix parser management is disabled for example.
> 2. filesPlugin should never be added as a plugin. This is wrong to begin with. That's because user files should always be first in rtp. When filesPlugins is in the mix with other plugins, overriding plugin files may not work (depending on sorting order).
> 3. I think that there should be a default internal variable with a path to user files. By default it point to filesPlugin, but other modules (wrappers) can override it with it's own path (home-manager should set it to xdg config path that is perfectly known to it without using raw lua, like I did now, nixos can override it with etc directory with configs and so on). Final rtp is: `configpath,plugins,nvimruntime,configafter`.
> 4. Potentially we can manage rtp partially with wrapper args, in that case we can even drop `wrapRc` completely and just place init.lua to filesPlugin. But this is debatable, maybe managing rtp paths as the first lines in init.lua is enough plus allow users to add paths with raw lua. Or it can mixed approach: user config path is added with command line args (so that the directory with init.lua is known to neovim), other paths are managed in init.lua.
> 
> In the end since all above as pretty large refactor and it should also change how things are managed right now, I put this task aside and decided to just change my lua code with a simple nix config in my config (in home-manager options I can get a static paths to my config files, to plugin path and to Nvim runtime, so no lua loops and matching needed, just plain vim.o.runtimepath assignment.

https://github.com/nix-community/nixvim/pull/1886#issuecomment-2247835633

Related issues:

- We're conflating `wrapRc` with "isolating" the rtp - these are really separate issues.
  - `wrapRc` should just be about listing the init file in the wrapper args, using `-u`
[...]

PR #4151, opened 2026-01-13
modules/performance/byte-compile-lua: support compiling directories and fixed-output derivations

stasjok:
This PR:

* re-enables `modules/performance` tests disabled in #3562
* fixes byte-compiling of fixed-output derivations, e.g. `fetchurl`
* adds ability to byte-compile directories added in #4016 

Note, that for now it supports only simple directories like `./dir`. It doesn't support derivations, like `fetchFromGitHub`.

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -67,17 +67,34 @@ let
           );
           finalSource =
             # Byte compile lua files if performance.byteCompileLua option is enabled
-            if byteCompileLua && lib.hasSuffix ".lua" config.target then
-              # It is possible to remove this condition entirely and use only
-              # `builders.byteCompileLuaFile` for all files, but this way it's
-              # slightly faster, because `finalSource` is built directly from
-              # `text`, not from intermediate `source`
-              if lib.isDerivation config.source && !config.source ? outputHash then
-                # Source is a derivation (not fixed-output)
-                builders.byteCompileLuaDrv config.source
+            if byteCompileLua then
+              if lib.hasSuffix ".lua" config.target then
+                # It is possible to remove this condition entirely and use only
+                # `builders.byteCompileLuaFile` for all files, but this way it's
+                # slightly faster, because `finalSource` is built directly from
+                # `text`, not from intermediate `source`
+                if lib.isDerivation config.source && !config.source ? outputHash then
+                  # Source is a derivation (not fixed-output)
+                  builders.byteCompileLuaDrv config.source
+                else
+                  # Source is a path, string or fixed-output derivation
+                  builders.byteCompileLuaFile derivationName config.source
+              else if
+                builtins.isPath config.source
+                && lib.filesystem.pathIsDirectory config.source
+                && builtins.any (lib.hasSuffix ".lua") (lib.filesystem.listFilesRecursive config.source)
+              then
+                # Source is a directory path with lua files
+                builders.byteCompileLuaDrv (
+                  pkgs.stdenvNoCC.mkDerivation {
+                    name = derivationName;
+                    src = config.source;
+                    phases = "unpackPhase installPhase fixupPhase";
+                    installPhase = "cp -R ./ $out";
+                  }
+                )
               else
-                # Source is a path, string or fixed-output derivation
-                builders.byteCompileLuaFile derivationName config.source
+                config.source
             else
               config.source;
         };
--- a/tests/test-sources/modules/performance/byte-compile-lua.nix
+++ b/tests/test-sources/modules/performance/byte-compile-lua.nix
@@ -128,8 +128,14 @@ in
           url = "file://${./files/fixed_output.lua}";
           hash = "sha256-MnPksxTLb5e47HOgtDUuxlEKJkP4av9zV/679MazN0E=";
         };
+        # Directory
+        "plugin/extra_file_dir".source = ./files/dir;
+        # Subdirectory
+        "plugin/extra_file_subdir".source = ./files/subdir;
         # Non-lua file
         "plugin/extra_file_non_lua.vim".text = "let g:extra_file_non_lua = 1";
+        # Directory with non-lua file
+        "plugin/extra_file_dir_non_lua".source = ./files/dir_non_lua;
         # Lua file with txt extension won't be byte compiled
         "text.txt".source = writeLua "text.txt" "vim.g.text = true";
       };
@@ -180,8 +186,11 @@ in
             { "plugin/extra_file_string.lua", true, "extra_file_string" },
             { "plugin/extra_file_drv_string.lua", true, "extra_file_drv_string" },
             { "plugin/extra_file_fixed_output.lua", true, "extra_file_fixed_output" },
+            { "plugin/extra_file_dir/dir.lua", true, "extra_file_dir" },
+            { "plugin/extra_file_subdir/sub/subdir.lua", true, "extra_file_subdir" },
             -- other 'extraFiles'
             { "plugin/extra_file_non_lua.vim", false, "extra_file_non_lua" },
+            { "plugin/extra_file_dir_non_lua/file.vim", false, "extra_file_dir_non_lua" },
             -- lua 'files' are byte compiled
             { "plugin/file.lua", true, "file_lua" },
             -- other 'files'
@@ -215,7 +224,10 @@ in
   disabled = {
     performance.byteCompileLua.enable = false;
 
-    extraFiles."plugin/test1.lua".text = "vim.opt.tabstop = 2";
+    extraFiles = {
+      "plugin/test1.lua".text = "vim.opt.tabstop = 2";
+      "plugin/dir".source = ./files/dir;
+    };
 
     files."plugin/test2.lua".opts.tabstop = 2;
 
@@ -232,6 +244,7 @@ in
       assert(not is_byte_compiled(init), "MYVIMRC is not expected to be byte compiled, but it is")
       -- extraFiles
       test_rtp_file("plugin/test1.lua", false)
+      test_rtp_file("plugin/dir/dir.lua", false)
       -- files
       test_rtp_file("plugin/test2.lua", false)
       -- Neovim runtime
--- a/tests/test-sources/modules/performance/files/dir/dir.lua
+++ b/tests/test-sources/modules/performance/files/dir/dir.lua
@@ -1,0 +1,1 @@
+vim.g.extra_file_dir = true
--- a/tests/test-sources/modules/performance/files/dir_non_lua/file.vim
+++ b/tests/test-sources/modules/performance/files/dir_non_lua/file.vim
@@ -1,0 +1,1 @@
+let g:extra_file_dir_non_lua = 1
--- a/tests/test-sources/modules/performance/files/subdir/sub/subdir.lua
+++ b/tests/test-sources/modules/performance/files/subdir/sub/subdir.lua
@@ -1,0 +1,1 @@
+vim.g.extra_file_subdir = true


MattSturgeon:
> * adds ability to byte-compile directories added in

I was recently considering a new approach to byte-compiling extraFiles, which might tie in nicely with another potential refactor to demote `init.lua` to being "just another" extraFile.

---

Essentially, we add `build.extraFiles.passthru.byteCompiled`, which is _effectively_ a symlinkJoin of `build.extraFiles` but every lua symlink gets replaced with a byte-comp version.

(psudocode):

```nix
passthru.byteCompiled = symlinkJoin {
  paths = [ finalAttrs.finalPackage ];
  postBuild = ''
    lua_files=("$out"/**/*.lua)
    byte_compile_and_replace "''${lua_files[@]}"
  '';
}
```

Alternatively, instead of a bash array of lua symlinks, we could:

1. construct an associative array of out-path -> symlink target
2. remove all lua symlinks (array keys)
3. byte-compile all targets (array values)
4. write the results to the respective out paths (array keys)

Or whatever variation on the concept that ends up working best.

---

Once we have this `build.extraFiles` and its `passthru.byteCompiled` wrapper, we can then select which one to include in the actual rtp: `if config.performance.byteCompileLua.enable then config.build.extraFiles.byteCompiled else config.build.extraFiles`

modules/files.nix
```diff
@@ -68,11 +68,15 @@ let
...
             if byteCompileLua && lib.hasSuffix ".lua" config.target then
-              if lib.isDerivation config.source then
-                # Source is a derivation
+              # It is possible to remove this condition entirely and use only
+              # `builders.byteCompileLuaFile` for all files, but this way it's
+              # slightly faster, because `finalSource` is built directly from
+              # `text`, not from intermediate `source`
+              if lib.isDerivation config.source && !config.source ? outputHash then
+                # Source is a derivation (not fixed-output)
                 builders.byteCompileLuaDrv config.source
```
MattSturgeon, line 77: > Fetchers doesn't execute postFixup hooks, therefore byteCompileLuaHook
> doesn't work. And even if it would work, fetchers are fixed-output
> derivation which has outputHash known beforehand. We can't override such
> derivations using byteCompileLuaDrv.
> This commit uses byteCompileLuaFile for fixed-output derivations.

Looking at this with fresh eyes, I'm questioning why we are overriding derivations at all; it's usually better to wrap them (e.g. with `symlinkJoin`) and then mess with the outputs in a fresh derivation instead of rebuilding the original derivation.

The exception being when the wrapped derivation and the wrapper derivation essentially need to do the same compilation step anyway.

I'm not keen on how we end up handling paths, FODs, drvs, files, directories, (etc) differently from one and other.

---

My proposal in https://github.com/nix-community/nixvim/pull/4151#issuecomment-3757075358 would mean dropping the `finalSource` option entirely and handling this much later, in a wrapper of `build.extraFiles`.

tests/test-sources/modules/performance/byte-compile-lua.nix
```diff
@@ -123,6 +123,11 @@ in
         "plugin/extra_file_drv_string.lua".source = toString (
           writeLua "extra_file_drv_string.lua" "vim.g.extra_file_drv_string = true"
         );
+        # By fixed-output derivation
+        "plugin/extra_file_fixed_output.lua".source = pkgs.fetchurl {
+          url = "file://${./files/fixed_output.lua}";
+          hash = "sha256-MnPksxTLb5e47HOgtDUuxlEKJkP4av9zV/679MazN0E=";
+        };
```
MattSturgeon, line 130: `pkgs.emptyFile` is a FOD:

```
nix-repl> pkgs.emptyFile.outputHash
"sha256-d6xi4mKdjkX2JFicDIv5niSzpyI0m/Hnm8GGAIU04kY="
```

That may be a cheaper test fixture if we end up sticking with this approach of handling FODs specially. I guess even if we don't, test coverage is always good.

modules/files.nix
```diff
@@ -67,17 +67,34 @@ let
...
+                if lib.isDerivation config.source && !config.source ? outputHash then
+                  # Source is a derivation (not fixed-output)
+                  builders.byteCompileLuaDrv config.source
+                else
+                  # Source is a path, string or fixed-output derivation
+                  builders.byteCompileLuaFile derivationName config.source
+              else if
+                builtins.isPath config.source
+                && lib.filesystem.pathIsDirectory config.source
+                && builtins.any (lib.hasSuffix ".lua") (lib.filesystem.listFilesRecursive config.source)
```
MattSturgeon, line 85: > This commit adds support for byte-compiling path directories. Only paths
> like `./dir` are supported. Directory derivations e.g. fetchers are not
> supported.

Again, my instinct is that delaying byte-comp until during/after building `build.extraFiles`, as per https://github.com/nix-community/nixvim/pull/4151#issuecomment-3757075358, would massively simplify this. We wouldn't need any eval-time inspection. Instead all inspection of "symlink, file, or directory" would be done in the build script.

tests/test-sources/modules/performance/byte-compile-lua.nix
MattSturgeon: Although, I guess you want an actual lua file for the test fixture?

MattSturgeon (commented):
Thanks for addressing these issues!

On my initial read through, I don't see any obvious technical issues with the changes. However, I do feel like the issues we're needing to solve highlight how we could benefit from a simpler, more delayed approach.

I'd appreciate you're thoughts on my above concept; whether you think the concept is better than the current architecture; whether you'd like to implement something like that; whether you'd like to continue with this PR (first/instead); whether you'd be interested in reviewing if someone else worked on the idea; etc

tests/test-sources/modules/performance/byte-compile-lua.nix
stasjok: > I guess you want an actual lua file for the test fixture?

Of course. Otherwise how would I test if the compilation actually works?

stasjok:
> Essentially, we add `build.extraFiles.passthru.byteCompiled`, which is _effectively_ a symlinkJoin of `build.extraFiles` but every lua symlink gets replaced with a byte-comp version.

There is one caveat. `build.extraFiles` aren't used when `wrapRc = false` (home-manager default, nixos default disabled, but can be enabled). In that case extraFiles are propagated to the corresponding `xdg.config` options skipping `build.extraFiles`.

I guess if the handling of runtimepath is changed (some of the discussion was in #1920), it could be the thing. For me it doesn't matter whether it `passthru.byteCompiled` attr or _yet another "final" option_.

There is another thing. The code need to be much smarter than https://github.com/nix-community/nixvim/blob/03a638205b5cb04ba9c2ed6c604e137b15f07fa1/lib/builders.nix#L129-L137 This simple approach doesn't handle symlinks to directories (which symlinkJoin produce). Technically symlink could point anywhere, so it need to resolve all links, verify that they point to the nix store, replace directory-simlinks with the actual directories and file-symlinks in them etc.

> Looking at this with fresh eyes, I'm questioning why we are overriding derivations at all; it's usually better to wrap them (e.g. with `symlinkJoin`) and then mess with the outputs in a fresh derivation instead of rebuilding the original derivation.

Originally this was implemented like this for optimization. In the build time, overridden derivation needs to be built once. With symlink it needs to be built twice: one derivation for `source` (converted from `text`, or any use-provided derivation), and another derivation for `finalSource` (which has a dependency on the `source` derivation). In the runtime, extra symlinks adds a tiny overhead for resolving them compared to regular files.

For this reason I wanted to determine if byte-compilation is actually needed during evaluation (also as a precaution, who knows what can be in `source`). This was simple enough when `extraFiles` were actually _files_. Now when it can be directories, it's not simple. It's impossible to determine if the directory needs compilation without IFD.

> my instinct is that delaying byte-comp until during/after building `build.extraFiles`, as per [#4151 (comment)](https://github.com/nix-community/nixvim/pull/4151#issuecomment-3757075358), would massively simplify this

Working on this, I already thought that the current approach is limiting. I don't like this _if_ monstrosity. And it was already an idea that we'll have to switch to build-time byte-compilation at some point. I opted to the most direct/simple approach for now.

To move forward, e.g. supporting byte-compilation of any directory-like path/string/derivation, I think the only option is to write a new `lib.builders` that would be smart enough to handle directory-symlinks and use it without conditions, effectively dropping optimizations I described above (I didn't resolved on that yet).

`byteCompileLuaDrv` can be used too, but it doesn't work for symlinked directories (usually it's a rare case like `symlinkJoin`). Technically this already passes all test cases:

```nix
          finalSource =
            # Byte compile lua files if performance.byteCompileLua option is enabled
            if byteCompileLua then
              builders.byteCompileLuaDrv (
                pkgs.stdenvNoCC.mkDerivation {
                  name = derivationName;
                  src = config.source;
                  phases = "installPhase fixupPhase";
                  installPhase = ''
                    if [ -d $src ]; then
                      cp -r --no-preserve=mode $src $out
                    else
                      cp --no-preserve=mode $src $out
                    fi
                  '';
                }
              )
            else
              config.source;
```

> I'd appreciate you're thoughts on my above concept; whether you think the concept is better than the current architecture;

I don't think there is much difference at what point it's compiled. Originally I searched for the single place to add my changes, so all other places would work without changes. Same for passthru attr/another option. I don't particularly favor one "storage" for the compiled derivation over another.

> whether you'd like to implement something like that;

I think I won't be doing a big refactoring. If `build.extraFiles` would be the only possible source for files (see home-manager), then it would be simple to move compilation there. Until that current `finalSource` approach looks simple enough.

> whether you'd like to continue with this PR (first/instead); whether you'd be interested in reviewing if someone else worked on the idea; etc

Currently I think this PR is a simple fix. The thing I'm not sure about is this complicated _if statement_. It can be dropped in favor of the code above with the potential of improving/rewriting of `byteCompileLuaDrv`. This will simplify the code at the cost of small performance overhead, but it will support other directory sources.

MattSturgeon:
> There is one caveat. `build.extraFiles` aren't used when `wrapRc = false` (home-manager default, nixos default disabled, but can be enabled). In that case extraFiles are propagated to the corresponding `xdg.config` options skipping `build.extraFiles`.

Yes, I hadn't thought about this when initially considering the idea, but quickly discovered it when experimenting locally.

I think there's a few ways to handle this:

1. Drop the idea of propagating `extraFiles` to the host
   One alternative being to install an entire directory to `~/.config/nvim`.
2. We could map over `extraFiles` while propagating, applying a lua builder to the `sources`.
   If the lua builder handles directories, then it could be the same one used for `build.extraFiles`.
   This does mean some duplication of impl (although the host-extraFiles-propagation is already a duplication in a way).

> This simple approach doesn't handle symlinks to directories (which symlinkJoin produce). Technically symlink could point anywhere, so it need to resolve all links, verify that they point to the nix store, replace directory-simlinks with the actual directories and file-symlinks in them etc.

Yes, I'd already considered this but wanted to go over high-level concepts before discussing details 🙂 

My idea would be to recursively _copy_ the base drv's output tree into the wrapper's `$out`, following symlinks-to-directories recursively (checking for loops and broken links). When reaching non-directory (leaf) nodes, it'd either symlink to the file or byte-compile it and install the result into `$out`.

This could work as a new builder in `lib.builders`, something like `lib.builders.byteCompileTree`?

> Originally this was implemented like this for optimization. In the build time, overridden derivation needs to be built once. With symlink it needs to be built twice: one derivation for `source` (converted from `text`, or any use-provided derivation), and another derivation for `finalSource` (which has a dependency on the `source` derivation).

This makes sense in the typical case, where `source` is a trivial derivation (e.g. `writeText`). If someone decided to assign an expensive derivation to `source`, then wrapping becomes cheaper than overriding, but I appreciate that is unlikely when we're mostly dealing with text (lua/vim) files.

> Currently I think this PR is a simple fix. The thing I'm not sure about is this complicated _if statement_. It can be dropped in favor of the code above with the potential of improving/rewriting of `byteCompileLuaDrv`. This will simplify the code at the cost of small performance overhead, but it will support other directory sources.

Is that something you'd like to explore here or in future work?

> I think I won't be doing a big refactoring.

Sure. I'll consider looking into it _after_ this PR then, if it's still bothering me by then.

stasjok:
> > Currently I think this PR is a simple fix. The thing I'm not sure about is this complicated _if statement_. It can be dropped in favor of the code above with the potential of improving/rewriting of `byteCompileLuaDrv`. This will simplify the code at the cost of small performance overhead, but it will support other directory sources.
> 
> Is that something you'd like to explore here or in future work?

To be honest, I'm fine with it as is. But if you think that the logic became too complicated, I can change it to something like above. Though I'm not a fan if all non-lua files will be symlinks (since most files are generated from `source` text, I think it's better to generate them directly in target).

PR #4161, opened 2026-01-19
plugins/lint: allow configuring linters with special names

PierreBorine:
This PR makes configuring linters with special characters in their names possible.

The following code would currently result in a compilation error because `markdownlint-cli2` contains a dash character.
```nix
plugins.lint.linters.markdownlint-cli2 = {
  cmd = lib.getExe pkgs.markdownlint-cli2;
  stdin = true;
  args = [
    "--config"
    "${mdConfig}"
    "-"
  ];
};
```

### DIFF
--- a/plugins/by-name/lint/default.nix
+++ b/plugins/by-name/lint/default.nix
@@ -228,7 +228,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
             linterConfig' =
               if builtins.isString linterConfig then lib.nixvim.mkRaw linterConfig else linterConfig;
           in
-          "__lint.linters.${customLinter} = ${toLuaObject linterConfig'}"
+          ''__lint.linters["${customLinter}"] = ${toLuaObject linterConfig'}''
         ) cfg.customLinters
       )
     ))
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN (approved):

khaneliman (approved):

plugins/by-name/lint/default.nix
```diff
@@ -228,7 +228,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
             linterConfig' =
               if builtins.isString linterConfig then lib.nixvim.mkRaw linterConfig else linterConfig;
           in
-          "__lint.linters.${customLinter} = ${toLuaObject linterConfig'}"
+          ''__lint.linters["${customLinter}"] = ${toLuaObject linterConfig'}''
```
MattSturgeon, line 231: This approach still fails if a linter name contains `"` , `\`, or other quoted-string syntax.

The correct approach would be:


```suggestion
          ''__lint.linters[${toLuaObject customLinter}] = ${toLuaObject linterConfig'}''
```

PR #4179, opened 2026-01-29
modules/dependencies: add `packageFallback` option

sglre6355:
This PR:
- introduces `packageFallback` option to `dependencies` module
  - When set to true, the package is added to `extraPackagesAfter` instead of `extraPackages`
  - This is aligned with the existing [`lsp.servers.*.packageFallback` definition](https://github.com/nix-community/nixvim/blob/3c27e1b35ca0fee6a89bfc20840654361ffe888d/modules/lsp/servers/server.nix#L56-L64)
- adds tests for the option

This allows users to fall back to local dependencies (e.g. via `direnv`) and prevents version mismatches when LSPs use dependencies from this module instead of project-local ones.

### DIFF
--- a/modules/dependencies.nix
+++ b/modules/dependencies.nix
@@ -30,6 +30,16 @@ let
       // lib.optionalAttrs (literalExpressionType.check properties.example) {
         inherit (properties) example;
       };
+
+    packageFallback = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        When enabled, the ${name} package will be added to the end of the `PATH` _(suffix)_ instead of the beginning _(prefix)_.
+
+        This can be useful if you want local versions (e.g. from a devshell) to override the Nixvim version.
+      '';
+    };
   };
 
   # Motivation:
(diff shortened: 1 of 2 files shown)


modules/dependencies.nix
```diff
@@ -96,7 +106,13 @@ in
...
-      (builtins.filter (p: p.enable))
+      (builtins.filter (p: p.enable && !p.packageFallback))
+      (map (p: p.package))
+    ];
+
+    extraPackagesAfter = lib.pipe cfg [
+      builtins.attrValues
+      (builtins.filter (p: p.enable && p.packageFallback))
       (map (p: p.package))
     ];
```
MattSturgeon: If we're going to use the same pipeline twice, let's DRY it up with a `partition`.

Arguably less efficient because we filter _then_ partition, arguably more efficient because we run the pipeline only once.

Given the pipeline already loops over the list multiple times, I think the latter is true.

Either way, I think the latter is easier to maintain.

```nix
# phone keyboard psudocode
config =
  let
    enabledPackages = lib.pipe cfg [ 
      builtins.attrValues
      (lib.filter (p: p.enable))
      (lib.partition (p: p.packageFallback))
      (lib.mapAttrs (_: map (p: p.package)))
      (parts: {
        prefix = parts.wrong;
        suffix = parts.right;
      })
    ];
  in
  {
    extraPackages = enabledPackages.prefix;
    extraPackagesAfter = enabledPackages.suffix;
  }
```

(Pulling the `let` block up to the top of the file would avoid an indentation change for the rest of `config`)

sglre6355:
Hi, thanks for your reviews. I've reflected the suggestion and hopefully it's good to go.

sglre6355:
@MattSturgeon Done!

PR #4191, opened 2026-02-06
plugins/gitlineage: init

carl0xs:
Add [gitlineage.nvim](https://github.com/LionyxML/gitlineage.nvim), a plugin for view git history for selected lines.

### DIFF
--- a/plugins/by-name/gitlineage/default.nix
+++ b/plugins/by-name/gitlineage/default.nix
@@ -1,0 +1,54 @@
+{
+  lib,
+  ...
+}:
+let
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gitlineage";
+  package = "gitlineage-nvim";
+
+  description = "Show git lineage information inside Neovim.";
+
+  maintainers = with lib.maintainers; [ LionyxML ];
+
+  settingsOptions = {
+    split = defaultNullOpts.mkStr "auto" ''
+      Where the git lineage window should open.
+      Accepts "vertical", "horizontal", or "auto".
+    '';
+
+    keymap = defaultNullOpts.mkStr "<leader>gl" ''
+      Default keymap for opening the git lineage window.
+      Set to `null` to disable.
+    '';
+
+    keys =
+      defaultNullOpts.mkAttrsOf (types.nullOr types.str)
+        {
+          close = "q";
+          next_commit = "]c";
+          prev_commit = "[c";
+          yank_commit = "yc";
+          open_diff = "<CR>";
+        }
+        ''
+          Keymaps inside the git lineage window.
+          Set individual entries to `null` to disable.
+        '';
+  };
+
+  settingsExample = {
+    split = "auto";
+    keymap = "<leader>gl";
+    keys = {
+      close = "q";
+      next_commit = "]c";
+      prev_commit = "[c";
+      yank_commit = "yc";
+      open_diff = "<CR>";
+    };
+  };
+}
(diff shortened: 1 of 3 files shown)


GaetanLepage (changes requested):
Thanks for the contribution.

If the plugin is not yet packaged in nixpkgs, then you need to add it there first (`nix run .#vimPluginsUpdater -- add LionyxML/gitlineage.nvim`).

plugins/by-name/gitlineage/default.nix
```diff
@@ -0,0 +1,21 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gitlineage";
+  moduleName = "gitlineage";
```
GaetanLepage:
```suggestion
  package = "gitlineage-nvim";
```

GaetanLepage (commented):
This should fix it! Also, please squash your commits into a single one:
`plugins/gitlineage: init`

plugins/by-name/gitlineage/default.nix
```diff
@@ -0,0 +1,54 @@
...
+          prev_commit = "[c";
+          yank_commit = "yc";
+          open_diff = "<CR>";
+        }
+        ''
+          Keymaps inside the git lineage window.
+          Set individual entries to `null` to disable.
+        '';
+  };
+
```
GaetanLepage, line 42: Now, we tend to avoid explicitly declaring `settingsOptions`. The `settings` option is already freeform by default, so it accepts arbitrary values.

GaetanLepage (commented):
You can now rebase on `main` and remove the change to `flake.lock`.

plugins/by-name/gitlineage/default.nix
```diff
@@ -0,0 +1,54 @@
...
+  inherit (lib) types;
+  inherit (lib.nixvim) defaultNullOpts;
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "gitlineage";
+  package = "gitlineage-nvim";
+
+  description = "Show git lineage information inside Neovim.";
+
+  maintainers = with lib.maintainers; [ LionyxML ];
```
khaneliman, line 15: issue (blocking): `lib.maintainers.LionyxML` does not exist in the pinned nixpkgs or in `lib/maintainers.nix`.

Every consumer of `meta.maintainers` fails with `attribute 'LionyxML' missing`, which is why all buildbot checks are red on this head. Add a maintainer entry in a separate `maintainers: add <name>` commit. Note the nixvim maintainer is the module maintainer, so listing yourself rather than the upstream plugin author is the usual choice.

plugins/by-name/gitlineage/default.nix
```diff
@@ -0,0 +1,54 @@
...
+          yank_commit = "yc";
+          open_diff = "<CR>";
+        }
+        ''
+          Keymaps inside the git lineage window.
+          Set individual entries to `null` to disable.
+        '';
+  };
+
+  settingsExample = {
```
khaneliman, line 43: suggestion: this example repeats the upstream defaults verbatim, so it adds nothing over the generated docs.

When you drop `settingsOptions` per the earlier review, switch the example to non-default values (e.g. `split = "vertical"` plus a disabled keymap) so the docs show a meaningful override.

tests/test-sources/plugins/by-name/gitlineage/default.nix
```diff
@@ -0,0 +1,41 @@
...
+
+      settings = {
+        split = "auto";
+        keymap = "<leader>gl";
+        keys = {
+          close = "q";
+          next_commit = "]c";
+          prev_commit = "[c";
+          yank_commit = "yc";
+          open_diff = "<CR>";
```
khaneliman, line 18: note: `keys.open_diff` requires diffview.nvim upstream and no test enables it.

Optional: add a case with `plugins.diffview.enable = true;` to cover the integration path, or mention the optional dependency in the module `description`.

khaneliman (changes requested):
Packaging and squash requests from earlier review landed, but the settingsOptions removal and flake.lock rebase are still outstanding. One new blocker below explains the all-red buildbot status.

Issue #4197, opened 2026-02-11
[BUG] mkdnflow - mappings syntax not documented


Ginner:
### Plugin

mkdnflow

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

- [x] I have read the [FAQ](https://nix-community.github.io/nixvim/user-guide/faq.html) and my bug is not listed there.

### Description

For a long time, I've been using the following mappings syntax for mkdnflow:
```
programs.nixvim.plugins.mkdnflow = {
  enable = true;
  settings = {
    mappings = {
      MkdnEnter = {
        key = "<cr>";
        modes = ["n" "i"];
      };
      MkdnToggleToDo = {
        key = "<c-space>";
        modes = ["n" "i"];
      };
    };
  };
};
```
Recently this has resulted in the mappings not working and the following error when opening a markdown file:
```
Error detected while processing BufWinEnter Autocommands for "*.md":                                               
                                                                                                                   
Error executing lua callback: ...NeovimPackages/start/mkdnflow.nvim/lua/mkdnflow/maps.lua:53: Invalid 'lhs': Expect
ed Lua string                                                                                                      
stack traceback:                                                                                                   
        [C]: in function 'nvim_buf_set_keymap'                                                                     
        ...NeovimPackages/start/mkdnflow.nvim/lua/mkdnflow/maps.lua:53: in function <...NeovimPackages/start/mkdnfl
ow.nvim/lua/mkdnflow/maps.lua:31>                                                                                  
```
After some troubleshooting the following syntax solved the issue:
```
programs.nixvim.plugins.mkdnflow = {
  enable = true;
  settings = {
    mappings = {
      MkdnEnter = [ [ "i" "n" ] "<cr>" ];
      MkdnToggleToDo = [ ["n" "i"] "<c-space>" ];
    };
  };
};
```
This syntax is not apparent from the documentation and I find the former syntax more logical and adherent with the other configurations.

If this change is intentional and made for good reason, I'll be happy to try and contribute some documentation.

[Link to documentation](https://nix-community.github.io/nixvim/plugins/mkdnflow/index.html?highlight=mkd#pluginsmkdnflowsettings)

### Minimal, Reproducible Example (MRE)

```nix
programs.nixvim.plugins.mkdnflow = {
  enable = true;
  settings = {
    to_do.symbols = [" " "⧖" "x"];
    mappings = {
      MkdnEnter = {
        key = "<cr>";
[...]

PR #4198, opened 2026-02-11
plugins/mkdnflow: properly handle old mappings

khaneliman:
Migrate user's with old mappings to new format and warn about change.

Closes https://github.com/nix-community/nixvim/issues/4197

### DIFF
--- a/tests/test-sources/plugins/by-name/mkdnflow/default.nix
+++ b/tests/test-sources/plugins/by-name/mkdnflow/default.nix
@@ -78,133 +78,153 @@
           };
         };
         mappings = {
+          MkdnEnter = [
+            [
+              "n"
+              "v"
+              "i"
+            ]
+            "<CR>"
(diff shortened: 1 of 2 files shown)


tests/test-sources/plugins/by-name/mkdnflow/default.nix
```diff
@@ -78,133 +78,153 @@
...
+            "n"
+            "<leader>F"
+          ];
+        };
+      };
+    };
+
+  };
+
+  mappings-deprecated = {
```
MattSturgeon, line 212: Should this test assert the warning?

MattSturgeon: I suppose it can't, as we're using `lib.warn` rather than the `warnings` option.

MattSturgeon (approved):

PR #4209, opened 2026-02-27
plugins/conform: change state semantics

khaneliman:
Didn't realize `darwinOnly` didn't actually warn/fail install on linux
only. It just passes a state message to the auto-install function.
Change the state function to automatically handle the expected
condition.

### DIFF
--- a/plugins/by-name/conform-nvim/formatter-packages.nix
+++ b/plugins/by-name/conform-nvim/formatter-packages.nix
@@ -7,6 +7,7 @@ let
   states = {
     broken = _package: "broken";
     darwinOnly = p: if stdenv.hostPlatform.isDarwin then p else "Darwin only";
+    linuxOnly = p: if stdenv.hostPlatform.isLinux then p else "Linux only";
     unpackaged = "unpackaged";
   };
 in
@@ -115,7 +116,7 @@ in
     inherit (ocamlPackages) ocp-indent;
     odinfmt = ols;
     # TODO: added 2026-02-08 broken on darwin
-    opa_fmt = if stdenv.isDarwin then states.broken open-policy-agent else open-policy-agent;
+    opa_fmt = states.linuxOnly open-policy-agent;
     perltidy = perlPackages.PerlTidy;
     pg_format = pgformatter;
     phpcbf = php84Packages.php-codesniffer;
@@ -126,9 +127,9 @@ in
     inherit (python313Packages) python-ly;
     qmlformat = libsForQt5.qt5.qtdeclarative;
     # TODO: added 2026-02-08 marked unsupported on darwin
-    racketfmt = if stdenv.hostPlatform.isLinux then racket else states.broken racket;
+    racketfmt = states.linuxOnly racket;
     inherit (python313Packages) reorder-python-imports;
-    roc = if stdenv.isDarwin then states.broken roc else roc;
+    roc = states.linuxOnly roc;
     ruff_fix = ruff;
     ruff_format = ruff;
     ruff_organize_imports = ruff;


plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -23,7 +23,7 @@ rec {
     name:
     let
       permittedNames = lib.optionals (lib.isAttrs configuredFormatters) (attrNames configuredFormatters);
-      stateList = map (state: lib.fix (lib.toFunction state)) (attrValues states);
+      stateList = filter isString (map (state: (lib.toFunction state) null) (attrValues states));
```
MattSturgeon: I don't follow from reading the diff why we need to change `lib.fix` to instead always supply `null`?

Also, this is growing to the point where it may be worth considering `lib.pipe`.

plugins/by-name/conform-nvim/formatter-packages.nix
```diff
@@ -6,7 +6,7 @@ with pkgs;
 let
   states = {
     broken = _package: "broken";
-    darwinOnly = _package: "Darwin only";
+    darwinOnly = package: if stdenv.hostPlatform.isDarwin then package else "Darwin only";
```
MattSturgeon: I wonder if it is more optimised to do:


```suggestion
    darwinOnly = if stdenv.hostPlatform.isDarwin then lib.id else (_: "Darwin only");
```

I.e. evaluate whether this is darwin earlier 

?

tests/test-sources/plugins/by-name/conform-nvim/default.nix
```diff
@@ -11,7 +11,9 @@
         formatter-packages
         states
         ;
-      stateList = map (state: lib.fix (lib.toFunction state)) (builtins.attrValues states);
+      stateList = builtins.filter builtins.isString (
```
MattSturgeon: Doesn't `builtins.toString` trigger nixd's linting (and fail our treefmt)? I see treefmt passed, so maybe we have that lint disabled? Still, I'd rather not introduce new cases of it unnecessarily.

(Also same feedback about `fix` and `pipe`)

tests/test-sources/plugins/by-name/conform-nvim/default.nix
```diff
@@ -31,6 +33,47 @@
...
+            ];
+      };
+
+      assertions = [
+        {
+          assertion =
+            if isDarwin then
+              lib.elem pkgs.swift config.extraPackages
+            else
+              !lib.elem pkgs.swift config.extraPackages;
```
MattSturgeon: This can be simplified to something like:


```suggestion
            isDarwin == lib.elem pkgs.swift config.extraPackages;
```

tests/test-sources/plugins/by-name/conform-nvim/default.nix
saygo-png: builitins.isString is not exported to the global scope so it's not linted. You can see the list of globals here https://nix.dev/manual/nix/2.33/language/builtins.html#built-ins

As for the PR overall I'll review and test it in more detail tommorow

plugins/by-name/conform-nvim/auto-install.nix
saygo-png: We need to change lib.fix as this would do infinite recursion otherwise. lib.fix provides "self" as an argument.
image

tests/test-sources/plugins/by-name/conform-nvim/default.nix
MattSturgeon: Ah yes. I confused it with `toString`

khaneliman:
> I find your description of the problem confusing since the warning does show on all systems, it just doesn't ever provide the package — even on Darwin systems. Other than that LGTM.

that's the entirety of the problem, the warning shows on darwin and doesn't install on darwin.

plugins/by-name/conform-nvim/formatter-packages.nix
khaneliman: Does this improve eval performance noticably? I find lib.id usage harder to read than explicit `package` argument.

MattSturgeon: `(p: p)` would also work, but no strong opinion. I'm not gonna benchmark it! 😅

tests/test-sources/plugins/by-name/conform-nvim/default.nix
```diff
@@ -11,7 +11,11 @@
         formatter-packages
         states
         ;
-      stateList = map (state: lib.fix (lib.toFunction state)) (builtins.attrValues states);
+      stateList = lib.pipe states [
+        builtins.attrValues
+        (map (state: (lib.toFunction state) null))
```
MattSturgeon: How about:


```suggestion
        (map lib.toFunction)
        (map (fn: fn null))
```

Sometimes I think it's more readable to break pipes up into smaller steps.

---

Aside: I think this is equivalent, too:

```suggestion
        (map (state: lib.toFunction state null))
```

plugins/by-name/conform-nvim/auto-install.nix
```diff
@@ -23,7 +23,11 @@ rec {
     name:
     let
       permittedNames = lib.optionals (lib.isAttrs configuredFormatters) (attrNames configuredFormatters);
-      stateList = map (state: lib.fix (lib.toFunction state)) (attrValues states);
+      stateList = lib.pipe states [
+        attrValues
+        (map (state: (lib.toFunction state) null))
```
MattSturgeon: Same


```suggestion
        (map lib.toFunction)
        (map (fn: fn null))
```

tests/test-sources/plugins/by-name/conform-nvim/default.nix
khaneliman: Yeah, i like that more.

PR #4211, opened 2026-03-03
modules/lsp: feat add ansiblels back and add defaults for ansiblels cmd

dtvillafana:
This might not be the best way to do this, but I figured out a way to add back in ansiblels and set the cmd.

did this because I added ansible-language-server to upstream nixpkgs so that I could use it in nixvim, so I figured I'd open a PR here.

### DIFF
--- a/modules/lsp/servers/custom/ansiblels.nix
+++ b/modules/lsp/servers/custom/ansiblels.nix
@@ -1,0 +1,8 @@
+{ lib, config, ... }:
+{
+  # cmd depends on evaluated config.package, so set via module config
+  config.config.cmd = lib.mkIf (config.package != null) [
+    (lib.getExe config.package)
+    "--stdio"
+  ];
+}
--- a/modules/lsp/servers/packages.nix
+++ b/modules/lsp/servers/packages.nix
@@ -5,7 +5,6 @@
     "alloy_ls"
     "anakin_language_server"
     "antlersls"
-    "ansiblels"
     "apex_ls"
     "autohotkey_lsp"
     "awk_ls"
@@ -191,6 +190,15 @@
     aiken = "aiken";
     air = "air-formatter";
     angularls = "angular-language-server";
+    ansiblels = {
+      package = "ansible-language-server";
+      config.default = {
+        cmd = [
+          "ansible-language-server"
+          "--stdio"
+        ];
+      };
+    };
     arduino_language_server = "arduino-language-server";
     asm_lsp = "asm-lsp";
     ast_grep = "ast-grep";
--- a/modules/lsp/servers/server.nix
+++ b/modules/lsp/servers/server.nix
@@ -84,7 +84,7 @@ in
       description = ''
         Configurations for ${displayName}. ${args.config.extraDescription or ""}
       '';
-      default = { };
+      default = args.config.default or { };
       example =
         args.config.example or {
           cmd = [
--- a/modules/lsp/servers/default.nix
+++ b/modules/lsp/servers/default.nix
@@ -70,7 +70,18 @@ in
         {
           # Declare explicit options for each `packages.nix` entry with a known package
           options = builtins.mapAttrs (
-            name: package: mkServerOption { inherit name package; }
+            name: value:
+            let
+              args =
+                if builtins.isAttrs value then
+                  { inherit name; } // value
+                else
+                  {
+                    inherit name;
+                    package = value;
+                  };
+            in
+            mkServerOption args
           ) (import ./packages.nix).packages;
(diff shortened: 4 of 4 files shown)


dtvillafana:
nixpkgs needs to be updated so that ansible-language-server is available, but updating it also causes build failures.
Fixing all those is beyond my time and expertise currently.
If this isn't helpful, please leave some advice before closing the PR.

modules/lsp/servers/server.nix
```diff
@@ -84,7 +84,7 @@ in
       description = ''
         Configurations for ${displayName}. ${args.config.extraDescription or ""}
       '';
-      default = { };
+      default = args.config.default or { };
```
MattSturgeon: This is the wrong place for server-specific defaults to be set. This is the actual default value of the entire submodule, which should be `{}` in order to use the sub-option's default values.

modules/lsp/servers/custom/ansiblels.nix
```diff
@@ -0,0 +1,8 @@
+{ lib, config, ... }:
+{
```
MattSturgeon, line 2:
```suggestion
{
  options.config = lib.mkOption {
    // Use option declaration merging to set defaultText
    type = with lib.types; attrsOf anything;
    defaultText = lib.literalExpression ''
      [
        (lib.getExe config.${options.package})
        "--stdio"
      ]
   '';
  };
```

That'll render the default in docs as something like:
```nix
[
  (lib.getExe config.lsp.servers.ansiblels.package)
  "--stdio"
]
```

However, we don't need this if we drop the below config definition.

modules/lsp/servers/custom/ansiblels.nix
```diff
@@ -0,0 +1,8 @@
...
+  # cmd depends on evaluated config.package, so set via module config
+  config.config.cmd = lib.mkIf (config.package != null) [
+    (lib.getExe config.package)
```
MattSturgeon, line 5: Do we really need `getExe` here? The `lsp` module installs `config.package` onto the neovim PATH. Most (all?) servers run the LSP via the path rather than an absolute path-in-store.


```suggestion
    "ansible-language-server"
```

(if so, we can drop this entire file)

modules/lsp/servers/custom/ansiblels.nix
```diff
@@ -0,0 +1,8 @@
+{ lib, config, ... }:
+{
+  # cmd depends on evaluated config.package, so set via module config
+  config.config.cmd = lib.mkIf (config.package != null) [
```
MattSturgeon, line 4: This sets `config` and `config.cmd` at the `defaultOverridePriority` (100). That's bad because it is the same override priority typically used by end-users, so their `cmd` definition will be merged with this one into a single list.

Ideally, we want `mkOptionDefault` as that is the override priority used by `mkOption`'s `default`. `mkDefault` is also acceptable, as is `mkOverride 1499` (1 less than `mkOptionDefault`).

https://github.com/NixOS/nixpkgs/blob/e3b13232390e3c72d8d6fe1214bd406cded4d9d0/lib/modules.nix#L1577-L1581

In this case, I don't think we need this definition at all. The default you passed into `config = mkOption` already specifies `[ "ansible-language-server" "--stdio" ]` and the `lsp` module ensures that `config.package` is on neovim's PATH.

modules/lsp/servers/packages.nix
```diff
@@ -195,6 +190,15 @@
     aiken = "aiken";
     air = "air-formatter";
```
MattSturgeon, line 191: This file is currently supposed to be a simple mapping of lsp->package. Having a mixture of plain package values and default config values complicates it, and complicates our usage in modules/lsp/servers/default.nix and tests.

---

We can simplify by doing this throughout the file:

```suggestion
    air.package = "air-formatter";
```

But this has a bunch of follow-up implications:

E.g. renaming `packages` to something like `configs`, `modules`, `args`, etc. Maybe rename the file to something like `known-servers.nix` or `servers/known.nix` (requires treewide updates).

...and we could also simplify `unpackaged = [ "foo" ]` to `foo = null` and filter out nulls in modules/lsp/servers/default.nix. Maybe.

Therefore, we'd want to isolate such a refactor to a separate PR, though. Especially as it'll require treewide updates everywhere the file is imported.

MattSturgeon (commented):
Some feedback on the approach taken and some comments on how this highlights issues with Nixvim's existing architecture.

Thanks for opening the PR!

> nixpkgs needs to be updated so that ansible-language-server is available

I think this is resolved now on `main`?

PR #4219, opened 2026-03-19
fix: adapt to upstream changes in provider generation

obreitwi:
This PR adapts to upstream https://github.com/NixOS/nixpkgs/pull/498687.

The biggest upstream change is that the host progs are no longer exposed. Hence, we have to include the upstream `providerLuaRc`.

Looking for feedback as I am unsure that a) `providerLuaRc` should be included unconditionally and that b) this is the correct place to do so.

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -241,8 +241,17 @@ in
             dontFixup = true;
           };
 
+      nvimPackage = wrappedNeovim.override (prev: {
+        wrapperArgs =
+          (if lib.isString prev.wrapperArgs then prev.wrapperArgs else lib.escapeShellArgs prev.wrapperArgs)
+          + " "
+          + extraWrapperArgs;
+        wrapRc = false;
+      });
+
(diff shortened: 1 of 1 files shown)


modules/top-level/output.nix
```diff
@@ -236,8 +236,17 @@ in
...
+        wrapperArgs =
+          (if lib.isString prev.wrapperArgs then prev.wrapperArgs else lib.escapeShellArgs prev.wrapperArgs)
+          + " "
+          + extraWrapperArgs;
+        wrapRc = false;
+      });
+
       customRC = lib.nixvim.concatNonEmptyLines [
         (lib.nixvim.wrapVimscriptForLua wrappedNeovim.initRc)
+        nvimPackage.passthru.providerLuaRc
```
MattSturgeon: I assume CI is failing because our pinned nixpkgs is still behind?

Would it make sense to do `or ""` here, to be compatible with old nixpkgs revisions _and_ new ones?


```suggestion
        (nvimPackage.passthru.providerLuaRc or "")
```

PR #4226, opened 2026-03-26
plugins/edgy: fix options for size and collapsed

ReeSilva:
This PR aims to solve https://github.com/nix-community/nixvim/issues/4225 by rewriting `size` options using `mkCompositeOption` to be nullable and an attribute set with optional `width` and `height`.

I've also added an option for `collapsed`. I've tested this in my config and everything seems work.


plugins/by-name/edgy/default.nix
```diff
@@ -36,16 +36,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+              `height` is the minimum height for panels.
+            ''
+            {
+              width = lib.nixvim.mkNullOrOption types.ints.unsigned ''
+                Minimum width of the view when it's in an edgebar.
+              '';
+              height = lib.nixvim.mkNullOrOption types.ints.unsigned ''
+                Minimum height of the view when it's in a panel.
+              '';
+            };
```
GaetanLepage:
```suggestion
```
Prefer simply removing the options. The `settings` option is already typed such as you can provide any value to sub-options. This is the design pattern we now prefer.

plugins/by-name/edgy/default.nix
```diff
@@ -36,16 +36,30 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
 
         pinned = lib.nixvim.mkNullOrOption types.bool ''
           If true, the view will always be shown in the edgebar even when it has no windows.
         '';
 
+        collapsed = lib.nixvim.mkNullOrOption types.bool ''
+          When a view is pinned and collapsed, it will be shown closed on start.
+        '';
+
```
GaetanLepage:
```suggestion
```
Same here, no need to explicitly add an option.

plugins/by-name/edgy/default.nix
ReeSilva: Agreed, and done.

plugins/by-name/edgy/default.nix
ReeSilva: Agreed and done.

GaetanLepage:
While removing all option declarations is a good idea in general, we can't do that for `mkLuaFn[']` ones.
Indeed, they automatically convert the value to "raw lua". So this would be a breaking change.

Either you leave those specific declarations, or you add proper `warnings` (grep the tree to find such examples) to warn the use when they put a plain string.

ReeSilva:
@GaetanLepage Changed. Please, let me know if there is any type in the `nixvim` lib that sets what I did that I couldn't find. I followed similar approaches in `dropbar` and `colorizer`. Thanks for all the tips, btw, and wish you a great weekend.

EDIT: I notice some of my commits are not following the standard (habits, mb). Should I rewrite them and force-push, or the fact that some of them are following the guidelines is enough?

khaneliman:
> EDIT: I notice some of my commits are not following the standard (habits, mb). Should I rewrite them and force-push, or the fact that some of them are following the guidelines is enough?

Could you please cleanup the commit history.

ReeSilva:
> > EDIT: I notice some of my commits are not following the standard (habits, mb). Should I rewrite them and force-push, or the fact that some of them are following the guidelines is enough?
> 
> Could you please cleanup the commit history. 
> 

@khaneliman Would a new branch from `main` cherry-pickingmy changes be enough?

khaneliman:
> > > EDIT: I notice some of my commits are not following the standard (habits, mb). Should I rewrite them and force-push, or the fact that some of them are following the guidelines is enough?
> > 
> > 
> > Could you please cleanup the commit history.
> 
> @khaneliman Would a new branch from `main` cherry-pickingmy changes be enough?

You can just squash all the commits with an interactive reabse, otherwise i can do it if you need help with it.

ReeSilva:
Well, it clearly didn't work and I made things worst, sorry. Should not be commiting during the night hehe, mb, I'll fix it tomorrow

ReeSilva:
I'll close this one and open another one, it become a mess. Sorry everyone who got accidentally notified.

PR #4235, opened 2026-04-08
fix: manpage NAME to use lowercase nixvim

bodzioney:
## Summary
  - Fix the manpage header to use the standard `name \- description` format in the NAME section
  - Previously, the NAME section used freeform text (`Nixvim options specification`), causing tools like fish's autocomplete to register the command as `Nixvim` instead of `nixvim`
  - Also dropped the redundant 5th `.TH` argument to match home-manager's manpage convention
  
  Fixes #4233


GaetanLepage (approved):
Thanks :)

PR #4236, opened 2026-04-10
plugins/sidekick: don't bundle providers default

### LINKED ISSUE #4218: Reconsider default agent packages for nvim-sidekick
Should we really default to including all these in the sidekick plugin?

https://github.com/nix-community/nixvim/blob/21ae25e13b01d3b4cdc750b5f9e7bad68b150c10/plugins/by-name/sidekick/default.nix#L14-L17

Isn't it safer to simply default to none and let users decide themselves?

khaneliman:
Default to no providers installed, opt-in vs opt-out to reduce closure size by default. Most of these require subscriptions or familiarity with anyways and doesn't make sense to default enable everything.

Closes https://github.com/nix-community/nixvim/issues/4218

Technically a breaking change, don't want to force explicit acknowledgement though for a list of providers to be explicitly disabled/enabled...

### DIFF
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -10,12 +10,19 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
   maintainers = [ lib.maintainers.khaneliman ];
 
-  dependencies = [
-    "claude-code"
-    "copilot"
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
I agree!

PR #4243, opened 2026-04-12
plugins/gitlab: use dependencies.nodejs.path as default for code_suggestions.lsp_binary_path

GaetanLepage:


plugins/by-name/gitlab/default.nix
```diff
@@ -84,7 +92,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
           The path where the `node` executable is available.
 
           By default, this option is set to `"node"` which will look for nodejs in your `$PATH`.
-          To ensure that `node` will always be in your `$PATH`, you may set the `nodePackage` option.
+          To ensure that `node` will always be in your `$PATH`, you may set the `dependencies.nodejs.enable` to `true`.
```
khaneliman: Curious, what does this generate when `config.dependencies.nodejs.enable = false`? Does it still contain the derivation accessible and would just install it in the settings anyways? Looks like package already defaults it enabled.

GaetanLepage: You are talking about the `mkDefault` in `extraConfig`? This one is conditioned on `dependencies.nodejs.enable`.

khaneliman: Ah alright. But the default is enabled so messaging is a little confusing, it seems like a user might need to take action. Wonder if we just word it differently based on that understanding that `nodejs` is automatically included and configured and they can disable it.

khaneliman (approved):

plugins/by-name/gitlab/default.nix
GaetanLepage: Indeed, I changed the description.

PR #4253, opened 2026-04-21
ci: lix parse; treewide: fix incorrect string escapes

### LINKED ISSUE #4251: [STYLE] `lix` 2.95 warns on many `broken-string-escape`s
Running `lix` 2.95 on a configuration warns many times on broken string escapes (escapes where the character escaped is returned directly, e.g. `"\." == "."`):

```log
...
warning: \< is an ill-defined escape. You can drop the \ and simply write < instead. Use --extra-deprecated-features broken-string-escape to silence this warning.
         at /nix/store/ab62cm07rw12mvpw307y7cl7867crb9y-source/plugins/by-name/vim-matchup/default.nix:296:21:
            295|       pluginDefault = "";
            296|       example = "vV\<c-v>";
               |                     ^
            297|       description = ''
warning: \{ is an ill-defined escape. You can drop the \ and simply write { instead. If you meant to escape an interpolation, write \${ instead. Use --extra-deprecated-features broken-string-escape to silence this warning.
         at /nix/store/ab62cm07rw12mvpw307y7cl7867crb9y-source/plugins/by-name/dap/dapHelpers.nix:58:68:
             57|
             58|     port = mkNullOrOption (types.either types.int (types.enum [ "$\{port}" ])) ''
               |                                                                    ^
             59|       Port to connect to.
warning: \{ is an ill-defined escape. You can drop the \ and simply write { instead. If you meant to escape an interpolation, write \${ instead. Use --extra-deprecated-features broken-string-escape to silence this warning.
         at /nix/store/ab62cm07rw12mvpw307y7cl7867crb9y-source/plugins/by-name/dap/dapHelpers.nix:81:48:
             80|   pipeAdapterOption = mkAdapterType {
             81|     pipe = lib.nixvim.defaultNullOpts.mkStr "$\{pipe}" "Pipe name.";
               |                                                ^
             82|
...
```

khaneliman:
Done with help of lix parsing for warnings.

Closes https://github.com/nix-community/nixvim/issues/4251 

Related https://github.com/nix-community/home-manager/pull/9150 https://github.com/NixOS/nixpkgs/pull/511836


ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+    export XDG_CONFIG_HOME=$TMPDIR
+    export NIX_CONF_DIR=$TMPDIR/nix-conf
+    mkdir -p "$NIX_CONF_DIR"
+    : > "$NIX_CONF_DIR/nix.conf"
+
+    export NIX_STORE_DIR=$TMPDIR/store
+    export NIX_STATE_DIR=$TMPDIR/state
+    nix-store --init
+
+    cd "${nixvim}"
```
MattSturgeon: Nit: putting quotes around a string interpolation is a bit of an anti-pattern imo.

Either the string is safe to interpolate and doesn't need quoting (most nix paths are safe), or you should use a proper escaping mechanism like `toShellVars`, `escapeShellArg`, or implicit drvAttrs→shellVars.

Wrapping with quotes gives the illusion of safety without addressing many interpolation pitfalls.

ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+  pipeOperatorFlag = if lib.hasPrefix "lix-" nix.name then "pipe-operator" else "pipe-operators";
+in
+runCommand "nix-parse-${nix.name}"
+  {
+    nativeBuildInputs = [ nix ];
+  }
+  ''
+    export HOME=$TMPDIR
+    export XDG_CONFIG_HOME=$TMPDIR
+    export NIX_CONF_DIR=$TMPDIR/nix-conf
```
MattSturgeon: We should always quote shell variables in case they contain whitespace 


```suggestion
    export HOME="$TMPDIR"
    export XDG_CONFIG_HOME="$TMPDIR"
    export NIX_CONF_DIR="$TMPDIR/nix-conf"
```

ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+    mkdir -p "$NIX_CONF_DIR"
+    : > "$NIX_CONF_DIR/nix.conf"
+
+    export NIX_STORE_DIR=$TMPDIR/store
+    export NIX_STATE_DIR=$TMPDIR/state
```
MattSturgeon:
```suggestion
    export NIX_STORE_DIR="$TMPDIR/store"
    export NIX_STATE_DIR="$TMPDIR/state"
```

ci/parse.nix
MattSturgeon: In this case I'd recommend defining `nixvim` as a derivation attr, then doing `cd "$nixvim"`.

ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+  runCommand,
+}:
+let
+  nixvim =
+    with lib.fileset;
+    toSource {
+      root = ../.;
+      fileset = fileFilter (file: file.hasExt "nix") ../.;
+    };
+  parseLog = "$TMPDIR/nix-parse.log";
```
MattSturgeon: Why does this need to be outside of the drv? I'd replace this nix binding with a shell variable for consistency and simplicity

ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+    export NIX_STATE_DIR=$TMPDIR/state
+    nix-store --init
+
+    cd "${nixvim}"
+
+    # This will only show the first parse error, not all of them. That's fine, because
+    # the other CI jobs will report in more detail. This job is about checking parsing
+    # across different implementations / versions, not about providing the best DX.
+    # Returning all parse errors requires significantly more resources.
+    if ! find . -type f -iname '*.nix' -print0 | xargs -0 -P "$(nproc)" nix-instantiate --extra-experimental-features "${pipeOperatorFlag}" --parse > /dev/null 2> "${parseLog}"; then
```
MattSturgeon: Good call with the lix vs nix pipe operators experimental flags. Same nit with the misleading quotes, though:


```suggestion
    if ! find . -type f -iname '*.nix' -print0 | xargs -0 -P "$(nproc)" nix-instantiate --extra-experimental-features ${pipeOperatorFlag} --parse > /dev/null 2> "$parseLog"; then
```

(And assuming `$parseLog` becomes a shell variable)

flake/dev/parse.nix
```diff
@@ -0,0 +1,14 @@
...
+    { pkgs, ... }:
+    {
+      checks = {
+        parseNix = pkgs.callPackage ../../ci/parse.nix {
+          nix = pkgs.nixVersions.latest;
+        };
+        parseLix = pkgs.callPackage ../../ci/parse.nix {
+          nix = pkgs.lixPackageSets.latest.lix;
+        };
+      };
```
MattSturgeon, line 14: Nice.

If this grows, then it may be nice to `mapAttrs` or `genAttrs` a name+package combo to reduce repetition.

On the other hand, simple and repetitive is sometimes best. Simple is certainly better for now.

tests/test-sources/plugins/by-name/dropbar/default.nix
```diff
@@ -658,7 +658,7 @@
           };
           treesitter = {
             max_depth = 16;
-            name_regex = "[=[[#~!@\*&.]*[[:keyword:]]\+!\?\(\(\(->\)\+\|-\+\|\.\+\|:\+\|\s\+\)\?[#~!@\*&.]*[[:keyword:]]\+!\?\)*]=]";
+            name_regex = "[=[[#~!@\\*&.]*[[:keyword:]]\\+!\\?\\(\\(\\(->\\)\\+\\|-\\+\\|\\.\\+\\|:\\+\\|\\s\\+\\)\\?[#~!@\\*&.]*[[:keyword:]]\\+!\\?\\)*]=]";
```
MattSturgeon: Well done gleaning the original author's intent here...

Do we know if the regex is valid (before or after)?

tests/test-sources/plugins/lsp/language-servers/lua-ls.nix
```diff
@@ -226,7 +226,7 @@
               utf8 = "default";
             };
             fileEncoding = "utf8";
-            meta = "$\{version} $\{language} $\{encoding}";
+            meta = "\${version} \${language} \${encoding}";
```
MattSturgeon, line 229: I wonder why we had so many `$\{foo}`s?

ci/parse.nix
```diff
@@ -0,0 +1,50 @@
...
+    export NIX_STORE_DIR=$TMPDIR/store
+    export NIX_STATE_DIR=$TMPDIR/state
+    nix-store --init
+
+    cd "${nixvim}"
+
+    # This will only show the first parse error, not all of them. That's fine, because
+    # the other CI jobs will report in more detail. This job is about checking parsing
+    # across different implementations / versions, not about providing the best DX.
+    # Returning all parse errors requires significantly more resources.
```
MattSturgeon: How would this look if we tried to use semantic wrapping?


```suggestion
    # This will only show the first parse error, not all of them.
    # That's fine, because the other CI jobs will report in more detail.
    # This job is about checking parsing across different implementations / versions, not about providing the best DX.
    # Returning all parse errors requires significantly more resources.
```

The third sentence ends up rather long when wrapped semantically...

.github/workflows/lint.yml
```diff
@@ -7,7 +7,6 @@ on:
       - "**.svg"
       - ".gitignore"
       - "LICENSE"
-      - "flake.lock"
```
MattSturgeon, line 10: Why?

MattSturgeon (commented):
Couple nits w.r.t. nix/shell variable hygiene. Otherwise sgtm.

Thanks for putting this together!

.github/workflows/lint.yml
MattSturgeon, line 16: This job should probably be renamed now.

Maybe `lint` is ok, but that'll show up in the UI as `lint / lint (pull_request)`.

MattSturgeon: This makes me think... Would `nix-instantiate --parse` make sense as a treefmt step? We'd need to wrap it in a treefmt-compliant tool of course, and have lix+nix permutations still. Maybe an idea for another day.

.github/workflows/lint.yml
khaneliman: So we can run the parse checks against changed nix/lix versions.

tests/test-sources/plugins/lsp/language-servers/lua-ls.nix
khaneliman: Harder to spot just trying to get lsp / eval happy

tests/test-sources/plugins/by-name/dropbar/default.nix
khaneliman: Didn't test it, just saw the upstream default https://github.com/Bekaboo/dropbar.nvim/blob/ce202248134e3949aac375fd66c28e5207785b10/lua/dropbar/configs.lua#L718-L719 that was used when creating the module and I am not sure if it's correct...

khaneliman: I think we might have a form that is lua syntax in a regular string and could need fixing for real usage.

khaneliman: I'll push a commit for fixing that. 

Here's a test script 

```lua
local upstream = [=[[#~!@\*&.]*[[:keyword:]]\+!\?]=]
  .. [=[\(\(\(->\)\+\|-\+\|\.\+\|:\+\|\s\+\)\?[#~!@\*&.]*[[:keyword:]]\+!\?\)*]=]

local wrapped =
  "[=[[#~!@\\*&.]*[[:keyword:]]\\+!\\?"
  .. "\\(\\(\\(->\\)\\+\\|-\\+\\|\\.\\+\\|:\\+\\|\\s\\+\\)\\?"
  .. "[#~!@\\*&.]*[[:keyword:]]\\+!\\?\\)*]=]"

local fixed =
  "[#~!@\\*&.]*[[:keyword:]]\\+!\\?"
  .. "\\(\\(\\(->\\)\\+\\|-\\+\\|\\.\\+\\|:\\+\\|\\s\\+\\)\\?"
  .. "[#~!@\\*&.]*[[:keyword:]]\\+!\\?\\)*"

local samples = {
  "foo",
  "foo.bar",
  "foo->bar",
  "#foo!",
  "foo bar baz",
  "x::y",
  "foo-bar",
}

for _, s in ipairs(samples) do
  local a = vim.fn.matchstr(s, upstream)
  local b = vim.fn.matchstr(s, wrapped)
  local c = vim.fn.matchstr(s, fixed)
  print(
    s
      .. " | upstream="
      .. vim.inspect(a)
      .. " | wrapped="
      .. vim.inspect(b)
      .. " | fixed="
      .. vim.inspect(c)
      .. " | wrapped_same="
      .. tostring(a == b)
      .. " | fixed_same="
      .. tostring(a == c)
  )
end

vim.cmd("qa")
```


```bash
nvim --headless -u NONE -S /tmp/check-dropbar.lua
foo | upstream="foo" | wrapped="" | fixed="foo" | wrapped_same=false | fixed_same=true
foo.bar | upstream="foo.bar" | wrapped="" | fixed="foo.bar" | wrapped_same=false | fixed_same=true
foo->bar | upstream="foo->bar" | wrapped="" | fixed="foo->bar" | wrapped_same=false | fixed_same=true
#foo! | upstream="#foo!" | wrapped="" | fixed="#foo!" | wrapped_same=false | fixed_same=true
foo bar baz | upstream="foo bar baz" | wrapped="" | fixed="foo bar baz" | wrapped_same=false | fixed_same=true
x::y | upstream="x::y" | wrapped="" | fixed="x::y" | wrapped_same=false | fixed_same=true
foo-bar | upstream="foo-bar" | wrapped="" | fixed="foo-bar" | wrapped_same=false | fixed_same=true%
```

MattSturgeon (approved):
Awesome thanks

PR #4255, opened 2026-04-22
CONTRIBUTING: explain common test attrset names

### LINKED ISSUE #3572: [docs/contributing] List the purpose of different test attrsets.
Currently the test section in the contributing section does not mention the purpose of attrsets like `default` `empty` `example` or `no-packages`.
The names are not self explanatory, [the difference between "default" and "empty" is unclear](https://github.com/nix-community/nixvim/discussions/3569).
Same with the difference between "empty" and "no-packages"

khaneliman:
Closes #3572

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -300,6 +300,13 @@ The tests are located in the [tests/test-sources](tests/test-sources) directory,
 
 Tests can either be a simple attribute set, or a function taking `{pkgs}` as an input. The keys of the set are configuration names, and the values are a Nixvim configuration.
 
+These configuration names are just conventions, not special values interpreted by the test runner. Common ones include:
+
+- `defaults`: explicitly sets the plugin or module to the default values documented by Nixvim's settings options. This is mainly useful when those defaults are part of Nixvim's option interface; otherwise `empty` is usually enough.
+- `empty`: enables the feature with no additional configuration.
+- `example`: uses a configuration that differs from the defaults and reflects something a user might realistically want to do.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -300,6 +300,13 @@ The tests are located in the [tests/test-sources](tests/test-sources) directory,
 
 Tests can either be a simple attribute set, or a function taking `{pkgs}` as an input. The keys of the set are configuration names, and the values are a Nixvim configuration.
 
+These configuration names are just conventions, not special values interpreted by the test runner. Common ones include:
+
+- `defaults`: explicitly sets the plugin or module to its upstream/default values so those defaults are covered by a test.
```
MattSturgeon: We could try to hint at `defaults` tests mainly existing to validate defaults "documented" by our settings options.

I don't see much value in these "explicit defaults" tests in modules that _don't_ have corresponding settings options. In that scenario, they're just a more brittle version of the `empty` test.

So maybe we could also state something like "only relevant when nixvim's options explicitly document these defaults"?

MattSturgeon (approved):
Nice

CONTRIBUTING.md
khaneliman: Hmm yeah, I can see that. As we've less time to micro manage the ci / tests / repo. It does create more friction when we have those test cases, but sometimes I appreciate the heads up from our ci that the upstream plugin has changed when there is a deprecation/removal of configuration.

khaneliman: I'll use that more conservative wording for the contributing docs to not encourage that extra maintenance debt though.

PR #4257, opened 2026-04-22
plugins/lint: add auto-install support

### LINKED ISSUE #3696: [Feature] Automatically install linters
Currently to add a linter you have to manually add its package after enabling it in the lint plugin. Would be nice if this was automatic like in the case of LSPs.

khaneliman:
Closes https://github.com/nix-community/nixvim/issues/3696

Lot of copy paste with minor tweaks from conform implementation.


MattSturgeon (approved):
I didn't look super close.

If the code is similar to existing code I'll try to avoid ~~looking too close~~ nit picking 😁 

Seems well tested and well thought out. LGTM.

PR #4258, opened 2026-04-22
lib/plugins: add optional setup plumbing; add public callSetup override

### LINKED ISSUE #3959: [Feature] option to disable calling `setup`
It is not necessary to call the setup for every plugin,
so is it possible to add a boolean to control this in nixvim?

khaneliman:
Closes https://github.com/nix-community/nixvim/issues/3959

Adding the optional `callSetup` functionality to allow us to respect modern neovim plugin direction of migrating away from explicit setup calls. Added an internal mechanism that we can use on plugins we know support it, while giving a public api option to allow a user to override the setup call themself. 

Have a fixup commit related to the discussion about whether we should be over protective or allow advanced usage.


GaetanLepage (approved):
Thanks. Good job on the testing too!

PR #4270, opened 2026-04-25
plugins/lz-n: support importing plugin specs

khaneliman:


GaetanLepage (approved):

plugins/pluginmanagers/lz-n.nix
```diff
@@ -144,6 +144,19 @@ lib.nixvim.plugins.mkNeovimPlugin {
       };
     in
     {
+      imports = mkOption {
```
MattSturgeon, line 147: The name `imports` would conflict if we were ever to make plugins themselves into submodules.

I think that's acceptable here.

MattSturgeon (approved):

PR #4287, opened 2026-05-12
mailmap: map NAHO developer identity to Noah Biewesch

trueNAHO:
Map my NAHO developer identity to Noah Biewesch to correct existing contributions, which would also be valuable to align with future contributions.

This PR is part of a coordinated effort to canonicalize my developer identity:

- https://github.com/NixOS/nixpkgs/pull/519557
- https://github.com/nix-community/home-manager/pull/9308
- https://github.com/nix-community/nixvim/pull/4287
- https://github.com/nix-community/stylix/pull/2309

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Noah Biewesch <dev@noahbiewesch.com> <90870942+trueNAHO@users.noreply.github.com>


MattSturgeon (commented):
This will allow `git log` to recognise both emails as the same author.

This doesn't seem _critical_. I'm not convinced it's worth the bytes in tree, but I'm won't block the PR yet.

It begs the question, would we accept an entry here for every contributor who has used multiple emails? Even GitHub anonymised emails?

You can already associate both emails with your GitHub account, by simply registering multiple emails (privately) in your account settings. I assume your goals go beyond that?

MattSturgeon (approved):
My main hesitation is whether we want to maintain a mailmap file _at all_.

Seeing the precident for how nixpkgs has used a mailmap file has shifted my perspective a little, so I'm happy to approve.

trueNAHO:
> It begs the question, would we accept an entry here for every contributor who has used multiple emails? Even GitHub anonymised emails?
>
> You can already associate both emails with your GitHub account, by simply registering multiple emails (privately) in your account settings. I assume your goals go beyond that?

The `.mailmap` introduction is heavily inspired by https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.mailmap. While the Linux Kernel also uses the Git history for its mailing list, the value here is that contacting me is possible privately outside GitHub, which is impossible with GitHub's private email feature.

> This doesn't seem _critical_. I'm not convinced it's worth the bytes in tree, but I'm won't block the PR yet.

IMHO, there are many better ways to "optimize" the archived tree size, like condensing the documentation and code, although this mostly seems like wasted effort to me. Since the purpose of Git is collaboration, these functionally useless bytes seem extremely valuable to me.

PR #4290, opened 2026-05-13
plugins/jj-nvim: init

sportshead:
Add [jj.nvim](https://github.com/NicolasGB/jj.nvim)


MattSturgeon (approved):
Thanks

khaneliman:
I was expecting an options eval error when this merged and realized it was `jj-nvim` instead of `jj`. I thought we were stripping `-nvim` from plugin names in our modules. Thoughts on a quick rename?

PR #4291, opened 2026-05-14
plugins/lazy: allow `pkg` option in `pluginType` to be null.

my7h3le:
- It's often required to define a spec that doesn't define a `pkg`, for example in the case of import statements.

```Nix
plugins = with pkgs.vimPlugins; [
  {
    pkg = LazyVim;
    import = "lazyvim.plugins";
    dependencies = [ trouble-nvim ];
  }
  # No `pkg` defined here as it's solely an import statement.
  { import = "lazyvim.plugins.extras.ai.copilot-chat"; }
];
```

In the case of import statements, often they can be defined with a plugin spec that includes a `pkg`, for example:
```Nix
plugins = with pkgs.vimPlugins; [
  {
    pkg = LazyVim;
    import = "lazyvim.plugins";
  }
];
```
But this can't always be done and can break functionality with some packages if done so, and as such import statements need to be able to be defined without explicitly having to define `pkg`.

### DIFF
--- a/plugins/pluginmanagers/lazy.nix
+++ b/plugins/pluginmanagers/lazy.nix
@@ -52,10 +52,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
               options = {
                 dir = mkNullOrOption str "A directory pointing to a local plugin";
 
-                pkg = mkOption {
-                  type = package;
-                  description = "Vim plugin to install";
-                };
+                pkg = mkNullOrOption package "Vim plugin to install";
 
                 name = mkNullOrOption str "Name of the plugin to install";
 
@@ -158,11 +155,13 @@ lib.nixvim.plugins.mkNeovimPlugin {
                   name = "${lib.getName p}";
                   path = p;
                 }
-              else
+              else if (p.pkg or null) != null then
                 {
-                  name = "${lib.getName p.pkg}";
+                  name = if (p.name or null) != null then p.name else lib.getName p.pkg;
                   path = p.pkg;
-                };
+                }
+              else
+                null;
             processDependencies =
               if plugin ? dependencies && plugin.dependencies != null then
                 builtins.concatMap processPlugin plugin.dependencies
@@ -171,37 +170,26 @@ lib.nixvim.plugins.mkNeovimPlugin {
           in
           [ (mkEntryFromDrv plugin) ] ++ processDependencies;
 
-        processedPlugins = builtins.concatLists (map processPlugin cfg.plugins);
+        processedPlugins = lib.remove null (builtins.concatMap processPlugin cfg.plugins);
         lazyPath = pkgs.linkFarm "lazy-plugins" processedPlugins;
 
         pluginToLua =
           plugin:
           if lib.isDerivation plugin then
-            { dir = "${lazyPath}/${lib.getName plugin}"; }
-          else
-            plugin
+            {
+              dir = "${plugin}";
+              name = "${lib.getName plugin}";
+            }
+          else if (plugin.pkg or null) != null then
+            let
+              name = if (plugin.name or null) != null then plugin.name else lib.getName plugin.pkg;
+            in
+            # Remove `pkg` from the plugin spec, since it's only used to
+            # determine the plugin directory and is not needed in the final
+            # spec.
+            removeAttrs plugin [ "pkg" ]
             // {
-              "__unkeyed" = plugin.name;
-
-              inherit (plugin)
-                cmd
-                cond
-                config
-                dev
-                enabled
-                event
-                ft
-                init
-                keys
-                lazy
-                main
-                module
-                name
-                optional
-                opts
-                priority
-                submodules
-                ;
+              "__unkeyed" = name;
 
               dependencies = lib.nixvim.ifNonNull' plugin.dependencies (
                 if lib.isList plugin.dependencies then
@@ -210,9 +198,10 @@ lib.nixvim.plugins.mkNeovimPlugin {
                   plugin.dependencies
               );
 
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
-            };
+              dir = if (plugin.dir or null) != null then plugin.dir else "${lazyPath}/${name}";
+            }
+          else
+            plugin;
 
         pluginListToLua = map pluginToLua;
 
--- a/tests/test-sources/plugins/pluginmanagers/lazy.nix
+++ b/tests/test-sources/plugins/pluginmanagers/lazy.nix
@@ -24,6 +24,10 @@
           ];
         }
 
+        # Test freeform and case when `pkg` is null.
+        { pkg = LazyVim; }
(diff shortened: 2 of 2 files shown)


plugins/pluginmanagers/lazy.nix
```diff
@@ -148,60 +144,20 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
           if lib.isDerivation plugin then
-            { dir = "${lazyPath}/${lib.getName plugin}"; }
-          else
+            {
+              dir = "${plugin}";
+              name = lib.getName plugin;
+            }
+          else if plugin ? pkg && plugin.pkg != null then
+            let
+              name = if plugin ? name && plugin.name != null then plugin.name else lib.getName plugin.pkg;
```
MattSturgeon: Existing (moved) code, but can be simplified:

```suggestion
              name = if plugin.name or null != null then plugin.name else lib.getName plugin.pkg;
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -210,9 +166,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
 
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
-            };
+              # Remove `pkg` from the plugin spec, since it's only used to
+              # determine the plugin directory and is not needed in the final
+              # spec.
+              pkg = null;
+
+              dir = if plugin ? dir && plugin.dir != null then plugin.dir else "${plugin.pkg}";
```
MattSturgeon: Same
```suggestion
              dir = if plugin.dir or null != null then plugin.dir else "${plugin.pkg}";
```

plugins/pluginmanagers/lazy.nix
```diff
@@ -210,9 +166,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
                   plugin.dependencies
               );
 
-              dir =
-                if plugin ? dir && plugin.dir != null then plugin.dir else "${lazyPath}/${lib.getName plugin.pkg}";
-            };
+              # Remove `pkg` from the plugin spec, since it's only used to
+              # determine the plugin directory and is not needed in the final
+              # spec.
+              pkg = null;
```
MattSturgeon: Why not use an actual `removeAttrs`?

I.e. `removeAttrs plugin [ "pkg" ]` on L158

plugins/pluginmanagers/lazy.nix
```diff
@@ -148,60 +144,20 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-            processDependencies =
-              if plugin ? dependencies && plugin.dependencies != null then
-                builtins.concatMap processPlugin plugin.dependencies
-              else
-                [ ];
-          in
-          [ (mkEntryFromDrv plugin) ] ++ processDependencies;
-
-        processedPlugins = builtins.concatLists (map processPlugin cfg.plugins);
-        lazyPath = pkgs.linkFarm "lazy-plugins" processedPlugins;
```
MattSturgeon, line 175: I don't fully follow why we're able to remove this? Where has the logic moved to or why isn't it needed anymore?

my7h3le: So this section is essentially doing a link farm and creating a directory path in the nix store that contains all the included plugins, it's then passing this directory path to lazy.nvim's dev.path option https://lazy.folke.io/configuration. To lazy.nvim it'll just look  like a directory that contains local plugins, and the existing code would define plugin paths relative to that dev path, so symbolically each nix vim plugin package would evaluate to ` { dir = <lazyPath>/path-to-plugin> }`.

This isn't necessary and it isn't necessary to define lazy.nvim's dev.path option either, as each nix vim plugin package can be simply evaluated with `"${pkg}"` and that nix store path can be passed directly to the `dir` option for any given plugin spec. So symbolically `{dir = "${plugin-pkg}"}`

plugins/pluginmanagers/lazy.nix
my7h3le: Good call, I actually wasn't aware that function existed

plugins/pluginmanagers/lazy.nix
my7h3le: Oooh I wasn't aware of the `or` operator https://nix.dev/manual/nix/2.34/language/operators#attribute-selection . I was initially very confused by the latter half until I added brackets `if (plugin.name or null) != null then`.

plugins/pluginmanagers/lazy.nix
MattSturgeon: My understanding was that it was needed to collect transitive dependencies? You'll notice that the link farm is not just of user-defined plugins, but also their dependencies.

How are transitive plugin dependencies handles now?

I also vaguely recall there being some performance benefit to having a single "meta" linkfarm-plugin; less runtimepath lookups maybe? I could be wrong here though, and we have a separate "combine plugins" performance option anyway.

Regardless, this seems like a tangential change to the core change of this PR. Is it worth extracting it to its own commit, before the main change?

my7h3le: So transitive plugin dependencies were already handled by pluginToLua https://github.com/nix-community/nixvim/blob/06cace835d7ee727852ac789e3dcd42fc2fd360e/plugins/pluginmanagers/lazy.nix#L206-L211

the `processDependencies` function is just to add dependencies to the link farmed nix store directory path. https://github.com/nix-community/nixvim/blob/06cace835d7ee727852ac789e3dcd42fc2fd360e/plugins/pluginmanagers/lazy.nix#L151-L172

My evaluated `init.lua` for example still has dependencies evaluated properly: 
```lua
        {
            "CopilotChat.nvim",
            build = "make tiktoken",
            dependencies = {
                {
                    dir = "/nix/store/il6y7c92wpnf08ppk8q6izngajnzvra3-vimplugin-luajit2.1-plenary.nvim-scm-1-unstable-scm-1",
                    name = "plenary.nvim",
                },
            },
            dir = "/nix/store/p51dgvdbrvsfd5ddm08w4dhn5yj19nkk-vimplugin-CopilotChat.nvim-4.7.4-unstable-2026-04-26",
            opts = { chat_autocomplete = false },
        },
```

As for perfomance benefits, that I'm unsure of and it's not something I could find in lazy.nvim's docs `https://lazy.folke.io` . I imagine it won't make a difference because we still need to define `dir` in `pluginToLua`  but I'm not entirely sure myself either. https://github.com/nix-community/nixvim/blob/06cace835d7ee727852ac789e3dcd42fc2fd360e/plugins/pluginmanagers/lazy.nix#L180 https://github.com/nix-community/nixvim/blob/06cace835d7ee727852ac789e3dcd42fc2fd360e/plugins/pluginmanagers/lazy.nix#L213-L214

I do agree that it kinda falls outside of the scope of the core change of this PR though, and it's more of a refactor. I've changed the commit to make use of the existing link farm functionality. Not sure if the commit to remove link farm and `processPlugins` should be it's own PR though.

MattSturgeon (approved):
SGTM. Let's give it a try. Thanks!

PR #4293, opened 2026-05-14
allow raw lua in blink-cmp keymap via lib.nixvim.mkRaw

DasGruene:
Allows to define raw lua commands to the keymap section of blink-cmp.

as im quite new to coding in nix, im not sure I've used the best approach, please do fell free to come with constructive suggestions.

Note: personally needed this to add code completion through minuet

### DIFF
--- a/plugins/by-name/blink-cmp/settings-options.nix
+++ b/plugins/by-name/blink-cmp/settings-options.nix
@@ -75,7 +75,8 @@ in
         "snippet_backward"
         "fallback"
       ];
-
+      # Example using raw Lua:
+      "<C-a>" = lib.nixvim.mkRaw "require('some-module').custom_function()";
     };
     description = ''
       The keymap can be:
@@ -83,6 +84,9 @@ in
       - A table of `keys => command[]` (optionally with a "preset" key to merge with a preset)
       When specifying 'preset' in the keymap table, the custom key mappings are merged with the preset,
       and any conflicting keys will overwrite the preset mappings.
+
+      For advanced use cases, you can use raw Lua with `lib.nixvim.mkRaw` to define custom
+      keymap logic.
     '';
   };
 


plugins/by-name/blink-cmp/settings-options.nix
```diff
@@ -28,7 +28,7 @@ in
   };
 
   keymap = defaultNullOpts.mkNullableWithRaw' {
-    type = with types; either (attrsOf anything) (enum [ false ]);
+    type = with types; either (attrsOf (maybeRaw anything)) (enum [ false ]);
```
MattSturgeon: I don't understand why this is needed. Surely `anything` already covers raw lua?

DasGruene: This is quit an embarrassing mistake 😅 in my search for a  solution I definitely mixed a few things together that that i shouldnt have.

turns out the solution is to "just" use lib.nixvim.mkRaw.

I will remove this, but then there is really only the examples/documentation left, I wil let you decide if there is any worth in that. 

But anyway, Thanks for taking the time to look at my mistakes.

MattSturgeon: > there is really only the examples/documentation left,

Improving documentation and examples is always valuable. In this case having better examples may have avoided your confusion.

When you rebase, could you use our path/scope-style commit messages? See `git log -- plugins/by-name/blink-cmp` for examples or `CONTRIBUTING.md` for docs.

DasGruene: Saw this a bit too late,  but I will do some light clean up and get back to you.

DasGruene: Ive cleaned up my commit messages now 😉

plugins/by-name/blink-cmp/settings-options.nix
```diff
@@ -75,14 +75,18 @@ in
         "snippet_backward"
         "fallback"
       ];
-
+      # Example using raw Lua:
+      "<C-a>" = lib.nixvim.mkRaw "require('some-module').custom_function()";
```
MattSturgeon, line 79: FYI: this won't render in docs as a `mkRaw`, instead the _result_ (a `__raw` attr) will render.

Two ways to solve that:

1. make the whole example a `literalExpression`
2. use Nixvim's `nestedLiteralLua` helper

MattSturgeon (commented):
Sorry for the delay, this slipped into my backlog.

I noticed the `treefmt` CI didn't run, maybe we need a rebase on a more up-to-date `main` to pick up CI changes?

image

plugins/by-name/blink-cmp/settings-options.nix
MattSturgeon:
```suggestion
      "<C-a>" = lib.nixvim.nestedLiteralLua "require('some-module').custom_function()";
```

PR #4294, opened 2026-05-15
most dev flake inputs flake=false

mightyiam:
This sets `flake=false` for 5 out of the 8 dev flake inputs.

These 5 inputs seem to support being used as vanilla nix.

This eliminates the `inputs.<name>.inputs.<name>.follows.<name>` shenanigans.
It also eliminates the risk of failing to notice that an input has a new input.

Yes, we benefit from some flake features, such as purity, eval cache and input pinning.
But we don't have to use inputs as flakes if that incurs a cost and no benefit.

This is how I personally would like to consume flake inputs whenever they support it, including NixVim.
That is discussed in https://github.com/nix-community/nixvim/discussions/4288
and is mostly decoupled from this change.

By the way, I tried `nix flake check` in this repo and ran out of memory.
What's the trick to succeed?

### DIFF
--- a/flake/dev/devshell.nix
+++ b/flake/dev/devshell.nix
@@ -1,7 +1,7 @@
 { lib, inputs, ... }:
 {
   imports = [
-    inputs.devshell.flakeModule
+    "${inputs.devshell}/flake-module.nix"
   ];
 
   perSystem =
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -133,12 +133,7 @@
       }
     },
     "nix-darwin": {
-      "inputs": {
-        "nixpkgs": [
-          "nixvim",
-          "nixpkgs"
-        ]
-      },
+      "flake": false,
       "locked": {
         "lastModified": 1781761792,
         "narHash": "sha256-rCPytmKNjctLloB6UgK5CRrHSwV4b0ygxtJLPPp8R14=",
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -91,12 +91,7 @@
       }
     },
     "home-manager": {
-      "inputs": {
-        "nixpkgs": [
-          "nixvim",
-          "nixpkgs"
-        ]
-      },
+      "flake": false,
       "locked": {
         "lastModified": 1782220169,
         "narHash": "sha256-NxykfEdGQjop9m49Bk61pXpAduH8ICHgXC1N3Ph3Jp0=",
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -254,12 +254,7 @@
       }
     },
     "treefmt-nix": {
-      "inputs": {
-        "nixpkgs": [
-          "nixvim",
-          "nixpkgs"
-        ]
-      },
+      "flake": false,
       "locked": {
         "lastModified": 1780220602,
         "narHash": "sha256-eynAfOmbmxJnkp7YewvCEbShNnnYJ9gLLqkzsYtBPeM=",
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -1,12 +1,7 @@
 {
   "nodes": {
     "devshell": {
-      "inputs": {
-        "nixpkgs": [
-          "nixvim",
-          "nixpkgs"
-        ]
-      },
+      "flake": false,
       "locked": {
         "lastModified": 1768818222,
         "narHash": "sha256-460jc0+CZfyaO8+w8JNtlClB2n4ui1RbHfPTLkpwhU8=",
--- a/flake/dev/flake.lock
+++ b/flake/dev/flake.lock
@@ -75,16 +75,7 @@
       }
     },
     "git-hooks": {
-      "inputs": {
-        "flake-compat": [
-          "flake-compat"
-        ],
-        "gitignore": "gitignore",
-        "nixpkgs": [
-          "nixvim",
-          "nixpkgs"
-        ]
-      },
+      "flake": false,
       "locked": {
         "lastModified": 1781733627,
         "narHash": "sha256-U3yTuGBnmXvXoQI3qkpfEDsn9RovQPAjN7ndRco+3u0=",
@@ -96,27 +87,6 @@
       "original": {
         "owner": "cachix",
         "repo": "git-hooks.nix",
-        "type": "github"
-      }
-    },
-    "gitignore": {
-      "inputs": {
-        "nixpkgs": [
-          "git-hooks",
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1709087332,
-        "narHash": "sha256-HG2cCnktfHsKV0s4XW83gU3F57gaTljL9KNSuG6bnQs=",
-        "owner": "hercules-ci",
-        "repo": "gitignore.nix",
-        "rev": "637db329424fd7e46cf4185293b9cc8c88c95394",
-        "type": "github"
-      },
-      "original": {
-        "owner": "hercules-ci",
-        "repo": "gitignore.nix",
         "type": "github"
       }
     },
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -16,7 +16,7 @@
     # keep-sorted start block=yes newline_separated=yes
     devshell = {
       url = "github:numtide/devshell";
-      inputs.nixpkgs.follows = "nixvim/nixpkgs";
+      flake = false;
     };
 
     git-hooks = {
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -21,8 +21,7 @@
 
     git-hooks = {
       url = "github:cachix/git-hooks.nix";
-      inputs.nixpkgs.follows = "nixvim/nixpkgs";
-      inputs.flake-compat.follows = "flake-compat";
+      flake = false;
     };
 
     home-manager = {
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -31,7 +31,7 @@
 
     nix-darwin = {
       url = "github:nix-darwin/nix-darwin";
-      inputs.nixpkgs.follows = "nixvim/nixpkgs";
+      flake = false;
     };
 
     nuschtosSearch = {
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -1,6 +1,8 @@
 {
   description = "Private inputs for development purposes. These are used by the top level flake in the `dev` partition, but do not appear in consumers' lock files.";
 
+  # Avoid `flake = true` when we do not need to evaluate flake outputs or fetch flake inputs.
+  # e.g. when we only import loose files.
   inputs = {
     # Nix 2.26 improved support for relative path flake inputs.
     # This dev-flake is only evaluated by flake-parts' flake-compat,
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -26,7 +26,7 @@
 
     home-manager = {
       url = "github:nix-community/home-manager";
-      inputs.nixpkgs.follows = "nixvim/nixpkgs";
+      flake = false;
     };
 
     nix-darwin = {
--- a/flake/dev/flake.nix
+++ b/flake/dev/flake.nix
@@ -42,7 +42,7 @@
 
     treefmt-nix = {
       url = "github:numtide/treefmt-nix";
-      inputs.nixpkgs.follows = "nixvim/nixpkgs";
+      flake = false;
     };
 
     # keep-sorted end
--- a/tests/platforms/hm-extra-files-byte-compiling.nix
+++ b/tests/platforms/hm-extra-files-byte-compiling.nix
@@ -1,11 +1,11 @@
 {
   self,
   pkgs,
+  lib,
 }:
 let
-  inherit (self.inputs.home-manager.lib)
-    homeManagerConfiguration
-    ;
+  hmLib = import "${self.inputs.home-manager}/lib" { inherit lib; };
+  inherit (hmLib) homeManagerConfiguration;
 
   config = {
     home = {
--- a/tests/platforms/hm-submodule-merge.nix
+++ b/tests/platforms/hm-submodule-merge.nix
@@ -1,6 +1,7 @@
 {
   self,
   pkgs,
+  lib,
 }:
 # This test covers a user-reported regression where nixvim's submodule-option (programs.nixvim)
 # cannot correctly merge options declared from the parent scope.
@@ -9,9 +10,8 @@
 #
 # To be clear, this is an upstream module system bug, this test validates our workaround.
 let
-  inherit (self.inputs.home-manager.lib)
-    homeManagerConfiguration
-    ;
+  hmLib = import "${self.inputs.home-manager}/lib" { inherit lib; };
+  inherit (hmLib) homeManagerConfiguration;
 
   # This test module declares a nixvim option from a Home Manager module
   # The module system will attempt an option-type merge on the `programs.nixvim` option,
--- a/tests/platforms/hm.nix
+++ b/tests/platforms/hm.nix
@@ -1,8 +1,12 @@
 {
   self,
   pkgs,
+  lib,
 }:
-self.inputs.home-manager.lib.homeManagerConfiguration {
+let
+  hmLib = import "${self.inputs.home-manager}/lib" { inherit lib; };
+in
+hmLib.homeManagerConfiguration {
   inherit pkgs;
 
   modules = [
--- a/flake/dev/fmt.nix
+++ b/flake/dev/fmt.nix
@@ -1,7 +1,7 @@
 { inputs, ... }:
 {
   imports = [
-    inputs.treefmt-nix.flakeModule
+    "${inputs.treefmt-nix}/flake-module.nix"
   ];
 
   perSystem =
--- a/flake/dev/git-hooks.nix
+++ b/flake/dev/git-hooks.nix
@@ -1,7 +1,7 @@
 { inputs, ... }:
 {
   imports = [
-    inputs.git-hooks.flakeModule
+    "${inputs.git-hooks}/flake-module.nix"
   ];
 
   perSystem =


flake/dev/flake.lock
```diff
@@ -93,33 +81,8 @@
...
-        "repo": "gitignore.nix",
-        "rev": "637db329424fd7e46cf4185293b9cc8c88c95394",
-        "type": "github"
-      },
-      "original": {
-        "owner": "hercules-ci",
-        "repo": "gitignore.nix",
-        "type": "github"
-      }
-    },
```
mightyiam, line 127: Here is one input less 🕺

flake/dev/devshell.nix
```diff
@@ -1,7 +1,7 @@
 { lib, inputs, ... }:
 {
   imports = [
-    inputs.devshell.flakeModule
+    (inputs.devshell + "/flake-module.nix")
```
mightyiam: Small price to pay?

flake/dev/flake.nix
mightyiam: Nixpkgs will likely never have its own flake inputs, so I didn't bother.

mightyiam: flake-compat is also unlikely to have any of its own flake inputs, so I didn't bother.

mightyiam, line 43: This input does not seem to support being used as vanilla. It seems possible with a few lines, but I don't think that we should take that on, especially since it seems to pin its `ixx` input to some particular version that corresponds with something internal.

flake/dev/flake.nix
```diff
@@ -13,23 +13,22 @@
     # keep-sorted start block=yes newline_separated=yes
     devshell = {
       url = "github:numtide/devshell";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
```
mightyiam: `follows`es be gone!

tests/platforms/darwin.nix
```diff
@@ -1,18 +1,27 @@
...
+  inherit lib;
+
   modules = [
-    {
-      nixpkgs.hostPlatform = system;
+    (
+      { config, ... }:
+      {
+        nixpkgs = {
+          source = self.inputs.dev-nixpkgs;
```
mightyiam: In the case of this input, the `follows` actually has a purpose. For what it's worth, I think that the passing of an explicit `nixpkgs` _this way_ makes more sense, due to it being located here instead of in the `flake.nix`.

tests/platforms/darwin.nix
```diff
@@ -1,18 +1,27 @@
...
-      programs.nixvim = {
-        enable = true;
-      };
+        programs.nixvim = {
+          enable = true;
+        };
 
-      system.stateVersion = 5;
-    }
+        system.stateVersion = config.system.maxStateVersion;
```
mightyiam: I'm not sure what was going on, but evaluation failed and this seems to have worked around the issue. Probably better to do this anyway? If you are reviewing this and are curious, just revert this line and `nix eval ".#checks.aarch64-darwin.darwin-module"`.

mightyiam (commented):
For what it's worth, I've seen worse lockfiles. I've had worse lockfiles. The `follows` situation here seems minimal. And I don't think that the approach I'm suggesting here is terribly popular, but I think it's based.

flake/dev/flake.nix
MattSturgeon: flake-compat documents `flake = false` as its recommended usage.

flake/dev/flake.nix
MattSturgeon: I'm not entirely sold on `flake = false` being worth it just for eliminating transitive `follows`.

Imagine the scenario where an input changes its inputs without us noticing; if we pull it as a flake, then we get their new inputs transitively. If we don't, then things potentially break.

---

If the reason we can set `flake = false` is that we are explicitly only using a part of the project that is designed to be used without flake inputs, then that's fine and should probably be documented in a comment.

```nix
# Don't pull flake inputs or evaluate outputs.
# We exclusively use the flake-module file.
flake = false;
```

Or perhaps a top-level comment along the lines of:
```nix
# Avoid `flake = true` when we do not need to evaluate flake outputs or fetch flake inputs.
# e.g. when we only import loose files.
```

---

TL;DR a) consider whether we're setting `flake = false` for the right reasons b) consider whether `flake = false` is fragile and may miss future transitive dependencies c) document why we enable or disable `flake` in general or for specific inputs.

tests/platforms/darwin.nix
MattSturgeon: We should prefer `inputs.nixpkgs` over `inputs.dev-nixpkgs`.

We enforce that they are the same FOD, but `dev-nixpkgs` is an implementation detail of the dev-lockfile.

tests/platforms/darwin.nix
```diff
@@ -1,18 +1,27 @@
 {
   self,
   system,
+  lib,
 }:
-self.inputs.nix-darwin.lib.darwinSystem {
+(import (self.inputs.nix-darwin + "/eval-config.nix")) {
```
MattSturgeon: Is it worth naming this function?


```suggestion
let
  darwinSystem = import (self.inputs.nix-darwin + "/eval-config.nix");
in
darwinSystem {
```

Is this a publicly documented usage pattern? I.e. the "official" way to consume nix-darwin without flakes?

tests/platforms/hm-extra-files-byte-compiling.nix
```diff
@@ -1,9 +1,11 @@
 {
   self,
   pkgs,
+  lib,
 }:
 let
-  inherit (self.inputs.home-manager.lib)
+
+  inherit (import (self.inputs.home-manager + "/lib") { inherit lib; })
```
MattSturgeon: For readability, is it worth doing something like:


```suggestion
  hmLib = import (self.inputs.home-manager + "/lib") { inherit lib; };
  inherit (hmLib)
```

?

tests/platforms/hm-submodule-merge.nix
```diff
@@ -9,7 +10,7 @@
 #
 # To be clear, this is an upstream module system bug, this test validates our workaround.
 let
-  inherit (self.inputs.home-manager.lib)
+  inherit (import (self.inputs.home-manager + "/lib") { inherit lib; })
```
MattSturgeon: Same

tests/platforms/hm.nix
```diff
@@ -1,8 +1,9 @@
 {
   self,
   pkgs,
+  lib,
 }:
-self.inputs.home-manager.lib.homeManagerConfiguration {
+(import (self.inputs.home-manager + "/lib") { inherit lib; }).homeManagerConfiguration {
```
MattSturgeon: Same

tests/platforms/darwin.nix
MattSturgeon: Also, because flake inputs are already in-store objects, we could consider using:

```nix
"${self.inputs.nix-darwin}/eval-config.nix"
```

string interpolation.

IIUC, this shouldn't copy-to-store if the path is already a store-object.

Either approach is fine by me, though. You current approach is technically more robust at avoiding potential copy-to-store operations.

MattSturgeon:
It's worth noting that the dev-lockfile you've worked on here is not user-facing.

It does not get evaluated by end-users unless they evaluate things like our test suite or dev shell.

MattSturgeon:
> By the way, I tried `nix flake check` in this repo and ran out of memory.
> What's the trick to succeed?

I believe this is documented in CONTRIBUTING.md.

IIRC, we still have our entire test suite in the `checks` output, and when `nix flake check` tries to evaluate it all in one go it runs out of memory.

We have a `tests` command in our dev shell that runs the test suite via chunked evals. We also chunk them up in our `ci.buildbot` output, which you could build using something like `nix-fast-build`. We should probably look into dropping tests that can't be evaluated from the flake checks output.

flake/dev/flake.nix
```diff
@@ -13,7 +13,7 @@
...
+      flake = false;
```
mightyiam, line 21: Exposes a vanilla `/flake-module.nix`, so we're good.

flake/dev/flake.nix
```diff
@@ -18,8 +18,7 @@
 
     git-hooks = {
       url = "github:cachix/git-hooks.nix";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
-      inputs.flake-compat.follows = "flake-compat";
+      flake = false;
```
mightyiam, line 26: Exposes a vanilla `/flake-module.nix`, so we're good.

flake/dev/flake.nix
```diff
@@ -23,7 +23,7 @@
 
     home-manager = {
       url = "github:nix-community/home-manager";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
+      flake = false;
```
mightyiam, line 31: Provides a `/lib/default.nix` that takes `{ inherit lib; }` so we're golden.

flake/dev/flake.nix
```diff
@@ -39,7 +39,7 @@
 
     treefmt-nix = {
       url = "github:numtide/treefmt-nix";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
+      flake = false;
```
mightyiam, line 47: Provides a vanilla `/flake-module.nix` so we're golden.

flake/dev/flake.nix
```diff
@@ -1,35 +1,42 @@
 {
   description = "Private inputs for development purposes. These are used by the top level flake in the `dev` partition, but do not appear in consumers' lock files.";
 
+  # Avoid `flake = true` when we do not need to evaluate flake outputs or fetch flake inputs.
+  # e.g. when we only import loose files.
```
mightyiam, line 5: Here's a comment you suggested, @MattSturgeon.

flake/dev/flake.nix
```diff
@@ -1,35 +1,39 @@
...
     # NOTE: Use a different name to the root flake's inputs.nixpkgs to avoid shadowing it.
     # NOTE: The only reason we specify a nixpkgs input at all here, is so the other inputs can follow it.
     # TODO: Once nix 2.26 is more prevalent, follow the root flake's inputs using a "path:../.." input.
     dev-nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
 
     # flake-compat is used by the root `default.nix` to allow non-flake users to import nixvim
-    flake-compat.url = "github:NixOS/flake-compat";
+    flake-compat = {
+      url = "github:NixOS/flake-compat";
+      flake = false;
```
mightyiam, line 15: As @MattSturgeon pointed, flake-compat instructs declaring it with `flake=false` in its readme.

flake/dev/flake.nix
mightyiam: I've added that single comment above the inputs attribute.

I've pointed out in this PR for each `flake=false` why it is okay.

I think that projects that expose truly loose files such as `/flake-module.nix` and `/lib/default.nix` are aware that users will use them and are not likely to turn them into functions that first take local context such as flake inputs. I prefer this risk over lockfile drift.

flake/dev/flake.nix
```diff
@@ -1,35 +1,39 @@
...
     home-manager = {
       url = "github:nix-community/home-manager";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
+      flake = false;
     };
 
     nix-darwin = {
       url = "github:lnl7/nix-darwin";
-      inputs.nixpkgs.follows = "dev-nixpkgs";
+      flake = false;
```
mightyiam, line 36: Provides a vanilla `/eval-config.nix` so we're golden.

tests/platforms/darwin.nix
mightyiam: I'll assume you are correct about what we _should_ be doing, but in `main`, the nix-darwin input has `inputs.nixpkgs.follows = "dev-nixpkgs";`. Would you like a separate PR that changes that?

mightyiam (commented):
I totally get that this whole change does not affect users.

Well, here are some changes. Please let me know whether I missed anything.

tests/platforms/darwin.nix
MattSturgeon: > in `main`, the nix-darwin input has `inputs.nixpkgs.follows = "dev-nixpkgs";`.

In `main` it is necessary, because of the way `nix flake lock` works; we cannot _follow_ an input from the top-level lockfile so we introduced `dev-nixpkgs` as a workaround.

`inputs.dev-nixpkgs` is a mirror of the top-level flake's `inputs.nixpkgs` and should only be used when the workaround is necessary.

Or to quote the inline comment:

> The only reason we specify a nixpkgs input at all here, is so the other inputs can follow it.

As of this PR, I believe the only input that needs `dev-nixpkgs` is `nuschtosSearch`.

MattSturgeon:
Closing as stale. Thanks for trying to improve things, regardless!

If you _are_ motivated to rebase and re-open, just know that I'd want to see buy-in from other maintainers before committing to actually merge this refactor.

Plus, I'm unsure how strong the benefits are in doing this to our _`dev`_ lockfile and I'm unconvinced we want to be "fighting" how flakes are intended to be used.

mightyiam:
Rebased and converted the import style from

```nix
(inputs.foo + "/module.nix")
```

to

```nix
"${inputs.foo}/module.nix"
```

tests/platforms/hm-extra-files-byte-compiling.nix
```diff
@@ -1,11 +1,11 @@
 {
   self,
   pkgs,
+  lib,
 }:
 let
-  inherit (self.inputs.home-manager.lib)
-    homeManagerConfiguration
-    ;
+  hmLib = import (self.inputs.home-manager + "/lib") { inherit lib; };
```
mightyiam: These can get a little shorter by bumping home-manager:
https://github.com/nix-community/home-manager/pull/9430

mightyiam (commented):
I think this is an improvement 🤷

mightyiam:
Rebased

MattSturgeon:
> Rebased

AFAICT GitHub refuses to re-open a PR after its branch has been force-pushed to.

image

If re-opening, you may need to reset the branch back to 4d35f4db3280907c8e4139231059762b9ee8c1ad first.

---

> I'd want to see buy-in from other maintainers before committing to actually merge this refactor.

I'm still reserving my judgement until other maintainers have given their views.

mightyiam:
Okay, I've force-pused to 4d35f4d.

MattSturgeon:
Re-opening as you're continuing to work; above comments still stand.

mightyiam:
I'd be happy to rebase :shrug:

MattSturgeon:
> If you _are_ motivated to rebase and re-open, just know that I'd want to see buy-in from other maintainers before committing to actually merge this refactor.

In 2 months, We've not seen any engagement from other maintainers. Either they're against the PR but keeping quiet or they're indifferent and don't wish to advocate for it.

Therefore I think we should re-close this and continue using flake inputs "as intended".


> Plus, I'm unsure how strong the benefits are in doing this to our _`dev`_ lockfile and I'm unconvinced we want to be "fighting" how flakes are intended to be used.

Honestly, the fact we have a separate "dev" lockfile at all is so that end users don't feel the need to "clean up" their downstream lockfiles by applying`follows=""` to our inputs.

In practice, the dev lockfile is only used when evaluating dev outputs, such as the `devShell` or `checks`; therefore I don't think it's necessary to over-optimise its dependency graph.

If upstream nix can implement lazier flake inputs fetching and/or avoid re-locking transitive inputs locally, then both issues will be solved upstream and we won't need _either_ a separate dev lockfile _or_ `flake=false` on inputs.

In the meantime, I'd rather reserve workarounds for where there is a tangible benefit to end users or a significant benefit to developers.

---

Sorry if I've strung you along by leaving the door open on this one.

GaetanLepage:
Hi!
This one slipped under my radar, sorry.
I agree with @MattSturgeon in the sense that benefits do not seem to be very limited.
Sorry for denying your contribution. I truly appreciate you taking the time to submit this PR.
If you feel like this is truly a mistake, please make your case.

All the best

mightyiam:
No worries, @GaetanLepage. Thank you for taking a look.

First, this is a low impact change, granted.

But I think the `flake=false` should obviously be preferred.

Let's take `inputs.devshell.inputs.nixpkgs.follows = "nixvim/nixpkgs"` as an example.
It prevents an extraneous input in the lockfile by deduplicating it with another.

`inputs.devshell.flake=false` acknowledges the truth, that that transitive input was never used anyway. And it guarantees that no new transitive inputs will be added for that input.

And what is the cost in this trade-off? Nothing? Is `flake=false` a dirty workaround? No, it's a clean way of saying "hey, I'm not using this input as a flake. Just fetch it for me, please".

This change is zero-cost and net-benefit.

MattSturgeon:
> Is `flake=false` a dirty workaround? No, it's a clean way of saying "hey, I'm not using this input as a flake. Just fetch it for me, please".

I agree with this in principle. It's why we fetch `flake-compat` with `flake=false`:[^1]
https://github.com/nix-community/nixvim/blob/f1311e1d8c101bf3fbce6804de144ca1546d0901/flake/dev/flake.nix#L11-L14

However, `flake=false` doesn't just affect transitive inputs; it tells Nix that the input is not a flake at all, meaning Nix won't evaluate its flake *outputs* either.
Therefore, the decision to use `flake=false` really comes down to whether we need or want an input's flake *outputs*, rather than just trying to avoid its transitive *inputs*.

In many cases, we actually *do* want to consume these inputs as flakes, because we prefer to use their flake *outputs* instead of importing from loose files. This is usually what I mean by consuming flakes "idiomatically."
Sometimes, flake outputs are the only "official" API endorsed by a project. Other times, it's just more convenient.

I'd fully support a PR that declares `flake=false` only on inputs where evaluating the flake provides zero benefit: i.e., where we don't use the flake outputs *and* we have no need for the transitive inputs.

I'd also be open to dropping unnecessary inputs altoghetr; for example we could switch from `treefmt-nix` to Nixpkgs's built-in `treefmt.withConfig` if we're willing to lose some of the pre-configured formatter defaults provided by `treefmt-nix`. Or we could drop `devshell` in favour of directly using `mkShell` (and maybe some flake `apps` outputs).

[^1]: The same applies to `nix-systems`, so I opened #4520.

PR #4308, opened 2026-05-19
modules/lsp/onAttach: handle dynamic registrations

### LINKED ISSUE #3323: `lsp.onAttach` should use `client/registerCapability` handler
The [`:h LspAttach`](https://neovim.io/doc/user/lsp.html#LspAttach) docs mention the following:

> Note: If the LSP server performs dynamic registration, capabilities may be registered any time _after_ `LspAttach`. In that case you may want to handle the "`registerCapability`" event.
>
> Example:
>
> ```lua
> vim.lsp.handlers['client/registerCapability'] = (function(overridden)
>   return function(err, res, ctx)
>     local result = overridden(err, res, ctx)
>     local client = vim.lsp.get_client_by_id(ctx.client_id)
>     if not client then
>       return
>     end
>     -- Call your custom on_attach logic...
>     -- my_on_attach(client, vim.api.nvim_get_current_buf())
>     return result
>   end
> end)(vim.lsp.handlers['client/registerCapability'])
> ```

Perhaps this is a more appropriate place to put our "global" onAttach code, instead of using a `LspAttach` autoCmd?

khaneliman:
Some LSP servers register capabilities after LspAttach, so the global onAttach body must also run from client/registerCapability for buffers already attached to that client.

Keep the initial LspAttach autocmd for servers that advertise capabilities up front, and preserve the wrapped handler's return value.

Closes #3323

### DIFF
--- a/modules/lsp/on-attach.nix
+++ b/modules/lsp/on-attach.nix
@@ -47,13 +47,38 @@ in
               -- client and bufnr are supplied to the builtin `on_attach` callback,
               -- so make them available in scope for our global `onAttach` impl
               local client = vim.lsp.get_client_by_id(event.data.client_id)
-              local bufnr = event.buf
-              ${cfg.onAttach}
+              if client == nil then
+                return
+              end
+
+              __nixvim_lsp_on_attach(client, event.buf)
             end
           end
         '';
         desc = "Run LSP onAttach";
       }
     ];
+
+    extraConfigLua = ''
+      local function __nixvim_lsp_on_attach(client, bufnr)
+        ${cfg.onAttach}
+      end
+
+      vim.lsp.handlers["client/registerCapability"] = (function(overridden)
+        return function(err, res, ctx, ...)
+          local result = overridden(err, res, ctx, ...)
+          local client = vim.lsp.get_client_by_id(ctx.client_id)
+          if client == nil then
+            return result
+          end
+
+          for bufnr, _ in pairs(client.attached_buffers) do
+            __nixvim_lsp_on_attach(client, bufnr)
+          end
+
+          return result
+        end
+      end)(vim.lsp.handlers["client/registerCapability"])
+    '';
   };
 }
--- a/tests/test-sources/modules/lsp.nix
+++ b/tests/test-sources/modules/lsp.nix
@@ -173,6 +173,83 @@
       '';
     };
 
+  on-attach-dynamic-registration = {
+    lsp.onAttach = ''
+      table.insert(_G.__nixvim_lsp_on_attach_calls, {
+        client = client.id,
+        bufnr = bufnr,
+      })
+    '';
+
+    extraConfigLuaPre = ''
+      _G.__nixvim_lsp_on_attach_calls = {}
+      vim.lsp.handlers["client/registerCapability"] = function()
+        return "registered"
+      end
+    '';
+
+    extraConfigLuaPost = ''
+      local get_client_by_id = vim.lsp.get_client_by_id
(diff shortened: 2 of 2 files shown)


modules/lsp/on-attach.nix
```diff
@@ -55,5 +59,27 @@ in
         desc = "Run LSP onAttach";
       }
     ];
+
+    extraConfigLua = ''
+      vim.lsp.handlers["client/registerCapability"] = (function(overridden)
```
MattSturgeon, line 67: Can there only be one handler for `client/registerCapability`? There's no "register a handler dynamically" API?

The `overridden` closure pattern is a fair workaround; are end users likely to follow it if they override/extend _our_ handler? E.g. is it included in neovim's `:help` examples?

modules/lsp/on-attach.nix
```diff
@@ -55,5 +59,27 @@ in
...
+        return function(err, res, ctx, ...)
+          local function __nixvim_lsp_on_attach(client, bufnr)
+            ${cfg.onAttach}
+          end
+
+          local result = overridden(err, res, ctx, ...)
```
MattSturgeon, line 69: Does this `...` work as an argument here? Surprising lua syntax if so; I'm used to it being used as a vararg parameter in other languages, or a "spread" operator. Here, is it a literal name representing "the other arguments"?

modules/lsp/on-attach.nix
```diff
@@ -55,5 +59,27 @@ in
         desc = "Run LSP onAttach";
       }
     ];
+
+    extraConfigLua = ''
+      vim.lsp.handlers["client/registerCapability"] = (function(overridden)
+        return function(err, res, ctx, ...)
+          local function __nixvim_lsp_on_attach(client, bufnr)
+            ${cfg.onAttach}
+          end
```
MattSturgeon: Can't we define this somewhere centrally and reuse it here and in `LspAttach`?

Or alternatively invoke the autocmd here instead of re-defining the function? Does that get too messy?

MattSturgeon (commented):
Concept SGTM. Questions and nits.

modules/lsp/on-attach.nix
khaneliman: Yep, this is modeled after neovim's own `:h LspAttach` dynamic registration example: https://neovim.io/doc/user/lsp/#LspAttach from the linked issue you created. Neovim documents `vim.lsp.handlers` as a method-to-handler table, with overrides done by setting a field https://neovim.io/doc/user/lsp/#lsp-handler-resolution

modules/lsp/on-attach.nix
khaneliman: `...` is a Lua `vararg` syntax. Since the wrapper is declared with it, the `overridden` will forward any extra arguments through. https://www.lua.org/manual/5.1/manual.html#2.5.9 Page doesn't have any real anchors, hard to link to `vararg` docs

modules/lsp/on-attach.nix
khaneliman: Updated to move scope out and reuse with both `autocmd` and `registerCapability`

modules/lsp/on-attach.nix
MattSturgeon: > [if] the function is a variadic or vararg function, which is indicated by three dots ('...') at the end of its parameter list [it] does not adjust its argument list; instead, it collects all extra arguments and supplies them to the function through a vararg expression, which is also written as three dots.

Interesting, thanks for pointing at the relevant docs

MattSturgeon (approved):

PR #4309, opened 2026-05-19
plugins/jdtls: prefer vim.lsp config registration

### LINKED ISSUE #3768: Migrate nvim-jdtls to `vim.lsp`/`lsp.servers`
Supported by upstream: https://codeberg.org/mfussenegger/nvim-jdtls#via-lsp-config

Like clangd-extensions at 8b3107ad6f520080135e06ab6c017bdd3952d210

### LINKED ISSUE #4165: [BUG] jdtls does not specify a command to run the language server by default
### Plugin

jdtls

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

### Description

By default, `plugins.jdtls.enable` does not populate `plugins.jdtls.settings.cmd` field. When this is not populated, an assertion fails because the plugin expects to have the command to start the language server in this field. Expected behavior is that it would populate this with the `plugins.jdtls.jdtLanguageServerPackage` field so it works without needing further configuration which is not clearly documented.

```nix
programs.nixvim = {
    plugins.jdtls.enable = true;
};
```

khaneliman:
nvim-jdtls 0.11+ can be initialized through native Vim LSP config, so explicit FileType start_or_attach is no longer needed.

This avoids passing an empty table into start_or_attach, which triggers the runtime assertion when cmd is unset.

Instead, map plugins.jdtls.settings to lsp.servers.jdtls.config and keep the server enabled by default.

close #4165
close #3768

### DIFF
--- a/plugins/by-name/jdtls/default.nix
+++ b/plugins/by-name/jdtls/default.nix
@@ -16,22 +16,15 @@ lib.nixvim.plugins.mkNeovimPlugin {
     };
   };
 
-  setup = ".start_or_attach"; # only used settingsDescription
+  settingsDescription = "LSP configuration passed to `vim.lsp.config('jdtls', ...)`.";
   callSetup = false;
-  extraConfig = cfg: {
+  extraConfig = cfg: opts: {
     extraPackages = [ cfg.jdtLanguageServerPackage ];
 
-    autoCmd = [
-      {
-        event = "FileType";
-        pattern = "java";
-        callback.__raw = ''
-          function ()
-            require('jdtls').start_or_attach(${lib.nixvim.toLuaObject cfg.settings})
-          end
-        '';
-      }
-    ];
+    lsp.servers.jdtls = {
+      enable = lib.mkDefault true;
+      config = lib.mkDerivedConfig opts.settings lib.id;
+    };
   };
 
   settingsOptions = import ./settings-options.nix lib;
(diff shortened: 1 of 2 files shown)


plugins/by-name/jdtls/default.nix
```diff
@@ -16,30 +16,23 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-        callback.__raw = ''
-          function ()
-            require('jdtls').start_or_attach(${lib.nixvim.toLuaObject cfg.settings})
-          end
-        '';
-      }
-    ];
+    lsp.servers.jdtls = {
+      enable = lib.mkDefault true;
+      config = cfg.settings;
```
MattSturgeon: I'd be tempted to use [`mkDerivedConfig`](https://noogle.dev/f/lib/mkDerivedConfig) here.

Although, IIRC it doesn't work _that_ well for attrsets anyway, as only the top-level `highestPrio` is preserved; still worth considering. The alternative is to copy what `doRename` does and `mkMerge` the original raw definitions; not worth it in this case without appropriate helpers, but we do do it in some other cases.

MattSturgeon (approved):

plugins/by-name/jdtls/default.nix
khaneliman: I completely forgot about that function. I agree it makes sense to use here.

PR #4310, opened 2026-05-19
plugins/direnv: fix `settingsOptions`

FovirDev:
This PR has done:

- Remove global prefixes in `settingsOptions`
- Correct the default value of `silent_load` [direnv/direnv.vim](https://github.com/direnv/direnv.vim/blob/ab2a7e08dd630060cd81d7946739ac7442a4f269/autoload/direnv.vim#L13-L15)


GaetanLepage (approved):
Good catch!

PR #4311, opened 2026-05-19
modules: allow `config.flake` to be undefined

MattSturgeon:
Using a similar technique to flake-compat, we can read our `flake.lock` and fetch the locked Nixpkgs revision using `fetchTree`. This is done in a new `nixpkgs.nix` file, which when imported evaluates to the same thing as `inputs.nixpkgs`:
```nix
$ nix-instantiate --eval nixpkgs.nix
{ lastModified = 1778869304; lastModifiedDate = "20260515182144"; narHash = "sha256-30sZNZoA1cqF5JNO9fVX+wgiQYjB7HJqqJ4ztCDeBZE="; outPath = "/nix/store/77dbgds155bbz3vd3qywq1sii07i5ljs-source"; rev = "d233902339c02a9c334e7e593de68855ad26c4cb"; shortRev = "d233902"; }
```

We use this to provide a fallback default for `nixpkgs.source`.

For `wrappers/`, we can access nixvim's lib and lib overlay internally from the `programs.nixvim` submodule configuration. There's no need to pass it in as an external argument.

For `build.manDocsPackage` things are a little more complicated; we currently encapsulate all docs-packages in `docs/default.nix` which itself needs an instance of the flake, so we always access docs packages via flake outputs. This is worth solving, but it too big a refactor for this PR. For now, I've changed `enableMan`'s default to check `options.flake.isDefined`.

`default.nix` could now be refactored to not rely on the flake at all, and drop flake-compat. But that's a bigger (potentially user-facing) change that I'd prefer to delay for a future PR. We also rely on flake-compat internally, e.g. in our update script we read `inputs.nixpkgs.rev` from `default.nix`.

This fixes #4288, and also allows us to warn users who have accidentally applied `follows` to our `nixpkgs.source` default.
We could also decide to drop the warning and circumvent `follows`, now or in the future.

To fully resolve #4288 we should address:

- [x] `nixpkgs.source` option.
- [x] `wrappers/_shared.nix` gets nixvim's `lib` from flake outputs.
- [x] manpage docs in `outputs.nix` are referenced via flake outputs.
      (Half-solved by having it be conditional; currently all docs packages are encapsulated in `docs/default.nix` and access to them isn't easy without flake-outputs)
- [x] `lib/top-level.nix` and `lib/overlay.nix` need a reference to `flake`. Remove this or make it optional.
- [ ] `default.nix` loads the flake via flake-compat, which is mostly unnecessary with this PR.
      (This is non-trivial to solve, as we use flake-compat for external access to flake input attributes)
      (I have ideas for deprecating this by applying `mapAttrs (_: warn "")` to the flake-compat result and shadowing it with `// {/* non-deprecated outputs */}`. For internal "what is `inputs.nixpkgs.rev`", we can actually use `(import ./nixpkgs.nix).rev`!)

cc @mightyiam

### DIFF
--- a/flake/dev/fmt.nix
+++ b/flake/dev/fmt.nix
@@ -25,6 +25,10 @@
             excludes = [
               # sema-unused-def-lambda-noarg-formal
               "ci/rust-analyzer/default.nix"
+
+              # sema-primop-overridden
+              # sema-primop-removed-prefix
+              "nixpkgs.nix"
             ];
           };
           nixfmt = {
--- a/flake/lib.nix
+++ b/flake/lib.nix
@@ -14,7 +14,7 @@
       # on pinning the flake-parts nixpkgs-lib to the nixpkgs pin
       inherit (inputs.nixpkgs) lib;
     };
-    overlay = import ../lib/overlay.nix {
+    overlay = import ../lib/overlay.internal.nix {
       flake = self;
     };
     # Top-top-level aliases
--- a/lib/overlay.internal.nix
+++ b/lib/overlay.internal.nix
@@ -1,0 +1,42 @@
+/**
+  Internal function returning Nixvim's Nixpkgs `lib` overlay.
+
+  Publicly accessible at `./overlay.nix` and the `nixvim.lib.overlay` flake output.
+*/
+{ flake }:
+
+/**
+  Overlay function extending the Nixpkgs `lib` with Nixvim's additions.
+
+  This function is intended to be passed to `lib.extend`.
+
+  # Examples
+
+  Importing the file directly:
+  ```
+  lib.extend (import (nixvim + "/lib/overlay.nix"))
+  ```
+
+  Using the overlay from flake outputs:
+  ```
+  lib.extend nixvim.lib.overlay
+  ```
+
+  # Inputs
+
+  `lib`
+  : The final lib instance; the fixpoint of all extensions, including this one.
+
+  `prevLib`
+  : The lib instance prior to applying this overlay.
+*/
+lib: prevLib: {
+  # Add Nixvim's section to the lib
+  nixvim = import ./top-level.nix { inherit flake lib; };
+
+  # Extend the maintainers set with Nixvim-specific maintainers
+  maintainers = prevLib.maintainers // import ./maintainers.nix;
+
+  # Extend lib.types with Nixvim's custom types
+  types = prevLib.types // import ./types.nix { inherit lib; };
+}
--- a/lib/overlay.nix
+++ b/lib/overlay.nix
@@ -1,37 +1,4 @@
-{ flake }:
-
-/**
-  Overlay function extending the Nixpkgs `lib` with Nixvim's additions.
-
-  This function is intended to be passed to `lib.extend`.
-
-  # Examples
-
-  Importing the file directly:
-  ```
-  lib.extend (import ./overlay.nix { flake = nixvim; })
-  ```
-
-  Using the overlay from flake outputs:
-  ```
-  lib.extend nixvim.lib.overlay
-  ```
-
-  # Inputs
-
-  `lib`
-  : The final lib instance; the fixpoint of all extensions, including this one.
-
-  `prevLib`
-  : The lib instance prior to applying this overlay.
-*/
-lib: prevLib: {
-  # Add Nixvim's section to the lib
-  nixvim = import ./top-level.nix { inherit flake lib; };
-
-  # Extend the maintainers set with Nixvim-specific maintainers
-  maintainers = prevLib.maintainers // import ./maintainers.nix;
-
-  # Extend lib.types with Nixvim's custom types
-  types = prevLib.types // import ./types.nix { inherit lib; };
+# Public non-flake entrypoint for Nixvim's Nixpkgs-lib overlay
+import ./overlay.internal.nix {
+  flake = null;
 }
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -9,6 +9,8 @@ let
   cfg = config.nixpkgs;
   opt = options.nixpkgs;
 
+  pinned = import ../../nixpkgs.nix;
+
   isConfig = x: builtins.isAttrs x || lib.isFunction x;
 
   mergeConfig =
@@ -188,7 +190,7 @@ in
     # NOTE: This is a nixvim-specific option; there's no equivalent in nixos
     source = lib.mkOption {
       type = lib.types.path;
-      default = config.flake.inputs.nixpkgs;
+      default = (lib.optionalAttrs options.flake.isDefined config.flake).inputs.nixpkgs or pinned;
       defaultText = lib.literalMD "Nixvim's flake `input.nixpkgs`";
       description = ''
         The path to import Nixpkgs from.
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -11,6 +11,8 @@ let
 
   pinned = import ../../nixpkgs.nix;
 
+  optionDefaultPrio = (lib.mkOptionDefault null).priority;
+
   isConfig = x: builtins.isAttrs x || lib.isFunction x;
 
   mergeConfig =
@@ -231,6 +233,16 @@ in
                 };
           in
           import cfg.source (args // systemArgs);
+
+      # Whether `pkgs` was constructed by this module.
+      # This is false when any of nixpkgs.pkgs or _module.args.pkgs is set.
+      constructedByMe =
+        # We set it with default priority and it can not be merged, so if the
+        # pkgs module argument has that priority, it's from us.
+        (lib.modules.mergeAttrDefinitionsWithPrio options._module.args).pkgs.highestPrio
+        == lib.modules.defaultOverridePriority
+        # Although, if nixpkgs.pkgs is set, we did forward it, but we did not construct it.
+        && !opt.pkgs.isDefined;
     in
     {
       # We explicitly set the default override priority, so that we do not need
@@ -256,5 +268,20 @@ in
           '';
         }
       ];
+
+      # TODO: consider `nixpkgs.source` always defaulting to `pinned`? That would bypass flake input following.
+      warnings =
+        lib.optional
+          (
+            constructedByMe
+            && opt.source.highestPrio == optionDefaultPrio
+            && pinned.narHash != cfg.source.narHash or null
+          )
+          ''
+            The `${opt.source}` default value has been affected by your flake input `follows`.
+            Nixvim's inputs pin Nixpkgs to `${pinned.rev}`.${
+              lib.optionalString (cfg.source ? rev) " Actual Nixpkgs is following `${cfg.source.rev}`."
+            }
+            Please remove your `inputs.nixvim.inputs.nixpkgs.follows` or explicitly define `${opt.source}` to suppress this warning.'';
     };
 }
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -351,7 +351,12 @@ in
           '';
         };
 
-        manDocsPackage = config.flake.packages.${system}.man-docs;
+        # TODO: make `build.manDocsPackage` available in non-flake evals.
+        # Also update `enableMan` default value.
+        manDocsPackage =
+          lib.throwIfNot options.flake.isDefined
+            "`${options.build.manDocsPackage}` currently requires `${options.flake}`, which is not defined."
+            config.flake.packages.${system}.man-docs;
       };
 
       # Set `wrapRc` and `impureRtp`s option defaults with even lower priority than `mkOptionDefault`
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -1,0 +1,91 @@
+/**
+  Fetches Nixvim's pinned revision of Nixpkgs without evaluating the flake.
+*/
+let
+  # fetchTree was added in Nix 2.4 as an experimental feature, so provide a fetchTarball-based polyfill
+  # Based on https://github.com/NixOS/flake-compat/blob/5edf11c4/default.nix#L20-L36
+  fetchTree =
+    builtins.fetchTree or (
+      info:
+      if info.type == "github" then
+        {
+          inherit (info) rev narHash lastModified;
+          shortRev = builtins.substring 0 7 info.rev;
+          lastModifiedDate = formatSecondsSinceEpoch info.lastModified;
+          outPath = fetchTarball {
+            url = "https://api.${info.host or "github.com"}/repos/${info.owner}/${info.repo}/tarball/${info.rev}";
+            sha256 = info.narHash;
+          };
+        }
+      else
+        throw "fetchTree: unsupported type '${info.type}'"
+    );
+
+  # Format number of seconds since the Unix epoch as %Y%m%d%H%M%S.
+  # Based on https://github.com/NixOS/flake-compat/blob/5edf11c4/default.nix#L172-L197
+  # Uses Howard Hinnant civil date algorithm.
+  formatSecondsSinceEpoch =
+    let
+      rem = x: y: x - x / y * y;
+      pad = s: if builtins.stringLength s < 2 then "0" + s else s;
+
+      second = 1;
+      minute = 60 * second;
+      hour = 60 * minute;
+      day = 24 * hour;
+
+      # Gregorian leap year rules:
+      # - leap year every 4 years
+      # - except every 100 years
+      # - except every 400 years
+      daysPerYear = 365;
+      daysPer4Years = daysPerYear * 4 + 1;
+      daysPer100Years = daysPerYear * 100 + 24;
+      daysPer400Years = daysPerYear * 400 + 97;
+
+      monthsToDaysDivisor = 153;
+      monthsToDaysOffset = 2;
+
+      # Days between civil date epoch 0000-03-01 and Unix epoch 1970-01-01.
+      civilDayOffset = 719468;
+    in
+    t:
+    let
+      z = (t / day) + civilDayOffset;
+      secondsInDay = rem t day;
+      hours = secondsInDay / hour;
+      minutes = (rem secondsInDay hour) / minute;
+      seconds = rem t minute;
+
+      # Gregorian calendars repeat every 400 years.
+      era = (if z >= 0 then z else z - (daysPer400Years - 1)) / daysPer400Years;
+
+      dayOfEra = z - era * daysPer400Years;
+
+      yearOfEra =
+        (
+          dayOfEra - dayOfEra / daysPer4Years + dayOfEra / daysPer100Years - dayOfEra / (daysPer400Years - 1)
+        )
+        / daysPerYear;
+
+      year' = yearOfEra + era * 400;
+      dayOfYear = dayOfEra - (daysPerYear * yearOfEra + yearOfEra / 4 - yearOfEra / 100);
+      month' = (5 * dayOfYear + monthsToDaysOffset) / monthsToDaysDivisor;
+      dayOfMonth = dayOfYear - (monthsToDaysDivisor * month' + monthsToDaysOffset) / 5 + 1;
+
+      # The algorithm uses March-indexed months, so offset back to January-index
+      month = month' + (if month' < 10 then 3 else -9);
+      year = year' + (if month <= 2 then 1 else 0);
+    in
+    toString year
+    + pad (toString month)
+    + pad (toString dayOfMonth)
+    + pad (toString hours)
+    + pad (toString minutes)
+    + pad (toString seconds);
+
+  lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  node = lock.nodes.${lock.nodes.${lock.root}.inputs.nixpkgs};
+in
+assert !(node.locked ? dir);
+fetchTree node.locked
--- a/tests/nixpkgs-fetch.nix
+++ b/tests/nixpkgs-fetch.nix
@@ -1,0 +1,153 @@
+{
+  lib,
+  self,
+  jq,
+  nix,
+  linkFarmFromDrvs,
+  runCommandLocal,
+  writableTmpDirAsHomeHook,
+}:
+
+# nixpkgs.nix fetches our pinned nixpkgs using the `fetchTree` primop.
+#
+# In this file, we test that it is identical to our `inputs.nixpkgs` flake input.
+#
+# The primop is only available with `experimental-features = flakes` or `experimental-features = fetch-tree`,
+# so we have a `fetchTree` polyfill that delegates to `fetchTarball`.
+#
+# Permutations to test:
+# 1. fetchTree primop
+#   a. attrs
+#   b. outPath derivation
+# 2. fetchTree polyfill
+#   a. attrs
+#   b. outPath derivation
+#
+# As we run this test from a flake, we have `experimental-features = flakes` and `fetchTree` primop is available.
+# Therefore, we can test the `fetchTree` primop easily.
+#
+# By running `nix-instantiate` _within_ a build sandbox, we can configure `experimental-features =`
+# to test the polyfill's attrs. However, we cannot do an actual fetch in the sandbox.
+#
+# Since we can't do an actual fetch in the sandbox, the polyfill's outPath can be tested manually.
+# These should both evaluate to the same derivation:
+#     nix-instantiate --eval nixpkgs.nix --attr outPath --option experimental-features flakes
+#     nix-instantiate --eval nixpkgs.nix --attr outPath --option experimental-features ''
+
+let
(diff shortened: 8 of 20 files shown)


flake/dev/fmt.nix
```diff
@@ -25,6 +25,10 @@
             excludes = [
               # sema-unused-def-lambda-noarg-formal
               "ci/rust-analyzer/default.nix"
+
+              # sema-primop-overridden
+              # sema-primop-removed-prefix
+              "nixpkgs.nix"
```
MattSturgeon, line 31: Needed for the `fetchTree = builtins.fetchTree or (...)` fallback

modules/top-level/nixpkgs.nix
```diff
@@ -254,5 +258,16 @@ in
           '';
         }
       ];
+
+      # TODO: consider `nixpkgs.source` always defaulting to `pinned`? That would bypass flake input following.
+      warnings =
+        lib.optional
+          (opt.source.highestPrio == optionDefaultPrio && pinned.narHash != cfg.source.narHash or null)
```
MattSturgeon: Maybe this warning should only show when `cfg.source` was used to construct `_module.args.pkgs`?

Not sure the exact condition to check for that, maybe something like `pkgs.path == (cfg.source.outPath or cfg.source)`?

nixpkgs.nix
```diff
@@ -0,0 +1,26 @@
...
+          shortRev = builtins.substring 0 7 info.rev;
+          lastModifiedDate = throw "lastModifiedDate is not available before Nix 2.4";
+          outPath = fetchTarball {
+            url = "https://api.${info.host or "github.com"}/repos/${info.owner}/${info.repo}/tarball/${info.rev}";
+            ${if info ? narHash then "sha256" else null} = info.narHash;
+          };
+        }
+      else
+        throw "fetchTree: unsupported type '${info.type}'"
+    );
```
MattSturgeon: Based on https://github.com/NixOS/flake-compat/blob/5edf11c44bc78a0d334f6334cdaf7d60d732daab/default.nix#L20-L36

tests/nixpkgs-fetch.nix
```diff
@@ -0,0 +1,37 @@
+{
+  self,
+  jq,
+  runCommandLocal,
+}:
+let
+  flake-input = self.inputs.nixpkgs.sourceInfo;
+  file-input = import ../nixpkgs.nix;
+in
+runCommandLocal "nixpkgs-fetch-test"
```
mightyiam: Is using `runLocal` for at least some category of tests a best practice? I'm not sure I can think of why this should not be subject for substitution. Perhaps for some sort of CLI invocation convenience that would make a certain flag unnecessary?

tests/nixpkgs-fetch.nix
```diff
@@ -0,0 +1,37 @@
...
+    flake_info = removeAttrs flake-input [ "outPath" ];
+    file_info = removeAttrs file-input [ "outPath" ];
+
+    nativeBuildInputs = [ jq ];
+  }
+  ''
+    mkdir "$out"
+
+    [ "$flake_drv" = "$file_drv" ] || {
+      echo 'Unexpected difference between `nixpkgs.nix` file and `inputs.nixpkgs` derivation hash' >&2
```
mightyiam: Is there a slight risk of `nixpkgs.nix` being wrong in the future if that file is moved?

> Unexpected difference between vanilla-fetched Nixpkgs and flake input Nixpkgs derivation hash

Interesting to note that `builtins.fetchTree` produces the same derivation as a flake input 🤔

tests/nixpkgs-fetch.nix
```diff
@@ -0,0 +1,37 @@
...
+      echo "'$flake_drv' vs '$file_drv'" >&2
+      exit 1
+    }
+
+    jq .flake_info "$NIX_ATTRS_JSON_FILE" > "$out/flake.json"
+    jq .file_info "$NIX_ATTRS_JSON_FILE" > "$out/file.json"
+
+    diff -q "$out/flake.json" "$out/file.json" || {
+      echo 'Unexpected difference between `nixpkgs.nix` file and `inputs.nixpkgs` flake input' >&2
```
mightyiam: Ditto

nixpkgs.nix
```diff
@@ -0,0 +1,26 @@
+/**
+  Fetches Nixvim's pinned revision of Nixpkgs without evaluating the flake.
+*/
+let
+  # builtins.fetchTree was added in Nix 2.4, so provide a fetchTarball-based polyfill
```
mightyiam: Does this project have a Nix support policy or at least a precedent for supporting such an old Nix? [November 2021](https://github.com/NixOS/nix/releases/tag/2.4).

flake/dev/fmt.nix
mightyiam: Would an alternative to this be to avoid naming the binding `fetchTree` and naming it `fetchTree_` or similar instead?

modules/top-level/nixpkgs.nix
MattSturgeon: Solved using `constructedByMe`: https://github.com/NixOS/nixpkgs/blob/8a3a8d8ed60085082c2a390bd33149c07e9e7ca4/nixos/modules/misc/nixpkgs.nix#L366-L374

wrappers/_shared.nix
```diff
@@ -50,6 +54,12 @@ let
     ];
   };
 
+  finalConfiguration =
+    options.programs.nixvim.valueMeta.configuration or (
+      # v2 check+merge was added 2025-08-28, halfway through the 25.11 cycle
+      throw "Nixvim must be evaluated with a recent revision of Nixpkgs lib (missing `valueMeta`)"
+    );
```
MattSturgeon: We probably don't need this `or throw`, but it handles the situation where the host module is from an old Nixpkgs channel while the Nixvim submodule is from a recent revision.

Happy to drop the `or` if preferred.

lib/overlay.nix
```diff
@@ -1,37 +1,4 @@
...
-  # Add Nixvim's section to the lib
-  nixvim = import ./top-level.nix { inherit flake lib; };
-
-  # Extend the maintainers set with Nixvim-specific maintainers
-  maintainers = prevLib.maintainers // import ./maintainers.nix;
-
-  # Extend lib.types with Nixvim's custom types
-  types = prevLib.types // import ./types.nix { inherit lib; };
+# Public non-flake entrypoint for Nixvim's Nixpkgs-lib overlay
+import ./overlay.internal.nix {
```
khaneliman, line 2: `lib/overlay.nix`: this changes the direct-import API shape from `{ flake } -> overlay` to `overlay`, so the previously documented form now fails.

Before this PR, the public file documented:

https://github.com/nix-community/nixvim/blob/13ba3f7ca9da3d49e959745f6d0d1e7a662c1984/lib/overlay.nix#L10-L13

This PR changes `lib/overlay.nix` to immediately return `import ./overlay.internal.nix { flake = null; }`:

https://github.com/nix-community/nixvim/blob/9584705d4f96d85a034087bb691a6831d551e5fc/lib/overlay.nix#L1-L4

I verified the old documented form against this PR head:

```sh
nix eval --impure --expr '
let
  nixvim = builtins.getFlake (toString ./.);
  pkgs = import nixvim.inputs.nixpkgs {};
  lib = pkgs.lib.extend (import ./lib/overlay.nix { flake = nixvim; });
in
builtins.hasAttr "evalNixvim" lib.nixvim
'
```

It fails with `attempt to call something which is not a function but a set`.

Can we preserve the old import form for compatibility, or intentionally document this as a breaking API change?

mightyiam:
I can try to review, so if you're interested then let me know when it's ready.

lib/overlay.nix
MattSturgeon: > This PR changes `lib/overlay.nix` to immediately return `import ./overlay.internal.nix { flake = null; }`:

Correct. That change was the main (only?) reason for splitting the file in two.

> so the previously documented form now fails.

Good point.

My thought process was that we didn't publicly document the previous `overlay.nix` file anywhere. In my view, this PR was the first to publicly document importing `overlay.nix`, promoting it from an internal file to a public entrypoing.

It _was_ shown in the example you reference though, which is visible in in-source documentation and `:doc` documentation (e.g. `:lf github:nix-community/nixvim` `:doc lib.overlay`), though, so I can see the "breaking change" argument.

I'd prefer to do it sooner than later, assuming we want to do it at all. If you still think it needs a "breaking change" announcement, where should that go?

> Can we preserve the old import form for compatibility, or intentionally document this as a breaking API change?

I don't think we can do that lazily; to provide a compatibility shim we'd have to evaluate the first argument passed in to detect wether it is `lib` or `{ flake = <...> }`.

MattSturgeon:
> I can try to review, so if you're interested then let me know when it's ready.

Thanks! I believe this is ready for review, https://github.com/nix-community/nixvim/pull/4311#discussion_r3267767311 not withstanding.

lib/overlay.nix
MattSturgeon: I experimented with:

```nix
# Public non-flake entrypoint for Nixvim's Nixpkgs-lib overlay

# Sadly, to provide backwards compatibility we must evaluate the first arg.
# This may be the internal set-pattern, or it may be the overlay's fixpoint.
#
# This breaks :doc, so we copy-paste the doccomment from overlay.internal.nix

let
  # Hacky check for "is this a nixpkgs-lib fixpoint"
  libAttrs = [ "meta" "misc" "trivial" ];

  fn = import ./overlay.internal.nix;
  fnArgs = builtins.functionArgs fn;
in

/**
  Overlay function extending the Nixpkgs `lib` with Nixvim's additions.
  ...
*/
arg:
let
  isFixpointArg = builtins.all (attr: builtins.hasAttr attr arg) libAttrs;
  isInternalArg = !isFixpointArg && builtins.intersectAttrs arg fnArgs == fnArgs;
  internalArg = if isInternalArg then arg else { flake = null; };
  overlay = fn internalArg;
in
if isFixpointArg then overlay arg else overlay
```

but this is infinitely recursive, since we're asking a extension fixpoint what attrs it has.

wrappers/_shared.nix
```diff
@@ -50,6 +54,12 @@ let
     ];
   };
 
+  finalConfiguration =
+    options.programs.nixvim.valueMeta.configuration or (
+      # v2 check+merge was added 2025-08-28, halfway through the 25.11 cycle
+      throw "Nixvim must be evaluated with a recent revision of Nixpkgs lib (missing `valueMeta`)"
```
khaneliman: Q: Do we care in master/26.05? Consumers are expected to be following matching releases, right?

khaneliman (approved):
Don't feel too strongly about the concept, but implementation seems to make sense.

nixpkgs.nix
mightyiam: Should this be a code comment, instead?

nixpkgs.nix
```diff
@@ -0,0 +1,27 @@
...
+          };
+        }
+      else
+        throw "fetchTree: unsupported type '${info.type}'"
+    );
+
+  lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  node = lock.nodes.${lock.nodes.${lock.root}.inputs.nixpkgs};
+in
+assert !(node.locked ? dir);
```
mightyiam, line 90: What's the `dir` attribute, please?

nixpkgs.nix
```diff
@@ -0,0 +1,27 @@
...
+fetchTree (node.info or { } // node.locked)
```
mightyiam: When does a node have `info`, please?

modules/top-level/nixpkgs.nix
```diff
@@ -231,6 +233,16 @@ in
...
+
+      # Whether `pkgs` was constructed by this module.
+      # This is false when any of nixpkgs.pkgs or _module.args.pkgs is set.
+      constructedByMe =
+        # We set it with default priority and it can not be merged, so if the
+        # pkgs module argument has that priority, it's from us.
+        (lib.modules.mergeAttrDefinitionsWithPrio options._module.args).pkgs.highestPrio
+        == lib.modules.defaultOverridePriority
+        # Although, if nixpkgs.pkgs is set, we did forward it, but we did not construct it.
+        && !opt.pkgs.isDefined;
```
mightyiam, line 245: Would an alternative to this be to avoid defining `pkgs` as `_module.args` directly and instead defining it via an option and then passing that option's value to `_module.args`? Or perhaps option values cannot be passed to `_module.args` 🤔

modules/top-level/nixpkgs.nix
```diff
@@ -256,5 +268,20 @@ in
...
+            constructedByMe
+            && opt.source.highestPrio == optionDefaultPrio
+            && pinned.narHash != cfg.source.narHash or null
+          )
+          ''
+            The `${opt.source}` default value has been affected by your flake input `follows`.
+            Nixvim's input is pinned Nixpkgs to `${pinned.rev}`.${
+              lib.optionalString (cfg.source ? rev) " Actual Nixpkgs is following `${cfg.source.rev}`."
+            }
+            Please remove your `inputs.nixvim.inputs.nixpkgs.follows` or explicitly define `${opt.source}` to suppress the warning.'';
```
mightyiam: I like the improved communication here!

flake/lib.nix
```diff
@@ -14,7 +14,7 @@
       # on pinning the flake-parts nixpkgs-lib to the nixpkgs pin
       inherit (inputs.nixpkgs) lib;
     };
-    overlay = import ../lib/overlay.nix {
+    overlay = import ../lib/overlay.internal.nix {
```
mightyiam, line 17: So, internal is used only for development?

lib/overlay.internal.nix
```diff
@@ -0,0 +1,42 @@
...
+/**
+  Overlay function extending the Nixpkgs `lib` with Nixvim's additions.
+
+  This function is intended to be passed to `lib.extend`.
+
+  # Examples
+
+  Importing the file directly:
+  ```
+  lib.extend (import ./overlay.nix)
```
mightyiam: Perhaps an expression that looks more like from a consumer's perspective, like `lib.extend (import (nixvim + "/lib/overlay.nix"))`?

mightyiam (commented):
My thoughts as far as `lib: allow instantiating without a flake instance`, for what they're worth. Will resume later, if you want to wait for me 🤷

wrappers/_shared.nix
MattSturgeon: Traditionally we required host and nixvim followed matching releases, yes. But, in theory, so long as Nixvim's `pkgs` and `lib` are from compatible versions, it shouldn't usually matter if the host is from an incompatible version.

This is one of a handful of cases where we are using the _host_ `lib` and are also using a _relatively_ recent feature.

Whether we care if a pre-August 2025 host can evaluate a nixvim submodule is an open question. nixos-25.11 will be end-of-life in ~3mo, and recent revisions of 25.11 should support `valueMeta`, so maybe not?

tests/nixpkgs-fetch.nix
MattSturgeon: [`preferLocalBuild`](https://nix.dev/manual/nix/2.34/language/advanced-attributes.html#adv-attr-preferLocalBuild) is useful for very cheap derivations, that are faster to build locally than to outsource to a remote builder.

[`allowSubstitutes`](https://nix.dev/manual/nix/2.34/language/advanced-attributes.html#adv-attr-allowSubstitutes) is a similar option; useful when it is faster to rebuild a derivation than to download its output from a binary substituter (like cachix or nixos-cache)

`runCommandLocal` (or passing `runLocal = true` to `runCommandWith`) configures these options:
https://github.com/NixOS/nixpkgs/blob/282deb2d89b6b60b4ca608d460b197e726d440a6/pkgs/build-support/trivial-builders/default.nix#L93-L94

tests/nixpkgs-fetch.nix
MattSturgeon: > Is there a slight risk of `nixpkgs.nix` being wrong in the future if that file is moved?

This is mostly for test coverage, to ensure we don't break it in future refactors. These kinda tests also help prove a new feature is doing what we expected.

> > Unexpected difference between vanilla-fetched Nixpkgs and flake input Nixpkgs derivation hash
> 
> Interesting to note that `builtins.fetchTree` produces the same derivation as a flake input 🤔

`fetchTree` was originally backported _from_ the flake branch to de-couple flake-style input fetching from flakes:

- https://github.com/NixOS/nix/pull/3459
- https://nix.dev/manual/nix/2.34/language/builtins.html#builtins-fetchTree

As is, this fetches in the `inputs.foo.flake = false` style, without evaluating the fetched source _as a flake_. Actual flake inputs have a second step where transitive inputs are fetched, the flake _outputs_ are evaluated, and the outputs are merged with the actual source.

https://github.com/NixOS/nix/blob/915772aa5ee56f639730cf616218aab5f8f68e07/src/libflake/call-flake.nix#L71-L89

nixpkgs.nix
MattSturgeon: > Does this project have a Nix support policy

No

> or at least a precedent for supporting such an old Nix? [November 2021](https://github.com/NixOS/nix/releases/tag/2.4).

As we have no policy, and no CI running against old nix versions, not really.

I mostly did this out of habbit, as we typically support Nix 2.18 in Nixpkgs (at least, in terms of language features).

I also didn't want to check if `fetchTree` was ever an "experimental feature" (I don't think it was, but I didn't check).

nixpkgs.nix
MattSturgeon: It's a flake input feature.

You can configure a flake input to be a sub-path within the fetched source.

E.g. `inputs.foo.url = "github:nix-community/nixvim?dir=flake/dev"` will fetch all of nixvim into the nix store, but the flake input will have `outPath = "${sourceInfo}/flake/dev"` and `flake/dev/flake.nix` will get evaluated instead of the top-level `flake.nix`.

This should never be used with our Nixpkgs input, and isn't supported by `nixpkgs.nix`, hence the assertion.

nixpkgs.nix
MattSturgeon: Good question. I don't know.

I copied this from flake-compat: https://github.com/NixOS/flake-compat/blob/5edf11c44bc78a0d334f6334cdaf7d60d732daab/default.nix#L212, however looking at upstream, it is described as obsolete and to be dropped: https://github.com/NixOS/nix/blob/915772aa5ee56f639730cf616218aab5f8f68e07/src/libflake/call-flake.nix#L53-L55

I'll drop it here, since it seems unlikely that any new `nix flake update` or `nix flake lock` will add it.

modules/top-level/nixpkgs.nix
MattSturgeon: This is copied from nixos: https://github.com/NixOS/nixpkgs/blob/0acbcf2ef991d8e757d989e263253119cb3fcb48/nixos/modules/misc/nixpkgs.nix#L366-L374

It is deliberately allowing end-users to override the constructed `pkgs` via _either_ `_module.args.pkgs = mkForce ...` or `nixpkgs.pkgs = ...`. If either of those things happen, then `constructedByMe` is false.

---

In theory, yes, we could have an intermediate `finalPkgs` option. However, that doesn't solve things.

1. We want to avoid evaluating nixpkgs instances that aren't in use; we solve that by evaluating override wrapper's priorities instead.
2. We want to keep our module in sync with NixOS's module, so copying its code is the preferred style.
3. Intermediate options are inefficient if they are not needed;
   - sometimes they add clarity, but they also increase API surface, meaning more things to keep track of, more things to decide whether they are public or internal, more things for nix to evaluate, etc.

flake/lib.nix
MattSturgeon: Sadly, nix doesn't have many ways to properly encapsulate files as "internal" or mark them as "public" (hence why I was cautious in your earlier PR about importing loose files directly from repos that may assume their flake is the only public API).

I'm hoping that `.internal.nix` is a clear enough signal that this file is an implementation detail not a public entrypoint, but in doing so we're also making a breaking change to the existing `overlay.nix`'s function signature, using the argument that it wasn't _publicly_ documented before to justify the breaking change. See https://github.com/nix-community/nixvim/pull/4311#discussion_r3267767311

We need the _internal_ file so that the public file doesn't need a `flake` argument. We need the `flake` argument internally, so that we can pass the already instantiated flake.

mightyiam: Unless we don't rely on a flake internally at all, but beggars can't be choosers.

lib/overlay.nix
mightyiam: Nixers love breaking changes 🙄 

Here's how I see it; a file named `overlay.nix` should really be just an overlay and not a function to one. So, this is a breaking fix 🤷

modules/top-level/nixpkgs.nix
mightyiam: Thank you. I wish we could have avoided this difficult expression, but there might not be a non-breaking way 🤷

modules/top-level/output.nix
```diff
@@ -351,7 +351,13 @@ in
...
         };
 
-        manDocsPackage = config.flake.packages.${system}.man-docs;
+        # TODO: make `build.manDocsPackage` available in non-flake evals.
+        # Also update `enableMan` default value.
+        manDocsPackage =
+          if options.flake.isDefined then
+            config.flake.packages.${system}.man-docs
+          else
+            throw "`${options.build.manDocsPackage}` currently requires `${options.flake}`, which is not defined.";
```
mightyiam: I'll take this trade-off.

mightyiam (commented):
Okay, I don't have much to add. I am grateful that you've picked this up, astounded at the level of complexity you've been processing here and looking forward to using it.

flake/lib.nix
MattSturgeon: Yes, it'd be much cleaner if the Nixvim configuration (and by extension, `lib.nixvim.evalNixvim`) didn't depend on the flake itself or any flake outputs.


If we can refactor everything using `config.flake`, then we can `mkRemovedOptionModule ["flake"]` in case anyone else is using it downstream.

That's a much bigger refactor/breaking change though. Blocked by:
- Make docs packages easier to evaluate without a `flake` instance.
- Change `nixpkgs.source`'s default from the flake input to the `nixpkgs.nix` pin.

Both of these are covered by TODO comments, but tracking issues or PRs would be good too.

This PR was at least able to fully separate wrapper modules from depending on a `flake` instance. But they _do_ still depend on the `extendModules` from a nixvim configuration, so the only way to practically get a wrapper module is _still_ via the flake or via a standalone nixvim configuration.

nixpkgs.nix
MattSturgeon: > I also didn't want to check if `fetchTree` was ever an "experimental feature" (I don't think it was, but I didn't check).

Ah, I just noticed [in the manual](https://nix.dev/manual/nix/2.34/language/builtins.html#builtins-fetchTree):

> This function is only available if the [`fetch-tree` experimental feature](https://nix.dev/manual/nix/2.34/development/experimental-features#xp-feature-fetch-tree) is enabled.

> The [`flakes`](https://nix.dev/manual/nix/2.34/development/experimental-features#xp-feature-flakes) feature flag always enables `fetch-tree`. This built-in was previously guarded by the `flakes` experimental feature because of that overlap.

Maybe I can come up with a test that does a nested eval with different experimental features configured within the sandbox

flake/dev/fmt.nix
MattSturgeon: No, that'd still suffer from `sema-primop-removed-prefix` because `fetchTree` can normally be used without the `builtins.` prefix.

MattSturgeon:
I tried to come up with a test that runs nix with different `experimental-features` in a sandbox with a dummy store. However I'm running into issues with `fetchTree` and `fetchTarball` being too eager to fetch (rather than being lazy or even using substituters), so I'll probably have to go back to something closer to the simplified test we had before.

I'll draft this for now as I have to step away from the PC.

modules/top-level/nixpkgs.nix
```diff
@@ -188,7 +192,7 @@ in
     # NOTE: This is a nixvim-specific option; there's no equivalent in nixos
     source = lib.mkOption {
       type = lib.types.path;
```
mightyiam, line 194: Will this type accept any store object?

MattSturgeon:
```
nix-repl> :p types.path.description
absolute path

nix-repl> types.path.check ./foo
true

nix-repl> types.path.check "/nix/store/foo"
true

nix-repl> types.path.check { outPath = "/nix/store/foo"; }
true
```

mightyiam:
Thank you so much, @MattSturgeon :pray: :bowing_man: :1st_place_medal:

mightyiam:
[Here's my personal usage](https://github.com/mightyiam/infra/commit/3fcd3f8de6218a6ce4c30a43f35753e95f825f26)!

Issue #3326, opened 2025-05-14
[PLUGIN REQUEST] vectorcode for codecompanion


auipga:
| Field    | Description |
|----------|-------------|
| Plugin   | `vectorcode`  | 
| Homepage | [Davidyz/VectorCode](https://github.com/Davidyz/VectorCode)     | 
| Nixpkgs  | `vectorcode` | 

## Extra Information

- https://github.com/Davidyz/VectorCode/blob/main/docs/neovim.md
- https://github.com/Davidyz/VectorCode/wiki/Neovim-Integrations#olimorriscodecompanionnvim
- https://github.com/olimorris/codecompanion.nvim/blob/main/doc/extensions/vectorcode.md

```nix
# Requirements
plugins.codecompanion.enable = true;

# Configure the extension
plugins.codecompanion.settings = {
  extensions.vectorcode = {
    enabled = true;
    opts = {
      add_tool = true; # the @vectorcode tool becomes available in the CodeCompanion chat buffer
      add_slash_command = true; # the /vectorcode slash command
      tool_opts = {};
    };
  };

# TODO: make this work
#  context.providers = [ ] ++ [
#    { name = "vectorcode"; opts = { top_k = 15; }; }
#  ];

  hooks = {
    on_pre_send.__raw = ''
      function(payload)
        if payload.dollar_cost and payload.dollar_cost > 1.00 then
          vim.notify("Request aborted: cost ≥ $1", vim.log.levels.WARN)
          return false   -- cancel
        end
      end,
    '';
  };
};
```

PR #4329, opened 2026-05-21
plugins/vectorcode: add codecompanion integration

khaneliman:
Forgot to include in the previous PR 

Related https://github.com/nix-community/nixvim/issues/3326
Related https://github.com/nix-community/nixvim/pull/4326

### DIFF
--- a/plugins/by-name/vectorcode/default.nix
+++ b/plugins/by-name/vectorcode/default.nix
@@ -1,4 +1,8 @@
-{ lib, ... }:
+{
+  config,
+  lib,
+  ...
+}:
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "vectorcode";
   package = "vectorcode-nvim";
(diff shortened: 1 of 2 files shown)


plugins/by-name/vectorcode/default.nix
```diff
@@ -1,4 +1,12 @@
-{ lib, ... }:
+{
+  config,
+  lib,
+  ...
+}:
+let
+  cfg = config.plugins.vectorcode;
```
MattSturgeon: Can you move this to `extraConfig`, or is it used somewhere else I'm not seeing?

plugins/by-name/vectorcode/default.nix
```diff
@@ -16,4 +24,41 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
+          };
+        };
+        description = ''
+          Options passed to CodeCompanion's vectorcode extension.
+        '';
+      };
+    };
+  };
+
+  extraConfig = {
```
MattSturgeon:
```suggestion
  extraConfig = cfg: {
```

plugins/by-name/vectorcode/default.nix
```diff
@@ -16,4 +24,41 @@ lib.nixvim.plugins.mkNeovimPlugin {
   dependencies = [
     "vectorcode"
   ];
+
+  extraOptions = {
+    integrations.codecompanion = {
+      enable = lib.mkEnableOption "the vectorcode extension for CodeCompanion";
+
+      settings = lib.mkOption {
+        type = types.attrsOf types.anything;
```
MattSturgeon: We don't normally define `settings` manually in `extraOptions`, why do we need to here?

If we're doing it, could you use nixvim's new `luaTypes.anything` or `luaTypes.tableOf luaTypes.anything`? Or will this not end up being serialized as lua?

khaneliman: > We don't normally define settings manually in extraOptions, why do we need to here?

A suboption with it's own configuration, it's not the parent module's configuration. Allows configuring parent module and the codecompanion integration. 

> If we're doing it, could you use nixvim's new luaTypes.anything or luaTypes.tableOf luaTypes.anything? Or will this not end up being serialized as lua?

What does that give us ? I don't remember using it before

MattSturgeon: > > We don't normally define settings manually in extraOptions, why do we need to here?
> 
> A suboption with it's own configuration, it's not the parent module's configuration. Allows configuring parent module and the codecompanion integration. 

Ah, right. I missed that this was     `integrations.codecompanion.settings` instead of just `settings`.


> What does [`luaTypes`] give us ? I don't remember using it before

Nothing much. Slightly stricter/more specific type checks, potential for more specific merge logic in the future, more specific type documentation.

Your right that we don't use it much, it's currently only used in a few places as we haven't put much work into migration.

MattSturgeon (approved):

PR #4333, opened 2026-05-21
modules/output: load wrapped config via VIMINIT

### LINKED ISSUE #3506: [BUG] `exrc` option not working for `makeNixvimWithModule`?
| Nixpkgs      | `2631b0b7abcea6e640ce31cd78ea58910d31e650` |

## Description

Afaik everything necessary to source a local `.nvim.lau` file is to have the `exrc` option set and the just run a `nvim .` in the directory where its placed. However, I can't get the file to be sourced with my nixvim instance using `nixvim.makeNixvimWithModule`...
Using this option, the file generated and added as `nvim` to the path is a bash script ending with something like this:
```bash
  exec "/nix/store/q570k9f80p21sbr4x74vbggjxhg82brz-neovim-unwrapped-0.11.2/bin/nvim"  --cmd "lua vim.g.loaded_node_provider=0;vim.g.loaded_perl_provider=0;vim.g.loaded_pytho
  n_provider=0;vim.g.loaded_python3_provider=0;vim.g.loaded_ruby_provider=0" --cmd "set packpath^=/nix/store/3wb9jlpx5i85wj4h7y61wwsms2mn7995-vim-pack-dir" --cmd "set rtp^=/n
  ix/store/3wb9jlpx5i85wj4h7y61wwsms2mn7995-vim-pack-dir" -u /nix/store/sav787yqb8z271hplrxdfp6rwq7mw5hb-init.lua "$@"
```

Can anyone confirm or contradict this claim? Would be really nice to know if this is a Nixvim problem or one related to my own config...

```nix
programs.nixvim = {
 options = {
    exrc = true;
    secure = true;
  };
}
```

```shell
$ l --group .nvim.lua
Permissions Size User Group Date Modified Name
.rw-r--r--   152 jnnk users 24 Jun 21:14  .nvim.lua
```

khaneliman:
Wrapped Nixvim configs previously passed the generated init.lua with -u. Neovim treats that startup path as an explicit user config and skips exrc processing, so project-local .nvim.lua files were not sourced even when users enabled the exrc option.

Load the generated config through a forced VIMINIT instead. This preserves the current wrapped-config precedence while allowing Neovim's normal exrc startup path to run for trusted project config files.

When impureRtp is disabled, clear XDG_CONFIG_DIRS and VIM from an early --cmd before sysinit is checked, then restore them before loading the generated config so runtime code and child processes see the original environment.

Keep the saved startup environment in a short-lived global Lua table shared between the --cmd and VIMINIT chunks, then clear it after restore.

Add a modules-output regression test that checks the wrapper args no longer include -u, verifies VIMINIT is set, confirms trusted .nvim.lua is sourced, blocks sysinit.vim, verifies XDG_CONFIG_DIRS and VIM are restored, and ensures the saved startup state is cleared.

Closes #3506

### DIFF
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -420,6 +420,13 @@ in
       wrapRc = lib.mkOverride 1501 true;
       impureRtp = lib.mkOverride 1501 false;
 
+      assertions = lib.nixvim.mkAssertions "output" [
+        {
+          assertion = config.impureRtp || config.wrapRc;
+          message = "`impureRtp = false` requires `wrapRc = true` so Nixvim can suppress system/XDG startup config.";
+        }
+      ];
+
       extraConfigLuaPre = lib.mkOrder 100 (
         lib.concatStringsSep "\n" (
           lib.optional (!config.impureRtp) ''
--- a/modules/top-level/output.nix
+++ b/modules/top-level/output.nix
@@ -77,7 +77,7 @@ in
       description = ''
         Whether the config will be included in the wrapper script.
 
-        When enabled, the Nixvim config will be passed to `nvim` using the `-u` option.
+        When enabled, the Nixvim config will be passed to `nvim` using the `VIMINIT` environment variable.
       '';
       defaultText = lib.literalMD ''
         Configured by your installation method: `false` when using the Home Manager module, `true` otherwise.
@@ -90,6 +90,7 @@ in
         Whether to keep the (impure) nvim config directory in the runtimepath.
 
         If disabled, the XDG config dirs `nvim` and `nvim/after` will be removed from the runtimepath.
+        When `wrapRc` is enabled, the wrapper will also ignore system/XDG config dirs during startup.
       '';
       defaultText = lib.literalMD ''
         Configured by your installation method: `true` when using the Home Manager module, `false` otherwise.
@@ -276,6 +277,48 @@ in
         else
           initSource;
 
+      # `:help initialization` says sysinit is sourced before `VIMINIT` from
+      # `$XDG_CONFIG_DIRS/nvim/sysinit.vim` or, as fallback, `$VIM/sysinit.vim`.
+      impureStartupEnvVars = [
+        "XDG_CONFIG_DIRS"
+        "VIM"
+      ];
+
+      # Runs via `--cmd` before sysinit, so system/XDG config dirs are hidden
+      # before Nvim checks them without relying on makeWrapper-only `--run`.
+      saveAndClearImpureStartupEnvVars = ''
+        _G.__nixvim_impure_startup_env = {}
+        for _, name in ipairs(${lib.nixvim.toLuaObject impureStartupEnvVars}) do
+          local value = vim.env[name]
+          if value ~= nil then
+            _G.__nixvim_impure_startup_env[name] = {
+              set = true,
+              value = value,
+            }
+          else
+            _G.__nixvim_impure_startup_env[name] = {
+              set = false,
+            }
+          end
+          vim.env[name] = nil
+        end
+      '';
+
+      # Restore before loading the generated config so user code sees the original
+      # environment while sysinit remains skipped.
+      restoreImpureStartupEnvVars = ''
+        local saved_impure_startup_env = _G.__nixvim_impure_startup_env or {}
+        for _, name in ipairs(${lib.nixvim.toLuaObject impureStartupEnvVars}) do
+          local saved = saved_impure_startup_env[name]
+          if saved ~= nil and saved.set then
+            vim.env[name] = saved.value
+          else
+            vim.env[name] = nil
+          end
+        end
+        _G.__nixvim_impure_startup_env = nil
+      '';
+
       extraWrapperArgs = builtins.concatStringsSep " " (
         # Setting environment variables in the wrapper
         (lib.mapAttrsToList (
@@ -292,7 +335,21 @@ in
         ++ (optional (
           config.extraPackagesAfter != [ ]
         ) ''--suffix PATH : "${lib.makeBinPath config.extraPackagesAfter}"'')
-        ++ (optional config.wrapRc ''--add-flags -u --add-flags "${initFile}"'')
+        ++ (optional (config.wrapRc && !config.impureRtp) (
+          lib.escapeShellArgs [
+            "--add-flag"
+            "--cmd"
+            "--add-flag"
+            "lua ${saveAndClearImpureStartupEnvVars}"
+          ]
+        ))
+        ++ (optional config.wrapRc (
+          lib.escapeShellArgs [
+            "--set"
+            "VIMINIT"
+            "lua ${lib.optionalString (!config.impureRtp) restoreImpureStartupEnvVars} dofile('${initFile}')"
+          ]
+        ))
       );
 
       package =
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -150,6 +150,127 @@
     '';
   };
 
+  wrapRc-uses-viminit-for-exrc =
+    {
+      config,
+      lib,
+      pkgs,
+      ...
+    }:
+    let
+      wrapperArgs = config.build.nvimPackage.wrapperArgs;
+      exrcTest =
+        pkgs.runCommandLocal "wraprc-viminit-exrc-test"
+          {
+            nativeBuildInputs = [ config.build.nvimPackage ];
+          }
+          ''
+            mkdir -p project home xdg/nvim vim
+
+            cat > project/.nvim.lua <<'EOF'
+            vim.g.nixvim_exrc_loaded = true
+            EOF
+
+            cat > project/check.lua <<'EOF'
+            assert(vim.g.nixvim_exrc_loaded == true, ".nvim.lua was not sourced")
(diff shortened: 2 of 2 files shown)


modules/top-level/output.nix
```diff
@@ -276,6 +277,52 @@ in
...
 
+      # `:help initialization` says sysinit is sourced before `VIMINIT` from
+      # `$XDG_CONFIG_DIRS/nvim/sysinit.vim` or, as fallback, `$VIM/sysinit.vim`.
+      impureStartupEnvVars = [
+        "XDG_CONFIG_DIRS"
+        "VIM"
+      ];
+
+      originalEnvVar = name: "__NIXVIM_ORIGINAL_${name}";
+      originalEnvVarSet = name: "__NIXVIM_ORIGINAL_${name}_SET";
```
MattSturgeon: This is clever. But it's kinda rubbing me the wrong way.

I wonder if we could do something like:
1. store the "saved" env in a bash associative array
2. or store the saved env in a `key=value` file
3. or store the saved env as json

(an associative array is probably neatest, easy to unset, doesn't get exported, etc)

EDIT: ah, I see this is unset in the wrapper and re-set in lua? So maybe passing the originals in as JSON is best, assuming lua is able to deserialize JSON

modules/top-level/output.nix
```diff
@@ -276,6 +277,52 @@ in
...
+
+      restoreEnvVar =
+        name:
+        let
+          original = originalEnvVar name;
+          originalSet = originalEnvVarSet name;
+        in
+        lib.concatStringsSep " " [
+          ''if vim.env.${originalSet} == "1" then''
+          "vim.env.${name} = vim.env.${original}"
```
MattSturgeon: This assumes `name` and `originalSet` are valid lua identifiers. Safe for most env vars, but not safe if we ever expand this to the general case. I assume that's not a goal here, though.

modules/top-level/output.nix
```diff
@@ -292,7 +339,29 @@ in
...
+            ]
+          ) impureStartupEnvVars
+        ))
+        ++ (optional config.wrapRc (
+          lib.escapeShellArgs [
+            "--set"
+            "VIMINIT"
+            "lua ${lib.optionalString (!config.impureRtp) restoreImpureStartupEnvVars} dofile('${initFile}')"
+          ]
+        ))
```
MattSturgeon: Seems every element in this `optionals` is using `escapeShellArgs`, so we can pull that up and wrap the entire `optionals` in a single call to `escapeShellArgs` to reduce noise.

There are also a few cases of unnecessary parens, usually nixfmt makes it pretty clear that `++` doesn't need parens around its operands.

Some of these multi-line lists _may_ be allowed to be on one line by nixfmt, if you try. Depending on whether the internal heuristics are happy.

modules/top-level/output.nix
khaneliman: Good call. I kept the pre-startup shell save, but now pass the restore metadata to Lua as JSON and restore via `vim.env[...]`.

modules/top-level/output.nix
khaneliman: similar to above, i believe this is safe with bracket access

modules/top-level/output.nix
```diff
@@ -276,6 +277,52 @@ in
...
+      savedImpureStartupEnv = map (name: {
+        inherit name;
+        original = originalEnvVar name;
+        originalSet = originalEnvVarSet name;
+      }) impureStartupEnvVars;
+
+      # Restore before loading the generated config so user code sees the original
+      # environment while sysinit remains skipped.
+      restoreImpureStartupEnvVars = ''
+        for _, saved in ipairs(vim.json.decode(${builtins.toJSON (builtins.toJSON savedImpureStartupEnv)})) do
```
MattSturgeon: Ah, if I'd realised we were embedding the savedImpureStartupEnv directly into lua code, I would've suggested `toLuaObject` instead to `toJSON`. To me, `toJSON` makes more sense if it's getting saved to a separate file.

I'm also unsure why we've got _two_ `toJSON`s here?

modules/top-level/output.nix
```diff
@@ -292,7 +339,26 @@ in
...
+          lib.escapeShellArgs (
+            [
+              "--run"
+              (lib.concatMapStrings saveEnvVar impureStartupEnvVars)
+            ]
+            ++ lib.concatMap (name: [
+              "--set"
+              name
+              ""
+            ]) impureStartupEnvVars
```
MattSturgeon:
```suggestion
            ++ lib.concatMap (name: [ "--unset" name ]) impureStartupEnvVars
```

https://github.com/NixOS/nixpkgs/blob/509878c3cebcbcc803bdafc551a84146f4c4e3ae/pkgs/by-name/ma/makeBinaryWrapper/make-binary-wrapper.sh#L26

https://github.com/NixOS/nixpkgs/blob/509878c3cebcbcc803bdafc551a84146f4c4e3ae/pkgs/build-support/setup-hooks/make-wrapper.sh#L22

modules/top-level/output.nix
khaneliman: Yeah.. i thought it was weird but just did it anyways lol

modules/top-level/output.nix
```diff
@@ -292,7 +339,25 @@ in
         ++ (optional (
           config.extraPackagesAfter != [ ]
         ) ''--suffix PATH : "${lib.makeBinPath config.extraPackagesAfter}"'')
-        ++ (optional config.wrapRc ''--add-flags -u --add-flags "${initFile}"'')
+        ++ (optional (config.wrapRc && !config.impureRtp) (
+          lib.escapeShellArgs (
+            [
+              "--run"
+              (lib.concatMapStrings saveEnvVar impureStartupEnvVars)
```
MattSturgeon: FYI `--run` is exclusive to `makeWrapper`, it's not supported by `makeBinaryWrapper`.

If us or nixpkgs ever change the backend this could break.

modules/top-level/output.nix
```diff
@@ -292,7 +339,25 @@ in
         ++ (optional (
           config.extraPackagesAfter != [ ]
         ) ''--suffix PATH : "${lib.makeBinPath config.extraPackagesAfter}"'')
-        ++ (optional config.wrapRc ''--add-flags -u --add-flags "${initFile}"'')
+        ++ (optional (config.wrapRc && !config.impureRtp) (
```
MattSturgeon, line 338: I can't see where we handle the `!wrapRc && impureRtp` permutation; does that need special handling?

modules/top-level/output.nix
khaneliman: Maybe we just move the save/clear back into `--cmd lua...` and avoid the `makeXWrapper` dependency issue. Looks like that should still run before `sysinit` so we can clear early enough.

MattSturgeon: Would that mean we can avoid the hacky temporary env variables? I like the _idea_ or having both sides using lua.

I'm not sure how a `save` lua script would be able to expose the saved variables to the `restore` script in `VIMINIT`; IIUC they will be in isolated environments?

modules/top-level/output.nix
khaneliman: Not sure if I understand question fully, but the idea of the condition here was to make `impureRtp` be the deciding factor in whether to "allow" those impure env variables to affect the `sysinit` phase.

modules/top-level/output.nix
```diff
@@ -292,7 +339,21 @@ in
...
+          lib.escapeShellArgs [
+            "--add-flag"
+            "--cmd"
+            "--add-flag"
+            "lua ${saveAndClearImpureStartupEnvVars}"
```
MattSturgeon, line 343: How do we test that `vim.env[saved.originalSet] = foo` actually leaks out of this `--cmd` run and into the sysinit and viminit environments? Do your "sysinit.vim was sourced" tests cover this?

modules/top-level/output.nix
MattSturgeon: Oops, I think I meant `!wrapRc && !impureRtp`. I only see code in the `wrapRc && !impureRtp` permutation.

modules/top-level/output.nix
khaneliman: Actually, I like that... We can just use a global table and save / restore from that I think.

modules/top-level/output.nix
khaneliman: Oh, I don't think so. If we don't wrap, then we don't control that startup in wrapper args anyways.

MattSturgeon: Probably a pre-existing issue then, unrelated to this PR. Maybe we should have an assertion if `impureRtp` is disabled without `wrapRc` being enabled?

khaneliman: Yeah, I think that makes sense. Otherwise, someone might think we are purifying their path when we really don't.

modules/top-level/output.nix
khaneliman: Updated the test after restructuring to use the lua table. It now checks the variables that point at sysinit files and then verify we don't source those files and that we repopulate the variables and clear that temp table.

MattSturgeon (approved):
Thanks for getting this off the backlog!

jfly:
In case this saves anyone from the marathon debugging session I just went through: one unfortunate consequence of this change is that it exposes a latent bug in neovim where explicitly set `background`s now get clobbered by neovim's `background` autodetection logic: https://github.com/neovim/neovim/issues/40631.

PR #4338, opened 2026-05-23
plugins/language-servers: fix warning about `propagatedBuildInputs`

kybe236:
```
warning: Dependency of package 'python3.13-python-lsp-server-1.14.0' uses a nested list in attribute 'propagatedBuildInputs'.
This is deprecated as of Nixpkgs release 26.05, and support will be removed in a future nixpkgs release.
```

in my case `propagatedBuildInputs` contained empty lists which this also fixes.

### DIFF
--- a/plugins/lsp/language-servers/pylsp.nix
+++ b/plugins/lsp/language-servers/pylsp.nix
@@ -508,7 +508,7 @@ in
         inherit (cfg.settings) plugins;
 
         nativePlugins =
-          (map
+          (lib.concatMap
             (
               pluginName:
               (lib.optionals (isEnabled plugins.${pluginName}) cfg.package.optional-dependencies.${pluginName})


plugins/lsp/language-servers/pylsp.nix
```diff
@@ -507,7 +507,7 @@ in
         isEnabled = x: (x.enabled != null && x.enabled);
         inherit (cfg.settings) plugins;
 
-        nativePlugins =
+        nativePlugins = lib.lists.flatten (
           (map
```
SuperSandro2000: I would rather recommend to use concatMap.

SuperSandro2000:
```suggestion
        nativePlugins =
          (lib.concatMap
```

plugins/lsp/language-servers/pylsp.nix
```diff
@@ -526,7 +526,8 @@ in
           )
           ++ (lib.optionals (
             (isEnabled plugins.rope_autoimport) || (isEnabled plugins.rope_completion)
-          ) cfg.package.optional-dependencies.rope);
+          ) cfg.package.optional-dependencies.rope)
+        );
```
SuperSandro2000:
```suggestion
          ) cfg.package.optional-dependencies.rope);
```

GaetanLepage (approved):
Thanks!

PR #4342, opened 2026-05-24
flake/dev: replace `dev-nixpkgs` hack with local-path flake input

MattSturgeon:
- **flake/dev: drop dev-nixpkgs for nixvim path ref**
- **flake/dev: have nuschtosSearch systems input follow nixivm's**
- **flake/dev: mark flake-compat as `flake = false`**

Simplifies the update script and our mental model of how dev-inputs are managed.

Instead of "manually" syncing two `nixpkgs` inputs (root `nixpkgs` and dev `dev-nixpkgs`), we have the dev flake inherit all of the root flake's inputs via a `../..` path reference.

Internally, a relative-path flake reference within the same root flake is seen as a path-offset within the same fixed-output-derivation. If one was to compare the root `nixvim`, `nixvim/flake/dev`, and `nixvim/flake/dev`'s `nixvim` input, they would find:
- All have identical `sourceInfo` derivations
- The root flake's `outPath` matches `sourceInfo`
- `nixvim/flake/dev`'s `outPath` is `"${sourceInfo}/flake/dev`
- `nixvim/flake/dev`'s `nixvim` input's `outPath` is `"${sourceInfo}/flake/dev/../..`
- All three therefore refer to the same store object

When running `nix flake lock ./flake/dev`, the root nixvim flake's inputs are synced into `flake/dev/flake.lock` (just like any other flake input's inputs). It is therefore still important to run `nix flake lock ./flake/dev` after running `nix flake lock`, to keep both in sync; our CI update script is still responsible for this.

As a side-effect, we now have easier access to other root-nixvim inputs, like our `systems` input. This allows `nuschtosSearch.inputs.flake-utils.inputs.systems` to follow it.

We held off on doing this sooner as nix and flake-compat have historically had flaky support for relative-path flake inputs. These days, support is good enough for our needs.

cc @mightyiam: this will conflict with #4294, sorry.

### DIFF
--- a/ci/update.nix
+++ b/ci/update.nix
@@ -108,12 +108,8 @@ writeShellApplication {
     fi
 
     # Update the dev lockfile
-    root_nixpkgs=$(nix eval --raw --file . 'inputs.nixpkgs.rev')
     echo "Updating dev lockfile"
-    nix flake update \
-        --override-input 'dev-nixpkgs' "github:NixOS/nixpkgs/$root_nixpkgs" \
-        --flake './flake/dev' \
-        2> >(tee "$dev_update" >&2)
+    nix flake update --flake './flake/dev' 2> >(tee "$dev_update" >&2)
     cleanUpdateOutput "$dev_update" > "$dev_msg"
     if [ -s "$dev_msg" ]; then
       {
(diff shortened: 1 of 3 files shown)


khaneliman (approved):
Forgot we didn't migrate to this approach already.

PR #4345, opened 2026-05-24
modules: drop `helpers` alias

MattSturgeon:
It has been soft-deprecated for years now, but the warning was introduced just before 25.11 was branched off. We've therefore had the warning for two stable releases and ~1 unstable cycle.

Normally I'd advocate for having another release cycle with a `helpers = throw` stub, but I think in this case it's very unlikely anyone is still using nixvim's `helpers` module argument, so it's better to remove `specialArgs.helpers` entirely, in case an end-user wants to define their own unrelated `_module.args.helpers`.\*

_\*The original motivation for deprecating `helpers` was that the name was too generic for a shared namespace._

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -48,12 +48,6 @@ in
         # however see https://github.com/nix-community/nixvim/issues/2879
         inherit lib;
         modulesPath = ../modules;
-        # TODO: deprecated 2025-11-19
-        helpers = lib.warn ''
-          nixvim: the `helpers` module arg has been renamed to `lib.nixvim`.
-          Nixvim modules can access this via the `lib` module arg.
-          For wrapper modules (e.g. NixOS or Home Manager modules), see:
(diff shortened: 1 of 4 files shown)


MattSturgeon:
I'm gonna draft this until Nixvim bumps to 26.11-unstable; I'd like Nixvim's 26.05-stable to still have the warning.

khaneliman:
I agree, just mark it ready and ping when we do branch off.

MattSturgeon:
> when we do branch off.

We're technically ready to branch off now. The next (successful) flake bump will likely move `main` onto 26.11, as nixpkgs has already done its branch-off.

MattSturgeon:
> We're technically ready to branch off now.

I've created the `nixos-26.05` branch, so if we don't backport this it's _effectively_ targeting 26.11.

I don't mind waiting to merge until we've actually updated our nixpkgs to a 26.11 revision, if preferred.

khaneliman:
i'm fine with early branch off and just merging if we're able

khaneliman (approved):

PR #4349, opened 2026-05-26
plugins/auto-dark-mode: init

vimlinuz:


vimlinuz:
`vimPlugins.auto-dark-mode-nvim` yet to be available in `nixos-25.11`

PR #4350, opened 2026-05-27
plugins/wildfire.nvim: init

FilippoAiraldi:
Adds [wildfire.nvim](https://github.com/sustech-data/wildfire.nvim).

First time contributing here, so please let me know if anything needs adjusting :)

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -120,6 +120,12 @@
     githubId = 50964650;
     name = "gwen";
   };
+  fairaldi = {
+    email = "filippo.airaldi@proton.me";
+    github = "FilippoAiraldi";
+    githubId = 48798518;
+    name = "Filippo Airaldi";
+  };
   fwastring = {
     email = "fredrik@wastring.com";
     github = "fwastring";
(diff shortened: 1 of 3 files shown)


lib/maintainers.nix
```diff
@@ -120,6 +120,12 @@
     githubId = 50964650;
     name = "gwen";
   };
+  fairaldi = {
+    email = "filippo.airaldi@proton.me";
+    github = "FilippoAiraldi";
+    githubId = 48798518;
+    name = "Filippo Airaldi";
+  };
```
MattSturgeon, line 128: We prefer this to be added first in its own commit, then the plugin gets added second in another commit.

FilippoAiraldi: @MattSturgeon do you mean to split the changes in two separate PRs? Or just to separate this PR into two commits?

MattSturgeon: Two commits, one PR.

In theory if you had multiple PRs, they could all be based on your "`maintainers: add fairaldi`" commit and it wouldn't matter which order we merged in.

FilippoAiraldi: Done!

MattSturgeon (approved):
SGTM, thanks. Waiting for treefmt CI.

PR #4372, opened 2026-06-03
plugins.jujutsu: initialise

auscyber:
add jujutsu-nvim as a plugin.

yes i know a lil weird to have the settings as such, but its worth having checking of keymaps as they are in a weird format, and also being able to enable diff viewer on demand as shown. the other option yes not needed necessarily, however it makes such if im defining the others

### DIFF
--- a/plugins/by-name/jujutsu/default.nix
+++ b/plugins/by-name/jujutsu/default.nix
@@ -1,0 +1,116 @@
+{ lib, ... }:
+let
+  inherit (lib) mkOption;
+
+  actions = {
+    "show_help" = "Show keybindings help window";
+    "quit" = "Close the log window";
+    "jump_to_next_change" = "Navigate to next change";
+    "jump_to_prev_change" = "Navigate to previous change";
+    "jump_to_current_change" = "Navigate to the currently edited change";
+    "refresh" = "Refresh the log view";
+    "undo" = "Undo the last operation";
+    "set_revset" = "Open log with custom revset";
+    "switch_diff_viewer" = "Switch between diff viewer presets";
+    "show_global_flags" = "Show menu to toggle global jj flags";
+    "open_diff" = "Open diff viewer for change";
+    "describe" = "Edit change description";
+    "new_change" = "Create new change";
+    "new_change_menu" = "Show new change options menu";
+    "abandon_changes" = "Abandon change(s)";
+    "absorb_changes" = "Absorb change(s)";
+    "edit_change" = "Check out change";
+    "edit_change_menu" = "Show edit change options menu";
+    "rebase_change" = "Rebase change";
+    "squash_change" = "Squash change";
+    "squash_to_target" = "Squash to specific target";
+    "bookmark_change" = "Set/create bookmark";
+    "bookmark_menu" = "Show bookmark operations";
+    "push_bookmarks" = "Push bookmarks";
+    "push_bookmarks_and_create" = "Push bookmarks with --allow-new";
+    "git_fetch" = "Fetch from remote";
+    "pull_bookmark" = "Pull bookmark from remote";
+    "toggle_change" = "Toggle multi-selection";
+    "clear_selections" = "Clear all selections";
+  };
+
+  keys = lib.nixvim.keymaps.mkMapOptionSubmodule {
+    key = false;
+
+    action = false;
+    defaults.mode = "";
+    extraOptions = {
+      cmd = mkOption {
+        type = lib.types.maybeRaw (lib.types.enum (builtins.attrNames actions));
+        description = "Command to execute";
+      };
+
+      desc = mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+      };
+    };
+
+    extraModules = [
+      (
+        { config, lib, ... }:
+        let
+          isBuiltin = actions ? config.action;
+
+          defaultDesc = if isBuiltin then actions.${config.action} else null;
+        in
+        {
+          config = {
+            desc = lib.mkDefault defaultDesc;
+          };
+        }
+      )
+    ];
+  };
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "jujutsu";
+  moduleName = "jujutsu-nvim";
+  package = "jujutsu-nvim";
+
+  maintainers = [ lib.maintainers.auscyber ];
+  extraConfig =
+    cfg:
+    let
+      possible-sources = [
+        "diffview"
+        "codediff"
+      ];
+    in
+    {
+      plugins = lib.genAttrs' possible-sources (name: {
+        inherit name;
+        value.enable = lib.mkIf (cfg.settings.diff_preset == name) true;
+      }
+
+      );
+    };
+
+  settingsOptions = {
+    diff_preset = lib.nixvim.defaultNullOpts.mkEnum [
+      "difftastic"
+      "diffview"
+      "codediff"
+      "none"
+    ] "none" "diff view preset";
+    help_position = lib.nixvim.defaultNullOpts.mkEnum [
+      "center"
+      "bottom_right"
+    ] "center" "The help window (opened with ?) can be positioned in different locations";
+    keymap = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.either keys lib.types.bool);
+      description = "Keymaps for jujutsu-nvim";
+      default = { };
+
+    };
+
+    # TODO provide an example for the `settings` option (or remove entirely if there is no useful example)
+    # NOTE you can use `lib.literalExpression` or `lib.literalMD` if needed
+
+  };
+}
--- a/tests/test-sources/plugins/by-name/jujutsu/default.nix
+++ b/tests/test-sources/plugins/by-name/jujutsu/default.nix
@@ -1,0 +1,20 @@
+{
+  empty = {
+    plugins.jujutsu.enable = true;
+  };
+
+  defaults = {
+    plugins.jujutsu = {
+      enable = true;
+      settings = {
+        keymap = {
+          "?" = {
+            cmd = "show_help";
+            desc = "Show keybindings help";
+          };
+        };
+      };
+
+    };
+  };
+}


auscyber:
once [difftastic-nvim](https://github.com/NixOS/nixpkgs/pull/526605)  is in unstable, i'll add difftastic-nvim as well

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    ] "center" "The help window (opened with ?) can be positioned in different locations";
+    keymap = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.either keys lib.types.bool);
+      description = "Keymaps for jujutsu-nvim";
+      default = { };
+
+    };
+
+    # TODO provide an example for the `settings` option (or remove entirely if there is no useful example)
+    # NOTE you can use `lib.literalExpression` or `lib.literalMD` if needed
```
MattSturgeon, line 113: This needs addressing.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    ] "none" "diff view preset";
+    help_position = lib.nixvim.defaultNullOpts.mkEnum [
+      "center"
+      "bottom_right"
+    ] "center" "The help window (opened with ?) can be positioned in different locations";
+    keymap = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.either keys lib.types.bool);
+      description = "Keymaps for jujutsu-nvim";
+      default = { };
+
```
MattSturgeon, line 109:
```suggestion
```

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    };
```
MattSturgeon, line 110: The reason we've stopped declaring settings-options in new plugins is because they don't do anything, users can already define anything in `settings` as it is freeform.

They do add documentation, which also adds a maintenance burden for us (and you) to keep in sync with upstream. And they add more restrictive option types, which can sometimes accidentally prevent users defining valid configs.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+        "codediff"
+      ];
+    in
+    {
+      plugins = lib.genAttrs' possible-sources (name: {
+        inherit name;
+        value.enable = lib.mkIf (cfg.settings.diff_preset == name) true;
+      }
+
+      );
```
MattSturgeon, line 91:
```suggestion
      });
```

You may need to run `nix fmt` or `nix develop --run treefmt` on this PR.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    "squash_to_target" = "Squash to specific target";
+    "bookmark_change" = "Set/create bookmark";
+    "bookmark_menu" = "Show bookmark operations";
+    "push_bookmarks" = "Push bookmarks";
+    "push_bookmarks_and_create" = "Push bookmarks with --allow-new";
+    "git_fetch" = "Fetch from remote";
+    "pull_bookmark" = "Pull bookmark from remote";
+    "toggle_change" = "Toggle multi-selection";
+    "clear_selections" = "Clear all selections";
+  };
```
MattSturgeon, line 35: This feels hard to maintain. Is it really worth it? Can't we link to upstream docs or refer to `:h` instead?

plugins/by-name/jujutsu/default.nix
auscyber: does settingsExample prevent this?

plugins/by-name/jujutsu/default.nix
auscyber: illl remove this as a guarantee, but it is a nice fallback for the desc which makes the documentation work correctly, ill make it less stringent on checking

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+
+    action = false;
+    defaults.mode = "";
+    extraOptions = {
+      cmd = mkOption {
+        type = lib.types.maybeRaw (lib.types.enum (builtins.attrNames actions));
+        description = "Command to execute";
+      };
+
+      desc = mkOption {
```
khaneliman, line 48: issue (blocking): read declared keymap field

This submodule disables `action` and declares `cmd`, so `config.action` does not exist. Use `config.cmd` for this lookup and the following description lookup.

khaneliman (commented):
Existing review requests remain unresolved. One additional keymap evaluation defect needs correction.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    let
+      possible-sources = [
+        "diffview"
+        "codediff"
+      ];
+    in
+    {
+      plugins = lib.genAttrs' possible-sources (name: {
+        inherit name;
+        value.enable = lib.mkIf (cfg.settings.diff_preset == name) true;
```
khaneliman, line 88: issue (blocking): `lib.mkIf ... true` defines `plugins.<viewer>.enable` at normal priority, so it conflicts with a user who explicitly disables that plugin.

Evaluating `plugins.jujutsu.settings.diff_preset = "diffview"` together with `plugins.diffview.enable = false` fails with "conflicting definition values". Use `lib.mkDefault true`, or drop the auto-enable and emit an assertion or warning instead.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+    "pull_bookmark" = "Pull bookmark from remote";
+    "toggle_change" = "Toggle multi-selection";
+    "clear_selections" = "Clear all selections";
+  };
+
+  keys = lib.nixvim.keymaps.mkMapOptionSubmodule {
+    key = false;
+
+    action = false;
+    defaults.mode = "";
```
khaneliman, line 41: issue (blocking): `mkMapOptionSubmodule` always declares `mode`, and `defaults.mode = ""` gives it a non-null default, so `mode` leaks into upstream's config table.

Evaluation produces `setup({ keymap = { ["?"] = { cmd = "show_help", mode = "" } } })`, but upstream's `KeymapBinding` only accepts `cmd`, `desc`, and `nowait`. Declare a plain submodule with just the upstream fields, or drop the option and rely on freeform `settings`.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+      );
+    };
+
+  settingsOptions = {
+    diff_preset = lib.nixvim.defaultNullOpts.mkEnum [
+      "difftastic"
+      "diffview"
+      "codediff"
+      "none"
+    ] "none" "diff view preset";
```
khaneliman, line 100: issue: the `pluginDefault` argument documents the upstream default, but jujutsu-nvim's `default_config.diff_preset` is `"difftastic"`, not `"none"`.

This publishes a wrong default in the generated option docs. This shows an example of why it was recommended to delete the option per the earlier settings-options feedback.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+      )
+    ];
+  };
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "jujutsu";
+  moduleName = "jujutsu-nvim";
+  package = "jujutsu-nvim";
+
+  maintainers = [ lib.maintainers.auscyber ];
```
khaneliman, line 76: suggestion: jujutsu-nvim shells out to the `jj` binary, and nixvim already declares `dependencies.jujutsu`.

Add `dependencies = [ "jujutsu" ];` so the runtime dependency is wired up like `plugins/by-name/conflict/default.nix` does with `git`.

tests/test-sources/plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,20 @@
+{
+  empty = {
+    plugins.jujutsu.enable = true;
+  };
+
+  defaults = {
```
khaneliman, line 6: suggestion: repo convention is `empty` plus a `defaults` case mirroring upstream's `default_config` and an `example` case matching `settingsExample` (see the conflict plugin tests).

This `defaults` case sets one custom keymap instead, and nothing covers the `keymap.<key> = false` branch or the `diff_preset` extraConfig path, which is why CI stayed green despite the enable conflict.

khaneliman (changes requested):

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+          };
+        }
+      )
+    ];
+  };
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "jujutsu";
+  moduleName = "jujutsu-nvim";
+  package = "jujutsu-nvim";
```
khaneliman: issue: the required `jj` dependency is not enabled

`jujutsu.nvim` invokes the external `jj` executable for `:JJ` and log actions ([upstream implementation](https://github.com/yannvanhalewyn/jujutsu.nvim/blob/17ab008d71cbcd31f8c2891e11cb758579f597c0/lua/jujutsu-nvim/init.lua)). With only `plugins.jujutsu.enable = true`, Nixvim leaves `dependencies.jujutsu` disabled, so the core command fails unless the user installs `jj` separately. Add `dependencies = [ "jujutsu" ];`.

plugins/by-name/jujutsu/default.nix
khaneliman: issue: the declared diff default is not applied

`defaultNullOpts.mkEnum` leaves `diff_preset` as `null`, so the `"none"` argument only changes option text and the generated setup call still lets upstream use `"difftastic"`. That default's `open_diff` path runs `:Difft`, but this module only enables `diffview` and `codediff`, so a minimal configuration has no viewer and the docs are wrong ([upstream implementation](https://github.com/yannvanhalewyn/jujutsu.nvim/blob/17ab008d71cbcd31f8c2891e11cb758579f597c0/lua/jujutsu-nvim/init.lua)). Either emit `diff_preset = "none"` or provide `difftastic-nvim`, then keep the option text aligned.

plugins/by-name/jujutsu/default.nix
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  keys = lib.nixvim.keymaps.mkMapOptionSubmodule {
+    key = false;
+
+    action = false;
+    defaults.mode = "";
+    extraOptions = {
+      cmd = mkOption {
+        type = lib.types.maybeRaw (lib.types.enum (builtins.attrNames actions));
```
khaneliman: issue: custom keymap commands are rejected

The plugin resolves `actions[binding.cmd] or binding.cmd`, so `cmd` can be a built-in action or a normal command string ([upstream implementation](https://github.com/yannvanhalewyn/jujutsu.nvim/blob/17ab008d71cbcd31f8c2891e11cb758579f597c0/lua/jujutsu-nvim/init.lua)). This option type accepts only `builtins.attrNames actions`, so `settings.keymap.x.cmd = "<Cmd>MyCommand<CR>"` fails Nix evaluation instead of reaching the plugin. Accept arbitrary strings and use the built-in set only for description lookup.

PR #4376, opened 2026-06-05
lib/keymaps: remove keymaps `lua` sub-option

MattSturgeon:
- **modules/test: add "all" expectation predicates**
- **lib/keymaps: remove `lua` submodule option**
- **lib/keymaps: drop `removeDeprecatedMapAttrs`**

### DIFF
--- a/lib/keymap-helpers.nix
+++ b/lib/keymap-helpers.nix
@@ -65,7 +65,11 @@ rec {
   # to ensure `lua` isn't evaluated when (e.g.) generating lua code.
   # Failure to do so will result in "option used but not defined" errors!
   deprecatedMapOptionSubmodule = mkMapOptionSubmodule { lua = true; };
-  removeDeprecatedMapAttrs = v: removeAttrs v [ "lua" ];
+  removeDeprecatedMapAttrs = lib.flip removeAttrs [
+    "lua"
+    "assertions"
+    "warnings"
+  ];
 
   mkModeOption =
     default:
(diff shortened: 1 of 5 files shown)


modules/keymaps.nix
```diff
@@ -49,51 +70,14 @@ in
...
-              ))
-            ]
-            [
-              lib.concatLists
-              (lib.filter (meta: meta.configuration.options.lua.isDefined or false))
-              (map (meta: meta.configuration.options))
-            ];
-      in
-      lib.optional (keymapsUsingLua != [ ]) ''
-        The `lua` keymap option is deprecated and will be removed in 26.05.
```
MattSturgeon, line 83: We're a release late, but that's ok 🙂

lib/keymap-helpers.nix
```diff
@@ -61,11 +61,42 @@ rec {
...
+          {
+            inherit (result) headError valueMeta;
+            value = removeAttrs result.value [
+              "lua"
+              "warnings"
+              "assertions"
+            ];
+          };
+      };
+    };
```
MattSturgeon: https://github.com/NixOS/nixpkgs/pull/528474 or https://github.com/NixOS/nixpkgs/pull/163601 would simplify this if we were doing this in the future. However, our `lua` deprecation will be finished before we are able to rely on nixpkgs-lib having that.

lib/keymap-helpers.nix
```diff
@@ -101,57 +132,45 @@ rec {
...
-            );
-          }
-          // {
-            mode = mkModeOption defaults.mode or "";
-            options = mapConfigOptions;
-          }
-          // extraOptions;
-      }
-    );
+    customKeymapSubmodule {
```
khaneliman: suggestion (non-blocking): scope merge wrapper to removed lua option

This wrapper drops `lua`, `warnings`, and `assertions` for every keymap submodule. Minimal eval confirms same names from `extraOptions` disappear too. Use wrapper only for legacy `lua` maps.

```suggestion
    (if isAttrs lua || lua then customKeymapSubmodule else types.submodule) {
```

tests/test-sources/modules/keymaps.nix
```diff
@@ -132,16 +132,15 @@
...
-      (expect "any" "- `plugins.lsp.keymaps.extra.\"[definition 1-entry 1]\".lua' is defined in `test-module'")
-      (expect "any" "- `plugins.tmux-navigator.keymaps.\"[definition 1-entry 1]\".lua' is defined in `test-module'")
-      (expect "any" "- `plugins.barbar.keymaps.first.lua' is defined in `test-module'")
+    test.assertions = expect: [
+      # Failed assertions:
+      #  - The option definition `lua' in `<unknown-file>' no longer has any effect; please remove it.
+      #  Use a "raw lua" `action` instead;
+      #  e.g. `action.__raw = "<lua code>"` or `action = lib.nixvim.mkRaw "<lua code>"`.
+      (expect "count" 5)
+      (expect "all" "The option definition `lua' in `test-module' no longer has any effect; please remove it.")
```
khaneliman, line 141: suggestion (non-blocking): assert path-specific lua diagnostics

This test now accepts five duplicate pathless messages. Since this is where the removed-option output is locked down, keep expectations that prove each legacy keymap location remains identifiable.

```suggestion
      (expect "any" "keymaps.\"[definition 1-entry 1]\".lua")
      (expect "any" "keymapsOnEvents.InsertEnter.\"[definition 1-entry 1]\".lua")
      (expect "any" "plugins.lsp.keymaps.extra.\"[definition 1-entry 1]\".lua")
      (expect "any" "plugins.tmux-navigator.keymaps.\"[definition 1-entry 1]\".lua")
      (expect "any" "plugins.barbar.keymaps.first.lua")
```

modules/keymaps.nix
```diff
@@ -6,10 +6,31 @@
...
 let
   inherit (lib.nixvim) toLuaObject;
-  inherit (lib.nixvim.keymaps)
-    removeDeprecatedMapAttrs
-    deprecatedMapOptionSubmodule
-    ;
+  inherit (lib.nixvim.keymaps) deprecatedMapOptionSubmodule;
+
+  # All keymap submodules that have historically supported the `lua` sub-option
+  keymapsUsingLua =
```
khaneliman: suggestion (non-blocking): rename collection to match contents

This now includes every legacy-capable keymap configuration, not only maps using `lua`. Rename it and update the two uses below.

```suggestion
  legacyKeymapConfigurations =
```

plugins/by-name/barbar/default.nix
```diff
@@ -82,7 +82,6 @@ lib.nixvim.plugins.mkNeovimPlugin {
           }
         );
         default = null;
-        apply = v: if v == null then null else keymaps.removeDeprecatedMapAttrs v;
         description = "Keymap for function Buffer${funcName}";
```
khaneliman, line 88: suggestion (non-blocking): update stale valueMeta comment

The comment above still says `lua` deprecation warning. This is now removed-option assertion plumbing. Lines 48-52 are outside this diff hunk; suggested replacement:

```nix
  # As of 2025-11-21, `either` supports `valueMeta`, while `nullOr` does not.
  # The `lua` removal assertions in `modules/keymaps.nix` require `valueMeta`,
  # so re-implement `nullOr` using `types.either` as a workaround.
  #
  # TODO: Remove once `nullOr` supports v2 check and merge.
```

lib/keymap-helpers.nix
MattSturgeon: > Minimal eval confirms same names from `extraOptions` disappear too

Interesting. So we have some keymap submodules that have non-internal `assertions`, `lua`, or `warnings` sub-options?

Do you know which?

khaneliman: No current ones found. I checked the existing `mkMapOptionSubmodule` callsites; current extra options are `cwd` and `plugin`, plus LSP uses `extraModules`.

So this is not an existing regression, more API surface/footgun: `mkMapOptionSubmodule { }` now reserves and strips `lua`, `warnings`, and `assertions` for all future keymap submodules. I think we can feel free to ignore, just wanted to call out a possible issue.

tests/test-sources/modules/keymaps.nix
khaneliman: My suggestion was wrong as a standalone test change. These path-specific strings are desired output, not current output.

Please drop the five added `expect "any" ... .lua` lines and keep the original count/shared-message checks. I verified that fix with `nix develop --command tests modules-keymaps`.

MattSturgeon: Interestingly we're seeing 

> ``The option definition `lua' in `test-module' no longer has any effect; please remove it.``

in the assertion, so the `lua` option's prefix is missing. That could be a bug in `mkRemovedOptionModule`?

MattSturgeon: Indeed: https://github.com/NixOS/nixpkgs/blob/2f7793061c0dd564abb22061b930ea8f94d85126/lib/modules.nix#L1727-L1730

That should be using `"${opt}"` or `showOption opt.loc` but it using the raw unprefixed list passed directly to mkRemovedOptionModule. Same in the option's `apply` error.

EDIT: https://github.com/NixOS/nixpkgs/pull/528556

MattSturgeon: As a workaround, I added this to the error message.
```nix
"Full option: `${options.lua}`"
```

If my Nixpkgs PR gets merged it'll fail our test, so effectively we will be notified to remove the workaround.

plugins/by-name/barbar/default.nix
MattSturgeon: Went with:
```nix
# As of 2025-11-21, `either` supports `valueMeta`, while `nullOr` does not.
# See https://github.com/NixOS/nixpkgs/pull/525519
#
# The `lua` removal assertions in `modules/keymaps.nix` require `valueMeta`,
# so re-implement `nullOr` using `types.either` as a workaround.
#
# TODO: Remove when assertions are dropped or `nullOr` supports v2 checkAndMerge.
```

MattSturgeon:
I'm going to drop af756174 `` lib/keymaps: drop `removeDeprecatedMapAttrs` `` as I'm unsure how reliable it'll be after a similar investigation into type-merging in https://github.com/NixOS/nixpkgs/pull/163601#issuecomment-4654533162

khaneliman (approved):

PR #4379, opened 2026-06-05
plugins/conflict.nvim: init

### LINKED ISSUE #4279: [PLUGIN REQUEST] conflict.nvim
### Plugin

conflict.nvim

### Homepage

https://github.com/niekdomi/conflict.nvim

### Nixpkgs URL

_No response_

### Extra Information

It is a maintained alternative to git-conflict.nvim.
Recently git-conflict started throwing errors whenever there are any conflicts in the buffer, presumably because it was last updated in 2024. So an up to date alternative would be nice to have.

jaredmontoya:
Generated with opus 4.6 based on the git-conflict module using opencode.
Works inside of my NixOS configuration so far.

closes #4279

### DIFF
--- a/plugins/by-name/conflict/default.nix
+++ b/plugins/by-name/conflict/default.nix
@@ -1,0 +1,29 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "conflict";
+  package = "conflict-nvim";
+  description = "A simple Neovim plugin to resolve merge conflicts with ease.";
+
+  maintainers = [ lib.maintainers.jaredmontoya ];
+
+  dependencies = [ "git" ];
+
+  settingsExample = {
+    default_mappings = {
+      current = "<leader>co";
+      incoming = "<leader>ct";
+      both = "<leader>ca";
+      base = "<leader>cb";
+      none = "<leader>cn";
+      next = "]e";
+      prev = "[e";
+    };
+    show_actions = false;
+    disable_diagnostics = false;
+    highlights = {
+      current = "DiffChange";
+      incoming = "DiffText";
+      ancestor = "DiffDelete";
+    };
+  };
+}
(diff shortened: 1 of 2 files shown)


plugins/by-name/conflict/default.nix
```diff
@@ -0,0 +1,82 @@
...
+      incoming = lib.nixvim.defaultNullOpts.mkStr "DiffAdd" ''
+        Which highlight group to use for incoming (remote) changes.
+      '';
+
+      ancestor = lib.nixvim.defaultNullOpts.mkStr "DiffChange" ''
+        Which highlight group to use for ancestor (base) changes.
+      '';
+    };
+  };
+
```
MattSturgeon:
```suggestion
```

Settings options are not typically being added to Nixvim anymore. They are now deemed to be a considerable maintenance burden, that provide little benefit.

`settingsExample` and `description` are already more than enough.

jaredmontoya: Done, please also backport to stable.

tests/test-sources/plugins/by-name/conflict/default.nix
```diff
@@ -0,0 +1,55 @@
...
+        show_actions = false;
+        disable_diagnostics = false;
+        highlights = {
+          current = "DiffText";
+          incoming = "DiffAdd";
+          ancestor = "DiffChange";
+        };
+      };
+    };
+  };
```
MattSturgeon, line 54: The `defaults` and `example` tests seem to be identical.

Are these upstream's defaults, or are they an example of what a user may configure?

It's fine to have one or both, but if present `defaults` should be the plugin's defaults explicitly defined and `example` should be a non-default practical example of configuring some options.

Usually, `settingsExample` in the main plugin should be similar or identical to the `example` test.

MattSturgeon (approved):

PR #4393, opened 2026-06-07
plugins/web-devicons: drop auto-enable and back-compat

MattSturgeon:
The `plugins.<name>.iconsPackage` options only briefly existed before being removed, we can safely drop their removal assertions now.

The `plugins.web-devicons.enable` auto-enable has been deprecated since the same time-period, originally marked as "remove after 24.11". We never actually stated "it will be removed after x" in the warning, but I think it is fair to drop it after having had a warning this long.

### DIFF
--- a/plugins/by-name/barbar/default.nix
+++ b/plugins/by-name/barbar/default.nix
@@ -1,4 +1,4 @@
-{ config, lib, ... }:
+{ lib, ... }:
 let
   inherit (lib) mkOption types;
   inherit (lib.nixvim)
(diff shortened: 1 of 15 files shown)


khaneliman (commented):
can we clean up the tests that enable it to get around the warning?

khaneliman (approved):

khaneliman:
Well, barbar apparently has another issue 

       > ERROR: barbar.nvim: `icons.filetype.enabled` is set to `true` but `nvim-web-devicons` was not found.
       > barbar.nvim: icons have been disabled. Set `icons.filetype.enabled` to `false` or install `nvim-web-devicons`as a non-optional plugin to prevent this message.

plugins/by-name/barbar/default.nix
```diff
@@ -89,18 +89,6 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-          config.plugins.mini.enable
-          && config.plugins.mini.modules ? icons
-          && config.plugins.mini.mockDevIcons
-        )
-        || (config.plugins.mini-icons.enable && config.plugins.mini-icons.mockDevIcons)
-      )
-    ) { enable = lib.mkOverride 1490 true; };
-
     keymaps = lib.filter (keymap: keymap != null) (
       # TODO: switch to `attrValues cfg.keymaps` when removing the deprecation warnings above:
```
MattSturgeon, line 93: 👀

MattSturgeon: Seems unrelated.

khaneliman (approved):

PR #4413, opened 2026-06-13
modules/output: reject vim plugins in extraPackages

khaneliman:
Prevent Vim plugin derivations from being silently treated as PATH packages and point users to extraPlugins instead.

Saw this on matrix and thought it fit into our recent footgun assertions path.

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -1,4 +1,9 @@
-{ lib, config, ... }:
+{
+  lib,
+  config,
+  options,
+  ...
+}:
 let
   inherit (lib) types mkOption;
 
@@ -20,6 +25,24 @@ let
       };
     };
   };
+
+  isVimPluginPackage = package: package.vimPlugin or false;
+
+  mkVimPluginPackageAssertion =
+    opt:
+    let
+      vimPlugins = builtins.filter isVimPluginPackage opt.value;
+    in
+    {
+      assertion = vimPlugins == [ ];
+      message = ''
+        `${opt}` is for executable packages added to Neovim's PATH, but it contains Vim plugin package(s):
+        ${lib.concatMapStringsSep "\n" (package: "  - ${lib.getName package}") vimPlugins}
+
+        Use `${options.extraPlugins}` for Vim plugin packages:
+          ${options.extraPlugins} = [ pkgs.vimPlugins.<plugin> ];
+      '';
+    };
 in
 {
   options = {
@@ -132,6 +155,11 @@ in
   };
 
   config = {
+    assertions = lib.nixvim.mkAssertions "output" [
+      (mkVimPluginPackageAssertion options.extraPackages)
+      (mkVimPluginPackageAssertion options.extraPackagesAfter)
+    ];
+
     content =
       if config.type == "lua" then
         # Lua
--- a/tests/test-sources/modules/output.nix
+++ b/tests/test-sources/modules/output.nix
@@ -295,6 +295,31 @@
       '';
     };
 
+  extraPackages-rejects-vim-plugins =
+    { pkgs, ... }:
+    let
+      plugin = pkgs.vimUtils.buildVimPlugin {
+        pname = "nixvim-extra-packages-vim-plugin-test";
+        version = "1";
+        src = pkgs.runCommandLocal "nixvim-extra-packages-vim-plugin-test-src" { } "mkdir -p $out";
+      };
+    in
+    {
+      extraPackages = [ plugin ];
+      extraPackagesAfter = [ plugin ];
+
+      test = {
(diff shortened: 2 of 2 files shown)


modules/output.nix
```diff
@@ -20,6 +20,24 @@ let
       };
     };
   };
+
+  isVimPluginPackage = package: (package.vimPlugin or false) || (package.passthru.vimPlugin or false);
```
MattSturgeon: If `passthru.x` is passed to mkDerivation, then `x` will always be present in the package. I don't think the second case will ever be needed:

```suggestion
  isVimPluginPackage = pkg: pkg.vimPlugin or false;
```

modules/output.nix
```diff
@@ -20,6 +20,24 @@ let
...
+
+  mkVimPluginPackageAssertion =
+    option:
```
MattSturgeon: I believe this is the conventional name:

```suggestion
    opt:
```

Either is fine, of course.

MattSturgeon: Actually, if this is not an option object, but an option attrName, we should clarify that:

```suggestion
    name:
```

```suggestion
    optionName:
```

Etc

(This tripped me up, original comment left for demonstration)

modules/output.nix
```diff
@@ -20,6 +20,24 @@ let
...
+    let
```
MattSturgeon, line 33: We can take a reference to the actual option, too. That way we can interpolate it into errors to print the full (prefixed) path.


```suggestion
    let
      opt = options.${optionName};
      value = config.${optionName};
```

modules/output.nix
```diff
@@ -20,6 +20,24 @@ let
...
+      packages = builtins.filter isVimPluginPackage config.${option};
```
MattSturgeon: Would `pluginPackages` or `vimPlugins` be better names?

`packages` could equally mean the unfiltered list, or the other side of a partitioned list.

modules/output.nix
```diff
@@ -20,6 +20,24 @@ let
...
+    in
+    {
+      assertion = packages == [ ];
+      message = ''
+        `${option}` is for executable packages added to Neovim's PATH, but it contains Vim plugin package(s):
+        ${lib.concatMapStringsSep "\n" (package: "  - ${lib.getName package}") packages}
+
+        Use `extraPlugins` for Vim plugin packages:
+          extraPlugins = [ pkgs.vimPlugins.<plugin> ];
```
MattSturgeon: At least one of these `extraPlugins` references should use the full interpolated path:

```suggestion
        Use `extraPlugins` for Vim plugin packages:
          ${options.extraPlugins} = [ pkgs.vimPlugins.<plugin> ];
```

MattSturgeon (commented):
SGTM semantically. Some nits on naming and handling of option paths in errors.

modules/output.nix
MattSturgeon: On reflection, I think this API works better if it receives an "actual option instance" instead of an attrName.

That makes usage more flexible; you aren't restricted to top-level options anymore, and this function's impl doesn't need access to full `config` or `options` sets.

From an option instance, you can access everything you need:
- `"${opt}"` aka `lib.showOption opt.loc`
- `lib.last opt.loc`
- `opt.value`
Etc

modules/output.nix
khaneliman: I think you're right... I'm used to checking `passthru` when dealing with internal nixpkgs logic where we do have to check for it explicitly while building a derivation.

modules/output.nix
khaneliman: I agree, that sounds cleaner.

MattSturgeon (approved):
Awesome!

modules/output.nix
MattSturgeon: Indeed, the `finalAtttrs` fixpoint seen by `overrideAttrs` and mkDerivation-inputs is _before_ mkDerivation applies passthru to the result.

tests/test-sources/modules/output.nix
```diff
@@ -295,6 +295,31 @@
       '';
     };
 
+  extraPackages-rejects-vim-plugins =
+    { pkgs, ... }:
+    let
+      plugin = pkgs.vimUtils.buildVimPlugin {
+        pname = "nixvim-extra-packages-vim-plugin-test";
+        version = "1";
+        src = pkgs.runCommandLocal "nixvim-extra-packages-vim-plugin-test-src" { } "mkdir -p $out";
```
MattSturgeon, line 304: FYI:

```suggestion
        src = pkgs.emptyDirectory;
```

https://github.com/nix-community/nixvim/pull/4419

PR #4415, opened 2026-06-13
Re-enable _module.check

### LINKED ISSUE #4411: [BUG] Don't accept arbitrary keys under `nixvim.plugins`
I already made a few times a mistake of setting `nixvim.plugins.X` instead of `nixvim.X`. Unfortunately nixvim just silently ignores that mistake, which led me to multiple hours wasted on debugging. Could something be done about it?

I've tried to find in the code where `options.plugins` is defined, but couldn't find it. The documentation doesn't say anything about `options.plugins` itself, so I don't even know for what reason it is possible to put arbitrary keys there, or is it just a bug?

MrQubo:
Fixes #4411

### DIFF
--- a/wrappers/_shared.nix
+++ b/wrappers/_shared.nix
@@ -52,7 +52,7 @@ let
   baseConfiguration = extendModules {
     modules = [
       nixpkgsModule
-      { disabledModules = [ "<internal:nixvim-nocheck-base-eval>" ]; }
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Ah, sorry I didn't seen this until after I pushed up #4417.

I'll go with my PR as it includes tests, but the fix is identical. As we both came up with the same solution, I'll mark you as co-author.

PR #4417, opened 2026-06-13
wrappers: fix `_module.check` disablement

### LINKED ISSUE #4411: [BUG] Don't accept arbitrary keys under `nixvim.plugins`

MattSturgeon:
When passing a string to `disabledModules`, it is appended to `modulesPath`. To avoid this, explicitly pass a `{key}` module.

Adds an assertion module to the platform-wrapper integration tests, which prints the following error:

```
Failed assertions:
    - Expected programs.nixvim._module.check to be true. Definitions:
    - In `/nix/store/XXXX-source/flake/wrappers.nix': false
```

Fixes #4411

### DIFF
--- a/flake/wrappers.nix
+++ b/flake/wrappers.nix
@@ -19,6 +19,7 @@ let
   configuration = self.lib.evalNixvim {
     modules = [
       {
+        _file = __curPos.file;
         key = "<internal:nixvim-nocheck-base-eval>";
         config._module.check = false;
       }
--- a/tests/platforms/darwin.nix
+++ b/tests/platforms/darwin.nix
@@ -9,6 +9,9 @@ self.inputs.nix-darwin.lib.darwinSystem {
 
       programs.nixvim = {
         enable = true;
+        imports = [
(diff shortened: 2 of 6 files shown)


GaetanLepage (approved):

MattSturgeon:
> Backport failed for `nixos-25.11`, because it was unable to cherry-pick the commit(s).

`nixos-25.11` pre-dates the "disable checks in the flake eval" then "disable the disable-checks module in the wrapper submodule" pattern, which was added in 26.05 and not backported to 25.11.

There's nothing to backport.

PR #4421, opened 2026-06-14
modules/output: make vimPlugin assertions lazier

MattSturgeon:
Yes, a second minor followup. Sorry...

1. We can make the "happy path" of the assertion a bit cheaper to evaluate.
   - This is more important than it sounds, as all assertions get evaluated by every Nixvim eval.
2. Avoid printing the same file multiple times on the same line, if one file defines multiple plugins with the same name

### DIFF
--- a/modules/output.nix
+++ b/modules/output.nix
@@ -28,22 +28,21 @@ let
 
   isVimPluginPackage = package: package.vimPlugin or false;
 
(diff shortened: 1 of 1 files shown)


khaneliman (approved):

khaneliman:
sorry, missed this

PR #4427, opened 2026-06-16
dev: update search

MarcelCoding:
Its now also possible to add packages to the search. Maybe I will do another PR.


MattSturgeon (approved):
Thanks for bumping! Seems to have built ok. A few warnings upstream may want to be aware of: https://buildbot.nix-community.org/#/builders/314/builds/381/steps/1/logs/stdio

PR #4430, opened 2026-06-16
various: add newly packaged to available lists

khaneliman:
Just visiting several of the various modules and their unpackaged vs packaged lists and trying to map some newly available packages.

### DIFF
--- a/modules/lsp/servers/packages.nix
+++ b/modules/lsp/servers/packages.nix
@@ -1,11 +1,12 @@
 {
   unpackaged = [
+    # keep-sorted start block=yes newline_separated=no
     "ada_ls"
     "agda_ls"
     "alloy_ls"
     "anakin_language_server"
-    "antlersls"
     "ansiblels"
+    "antlersls"
     "apex_ls"
     "autohotkey_lsp"
     "awk_ls"
@@ -20,7 +21,6 @@
     "bsl_ls"
     "buddy_ls"
     "bzl"
-    "c3_lsp"
     "cairo_ls"
     "cds_lsp"
     "cir_lsp_server"
@@ -45,7 +45,6 @@
     "djls"
     "djlsp"
     "ds_pinyin_lsp"
-    "dts_lsp"
     "ecsact"
     "ember"
     "esbonio"
@@ -63,8 +62,8 @@
     "gitlab_duo"
     "glasgow"
     "glint"
-    "gnls"
     "gn_language_server"
+    "gnls"
     "gradle_ls"
     "grammarly"
     "graphql" # nodePackages.graphql-language-service-cli was removed in https://github.com/NixOS/nixpkgs/pull/382557
@@ -80,9 +79,7 @@
     "hoon_ls"
     "hydra_lsp"
     "hylo_ls"
-    "intelephense"
     "janet_lsp"
-    "jinja_lsp"
     "jls"
     "julials"
     "kakehashi"
@@ -97,7 +94,6 @@
     "m68k"
     "marko-js"
     "mdx_analyzer"
-    "millet"
     "mm0_ls"
     "mojo"
     "motoko_lsp"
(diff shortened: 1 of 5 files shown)


modules/lsp/servers/packages.nix
```diff
@@ -258,6 +250,14 @@
...
     emmet_ls = "emmet-ls";
     emmylua_ls = "emmylua-ls";
+    c3_lsp = "c3-lsp";
+    dts_lsp = "dts-lsp";
+    intelephense = "intelephense";
+    jinja_lsp = "jinja-lsp";
+    millet = "millet";
+    textlsp = "textlsp";
+    veridian = "veridian";
+    wasm_language_tools = "wasm-language-tools";
```
MattSturgeon: Do you mind `:sort`-ing these with the rest of the file?

plugins/by-name/conform-nvim/formatter-packages.nix
```diff
@@ -71,6 +67,10 @@ in
     ziggy_schema = states.unpackaged;
 
     inherit (python313Packages) autopep8;
+    inherit findent;
+    inherit hledger-fmt;
+    json_repair = json-repair;
+    inherit markdown-toc;
```
MattSturgeon: The below block is also mostly sorted. Though maybe we can have a separate `inherit` block (rather than three inherits)?

plugins/by-name/efmls-configs/packages.nix
```diff
@@ -131,6 +122,15 @@ in
...
     // {
+      alex = "alex";
+      blade_formatter = "blade-formatter";
+      cljstyle = "cljstyle";
+      cspell = "cspell";
+      fixjson = "fixjson";
+      gersemi = "gersemi";
+      kdlfmt = "kdlfmt";
+      swiftformat = "swiftformat";
+      swiftlint = "swiftlint";
```
MattSturgeon: Again, can be sorted with the rest of the list.

plugins/by-name/none-ls/packages.nix
```diff
@@ -121,6 +121,17 @@ in
...
+      cljstyle = "cljstyle";
+      findent = "findent";
+      packer = "packer";
+      pretty_php = "pretty-php";
+      regal = "regal";
+      sql_formatter = "sql-formatter";
+      swiftformat = "swiftformat";
+      swiftlint = "swiftlint";
+      textidote = "textidote";
+      textlint = "textlint";
```
MattSturgeon: Same

MattSturgeon (approved):
We probably need to set up some "keep-sorted" formatting directives, as a lot of these lists are sorted, but not enforced by treefmt.

Ideally, you'd sort in each of your commits. But since it's not enforced and that's a pain to rebase, I won't block on it.

modules/lsp/servers/packages.nix
```diff
@@ -1,4 +1,5 @@
 {
+  # keep-sorted start block=yes newline_separated=no
   unpackaged = [
```
MattSturgeon: Should this be:
```suggestion
  unpackaged = [
    # keep-sorted start block=yes newline_separated=no
```
?

khaneliman: oops, good catch

MattSturgeon: I guess if treefmt-ci didn't complain, it was probably valid? 🤷🏻 
but best to keep it aligned with the closing tag

khaneliman: nah, invalid keep-sorted fail silently i think

khaneliman: https://github.com/nix-community/nixvim/compare/c5f36d546d850277d5d49bd93a21c731afb49e4c..a6c84ee0dcac24f076041d8125045991f4eb822e shows it sorted stuff after i fixed the location

PR #4434, opened 2026-06-17
plugins.lsp: restore `autostart` compatibility

### LINKED ISSUE #4432: [BUG] lsp autostart=false does not disable autostart
### Plugin

Seemingly all lsps

### Nixpkgs Release

unstable

### Home Manager Release

None

### I have read the FAQ

### Description

Currently every server regardless of autostart being true of false has a `vim.lsp.enable("$SERVER_NAME")` call.
This does the following.
> Auto-activates LSP in each buffer based on the lsp-config filetypes, root_markers, and root_dir.

Which we don't want when autostart is false.

```nix
programs.nixvim = {
  plugins.lsp = {
    enable = true;
    servers.lua_ls = {
      enable = true;
      autostart = false;
    };
  };
}

Open a lua file and lua-ls starts.
```

MattSturgeon:
`autostart` was a legacy nvim-lspconfig option controlling whether a server was automatically started for matching buffers.

Neovim's `vim.lsp.config` API does not have an `autostart` field, but the `vim.lsp.enable()` function is effectively equivalent.

Alias legacy `plugins.lsp.servers.<name>.autostart` definitions to `lsp.servers.<name>.activate`, which controls `vim.lsp.enable()`.

Fixes #4432

### DIFF
--- a/plugins/lsp/language-servers/_mk-lsp.nix
+++ b/plugins/lsp/language-servers/_mk-lsp.nix
@@ -31,6 +31,16 @@ let
 
   enabled = config.plugins.lsp.enable && cfg.enable;
 
+  # Note: getSubOptions returns docs-eval options, not fully-realized options with user-definitions.
+  getSubOptions = opt: opt.type.getSubOptions opt.loc;
+  documentedServers = getSubOptions options.lsp.servers;
+  hasBespokeDocOptions = documentedServers ? ${serverName};
+  newOptionDocs =
+    if hasBespokeDocOptions then
+      getSubOptions documentedServers.${serverName}
+    else
+      documentedServers._freeformOptions;
+
   inherit (lib) mkOption types;
 in
 {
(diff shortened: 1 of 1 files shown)


plugins/lsp/language-servers/_mk-lsp.nix
```diff
@@ -219,14 +228,16 @@ in
...
+        );
         package = lib.mkIf (opts.package.highestPrio < 1500) (
           lib.modules.mkAliasAndWrapDefsWithPriority lib.id opts.package
         );
         packageFallback = lib.modules.mkAliasAndWrapDefsWithPriority lib.id opts.packageFallback;
         __configWrapper =
           luaCfg: "__wrapConfig(${lib.foldr lib.id luaCfg config.plugins.lsp.setupWrappers})";
         config = {
-          autostart = lib.mkIf (cfg.autostart != null) cfg.autostart;
           cmd = lib.mkIf (cfg.cmd != null) cfg.cmd;
```
GaetanLepage: What's the effect of moving `activate` from down there to up there?

MattSturgeon: Having it in `config` passes it to `vim.lsp.config()`, via our new module's `config` option.

That essentially just defines useless no-op lua config, as `vim.lsp.Config` does not recognize an "autostart" field.

Instead, we want to define the new module's `activate` option, which in turn controls whether we include `vim.lsp.enable()` in the generated lua.

GaetanLepage: Oh right, makes sense.

GaetanLepage (approved):

PR #4436, opened 2026-06-17
plugins/sidekick: Fix broken `nes` assertion

dwf:
This assertion is supposed to ensure that either `copilot-lua` or the copilot LSP is enabled if the Next Edit Suggestion option is enabled, but the `or true` renders this equivalent to asserting one of these plugins is installed, regardless of the value of `nes.enabled`.

### DIFF
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -27,7 +27,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   extraConfig = cfg: {
     assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
       assertion =
-        (cfg.settings.opts.nes.enabled or true)
+        cfg.settings.opts.nes.enabled
         -> (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable);
       message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled when NES is enabled";
     };


plugins/by-name/sidekick/default.nix
```diff
@@ -27,7 +27,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   extraConfig = cfg: {
     assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
       assertion =
-        (cfg.settings.opts.nes.enabled or true)
+        cfg.settings.opts.nes.enabled
```
khaneliman, line 30: IIRC this was by design? Since the default upstream was that it is enabled and we let someone know that if they haven't configured it, it will throw an error. If the user explicitly disables it, then we don't assert.

MattSturgeon:
> the `or true` renders this equivalent to asserting one of these plugins is installed, regardless of the value of `nes.enabled`.

Sorry for asking the obvious, but are you confusing the [`or`](https://nix.dev/manual/nix/latest/language/operators#attribute-selection) (attribute selection fallback) operator with the [`||`](https://nix.dev/manual/nix/latest/language/operators#logical-disjunction) (logical OR) operator?

dwf:
> Sorry for asking the obvious, but are you confusing the [or](https://nix.dev/manual/nix/latest/language/operators#attribute-selection) (attribute selection fallback) operator with the [||](https://nix.dev/manual/nix/latest/language/operators#logical-disjunction) (logical OR) operator?

I think I am, and I just realized that I was building from 25.11. Seems like it's WAI when building from 26.05. Sorry for the noise!

dwf:
Actually, this still seems broken? The calling convention seems to be for the `settings` member to be passed as `opts` to `setup()`, but this expects there to be an `opts` attrset inside `settings`. 

To confirm, when I define a tool as `settings.opts.cli.tools.foo.cmd = [ "foo" ]`, I don't see the `foo` tool defined in `:lua =require("sidekick.config").tools()`, but if I use `settings.cli.tools.foo.cmd`, I do. So I think this assertion is against an element the plugin will ignore.

khaneliman:
> Actually, this still seems broken? The calling convention seems to be for the `settings` member to be passed as `opts` to `setup()`, but this expects there to be an `opts` attrset inside `settings`.
> 
> To confirm, when I define a tool as `settings.opts.cli.tools.foo.cmd = [ "foo" ]`, I don't see the `foo` tool defined in `:lua =require("sidekick.config").tools()`, but if I use `settings.cli.tools.foo.cmd`, I do. So I think this assertion is against an element the plugin will ignore.

I think you're correct with that, would welcome a PR to fix it.

PR #4439, opened 2026-06-19
plugins/sidekick: fix incorrect assertion

dwf:
The calling convention is that the `settings` member will be passed as `opts` to `setup()`, but the assertion expects there to be an `opts` attrset inside `settings`.

### DIFF
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -15,7 +15,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
     Sidekick's NES feature requires the Copilot language server. Enable either
     `plugins.copilot-lua.enable` or `lsp.servers.copilot.enable`, or disable
-    NES in `plugins.sidekick.settings.opts.nes.enabled`.
+    NES in `plugins.sidekick.settings.nes.enabled`.
 
     Sidekick supports several external CLI tools, but nixvim does not enable
     them automatically. Enable the tools you use explicitly, for example with
--- a/plugins/by-name/sidekick/default.nix
+++ b/plugins/by-name/sidekick/default.nix
@@ -27,7 +27,8 @@ lib.nixvim.plugins.mkNeovimPlugin {
   extraConfig = cfg: {
     assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
       assertion =
-        (cfg.settings.nes.enabled or true)
+        # Compare with false to handle settings.nes.enabled being a Lua function.
+        (cfg.settings.nes.enabled or true != false)
         -> (config.plugins.copilot-lua.enable || config.lsp.servers.copilot.enable);
       message = "sidekick requires either copilot-lua (${options.plugins.copilot-lua.enable}) or copilot LSP (${options.lsp.servers}.copilot.enable) to be enabled when NES is enabled";
     };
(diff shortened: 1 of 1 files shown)


khaneliman (dismissed):
ty

khaneliman:
Just need to fix the test

dwf:
Now that I look at it, I'm actually not sure how to fix the test. The default isn't a boolean, so it looks like the assertion needs to somehow accommodate `nes.enabled` being a Lua function?

plugins/by-name/sidekick/default.nix
```diff
@@ -27,7 +27,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
   extraConfig = cfg: {
     assertions = lib.nixvim.mkAssertions "plugins.sidekick" {
       assertion =
-        (cfg.settings.opts.nes.enabled or true)
+        (cfg.settings.nes.enabled or true)
```
khaneliman: The `defaults` test sets `nes.enabled` to a raw Lua function (`{ __raw = "function(buf)..."; }`). With the old `settings.opts.*` path the attr was always missing, so `or true` fired and the LHS was always a boolean. Now that the path resolves, `or` no longer defaults (the attr is present, just non-bool), and `->` can't coerce the attrset → `error: expected a Boolean but found a set`.

Comparing against `false` keeps the LHS boolean for any value:

```suggestion
        (cfg.settings.nes.enabled or true != false)
```

NES counts as enabled unless explicitly `false`; a function value (per-buffer toggle) conservatively still requires copilot. Verified both test cases evaluate cleanly with this.

(Nit: the docstring above still says `settings.opts.nes.enabled` — same stale `.opts` path, worth dropping to `settings.nes.enabled`: https://github.com/nix-community/nixvim/pull/4439#discussion_r3444659910)

MattSturgeon:
For reference, upstream docs show:
```lua
  nes = {
    ---@type boolean|fun(buf:integer):boolean?
    enabled = function(buf)
      return vim.g.sidekick_nes ~= false and vim.b.sidekick_nes ~= false
    end,
```

https://github.com/folke/sidekick.nvim#%EF%B8%8F-configuration

plugins/by-name/sidekick/default.nix
MattSturgeon: Is it enough for the assertion to just check the value with `isBool` and assume for any explicitly non-bool values the user knows what they're doing?

> Comparing against `false`

EDIT: that works too

khaneliman: Yeah I think I saw we used `isBool` in some other assertions in the repo

plugins/by-name/sidekick/default.nix
khaneliman:
```suggestion
    NES in `plugins.sidekick.settings.nes.enabled`.
```

dwf:
Relatedly, for some reason even with NES this disabled I am seeing the derivation try to install copilot (I know because it's tripping the unfree check). I see nothing in the module as to why this would be the case. Any idea?

khaneliman:
> Relatedly, for some reason even with NES this disabled I am seeing the derivation try to install copilot (I know because it's tripping the unfree check). I see nothing in the module as to why this would be the case. Any idea?

Looks like we have `copilot-language-server` as `runtimeDeps` upstream in nixpkgs.

dwf:
I pushed the`isBool` solution but the `!= false` is simpler. I added an explanatory comment as it does look pretty weird.

dwf:
> Looks like we have `copilot-language-server` as `runtimeDeps` upstream in nixpkgs.

Hmm, not ideal. I'll look at fixing that upstream as well.

khaneliman (approved):

PR #4444, opened 2026-06-20
modules/nixvim-info: flatten and simplify

MattSturgeon:
Instead of recursively merging a nested tree, construct a flat map of `"option-path" → {info}`.

This should be more efficient to eval, and simpler to introspect.

Needed this while working on some other docs improvements; replacing the custom docs-eval logic with something closer to Nixpkgs' option-docs tooling.


khaneliman (approved):
LGTM brief measurements showed less gc and thunks.

PR #4451, opened 2026-06-22
docs/modules: model page content as sections

MattSturgeon:
Represent page content as a list of `attrTag`-typed sections, rather than a fixed set of page attributes.

This allows pages to be composed from multiple content sources (in any order) while simplifying the rendering logic and prepares the new docs system for additional section types such as option documentation.

Next steps will be renaming `docs/lib` to something like `docs/pages`, adding "options"-type support to `renderSection`, and rendering option-docs pages via this system too.

Eventually when we can remove the existing option-docs system, we'll be replacing thousands of optionJSON + optionMD derivations with _one_ docs-pages derivation. That'll be a massive performance boost when building the docs.

### DIFF
--- a/docs/lib/default.nix
+++ b/docs/lib/default.nix
@@ -1,5 +1,3 @@
-# Generates the documentation for library functions using nixdoc.
-# See https://github.com/nix-community/nixdoc
 {
   lib,
   runCommand,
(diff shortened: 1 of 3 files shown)


docs/lib/default.nix
```diff
@@ -1,5 +1,3 @@
-# Generates the documentation for library functions using nixdoc.
-# See https://github.com/nix-community/nixdoc
```
MattSturgeon, line 2: I moved this comment down to the nixdoc-specific rendering, as this file could become a more general docs-page renderer in a future PR.

khaneliman (approved):
No issues found.

PR #4461, opened 2026-06-27
wrappers: inherit platform config tuples

### LINKED ISSUE #4426: [BUG] Home Manager eval fails with programs.nixvim.enable: lib.systems.elaborate: linux-kernel has been removed
### Plugin

_No response_

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

### Description

After updating nixvim to `1d91fc6f88e1bda1f36fc4cf261e77458a125926`, Home Manager evaluation fails before building the activation package.

The failure occurs with only the nixvim Home Manager module imported and `programs.nixvim.enable = true`; no plugins are required.

The trace ends in:

```text
error: lib.systems.elaborate: linux-kernel has been removed; see the 26.11 release notes
```

In my real config, reverting only the flake lock update for nixvim/nixvim’s nixpkgs makes evaluation succeed again.

The failing updated inputs are:

- nixvim: 1d91fc6f88e1bda1f36fc4cf261e77458a125926
- nixvim/nixpkgs: 9eac87a12312b8f60dd52e1c6e1a265f6fc7f5fc
- root nixpkgs: 9ae611a455b90cf061d8f332b977e387bda8e1ca
- home-manager: 8aec76cc1e045f37b55d82ca3cee4910ae04d3db

A workaround is to set `programs.nixvim.nixpkgs.pkgs = pkgs;`, or make nixvim’s nixpkgs input follow the root nixpkgs input.

```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/9ae611a455b90cf061d8f332b977e387bda8e1ca";

    home-manager.url = "github:nix-community/home-manager/8aec76cc1e045f37b55d82ca3cee4910ae04d3db";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";

    nixvim.url = "github:nix-community/nixvim/1d91fc6f88e1bda1f36fc4cf261e77458a125926";
  };

  outputs = { nixpkgs, home-manager, nixvim, ... }: {
    homeConfigurations.repro = home-manager.lib.homeManagerConfiguration {
      pkgs = import nixpkgs { system = "x86_64-linux"; };

      modules = [
        nixvim.homeModules.nixvim
        {
          home.username = "repro";
          home.homeDirectory = "/home/repro";
          home.stateVersion = "25.05";

          programs.nixvim.enable = true;
        }
      ];
    };
  };
}
```

### LINKED ISSUE #4460: [BUG] Default LSP server package resolution can infinite recurse on aarch64-darwin with NixVim's nixpkgs
### Plugin

lsp

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

### Description

Edited: this is **not HLS-specific**. HLS was just the first LSP server in my config that exposed the issue.

I’m not certain this is ultimately a NixVim bug; it may be exposing a nixpkgs/Darwin module-system or stdenv recursion. I’m filing here because the smallest reproduction I have is through the NixVim Home Manager module, specifically when NixVim resolves default packages for enabled LSP servers, and because the workaround is NixVim-specific (`programs.nixvim.nixpkgs.useGlobalPackages = true`).

This may also be related to my previous report, #4426, which similarly involved NixVim’s own nixpkgs input interacting badly with Home Manager NixVim module evaluation. That issue looked related to NixOS/nixpkgs#530133. I’m not sure whether this is the same underlying problem, but the shape is similar enough that I wanted to mention it.

Originally, the failure showed up with:

- `programs.nixvim.lsp.servers.hls.packages.prefix`
- `programs.nixvim.lsp.servers.hls.package`

However, after removing HLS, rust-analyzer, and the `nixpkgs.useGlobalPackages = true` workaround, evaluation still fails, now while resolving `marksman`:

- `programs.nixvim.extraPackages`
- `programs.nixvim.lsp.servers.marksman.packages.prefix`
- `programs.nixvim.lsp.servers.marksman.package`

So the issue appears to be default package resolution for enabled LSP servers, not the `hls` server specifically.

The trace indicates that `pkgs` is not externally provided for the LSP server submodule, so the module system queries `_module.args.pkgs`, which requires `config`. NixVim then imports its own
nixpkgs and eventually hits an infinite recursion in Darwin stdenv/runtimeShell.

This reproduces with the default NixVim nixpkgs behavior. Setting:

```nix
programs.nixvim.nixpkgs.useGlobalPackages = true;
```

works around the issue.

Observed with:

- nixvim rev: dbf9550dba8448b03e11d58e5695d6c44a464554
- nixvim nixpkgs input: 3e41b24abd260e8f71dbe2f5737d24122f972158
- root nixpkgs: e73de5be04e0eff4190a1432b946d469c794e7b4
- system: aarch64-darwin

Relevant error shape:

```text
  while evaluating the option `programs.nixvim.extraPackages':

  while evaluating the option `programs.nixvim.lsp.servers.<server>.packages.prefix':

  while evaluating the option `programs.nixvim.lsp.servers.<server>.package':

  while evaluating the module argument `pkgs' in ":anon-2:anon-1:anon-1":

  noting that argument `pkgs` is not externally provided, so querying `_module.args` instead, requiring `config`

  while evaluating definitions from `.../modules/lsp/servers':

  while evaluating the module argument `pkgs' in ".../modules/lsp/servers":

  at .../pkgs/top-level/all-packages.nix:3238:3:

    runtimeShell = "${runtimeShellPackage}${runtimeShellPackage.shellPath}";
```

```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";

    home-manager.url = "github:nix-community/home-manager";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";

    nixvim.url = "github:nix-community/nixvim/dbf9550dba8448b03e11d58e5695d6c44a464554";
  };

  outputs =
    {
      nixpkgs,
      home-manager,
      nixvim,
      ...
    }:
    let
      system = "aarch64-darwin";
      pkgs = nixpkgs.legacyPackages.${system};
    in
    {
      homeConfigurations.repro = home-manager.lib.homeManagerConfiguration {
        inherit pkgs;

        modules = [
          nixvim.homeModules.nixvim
          {
            home.username = "repro";
            home.homeDirectory = "/Users/repro";
            home.stateVersion = "25.05";

            programs.nixvim = {
              enable = true;

              plugins.lsp = {
                enable = true;
                servers.hls = {
                  enable = true;
                  installGhc = false;
                };
              };

              # Uncommenting this works around the infinite recursion:
              # nixpkgs.useGlobalPackages = true;
            };
          }
        ];
      };
    };
}
```

### LINKED ISSUE #4463: [BUG] NixOS unstable rebuild ERROR due to "infinite recursion encountered"
### Plugin

_No response_

### Nixpkgs Release

unstable

### Home Manager Release

unstable

### I have read the FAQ

### Description

When I run `sudo nixos-rebuild boot --upgrade` and it calls ERROR :
```
error:
       … while calling the 'head' builtin
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1719:13:
         1718|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1719|             head values
             |             ^
         1720|           else

       … while evaluating the attribute 'value'
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1147:7:
         1146|     // {
         1147|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1148|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `programs.nixvim.extraPackages':

       … while evaluating definitions from `/nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix':

       … while evaluating the option `programs.nixvim.dependencies.bat.package':

       … while evaluating the module argument `pkgs' in "/nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix":

       … noting that argument `pkgs` is not externally provided, so querying `_module.args` instead, requiring `config`

Command 'nix-build '<nixpkgs/nixos>' --attr config.system.build.toplevel --no-out-link' returned non-zero exit status 1.
```

And it has appeared before , but it was solved after an update . Now it appeared again . 
When I run `sudo nixos-rebuild boot --upgrade --show-trace` and it calls ERROR :
```
error:
       … from call site
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1714:11:
         1713|         zipAttrsWith (
         1714|           name: values:
             |           ^
         1715|           let

       … while calling anonymous lambda
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1714:17:
         1713|         zipAttrsWith (
         1714|           name: values:
             |                 ^
         1715|           let

       … while calling the 'head' builtin
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1719:13:
         1718|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1719|             head values
             |             ^
         1720|           else

       … while calling anonymous lambda
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1192:17:
         1191|         mapAttrs (
         1192|           name: value:
             |                 ^
         1193|           if isAttrs value && cond value then recurse (path ++ [ name ]) value else f (path ++ [ name ]) value

       … from call site
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/attrsets.nix:1193:85:
         1192|           name: value:
         1193|           if isAttrs value && cond value then recurse (path ++ [ name ]) value else f (path ++ [ name ]) value
             |                                                                                     ^
         1194|         );

       … while calling anonymous lambda
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:283:71:
          282|           # For definitions that have an associated option
          283|           declaredConfig = mapAttrsRecursiveCond (v: !isOption v) (_: v: v.value) options;
             |                                                                       ^
          284|

       … while evaluating the attribute 'value'
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1147:7:
         1146|     // {
         1147|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1148|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating the attribute 'mergedValue'
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1244:5:
         1243|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1244|     mergedValue =
             |     ^
         1245|       if isDefined then

         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1245:7:
         1244|     mergedValue =
         1245|       if isDefined then
             |       ^
         1246|         if type.merge ? v2 then

         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1221:7:
         1220|       # above are lazy and thus never forced on this branch.
         1221|       if
             |       ^
         1222|         length defs == 1

       … in the right operand of the AND (&&) operator
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1223:9:
         1222|         length defs == 1
         1223|         && (
             |         ^
         1224|           let

       … while evaluating definitions from `/nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/nixos/modules/system/activation/top-level.nix':

       … in the argument of the not operator
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1228:31:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |                               ^
         1229|           )

       … in the left operand of the AND (&&) operator
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1228:31:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |                               ^
         1229|           )

       … while calling the 'isAttrs' builtin
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:1228:15:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |               ^
         1229|           )

       … while evaluating the attribute 'value'
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/modules.nix:849:21:
          848|             inherit (module) file;
          849|             inherit value;
             |                     ^
          850|           }) module.config

       … from call site
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/nixos/modules/system/activation/top-level.nix:78:26:
           77|   # Handle assertions and warnings
           78|   baseSystemAssertWarn = lib.asserts.checkAssertWarn config.assertions config.warnings baseSystem;
             |                          ^
           79|

       … while calling 'checkAssertWarn'
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:193:27:
          192|   checkAssertWarn =
          193|     assertions: warnings: val:
             |                           ^
          194|     let

         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:197:5:
          196|     in
          197|     if failedAssertions != [ ] then
             |     ^
          198|       throw "\nFailed assertions:\n${concatStringsSep "\n" (map (x: "- ${x}") failedAssertions)}"

       … while calling the 'map' builtin
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:195:26:
          194|     let
          195|       failedAssertions = map (x: x.message) (filter (x: !x.assertion) assertions);
             |                          ^
          196|     in

       … while calling the 'filter' builtin
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:195:46:
          194|     let
          195|       failedAssertions = map (x: x.message) (filter (x: !x.assertion) assertions);
             |                                              ^
          196|     in

       … while calling anonymous lambda
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:195:54:
          194|     let
          195|       failedAssertions = map (x: x.message) (filter (x: !x.assertion) assertions);
          196|     in

       … in the argument of the not operator
         at /nix/store/fhzxfsn1qknfgp5kkx8d46cmc042gfj2-nixos/nixos/lib/asserts.nix:195:58:
          194|     let
          195|       failedAssertions = map (x: x.message) (filter (x: !x.assertion) assertions);
             |                                                          ^
          196|     in

       … while evaluating the attribute 'assertion'
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/lib/utils.internal.nix:137:21:
          136|       processAssertion = a: {
          137|         inherit (a) assertion;
             |                     ^
          138|         message = "${prefix} ${lib.trim a.message}";

       … while evaluating the attribute 'assertion'
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/output.nix:32:5:
           31|   mkVimPluginPackageAssertion = opt: {
           32|     assertion = !(lib.any isVimPluginPackage opt.value);
             |     ^
           33|     message =

       … in the argument of the not operator
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/output.nix:32:19:
           31|   mkVimPluginPackageAssertion = opt: {
           32|     assertion = !(lib.any isVimPluginPackage opt.value);
             |                   ^
           33|     message =

       … while calling the 'any' builtin
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/output.nix:32:19:
           31|   mkVimPluginPackageAssertion = opt: {
           32|     assertion = !(lib.any isVimPluginPackage opt.value);
             |                   ^
           33|     message =

       … while evaluating the attribute 'value'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1147:7:
         1146|     // {
         1147|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1148|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `programs.nixvim.extraPackages':

       … while calling the 'filter' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1136:35:
         1135|       # yield a value computed from the definitions
         1136|       value = if opt ? apply then opt.apply res.mergedValue else res.mergedValue;
         1137|

       … while evaluating the attribute 'mergedValue'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1244:5:
         1243|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1244|     mergedValue =
             |     ^
         1245|       if isDefined then

       … while evaluating the attribute 'value'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:743:13:
          742|             headError = checkDefsForError check loc defs;
          743|             value = map (x: x.optionalValue.value or x.mergedValue) evals;
             |             ^
          744|             valueMeta.list = map (v: v.checkedAndMerged.valueMeta) evals;

       … while calling the 'map' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:743:21:
          742|             headError = checkDefsForError check loc defs;
          743|             value = map (x: x.optionalValue.value or x.mergedValue) evals;
             |                     ^
          744|             valueMeta.list = map (v: v.checkedAndMerged.valueMeta) evals;

       … while calling the 'filter' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:724:21:
          723|           let
          724|             evals = filter (x: x.optionalValue ? value) (
             |                     ^
          725|               concatLists (

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:724:29:
          723|           let
          724|             evals = filter (x: x.optionalValue ? value) (
             |                             ^
          725|               concatLists (

       … while evaluating the attribute 'optionalValue'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1310:5:
         1309|
         1310|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |     ^
         1311|   };

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1310:21:
         1309|
         1310|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                     ^
         1311|   };

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1221:7:
         1220|       # above are lazy and thus never forced on this branch.
         1221|       if
             |       ^
         1222|         length defs == 1

       … in the right operand of the AND (&&) operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1223:9:
         1222|         length defs == 1
         1223|         && (
             |         ^
         1224|           let

       … while evaluating definitions from `/nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix':

       … in the argument of the not operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1228:31:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |                               ^
         1229|           )

       … in the left operand of the AND (&&) operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1228:31:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |                               ^
         1229|           )

       … while calling the 'isAttrs' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1228:15:
         1227|           addErrorContext "while evaluating definitions from `${d.file}':" (
         1228|             !(isAttrs d.value && d.value ? _type)
             |               ^
         1229|           )

       … while evaluating the attribute 'value'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:733:25:
          732|                         inherit (def) file;
          733|                         value = def';
             |                         ^
          734|                       }

       … while calling the 'elemAt' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/lists.nix:347:43:
          346|   */
          347|   imap1 = f: list: genList (n: f (n + 1) (elemAt list n)) (length list);
             |                                           ^
          348|

       … while calling anonymous lambda
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix:67:28:
           66|     (lib.partition (p: p.packageFallback))
           67|     (lib.mapAttrs (_: map (p: p.package)))
           68|     (parts: {

       … from call site
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix:67:31:
           66|     (lib.partition (p: p.packageFallback))
           67|     (lib.mapAttrs (_: map (p: p.package)))
             |                               ^
           68|     (parts: {

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:1187:17:
         1186|         mapAttrs (
         1187|           name: value:
             |                 ^
         1188|           if isAttrs value && cond value then recurse (path ++ [ name ]) value else f (path ++ [ name ]) value

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:1188:85:
         1187|           name: value:
         1188|           if isAttrs value && cond value then recurse (path ++ [ name ]) value else f (path ++ [ name ]) value
             |                                                                                     ^
         1189|         );

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:283:71:
          282|           # For definitions that have an associated option
          283|           declaredConfig = mapAttrsRecursiveCond (v: !isOption v) (_: v: v.value) options;
             |                                                                       ^
          284|

       … while evaluating the attribute 'value'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1147:7:
         1146|     // {
         1147|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1148|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `programs.nixvim.dependencies.bat.package':

       … while evaluating the attribute 'mergedValue'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1244:5:
         1243|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1244|     mergedValue =
             |     ^
         1245|       if isDefined then

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1245:7:
         1244|     mergedValue =
         1245|       if isDefined then
             |       ^
         1246|         if type.merge ? v2 then

       … while evaluating the attribute 'values'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1238:11:
         1237|         {
         1238|           values = defsSorted;
             |           ^
         1239|           inherit (defsFiltered) highestPrio;

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1209:11:
         1208|           # Avoid sorting if we don't have to.
         1209|           if any (def: def.value._type or "" == "order") defsFiltered.values then
             |           ^
         1210|             sortProperties defsFiltered.values

       … while calling the 'any' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1209:14:
         1208|           # Avoid sorting if we don't have to.
         1209|           if any (def: def.value._type or "" == "order") defsFiltered.values then
             |              ^
         1210|             sortProperties defsFiltered.values

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1209:19:
         1208|           # Avoid sorting if we don't have to.
         1209|           if any (def: def.value._type or "" == "order") defsFiltered.values then
             |                   ^
         1210|             sortProperties defsFiltered.values

       … while evaluating the attribute 'value._type'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1420:67:
         1419|       strip =
         1420|         def: if def.value._type or "" == "override" then def // { value = def.value.content; } else def;
             |                                                                   ^
         1421|     in

       … while evaluating the attribute 'value.content'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1564:22:
         1563|     _type = "override";
         1564|     inherit priority content;
             |                      ^
         1565|   };

       … while evaluating the attribute 'default'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/options.nix:325:13:
          324|           {
          325|             default = defaultValue;
             |             ^
          326|             defaultText = literalExpression "${pkgsText}.${defaultText}";

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/options.nix:321:22:
          320|       defaultText = showAttrPath default';
          321|       defaultValue = attrByPath default' (throw "${defaultText} cannot be found in ${pkgsText}") pkgs;
             |                      ^
          322|       defaults =

       … while calling 'attrByPath'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:88:24:
           87|   attrByPath =
           88|     attrPath: default: set:
             |                        ^
           89|     let

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:105:5:
          104|     in
          105|     attrByPath' 0 set;
             |     ^
          106|

       … while calling 'attrByPath''
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:92:12:
           91|       attrByPath' =
           92|         n: s:
             |            ^
           93|         (

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:101:15:
          100|               in
          101|               if s ? ${attr} then attrByPath' (n + 1) s.${attr} else default
             |               ^
          102|             )

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/attrsets.nix:101:18:
          100|               in
          101|               if s ? ${attr} then attrByPath' (n + 1) s.${attr} else default
             |                  ^
          102|             )

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:725:15:
          724|       extraArgs = mapAttrs (
          725|         name: _:
             |               ^
          726|         addErrorContext (context name) (

       … while evaluating the module argument `pkgs' in "/nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/dependencies.nix":

       … noting that argument `pkgs` is not externally provided, so querying `_module.args` instead, requiring `config`

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:729:13:
          728|             "noting that argument `${name}` is not externally provided, so querying `_module.args` instead, requiring `config`"
          729|             config._module.args.${name}
             |             ^
          730|           )

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/types.nix:1026:18:
         1025|             value = mapAttrs (
         1026|               n: v:
             |                  ^
         1027|               if lazy then

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/modules.nix:1310:5:
         1309|
         1310|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |     ^
         1311|   };

       (7 duplicate frames omitted)

       … from call site
         at /nix/store/w6727xld8vsc1hb2yqm7p2z2nzzxlcy5-source/modules/top-level/nixpkgs.nix:240:11:
          239|           in
          240|           import cfg.source (args // systemArgs);
             |           ^
          241|

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/impure.nix:12:1:
           11|
           12| {
             | ^
           13|   # We put legacy `system` into `localSystem`, if `localSystem` was not passed.

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/impure.nix:55:1:
           54|
           55| import ./. (
             | ^
           56|   removeAttrs args [ "system" ]

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/default.nix:40:1:
           39|
           40| {
             | ^
           41|   # The system packages will be built on. See the manual for the

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/default.nix:220:1:
          219| in
          220| checked pkgs
             | ^
          221|

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/default.nix:112:11:
          111|         else
          112|           x: x
             |           ^
          113|       );

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/default.nix:217:10:
          216|
          217|   pkgs = boot stages;
             |          ^
          218|

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:42:1:
           41| # other words, this does a foldr not foldl.
           42| stageFuns:
             | ^
           43| let

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:165:10:
          164|
          165|   pkgs = dfold folder postStage (_: { }) withAllowCustomOverrides;
             |          ^
          166|

       … while calling 'dfold'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:63:21:
           62|   dfold =
           63|     op: lnul: rnul: list:
             |                     ^
           64|     let

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:78:13:
           77|       lapp = lnul cur;
           78|       cur = go lapp 0;
             |             ^
           79|     in

       … while calling 'go'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:67:15:
           66|       go =
           67|         pred: n:
             |               ^
           68|         if n == len then

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:73:19:
           72|             # Note the cycle -- call-by-need ensures finite fold.
           73|             cur = op pred (builtins.elemAt list n) succ;
             |                   ^
           74|             succ = go cur (n + 1);

       … while calling 'folder'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:96:26:
           95|   folder =
           96|     nextStage: stageFun: prevStage:
             |                          ^
           97|     let

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:107:9:
          106|       thisStage =
          107|         if args.__raw or false then
             |         ^
          108|           args'

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:98:14:
           97|     let
           98|       args = stageFun prevStage;
             |              ^
           99|       args' = args // {

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:86:22:
           85|   withAllowCustomOverrides = lib.lists.imap1 (
           86|     index: stageFun: prevStage:
             |                      ^
           87|     # So true by default for only the first element because one

       … in the right operand of the update (//) operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:90:48:
           89|     # for the final stage.
           90|     { allowCustomOverrides = index == 1; } // (stageFun prevStage)
             |                                                ^
           91|   ) (lib.lists.reverseList stageFuns);

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:90:48:
           89|     # for the final stage.
           90|     { allowCustomOverrides = index == 1; } // (stageFun prevStage)
             |                                                ^
           91|   ) (lib.lists.reverseList stageFuns);

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:879:5:
          878|   (
          879|     prevStage:
             |     ^
          880|     # previous stage5 stdenv; see stage3 comment regarding gcc,

       … in the condition of the assert statement
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:882:5:
          881|     # which applies here as well.
          882|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |     ^
          883|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:882:12:
          881|     # which applies here as well.
          882|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |            ^
          883|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       … while calling 'isBuiltByNixpkgsCompiler'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:30:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
             |                              ^
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … in the left operand of the AND (&&) operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:53:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
             |                                                     ^
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:35:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … while calling 'isFromNixpkgs'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:19:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                   ^
          127|   isFromBootstrapFiles =

       … in the argument of the not operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:26:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                          ^
          127|   isFromBootstrapFiles =

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:26:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                          ^
          127|   isFromBootstrapFiles =

       … while calling 'isFromBootstrapFiles'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:128:5:
          127|   isFromBootstrapFiles =
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
             |     ^
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/booter.nix:74:20:
           73|             cur = op pred (builtins.elemAt list n) succ;
           74|             succ = go cur (n + 1);
             |                    ^
           75|           in

       (8 duplicate frames omitted)

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:687:5:
          686|   (
          687|     prevStage:
             |     ^
          688|     # previous stage4 stdenv; see stage3 comment regarding gcc,

       … in the condition of the assert statement
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:690:5:
          689|     # which applies here as well.
          690|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |     ^
          691|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:690:12:
          689|     # which applies here as well.
          690|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |            ^
          691|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       (16 duplicate frames omitted)

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:616:5:
          615|   (
          616|     prevStage:
             |     ^
          617|     # previous stage3 stdenv:

       … in the condition of the assert statement
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:618:5:
          617|     # previous stage3 stdenv:
          618|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |     ^
          619|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:618:12:
          617|     # previous stage3 stdenv:
          618|     assert isBuiltByNixpkgsCompiler prevStage.binutils-unwrapped;
             |            ^
          619|     assert isBuiltByNixpkgsCompiler prevStage.${localSystem.libc};

       … while calling 'isBuiltByNixpkgsCompiler'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:30:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
             |                              ^
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … in the right operand of the AND (&&) operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:53:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
             |                                                     ^
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:129:56:
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;
             |                                                        ^
          130|   isBuiltByBootstrapFilesCompiler = pkg: isFromNixpkgs pkg && isFromBootstrapFiles pkg.stdenv.cc.cc;

       … while calling 'isFromNixpkgs'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:19:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                   ^
          127|   isFromBootstrapFiles =

       … in the argument of the not operator
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:26:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                          ^
          127|   isFromBootstrapFiles =

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:126:26:
          125|
          126|   isFromNixpkgs = pkg: !(isFromBootstrapFiles pkg);
             |                          ^
          127|   isFromBootstrapFiles =

       … while calling 'isFromBootstrapFiles'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:128:5:
          127|   isFromBootstrapFiles =
          128|     pkg: pkg.passthru.isFromBootstrapFiles or pkg.passthru.isFromMinBootstrap or false;
             |     ^
          129|   isBuiltByNixpkgsCompiler = pkg: isFromNixpkgs pkg && isFromNixpkgs pkg.stdenv.cc.cc;

       … while evaluating the attribute 'stdenv.cc.cc'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/default.nix:236:15:
          235|
          236|       inherit cc hasCC;
             |               ^
          237|

       … while calling a functor (an attribute set with a '__functor' attribute)
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/linux/default.nix:201:14:
          200|           else
          201|             (lib.makeOverridable (import ../../build-support/cc-wrapper) {
             |              ^
          202|               name = "${name}-gcc-wrapper";

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/trivial.nix:1183:7:
         1182|     g: {
         1183|       __functor = self: g;
             |       ^
         1184|       __functionArgs = fArgs;

       … while calling 'f''
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/customisation.nix:158:9:
          157|       f' =
          158|         origArgs:
             |         ^
          159|         let

         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/customisation.nix:174:9:
          173|         in
          174|         if isAttrs result then
             |         ^
          175|           result

       … while calling the 'isAttrs' builtin
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/customisation.nix:174:12:
          173|         in
          174|         if isAttrs result then
             |            ^
          175|           result

       … from call site
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/lib/customisation.nix:160:20:
          159|         let
          160|           result = f origArgs;
             |                    ^
          161|

       … while calling anonymous lambda
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/build-support/cc-wrapper/default.nix:8:1:
            7|
            8| {
             | ^
            9|   name ? "",

       … in the condition of the assert statement
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/build-support/cc-wrapper/default.nix:396:1:
          395| # Ensure bintools matches
          396| assert libc_bin == bintools.libc_bin;
             | ^
          397| assert libc_dev == bintools.libc_dev;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'glibc-2.42-61'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'buildInputs' of derivation 'glibc-2.42-61'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:722:11:
          721|           depsHostHost = hostHostOutputs;
          722|           buildInputs = hostTargetOutputs;
             |           ^
          723|           depsTargetTarget = targetTargetOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'linux-headers-6.18.7'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'depsBuildBuild' of derivation 'linux-headers-6.18.7'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:718:11:
          717|
          718|           depsBuildBuild = buildBuildOutputs;
             |           ^
          719|           nativeBuildInputs = buildHostOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'bootstrap-stage2-gcc-wrapper-15.2.0'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'bintools' of derivation 'bootstrap-stage2-gcc-wrapper-15.2.0'

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'binutils-wrapper-2.46'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'bintools_bin' of derivation 'binutils-wrapper-2.46'

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'binutils-2.46'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'buildInputs' of derivation 'binutils-2.46'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:722:11:
          721|           depsHostHost = hostHostOutputs;
          722|           buildInputs = hostTargetOutputs;
             |           ^
          723|           depsTargetTarget = targetTargetOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'gettext-1.0'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'buildInputs' of derivation 'gettext-1.0'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:722:11:
          721|           depsHostHost = hostHostOutputs;
          722|           buildInputs = hostTargetOutputs;
             |           ^
          723|           depsTargetTarget = targetTargetOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'bash-5.3p9'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'nativeBuildInputs' of derivation 'bash-5.3p9'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:719:11:
          718|           depsBuildBuild = buildBuildOutputs;
          719|           nativeBuildInputs = buildHostOutputs;
             |           ^
          720|           depsBuildTarget = buildTargetOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'bison-3.8.2'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'nativeBuildInputs' of derivation 'bison-3.8.2'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:719:11:
          718|           depsBuildBuild = buildBuildOutputs;
          719|           nativeBuildInputs = buildHostOutputs;
             |           ^
          720|           depsBuildTarget = buildTargetOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'perl-5.42.0'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'depsBuildBuild' of derivation 'perl-5.42.0'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:718:11:
          717|
          718|           depsBuildBuild = buildBuildOutputs;
             |           ^
          719|           nativeBuildInputs = buildHostOutputs;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'make-shell-wrapper-hook'
         whose name attribute is located at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/stdenv/generic/make-derivation.nix:651:11

       … while evaluating attribute 'shell' of derivation 'make-shell-wrapper-hook'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/all-packages.nix:781:7:
          780|       # targetPackages.runtimeShell only exists when pkgs == targetPackages (when targetPackages is not  __raw)
          781|       shell =
             |       ^
          782|         if targetPackages ? runtimeShell then

       … while evaluating the attribute 'runtimeShell'
         at /nix/store/j2r11kxv91yl5xqppy3vy84klwxjbz1i-source/pkgs/top-level/all-packages.nix:3238:3:
         3237|
         3238|   runtimeShell = "${runtimeShellPackage}${runtimeShellPackage.shellPath}";
             |   ^
         3239|   runtimeShellPackage = bashNonInteractive;

       … while calling the 'getAttr' builtin
         at «nix-internal»/derivation-internal.nix:50:17:
           49|     value = commonAttrs // {
           50|       outPath = builtins.getAttr outputName strict;
             |                 ^
           51|       drvPath = strict.drvPath;

Command 'nix-build '<nixpkgs/nixos>' --attr config.system.build.toplevel --no-out-link --show-trace' returned non-zero exit status 1.
```

My configuration is just like [this](https://github.com/baizhu945/BaizhuNix/blob/main/nixos/neovim.nix) .

```nix

```

khaneliman:
Nixvim wrapper modules passed the host nixpkgs' fully elaborated stdenv.{host,build}Platform attrsets into Nixvim's own nixpkgs import.

When the host and Nixvim nixpkgs revisions disagreed, those platform attrsets could be re-elaborated by a different lib.systems, causing eval recursion or platform-schema errors.

Pass only { config = ...; } as the inherited platform spec instead. This keeps ABI/libc details while letting Nixvim's selected nixpkgs elaborate the platform itself.

Closes https://github.com/nix-community/nixvim/issues/4460
Closes https://github.com/nix-community/nixvim/issues/4426

### DIFF
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -168,6 +168,12 @@ in
         To cross-compile, also set `nixpkgs.buildPlatform`.
 
         Ignored when `nixpkgs.pkgs` is set.
+
+        > [!WARNING]
+        > Avoid passing a fully elaborated platform from another Nixpkgs instance,
+        > such as `pkgs.stdenv.hostPlatform`, when `nixpkgs.source` may point to a
+        > different Nixpkgs revision. Prefer a platform spec, such as
+        > `{ inherit (pkgs.stdenv.hostPlatform) config; }`, or a system string.
       '';
     };
 
@@ -195,6 +201,12 @@ in
         Setting this option will cause Nixvim to be cross-compiled.
 
         Ignored when `nixpkgs.pkgs` is set.
+
+        > [!WARNING]
+        > Avoid passing a fully elaborated platform from another Nixpkgs instance,
+        > such as `pkgs.stdenv.buildPlatform`, when `nixpkgs.source` may point to a
+        > different Nixpkgs revision. Prefer a platform spec, such as
+        > `{ inherit (pkgs.stdenv.buildPlatform) config; }`, or a system string.
       '';
     };
 
--- a/wrappers/_shared.nix
+++ b/wrappers/_shared.nix
@@ -44,8 +44,8 @@ let
         pkgs = lib.mkIf config.nixpkgs.useGlobalPackages (lib.mkDefault pkgs);
 
         # Inherit platform spec
-        hostPlatform = lib.mkOptionDefault pkgs.stdenv.hostPlatform;
-        buildPlatform = lib.mkOverride buildPlatformPrio pkgs.stdenv.buildPlatform;
+        hostPlatform = lib.mkOptionDefault { inherit (pkgs.stdenv.hostPlatform) config; };
+        buildPlatform = lib.mkOverride buildPlatformPrio { inherit (pkgs.stdenv.buildPlatform) config; };
       };
     };
 


modules/top-level/nixpkgs.nix
```diff
@@ -168,6 +168,12 @@ in
         To cross-compile, also set `nixpkgs.buildPlatform`.
 
         Ignored when `nixpkgs.pkgs` is set.
+
+        > [!WARNING]
+        > Avoid passing a fully elaborated platform from another Nixpkgs instance,
+        > such as `pkgs.stdenv.hostPlatform`, when `nixpkgs.source` may point to a
+        > different Nixpkgs revision. Prefer a platform spec, such as
+        > `{ config = pkgs.stdenv.hostPlatform.config; }`, or a system string.
```
MattSturgeon: Should we encourage using `inherit` here?

```suggestion
        > `{ inherit (pkgs.stdenv.hostPlatform) config; }`, or a system string.
```


 I'm also a little concerned this example could be confused with the pkgs `config` used for things like `allowUnfree`.

MattSturgeon (approved):
This should help, but I'm not available to test it out today. I suspect even `(platform).config` has its own footguns, perhaps w.r.t. overridden/modified platforms.[^1]

I also think this may not be enough to _fully_ address ABI compatibility. If I've understood @emilazy's comments [here](https://github.com/NixOS/nixpkgs/pull/530133#discussion_r3417428329) and [here](https://github.com/NixOS/nixpkgs/pull/532194#issuecomment-4785107844) correctly, we may need to elaborate systems using a `lib` from the same Nixpkgs revisions we plan to instantiate `pkgs` from. I.e.:

```nix
libForPkgs = import (cfg.source + "/lib");
```

Then use `libForPkgs.systems.elaborate`. We'd also have to be careful to assert anything the `assertions` option would no longer catch, but I think this lib-import would only error if `cfg.source` isn't a Nixpkgs path.

It's a little inefficient, to instantiate another `lib` just for this tiny use case, when we already have a `lib` available, but maybe a necessary evil for ABI compatibility.

This would also apply to other Nixpkgs instantiating modules, like home-manager and nix-darwin, if they elaborate their platform options like upstream NixOS does.

This would not apply to NixOS because it has no `nixpkgs.source` option, so the NixOS `lib` argument is guaranteed to be instantiated from the same Nixpkgs revisions it will pass the platforms to when instantiating.

---

I've not thought too hard about the warning you added to the docs descriptions, but they seem reasonable at first glance. If a little unfortunate.

[^1]: There was a lot of discussion on this kinda thing back when Wolfgang was (trying to) work towards extensible platform specs, and when NixOS flip-flopped on whether it would elaborate platform options.

emilazy:
> This would also apply to other Nixpkgs instantiating modules, like home-manager and nix-darwin, if they elaborate their platform options like upstream NixOS does.
> 
> This would not apply to NixOS because it has no `nixpkgs.source` option, so the NixOS `lib` argument is guaranteed to be instantiated from the same Nixpkgs revisions it will pass the platforms to when instantiating.

I would say instead: setting `nixpkgs.source` in such a way that you get a mismatched `lib` is a configuration error (that may not be explicitly detected). nix-darwin [does try to avoid this happening by accident](https://github.com/nix-darwin/nix-darwin/blob/a1fa429e945becaf60468600daf649be4ba0350c/flake.nix#L22-L23). Essentially, `lib` and `pkgs.lib` differing is a recipe for sadness.

(In other words, it shouldn’t be necessary to instantiate another `lib` because the case where you’d need to is already screwy. But I don’t know much about NixVim or what it’s doing specifically.)

MattSturgeon:
> setting `nixpkgs.source` in such a way that you get a mismatched `lib` is a configuration error

I don't see how this can be avoided. Setting `nixpkgs.source` _at all_ implies that `pkgs` will be using a different Nixpkgs revision to `lib`'s revision. Due to the module system's fixpoints we can't construct a different `lib`; `lib` is fixed outside of the module eval (in fact it evaluates the modules themselves) while `_module.args.pkgs` is defined by modules, within the `config` fixpoint.

My understanding is that slightly mismatched libs should not usually be a problem:
1. Packages use `lib` as supplied by `callPackage`, which is instantiated from the same tree
2. One `lib` being combined with another is usually handled gracefully:
   - most of `lib` is pure functions
   - module system code is usually resilient to (e.g.) types constructed by other `lib`s
3. `lib.systems.elaborate` seems to be the exception to the rule

In fact, if we had an externally defined `pkgs` then `lib.systems.elaborate` is no longer relevant; as we only evaluate our `nixpkgs.*Platform` options when importing `nixpkgs.source` anyway.

> nix-darwin [does try to avoid this happening by accident](https://github.com/nix-darwin/nix-darwin/blob/a1fa429e945becaf60468600daf649be4ba0350c/flake.nix#L22-L23).

Looks like the main difference is that nix-darwin accepts an externally defined `pkgs` in its entrypoint; Nixvim does not, [our entrypoint](https://nix-community.github.io/nixvim/lib/nixvim/modules/index.html#lib.nixvim.modules.evalNixvim) only expects `modules` and optionally `system`.

An end-user configuring nix-darwins's `pkgs` via nix-darwin's [`nixpkgs.source`](https://github.com/nix-darwin/nix-darwin/blob/a1fa429e945becaf60468600daf649be4ba0350c/modules/nix/nixpkgs.nix#L246-L261) option would still be essentially doing the same thing as an end user configuring Nixvim or Home Manager's respective `nixpkgs.source` options.

emilazy:
> Setting `nixpkgs.source` _at all_ implies that `pkgs` will be using a different Nixpkgs revision to `lib`'s revision.

I added `nixpkgs.source` to nix-darwin and I certainly didn’t intend to imply it’s supported to use it in such a way; if I’ve caused the proliferation of that presumption in the ecosystem then I regret it. `nixpkgs.source` is there to pipe the relevant default through and to ensure that version information propagates when you customize `pkgs` in other ways, not to allow mixing `pkgs` and `lib` versions.

> * One `lib` being combined with another is usually handled gracefully:
>   
>   * most of `lib` is pure functions
>   * module system code is usually resilient to (e.g.) types constructed by other `lib`s

Well, sure; mixing and matching `lib` versions works fine *other than* the parts that are too coupled to mix :)

Note that even in the former case there’s no guarantee of that though: those pure functions can still change behaviour sometimes, and the internal representation of things like `lib.fileset` are explicitly documented as being unstable implementation details, so they can change in a way that makes mixing versions incompatible without breaking supported API surface.

With the parts that are coupled to Nixpkgs, which is a bit more than just `lib.systems`, it’s definitely fully unsupported.

It might be nice if this was a thing that worked and was supported, but it’s not. I’ve seen people end up with horribly tricky‐to‐debug mass rebuilds because of mismatched `lib`/`pkgs` versions. If NixVim doesn’t let people override the Nixpkgs to use in a way that makes the `lib` match up, I think that’s a NixVim bug.

MattSturgeon:
> > Setting `nixpkgs.source` _at all_ implies that `pkgs` will be using a different Nixpkgs revision to `lib`'s revision.
> 
> I added `nixpkgs.source` to nix-darwin and I certainly didn’t intend to imply it’s supported to use it in such a way; if I’ve caused the proliferation of that presumption in the ecosystem then I regret it. `nixpkgs.source` is there to pipe the relevant default through [...] not to allow mixing `pkgs` and `lib` versions.

That would make sense to me if `nixpkgs.source` was a readOnly option or an internal let binding, only used to propagate a default. But as a user-writable option, the only reason a user will define it is to _change_ the default to a different Nixpkgs source (implying a different Nixpkgs revision).

If that same `nixpkgs.source` is used to construct `pkgs` internally within the module eval, then `pkgs` and `lib` are almost certainly from different revisions.

If we consider this to be a design flaw in `nixpkgs.source`, then nix-darwin, Home Manager, Nixvim, etc should add warnings (and later assertions) when `constructedByMe && options.nixpkgs.source.highestPrio != 1500`.

> and to ensure that version information propagates when you customize `pkgs` in other ways

I guess that's why `nixpkgs.source` isn't readOnly, but perhaps in hindsight dedicated `nixpkgs.rev` (etc) version metadata options would have better represented that abstraction?

---

Further, the `nixpkgs.pkgs` option (also present in NixOS), `_module.args.pkgs`, and even `specialArgs.pkgs`[^1] suffer the same issue; there are two reasons to define these manually:
1. To reuse an already instantiated Nixpkgs
2. To use a different revision of Nixpkgs

If a user defines these to a different revision than the `lib` used to eval the modules, then they are exposed to the latent bug you describe.

In practice, I think these subtle mis-matches are a non-issue,[^2][^3] as the `lib` used to evaluate modules almost never interacts with package internals or the `lib` used by packages. But if we were to be strict about `nixpkgs.source`, the same issue also applies to other methods of providing `pkgs`.

Another footgun would be defining `specialArgs.lib`. That means that some parts of the module eval will be done by the `lib` that `evalModules` came from, while other parts of the module eval will use the `lib` from `specialArgs`. This seems more problematic to me than the mostly isolated lib+pkgs mismatch.

[^1]: none of _these_ examples are likely to use the `nixpkgs.{host,build}Platform` options, so are less exposed to `lib.systems.elaborate` differences.
[^2]: though I'm interested to learn about edge cases where it _does_ become a real issue.
[^3]: one example is, of course, elaborating a system using one revision's `lib`, then using that system to instantiate another Nixpkgs revision.

PR #4464, opened 2026-06-27
plugins/tiny-cmdline: init

### LINKED ISSUE #4422: [PLUGIN REQUEST] tiny-cmdline.nvim
### Plugin

tiny-cmdline.nvim

### Homepage

https://github.com/rachartier/tiny-cmdline.nvim

### Nixpkgs URL

_No response_

### Extra Information

This uses the new ui2 component present in neovim 0.12+.
So Neovim 0.12+ is hard requirement for this plugin.

ZainKergaye:
Fixes: #4422 

No AI was used to make this. 

Plugin was easy to implement. Requires `vim._core.ui2`. Not sure how to add that in. Any suggestions are appreciated.

### DIFF
--- a/plugins/by-name/tiny-cmdline/default.nix
+++ b/plugins/by-name/tiny-cmdline/default.nix
@@ -1,0 +1,17 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "tiny-cmdline";
+  package = "tiny-cmdline-nvim";
+
+  maintainers = [ lib.maintainers.zainkergaye ];
+
+  description = "A Neovim plugin that repositions the cmdline as a centered floating window, powered by Neovim's native ui2 system.";
+
+  callSetup = false;
+  extraConfig.plugins.tiny-cmdline.luaConfig.content = ''
+    	
+    		require('vim._core.ui2').enable({})
+    		require('tiny-cmdline')
+  '';
+
+}
--- a/tests/test-sources/plugins/by-name/tiny-cmdline/default.nix
+++ b/tests/test-sources/plugins/by-name/tiny-cmdline/default.nix
@@ -1,0 +1,3 @@
+{
+  empty.plugins.tiny-cmdline.enable = true;
+}


ZainKergaye:
Got inspiration from plugins/faust. This does feel like it should be at the top of the config. I could add: 
```nix 
  configLocation = lib.mkOrder 900 "extraConfigLua";
```
but i'm not sure (taken from java plugin)

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "tiny-cmdline";
+  package = "tiny-cmdline-nvim";
+
+  maintainers = [ lib.maintainers.zainkergaye ];
+
+  description = "A Neovim plugin that repositions the cmdline as a centered floating window, powered by Neovim's native ui2 system.";
+
+  callSetup = false;
```
khaneliman, line 10: issue (blocking): `callSetup = false` while `hasSettings` stays `true`, so `plugins.tiny-cmdline.settings` is generated and documented but never reaches the plugin.

Upstream supports `setup()` and documents that it takes priority over `vim.g`. Use `callSetup = "optional"` instead: setup then runs only when the user defines settings, and the UIEnter auto-init still covers the empty case (it bails on `_initialized`).

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  extraConfig.plugins.tiny-cmdline.luaConfig.content = ''
+    	
+    		require('vim._core.ui2').enable({})
+    		require('tiny-cmdline')
```
khaneliman, line 14: issue: `require('tiny-cmdline')` is a no-op here.

Upstream ships `plugin/tiny-cmdline.lua`, which nixvim loads from the packpath and which calls `setup()` itself on `UIEnter`. Requiring the module at config time only loads it without initializing. Drop this line.

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  name = "tiny-cmdline";
+  package = "tiny-cmdline-nvim";
+
+  maintainers = [ lib.maintainers.zainkergaye ];
+
+  description = "A Neovim plugin that repositions the cmdline as a centered floating window, powered by Neovim's native ui2 system.";
+
+  callSetup = false;
+  extraConfig.plugins.tiny-cmdline.luaConfig.content = ''
+
```
khaneliman, line 12: issue: Nix indented-string stripping only removes leading spaces, not tabs.

The generated Lua literally contains a tab-only first line plus tab-indented statements in the user's init.lua. Re-indent the block with spaces aligned to the `''` opener.

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  package = "tiny-cmdline-nvim";
+
+  maintainers = [ lib.maintainers.zainkergaye ];
+
+  description = "A Neovim plugin that repositions the cmdline as a centered floating window, powered by Neovim's native ui2 system.";
+
+  callSetup = false;
+  extraConfig.plugins.tiny-cmdline.luaConfig.content = ''
+    	
+    		require('vim._core.ui2').enable({})
```
khaneliman, line 13: suggestion: `require('vim._core.ui2').enable({})` unconditionally flips a global, private Neovim UI subsystem with an empty opts table, overriding any ui2 configuration the user set elsewhere.

Consider an `extraOptions` toggle (default true) so users who manage ui2 themselves can opt out, and note in the description that upstream also expects `cmdheight = 0`.

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "tiny-cmdline";
+  package = "tiny-cmdline-nvim";
+
+  maintainers = [ lib.maintainers.zainkergaye ];
+
+  description = "A Neovim plugin that repositions the cmdline as a centered floating window, powered by Neovim's native ui2 system.";
```
khaneliman, line 8: suggestion: no `settingsExample`, which CONTRIBUTING.md calls highly recommended.

A small example such as `{ width.value = "70%"; position = { x = "50%"; y = "50%"; }; }` renders in the generated docs and is enough.

tests/test-sources/plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,3 @@
+{
+  empty.plugins.tiny-cmdline.enable = true;
```
khaneliman, line 2: suggestion: only the `empty` case is covered.

Once setup is wired up, add an `example` case with a realistic settings table (matching `settingsExample`). Recently merged inits include both `empty` and `example`.

khaneliman (changes requested):

plugins/by-name/tiny-cmdline/default.nix
```diff
@@ -0,0 +1,17 @@
...
+
+  callSetup = false;
+  extraConfig.plugins.tiny-cmdline.luaConfig.content = ''
```
khaneliman: issue: plugin settings are silently ignored

`mkNeovimPlugin` still exposes `plugins.tiny-cmdline.settings`, but this static `extraConfig` never uses that value, so the plugin receives only `vim.g.tiny_cmdline` or `{}`. Any Nixvim setting is silently ignored; pass `cfg.settings` to `require('tiny-cmdline').setup(...)` or disable the generated `settings` option.

plugins/by-name/tiny-cmdline/default.nix
khaneliman: issue: lazy loading leaves the plugin uninitialized

When `lazyLoad` is configured, Nixvim moves this content into lz.n's `after` callback, which runs only when the trigger fires. The pinned plugin registers its automatic setup call on one-shot `UIEnter`, so a `cmd` or `keys` trigger after startup runs only this `require` and leaves the plugin inactive; call `setup` from the lazy callback or reject lazy loading for this module.

PR #4467, opened 2026-06-30
modules/nixpkgs: use `libForPkgs` for `elaborate`

MattSturgeon:
Followup to #4461

While all of `lib` is _technically_ coupled to its Nixpkgs revision and should ideally be kept in sync with any `pkgs` instances it'll be exposed to, `lib.systems.elaborate` is particularly sensitive to differences in Nixpkgs revision. We can mitigate that by importing a `lib` instance from the Nixpkgs revision we're constructing (`nixpkgs.source`).

This means when our Nixpkgs module constructs a Nixpkgs instance (`constructedByMe == true`), it'll instantiate `nixpkgs.source` with platforms elaborated by a `lib` from that same Nixpkgs revision.

For efficiency in the happy path, we only do this when `nixpkgs.source` has been overridden (`highestPrio != 1500`). The default value should already be aligned with our `lib`, as both are derived from our `nixpkgs` flake input.[^1]

[^1]: there's a case to be made that our `sepcialArgs.lib` normalization could also be problematic in the nested module scenario, as we're then mixing lib revisions with outer/inner module evaluations.

### DIFF
--- a/modules/top-level/nixpkgs.nix
+++ b/modules/top-level/nixpkgs.nix
@@ -16,6 +16,13 @@ let
 
   optionDefaultPrio = (lib.mkOptionDefault null).priority;
 
+  # lib.systems.elaborate is sensitive to Nixpkgs revision, so import a `lib` instance from the Nixpkgs revision we're constructing.
+  libForPkgs =
+    if opt.source.highestPrio == optionDefaultPrio then
+      lib
+    else
+      cfg.source.lib or (import cfg.source + "/lib");
+
   isConfig = x: builtins.isAttrs x || lib.isFunction x;
 
   mergeConfig =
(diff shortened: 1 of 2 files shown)


khaneliman (approved):

modules/top-level/nixpkgs.nix
```diff
@@ -16,6 +16,13 @@ let
 
   optionDefaultPrio = (lib.mkOptionDefault null).priority;
 
+  # lib.systems.elaborate is sensitive to Nixpkgs revision, so import a `lib` instance from the Nixpkgs revision we're constructing.
+  libForPkgs =
+    if opt.source.highestPrio == optionDefaultPrio then
+      lib
+    else
+      cfg.source.lib or (import cfg.source + "/lib");
```
MattSturgeon, line 24: Oops, typo:
```suggestion
      cfg.source.lib or (import (cfg.source + "/lib"));
```

PR #4478, opened 2026-07-03
docs: opts differences

belamu:
While trying to create my nixvim setup, I will try to add the documentation to options that would have helped me to understand faster.

Since documentation does not depend a lot on other documentation, the commits can also be cherry-picked if only some of them are wanted.

### DIFF
--- a/modules/opts.nix
+++ b/modules/opts.nix
@@ -5,21 +5,42 @@ let
       prettyName = "options";
       luaVariableName = "options";
       luaApi = "opt";
-      description = "The configuration options, e.g. line numbers (`vim.opt.*`)";
+      description = ''
+        The configuration options, e.g. line numbers (`vim.opt.*`)
+
+        These options are set as global and local.
+
+        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
     };
 
     globalOpts = {
       prettyName = "global options";
       luaVariableName = "global_options";
       luaApi = "opt_global";
-      description = "The configuration global options (`vim.opt_global.*`)";
+      description = ''
+        The configuration global options (`vim.opt_global.*`)
+
+        These options are overridden for buffers that have the same option set as a local option.
+        If a configuration set in `globalOpts` seems to have no effect, check with `:setglobal <option>?`, `:setlocal <option>?`
+        and move the option declaration into `opts`.
+
+        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
     };
 
     localOpts = {
       prettyName = "local options";
       luaVariableName = "local_options";
       luaApi = "opt_local";
-      description = "The configuration local options (`vim.opt_local.*`)";
+      description = ''
+        The configuration local options (`vim.opt_local.*`)
+
+        These options are applied to the buffer on startup of neovim with the global options (`opt_global`) as a fallback.
+        When opening a new buffer, e.g. with `:new`, the global options are used, making `localOpts` irrelevant for the new buffer.
+
+        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
     };
 
     globals = {
--- a/docs/user-configs/list.toml
+++ b/docs/user-configs/list.toml
@@ -190,7 +190,7 @@ url = "https://github.com/literally-sai/vermvim"
 owner = "carl0xs"
 title = "nixcfg"
 description = "Nix configuration setup, including nixvim"
-url = "https://github.com/carl0xs/nixcfg/blob/main/modules/workstation/neovim.nix"
+url = "https://github.com/carl0xs/nixcfg/blob/main/modules/nvim/default.nix"
 
 [[config]]
 owner = "ar-at-localhost"


belamu:
When I ran `nix run .#docs` in the repository root, I get the error `error: cannot find flake 'flake:docs' in the flake registries`. Therefore I cannot test how my changes look.

modules/opts.nix
```diff
@@ -5,21 +5,42 @@ let
...
       luaVariableName = "options";
       luaApi = "opt";
-      description = "The configuration options, e.g. line numbers (`vim.opt.*`)";
+      description = ''
+        The configuration options, e.g. line numbers (`vim.opt.*`)
+
+              These options are set as global and local.
+
+              See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
```
khaneliman, line 14: This paragraph keeps six leading spaces after Nix strips indentation, so CommonMark renders it as a code block.

```suggestion
      description = ''
        The configuration options, e.g. line numbers (`vim.opt.*`)

        These options are set as global and local.

        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
      '';
```

modules/opts.nix
```diff
@@ -5,21 +5,42 @@ let
...
-      description = "The configuration global options (`vim.opt_global.*`)";
+      description = ''
+        The configuration global options (`vim.opt_global.*`)
+
+              These options are overridden for buffers that have the same option set as a local option.
+              If a configuration set in `globalOpts` seems to have no effect, check with `:setglobal <option>?`, `:setlocal <option>?`
+              and move the option declaration into `opts`.
+
+              See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
```
khaneliman, line 29: Same rendering issue here: these continuation lines are indented as a code block in the generated docs.

```suggestion
      description = ''
        The configuration global options (`vim.opt_global.*`)

        These options are overridden for buffers that have the same option set as a local option.
        If a configuration set in `globalOpts` seems to have no effect, check with `:setglobal <option>?`, `:setlocal <option>?`
        and move the option declaration into `opts`.

        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
      '';
```

modules/opts.nix
```diff
@@ -5,21 +5,42 @@ let
...
       luaApi = "opt_local";
-      description = "The configuration local options (`vim.opt_local.*`)";
+      description = ''
+        The configuration local options (`vim.opt_local.*`)
+
+              These options are applied to the buffer on startup of neovim with the global options (`opt_global`) as a fallback.
+              When opening a new buffer, e.g. with `:new`, the global options are used, making `localOpts` irrelevant for the new buffer.
+
+              See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
+      '';
```
khaneliman, line 43: Same here: align prose with the first description line so it renders as text instead of a code block.

```suggestion
      description = ''
        The configuration local options (`vim.opt_local.*`)

        These options are applied to the buffer on startup of neovim with the global options (`opt_global`) as a fallback.
        When opening a new buffer, e.g. with `:new`, the global options are used, making `localOpts` irrelevant for the new buffer.

        See [:h option-list](https://neovim.io/doc/user/quickref/#option-list) for all available options.
      '';
```

khaneliman (dismissed):

MattSturgeon:
> When I ran `nix run .#docs` in the repository root, I get the error `error: cannot find flake 'flake:docs' in the flake registries`. Therefore I cannot test how my changes look.

This _sounds_ like `nix run` recieved `docs` instead of `.#docs`. Have you tried quoting the flakeref argument to avoid your shell doing weird things?

```sh
nix run '.#docs'
```

modules/opts.nix
belamu: Thanks, applied.

This extra whitespace was introduced by the autoformatting of vim. Apparently I have to configure it properly.

modules/opts.nix
belamu: Thanks, applied.

This extra whitespace was introduced by the autoformatting of vim. Apparently I have to configure it properly.

belamu:
> > When I ran `nix run .#docs` in the repository root, I get the error `error: cannot find flake 'flake:docs' in the flake registries`. Therefore I cannot test how my changes look.
> 
> This _sounds_ like `nix run` recieved `docs` instead of `.#docs`. Have you tried quoting the flakeref argument to avoid your shell doing weird things?
> 
> ```shell
> nix run '.#docs'
> ```

Thanks, that worked! I have added it as another commit in `CONTRIBUTING.md`.

MattSturgeon:
> > ```shell
> > nix run '.#docs'
> > ```
> Thanks, that worked!

Odd. What shell do you use? Both bash and fish support unquoted `.#docs` correctly.

I'm not convinced we actually want to update our contributing guide; the existing unquoted style matches how flake-commands are documented almost everywhere.

belamu:
I am using zsh (with grml-config).

Concerning the contribution guide, I would like it to be zsh compatible (without getting incompatible with other shells) but I understand your point and wouldn't insist.

`#` seems to be used for getting the length of arrays and strings. But I don't know how it is understood in our case.

Your help will surely help me in the future!

MattSturgeon (approved):
I cleaned up the commit history a bit. SGTM

PR #4484, opened 2026-07-06
modules/ui_two: init

Guelakais:
I've added `modules/ui_two.nix` to enable the neovim ui_two features through the nixvim declaration. That's my first Pull request and according to the contribution guidelines it seems fine.

Closes https://github.com/nix-community/nixvim/discussions/4477

### DIFF
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -52,6 +52,12 @@
     githubId = 51029895;
     name = "GGORG";
   };
+  guelakais = {
+    email = "koroyeldiores@gmail.com";
+    github = "Guelakais";
+    githubId = 76840985;
+    name = "GueLaKais";
+  };
   JanKremer = {
     email = "mail@jankremer.eu";
     matrix = "@jankremer:matrix.org";
--- a/lib/maintainers.nix
+++ b/lib/maintainers.nix
@@ -52,12 +52,6 @@
     githubId = 51029895;
     name = "GGORG";
   };
-  guelakais = {
-    email = "koroyeldiores@gmail.com";
-    github = "Guelakais";
-    githubId = 76840985;
-    name = "GueLaKais";
-  };
   JanKremer = {
     email = "mail@jankremer.eu";
     matrix = "@jankremer:matrix.org";
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -24,6 +24,7 @@
     ./output.nix
     ./performance.nix
     ./plugins.nix
+    ./ui_two.nix
     ./wrappers.nix
   ];
 }
--- a/modules/ui_two.nix
+++ b/modules/ui_two.nix
@@ -9,7 +9,7 @@ in
 {
   options.ui2.enable = lib.mkEnableOption "Neovim's [experimental UI2](https://neovim.io/doc/user/lua/#ui2)";
 
-  config = lib.mkIf (config.ui_two.enable) {
+  config = lib.mkIf cfg.enable {
     extraConfigLua = ''
       require('vim._core.ui2').enable(${lib.nixvim.toLuaObject cfg.settings})
     '';
--- a/modules/ui_two.nix
+++ b/modules/ui_two.nix
@@ -3,6 +3,9 @@
   config,
   ...
 }:
+let
+  cfg = config.ui2;
+in
 {
   options.ui_two.enable = lib.mkOption {
     type = lib.types.bool;
--- a/modules/ui_two.nix
+++ b/modules/ui_two.nix
@@ -13,7 +13,7 @@
 
   config = lib.mkIf (config.ui_two.enable) {
     extraConfigLua = ''
-      require('vim._core.ui2').enable()
+      require('vim._core.ui2').enable(${lib.nixvim.toLuaObject cfg.settings})
     '';
   };
 }
--- a/modules/ui_two.nix
+++ b/modules/ui_two.nix
@@ -7,12 +7,7 @@ let
   cfg = config.ui2;
 in
 {
-  options.ui_two.enable = lib.mkOption {
-    type = lib.types.bool;
-    description = "enables neovim ui2 features";
-    default = false;
-    example = true;
-  };
+  options.ui2.enable = lib.mkEnableOption "Neovim's [experimental UI2](https://neovim.io/doc/user/lua/#ui2)";
 
   config = lib.mkIf (config.ui_two.enable) {
     extraConfigLua = ''
--- a/modules/ui_two.nix
+++ b/modules/ui_two.nix
@@ -1,0 +1,19 @@
+{
+  lib,
+  config,
+  ...
+}:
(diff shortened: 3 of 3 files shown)


modules/ui_two.nix
```diff
@@ -0,0 +1,19 @@
...
+  options.ui_two.enable = lib.mkOption {
+    type = lib.types.bool;
+    description = "enables neovim ui2 features";
+    default = false;
+    example = true;
+  };
+
+  config = lib.mkIf (config.ui_two.enable) {
+    extraConfigLua = ''
+      require('vim._core.ui2').enable()
```
MattSturgeon: Looking at https://neovim.io/doc/user/lua/#ui2 , this `enable()` method also accepts a config table.

We can add a `settings` option (alongside `enable`) of type `lib.nixvim.lua-types.tableOf lib.nixvim.lua-types.anything` and then include it here as


```suggestion
      require('vim._core.ui2').enable(${lib.nixvim.toLuaObject cfg.settings})
```

modules/ui_two.nix
```diff
@@ -0,0 +1,19 @@
+{
+  lib,
+  config,
+  ...
+}:
+{
```
MattSturgeon, line 9:
```suggestion
let
  cfg = config.ui2;
in
{
```

modules/ui_two.nix
```diff
@@ -0,0 +1,19 @@
...
+  options.ui_two.enable = lib.mkOption {
+    type = lib.types.bool;
+    description = "enables neovim ui2 features";
+    default = false;
+    example = true;
+  };
```
MattSturgeon: 1. This can use `mkEnableOption`
2. Let's rename `ui_two` -> `ui2` to match Neovim

```suggestion
  options.ui2.enable = lib.mkEnableOption "Neovim's [experimental UI2](https://neovim.io/doc/user/lua/#ui2)";
```

modules/ui_two.nix
```diff
@@ -0,0 +1,19 @@
...
+
+  config = lib.mkIf (config.ui_two.enable) {
```
MattSturgeon:
```suggestion
  config = lib.mkIf cfg.enable {
```

modules/ui_two.nix
MattSturgeon: Typically settings options in Nixvim are created via `lib.nixvim.mkSettingsOption`: https://github.com/nix-community/nixvim/blob/4fd45857deecb90d2732c87e0315daa48505515e/lib/options.nix#L318-L325

By default that uses `lib.nixvim.lua-types.anything` instead of `lib.nixvim.lua-types.tableOf lib.nixvim.lua-types.anything`, though. Maybe we need a way of telling `mkSettingsOption` that the type must be a table?

modules/ui_two.nix
MattSturgeon, line 1: Could you add some tests asserting that this does what we expect?

Ideally use `extraConfigLuaPost` to do actual runtime assertions; you can fail an assertion by using lua's `print()` to write to stdout.

We can also do eval-time assertions using the normal `assertions` option, but the most you can do here is test that `extraConfigLua` contains the expected substrings (e.g. with `builtins.match` or `lib.hasInfix`)

See examples in `tests/test-sources/modules`.

lib/maintainers.nix
```diff
@@ -52,6 +52,12 @@
     githubId = 51029895;
     name = "GGORG";
   };
+  guelakais = {
+    email = "koroyeldiores@gmail.com";
+    github = "Guelakais";
+    githubId = 76840985;
+    name = "GueLaKais";
+  };
```
MattSturgeon:
```suggestion
```

> maintainers-test:
> 1 Nixvim maintainers are also nixpkgs maintainers:
> - guelakais

MattSturgeon (changes requested):

modules/ui_two.nix
Guelakais: Even If you write tests, how would you test them? `nix build .` in the source root produces:
```nix
error (ignored): SQLite database '/home/ulysses/.cache/nix/eval-cache-v6/9b091a8e0dbea87d7dd4b8aa7602fbd1a4677bc4a5b6a990b94cd997ecb717a0.sqlite' is busy
error: flake 'git+file:///home/ulysses/projects/nix/nixvim_source' does not provide attribute 'packages.x86_64-linux.default' or 'defaultPackage.x86_64-linux'
```

MattSturgeon: `nix develop . --command tests -i` will run our `tests` command from our dev-shell, and let you search for the test you added.

modules/ui_two.nix
```diff
@@ -0,0 +1,17 @@
...
+}:
+let
+  cfg = config.ui2;
+in
+{
+  options.ui2.enable = lib.mkEnableOption "Neovim's [experimental UI2](https://neovim.io/doc/user/lua/#ui2)";
+
+  config = lib.mkIf cfg.enable {
+    extraConfigLua = ''
+      require('vim._core.ui2').enable(${lib.nixvim.toLuaObject cfg.settings})
```
khaneliman, line 14: issue (blocking): `cfg.settings` is referenced but `options.ui2` only declares `enable`.

Any config with `ui2.enable = true` fails with `attribute 'settings' missing`. Declare the option, e.g. `options.ui2.settings = lib.nixvim.mkSettingsOption { ... }` as suggested in https://github.com/nix-community/nixvim/pull/4484#discussion_r3536757956, and add a test under `tests/test-sources/modules/` that enables the module so CI catches this.

khaneliman: note: `vim._core.ui2` only exists on Neovim 0.12+ (0.11 shipped this as `vim._extui`).

Enabling the module on an older package fails at runtime inside init.lua with no eval-time signal. Consider an assertion or warning gated on the configured Neovim version.

modules/default.nix
```diff
@@ -24,6 +24,7 @@
     ./output.nix
     ./performance.nix
     ./plugins.nix
+    ./ui_two.nix
```
khaneliman, line 27: suggestion: the option namespace was renamed to `ui2` but the file, this import, and the commit/PR titles still say `ui_two`.

Rename to `modules/ui2.nix`, update this import, and retitle as `modules/ui2: init`. While at it, squash the six commits (four titled "Update modules/ui_two.nix") into one atomic init commit per CONTRIBUTING.md.

khaneliman (changes requested):

modules/ui_two.nix
khaneliman: issue: ui2 enable path references an undeclared settings option

When `ui2.enable = true`, this evaluates `cfg.settings`, but `options.ui2` declares only `enable`. Nix aborts with `attribute 'settings' missing`, so the feature cannot produce a configuration. Declare `ui2.settings` with a default, or call `enable()` without a settings argument.

PR #4485, opened 2026-07-07
docs: use GFM alert patches from Nixpkgs PR

MattSturgeon:
Replace our `gfm-alerts-to-admonitions` plugin and `nixos-render-docs` patches with patches from the upstream Nixpkgs PR https://github.com/NixOS/nixpkgs/pull/538502

This allows us to test out that PR in practice, as the intended target audience, before it is merged. It also allows us to drop our custom in-tree implementation early.

If/when the upstream PR merges and makes it to `nixpkgs-unstable`, it will be trivial to drop the patches from the `docs/pkgs.nix` overlay, as this PR will have already handled other cleanup & migration.

Built the patched `nixos-render-docs` locally. Will run `docs` shortly, to verify no visual changes.


MattSturgeon:
> Will run `docs` shortly, to verify no visual changes.

Done, SGTM.

image

khaneliman (approved):

khaneliman:
> This allows us to test out that PR in practice, as the intended target audience, before it is merged. It also allows us to drop our custom in-tree implementation early.

Now that it's already merged, won't be much of a wait before removing this. But, doesn't hurt, I guess.

MattSturgeon:
Yeah. The other advantage to merging now is it will make the eventual nixpkgs-bump much simpler; it'll only need to drop the `nixos-render-docs` overlay, rather than do any other cleanup or migration.

PR #4496, opened 2026-07-12
Remove GFM alert patches

korrat:
PR #4485 added an overlay applying GFM alerts patches from NixOS/nixpkgs#538502 to `nix-render-docs`. With the `nixpkgs` PR merged and propagated to `nixpkgs-unstable`, this fails due to trying to apply the patches twice.

Therefore, this change updates the `nixpkgs` input to get the updated `nix-render-docs` and drops the patches from the overlay.

### DIFF
--- a/docs/pkgs.nix
+++ b/docs/pkgs.nix
@@ -17,17 +17,6 @@ let
 
   overlay = final: prev: {
     lib = prev.lib.extend libOverlay;
-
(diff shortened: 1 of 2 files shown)


MattSturgeon:
We will handle this in the automated nixpkgs bump PR

If this is affecting you a temporary solution is `enableMan = false`

PR #4498, opened 2026-07-13
readme: remove nixpkgs follows recommendation

### LINKED ISSUE #4367: [DOCS] Inconsistency in flake input instructions between README and documentation
There's an inconsistency between the repository README and the documentation on whether to use the `inputs.nixpkgs.follows` argument in the flake installation of nixvim:

- [The nixvim README flake input instructions as of May 31 2026](https://github.com/nix-community/nixvim/blob/e5c7b40dc569f5c97ba2182d409f0fb54c02d7c1/README.md?plain=1#L156): shows an example using `follows`
- The [nixvim documentation on flake input instructions](https://nix-community.github.io/nixvim/user-guide/install.html#accessing-nixvim): explains that `follows` should not be used

Semak12345:
Closes #4367.

Removes the README example that makes Nixvim's Nixpkgs input follow the system input. Adds a short explanation and links to the installation guide, matching the project's tested-revision recommendation.

The standalone example is intentionally unchanged: it makes the top-level Nixpkgs input follow Nixvim's tested revision, which is the opposite direction.

Checks:
- git diff --check
- installation guide link returns HTTP 200

This is a documentation-only change. Nix-based checks were not run locally because Nix is not installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,11 +152,14 @@ flakes, just add the `nixvim` input:
     url = "github:nix-community/nixvim";
     # If you are not running an unstable channel of nixpkgs, select the corresponding branch of Nixvim.
     # url = "github:nix-community/nixvim/nixos-26.05";
-
-    inputs.nixpkgs.follows = "nixpkgs";
   };
 }
 ```
+
+Nixvim is tested against its own Nixpkgs revision. We recommend not overriding
+its `nixpkgs` input with `follows`; see the
+[installation guide](https://nix-community.github.io/nixvim/user-guide/install.html#accessing-nixvim)
+for details.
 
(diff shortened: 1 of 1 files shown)


MattSturgeon (approved):
Thanks for resolving the inconsistency and improving the docs!

I think there's a separate conversation to be had about _whether_ Nixvim still needs to be so insistent about using its pinned nixpkgs, but so far the most discussion we've had was this matrix thread: https://matrix.to/#/!iBrKeqhFRZaBBGodLH:matrix.org/$uaxxPgYyzHtcjhsdOi9uNtwGD5Fc_x0wG7phyt-Eo8A

PR #4503, opened 2026-07-17
plugins/treesitter: improve Nixvim's native tree-sitter setup

### LINKED ISSUE #4458: [Feature] Allow disabling treesitter indentation per filetype
It would be useful to be able to allow disabling tree-sitter indentation per filetype, similar to how highlighting can be disabled by filetype (https://github.com/nix-community/nixvim/issues/4163). Some filetypes with less developed grammars (in my case Puppet), currently have broken indentation which means I have to use an "all or nothing" approach, or manually set `indentexpr` with an autocmd.

stasjok:
Recently I switched from NixOS 25.11 to 26.05 and noticed significant limitations due to switching to nvim-treesitter's main branch.

This PR tries to lift these limitations. The intent was to bring more customizations
that were lost after switching from master branch. In particular:

- `highlight.disable` option now accepts a raw function compatible with
  disable parameter from master branch. Additionally it accepts a third
  filetype parameter
- add `highlight.enableVimSyntax` option to enable :h syntax for chosen
  filetypes. Syntax is sometimes needed for indent files, some vim
  scripts, or to improve highlighting for compopund filetypes
- `indent` and `folding` now also have `disable` support
- all features are now enabled only when the parser and queries are
  available for current filetype. This avoids setting 'indentexpr' and
  'foldexpr' when they cannot work because of missing parser or queries
- adds commands to `b:undo_ftplugin` variable to undo options set. The
  content of this variable is automatically executed by Nvim when
  switching filetype (by ftplugin). This is needed so that options were
  reset if the filetype is changed. Previously stale options were left.

Development was done in my dotfiles, and not in Nixvim repo, so there is no dev history. Everything is in one commit. I didn't done a thorough functional tests (like I did with performance modules), but added a couple of assertions based on tests that are already here. They test only the presence of relevant config lines, not absence or actual functionality. But I tested it in my dotfiles.

I think it can close #4458 too.

Technically I think `disable` now compatible enough with the old option from settings that old option can be used directly. Previous behavior was to only show a warning, but the value isn't used. I didn't change that behavior.

### DIFF
--- a/plugins/by-name/treesitter/default.nix
+++ b/plugins/by-name/treesitter/default.nix
@@ -373,6 +373,9 @@ lib.nixvim.plugins.mkNeovimPlugin {
                   return
                 end
                 local has_parser = vim.treesitter.language.add(lang)
+                if not has_parser then
+                  return
+                end
 
                 local function has_query(query_name)
                   return vim.treesitter.query.get(lang, query_name) ~= nil
@@ -398,7 +401,7 @@ lib.nixvim.plugins.mkNeovimPlugin {
 
               ${optionalString highlightEnabled ''
                 local disabled_highlight = ${lib.nixvim.toLuaObject cfg.highlight.disable}
-                if has_parser and has_query('highlights') and not is_disabled(disabled_highlight) then
+                if has_query('highlights') and not is_disabled(disabled_highlight) then
                   vim.treesitter.start(buf, lang)
                   add_undo_ftplugin(("call v:lua.vim.treesitter.stop(%d)"):format(buf))
 
@@ -409,13 +412,13 @@ lib.nixvim.plugins.mkNeovimPlugin {
                 end
               ''}${optionalString indentEnabled ''
                 local disabled_indent = ${lib.nixvim.toLuaObject cfg.indent.disable}
-                if has_parser and has_query('indents') and not is_disabled(disabled_indent) then
+                if has_query('indents') and not is_disabled(disabled_indent) then
                   vim.bo[buf].indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
                   add_undo_ftplugin('setlocal indentexpr<')
                 end
               ''}${optionalString cfg.folding.enable ''
                 local disabled_folding = ${lib.nixvim.toLuaObject cfg.folding.disable}
-                if has_parser and has_query('folds') and not is_disabled(disabled_folding) then
+                if has_query('folds') and not is_disabled(disabled_folding) then
                   vim.wo[0][0].foldexpr = 'v:lua.vim.treesitter.foldexpr()'
                   vim.wo[0][0].foldmethod = 'expr'
                   add_undo_ftplugin('setlocal foldexpr< foldmethod<')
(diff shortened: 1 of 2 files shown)


khaneliman:
I'll try to review later, but have you tried something like https://github.com/nix-community/nixvim/pull/4491 that was designed specifically to add features that were missing from the branch change? I'd rather not inline a whole plugin configuration for things that we can already leverage elsewhere.

khaneliman:
Actually, might prefer doing this in house instead of using the other module just because upstream is recommending also leveraging the, now upstreamed, api more.

plugins/by-name/treesitter/default.nix
```diff
@@ -309,34 +360,67 @@ lib.nixvim.plugins.mkNeovimPlugin {
...
-                  vim.bo[args.buf].indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
-                ''}
+                end
+              ''}${optionalString indentEnabled ''
+                local disabled_indent = ${lib.nixvim.toLuaObject cfg.indent.disable}
+                if has_parser and has_query('indents') and not is_disabled(disabled_indent) then
+                  vim.bo[buf].indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
+                  add_undo_ftplugin('setlocal indentexpr<')
+                end
+              ''}${optionalString cfg.folding.enable ''
```
khaneliman, line 441: issue (blocking): keep existing folding behavior with the legacy package

Before this PR, `cfg.folding.enable` created an autocmd outside the `has_configs_module` check, so it ran with both modern and legacy nvim-treesitter. This PR removes that autocmd and recreates folding only inside the `else` branch; because `nvim-treesitter-legacy` provides `nvim-treesitter.configs`, it takes the `if has_configs_module` branch and never reaches this folding code. Existing configurations using `package = pkgs.vimPlugins.nvim-treesitter-legacy` with `folding.enable = true` therefore lose folding; please keep folding setup reachable in both branches until we remove support for legacy branch. (Unless we want to just tackle that now, I suppose. We really should be dropping that upstream soon, tbh. )

stasjok: Right. That was my oversight. I though that I was doing an optimization by moving it there.

I see three possibility:
1. Create a similar `vim.api.nvim_create_autocmd` in has_configs_module branch (there is more boilerplate, but it's most optimized solution, no extra work done in any version of nvim-treesitter)
2. Restore an old autoCmd, but add a condition similar to the warnings (`lib.hasInfix "nvim-treesitter-legacy" (lib.getName cfg.package)`). It should work unless user is providing some custom nvim-treesitter package
3. Drop `folding.disable` option. Technically there wasn't such option in legacy nvim-treesitter, because folding was provided only as a function, not as module. I did it here, because I wanted to be consistent for all options. In that case autoCmd can be restored. But need to add extra conditions (has parser/query), because I think it's important.

stasjok: Done 1 for now. When it is decided to drop support for nvim-treesitter-legacy, it'll be easy to just delete an if condition.

stasjok:
> I'd rather not inline a whole plugin configuration for things that we can already leverage elsewhere.

I think that implementation is simple enough and it's using a features integrated in Nvim (except for indent) and it's as recommended to set it up (except maybe for folding due to it be window local option, so it doesn't work properly when the buffer is displayed in multiple windows simultaneously, in that case it should loop over all windows and set the option for each). Most of the added staff is extra conditions, which I think should be there (checks for query, parser). Maybe in the feature there will be more definitive setup in Nvim, like it became with `vim.lsp`, but for now some boilerplate is required, and I like that Nix can handle this.

khaneliman (approved):

PR #4512, opened 2026-07-30
plugins/hover: init

libewa:
Add a simple module for [hover.nvim](https://github.com/lewis6991/hover.nvim).

### Should there be an option to automatically set keybinds, e.g.:
```nix
{
  plugins.hover.keybinds.open.key = "K";
}
```
setting:
```nix
{
  keymaps = [{
    key = "K";
    action = lib.mkRaw "require('hover').open()";
    options.desc = "hover.nvim (open)";
  }];
}
```

### Should there be automatic setup of the mouse?
Either through some kind of global toggle (which should be coordinated with modules such as `scrollview`, or through `plugins.hover.enableMouse`.
Should this option just add a keybind for `<MouseMove>`, or should it also enable `opts.mousemoveevent`? `scrollview` does the former (or rather, the plugin does it), but not the latter.

### DIFF
--- a/plugins/by-name/hover/default.nix
+++ b/plugins/by-name/hover/default.nix
@@ -1,0 +1,36 @@
+{ lib, ... }:
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "hover";
+  package = "hover-nvim";
+  url = "https://github.com/lewis6991/hover.nvim";
+  setup = ".config";
+
+  description = /* markdown */ ''
+    General framework for context aware hover providers (similar to vim.lsp.buf.hover).
+    You should also create keybinds for:
+    - `require("hover").open()`
+    - `require("hover").enter()`
+    - `require("hover").switch()`
+    - `require("hover").switch("previous")`
+    - `require("hover").switch("next")`
+
+    Also, create a bind from `<MouseMove>` to `require("hover").mouse()` and set `opts.mousemoveevent = true`.
+  '';
+
+  maintainers = [ lib.maintainers.libewa ];
+
+  settingsExample = {
+    providers = [
+      "hover.providers.diagnostic"
+      "hover.providers.lsp"
+      "hover.providers.dap"
+      "hover.providers.man"
+      "hover.providers.dictionary"
+    ];
+    preview_opts.border = "single";
+    preview_window = false;
+    title = true;
+    mouse_providers = [ "hover.providers.lsp" ];
+    mouse_delay = 1000;
+  };
+}
--- a/plugins/by-name/hover/default.nix
+++ b/plugins/by-name/hover/default.nix
@@ -1,8 +1,42 @@
-{ lib, ... }:
+{ lib, config, ... }:
+let
+  keymapType = lib.nixvim.keymaps.mkMapOptionSubmodule { action = false; };
+  hoverKeymap =
+    { description, exampleKey }:
+    lib.mkOption {
+      type = lib.types.nullOr keymapType;
+      default = null;
+      inherit description;
+      example = {
+        key = exampleKey;
+        mode = "n";
+      };
+    };
+  hoverKeymapEntry =
+    {
+      action,
+      function ? "${action}()",
+    }:
+    let
+      mapping = config.plugins.hover.keymaps.${action};
+    in
+    lib.mkIf (mapping != null) (
+      lib.mkMerge [
+        mapping
+        {
+          action = lib.nixvim.mkRaw /* lua */ ''
+            function()
+              require('hover').${function}
+            end
+          '';
+          options.desc = lib.mkDefault "hover.nvim (${function})";
+        }
+      ]
+    );
+in
 lib.nixvim.plugins.mkNeovimPlugin {
   name = "hover";
   package = "hover-nvim";
-  url = "https://github.com/lewis6991/hover.nvim";
   setup = ".config";
 
   description = /* markdown */ ''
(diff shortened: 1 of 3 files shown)


tests/test-sources/plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,25 @@
+{
+  empty = {
+    plugins.hover.enable = true;
+  };
+
+  defaults = {
```
GaetanLepage, line 6: Can you also add a test case named `example` where you set the `plugins.hover.keymaps` option?

plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,114 @@
...
+    keymaps = {
+      open = hoverKeymap {
+        description = "A keymap to open the hover window.";
+        exampleKey = "K";
+      };
+      enter = hoverKeymap {
+        description = "A keymap to enter the hover window for scrolling.";
+        exampleKey = "gK";
+      };
+      previous = hoverKeymap {
```
khaneliman, line 80: issue: fix malformed key notation

This example is missing its closing angle bracket. Use `<C-p>`.

khaneliman (commented):
Top-level keymap and mouse options fit Nixvim integration policy. One example typo remains, and existing review asks for integration coverage.

plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,114 @@
...
+    lib.mkIf (mapping != null) (
+      lib.mkMerge [
+        mapping
+        {
+          action = lib.nixvim.mkRaw /* lua */ ''
+            function()
+              require('hover').${function}
+            end
+          '';
+          options.desc = lib.mkDefault "hover.nvim (${function})";
```
khaneliman, line 32: issue: When a user sets a keymap without a description, it gets no description instead of the fallback defined here.

Nix has already filled in `mapping.options.desc = null` before this merge. When `mapping` is merged again, that `null` takes priority over `mkDefault`, so the fallback is ignored.

Remove the null options before merging. This lets the fallback apply when no description was supplied, while keeping any description the user explicitly set.

```diff
--- a/plugins/by-name/hover/default.nix
+++ b/plugins/by-name/hover/default.nix
@@
-        mapping
+        (mapping // { options = lib.filterAttrs (_: value: value != null) mapping.options; })
```

Checked: focused `nix eval` of the hover module; the default changes from `null` to `"hover.nvim (open())"`, and a custom description remains unchanged.

plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,114 @@
...
+        action = "previous";
+        function = "switch('previous')";
+      })
+      (hoverKeymapEntry {
+        action = "next";
+        function = "switch('next')";
+      })
+      (lib.mkIf config.plugins.hover.enableMouse {
+        key = "<MouseMove>";
+        action = lib.nixvim.mkRaw "require('hover').mouse";
```
khaneliman, line 105: issue: With `enableMouse` and lz-n deferred loading, this eager `require` fails at startup before hover is on the runtimepath. Defer it until the mapping runs, as with the other keymaps.

```suggestion
        action = lib.nixvim.mkRaw "function() require('hover').mouse() end";
```

Checked: headless Neovim with the packaged hover.nvim; the original raises `module 'hover' not found`. The replacement registers without loading hover, and its callback runs after the plugin loads.

plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,114 @@
...
+        options.remap = true;
```
khaneliman: suggestion: `remap = true` has no effect for a Lua-function rhs, and the entry omits `mode`, so it defaults to `""` (normal, visual, select, operator-pending) while upstream maps `<MouseMove>` in normal mode only.

Drop `options.remap` and add `mode = "n"` to match the hover.nvim README.

plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,114 @@
...
+    );
+in
+lib.nixvim.plugins.mkNeovimPlugin {
+  name = "hover";
+  package = "hover-nvim";
+  setup = ".config";
+
+  description = /* markdown */ ''
+    General framework for context aware hover providers (similar to vim.lsp.buf.hover).
+    You should also create keybinds for:
```
khaneliman: suggestion: this description still tells users to create the keybinds and set `mousemoveevent` by hand, which contradicts the `keymaps` and `enableMouse` options added in the second commit.

Rewrite it to point at those options, and drop the bare `require("hover").switch()` entry since `switch` requires a `'previous'|'next'` argument.

tests/test-sources/plugins/by-name/hover/default.nix
```diff
@@ -0,0 +1,25 @@
...
+          "hover.providers.dictionary"
+        ];
+        preview_opts.border = "single";
+        preview_window = false;
+        title = true;
+        mouse_providers = [ "hover.providers.lsp" ];
+        mouse_delay = 1000;
+      };
+    };
+  };
```
khaneliman, line 24: issue: The new `example` never enables hover, so it generates neither keymap nor `mousemoveevent`. Enabling it also exposes a type error: `keymaps.open` requires an attribute set, not `"K"`.

```diff
--- a/tests/test-sources/plugins/by-name/hover/default.nix
+++ b/tests/test-sources/plugins/by-name/hover/default.nix
@@
   example = {
     plugins.hover = {
-      keymaps.open = "K";
+      enable = true;
+      keymaps.open = { key = "K"; mode = "n"; };
       enableMouse = true;
```

Checked: focused `nix eval`; the original produces no mappings, and enabling it alone fails type checking. This replacement generates `K`, `<MouseMove>`, and `mousemoveevent = true`.

PR #4514, opened 2026-08-01
modules/lps/servers/tinymist: add some more documentation

belamu:
Based on the very helpful PR #4510 this adds some information about the tinymist language server that I would have found helpful when configuring my nixvim.

I wanted to reference the nixvim documentation of the plugin `tinymist-preview` that provides a preview for Typst documents and has tinymist as a dependency. Currently I just added the web url but this is surely not a desired way of linking the typst-preview documentation since it depends on the docs being hosted where they are currently hosted. Is there a good way or should the link just be removed and the user can navigate by themselves to the needed page?

# Testing

```
nix fmt
nix build '.#checks.x86_64-linux.docs-options' '#checks.x86_64-linux.generated' --no-link
nix develop --command tests modules-lsp
```

based on the testing in #4510 .

### DIFF
--- a/modules/lsp/servers/packages.nix
+++ b/modules/lsp/servers/packages.nix
@@ -274,6 +274,15 @@
         };
       };
     };
+
+    zuban = {
+      config = {
+        extraDescription = ''
+          As described in the [documentation](https://docs.zubanls.com/en/latest/usage.html#configuration) Zuban is configured in the mypy or `pyproject.toml` configuration file of the project, not globally.
+        '';
+        example = { };
+      };
+    };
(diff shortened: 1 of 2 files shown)


belamu:
Maybe @khaneliman is willing to review since you introduced the possibility for these docs?

khaneliman (approved):

MattSturgeon:
> Maybe @khaneliman is willing to review

Sorry, this slipped into my backlog as I wasn't able to review when I first saw this!

Thanks for bumping the PR and thanks @khaneliman for handling the review/merge.

PR #4519, opened 2026-08-14
modules/lsp: add `codelens` option

k2on:
Adds the [codelens](https://neovim.io/doc/user/lsp/#lsp-codelens) options

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -15,6 +15,15 @@ in
     inlayHints = {
       enable = lib.mkEnableOption "inlay hints globally";
     };
+    codelens = {
+      enable = lib.mkEnableOption null // {
+        description = ''
+          Whether to enable codelens globally.
+
+          See [`:h lsp-codelens`](https://neovim.io/doc/user/lsp/#lsp-codelens)
+        '';
+      };
+    };
   };
 
   imports = [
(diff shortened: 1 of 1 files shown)


modules/lsp/default.nix
```diff
@@ -15,6 +15,9 @@ in
     inlayHints = {
       enable = lib.mkEnableOption "inlay hints globally";
     };
+    codelens = {
+      enable = lib.mkEnableOption "codelens globally";
```
MattSturgeon: The same could be said for the existing `inlayHints` option, but I wonder if we should reference `:h lsp-codelens` and/or link to https://neovim.io/doc/user/lsp/#lsp-codelens ?

If so, we'd need to shadow the option's `description` as something like:

```nix
mkEnableOption null // {
  description = ''
    Whether to enable codelens globally.

    See [`:h lsp-codelens`](https://neovim.io/doc/user/lsp/#lsp-codelens)
  '';
}
```

MattSturgeon (approved):
LGTM, thanks!

modules/lsp/default.nix
k2on: Ok, I updated both descriptions.

MattSturgeon (approved):

PR #4527, opened 2026-08-22
modules/performance: match vimGenDocHook by name

stasjok:
Current implementation has a problem: if the user is using a plugin from
a different nixpkgs set, vimGenDocHook isn't filtered out correctly,
because it doesn't match literally. In order to handle this, match
vimGenDocHook by name.


khaneliman:
I believe this explains the issue I had run into numerous times that confused me... thanks for looking into that. I was unsure why some flake input updates would cause these docs/tags errors from the combination.

khaneliman (approved):

PR #4532, opened 2026-08-24
lib/plugins: expose callSetup for all defaults

khaneliman:
PR #4258 added the override but only exposed it when the built-in default was not `false`.

This prevented users from forcing setup generation for modules that support Lua configuration. The built-in value should control fallback behavior, not option availability.


lib/plugins/mk-neovim-plugin.nix
```diff
@@ -94,28 +94,28 @@ let
...
-          lib.optionalAttrs (defaultCallSetup != false) {
-            callSetup = lib.mkOption {
-              type = with lib.types; nullOr bool;
-              default = null;
-              description = ''
-                Whether to generate the standard `require('${moduleName}')${setup}(...)` call for this plugin.
+        // lib.optionalAttrs hasLuaConfig {
+          callSetup = lib.mkOption {
+            type = with lib.types; nullOr bool;
+            default = null;
```
MattSturgeon: Why not expose our internal behaviour as the default value of the public option?

I.e.
```nix
default =
  if callSetup == "optional"
    # move implementation here
  else
    callSetup;
${if callSetup == "optional" then "defaultText" else null} = lib.literalMD ''
  Markdown explanation of 'optional' behaviour.
'';
```

khaneliman: Done in d19ecee3. The option is now a plain `bool` whose default is the computed built-in behavior, with `defaultText` describing the `"optional"` case. The `null` sentinel and the internal fallback logic are gone.

MattSturgeon (approved):
Nice simplification

PR #4536, opened 2026-08-27
modules/performance: keep plugin runtimeDeps when combining

khaneliman:
`combinePlugins` collapses its members into a single pack derivation, which dropped their `runtimeDeps`, so `autowrapRuntimeDeps` had nothing to add to the wrapper's PATH.

### DIFF
--- a/modules/top-level/plugins/mk-plugin-pack.nix
+++ b/modules/top-level/plugins/mk-plugin-pack.nix
@@ -34,6 +34,15 @@ let
       (builtins.concatMap (f: f ps))
     ];
 
+  # nixpkgs reads `plugin.runtimeDeps` in plugin-submodule.nix to suffix PATH
+  # in the neovim wrapper when `autowrapRuntimeDeps` is enabled.
+  runtimeDeps = lib.pipe pluginsToCombine [
+    (builtins.catAttrs "plugin")
+    (builtins.catAttrs "runtimeDeps")
+    builtins.concatLists
+    lib.unique
+  ];
+
   requiredLuaModules = lib.pipe pluginsToCombine [
     (builtins.catAttrs "plugin")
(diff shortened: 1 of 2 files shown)


modules/top-level/plugins/mk-plugin-pack.nix
```diff
@@ -65,7 +74,7 @@ let
           fixupPhase
         '';
         passthru = {
-          inherit python3Dependencies;
+          inherit python3Dependencies runtimeDeps;
```
MattSturgeon, line 81: Why as a passthru? Seems it'd be better to inherit an actual drv-attr, e.g. directly in this `buildEnv` call (if supported) or below in the `overrideAttrs` call.

It seems odd to me that `autowrapRuntimeDeps` would read a passthru attr at all: their usual intended purpose is to pass through metadata _without_ causing rebuilds.

MattSturgeon: Or is `runtimeDeps` _not_ used by the plugin drv and is instead used by another drv (e.g. the vim wrapper)?

MattSturgeon (approved):

modules/top-level/plugins/mk-plugin-pack.nix
khaneliman: `runtimeDeps` is read by the wrapper, not the plugin build, so passthru avoids a rebuild. `python3Dependencies` beside it has no choice at all: it's a function, and `overrideAttrs` fails with `cannot convert a function to JSON`. I checked the alternative too, as a real attribute the pack rehashes and the runtime deps land in its build closure.

`requiredLuaModules` is inconsistent with that, but I'd rather fix it in its own PR. Added a comment since the placement clearly isn't obvious.

PR #4537, opened 2026-08-27
modules/lsp: add native LSP feature toggles

khaneliman:
Adds toggles for the native `vim.lsp` features Neovim 0.12 exposes: completion, semantic tokens, document color, linked editing range, on-type formatting, and inline completion. Follows the existing `inlayHints` and `codelens` options.

Superseded by #4542, which keeps `enable` out of the generated Lua instead of
making it nullable. Only one of these two should land. Merging this closes
#4542.

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -1,6 +1,12 @@
-{ lib, config, ... }:
+{
+  lib,
+  config,
+  options,
+  ...
+}:
 let
   cfg = config.lsp;
+  opts = options.lsp;
 in
 {
   options.lsp = {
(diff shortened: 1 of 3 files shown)


modules/lsp/default.nix
```diff
@@ -30,6 +30,100 @@ in
...
+          See [`:h lsp-completion`](https://neovim.io/doc/user/lsp/#lsp-completion)
+        '';
+        example = {
+          autotrigger = true;
+        };
+      };
+    };
+    semanticTokens = {
+      enable = lib.mkOption {
+        type = lib.types.nullOr lib.types.bool;
```
MattSturgeon, line 62: I suspect these options would be easier to reason about, for both users and maintainers, if we drop the `enable` option's nullability and the `settings` option's `== {}` handling.

The `settings` option's descriptions could note something like
```md
> [!important]
> Neovim enables semantic tokens by default, however `settings` are only applied by Nixvim when `${opts.enable}` is explicitly set to `true`.
```

If we _really_ need a way to handle "enabling the Nixvim options" separately from "enabling the Neovim feature" we could use a similar pattern to the LSP server options (`enable` defaulting to `config is defined` and `activate` defaulting to `true`), but that's probably overkill in this case.

khaneliman: The nullability is intentional. Semantic tokens and document color are on by default in Neovim, so the option needs three states: leave Neovim's default alone (`null`), force on (`true`), and force off (`false`). With a plain default-false bool, `enable = false` emits nothing, and users lose the ability to turn these features off through the option. That's the main value it provides; for semantic tokens especially, `enable = true` is essentially a no-op, so a default-false bool would leave the option with nothing useful to do.

I'm happy to drop the `settings != { }` inference on `documentColor`, though: settings would only apply when `enable` is explicitly set, with a note like the one you suggested added to the description. Does that address the readability concern while keeping the off switch?

MattSturgeon: If we kept the current behaviour, I'd prefer it be implemented within the option's default, e.g.:
```nix
default = if cfg.settings == {} then null else true;
defaultText = lib.literalMD ''
  `true` when `${opts.settings}` is not empty, otherwise `null`
'';
description = ''
  ...

  When `null`, semantic tokens are not explicitly enabled/desabled or configured
  and Neovim's defaults are used.

  ...
'';
```
because at least then, the internal behaviour gets explicitly documented as the option's default.

---

However, I think that highlights how we're trying to overload the `enable` nix option to represent two things: enabling Nixvim's functionality and enabling/disabling Neovim's feature.

Here we're using `enable = null` as the equivalent of a traditional module's `enable = false`. The convention is `enable` toggles the Nixvim feature and sibling options configure the Neovim feature.

Perhaps it _should_ be two options like we did in `lsp.servers` (there we used `enable` and `activate`, but perhaps a different name fits better here)? But I understand that's a bit awkward.

MattSturgeon (approved):
I'm still not super keen on `enable` being overloaded to control _both_ Nixvim's feature (via nullability) _and_ Neovim's feature (via boolean value), but I won't block on that.

_(In an ideal world, we'd consider `enable` a reserved word, only used for toggling Nixvim features and never directly serialized into lua config for Neovim)_

Rant over, this SGTM overall. Thanks!

PR #4540, opened 2026-08-27
lib/types: add missing nvim_set_hl attrs

khaneliman:
Declares the ten `nvim_set_hl` attributes from 0.12.4's `api.txt` that the
highlight submodule lacked. The submodule is freeform so they already reached
Neovim; this puts them in the generated option reference.

Related https://github.com/nix-community/nixvim/pull/4538


khaneliman:
@MattSturgeon @GaetanLepage thoughts on maintenance debt when it comes to things that are actually neovim aligned. I know we didn't like plugins tracking options closely, but didn't know if we wanted to cover some of this stuff.

PR #4542, opened 2026-08-27
modules/lsp: add native LSP feature toggles

khaneliman:
Alternative to #4537, reshaped after the review feedback there that `enable`
should never be serialized into Neovim's config.

Each feature takes `enable`, whether nixvim manages it, and `activate`, the
boolean passed to `vim.lsp.<feature>.enable()`. `activate` defaults to true, so
`inlayHints.enable` and `codelens.enable` emit exactly what they do today and
nothing needs deprecating later.

Only one of these two should land. Merging this closes #4537.

### DIFF
--- a/modules/lsp/default.nix
+++ b/modules/lsp/default.nix
@@ -1,6 +1,89 @@
-{ lib, config, ... }:
+{
+  lib,
+  config,
+  options,
+  ...
+}:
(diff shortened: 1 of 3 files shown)


GaetanLepage (approved):
Nice one!

PR #4546, opened 2026-08-31
treewide: convert stdenv.{isLinux,isDarwin} alias to stdenv.hostPlatform.{isLinux,isDarwin}

SuperSandro2000:


khaneliman:
Could you just reword it to something like `treewide: X`

SuperSandro2000:
changed that and also added isDarwin as I first missed that

khaneliman (approved):

PR #4556, opened 2026-09-13
plugins/lsp: fix building pylsp-mypy 0.8.0

squat:
The current pinned version of nixpkgs brings in pylsp-mypy 0.8.0, which
no longer tracks dependencies in setup.cfg but rather in pyproject.toml,
which makes the current `substituteInPlace` fail. This changes the rule
to match the patches for `rope`, `ruff` and others.

Note that the pylsp-mypy project currently has an issue with their repo
governance which has caused the module as it exists in pypi to diverge
from the code in GitHub: https://github.com/python-lsp/pylsp-mypy/issues/113.

Signed-off-by: squat <lserven@gmail.com>


plugins/lsp/language-servers/pylsp.nix
```diff
@@ -549,8 +549,7 @@ in
               {
                 pylsp_mypy = pylsp-mypy.overridePythonAttrs (old: {
                   postPatch = old.postPatch or "" + ''
-                    substituteInPlace setup.cfg \
-                      --replace-fail "python-lsp-server >=1.7.0" ""
+                    sed -i '/python-lsp-server/d' pyproject.toml
```
khaneliman: Use `pythonRemoveDeps` instead of patching either source file; it handles both 0.7.0 and 0.8.0 through wheel metadata. My earlier `--replace-fail` request was too narrow: 0.7.0 also contains `pyproject.toml`, so this `sed` succeeds without removing its dependency from `setup.cfg`.

```diff
--- a/plugins/lsp/language-servers/pylsp.nix
+++ b/plugins/lsp/language-servers/pylsp.nix
@@ -550,5 +550,3 @@
                 pylsp_mypy = pylsp-mypy.overridePythonAttrs (old: {
-                  postPatch = old.postPatch or "" + ''
-                    sed -i '/python-lsp-server/d' pyproject.toml
-                  '';
+                  pythonRemoveDeps = (old.pythonRemoveDeps or [ ]) ++ [ "python-lsp-server" ];
                 });
```

Checked: isolated `nix build` of both PyPI releases with this override and the module's existing dependency filter; both passed. Installed metadata retained `mypy` and the pylsp entry point, with no `python-lsp-server` requirement.

squat: Sounds good, I'm happy to make the change. I notice that we don't use `pythonRemoveDeps` elsewhere in the file and instead use the exact same `sed` instruction to remove the dep from pyproject.toml, e.g. for `ruff`. Should we follow up this PR with a cleanup?

khaneliman: Yeah, we should leverage proper tooling methods instead of `sed` commands, where possible.

khaneliman:
Please squash the follow up commit into the first.

khaneliman (approved):

PR #4557, opened 2026-09-14
Fix tsgo to point to typescript instead of typescript-go

ZacheryFaria:
Typescript 7 (TSGO) is now just `typescript` in mainline nixpkgs, `typescript-go` has been removed.


modules/lsp/servers/packages.nix
```diff
@@ -537,7 +537,7 @@
     tombi = "tombi";
     ts_ls = "typescript-language-server";
     ts_query_ls = "ts_query_ls";
-    tsgo = "typescript-go";
+    tsgo = "typescript";
```
protoniac: Shouldn't we also rename the server to tsc?

ZacheryFaria: good point. i think we should

modules/lsp/servers/packages.nix
```diff
@@ -537,7 +537,7 @@
     tombi = "tombi";
     ts_ls = "typescript-language-server";
     ts_query_ls = "ts_query_ls";
-    tsgo = "typescript-go";
+    tsc = "typescript";
```
khaneliman: issue (blocking): This silently drops the package default for `lsp.servers.tsgo`: evaluating `enable = true` now gives `package = null` with no warning. The pinned nvim-lspconfig still defines `tsgo`, not `tsc`; the executable rename does not require an option rename.

```suggestion
    tsgo = "typescript";
```

This needs the companion module below: TypeScript 7 installs `tsc`, but the upstream `tsgo` config still launches `tsgo`. It retains a usable `typescript-go` package on older nixpkgs, handles the throwing alias on newer revisions, and preserves project-local executable lookup and user overrides.

```diff
--- /dev/null
+++ b/modules/lsp/servers/custom/tsgo.nix
@@ -0,0 +1,35 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  legacyPackage = builtins.tryEval (pkgs.typescript-go or null);
+  command = builtins.baseNameOf (lib.getExe config.package);
+in
+{
+  config = {
+    package = lib.mkDefault (
+      if legacyPackage.success && legacyPackage.value != null then
+        legacyPackage.value
+      else
+        pkgs.typescript
+    );
+
+    config.cmd = lib.mkIf (config.package != null) (
+      lib.mkDefault (lib.nixvim.mkRaw ''
+        function(dispatchers, config)
+          local cmd = ${lib.nixvim.toLuaObject command}
+          if (config or {}).root_dir then
+            local local_cmd = vim.fs.joinpath(config.root_dir, 'node_modules/.bin', cmd)
+            if vim.fn.executable(local_cmd) == 1 then
+              cmd = local_cmd
+            end
+          end
+          return vim.lsp.rpc.start({ cmd, '--lsp', '--stdio' }, dispatchers)
+        end
+      '')
+    );
+  };
+}
```

Checked: `nix eval --impure --json` on the patched PR with its pinned nixpkgs and simulated older/missing-attribute package sets; package selection, legacy `plugins.lsp` configuration, explicit overrides, and `package = null` passed. A `nvim --headless -u NONE` probe passed PATH/project-local command selection, LSP arguments, and inheritance of upstream filetypes/root detection; it did not start a real TypeScript server.

If we rename the public option later, please use `mkRenamedOptionModule` to migrate existing definitions with a warning rather than removing the old key.

modules/lsp/servers/packages.nix
MattSturgeon: We don't define the lsp module names; we inherit them from upstream nvim-lspconfig. Attempting to rename them like this results in the following CI failure:

> lsp: The following are not declared in modules/lsp/servers/packages.nix:
> - tsgo
> lsp: The following are not listed upstream, but are declared in modules/lsp/servers/packages.nix:
> - tsc

protoniac: I think that tsc is available in [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/issues/4467)

